PR #8, opened 2016-06-02
Authentication config file for in-memory backend

DavidPineauScality:
This PR provides a new config file format for the memory backend's authentication data, as well as a number of small utilities.

Edit: Please note that the checker writes the errors into the standard error output stream, in order for the user to see them. I am unsure whether this is the best solution or not, but I wanted to do that to ensure early failure if the configuration wasn't right. The thrown error in this case might deserve improvement also. I'm open to suggestions on that side :)


lib/auth/in_memory/checker.js
```diff
@@ -0,0 +1,173 @@
...
+                data.keys[keyObj.access] = incr(data.keys[keyObj.access]);
+            });
+        }
+    }
+}
+
+function dumpCountError(property, obj) {
+    Object.keys(obj).forEach(key => {
+        if (obj[key] > 1) {
+            log.error('property should be unique', {
```
DavidPineauScality: Btw, how will it behave ? I mean, the Config.js is using the authchecker inside.

Which means that the authchecker will be called while the Config is being initialized. Isn't there a risk of stack overflow through recursion here ? I'll try it anyways.

lib/auth/in_memory/checker.js
```diff
@@ -0,0 +1,173 @@
+import Logger from 'werelogs';
+
+// Here, we expect the logger to have already been configured in S3
+const log = new Logger('S3');
```
DavidPineauScality: will do, thanks (and will check how the circular dep behaves there)

lib/auth/in_memory/checker.js
rahulreddy: I was thinking that it should be ok as the config has been parsed and the var should be available to setup the logger, I don't see a recursion happening here, I could be wrong.

DavidPineauScality: I was thinking of the following inclusion cycle:
 - Config includes checkAuth
 - checkAuth includes log utilities
 - log utilities include Config.

Then, when we instanciate the server:
1. the first file imports the configuration
2. load the configuration
3. call authdataCheck
4. imports Config.js (going back to 1.)

I didn't check yet, but I do think there is a risk here.

conf/authdata.json
```diff
@@ -0,0 +1,32 @@
+{
+    "accounts": [{
+        "name": "Bart",
```
ThibaultRiviere: Simpson ?

LaurenSpiegel: You're the first person to notice 😃

lib/auth/in_memory/checker.js
rahulreddy: I found this in node docs

> Multiple calls to require('foo') may not cause the module code to be executed multiple times. This is an important feature. With it, "partially done" objects can be returned, thus allowing transitive dependencies to be loaded even when they would cause cycles.

DavidPineauScality: Which is a good news, but there's still the point of how much can the config object be used, if it's incompletely built...
I'll test it anyways ^^' Will keep you posted

rahulreddy:
Other than the comments relating to logs, I am ok with the rest of the code.

alexandre-merle:
Updated, unit tests for checker added

alexandre-merle:
Updated with unit test for indexer

tests/unit/auth/checker.js
```diff
@@ -0,0 +1,134 @@
...
+    done();
+}
+
+function shouldSuccess(obj, done) {
+    const res = checker(obj);
+    assert.strictEqual(res, false);
+    done();
+}
+
+export const should = {
```
MichaelZapataScality: :heart_eyes:

alexandre-merle:
@ironman-machine let me know the bugs

DavidPineauScality:
@ironman-machine and now ? :D May I keep hope ?

PR #14, opened 2016-06-04
Ft/metadata file backend

vrancurel:
S3 File metadata backend.


vrancurel:
Experimental implementation that works with cluster mode

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,402 @@
+import { errors } from 'arsenal';
+import level from 'level';
```
rahulreddy: I noticed that `level` is not part of package.json, it needs to be added to the list of dependencies.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,402 @@
...
+import multilevel from 'multilevel';
+import sublevel from 'level-sublevel';
+import net from 'net';
+import BucketInfo from '../BucketInfo';
+import arsenal from 'arsenal';
+import events from 'events';
+import assert from 'assert';
+import cluster from 'cluster';
+import fs from 'fs';
+import constants from '../../../constants';
```
rahulreddy: Can you please re-order the modules in this order?
1) core modules
2) dependency modules
3) local modules

vrancurel:
updated

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,406 @@
...
+                                               new Date().toJSON(), 2);
+        metastore.put(constants.usersBucket, usersBucketAttr.serialize());
+        const stream = metastore.createKeyStream();
+        stream
+            .on('data', e => {
+                // automatically recreate existing sublevels
+                sub.sublevel(e);
+            })
+            .on('error', () => {
+                throw (errors.InternalError);
```
rahulreddy: I think it's a good idea to send Werelogs logger here and log the actual error.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,406 @@
...
+
+const METADATA_PORT = 9990;
+const METADATA_PATH = '/metadata/';
+const MANIFEST_JSON = 'manifest.json';
+const MANIFEST_JSON_TMP = 'manifest.json.tmp';
+const ROOT_DB = 'rootDB';
+const METASTORE = '__metastore';
+const OPTIONS = { sync: true };
+
+let rootDB;
```
rahulreddy: These two vars can be moved to `startServer()` method as they are not being used anywhere else.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,406 @@
...
+        sub.methods.createSub = { type: 'async' };
+        sub.createSub = (subName, cb) => {
+            try {
+                sub.sublevel(subName);
+                multilevel.writeManifest(sub,
+                                         METADATA_PATH +
+                                         MANIFEST_JSON_TMP);
+                fs.renameSync(METADATA_PATH + MANIFEST_JSON_TMP,
+                              METADATA_PATH + MANIFEST_JSON);
+                cb(null);
```
rahulreddy: You can use `cb()` instead of `cb(null)`

vrancurel:
updated

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,302 @@
...
+     *  @return {undefined}
+     */
+    internalListObject(bucketName, params, log, cb) {
+        const requestParams = {};
+        let Ext;
+        let extParams = {};
+        // multipart upload listing
+        if (params.listingType === 'multipartuploads') {
+            Ext = arsenal.listMPU.ListMultipartUploads;
+            this.addExtensionParam('queryPrefixLength', params, extParams);
```
rahulreddy: This code comes from MetaData and there is a bit of context here. I believe extParams has a list of params supported by levelDB and we want to assign/change only those parameters based on the request. Isolating that logic into its own method(even though its just one line) would be more meaningful as we can expand on it later.

lib/auth/vault.js
```diff
@@ -8,6 +8,7 @@ import backend from './in_memory/backend';
 let client;
 
 if ((process.env.S3BACKEND && process.env.S3BACKEND === 'mem')
+    || (process.env.S3BACKEND && process.env.S3BACKEND === 'file')
```
LaurenSpiegel: Please rebase.  This is out of date.

lib/data/wrapper.js
```diff
@@ -9,6 +9,7 @@ let client;
 let implName;
 
 if ((process.env.S3BACKEND && process.env.S3BACKEND === 'mem')
+    || (process.env.S3BACKEND && process.env.S3BACKEND === 'file')
```
LaurenSpiegel: This is out of date as well.  We now have a config.backends

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,450 @@
...
+import sublevel from 'level-sublevel';
+import arsenal from 'arsenal';
+import { logger } from '../../utilities/logger';
+
+import { errors } from 'arsenal';
+import BucketInfo from '../BucketInfo';
+import constants from '../../../constants';
+
+const METADATA_PORT = 9990;
+const METADATA_PATH = '/metadata/';
```
LaurenSpiegel: Please use:
const METADATA_PATH = config.filePaths.metadataPath;

This will be available in the data backend PR.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,450 @@
...
+import events from 'events';
+import assert from 'assert';
+
+import level from 'level';
+import multilevel from 'multilevel';
+import sublevel from 'level-sublevel';
+import arsenal from 'arsenal';
+import { logger } from '../../utilities/logger';
+
+import { errors } from 'arsenal';
```
RachedBenMustapha: Couldn't this be `const errors = arsenal.errors;` since `arsenal` is already imported once?

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,450 @@
...
+import fs from 'fs';
+import cluster from 'cluster';
+import events from 'events';
+import assert from 'assert';
+
+import level from 'level';
+import multilevel from 'multilevel';
+import sublevel from 'level-sublevel';
+import arsenal from 'arsenal';
+import { logger } from '../../utilities/logger';
```
RachedBenMustapha: Since this is in-tree, this import should be in the next import group.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,451 @@
...
+            return undefined;
+        });
+    }
+
+    listObject(bucketName, params, log, cb) {
+        return this.internalListObject(bucketName, params, log, cb);
+    }
+
+    listMultipartUploads(bucketName, params, log, cb) {
+        return this.internalListObject(bucketName, params, log, cb);
```
RachedBenMustapha: why is it exactly the same code as `listObject`?

RachedBenMustapha:
Just 1 comment, then 👍

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,451 @@
...
+class BucketFileInterface {
+
+    constructor() {
+        this.logger = logger;
+        if (cluster.isMaster) {
+            this.startServer();
+        } else {
+            this.refcnt = 0;
+            this.reConnect();
+            this.notifyer = new events.EventEmitter();
```
LaurenSpiegel: notifier

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,451 @@
...
+import arsenal from 'arsenal';
+
+import { logger } from '../../utilities/logger';
+import BucketInfo from '../BucketInfo';
+import constants from '../../../constants';
+
+const errors = arsenal.errors;
+
+const METADATA_PORT = 9990;
+const METADATA_PATH = '/metadata/';
```
LaurenSpiegel: When you rebase and use the metadata path from the config, note that you are going to have to add a "/" to the end here.

lib/metadata/bucketfile/backend.js
rahulreddy: listObject was a sample I believe.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,451 @@
...
+                fs.renameSync(METADATA_PATH + MANIFEST_JSON_TMP,
+                              METADATA_PATH + MANIFEST_JSON);
+                cb();
+            } catch (err) {
+                cb(err);
+            }
+        };
+        const metastore = sub.sublevel(METASTORE);
+        /* due to a chicken-and-egg issue we need to create
+           usersBucket here */
```
LaurenSpiegel: No similar issue with the mpu buckets?

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,451 @@
...
+                sub.sublevel(e);
+            })
+            .on('error', err => {
+                this.logger.fatal('error listing metastore', { error: err });
+                throw (errors.InternalError);
+            })
+            .on('end', () => {
+                multilevel.writeManifest(sub, METADATA_PATH + MANIFEST_JSON);
+                this.logger.info('starting metadata file backend server');
+                net.createServer(con => {
```
LaurenSpiegel: Perhaps a comment for what this is doing 😄

lib/metadata/bucketfile/backend.js
vrancurel: The internalListObject() will use params.listingtype to differenciate different behavior. The fact that we use a common function to deal with many cases is specific to our implementation, but could be different in some other implementations of the backend class.

DavidPineauScality:
Updated, rebased.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,453 @@
...
+     * Reconnect to the server while possibly waiting for
+     * it to be idle
+     * @return {undefined}
+     */
+    reConnect() {
+        if (this.refcnt === 0) {
+            this.realReConnect();
+        } else {
+            this.notifier.on('idle', () => {
+                this.realReConnect();
```
vrancurel: The condition might have changed since we've been notified. How to properly loop and wait over the condition ?

lib/Config.js
```diff
@@ -193,7 +193,7 @@ class Config {
         const dataPath = process.env.S3DATAPATH ?
             process.env.S3DATAPATH : 'localData';
```
LaurenSpiegel: Let's be consistent and set this one with `${_dirname}/../localData` like we are doing now for metadata.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,453 @@
...
+                                         MANIFEST_JSON_TMP);
+                fs.renameSync(METADATA_PATH + MANIFEST_JSON_TMP,
+                              METADATA_PATH + MANIFEST_JSON);
+                cb();
+            } catch (err) {
+                cb(err);
+            }
+        };
+        const metastore = sub.sublevel(METASTORE);
+        /* due to a chicken-and-egg issue we need to create
```
LaurenSpiegel: Based on our discussion, I would change this to:

Since the bucket creation API is expecting the usersBucket to have attributes, we pre-create the usersBucket.

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,453 @@
...
+        this.metastore = this.client.sublevel(METASTORE);
+    }
+
+    /**
+     * Reconnect to the server while possibly waiting for
+     * it to be idle
+     * @return {undefined}
+     */
+    reConnect() {
+        if (this.refcnt === 0) {
```
rahulreddy: per our discussion
```javascript
relReconnect(cb)
```
would take callback and notify the callee when it's done. You can make the cb conditional, if you have to.

vrancurel:
Did a big squash before some commits conflicted.

Fixed race conditions thanks to @RachedBenMustapha help !

LaurenSpiegel:
Good stuff 👍 

@ironman-machine thoughts?

rahulreddy:
👍

lib/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,520 @@
...
+        delete require.cache[require.resolve(METADATA_PATH + MANIFEST_JSON)];
+        const manifest = require(METADATA_PATH + MANIFEST_JSON);
+        this.client = multilevel.client(manifest);
+        const con = net.connect(METADATA_PORT);
+        con.pipe(this.client.createRpcStream()).pipe(con);
+        this.metastore = this.client.sublevel(METASTORE);
+    }
+
+    /**
+     * Wait for reconnect do be done
```
vrancurel: the comment should be: try to reconnect possibly waiting for other operations to complete

PR #21, opened 2016-06-07
Dev/databackendretake

LaurenSpiegel:
Implement a data file backend.

Functional tests pass.

This assumes that /data/pathNumbers exist on machine.  Do we want to create that through the Docker image and instruct in the readme or handle some other way?

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -49,4 +49,7 @@ export default {
     // ACLs.
     logId: 'http://acs.amazonaws.com/groups/s3/LogDelivery',
     emptyFileMd5: 'd41d8cd98f00b204e9800998ecf8427e',
+
+    // Number of sub-directories for file backend
+    folderHash: 3511, // Prime number
 };
--- a/init.js
+++ b/init.js
@@ -1,0 +1,73 @@
+'use strict'; // eslint-disable-line strict
+require('babel-core/register');
+
+const assert = require('assert');
+const fs = require('fs');
+const os = require('os');
+
+const async = require('async');
+const ioctl = require('ioctl');
+const constants = require('./constants').default;
+const config = require('./lib/Config.js').default;
+const logger = require('./lib/utilities/logger.js').logger;
+
+function _setDirSyncFlag(path) {
+    const GETFLAGS = 2148034049;
+    const SETFLAGS = 1074292226;
+    const FS_DIRSYNC_FL = 65536;
+    const buffer = new Buffer(8);
+    const pathFD = fs.openSync(path, 'r');
+    const status = ioctl(pathFD, GETFLAGS, buffer);
+    assert.strictEqual(status, 0);
+    const currentFlags = buffer.readUIntLE(0, 8);
+    const flags = currentFlags | FS_DIRSYNC_FL;
+    buffer.writeUIntLE(flags, 0, 8);
+    const status2 = ioctl(pathFD, SETFLAGS, buffer);
+    assert.strictEqual(status2, 0);
+    fs.closeSync(pathFD);
+    const pathFD2 = fs.openSync(path, 'r');
+    const confirmBuffer = new Buffer(8);
+    ioctl(pathFD2, GETFLAGS, confirmBuffer);
+    assert.strictEqual(confirmBuffer.readUIntLE(0, 8),
+        currentFlags | FS_DIRSYNC_FL, 'FS_DIRSYNC_FL not set');
+    logger.info('FS_DIRSYNC_FL set');
+    fs.closeSync(pathFD2);
+}
+
+if (config.backends.data !== 'file' && config.backends.metadata !== 'file') {
+    logger.info('No init required. Go forth and store data.');
+    process.exit(0);
+}
+
+const dataPath = config.filePaths.dataPath;
+const metadataPath = config.filePaths.metadataPath;
+
+fs.accessSync(dataPath, fs.F_OK | fs.R_OK | fs.W_OK);
+fs.accessSync(metadataPath, fs.F_OK | fs.R_OK | fs.W_OK);
+
+if (os.type() === 'Linux' && os.endianness() === 'LE') {
+    _setDirSyncFlag(dataPath);
+    _setDirSyncFlag(metadataPath);
+} else {
+    logger.warn('WARNING: Synchronization directory updates are not ' +
+        'supported on this platform. Newly written data could be lost ' +
+        'if your system crashes before the operating system is able to ' +
+        'write directory updates.');
+}
+
+// Create 3511 subdirectories for the data file backend
+const subDirs = Array.from({ length: constants.folderHash },
+    (v, k) => (k).toString());
+async.eachSeries(subDirs, (subDirName, next) => {
+    fs.mkdir(`${dataPath}/${subDirName}`, err => {
+        // If already exists, move on
+        if (err && err.errno !== -17) {
+            return next(err);
+        }
+        return next();
+    });
+},
+ err => {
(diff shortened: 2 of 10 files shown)


LaurenSpiegel:
Ready for final reviews.  Thanks, @RachedBenMustapha for all of the heavy lifting on the file flags!

I can drop the changes to the metadata wrapper depending on which PR gets merged first (this or the metadata file backend).

Since the ReadMe for master has diverged, I will update the readMe separately.

LaurenSpiegel:
@ironman-machine any concerns?

lib/data/file/backend.js
```diff
@@ -0,0 +1,110 @@
...
+                log.error('error streaming data from request', { error: err });
+                fs.close(fd);
+                return callback(errors.InternalError);
+            })
+            .on('end', () => {
+                // eslint-disable-next-line no-param-reassign
+                request.calculatedHash = hash.digest('hex');
+                log.debug('finished writing data', { key,
+                    calculatedHash: request.calculatedHash });
+                fs.close(fd);
```
RachedBenMustapha: Here would be a great place to issue a please-really-write-to-disk request: https://nodejs.org/docs/v4.4.4/api/fs.html#fs_fs_fsync_fd_callback

It still does not guarantee 100% that all is written to the physical device, but it gets us closer.

lib/data/file/backend.js
```diff
@@ -0,0 +1,111 @@
...
+        }
+        const rs = fs.createReadStream(filePath, readStreamOptions);
+        return callback(null, rs);
+    },
+
+    delete: function delFile(key, reqUids, callback) {
+        const log = createLogger(reqUids);
+        const filePath = getFilePath(key);
+        log.debug('deleting file', { filePath });
+        return fs.unlink(filePath, callback);
```
RachedBenMustapha: `fs.unlink()` will give an `ENOENT` error to the callback so I think the error is not silently ignored. In any case, we should be aware if the metadata backend thinks an object still exists while it's missing on disk: it meands there is either a bug in some of our code or data/fs integrity issue - and we should be aware if that's the case.

init.js
```diff
@@ -0,0 +1,71 @@
...
+    const status = ioctl(pathFD, GETFLAGS, buffer);
+    assert.strictEqual(status, 0);
+    const currentFlags = buffer.readUIntLE(0, 8);
+    const flags = currentFlags | FS_DIRSYNC_FL;
+    buffer.writeUIntLE(flags, 0, 8);
+    const status2 = ioctl(pathFD, SETFLAGS, buffer);
+    assert.strictEqual(status2, 0);
+    const confirmBuffer = new Buffer(8);
+    ioctl(pathFD, GETFLAGS, confirmBuffer);
+    assert.strictEqual(confirmBuffer.readUIntLE(0, 8),
```
RachedBenMustapha: Not a big deal, but we could arguably want to do this assert on the value retrieved from a closed and re-opened fd.

init.js
```diff
@@ -0,0 +1,71 @@
...
+} else {
+    logger.warn('WARNING: Synchronization directory updates are not ' +
+        'supported on this platform. Newly written data could be lost ' +
+        'if your system crashes before the operating system is able to ' +
+        'write directory updates.');
+}
+
+// Create 3511 subdirectories for the data file backend
+const subDirs = Array.from({ length: constants.folderHash },
+    (v, k) => (k + 1).toString());
```
RachedBenMustapha: Why aren't we zero-based?

init.js
```diff
@@ -0,0 +1,71 @@
...
+    fs.mkdir(`${dataPath}/${subDirName}`, err => {
+        // If already exists, move on
+        if (err && err.errno !== -17) {
+            return next(err);
+        }
+        return next();
+    });
+},
+ err => {
+     assert.strictEqual(err, null, `Error creating data files ${err}`);
```
RachedBenMustapha: Does `assert` work as expected (interrupting the whole program flow) when called from several layers of callbacks?

lib/data/file/backend.js
```diff
@@ -0,0 +1,110 @@
...
+
+
+function getFilePath(key) {
+    const hash = stringHash(key);
+    const folderHashPath = ((hash % FOLDER_HASH) + 1).toString();
+    return `${STORAGE_PATH}/${folderHashPath}/${key}`;
+}
+
+export const backend = {
+    put: function putFile(request, keyContext, reqUids, callback) {
```
RachedBenMustapha: I'm not a fan of leaking the HTTP request here but I guess it's not specific to this file backend. Would it work like a `ReadableStream` if we just call it `stream` or do we actually use features from the request?

RachedBenMustapha:
A few minor comments/questions, but I'm 👍 .

lib/data/file/backend.js
LaurenSpiegel: We add the calculatedHash to it like with the real data backend.  Should be cleaned up in a separate PR.

init.js
LaurenSpiegel: Yes.  Just confirmed by testing.

vrancurel:
let's use localData and localMetadata names.

lib/data/file/backend.js
```diff
@@ -1,93 +1,110 @@
 import { errors } from 'arsenal';
 import crypto from 'crypto';
-import stream from 'stream';
 import fs from 'fs';
+import config from '../../Config';
+import Logger from 'werelogs';
```
vrancurel: Please use in order:
core dependencies
nl
external dependencies
nl
internal dependencies

lib/data/file/backend.js
```diff
@@ -0,0 +1,118 @@
...
+            .on('open', () => { callback(null, rs); });
+    },
+
+    delete: function delFile(key, reqUids, callback) {
+        const log = createLogger(reqUids);
+        const filePath = getFilePath(key);
+        log.debug('deleting file', { filePath });
+        return fs.unlink(filePath, err => {
+            if (err) {
+                log.error('error deleting file', { error: err });
```
RachedBenMustapha: We could maybe include the filename in the log attributes for ops people to easily look at the issue?

constants.js
```diff
@@ -49,4 +49,7 @@ export default {
     // ACLs.
     logId: 'http://acs.amazonaws.com/groups/s3/LogDelivery',
     emptyFileMd5: 'd41d8cd98f00b204e9800998ecf8427e',
+
+    // Number of sub-directories for file backend
+    folderHash: 3511, // Prime number
```
RachedBenMustapha: Did we run tests to see which prime number yields better results depending on the filesystem type? Just kidding 😆

LaurenSpiegel: That's @GiorgioRegni 's hobby.

vrancurel:
👍

LaurenSpiegel:
@ironman-machine SVC_BASE_FORCE_BUILD=1
NODESVC_BASE_FORCE_BUILD=1

LaurenSpiegel:
@ironman-machine SVC_BASE_FORCE_BUILD=1
NODESVC_BASE_FORCE_BUILD=1

LaurenSpiegel:
@ironman-machine where are you
SVC_BASE_FORCE_BUILD=1
NODESVC_BASE_FORCE_BUILD=1

rahulreddy:
👍

LaurenSpiegel:
@ironman-machine get to it

DavidPineauScality:
@ironman-machine No sleep for the warriors, go and charge !

DavidPineauScality:
Merging on behalf of @LaurenSpiegel since End-to-End passes :) Congrats ;)

PR #36, opened 2016-06-08
Update README.md

GiorgioRegni:
Updated to public URL now that the project is open.


DavidPineauScality:
Necessary to provide a public guide to run S3.
Exceptionally merging due to the current circumstances.

PR #40, opened 2016-06-10
[Listing](ft): Set an hard limit to listing

alexandre-merle:
- Amazon aws set an hard limit to 1000
  This config can be override in constants.js to increase or
  decrease the max keys we will receive

see http://docs.aws.amazon.com/AmazonS3/latest/API/
    RESTBucketGET.html#RESTBucketGET-requests

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -52,4 +52,8 @@ export default {
 
     // Number of sub-directories for file backend
     folderHash: 3511, // Prime number
+    // Aws amazon set an hard limit to the listing maxKeys
+    // http://docs.aws.amazon.com/AmazonS3/latest/API/
+    //      RESTBucketGET.html#RESTBucketGET-requests
+    listingHardLimit: 1000,
 };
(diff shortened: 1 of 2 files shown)


rahulreddy:
It's good to have a hard limit but I don't see it being mentioned in the link you provided.

alexandre-merle:
it's in the maxkeys description, the text is 
```
Sets the maximum number of keys returned in the response body. If you want to retrieve fewer than the default 1,000 keys, you can add this to your request.

The response might contain fewer keys, but it will never contain more.
```

rahulreddy:
Ah, I think I am recollecting things a bit by bit. We had discussion about this and we didn't want to impose a hard limit as we didn't want to be strict but given how expensive listing operations are imposing the hard limit is a good idea.
Btw, your code sets the hard limit as 10K not 1K as mentioned in the docs.

alexandre-merle:
Updated :)

rahulreddy:
👍

LaurenSpiegel:
👍

LaurenSpiegel:
Please change to a PR against master and reference this PR.

PR #42, opened 2016-06-10

rahulreddy:
By default S3 will be using file backend. For in-memory usage, `S3BACKEND` env var should be `mem` and for enterprise/federation usage S3BACKEND should be defined and set to something other than mem/file.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -150,16 +150,21 @@ class Config {
                 cert: certpath,
             };
         } else if (key || cert) {
-            throw new Error( 'bad config: both certFilePaths.key and ' +
+            throw new Error('bad config: both certFilePaths.key and ' +
                 'certFilePaths.cert must be defined');
         }
         /**
          * Configure the backends for Authentication, Data and Metadata.
          */
-        let auth = '';
-        let data = '';
-        let metadata = '';
+        let auth = 'mem';
+        let data = 'file';
+        let metadata = 'file';
         if (process.env.S3BACKEND) {
+            const validBackends = ['mem', 'file', 'scality'];
+            assert(validBackends.indexOf(process.env.S3BACKEND) > -1,
+                'bad environment variable: S3BACKEND environment variable ' +
+                'should be one of mem/file/scality'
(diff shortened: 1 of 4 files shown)


LaurenSpiegel:
Once https://github.com/scality/S3/pull/14 is merged, we should test this end to end.

DavidPineauScality:
I do not see any check for a value of the backends matching the scality-specific backends ?
How should they be managed ? I would have seen a "scality" value for this, in each of the wrappers.

rahulreddy:
scality specific backends fall in the else part. As long S3BACKEND is specified and not set to mem or file, scality-specific backends are used.

rahulreddy:
@DavidPineauScality What do you think?

DavidPineauScality:
We should really avoid setting-up such a kitchen sink for a specific backend. At best, it should either yield an error or fallback to the default backend (which given the open-source aspect of the project should be the file backend).

rahulreddy:
I have updated so that config throws an error on server startup if S3BACKEND is not set to one of the valid backends. If S3BACKEND is not set the default will be file.

LaurenSpiegel:
👍

lib/Config.js
```diff
@@ -150,16 +150,21 @@ class Config {
...
+        let metadata = 'file';
         if (process.env.S3BACKEND) {
+            const validBackends = ['mem', 'file', 'scality'];
+            assert(validBackends.indexOf(process.env.S3BACKEND) > -1,
+                'bad environment variable: S3BACKEND environment variable ' +
+                'should be one of mem/file/scality'
+            );
             auth = process.env.S3BACKEND;
             data = process.env.S3BACKEND;
             metadata = process.env.S3BACKEND;
```
jmunoznaranjo: This will set everything to the same backend. Say that i want file auth and scality metadat, this won't allow it. Is taht intended?

rahulreddy: We have specific overrides defined after this section.

jmunoznaranjo:
:+1:

PR #49, opened 2016-06-13

rahulreddy:
By default S3 will be using file backend. Other valid backends are
file and scality


DavidPineauScality:
:+1: for me, thanks @rahulreddy

LaurenSpiegel:
👍

PR #50, opened 2016-06-13
Perf--gets of objects put by MPU

LaurenSpiegel:
Uses double buffering to increase the speed of gets of objects put by MPU.

Perf testing with 1GB object:

205 parts with specified part size of 5MB
-with rel/1.1:  about 13-14 seconds
-with double buffer: about 7-8 seconds

103 parts with specified part size of 10 MB
-with rel/1.1: about 15-16 seconds
-with double buffer: about 7-8 seconds

52 parts with specified part size of 20 MB
-with rel/1.1: about 12-13 seconds
-with double buffer: about 5-6 seconds

### DIFF
--- a/lib/routes/routesUtils.js
+++ b/lib/routes/routesUtils.js
@@ -1,3 +1,5 @@
+import { readySetStream } from 'ready-set-stream';
+
 import data from '../data/wrapper';
 
 /**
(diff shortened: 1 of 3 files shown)


DavidPineauScality:
Hehe, I see you've made use of your personal time there :)

By double streaming, do you mean that this module keeps reading from the source on a secondary buffer while writing to the destination using the first, already-filled one ?

alexandre-merle:
It's an interesting improvment :) my only concern is for the memory, we should perform some tests to see if that will not be a problem, personnaly, i don't think so, we have much memory :)

LaurenSpiegel:
Based on testing with s3blaster, memory use is not increased significantly:

@lamphamsy 's analysis:

1. For S3:
    - Memory consumption is similar between your branch and rel1.1
    - Your branch makes S3 consume CPU more than that of rel1.1, but the difference is not much, about 50% vs. 30% for 5MB part size.
2. For Sproxyd:
    It's almost same as S3

alexandre-merle:
fine for me so :+1:

ThibaultRiviere:
:+1:

rahulreddy:
👍

LaurenSpiegel:
@ironman-machine thoughts?

PR #56, opened 2016-06-16
fix(logs): change NoSuchBucket case log level

michael-zapata:
From `warn` to `debug`

This was also highlighted during testing.


MathieuCassagne:
LGTM :+1:

PR #65, opened 2016-06-20
[Listing](ft): Set an hard limit to listing

LaurenSpiegel:
Reopening against rel/1.1

- Amazon aws sets a hard limit to 1000
  This config can be override in constants.js to increase or
  decrease the max keys we will receive

see http://docs.aws.amazon.com/AmazonS3/latest/API/
    RESTBucketGET.html#RESTBucketGET-requests

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -52,4 +52,8 @@ export default {
 
     // Number of sub-directories for file backend
     folderHash: 3511, // Prime number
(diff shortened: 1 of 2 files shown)


LaurenSpiegel:
👍

rahulreddy:
👍

PR #69, opened 2016-06-21
docs(DESIGN): update architectural representation

### LINKED ISSUE #67: DESIGN.md is not up to date
At a quick glance, there's still an image with node simple router in it....

michael-zapata:
With node simple router a thing of the past, we can replace it with the
authorisation step

Fix #67


LaurenSpiegel:
Thanks for updating!  Instead of services, I think it might make more sense to do a "backend wrapper" layer.  Then show the possible backends (and remove any IM reference).

rahulreddy:
Minor remark, can we have the text in a contrasting color, something like white or black would look nice.

LaurenSpiegel:
I still find it a bit confusing.  Perhaps "backends" can be "backend wrappers" and then both metadata and data could be at the bottom with each broken into the different types of backends.  I don't think it is necessary to have metadata up by authorization.  In fact, I might get rid of authentication and authorization and replace them with routes and vault for authentication so it is more about the structure of components rather than process steps.

michael-zapata:
@ironman-machine Let's test with rel/1.1

michael-zapata:
@ironman-machine hello

michael-zapata:
@ironman-machine plz

PR #71, opened 2016-06-21

LaurenSpiegel:
Merging this commit (which is already in master) will prevent only one worker from being created when S3BACKEND=scality

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -100,7 +100,7 @@ class S3Server {
 
 export default function main() {
     let clusters = _config.clusters || 1;
-    if (process.env.S3BACKEND) {
+    if (process.env.S3BACKEND === 'mem') {
         clusters = 1;
     }
     if (cluster.isMaster) {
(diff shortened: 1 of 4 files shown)


LaurenSpiegel:
@ironman-machine thoughts?

lib/server.js
```diff
@@ -100,7 +100,7 @@ class S3Server {
 
 export default function main() {
     let clusters = _config.clusters || 1;
-    if (process.env.S3BACKEND) {
+    if (process.env.S3BACKEND === 'mem') {
```
DavidPineauScality: wouldn't that line be enough for rel/1.1, rather than bringing the whole default commit ?

DavidPineauScality: Note: As much as possible, I'd like to only bring the real fix required, rather than the full commit, if avoidable.

LaurenSpiegel: Yes, we could do just that one line.

PR #72, opened 2016-06-21
Specify backend variable for one worker.

LaurenSpiegel:
If the backend is anything other than 'mem', the number of workers should be set by the config.


rahulreddy:
👍

alexandre-merle:
:+1:

PR #79, opened 2016-06-23
BF:Return error when invalid MaxKeys for bucketGet

nicolas2bert:
In order to fix this issue : https://github.com/scality/S3/issues/75


ThibaultRiviere:
In s3 the maxKey is 1000 .
```
Sets the maximum number of keys returned in the response body. If you want to retrieve fewer than the default 1,000 keys, you can add this to your request.
```
from: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html

You can't retrieve more than 1000 in once . 
Should we do the same here ?
Except this question . :+1:

LaurenSpiegel:
Replaced by https://github.com/scality/S3/pull/82

PR #74, opened 2016-06-22
Dev/ft/test get object bucket

YiuRULE:
Create tests for amazon S3 server for get object on bucket.
@LaurenSpiegel

PR #82, opened 2016-06-24
BF: Limiting maxKeys/maxUploads/maxParts

nicolas2bert:
**bucketGet.js** : 
    -  max-keys can not be negative,
     if the value is negative we return an error message to the user " errors.InvalidArgument".
**listParts.js** : 
    - Making 1000 uploads, the maximum number of uploads a response can include.
    - Making 1000 the default value.
    - The response can not include negative value,
     if the value is negative we return an error message to the user " errors.InvalidArgument".
    [http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListParts.html](http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListParts.html)
**listMultipartUploads.js** :
    - Making 1000 multipart uploads, the maximum number of uploads a response can include.
    - Making 1000 the default value.
    - The response can not include negative value,
     if the value is negative we return an error message to the user " errors.InvalidArgument".
  [http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListMPUpload.html]( http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListMPUpload.html)

### DIFF
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -3,6 +3,7 @@ import constants from '../../constants';
 
 import services from '../services';
 import escapeForXML from '../utilities/escapeForXML';
+import { errors } from 'arsenal';
 
 //	Sample XML response:
 /*	<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
@@ -45,6 +46,9 @@ export default function bucketGet(authInfo, request, log, callback) {
     const encoding = params['encoding-type'];
     let maxKeys = params['max-keys'] ?
         Number.parseInt(params['max-keys'], 10) : 1000;
+    if (maxKeys < 0) {
+        return callback(errors.InvalidArgument);
+    }
     if (maxKeys > constants.listingHardLimit) {
         maxKeys = constants.listingHardLimit;
     }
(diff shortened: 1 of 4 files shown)


DavidPineauScality:
We agree that this change only limits the listing capabilities rather than limiting the number of parts that we can upload ?

I'd like to see that properly explained in the commit message (it makes it much easier to track later on than tracking from the PRs/issues), saying that this is effectively enforcing the limit of   "0 <= n <= 1000" items into any kind of listing.

otherwise, the change looks good.

nicolas2bert:
Yes we agree on that.
Made it clearer, let me know.
Thanks.

DavidPineauScality:
@nicolas2bert you updated the PR's message, not the commit's, while I was mostly whishing for the commit's message to be updated :)
That being said, the new PR message would be a good fit for the commit message, you could just include it as-is into it !

DavidPineauScality:
thanks, good for me now :+1:

DavidPineauScality:
@scality/team-ironman-core Please review this, as it's a bugfix against rel/1.1, and required for @YiuRULE's pr to be merged.

Note: A forward port will be necessary before the PR #74 can be merged into master.

alexandre-merle:
LGTM :+1:

LaurenSpiegel:
@ironman-machine key thoughts?

PR #85, opened 2016-06-28
Prevent overriding of file if hash collision.

LaurenSpiegel:
See https://nodejs.org/api/fs.html#fs_fs_open_path_flags_mode_callback

'w' - Open file for writing. The file is created (if it does not exist) or truncated (if it exists).

'wx' - Like 'w' but fails if path exists.

### DIFF
--- a/lib/data/file/backend.js
+++ b/lib/data/file/backend.js
@@ -41,7 +41,7 @@ export const backend = {
 
         request.pause();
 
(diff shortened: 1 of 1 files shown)


LaurenSpiegel:
@mastachand

RachedBenMustapha:
👍

rahulreddy:
👍

PR #87, opened 2016-07-08
Ft/policy auth

LaurenSpiegel:
This sends a RequestContext to vault with an authentication request so that Vault can evaluate the request against any policies attached to a user (if the request is from a user).

The RequestContext requires information from the request (such as the locationConstraint on a create bucket request) so some refactoring was required to move the parsing of the locationConstraint to the routes.  Most of the changes are modifying the units tests to handle this.

This is dependent on:
1) Arsenal: https://github.com/scality/Arsenal/pull/109
2) VaultClient: https://github.com/scality/vaultclient/pull/91
3) Vault: https://github.com/scality/Vault/pull/415

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -1,4 +1,5 @@
 import { auth } from 'arsenal';
+import { policies } from 'arsenal';
 
 import bucketDelete from './bucketDelete';
 import bucketGet from './bucketGet';
@@ -21,17 +22,19 @@ import objectPutPart from './objectPutPart';
 import serviceGet from './serviceGet';
 import vault from '../auth/vault';
 
+const RequestContext = policies.RequestContext;
 auth.setAuthHandler(vault);
 
 const api = {
     callApiMethod(apiMethod, request, log, callback) {
+        const requestContext = new RequestContext(request, apiMethod, 's3', 1);
         auth.doAuth(request, log, (err, authInfo) => {
             if (err) {
                 log.trace('authentication error', { error: err });
                 return callback(err);
             }
             return this[apiMethod](authInfo, request, log, callback);
-        }, 's3', request.query);
+        }, 's3', requestContext);
     },
     bucketDelete,
     bucketGet,
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -26,12 +26,19 @@ const RequestContext = policies.RequestContext;
 auth.setAuthHandler(vault);
 
 const api = {
-    callApiMethod(apiMethod, request, log, callback) {
-        const requestContext = new RequestContext(request, apiMethod, 's3', 1);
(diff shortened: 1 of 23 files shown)


LaurenSpiegel:
@ironman-machine check it
SCALITY_VAULT_BRANCH=ft/policyAuth

LaurenSpiegel:
@ironman-machine check it without accounts instead of users
SCALITY_VAULT_BRANCH=ft/policyAuth
SCALITY_INTEGRATION_BRANCH=ft/useAccountsForTests

lib/api/bucketPut.js
```diff
@@ -29,24 +31,6 @@ export default function bucketPut(authInfo, request, log, callback) {
     }
     const bucketName = request.bucketName;
 
-    if (request.post) {
```
alexandre-merle: Just a question, it's not usefull anymore ?

lib/routes/routePUT.js
```diff
@@ -19,7 +20,31 @@ export default function routePUT(request, response, log) {
                     routesUtils.responseNoBody(err, null, response, 200, log));
             } else if (request.query.acl === undefined) {
                 // PUT bucket
-                api.callApiMethod('bucketPut', request, log, (err) =>
+                if (request.post) {
```
alexandre-merle: I don't know if it's the right place for this parsing, may be in a function where you create the requestContext can be better ? :)

alexandre-merle:
A little comment, LGTM :+1: otherwise :)

lib/routes/routePUT.js
LaurenSpiegel: I think it makes sense here since this parsing relates specifically to this route/api.

lib/api/api.js
```diff
@@ -21,17 +22,26 @@ import objectPutPart from './objectPutPart';
 import serviceGet from './serviceGet';
 import vault from '../auth/vault';
 
+const RequestContext = policies.RequestContext;
 auth.setAuthHandler(vault);
 
 const api = {
-    callApiMethod(apiMethod, request, log, callback) {
+    callApiMethod(apiMethod, request, log, callback, locationConstraint) {
```
rahulreddy: It's odd to see a param after callback. Common accepted pattern is to use
```javascript
if (typeof locationConstraint === 'function') {
    callback = locationConstraint;
    locationConstraint = null;
}
```

LaurenSpiegel: as discussed, this would result in parameter reassignment and the locationConstraint argument after the callback is only used in the putBucket call.

LaurenSpiegel:
@ironman-machine are any of the refactors not to your liking?
SCALITY_VAULT_BRANCH=ft/policyAuth
SCALITY_INTEGRATION_BRANCH=ft/useAccountsForTests

rahulreddy:
👍

PR #90, opened 2016-07-14
test bucketclient

ttvi:
To test versioning-enabled bucketclient's backward-compatibility with existing S3 and MetaData.


ttvi:
@ironman-machine your important part

PR #91, opened 2016-07-18
bucket versioning #1 : prerequisites

ttvi:
Step 1 (this PR): S3 to support versioning for buckets.
Step 2: S3 to provide routes for interacting with Metadata.
Step 3: S3 to provide versioning support for each operation.
Step 4: MetaData to support versioning.

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -5,9 +5,11 @@ import { auth, errors } from 'arsenal';
 import bucketDelete from './bucketDelete';
 import bucketGet from './bucketGet';
 import bucketGetACL from './bucketGetACL';
+import bucketGetVersioning from './bucketGetVersioning';
 import bucketHead from './bucketHead';
 import bucketPut from './bucketPut';
 import bucketPutACL from './bucketPutACL';
+import bucketPutVersioning from './bucketPutVersioning';
 import completeMultipartUpload from './completeMultipartUpload';
 import initiateMultipartUpload from './initiateMultipartUpload';
 import listMultipartUploads from './listMultipartUploads';
@@ -87,9 +89,11 @@ const api = {
     bucketDelete,
     bucketGet,
     bucketGetACL,
+    bucketGetVersioning,
     bucketHead,
     bucketPut,
     bucketPutACL,
+    bucketPutVersioning,
     completeMultipartUpload,
     initiateMultipartUpload,
     listMultipartUploads,
--- a/lib/api/apiUtils/bucket/bucketDeletion.js
+++ b/lib/api/apiUtils/bucket/bucketDeletion.js
@@ -110,7 +110,7 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
 
     return async.waterfall([
         function checkForObjectsStep(next) {
-            return metadata.listObject(bucketName, null, null, null, 1, log,
+            return metadata.listObject(bucketName, { maxKeys: 1 }, log,
                 (err, objectsListRes) => {
                     if (err) {
                         log.error('error from metadata', { error: err });
@@ -128,8 +128,8 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
         // delete a bucket even if there are ongoing multipart uploads.
         function deleteMPUbucketStep(next) {
             const MPUBucketName = `${constants.mpuBucketPrefix}${bucketName}`;
-            return metadata.listObject(MPUBucketName, null, null, null,
-                1, log, (err, objectsListRes) => {
+            return metadata.listObject(MPUBucketName,
+                { maxKeys: 1 }, log, (err, objectsListRes) => {
                     // If no shadow bucket ever created, no ongoing MPU's, so
                     // continue with deletion
                     if (err && err.NoSuchBucket) {
--- a/lib/api/bucketGetVersioning.js
+++ b/lib/api/bucketGetVersioning.js
@@ -1,0 +1,67 @@
+import xml from 'xml';
+import services from '../services';
+
+//	Sample XML response:
+/*
+   <VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+       <Status>VersioningState</Status>
+       <MfaDelete>MfaDeleteState</MfaDelete>
+   </VersioningConfiguration>
+*/
+
+/**
+ * Convert Versioning Configuration object of a bucket into xml format.
+ * @param {object} versioningConfiguration - versioning configuration object
+ * @return {string} - the converted xml string of the versioning configuration
+ */
+function convertToXml(versioningConfiguration) {
+    const options = { declaration: { standalone: 'yes', encoding: 'UTF-8' } };
+    const constructedJSON = {
+        VersioningConfiguration: [
+            { _attr: { xmlns: 'http://s3.amazonaws.com/doc/2006-03-01/' } },
+        ],
+    };
+    if (versioningConfiguration && versioningConfiguration.Status) {
+        constructedJSON.VersioningConfiguration.push(
+            { Status: versioningConfiguration.Status });
+    }
+    if (versioningConfiguration && versioningConfiguration.MfaDelete) {
+        constructedJSON.VersioningConfiguration.push(
+            { MfaDelete: versioningConfiguration.MfaDelete });
+    }
+    return xml(constructedJSON, options);
+}
+
+
+/**
+ * bucketGetVersioning - Return Versioning Configuration for bucket
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to respond to http request
+ *  with either error code or xml response body
+ * @return {undefined}
+ */
+export default function bucketGetVersioning(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketGetVersioning' });
+
+    const bucketName = request.bucketName;
+
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketGetVersioning',
+        log,
+    };
+
+    services.metadataValidateAuthorization(metadataValParams, (err, bucket) => {
+        if (err) {
+            log.debug('error processing request',
+                { method: 'bucketGetVersioning', error: err });
+            return callback(err);
+        }
+        const versioningConfiguration = bucket.getVersioningConfiguration();
+        const xml = convertToXml(versioningConfiguration);
+        return callback(null, xml);
+    });
+}
--- a/lib/api/bucketPutVersioning.js
+++ b/lib/api/bucketPutVersioning.js
@@ -1,0 +1,71 @@
+import { waterfall } from 'async';
+import { parseString } from 'xml2js';
+
+import metadata from '../metadata/wrapper';
+import services from '../services';
+
+/**
+ * Format of xml request:
+
+ <VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+    <Status>VersioningState</Status>
+    <MfaDelete>MfaDeleteState</MfaDelete>
+ </VersioningConfiguration>
+
+ Note that there is the header in the request if setting MfaDelete:
+    x-amz-mfa: [SerialNumber] [TokenCode]
+ */
+
+/**
+ * Bucket Put Versioning - Create or update bucket Versioning
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+export default function bucketPutVersioning(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketPutVersioning' });
+
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketPutVersioning',
+        log,
+    };
+
+    return waterfall([
+        next => services.metadataValidateAuthorization(metadataValParams,
+            (err, bucket) => next(err, bucket)), // to ignore null object
+        (bucket, next) => parseString(request.post, (err, result) => {
+            // just for linting; there should not be any parsing error here
+            if (err) {
+                return next(err);
+            }
+            const versioningConfiguration = {};
+            if (result.VersioningConfiguration.Status) {
+                versioningConfiguration.Status =
+                    result.VersioningConfiguration.Status[0];
+            }
+            if (result.VersioningConfiguration.MfaDelete) {
+                versioningConfiguration.MfaDelete =
+                    result.VersioningConfiguration.MfaDelete[0];
+            }
+            // the configuration has been checked before
+            return next(null, bucket, versioningConfiguration);
+        }),
+        (bucket, versioningConfiguration, next) => {
+            bucket.setVersioningConfiguration(versioningConfiguration);
+            // TODO all metadata updates of bucket should be using CAS
+            metadata.updateBucket(bucket.getName(), bucket, log, next);
+        },
+    ], err => {
+        if (err) {
+            log.trace('error processing request', { error: err,
+                method: 'bucketPutVersioning' });
+            return callback(err);
+        }
+        return callback(err, 'Versioning Set');
+    });
+}
--- a/lib/metadata/wrapper.js
+++ b/lib/metadata/wrapper.js
@@ -117,9 +117,9 @@ const metadata = {
         });
     },
 
-    listObject: (bucketName, prefix, marker, delimiter, maxKeys, log, cb) => {
+    listObject: (bucketName, listingParams, log, cb) => {
         client
-            .listObject(bucketName, { prefix, marker, maxKeys, delimiter },
+            .listObject(bucketName, listingParams,
                     log, (err, data) => {
                         log.debug('getting object listing from metadata');
                         if (err) {
--- a/lib/routes/routeGET.js
+++ b/lib/routes/routeGET.js
@@ -20,6 +20,14 @@ export default function routerGET(request, response, log, utapi) {
                     request.bucketName);
                 return routesUtils.responseXMLBody(err, xml, response, log);
             });
+        } else if (request.query.versioning !== undefined) {
+            api.callApiMethod('bucketGetVersioning', request, log,
+                (err, xml) => {
+                    // TODO push metrics for gucketGetVersioning
+                    // pushMetrics(err, log, utapi, 'bucketGetVersioning',
+                    //     request.bucketName);
+                    routesUtils.responseXMLBody(err, xml, response, log);
+                });
         } else if (request.query.uploads !== undefined) {
             // List MultipartUploads
             api.callApiMethod('listMultipartUploads', request, log,
--- a/lib/routes/routePUT.js
+++ b/lib/routes/routePUT.js
@@ -14,6 +14,9 @@ const encryptionHeaders = [
     'x-amz-server-side-encryption-customer-key-md5',
 ];
 
+const validStatuses = ['Enabled', 'Suspended'];
+const validMfaDeletes = [undefined, 'Enabled', 'Disabled'];
+
 /* eslint-disable no-param-reassign */
 export default function routePUT(request, response, log, utapi) {
     log.debug('routing request', { method: 'routePUT' });
@@ -32,6 +35,47 @@ export default function routePUT(request, response, log, utapi) {
                         request.bucketName);
                     return routesUtils.responseNoBody(err, null, response, 200,
                         log);
+                });
+            } else if (request.query.versioning !== undefined) {
+                if (request.post === '') {
+                    log.debug('request xml is missing');
+                    return routesUtils.responseNoBody(
+                        errors.MalformedXML, null, response, null, log);
+                }
+                const xmlToParse = request.post;
+                return parseString(xmlToParse, (err, result) => {
+                    if (err) {
+                        log.debug('request xml is malformed');
+                        return routesUtils.responseNoBody(
+                            errors.MalformedXML, null, response, null, log);
+                    }
+                    const status = result.VersioningConfiguration.Status ?
+                        result.VersioningConfiguration.Status[0] : undefined;
+                    const mfaDelete = result.VersioningConfiguration.MfaDelete ?
+                        result.VersioningConfiguration.MfaDelete[0] : undefined;
+                    if (validStatuses.indexOf(status) < 0 ||
+                        validMfaDeletes.indexOf(mfaDelete) < 0) {
+                        log.debug('illegal versioning configuration');
+                        return routesUtils.responseNoBody(
+                            errors.IllegalVersioningConfigurationException,
+                            null, response, null, log);
+                    }
+                    if (mfaDelete) {
+                        log.debug('mfa deletion is not implemented');
+                        return routesUtils.responseNoBody(
+                            errors.NotImplemented.customizedDescription(
+                                'MFA Deletion is not supported yet.'), null,
+                            response, null, log);
+                    }
+                    return api.callApiMethod('bucketPutVersioning', request,
+                        log, err => {
+                            // TODO push metrics for bucketPutVersioning
+                            // pushMetrics(err, log, utapi,
+                            //         'bucketPutVersioning',
+                            //         request.bucketName);
+                            routesUtils.responseNoBody(
+                                err, null, response, 200, log);
+                        });
                 });
             } else if (request.query.acl === undefined) {
                 // PUT bucket
--- a/lib/services.js
+++ b/lib/services.js
@@ -25,7 +25,7 @@ export default {
         const bucketUsers = overrideUserbucket || usersBucket;
         // Note that since maxKeys on a listObject request is 10,000,
         // this request will retrieve up to 10,000 bucket names for a user.
-        metadata.listObject(bucketUsers, prefix, null, null, null, log,
+        metadata.listObject(bucketUsers, { prefix }, log,
             (err, listResponse) => {
                 // If MD responds with NoSuchBucket, this means the
                 // hidden usersBucket has not yet been created for
@@ -383,11 +383,9 @@ export default {
      */
     getObjectListing(bucketName, listingParams, log, cb) {
         assert.strictEqual(typeof bucketName, 'string');
-        const { delimiter, marker, prefix } = listingParams;
-        const maxKeys = Number(listingParams.maxKeys);
         log.trace('performing metadata get object listing',
-            { listingParams, maxKeys });
-        metadata.listObject(bucketName, prefix, marker, delimiter, maxKeys, log,
+            { listingParams });
+        metadata.listObject(bucketName, listingParams, log,
             (err, listResponse) => {
                 if (err) {
                     log.warn('error from metadata', { error: err });
@@ -521,8 +519,7 @@ export default {
                 delimiter: undefined,
                 maxKeys: 1,
             };
-            metadata.listObject(mpuBucketName, searchArgs.prefix,
-                searchArgs.marker, searchArgs.delimiter, searchArgs.maxKeys,
+            metadata.listObject(mpuBucketName, searchArgs,
                 log, (err, response) => {
                     if (err) {
                         return cb(err);
@@ -739,8 +736,7 @@ export default {
             delimiter: undefined,
             maxKeys: 10000,
         };
-        metadata.listObject(mpuBucketName, searchArgs.prefix, searchArgs.marker,
-            searchArgs.delimiter, searchArgs.maxKeys, log, cb);
+        metadata.listObject(mpuBucketName, searchArgs, log, cb);
     },
 
     getSomeMPUparts(params, cb) {
@@ -754,8 +750,7 @@ export default {
             delimiter: undefined,
             maxKeys: maxParts,
         };
-        metadata.listObject(mpuBucketName, searchArgs.prefix, searchArgs.marker,
-            searchArgs.delimiter, searchArgs.maxKeys, log, cb);
+        metadata.listObject(mpuBucketName, searchArgs, log, cb);
     },
 
     batchDeleteObjectMetadata(mpuBucketName, keysToDelete, log, cb) {
--- a/tests/functional/aws-node-sdk/test/bucket/testBucketVersioning.js
+++ b/tests/functional/aws-node-sdk/test/bucket/testBucketVersioning.js
@@ -1,0 +1,124 @@
+import assert from 'assert';
+import { S3 } from 'aws-sdk';
+
+import getConfig from '../support/config';
+
+const bucket = `versioning-bucket-${Date.now()}`;
+
+describe('aws-node-sdk test bucket versioning', function testSuite() {
+    this.timeout(60000);
+    let s3;
+
+    // setup test
+    before(done => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+        s3.createBucket({ Bucket: bucket }, done);
+    });
+
+    // delete bucket after testing
+    after(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+    it('should not accept empty versioning configuration', done => {
+        const params = {
+            Bucket: bucket,
+            VersioningConfiguration: {},
+        };
+        s3.putBucketVersioning(params, error => {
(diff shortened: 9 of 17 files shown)


ttvi:
@ironman-machine let's go beta (then ramen for lunch)

SCALITY_FEDERATION_BRANCH=FT/metadata-vault-disable-versioning
SCALITY_METADATA_BRANCH=exp/geobucket_versioning
SCALITY_VAULT_BRANCH=master

lib/api/bucketGet.js
```diff
@@ -29,6 +31,177 @@ import { errors } from 'arsenal';
...
+    list.CommonPrefixes.forEach(item => {
+        xml.push(
+            `<CommonPrefixes><Prefix>${item}</Prefix></CommonPrefixes>`
+        );
+    });
+    xml.push('</ListBucketResult>');
+    return xml.join('');
+}
+
+// Sample XML response for versioning:
```
LaurenSpiegel: Sample XML response for GET Bucket Object Versions API

lib/api/bucketGet.js
```diff
@@ -63,9 +238,13 @@ export default function bucketGet(authInfo, request, log, callback) {
         delimiter: params.delimiter,
         marker: params.marker,
         prefix: params.prefix,
+
     };
+    if (versions) {
+        listParams.versions = versions;
+    }
```
LaurenSpiegel: I don't think you use listParams.versions so don't need to do this. If you do need it, it would be clearer to just set listParams.versions to true.

lib/api/bucketGet.js
```diff
@@ -29,6 +31,177 @@ import { errors } from 'arsenal';
...
+            xml.push(`<${param.tag}>${param.value}</${param.tag}>`);
+        } else {
+            xml.push(`<${param.tag}/>`);
+        }
+    });
+
+    list.Contents.forEach(item => {
+        const key = encoding === 'url' ?
+            querystring.escape(item.key) : escapeForXML(item.key);
+        const v = VSUtils.decodeVersion(item.value);
```
LaurenSpiegel: Since this is a json.parse it should be wrapped in a try/catch here or in the utils function itself.

lib/api/bucketGet.js
```diff
@@ -29,6 +31,177 @@ import { errors } from 'arsenal';
...
+        const size = v['content-length'];
+        const ETag = v['content-md5'];
+        const lastModified = v['last-modified'];
+        const owner = {
+            DisplayName: v['owner-display-name'],
+            ID: v['owner-id'],
+        };
+        const storageClass = v['x-amz-storage-class'];
+
+        const versionId = VSUtils.getts(v);
```
LaurenSpiegel: If you are using getters on some attributes might as well be consistent and do it for all.

lib/api/bucketGet.js
```diff
@@ -29,6 +31,177 @@ import { errors } from 'arsenal';
...
+        const isLatest = false;
```
LaurenSpiegel: How do you know this?  Is this a todo to make dynamic later?

lib/api/bucketGet.js
```diff
@@ -29,6 +31,177 @@ import { errors } from 'arsenal';
...
+                '</DeleteMarker>'
+            );
+        } else {
+            xml.push(
+                '<Version>',
+                `<Key>${key}</Key>`,
+                `<VersionId>${versionId}</VersionId>`,
+                `<IsLatest>${isLatest}</IsLatest>`,
+                `<LastModified>${lastModified}</LastModified>`,
+                `<ETag>${ETag}</ETag>`,
```
LaurenSpiegel: Your code made me realize that the etag listing should appear like this:
`<ETag>&quot;${ETag}&quot;</ETag>`

It should be this way in both listings.

lib/api/objectGet.js
```diff
@@ -36,6 +39,10 @@ function objectGet(authInfo, request, log, callback) {
                 log.debug('error processing request', { error });
                 return callback(error);
             }
+            // return ObjNotFound if the version is a delete marker
+            if (VSUtils.getdm(objMD)) {
+                return callback(errors.ObjNotFound);
```
LaurenSpiegel: I think the correct error is NoSuchKey.  See s3 errors: http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

Also, it looks like you need to add a header to the response:
"If the current version of the object is a delete marker, Amazon S3 behaves as if the object was deleted and includes x-amz-delete-marker: true in the response."

lib/api/objectPut.js
```diff
@@ -46,6 +46,10 @@ function _storeIt(bucketName, objectKey, objMD, authInfo, canonicalID,
         dataToDelete = Array.isArray(objMD.location) ?
             objMD.location : [objMD.location];
     }
+    // not deleting old data in the case of versioning
+    if (objMD && objMD['x-amz-version-id'] !== 'null') {
+        dataToDelete = undefined;
+    }
```
LaurenSpiegel: To be backwards compatible, shouldn't you do:
`if(objMD && objMD['x-amz-version-id'] && objMD['x-amz-version-id'] !== 'null')`

This assumes that you will be setting the objMD['x-amz-version-id'] to null for all non-versioned objects.

lib/api/objectPut.js
```diff
@@ -10,14 +10,14 @@ import { logger } from '../utilities/logger';
 function _storeInMDandDeleteData(bucketName, objMD, dataGetInfo,
     metadataStoreParams, dataToDelete, deleteLog, callback) {
     services.metadataStoreObject(bucketName, objMD, dataGetInfo,
-        metadataStoreParams, (err, contentMD5) => {
+        metadataStoreParams, (err, value) => {
```
LaurenSpiegel: what is being sent back now?

lib/routes/routePUT.js
```diff
@@ -104,6 +104,11 @@ export default function routePUT(request, response, log) {
                 }
                 // ETag's hex should always be enclosed in quotes
                 const resMetaHeaders = { ETag: `"${request.calculatedHash}"` };
+                if (data) {
+                    const _data = typeof data === 'object' ?
```
LaurenSpiegel: Why is this sometimes an object?

lib/routes/routePUT.js
```diff
@@ -104,6 +104,11 @@ export default function routePUT(request, response, log) {
...
+                        data : JSON.parse(data);
```
LaurenSpiegel: should put in a try/catch

lib/services.js
```diff
@@ -258,24 +260,30 @@ export default {
                     return cb(err);
                 }
                 omVal.acl = parsedACL;
-                metadata.putObjectMD(bucketName, objectKey, omVal, log, err => {
+                metadata.putObjectMD(bucketName, objectKey, omVal, log,
```
LaurenSpiegel: shouldn't you be updating the omVal['x-amz-server-version-id'] above?

lib/api/objectPut.js
LaurenSpiegel: I see now that objMD['x-amz-version-id'] was set to 'null' in current implementation so ok on backwards compatibility but shouldn't we still be able to delete data if versioning is not enabled?

lib/services.js
```diff
@@ -258,24 +260,30 @@ export default {
...
+                    (err, data) => {
+                        if (err) {
+                            log.error('error from metadata', { error: err });
+                            return cb(err);
+                        }
+                        log.trace('object successfully stored in metadata');
+                        const value = data ? JSON.parse(data) : {};
```
LaurenSpiegel: Could we have a class for the data coming back with a deserialize method that is called in the wrapper?

lib/services.js
```diff
@@ -288,18 +296,21 @@ export default {
...
      * @return {undefined}
      */
-    deleteObject(bucketName, objectMD, responseMDHeaders, objectKey, log, cb) {
+    deleteObject(bucketName, objectMD, responseMDHeaders, objectKey, log, cb,
+        query) {
         log.trace('deleting object from bucket');
         assert.strictEqual(typeof bucketName, 'string');
         assert.strictEqual(typeof objectMD, 'object');
-        if (objectMD['x-amz-version-id'] === 'null') {
+        if (objectMD['x-amz-version-id'] === 'null' || query.versionId) {
```
LaurenSpiegel: why does having a query.versionId indicate that the object is non-versioned?

ttvi: No it's part of the opposite. This check is to see if we are deleting a non-versioned object or a specific version---cases when we need to physically delete object's data. The later part of the condition means when the object is versioning-enabled and the request is to delete a specific version; it is equivalent to `objectMD['x-amz-version-id'] !== 'null' && query.versionId`.

lib/api/bucketGet.js
ttvi: I was hesitate to do that because I did not know whether these attributes should be known locally to S3 only or should metadata also know about them.

lib/api/objectPut.js
ttvi: This is another change in the S3-metadata communication. Now metadata is sending back `ETag` to S3. In the case of versioning, it needs to include the `versionId` as well.

lib/api/objectPut.js
ttvi: Yes we should. In normal cases, `objMD['x-amz-version-id'] === 'null'`.

lib/routes/routePUT.js
ttvi: Right. It's the very first patch for versioning put to receive a `versionId` back after an `objectPut`. I will need to check with the call path and remove it.

lib/services.js
ttvi: I don't know what it is. WIll take a look.

lib/services.js
ttvi: Yes. It'd be better.

lib/api/bucketGet.js
ttvi: The `decodeVersion` might or might not need `JSON.parse`. In any case try/catch should be in the utils.

lib/api/bucketGet.js
ttvi: Yes, this will be changed.

lib/api/bucketGet.js
ttvi: Hmm, it should be in `metadataValParams`.

ttvi:
Thanks @LaurenSpiegel , updates will be done soon.

lib/services.js
LaurenSpiegel: ok.  then, the log messages need to be updated.

package.json
```diff
@@ -58,6 +58,7 @@
   },
   "scripts": {
     "ft_test": "mocha tester.js --compilers js:babel-core/register",
+    "test_aws": "babel-node tests/functional/aws-node-sdk/test/legacy/test.js",
```
LaurenSpiegel: I assume we don't want this in the final PR.  Note that this is not all of the node aws sdk tests.

tests/unit/api/objectPut.js
```diff
@@ -201,6 +201,10 @@ describe('objectPut API', () => {
             });
     });
 
+    /**
+     * now overwriting an object in versioning means adding a new version, thus
+     * old data will node be deleted just be overwriting an existing object
```
LaurenSpiegel: now overwriting an object in versioning means adding a new version, thus old data will not be deleted

package.json
ttvi: oops

lib/api/bucketGet.js
ttvi: Actually I do use `versions` in `listParams` to inform metadata that this is a listing of all versions.

LaurenSpiegel: ok.  Still would be clearer to just assign it to true explicitly.

vrancurel:
We can merge that as there is no implication if we don't use versioning

ttvi:
@vrancurel I'm separating the versioning and replication parts. I will call for review again after that.

LaurenSpiegel:
@ttvi, would it be possible to keep the cb as the last argument in the functions if we are going to merge this?

ttvi:
@LaurenSpiegel yes of course. But it will require searching for all usages and modify them. I will make it the last step of this PR.

ttvi:
@ironman-machine please

lib/api/apiUtils/bucket/bucketDeletion.js
```diff
@@ -110,7 +110,7 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
 
     return async.waterfall([
         function checkForObjectsStep(next) {
-            return metadata.listObject(bucketName, null, null, null, 1, log,
+            return metadata.listObject(bucketName, { maxKeys: 1 }, log,
```
MathieuCassagne: Not really in the scope of the PR, but why do a listing of one key and not a get ?

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,84 @@
...
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketPutVersioning',
+        log,
+    };
+
+    return async.waterfall([
+        function waterfall1(next) {
```
MathieuCassagne: Can you put an explicit name for the funciton in the waterfall ? :)

lib/api/apiUtils/bucket/bucketDeletion.js
ttvi: This is in the context of deleting a bucket. In order to delete a bucket, we need to ensure that it doesn't contain any object, and listing is a way to check that. I think using `maxKeys: 1` here is an optimization; you can also use the max threshold, but it just wastes disk ios and bandwidth.

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,84 @@
+import async from 'async';
```
michael-zapata: Nit: `import { waterfall } from 'async';` would work as well.

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,84 @@
...
+
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketPutVersioning',
+        log,
+    };
+
+    return async.waterfall([
```
michael-zapata: Could we have each function defined out of the waterfall to avoid allocating it every time and not doing any v8 optimisation?

lib/metadata/wrapper.js
```diff
@@ -117,9 +117,9 @@ const metadata = {
         });
     },
 
-    listObject: (bucketName, prefix, marker, delimiter, maxKeys, log, cb) => {
+    listObject: (bucketName, listingParams, log, cb) => {
```
michael-zapata: Why this change? The previous signature was explicit and avoided creating/allocating an object.

lib/routes/routePUT.js
```diff
@@ -33,6 +33,42 @@ export default function routePUT(request, response, log, utapi) {
...
+                        errors.MalformedXML, null, response, null, log);
+                }
+                const xmlToParse = request.post;
+                return parseString(xmlToParse, (err, result) => {
+                    if (err) {
+                        log.debug('request xml is malformed');
+                        return routesUtils.responseNoBody(
+                            errors.MalformedXML, null, response, null, log);
+                    }
+                    const Status = result.VersioningConfiguration.Status ?
```
michael-zapata: Nit: camelCasing would be welcome, the linter should complain about that.

lib/routes/routePUT.js
```diff
@@ -33,6 +33,42 @@ export default function routePUT(request, response, log, utapi) {
...
+                        result.VersioningConfiguration.Status[0] : undefined;
+                    const MfaDelete = result.VersioningConfiguration.MfaDelete ?
+                        result.VersioningConfiguration.MfaDelete[0] : undefined;
+                    const validStatuses = ['Enabled', 'Suspended'];
```
michael-zapata: Since those are defined once and for all, could we have them in the global scope instead of allocating them every time we go into this call? Also shouldn't this have `undefined` as well?

lib/routes/routePUT.js
```diff
@@ -33,6 +33,42 @@ export default function routePUT(request, response, log, utapi) {
...
+                    const validMfaDeletes = [undefined, 'Enabled', 'Disabled'];
```
michael-zapata: Same here.

tests/functional/aws-node-sdk/test/legacy/testBucketVersioning.js
```diff
@@ -0,0 +1,131 @@
+import assert from 'assert';
```
michael-zapata: These aren't legacy tests, right?

Also, tests are not self-contained. (They should be able to run with an `.only()`)

lib/metadata/wrapper.js
ttvi: For the ability to extend. Listing does not always take only these arguments, for example listing all versions of a bucket uses `max-keys`, `key-marker`, and `version-id-marker`.

lib/routes/routePUT.js
ttvi: It's AWS' format. See http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTVersioningStatus.html?shortFooter=true.

lib/routes/routePUT.js
ttvi: It will be at the global scope.
A valid `Status` should not be undefined while `MfaDelete` could be.

tests/functional/aws-node-sdk/test/legacy/testBucketVersioning.js
ttvi: I'm not very familiar with the organization of the tests here.
Where do you think it should go?

lib/routes/routePUT.js
michael-zapata: Yeah, but we're talking about an internal variable here, hence my comment.

lib/metadata/wrapper.js
michael-zapata: We should solve this by adding arguments as needed, then.

lib/routes/routePUT.js
ttvi: Oh right, the `Status` and `MfaDelete`. Thanks.

lib/metadata/wrapper.js
ttvi: It would be a nightmare for readability. There is [List Objects Version 2](http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html?shortFooter=true) with its own set of parameters coming as well. I will keep this unless there are more requests like yours (consensus).

ttvi: Performance-wise, they are not different because inside this function, we have to create an object out of the listing parameters anyway.

ttvi:
@MathieuCassagne @michael-zapata thanks for comments. I keep the revision in a new commit (to be squashed) to keep track of the changes.

tests/functional/aws-node-sdk/test/legacy/testBucketVersioning.js
michael-zapata: We have `./tests/functional/aws-node-sdk/test/bucket` already created, you can add the file there.

lib/api/bucketPutVersioning.js
```diff
@@ -16,6 +16,46 @@ import services from '../services';
...
+            });
+            return next(err);
+        }
+        return next(null, bucket);
+    });
+}
+
+function waterfall2ParseVersioningConfiguration(request, bucket, next) {
+    parseString(request.post, (err, result) => {
+        // just for linting; there should not be any parsing error here
```
michael-zapata: Since this is based on user-input, you may very well have a parsing error (malicious intent or not).

lib/api/bucketPutVersioning.js
```diff
@@ -16,6 +16,46 @@ import services from '../services';
     x-amz-mfa: [SerialNumber] [TokenCode]
  */
 
+
+function waterfall1ValidateAuthorization(mdValParams, log, next) {
```
michael-zapata: I think you can skip the `waterfallX` here, it's a bit too verbose.

lib/api/bucketPutVersioning.js
```diff
@@ -16,6 +16,46 @@ import services from '../services';
...
+    services.metadataValidateAuthorization(mdValParams, (err, bucket) => {
+        if (err) {
+            log.trace('request authorization failed', {
```
michael-zapata: There's already logs for that, I don't think this is necessary. Especially since that would just get rid of that function since you could then use `services.metadataValidateAuthorization()` directly in the waterfall.

lib/api/bucketPutVersioning.js
```diff
@@ -16,6 +16,46 @@ import services from '../services';
...
+                method: 'services.metadataValidateAuthorization',
+            });
+            return next(err);
+        }
+        return next(null, bucket);
+    });
+}
+
+function waterfall2ParseVersioningConfiguration(request, bucket, next) {
+    parseString(request.post, (err, result) => {
```
michael-zapata: Nit: `parseString()` could also be in the waterfall itself. Nothing major there.

ttvi:
Updated.

lib/api/bucketGetVersioning.js
```diff
@@ -0,0 +1,67 @@
...
+ * bucketGetVersioning - Return Versioning Configuration for bucket
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to respond to http request
+ *  with either error code or xml response body
+ * @return {undefined}
+ */
+export default function bucketGetVersioning(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketGetACL' });
```
alexandre-merle: BucketGetVersionning :)

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,71 @@
...
+        log,
+    };
+
+    return waterfall([
+        next => services.metadataValidateAuthorization(metadataValParams,
+            (err, bucket) => next(err, bucket)), // to ignore null object
+        (bucket, next) => parseString(request.post, (err, result) => {
+            // just for linting; there should not be any parsing error here
+            if (err) {
+                return next(err);
```
alexandre-merle: here its more a log of the error and send back an internal error :)

lib/metadata/wrapper.js
alexandre-merle: i'm on @ttvi side, options should always be an object, to avoid modify everything when we add an option :)

lib/routes/routePUT.js
```diff
@@ -33,6 +36,40 @@ export default function routePUT(request, response, log, utapi) {
...
+            } else if (request.query.versioning !== undefined) {
+                if (request.post === '') {
+                    log.debug('request xml is missing');
+                    return routesUtils.responseNoBody(
+                        errors.MalformedXML, null, response, null, log);
+                }
+                const xmlToParse = request.post;
+                return parseString(xmlToParse, (err, result) => {
+                    if (err) {
+                        log.debug('request xml is malformed');
```
alexandre-merle: can you log the error here ? :) like err.stack :)

lib/api/bucketGetVersioning.js
ttvi: Thanks. Fixed.

lib/routes/routePUT.js
ttvi: I don't mind modifying it but don't you think the logging message is expressive enough?

lib/api/bucketPutVersioning.js
ttvi: The xml has actually been parsed here https://github.com/scality/S3/pull/91/files#diff-25f86c99d9d8badaa3509e13e8cc8edbR49.

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,71 @@
...
+        bucketName,
+        requestType: 'bucketPutVersioning',
+        log,
+    };
+
+    return waterfall([
+        next => services.metadataValidateAuthorization(metadataValParams,
+            (err, bucket) => next(err, bucket)), // to ignore null object
+        (bucket, next) => parseString(request.post, (err, result) => {
+            // just for linting; there should not be any parsing error here
```
LaurenSpiegel: I think you can remove this comment.

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,71 @@
...
+            if (err) {
+                return next(err);
+            }
+            const versioningConfiguration = {};
+            if (result.VersioningConfiguration.Status) {
+                versioningConfiguration.Status =
+                    result.VersioningConfiguration.Status[0];
+            }
+            if (result.VersioningConfiguration.MfaDelete) {
```
LaurenSpiegel: I think we should return a NotImplemented if this is in the request.  We do not want the user to think we are honoring this.

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,71 @@
...
+ * @return {undefined}
+ */
+export default function bucketPutVersioning(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketPutVersioning' });
+
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketPutVersioning',
```
LaurenSpiegel: This has to be addressed in the acl check functions.  The purpose of sending in the requestType is to know whether the requester has the authority under the acl's to perform the given action.  I don't think putBucketVersioning is addressed directly in the aws acl docs.  Perhaps only the bucket owner can call this api?

lib/api/bucketGetVersioning.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+export default function bucketGetVersioning(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketGetVersioning' });
+
+    const bucketName = request.bucketName;
+
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketGetVersioning',
```
LaurenSpiegel: See comment regarding getBucketVersioning below and acl's.

lib/metadata/wrapper.js
LaurenSpiegel: Vote for the @ttvi  approach. This way is so much more readable.

lib/routes/routeGET.js
```diff
@@ -20,6 +20,14 @@ export default function routerGET(request, response, log, utapi) {
                     request.bucketName);
                 return routesUtils.responseXMLBody(err, xml, response, log);
             });
+        } else if (request.query.versioning !== undefined) {
+            api.callApiMethod('bucketGetVersioning', request, log,
+                (err, xml) => {
+                    // TODO push metrics for gucketGetVersioning
+                    // pushMetrics(err, log, utapi, 'bucketGetVersioning',
+                    //     request.bucketName);
```
LaurenSpiegel: Please open an issue in s3 and utapi so this does not get forgotten.

lib/routes/routePUT.js
```diff
@@ -33,6 +36,40 @@ export default function routePUT(request, response, log, utapi) {
...
                         log);
                 });
+            } else if (request.query.versioning !== undefined) {
+                if (request.post === '') {
+                    log.debug('request xml is missing');
+                    return routesUtils.responseNoBody(
+                        errors.MalformedXML, null, response, null, log);
+                }
+                const xmlToParse = request.post;
+                return parseString(xmlToParse, (err, result) => {
```
LaurenSpiegel: Is there a reason to do the parsing here in the routes?  I'd prefer it in the api if possible, but it might be the right way to do things for error compatibility.  @nicolas2bert , do you have a view?

tests/functional/aws-node-sdk/test/bucket/testBucketVersioning.js
```diff
@@ -0,0 +1,131 @@
...
+        const params = { Bucket: bucket };
+        // s3.getBucketVersioning(params, done);
+        s3.getBucketVersioning(params, (error, data) => {
+            assert.strictEqual(error, null);
+            assert.deepStrictEqual(data, { Status: 'Enabled' });
+            done();
+        });
+    });
+
+    // deletebucket test
```
LaurenSpiegel: you can put this into an after block rather than its own test.

tests/functional/aws-node-sdk/test/bucket/testBucketVersioning.js
```diff
@@ -0,0 +1,131 @@
...
+
+    // setup test
+    before(() => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+
+        s3 = new S3(config);
+    });
+
+    // createbucket test
+    it('should create a bucket', done => {
```
LaurenSpiegel: probably makes more sense to put this in the before block rather than its own test.

lib/routes/routePUT.js
```diff
@@ -33,6 +36,40 @@ export default function routePUT(request, response, log, utapi) {
...
+                    if (validStatuses.indexOf(status) < 0 ||
+                        validMfaDeletes.indexOf(mfaDelete) < 0) {
+                        log.debug('illegal versioning configuration');
+                        return routesUtils.responseNoBody(
+                            errors.IllegalVersioningConfigurationException,
+                            null, response, null, log);
+                    }
+                    return api.callApiMethod('bucketPutVersioning', request,
+                        log, err => {
+                            // TODO push metrics for bucketPutVersioning
```
LaurenSpiegel: this could be in the same issue as the getBucketVersioning utapi integration.

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,71 @@
...
+        log,
+    };
+
+    return waterfall([
+        next => services.metadataValidateAuthorization(metadataValParams,
+            (err, bucket) => next(err, bucket)), // to ignore null object
+        (bucket, next) => parseString(request.post, (err, result) => {
```
ttvi: I think it's not feasible to make them static the way you want. Look at the full functions of the second waterfall:
```js
(bucket, cb) => parseString(req, (err, res) => {
    cb();
});
```
- The first anonymous function uses an extra variable `req` apart from the inputs so we cannot make it a static function unless we use `req` as a global variable.
- The second anonymous function uses the extra variable `next` as well so it's difficult to make it static.
You have any other solution?

And also, practically the function `bucketGetVersioning` should not be a frequent operation.

tests/functional/aws-node-sdk/test/bucket/testBucketVersioning.js
ttvi: You have this.

tests/functional/aws-node-sdk/test/bucket/testBucketVersioning.js
ttvi: and this.

lib/api/bucketPutVersioning.js
ttvi: I put it there, though the `NotImplemented` error will also confuse users whether versioning or only mfa deletion was not implemented.

lib/api/bucketPutVersioning.js
ttvi: Addressed.

lib/api/api.js
```diff
@@ -87,9 +89,11 @@ const api = {
     bucketDelete,
     bucketGet,
     bucketGetACL,
+    bucketGetVersioning,
```
ttvi: Mapped.

ttvi:
All good. Waiting for https://github.com/scality/Arsenal/pull/170 to land.
Please don't make the comment count 100 :|.

ttvi:
@ironman-machine

ttvi:
CI failed because of using non-existing branch #170 of Arsenal. Retrying and passing here http://ci.ironmann.io/gh/scality/Integration/5116.

lib/api/api.js
ttvi: I don't think we need the integration test wrt policies because only the bucket owner can perform these actions.

ttvi:
@LaurenSpiegel @michael-zapata your comments have been addressed/discussed. Let's finalize this.

lib/api/api.js
LaurenSpiegel: I think you mean the bucket owner can do so with respect to ACL's.  This doesn't mean that a user under the owner account does not have the right if a proper policy is attached.  I just tested with a user in real AWS and users can put and get bucket versioning.  That being said we can add an issue to integration to add this test after this PR.

PR #96, opened 2016-07-25
deleteObject's compat. w/ versioning

ttvi:
Tests are not handling the case when S3 returns an [error](https://github.com/scality/S3/blob/master/lib/services.js#L337) if users delete an object with a versioning-enabled metadata backend.
This PR fixes the behavior of S3 instead of changing the tests; S3 will not return the error and will continue to delete the object in metadata in such case. This ensures the same metadata behavior though with different backends (versioning and non-versioning).
This fix is also inline with how S3 should handled versioning in future versions.

### DIFF
--- a/lib/services.js
+++ b/lib/services.js
@@ -333,8 +333,8 @@ export default {
             });
         } else {
             // versioning
-            log.warn('deleteObject: versioning not implemented');
-            cb(errors.NotImplemented);
+            log.warn('deleteObject: versioning not fully implemented');
+            return metadata.deleteObjectMD(bucketName, objectKey, log, cb);
         }
     },
 


ttvi:
@ironman-machine end-to-end for the peace of mind

lib/services.js
```diff
@@ -333,8 +333,9 @@ export default {
             });
         } else {
             // versioning
-            log.warn('deleteObject: versioning not implemented');
-            cb(errors.NotImplemented);
+            log.warn('deleteObject: versioning not fully implemented');
+            return metadata.deleteObjectMD(bucketName, objectKey, log, cb);
+            // cb(errors.NotImplemented);
```
michael-zapata: This is already sourced, no need to keep it as a comment.

michael-zapata:
One small nit, nothing major!

ttvi:
@michael-zapata more than 30% of my PR has been modified to address your comment. Now what?

michael-zapata:
:+1:

LaurenSpiegel:
As discussed, this is step one and there will be future changes to delete data on a version specific delete and perhaps to send along the x-amz-version-id.  👍

ttvi:
@LaurenSpiegel yes, this is the minimal step for _legacy_ S3 to make versioning backends meaningful. Because present S3 deletes data on every delete, metadata's versioning and then geo-replication would be useless and not functional without this.

Now merging.

PR #97, opened 2016-07-25
Clean/md5

ploki:
needed by upcoming at REST encryption feature

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -56,4 +56,8 @@ export default {
     // http://docs.aws.amazon.com/AmazonS3/latest/API/
     //      RESTBucketGET.html#RESTBucketGET-requests
     listingHardLimit: 1000,
+
+    // AWS sets a minimum size limit for parts except for the last part.
+    // http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html
+    minimumAllowedPartSize: 5242880,
 };
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -212,8 +212,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 const partSize = Number.parseInt(storedPartsAsObjects[i]
                     .size, 10);
                 if (i < jsonList.Part.length - 1 &&
-                    partSize < constants.minimumAllowedPartSize &&
-                    request.relaxedSizeCheckingForTests !== true) {
+                    partSize < constants.minimumAllowedPartSize) {
                     return next(errors.EntityTooSmall);
                 }
                 // Assemble array of part locations, aggregate size and build
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -212,7 +212,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 const partSize = Number.parseInt(storedPartsAsObjects[i]
                     .size, 10);
                 if (i < jsonList.Part.length - 1 &&
-                    partSize < 5242880) {
+                    partSize < constants.minimumAllowedPartSize) {
                     return next(errors.EntityTooSmall);
                 }
                 // Assemble array of part locations, aggregate size and build
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -212,7 +212,8 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 const partSize = Number.parseInt(storedPartsAsObjects[i]
                     .size, 10);
                 if (i < jsonList.Part.length - 1 &&
-                    partSize < constants.minimumAllowedPartSize) {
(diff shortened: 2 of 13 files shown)


constants.js
```diff
@@ -56,4 +56,8 @@ export default {
     // http://docs.aws.amazon.com/AmazonS3/latest/API/
     //      RESTBucketGET.html#RESTBucketGET-requests
     listingHardLimit: 1000,
+
+    // AWS sets a minimum size limit for parts excepted ont the last one
```
LaurenSpiegel: AWS sets a minimum size threshold for parts in an MPU except for the last part.

lib/api/completeMultipartUpload.js
```diff
@@ -212,7 +212,8 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 const partSize = Number.parseInt(storedPartsAsObjects[i]
                     .size, 10);
                 if (i < jsonList.Part.length - 1 &&
-                    partSize < 5242880) {
+                    partSize < constants.minimumAllowedPartSize &&
+                    request.relaxedSizeCheckingForTests !== true) {
```
LaurenSpiegel: I'd rather not complicate the production code to accommodate tests which can be modified.

ploki: I agree

lib/utilities/MD5Sum.js
```diff
@@ -0,0 +1,24 @@
...
+
+class MD5Sum extends Transform {
+
+    constructor(callback) {
+        super({});
+        this.hash = crypto.createHash('md5');
+        this.callback = callback;
+    }
+
+    _transform(chunk, encoding, callback) {
```
LaurenSpiegel: let's call this callback something else so as not to be confused with the callback in the constructor -- done or cb

lib/utilities/MD5Sum.js
```diff
@@ -0,0 +1,24 @@
...
+        this.hash.update(chunk, encoding);
+        callback(null, chunk, encoding);
+    }
+
+    _flush(callback) {
```
LaurenSpiegel: same here -- done or cb

lib/data/file/backend.js
```diff
@@ -68,9 +66,7 @@ export const backend = {
             })
             .on('end', () => {
                 // eslint-disable-next-line no-param-reassign
```
michael-zapata: No need for that line since you removed the actual problem here :dancer:

lib/data/wrapper.js
```diff
@@ -24,20 +26,27 @@ if (config.backends.data === 'mem') {
...
-                key,
-                dataStoreName: implName,
-            };
-            return cb(null, dataRetrievalInfo);
+                                                method: 'put' });
+        const hashedStream = new MD5Sum((err, hash) => {
+            value.calculatedHash = hash;
         });
+        value.pipe(hashedStream);
+        client.put(hashedStream, valueSize, keyContext, log.getSerializedUids(),
```
alexandre-merle: This call not needed to be done after the md5 calculation, in the done callback ? :)

lib/utilities/MD5Sum.js
```diff
@@ -0,0 +1,24 @@
...
+        this.done = done;
+    }
+
+    _transform(chunk, encoding, callback) {
+        this.hash.update(chunk, encoding);
+        callback(null, chunk, encoding);
+    }
+
+    _flush(callback) {
+        this.done(null, this.hash.digest('hex'));
```
alexandre-merle: If the done callback we never get back an error, we can just remove the parameter :)

lib/utilities/MD5Sum.js
```diff
@@ -0,0 +1,24 @@
+const Transform = require('stream').Transform;
+const crypto = require('crypto');
+
+class MD5Sum extends Transform {
```
alexandre-merle: Can you add a comment here, about the design of the done callback, saying the done callback need to be synchronous and will be called before the stream end to avoid any strange behavior with this :)

lib/utilities/MD5Sum.js
alexandre-merle: the done callback is now sending to the constructor, and callback is used as usual in functions :)

alexandre-merle:
@ironman-machine let's check this

alexandre-merle:
End to end passed http://ci.ironmann.io/gh/scality/Integration/2486 :)

alexandre-merle:
@ironman-machine let's check this

SCALITY_METADATA_BRANCH=clean/md5

alexandre-merle:
End to end with metadata passing :)

LaurenSpiegel:
👍

ThibaultRiviere:
It can be good to have the md5sum class in the Arsenal repo. 
:+1:

alexandre-merle:
@ThibaultRiviere good idea, will be done after the release, you can open an issue for that :)

PR #98, opened 2016-07-25
Perf/in memory

ploki:
nice to have (may speedup tests)
~~this branch depends on clean/md5~~

### DIFF
--- a/lib/data/in_memory/backend.js
+++ b/lib/data/in_memory/backend.js
@@ -1,5 +1,18 @@
 import { errors } from 'arsenal';
 import stream from 'stream';
+import config from '../../Config';
+import { Logger } from 'werelogs';
+
+const logger = new Logger('MemDataBackend', {
+    logLevel: config.log.logLevel,
+    dumpLevel: config.log.dumpLevel,
+});
+
+function createLogger(reqUids) {
+    return reqUids ?
+        logger.newRequestLoggerFromSerializedUids(reqUids) :
+        logger.newRequestLogger();
+}
 
 export const ds = [];
 let count = 1; // keys are assessed with if (!key)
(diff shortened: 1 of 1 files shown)


lib/data/in_memory/backend.js
```diff
@@ -10,14 +9,26 @@ export function resetCount() {
...
+            }
+            cursor += data.length;
+        })
         .on('end', () => {
-            request.calculatedHash = crypto.createHash('md5')
-                                           .update(value).digest('hex');
-            ds[count] = { value, keyContext };
-            callback(null, count++);
+            if (exceeded) {
+                callback('Data stream exceed announced size');
```
michael-zapata: We use arsenal errors as the error in callbacks, strings are not meaningful enough.

lib/data/file/backend.js
```diff
@@ -68,9 +66,7 @@ export const backend = {
             })
             .on('end', () => {
                 // eslint-disable-next-line no-param-reassign
```
michael-zapata: No need for that line since you removed the actual problem here :dancer:

lib/data/in_memory/backend.js
```diff
@@ -10,14 +9,26 @@ export function resetCount() {
...
+    put: function putMem(request, size, keyContext, reqUids, callback) {
+        const value = new Buffer(size);
+        let cursor = 0;
+        let exceeded = false;
+        request.on('data', data => {
+            if (cursor + data.length > size) {
+                exceeded = true;
+            }
+            if (!exceeded) {
+                data.copy(value, cursor);
```
ThibaultRiviere: You can also do : 
```js
    cursor += data.copy(value, cursor); 
```

ploki: Nope, it can't. the increment is not conditional in the logic of the code

ThibaultRiviere: I don't understand because in case of exceeded you don't use ```cursor``` anymore . So why still increment it ?

lib/api/completeMultipartUpload.js
```diff
@@ -212,7 +212,8 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 const partSize = Number.parseInt(storedPartsAsObjects[i]
                     .size, 10);
                 if (i < jsonList.Part.length - 1 &&
-                    partSize < 5242880) {
+                    partSize < constants.minimumAllowedPartSize &&
+                    request.relaxedSizeCheckingForTests !== true) {
```
michael-zapata: Still adding conditionals in code for tests?

ploki: Nope, it's part of another pool request which has to be fixed

lib/data/in_memory/backend.js
ploki: to log announced and received size

PR #100, opened 2016-07-27
Use refactored Arsenal's Auth API - Step 1

DavidPineauScality:
This is the first step towards using the cleaned-up Auth API from arsenal.
The work is still ongoing, but this will allow us to avoid falling out-of-sync.

Depends on: https://github.com/scality/Arsenal/pull/119 then https://github.com/scality/vaultclient/pull/94

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -26,7 +26,7 @@ import prepareRequestContexts from
 import serviceGet from './serviceGet';
 import vault from '../auth/vault';
 
-auth.setAuthHandler(vault);
+auth.setHandler(vault);
 
 const api = {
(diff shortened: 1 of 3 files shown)


alexandre-merle:
The vault backend in lib/auth/vault.js is not updated to reflect the changes here, was update on the file inside arsenal, but it's the wrong one

AntoninCoulibaly:
updated

AntoninCoulibaly:
@ironman-machine 

SCALITY_VAULT_BRANCH=dev/cleanup/AuthAPI-Step1

michael-zapata:
:+1:

PR #110, opened 2016-08-02
Ft/kms

ploki:
Key Management System (for server side encryption usage)

please, review the last commit, as others are part of different pull requests.


lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+            }
+            return cb(err, cipherBundle);
+        });
+    }
+    /*
+     * cb(err, decipherBundle)
+     */
+    static createDecipherBundle(serverSideEncryption, offset,
+                                log, cb) {
+        const decipherBundle = {
```
ThibaultRiviere: Maybe you can declare the variable after check the argument

ThibaultRiviere:
:+1:

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
+import crypto from 'crypto';
+import { errors } from 'arsenal';
+
+export default class common {
```
LaurenSpiegel: capitalize class name "Common"

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+                });
+        } else {
+            log.error('Unknown cryptographic scheme', { cryptoScheme });
+            cb(errors.InternalError);
+        }
+    }
+
+    /*
+     * cb(err, decipher: ReadWritable.stream)
+     */
```
LaurenSpiegel: Please add full jsdoc

lib/kms/file/backend.js
```diff
@@ -0,0 +1,80 @@
+import common from '../common';
+
+export const backend = {
+    /*
+     * Target implementation will be async in some aspects. let mimic it
```
LaurenSpiegel: in some aspects?

let s/b let's

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+            return cb(null, masterKeyId);
+        });
+    }
+
+    /*
+     * cb(err, serverSideEncryptionInfo: object)masterKeyId: string)
+     */
+    static bucketLevelEncryption(bucketName, headers, log, cb) {
+        const sseAlgorithm = headers['x-amz-scal-server-side-encryption'];
+        const sseMasterKeyId =
```
LaurenSpiegel: do they need to send this?

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+    static _IVSize() {
+        return 16;
+    }
+
+    /* block size is 128bit for AES-256-CTR */
+    static _aesBlockSize() {
+        return 16;
+    }
+
+    static createDataKey() {
```
LaurenSpiegel: Let's add a comment:
Creates data key to encrypt and decrypt the actual data (which data key is ciphered and deciphered by the appliance).  The encrypted data key is stored in the object's metadata. We also use this function to create the bucket key for the file and in memory implementations.

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+            /* we are not storing hashed human password.
+             * It's a random key, so 1 iteration and
+             * a fixed salt is enough for our usecase.
+             * don't change these without bumping
+             * scheme number
+             */
+            const salt = new Buffer('ItsTasty');
+            const iterations = 1;
+            crypto.pbkdf2(
+                dataKey, salt, iterations,
```
LaurenSpiegel: Let's add a comment:
dataKey here is the unencrypted key (either from the appliance on a get or originally generated by kms in the case of a put)

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+                this._keySize(), 'sha1', (err, derivedKey) => {
+                    if (err) {
+                        log.error('pbkdf2 function failed on key derivation',
+                                  { err });
+                        cb(errors.InternalError);
+                        return;
+                    }
+                    crypto.pbkdf2(
+                        derivedKey, salt, iterations,
```
LaurenSpiegel: derivedKey is the actual data encryption or decryption key used in the AES ctr cipher

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+
+    /*
+     * cb(err, derivedKey, derivedIV)
+     */
+    static _deriveKey(cryptoScheme, dataKey, log, cb) {
+        if (cryptoScheme <= 1) {
+            /* we are not storing hashed human password.
+             * It's a random key, so 1 iteration and
+             * a fixed salt is enough for our usecase.
+             * don't change these without bumping
```
LaurenSpiegel: instead of 'these': the salt, the iteration number or the digest algorithm (sha1 here) without bumping the scheme number saved in the object metadata along with the dataKey.

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+    /* block size is 128bit for AES-256-CTR */
+    static _aesBlockSize() {
+        return 16;
+    }
+
+    static createDataKey() {
+        return new Buffer(crypto.randomBytes(this._keySize()));
+    }
+
+    static _incrementIV(derivedIV, counter) {
```
LaurenSpiegel: please add unit tests for this function

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+                    return;
+                }
+                const aesBlockSize = this._aesBlockSize();
+                const blocks = Math.floor(offset / aesBlockSize);
+                const toSkip = offset % aesBlockSize;
+                const iv = this._incrementIV(derivedIV, blocks);
+                const cipher = crypto.createDecipheriv(this._algorithm(),
+                                                       derivedKey, iv);
+                if (toSkip) {
+                    /* seek within the block to finish cipher initialization */
```
LaurenSpiegel: Above, we advanced to the latest boundary not greater than the offset amount.  Here we advance by the toSkip amount if necessary.

lib/kms/common.js
```diff
@@ -0,0 +1,122 @@
...
+                    cipher.write(dummyBuffer);
+                    cipher.read();
+                }
+                cb(null, cipher);
+            });
+    }
+
+    /*
+     * cb(err, cipher: ReadWritable.stream)
+     */
```
LaurenSpiegel: Included to give us flexibility for different functions for cipher and decipher.

lib/kms/file/backend.js
```diff
@@ -0,0 +1,80 @@
...
+     */
+
+    /*
+     * cb(err, masterKeyId: string)
+     */
```
LaurenSpiegel: Add comment: 
Using createDataKey here for purposes of createBucketKeyMem so that we do not need a separate function.

or just use random bytes.

lib/kms/file/backend.js
```diff
@@ -0,0 +1,80 @@
...
+            const newKey = common.createDataKey().toString('hex');
+            cb(null, newKey);
+        });
+    },
+
+    /*
+     * cb(err, cipheredDataKey: Buffer)
+     */
+    cipherDataKey: function cipherDataKeyMem(cryptoScheme,
+                                             masterKeyId,
```
LaurenSpiegel: masterKeyId -- stored with the bucket metadata.  for file backend, the masterKeyId is the actual bucket key rather than the key to retrieve the actual key from a dictionary

lib/kms/file/backend.js
```diff
@@ -0,0 +1,80 @@
...
+            common.createCipher(
+                cryptoScheme, masterKey, 0, log,
+                (err, cipher) => {
+                    if (err) {
+                        cb(err);
+                        return;
+                    }
+                    let cipheredDataKey =
+                            cipher.update(plainTextDataKey);
+                    const final = cipher.final();
```
LaurenSpiegel: Here to ensure that any bytes remaining in the output of the stream are captured

lib/kms/in_memory/backend.js
```diff
@@ -0,0 +1,81 @@
...
+    /*
+     * Target implementation will be async in some aspects. let mimic it
+     */
+
+    /*
+     * cb(err, masterKeyId: string)
+     */
+    createBucketKey: function createBucketKeyMem(bucketName, log, cb) {
+        process.nextTick(() => {
+            kms[count] = common.createDataKey();
```
LaurenSpiegel: either comment that using the object data key for convenience or just create random bytes

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+    client = file;
+    implName = 'fileKms';
+} else if (config.backends.kms === 'scality') {
+    throw new Error('KMS backend not implemented');
+} else {
+    throw new Error('KMS backend is not configured');
+}
+
+
+export default class kms {
```
LaurenSpiegel: KMS

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+        const sseMasterKeyId =
+                  headers['x-amz-scal-server-side-encryption-aws-kms-key-id'];
+        /*
+         * the purpose of bucket level encryption is to exempt client
+         * from sending appropriate headers to trigger encryption.
+         * Since server side encryption with customer provided key needs
+         * the key for every access to the ciphered objects and we don't
+         * want to store it in the bucket metadata. It is not feasible
+         * in bucket level encryption.
+         */
```
LaurenSpiegel: The purpose of bucket level encryption is so that the client does not have to send appropriate headers to trigger encryption on each object put in an "encrypted bucket".
Customer provided keys are not feasible in this system because we do not want to store this key in the bucket metadata.

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+                    return cb(err);
+                }
+                log.trace('cipher created by the kms');
+                return cb(null, cipher);
+            });
+    }
+
+    /*
+     * cb(err, decipher: ReadWritable.stream)
+     */
```
LaurenSpiegel: maybe call common.createDecipher directly from the code without this extra function layer

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+            cb(null, serverSideEncryptionInfo);
+        } else {
+            /*
+             * no encryption
+             */
+            cb(null, null);
+        }
+    }
+
+    static createDataKey(log) {
```
LaurenSpiegel: is this supposed to mimic an async function?  It's sort of odd to have this function rather than just calling common.createDataKey() directly in createCipherBundle function below

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+                }
+                log.trace('decipher created by the kms');
+                return cb(null, decipher);
+            });
+    }
+
+    /*
+     * cb(err, cipherBundle)
+     */
+    static createCipherBundle(serverSideEncryption,
```
LaurenSpiegel: I think it would be clearer to call this serverSideEncryptionInfo as you did above.

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+            function cipherDataKey(next) {
+                return kms._cipherDataKey(cipherBundle.cryptoScheme,
+                                          serverSideEncryption.masterKeyId,
+                                          dataKey, log, next);
+            },
+            function createCipher(cipheredDataKey, next) {
+                cipherBundle.cipheredDataKey =
+                    cipheredDataKey.toString('base64');
+                return kms._createCipher(cipherBundle.cryptoScheme,
+                                         dataKey, 0, log, next);
```
LaurenSpiegel: should this be cipheredDataKey or dataKey?

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+                log.error('error processing cipher bundle',
+                          { implName, error: err });
+            }
+            return cb(err, cipherBundle);
+        });
+    }
+    /*
+     * cb(err, decipherBundle)
+     */
+    static createDecipherBundle(serverSideEncryption, offset,
```
LaurenSpiegel: same as above -- serverSideEncryptionInfo

lib/kms/wrapper.js
```diff
@@ -0,0 +1,246 @@
...
+                    serverSideEncryption.masterKeyId,
+                    serverSideEncryption.cipheredDataKey,
+                    log, next);
+            },
+            function createDecipher(plainTextDataKey, next) {
+                return kms._createDecipher(decipherBundle.cryptoScheme,
+                                           plainTextDataKey,
+                                           offset, log, next);
+            },
+            function finishDecipherBundle(decipher, next) {
```
LaurenSpiegel: You could combine this with the final callback (or leave as is if you think clearer).

tests/unit/kms.js
```diff
@@ -0,0 +1,181 @@
...
+                assert.strictEqual(err, null);
+                assert.strictEqual(sseInfo.cryptoScheme, 1);
+                assert.strictEqual(sseInfo.mandatory, true);
+                assert.strictEqual(sseInfo.algorithm, 'aws:kms');
+                assert.strictEqual(sseInfo.masterKeyId, masterKeyId);
+                done();
+            });
+    });
+
+    it('should not construct a sse info object on garbage', done => {
```
LaurenSpiegel: 'should not construct a sse info object if x-amz-scal-server-side-encryption header contains invalid algorithm option'

tests/unit/kms.js
```diff
@@ -0,0 +1,181 @@
...
+        kms.bucketLevelEncryption(
+            'dummyBucket', headers, log,
+            (err, sseInfo) => {
+                assert.strictEqual(err, null);
+                assert.strictEqual(sseInfo, null);
+                done();
+            });
+    });
+
+    it('should not construct a sse info object on no header', done => {
```
LaurenSpiegel: 'should not construct a sse info object if no x-amz-scal-server-side-encryption header included with request'

tests/unit/kms.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    it('should cipher and decipher a datastream', done => {
+        _utestCreateBundlePair(log, (err, cipherBundle, decipherBundle) => {
+            assert.strictEqual(err, null);
+            cipherBundle.cipher.pipe(decipherBundle.decipher);
+            const target = new Buffer(10000);
+            target.fill('e');
+            cipherBundle.cipher.write(target);
+            const result = decipherBundle.decipher.read();
```
LaurenSpiegel: I believe this will just read what is contained in the readable stream buffer.  So, you are relying on the high water mark default size of 16 KB for the test to pass.  You should note that or put a listener on the stream and concatenate the result.

LaurenSpiegel:
@ironman-machine cool with this?

LaurenSpiegel:
@ironman-machine let's start fresh

lib/kms/common.js
```diff
@@ -0,0 +1,167 @@
...
+             * metadata along with the dataKey.
+             */
+            const salt = new Buffer('ItsTasty');
+            const iterations = 1;
+            crypto.pbkdf2(
+                dataKey, salt, iterations,
+                this._keySize(), 'sha1', (err, derivedKey) => {
+                    if (err) {
+                        log.error('pbkdf2 function failed on key derivation',
+                                  { err });
```
rahulreddy: Use error as key for err

lib/kms/common.js
```diff
@@ -0,0 +1,167 @@
...
+                        cb(errors.InternalError);
```
rahulreddy: return callbacks instead if returning undefined.

lib/kms/common.js
```diff
@@ -0,0 +1,167 @@
...
+                            if (err) {
+                                log.error(
+                                    'pbkdf2 function failed on IV derivation',
+                                    { err });
+                                cb(errors.InternalError);
+                                return;
+                            }
+                            // derivedKey is the actual data encryption or
+                            // decryption key used in the AES ctr cipher
+                            cb(null, derivedKey, derivedIV);
```
rahulreddy: return callback

lib/kms/common.js
```diff
@@ -0,0 +1,167 @@
...
+                        });
+                });
+        } else {
+            log.error('Unknown cryptographic scheme', { cryptoScheme });
+            cb(errors.InternalError);
```
rahulreddy: Return callback here as well.

lib/kms/wrapper.js
```diff
@@ -0,0 +1,255 @@
...
+                algorithm: sseAlgorithm,
+                masterKeyId: sseMasterKeyId,
+                mandatory: true,
+            };
+            cb(null, serverSideEncryptionInfo);
+        } else {
+            /*
+             * no encryption
+             */
+            cb(null, null);
```
rahulreddy: If you are just checking for falsey values `return cb()` should be sufficient.

lib/kms/file/backend.js
```diff
@@ -0,0 +1,107 @@
...
+                    }
+                    cb(null, cipheredDataKey);
+                });
+        });
+    },
+
+     /**
+      *
+      * @param {number} cryptoScheme - crypto scheme version number
+      * @param {string} masterKeyId - key to retrieve master key
```
alexandre-merle: Confusing, we named it masterKeyId, but the key is encrypted and use to retrieve the master key

alexandre-merle:
Minor stuff, but really good job :) :+1: (i will update the PR following the remaining comments)

lib/kms/common.js
```diff
@@ -0,0 +1,165 @@
+import crypto from 'crypto';
+import { errors } from 'arsenal';
+
+/** Class exposing common createDataKey,
+createDecipher and createCipher functions. */
+export default class Common {
```
michael-zapata: If everything is static, why do a class instead of an object?

lib/kms/file/backend.js
```diff
@@ -0,0 +1,109 @@
...
+      * @param {string} masterKeyId - master key; for the file backend
+      * the master key is the actual bucket master key rather than the key to
+      * retrieve the actual key from a dictionary
+      * @param {buffer} plainTextDataKey - data key
+      * @param {object} log - logger object
+      * @param {function} cb - callback
+      * @returns {undefined}
+      * @callback called with (err, cipheredDataKey: Buffer)
+      */
+    cipherDataKey: function cipherDataKeyMem(cryptoScheme,
```
michael-zapata: decipher is identical aside from a single call, could we just parametrize the call?

michael-zapata: Or well, the callback to that different function is similar, you can pretty much extract it.

lib/kms/common.js
```diff
@@ -0,0 +1,165 @@
...
+     * @param {number} counter - quotient of the offset and blocksize
+     * @return {buffer} - the incremented IV
+     */
+    static _incrementIV(derivedIV, counter) {
+        const newIV = derivedIV;
+        const len = derivedIV.length;
+        let i = len - 1;
+        let ctr = counter;
+        while (ctr !== 0) {
+            const mod = (ctr + newIV[i]) % 256;
```
MathieuCassagne: I think the variable "mod' is not updated at each iteration, it's a const variable.
```es6
➜  ~ cat test.js 
var i = 0;
while (i < 10) {
	const a = i;
	++i;
	console.log(a);
}
➜  ~ node test.js 
0
0
0
0
0
0
0
0
0
0
```
```es6
➜  ~ cat test.js 
var i = 0;
while (i < 10) {
	var a = i;
	++i;
	console.log(a);
}
➜  ~ node test.js
0
1
2
3
4
5
6
7
8
9
```

lib/kms/in_memory/backend.js
```diff
@@ -0,0 +1,104 @@
...
+      */
+    cipherDataKey: function cipherDataKeyMem(cryptoScheme,
+                                             masterKeyId,
+                                             plainTextDataKey,
+                                             log,
+                                             cb) {
+        process.nextTick(() => {
+            Common.createCipher(
+                cryptoScheme, kms[masterKeyId], 0, log,
+                (err, cipher) => {
```
michael-zapata: Still the same callback.

lib/kms/wrapper.js
```diff
@@ -0,0 +1,255 @@
...
+            !serverSideEncryptionInfo.cryptoScheme) {
+            log.error('Invalid cryptographic information', { implName });
+            return cb(errors.InternalError);
+        }
+        const decipherBundle = {
+            cryptoScheme: serverSideEncryptionInfo.cryptoScheme,
+            decipher: null,
+        };
+        return async.waterfall([
+            function decipherDataKey(next) {
```
michael-zapata: This function and the following ones are using parameters that are out of their own scope, it makes it hard to follow, and means we cannot rely on the function signature.

lib/kms/common.js
```diff
@@ -0,0 +1,165 @@
...
+      * @returns {undefined}
+      * @callback called with (err, decipher: ReadWritable.stream)
+      */
+    static createDecipher(cryptoScheme, dataKey, offset, log, cb) {
+        this._deriveKey(
+            cryptoScheme, dataKey, log,
+            (err, derivedKey, derivedIV) => {
+                if (err) {
+                    log.warn('key derivation failed', { error: err });
+                    cb(err);
```
MathieuCassagne: Why not ```return cb(err);``` ?

lib/kms/common.js
alexandre-merle: but for each iteration you can create the variable :) anyway you first example seem a little bit strange ^^

alexandre-merle: try to use 'use strict' and 'let i = 0' :)

lib/kms/common.js
alexandre-merle: both are okay :)

lib/kms/file/backend.js
alexandre-merle: i don't get it :) what do you mean ? :)

lib/kms/common.js
MathieuCassagne: Oh yes you're right, by using "use strict" there is a new variable at each iteration, my bad :)

alexandre-merle: :)

MathieuCassagne:
:+1:

alexandre-merle:
@michael-zapata  The modification for code duplication will be done in the final code, because this PR is actually the base PR for three others ones

michael-zapata:
:+1: then.

alexandre-merle:
Thank's :) Merging on behalf of @LaurenSpiegel

alexandre-merle:
Squashing first :)

PR #119, opened 2016-08-06
scality-kms integration and bug fixes

ploki:
- S3KMS environment variable added
- bucketKeyId as a number breaks sanity checks in BucketInfo
- scality-kms integration + fallback to file if not available
- location array in objectmd may be undefined

### DIFF
--- a/lib/kms/wrapper.js
+++ b/lib/kms/wrapper.js
@@ -1,22 +1,37 @@
+import async from 'async';
+
+import { errors } from 'arsenal';
+
 import config from '../Config';
+import { logger } from '../utilities/logger';
 import inMemory from './in_memory/backend';
 import file from './file/backend';
 import Common from './common';
-import async from 'async';
-import { errors } from 'arsenal';
+let scalityKMS;
+let scalityKMSImpl;
+try {
+    const ScalityKMS = require('scality-kms');
+    scalityKMS = new ScalityKMS(config.kms);
+    scalityKMSImpl = 'scalityKms';
+} catch (err) {
+    logger.warn('scality kms selected but unavailable, using file backend');
+    scalityKMS = file;
+    scalityKMSImpl = 'fileKms';
+}
+
 
 let client;
 let implName;
 
-// TODO: Remove when kms backend is implemented
-if (config.backends.kms === 'mem' || config.backends.kms === 'scality') {
+if (config.backends.kms === 'mem') {
     client = inMemory;
     implName = 'memoryKms';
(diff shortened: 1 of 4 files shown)


alexandre-merle:
LGTM :+1:

alexandre-merle:
@ironman-machine check that :)

alexandre-merle:
end to end passing :)

lib/kms/wrapper.js
```diff
@@ -1,22 +1,37 @@
...
+const logger = require('../utilities/logger.js').logger;
 import inMemory from './in_memory/backend';
 import file from './file/backend';
 import Common from './common';
-import async from 'async';
-import { errors } from 'arsenal';
+let scalityKMS;
+let scalityKMSImpl;
+try {
+    const ScalityKMS = require('scality-kms');
```
LaurenSpiegel: how does this work without changing the package.json?

lib/kms/wrapper.js
```diff
@@ -1,22 +1,37 @@
+import async from 'async';
+
+import { errors } from 'arsenal';
+
 import config from '../Config';
+const logger = require('../utilities/logger.js').logger;
```
michael-zapata: Why not use the ES6 syntax?

michael-zapata:
Could we have a non-regression test available?

ploki:
@michael-zapata what do you mean?

GiorgioRegni:
We need functional tests to go along with the PR, I guess that's what @michael-zapata means

ploki:
functional tests are included in #110

michael-zapata:
Hello,

To me, there is two concerns:
- Since this is a fix, it would be welcome to add a test confirming that the problem has been fixed (This is what I meant by non-regression)
- The PR you link to only has unit testing, and @GiorgioRegni addresses my other concern: we should have something that would actually test the features using an end-to-end scenario, which is something we actually do with functional tests

ploki:
> - bucketKeyId as a number breaks sanity checks in BucketInfo

sanity checks did work, mission accomplished

> - location array in objectmd may be undefined

spotted by existing tests on zero sized objects

The point here is that a large part of the "GET/PUT" unit tests rely on weak assumptions regarding how the code react in a given context and cannot be reused as is with encryption enabled.
Tests already exist but must be reworked and it is beyond the scope of this PR.

Are the end-to-end tests implemented in this repository?

vrancurel:
@ironman-machine check that :)

vrancurel:
👍

PR #121, opened 2016-08-06
Convert in memory bucket key to string to store.

LaurenSpiegel:
The in memory implementation is using a counter so the id was being stored as a number.  This was causing the assertion in BucketInfo.js (checks for string) to fail.


alexandre-merle:
Already fix in https://github.com/scality/S3/pull/119

PR #127, opened 2016-08-09
Only new metadata stored when obj put overwrite

### LINKED ISSUE #126: Mixed metadata on object put override
When putting an object with the same key, we are not overwriting all user metadata if the new user metadata and the old user metadata do not have the same keys.

LaurenSpiegel:
This closes #126.


LaurenSpiegel:
@ironman-machine work to do...

rahulreddy:
👍 
Please add BF tag to the commit before you merge.

LaurenSpiegel:
@ironman-machine try that again

alexandre-merle:
LGTM :+1:

PR #132, opened 2016-08-10
FIX: Object overwrite test

alexandre-merle:
This test is not working properly, because the
promise is not return to mocha.
This Pull request fix that

Fixes https://github.com/scality/S3/issues/131


LaurenSpiegel:
👍

rahulreddy:
👍

PR #137, opened 2016-08-12

nicolas2bert:
When setting the 'public-read' ACL permission of the object, any account can get an object (objectGet) / get its metadata (objectHead) **even if the account has no right to the bucket**

### DIFF
--- a/lib/api/apiUtils/authorization/aclChecks.js
+++ b/lib/api/apiUtils/authorization/aclChecks.js
@@ -8,8 +8,7 @@ export function isBucketAuthorized(bucket, requestType, canonicalID) {
         return true;
     }
     const bucketAcl = bucket.getAcl();
-    if (requestType === 'bucketGet' || requestType === 'bucketHead'
-            || requestType === 'objectGet' || requestType === 'objectHead') {
+    if (requestType === 'bucketGet' || requestType === 'bucketHead') {
         if (bucketAcl.Canned === 'public-read'
             || bucketAcl.Canned === 'public-read-write'
             || (bucketAcl.Canned === 'authenticated-read'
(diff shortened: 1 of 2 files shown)


lib/api/apiUtils/authorization/aclChecks.js
```diff
@@ -53,7 +52,11 @@ export function isBucketAuthorized(bucket, requestType, canonicalID) {
     // holding the object.  So, if the requst type is objectPutACL
     // or objectGetACL, the bucket authorization check should just
     // return true so can move on to check rights at the object level.
-    return (requestType === 'objectPutACL' || requestType === 'objectGetACL');
+
+    // An account can also get an object(objectGet) / get its metadata
+    // (objectHead) even if the account has no right to the bucket
```
LaurenSpiegel: Please combine comment with comment above.

tests/unit/api/bucketACLauth.js
```diff
@@ -26,7 +26,7 @@ describe('bucket authorization for bucketGet, bucketHead, ' +
     const requestTypes = ['bucketGet', 'bucketHead', 'objectGet', 'objectHead'];
 
     const trueArray = [true, true, true, true];
-    const falseArray = [false, false, false, false];
+    const falseArrayBucketTrueArrayObject = [false, false, true, true];
```
LaurenSpiegel: please add a comment as to what these means

tests/unit/api/bucketACLauth.js
```diff
@@ -37,14 +37,16 @@ describe('bucket authorization for bucketGet, bucketHead, ' +
           canned: 'public-read-write', id: accountToVet, response: trueArray },
         { it: 'should not allow access to public user if authenticated-read ' +
```
LaurenSpiegel: this it statement needs to be updated since now you have some true responses

tests/unit/api/bucketACLauth.js
```diff
@@ -37,14 +37,16 @@ describe('bucket authorization for bucketGet, bucketHead, ' +
...
           'ACL', canned: 'authenticated-read', id: constants.publicId,
-          response: falseArray },
+          response: falseArrayBucketTrueArrayObject },
         { it: 'should allow access to any authenticated user if authenticated' +
           '-read ACL', canned: 'authenticated-read', id: accountToVet,
           response: trueArray },
         { it: 'should not allow access to public user if private canned ACL',
```
LaurenSpiegel: same as above

tests/unit/api/bucketACLauth.js
```diff
@@ -37,14 +37,16 @@ describe('bucket authorization for bucketGet, bucketHead, ' +
...
-          canned: '', id: accountToVet, response: falseArray },
+          canned: '', id: accountToVet,
+          response: falseArrayBucketTrueArrayObject },
         { it: 'should not allow access to just any user if private canned ACL',
```
LaurenSpiegel: same as above

alexandre-merle:
LGTM :+1:

LaurenSpiegel:
👍 but let's do it against rel/1.1

rahulreddy:
Also please don't forget to mention the issue in the commit body.

PR #138, opened 2016-08-12
Object protected by erasure codes

lamphamsy:
Object sent from S3 Server will be protected by use a (k, m) erasure codes.
Each object is encoded into (k + m) fragments of the same size. These
fragments will be sent to object client to store them. Each fragment has its
own key. The object's key is a string composed of (k + m) keys of its fragments.

There are put/get/delete operations that use erasure coding techniques.
Eclib library is used to perform coding operations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.2.0",
     "babel-plugin-transform-es2015-parameters": "^6.2.0",
     "bucketclient": "scality/bucketclient#rel/1.1",
+    "eclib": "scality/eclib#master",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.2.0",
     "babel-plugin-transform-es2015-parameters": "^6.2.0",
     "bucketclient": "scality/bucketclient#rel/1.1",
```
LaurenSpiegel: Why are you working off of rel/1.1?

lamphamsy: I'm sorry. The PR is moved to #147 . Thanks

PR #139, opened 2016-08-12

nicolas2bert:
When setting the 'public-read' ACL permission of the object, any account can get an object (objectGet) / get its metadata (objectHead) even if the account has no right to the bucket


alexandre-merle:
again :+1:

PR #140, opened 2016-08-12
Implement object copy.

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine wanna copy some objects?

lib/api/objectCopy.js
```diff
@@ -0,0 +1,288 @@
...
+                    if (flag) {
+                        log.trace('deleted flag or transient flag ' +
+                        'on destination bucket', { flag });
+                        return next(errors.NoSuchBucket);
+                    }
+                    return next(null, storeMetadataParams,
+                        dataLocator, destBucketMD, destObjMD);
+                });
+        },
+        function getCipherBundle(storeMetadataParams, dataLocator, destBucketMD,
```
ploki: the cipher bundle should be created for each part. the crypto stream will not be reusable across all data.put calls

lib/api/objectCopy.js
```diff
@@ -0,0 +1,288 @@
...
+             // argument here to increase the number of parts
+             // copied at once.
+            return async.mapLimit(dataLocator, 1,
+                // eslint-disable-next-line prefer-arrow-callback
+                function copyPart(part, cb) {
+                    return data.get(part, log, (err, stream) => {
+                        if (err) {
+                            return cb(err);
+                        }
+                        return data.put(cipherBundle, stream, part.size,
```
ploki: I think enclosing this call in the createCipherBundle callback would be in order

ploki: as a consequence of this, cipheredDataKey from the cipher bundle should be kept per part and registered in the destination's location array

lib/api/objectCopy.js
LaurenSpiegel: what do you think about getting an array of cipher bundles at once with a length equal to the number of parts?  would that work?

ploki: getting a cipher bundle implies a network communication with the KMS. It may be best amortized close to the data.put. e.g. data transfers starting independently of other parts being ready regarding there cipherbundle

lib/api/objectCopy.js
```diff
@@ -0,0 +1,288 @@
...
+        },
+        function goGetData(storeMetadataParams, dataLocator,
+            destObjMD, cipherBundle, next) {
+            // skip if source and dest the same or 0 byte object
+            if (sourceIsDestination || dataLocator.length === 0) {
+                return next(null, storeMetadataParams, dataLocator, destObjMD,
+                    cipherBundle);
+            }
+             // dataLocator is an array.  need to get and put all parts
+             // For now, copy 1 part at a time. Could increase the second
```
ploki: I think, the copy is concurrent. What is made serially here is the connection of streams ends  with ```pipe()``` methods. Once streams are connected they are ready to be consumed by the node event loop

LaurenSpiegel: as discussed, the async.mapLimit is finishing the put for each one at a time.

LaurenSpiegel:
@ironman-machine like this better?

LaurenSpiegel:
end to end passed

lib/api/objectCopy.js
```diff
@@ -0,0 +1,306 @@
...
+function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
+    objectKey, log) {
+    let whichMetadata = headers['x-amz-metadata-directive'];
+    // Default is COPY
+    whichMetadata = whichMetadata === undefined ? 'COPY' : whichMetadata;
+    if (whichMetadata !== 'COPY' && whichMetadata !== 'REPLACE') {
+        return { error: errors.InvalidArgument };
+    }
+    const overrideMetadata = {};
+    if (headers['x-amz-server-side-encryption']) {
```
alexandre-merle: you can remove the if here :)

lib/api/objectCopy.js
```diff
@@ -0,0 +1,306 @@
...
+        overrideMetadata['x-amz-server-side-encryption'] =
+            headers['x-amz-server-side-encryption'];
+    }
+    if (headers['x-amz-storage-class']) {
```
alexandre-merle: same :)

lib/api/objectCopy.js
```diff
@@ -0,0 +1,306 @@
...
+        overrideMetadata['x-amz-storage-class'] =
+            headers['x-amz-storage-class'];
+    }
+    if (headers['x-amz-website-redirect-location']) {
```
alexandre-merle: same ^^

lib/api/objectCopy.js
```diff
@@ -0,0 +1,306 @@
...
+                        log.debug('no source object', { sourceObject });
+                        return next(errors.NoSuchKey);
+                    }
+                    const headerValResult =
+                        validateHeaders(sourceObjMD, request.headers);
+                    if (headerValResult.error) {
+                        return next(errors.PreconditionFailed);
+                    }
+
+                    if (headerValResult.error) {
```
alexandre-merle: you can remove this one :)

lib/api/objectCopy.js
```diff
@@ -0,0 +1,306 @@
...
+        const xml = [
+            '<?xml version="1.0" encoding="UTF-8"?>',
+            '<CopyObjectResult>',
+            '<LastModified>', new Date(storeMetadataParams.lastModifiedDate)
+                .toUTCString(), '</LastModified>',
+            '<ETag>&quot;', storeMetadataParams.contentMD5, '&quot;</ETag>',
+            '</CopyObjectResult>',
+        ].join('');
+        const additionalHeaders = {};
+        if (serverSideEncryption) {
```
alexandre-merle: may be we need to check here same as https://github.com/scality/S3/pull/146, to send back the key only for custom key id ?

alexandre-merle:
Some small comments, otherwise, LGTM :+1:

lib/api/objectCopy.js
LaurenSpiegel: I check the length of the keys of overrideMetadata below to see if any of these have been added so I can't assign these here if no data.

lib/api/objectCopy.js
LaurenSpiegel: good catch

lib/api/api.js
```diff
@@ -26,11 +29,50 @@ auth.setAuthHandler(vault);
...
+            }
+            // Pull the source bucket and source object separated by /
+            const sourceBucket = source.slice(0, slashSeparator);
+            const sourceObject = source.slice(slashSeparator + 1);
+            const getRequestContext = new RequestContext(request.headers,
+                request.query, sourceBucket, sourceObject,
+                request.socket.remoteAddress, request.connection.encrypted,
+                'objectGet', 's3', locationConstraint);
+
+            return auth.doAuth(request, log, err => {
```
rahulreddy: Why not use `async.parallel` since both auth can be performed at the same time?

lib/api/objectCopy.js
```diff
@@ -0,0 +1,304 @@
...
+    }
+    if (headers['x-amz-website-redirect-location']) {
+        overrideMetadata['x-amz-website-redirect-location'] =
+            headers['x-amz-website-redirect-location'];
+    }
+    // Cannot copy from same source and destination if no MD
+    // changed
+    if (sourceIsDestination && whichMetadata === 'COPY' &&
+        Object.keys(overrideMetadata).length === 0) {
+        return { error: errors.InvalidRequest };
```
rahulreddy: Let's use Arsenal's custom description here as the default desc. relates to SOAP requests. AWS is using

> This copy request is illegal because it is trying to copy an object to itself without changing the object\'s metadata, storage class, website redirect location or encryption attributes.

lib/api/objectCopy.js
```diff
@@ -0,0 +1,304 @@
...
+                return next(null, storeMetadataParams, dataLocator, destObjMD,
+                    serverSideEncryption);
+            }
+             // dataLocator is an array.  need to get and put all parts
+             // For now, copy 1 part at a time. Could increase the second
+             // argument here to increase the number of parts
+             // copied at once.
+            return async.mapLimit(dataLocator, 1,
+                // eslint-disable-next-line prefer-arrow-callback
+                function copyPart(part, cb) {
```
rahulreddy: I think it's worthwhile separating these functions into utility methods and call them. It would make it easier to follow the flow of logic. 
I would recommend 3 methods at least 

- copyPart
- encryptedPut
- unencryptedPut

lib/api/objectCopy.js
```diff
@@ -0,0 +1,304 @@
...
+                        destObjMD, serverSideEncryption);
+                });
+        },
+        function storeNewMetadata(storeMetadataParams, destDataGetInfoArr,
+            destObjMD, serverSideEncryption, next) {
+            return services.metadataStoreObject(destBucketName,
+                destDataGetInfoArr,
+                serverSideEncryption, storeMetadataParams, err => {
+                    if (err) {
+                        log.debug('error storing new metadata');
```
rahulreddy: Log the err object here

lib/api/objectCopy.js
```diff
@@ -0,0 +1,304 @@
...
+                                log, (err, cipherBundle) => {
+                                    if (err) {
+                                        log.debug('error getting cipherBundle');
+                                        return cb(errors.InternalError);
+                                    }
+                                    return data.put(cipherBundle, stream,
+                                        part.size, dataStoreContext, log,
+                                        (error, partRetrievalInfo) => {
+                                            if (error) {
+                                                return cb(error);
```
rahulreddy: need a log here

lib/api/objectCopy.js
```diff
@@ -0,0 +1,304 @@
...
+             // dataLocator is an array.  need to get and put all parts
+             // For now, copy 1 part at a time. Could increase the second
+             // argument here to increase the number of parts
+             // copied at once.
+            return async.mapLimit(dataLocator, 1,
+                // eslint-disable-next-line prefer-arrow-callback
+                function copyPart(part, cb) {
+                    return data.get(part, log, (err, stream) => {
+                        if (err) {
+                            return cb(err);
```
rahulreddy: need a log here

lib/api/objectCopy.js
```diff
@@ -0,0 +1,304 @@
...
+                                            return cb(null, partResult);
+                                        });
+                                });
+                        }
+                        // Copied object is not encrypted so just put it
+                        // without a cipherBundle
+                        return data.put(null, stream, part.size,
+                        dataStoreContext, log, (error, partRetrievalInfo) => {
+                            if (error) {
+                                return cb(error);
```
rahulreddy: need a log here

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -0,0 +1,834 @@
...
+                Bucket: destBucketName,
+                Key: destObjName,
+                CopySource: `${sourceBucketName}/${sourceObjName}`,
+            }, fields), cb);
+        }
+
+        it('should copy an object from a source bucket to a different ' +
+            'destination bucket and copy the metadata if no metadata directve' +
+            'header provided', done => {
+            s3.copyObject({ Bucket: destBucketName, Key: destObjName,
```
rahulreddy: Since most of the tests are similar, this could be consolidated into a util function and used throughout the test suite.

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -0,0 +1,834 @@
...
+                            assert.deepStrictEqual(res.StorageClass,
+                                'REDUCED_REDUNDANCY');
+                            assert.strictEqual(res.ETag, emptyFileETag);
+                            done();
+                        });
+                    });
+            });
+        });
+
+        it('should copy an object to a differnt destinagtion and change ' +
```
rahulreddy: should copy an object to a **different destination** and change

lib/api/api.js
LaurenSpiegel: An extra call for fail requests is not desired.  I don't think it would be much of a benefit for successful and there is already so much going on that I prefer to keep it separate.

lib/api/objectCopy.js
LaurenSpiegel: you can't separate the get from the put.  so i could do one copypart for encrypted versus nonencrypted but there will be duplication on the get side and there are a lot of parameters being pulled in that will have to be used in utility functions.

lib/api/objectCopy.js
LaurenSpiegel: I'm logging the error at the end of the async.map.

lib/api/objectCopy.js
LaurenSpiegel: same.

lib/api/objectCopy.js
LaurenSpiegel: same

tests/functional/aws-node-sdk/test/object/objectCopy.js
LaurenSpiegel: done

LaurenSpiegel:
@ironman-machine any issue with the updates?

rahulreddy:
👍

lib/api/objectCopy.js
rahulreddy: Ok.

LaurenSpiegel:
end to end passed

Issue #136, opened 2016-08-11
ACL public-read not working? 


Pathio:
Hi, I've been playing around with your S3 mock and it's really useful, so thanks for that!

I've come across an issue with the ACL. I've created an object and set the ACL to public-read. Double-checking the ACL for the gives me this:

{u'Owner': {u'DisplayName': 'Bart', u'ID': '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be'}, u'Grants': [{u'Grantee': {u'Type': 'CanonicalUser', u'DisplayName': 'Bart', u'ID': '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be'}, u'Permission': 'FULL_CONTROL'}, {u'Grantee': {u'Type': 'Group', u'URI': 'http://acs.amazonaws.com/groups/global/AllUsers'}, u'Permission': 'READ'}], 'ResponseMetadata': {'HTTPStatusCode': 200, 'HostId': 'bcfac69272b9181ff6e2', 'RequestId': 'bcfac69272b9181ff6e2'}}

So it looks as expected. However, when I do a GET on the object's URL I get an Access Denied (signing the URL and using that works ok so it behaves as if it were private). Any idea?

Thanks!

nicolas2bert:
Hey @Pathio , we are working on it.
I will come back to you when it is done.

Nicolas

LaurenSpiegel:
Fixed by https://github.com/scality/S3/pull/143#issuecomment-239588941

PR #142, opened 2016-08-12

nicolas2bert:
Fixing issue #136

When setting the 'public-read' ACL permission of the object, any account can get an object (objectGet) / get its metadata (objectHead) even if the account has no right to the bucket


LaurenSpiegel:
👍

alexandre-merle:
LGTM :+1: another time :D

LaurenSpiegel:
@ironman-machine any plans for the weekend?

PR #146, opened 2016-08-15
should not return sse key id unless sse is aws:kms

ploki:
by return, I mean in the response header


ploki:
returned key id is the actual bucket key when S3 is deployed with ```file``` backend.

alexandre-merle:
:+1:

LaurenSpiegel:
👍

LaurenSpiegel:
@ironman-machine just to be sure

PR #147, opened 2016-08-16
Object protected by erasure codes

lamphamsy:
Object (or MPU part) sent from S3 Server will be protected by use a (k, m) erasure codes.
Each object is encoded into (k + m) fragments of the same size. These
fragments will be sent to object client to store them. Each fragment has its
own key. The object's key is a string composed of (k + m) keys of its fragments.

There are put/get/delete operations that use erasure coding techniques.
Eclib library is used to perform coding operations.

PUT object operation is split into two parallel processes
1. split and store object as `k` data fragments (with valid header)
2. generate parity fragments via `eclib`, then store them


lib/Config.js
```diff
@@ -243,6 +243,9 @@ class Config {
         if (process.env.S3KMS) {
             kms = process.env.S3KMS;
         }
+        if (process.env.S3DP) {
+            data = process.env.S3DP;
+        }
```
LaurenSpiegel: I think it would be clearer if you had a separate variable called db or erasure coding and set that separately from data.  Then that variable would be set to on or off.

lib/data/dpa/dpClient.js
```diff
@@ -0,0 +1,280 @@
...
+     * @param {String} params.algo - algorithm for the hash
+     * @param {String} reqUids - The serialized request id
+     * @param {Callback} callback - callback
+     * @returns {undefined}
+     */
+    put(stream, size, params, reqUids, callback) {
+        assert(stream.readable, 'stream should be readable');
+        const log = this.createLogger(reqUids);
+
+        this.encode.bind(this)(stream, (err, fragments) => {
```
LaurenSpiegel: is bind needed here?

lib/data/dpa/dpClient.js
```diff
@@ -0,0 +1,280 @@
...
+     * @param {Object} params - parameters for key generation
+     * @param {String} params.bucketName - name of the object's bucket
+     * @param {String} params.owner - owner of the object
+     * @param {String} params.namespace - namespace of the S3 request
+     * @param {String} params.algo - algorithm for the hash
+     * @param {String} reqUids - The serialized request id
+     * @param {Callback} callback - callback
+     * @returns {undefined}
+     */
+    put(stream, size, params, reqUids, callback) {
```
LaurenSpiegel: you're not taking into account the cipherBundle argument that is now sent in

lib/data/wrapper.js
```diff
@@ -6,6 +6,7 @@ import inMemory from './in_memory/backend';
 import config from '../Config';
 import MD5Sum from '../utilities/MD5Sum';
 import assert from 'assert';
+import DpClient from './dpa/dpClient';
```
LaurenSpiegel: I think it would be clearer to call this middleware rather than a client because really what it is doing is calling each of the individual backend clients.

lib/data/dpa/dpClient.js
```diff
@@ -0,0 +1,280 @@
...
+     * This sends a PUT request to object client.
+     * @param {http.IncomingMessage} stream - Request with the data to send
+     * @param {string} stream.contentHash - hash of the data to send
+     * @param {number} size - data size in the stream
+     * @param {Object} params - parameters for key generation
+     * @param {String} params.bucketName - name of the object's bucket
+     * @param {String} params.owner - owner of the object
+     * @param {String} params.namespace - namespace of the S3 request
+     * @param {String} params.algo - algorithm for the hash
+     * @param {String} reqUids - The serialized request id
```
LaurenSpiegel: the log is actually being sent through.  why not use the logger being sent rather than recreating?

lib/data/dpa/dpClient.js
```diff
@@ -0,0 +1,280 @@
...
+        const log = this.createLogger(reqUids);
+
+        this.encode.bind(this)(stream, (err, fragments) => {
+            if (err) {
+                return callback(err);
+            }
+
+            const keys = new Array(fragments.length);
+            async.forEachOf(fragments, (frag, idx, cb) => {
+                const streamFrag = new PassThrough;
```
LaurenSpiegel: why a passthrough stream?  why not a transform stream that does the encoding?

lib/Config.js
lamphamsy: thanks. It's now

lib/data/dpa/dpClient.js
lamphamsy: in fact, these infos could be optional. So I removed such detailed infos

lib/data/dpa/dpClient.js
lamphamsy: as (k+m) encoded fragments are stored separately, hence there are (k+m) output streams. I'm not sure that we can do that with a transform stream..

lib/data/dpa/dpClient.js
lamphamsy: are you sure?  `log.getSerializedUids()` is actually sent to data client, not `log`

LaurenSpiegel: you're right.  i was looking at the wrapper put signature.

lib/data/dpa/dpClient.js
LaurenSpiegel: got it.

lib/data/dpa/dpClient.js
LaurenSpiegel: I see.  why do you need the passthrough?

lamphamsy: ah, I just want to send a buffer as a stream. Is `passthrough` not a good way for that?

LaurenSpiegel: sounds like you just want a readable stream -- https://nodejs.org/api/stream.html#stream_implementing_a_readable_stream

lamphamsy: Yes, it's updated. Thanks

lib/data/wrapper.js
lamphamsy: For now it calls individual backend clients. Our next purpose is to store fragments in multiple backend clients

lamphamsy:
This PR is replaced by #293

PR #161, opened 2016-08-23
FIX: Ensure bucket is empty at end of test.

LaurenSpiegel:
Without this change, the after all delete was being
attempted before the bucket was empty.

This closes https://github.com/scality/S3/issues/160


rahulreddy:
👍

PR #172, opened 2016-08-26
BF: Move commander module to be prod. dependency

### LINKED ISSUE #170: Encrypted bucket creation tool fails in production
Encrypted bucket creation tool uses the commander module which is currently a dev dependency. It should be moved to be a regular dependency so that npm install --production install the module.

rahulreddy:
Fix #170


LaurenSpiegel:
👍

RachedBenMustapha:
👍

PR #174, opened 2016-08-27
Ft/copy part

LaurenSpiegel:
Implements upload part copy feature.

Note: THIS CONTAINS A SMALL BREAKING CHANGE.
This modifies the format in which a put part stores data.
As a consequence, the way that an MPU is completed is altered.
This means that an MPU that was initiated prior to May 2nd,
will not be able to be completed.

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -225,8 +225,10 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                     if (!location || typeof location !== 'object') {
                         continue;
                     }
-                    const pieceSize = Number.isInteger(location.size) ?
-                        location.size : partSize;
+                    let pieceSize = partSize;
+                    if (location.size) {
+                        pieceSize = Number.parseInt(location.size, 10);
+                    }
                     const pieceRetrievalInfo = {
                         key: location.key,
                         size: pieceSize,
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -142,21 +142,9 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 key: item.key,
                 ETag: `"${item.value.ETag}"`,
                 size: item.value.Size,
-                // If the stored part has a partLocations.key property,
-                // use that for the actual key.  If not, it means we
-                // have an object using an md-model-version prior to 2 so
-                // item.vaule.partLocations will just contain the key
-                location: item.value.partLocations.key ?
-                    item.value.partLocations.key : item.value.partLocations,
-                // If the stored part has a partLocations.dataStoreName
-                // property use that for the actual dataStoreName.
-                // If not, default to sproxyd
-                dataStoreName: item.value.partLocations.dataStoreName ?
-                    item.value.partLocations.dataStoreName : 'sproxyd',
-                sseAlgorithm: item.value.partLocations.sseAlgorithm,
-                sseMasterKeyId: item.value.partLocations.sseMasterKeyId,
-                sseCryptoScheme: item.value.partLocations.sseCryptoScheme,
-                sseCipheredDataKey: item.value.partLocations.sseCipheredDataKey,
+
+                locations: Array.isArray(item.value.partLocations) ?
+                    item.value.partLocations : [item.value.partLocations],
             }));
             let splitter = constants.splitter;
             // BACKWARD: Remove to remove the old splitter
@@ -221,17 +209,36 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 }
                 // Assemble array of part locations, aggregate size and build
                 // string to create aggregate ETag
-                const partRetrievalInfo = {
-                    key: storedPartsAsObjects[i].location,
-                    size: partSize,
-                    start: calculatedSize,
-                    dataStoreName: storedPartsAsObjects[i].dataStoreName,
-                    cryptoScheme: storedPartsAsObjects[i].sseCryptoScheme,
-                    cipheredDataKey: storedPartsAsObjects[i].sseCipheredDataKey,
-                };
 
-                dataLocations.push(partRetrievalInfo);
-                calculatedSize += partSize;
+                // If part was put by a regular put part rather than a copy
+                // it is always one location.  With a put part copy, could
+                // be multiple locations so loop over array of locations.
+                for (let j = 0; j < storedPartsAsObjects[i].locations.length;
+                    j ++) {
+                    // If the piece has parts (was a put part object
+                    // copy) each piece will have a size attribute.
+                    // Otherwise, the piece was put by a regular put
+                    // part and the size the of the piece is the full
+                    // part size.
+                    const location = storedPartsAsObjects[i].locations[j];
+                    // If there is no location, move on
+                    if (!location || typeof location !== 'object') {
+                        continue;
+                    }
+                    const pieceSize = Number.isInteger(location.size) ?
+                        location.size : partSize;
+                    const pieceRetrievalInfo = {
+                        key: location.key,
+                        size: pieceSize,
+                        start: calculatedSize,
+                        dataStoreName: location.dataStoreName,
+                        cryptoScheme: location.sseCryptoScheme,
+                        cipheredDataKey: location.sseCipheredDataKey,
+                    };
+                    dataLocations.push(pieceRetrievalInfo);
+                    calculatedSize += pieceSize;
+                }
+
                 const partETagWithoutQuotes =
                     storedPartsAsObjects[i].ETag.slice(1, -1);
                 concatETags += partETagWithoutQuotes;
--- a/lib/api/multipartDelete.js
+++ b/lib/api/multipartDelete.js
@@ -72,11 +72,12 @@ function multipartDelete(authInfo, request, log, callback) {
         function deleteData(mpuBucket, storedParts, next) {
             // The locations were sent to metadata as an array
             // under partLocations.  Pull the partLocations.
-            const locations = storedParts.map(item => item.value.partLocations);
+            let locations = storedParts.map(item => item.value.partLocations);
             if (locations.length === 0) {
                 return next(null, mpuBucket, storedParts);
             }
-            // If have locations, flatten the array
+            // flatten the array
+            locations = [].concat.apply([], locations);
             return async.eachLimit(locations, 5, (loc, cb) => {
                 data.delete(loc, log, err => {
                     if (err) {
--- a/lib/api/objectPutCopyPart.js
+++ b/lib/api/objectPutCopyPart.js
@@ -1,0 +1,370 @@
+import async from 'async';
+import { errors } from 'arsenal';
+
+import constants from '../../constants';
+import data from '../data/wrapper';
+import kms from '../kms/wrapper';
+import metadata from '../metadata/wrapper';
+import RelayMD5Sum from '../utilities/RelayMD5Sum';
+import { logger } from '../utilities/logger';
+import routesUtils from '../routes/routesUtils';
+import services from '../services';
+import { parseRange } from './apiUtils/object/parseRange';
+import validateHeaders from '../utilities/validateHeaders';
+
+
+/**
+ * PUT Part Copy during a multipart upload.
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with
+ * requester's info
+ * @param {request} request - request object given by router,
+ *                            includes normalized headers
+ * @param {string} sourceBucket - name of source bucket for object copy
+ * @param {string} sourceObject - name of source object for object copy
+ * @param {object} log - the request logger
+ * @param {function} callback - final callback to call with the result
+ * @return {undefined}
+ */
+export default
+function objectPutCopyPart(authInfo, request, sourceBucket,
+    sourceObject, log, callback) {
+    log.debug('processing request', { method: 'objectPutCopyPart' });
+    const destBucketName = request.bucketName;
+    const destObjectKey = request.objectKey;
+    const valGetParams = {
+        authInfo,
+        bucketName: sourceBucket,
+        objectKey: sourceObject,
+        requestType: 'objectGet',
+        log,
+    };
+
+
+    const partNumber = Number.parseInt(request.query.partNumber, 10);
+    // AWS caps partNumbers at 10,000
+    if (partNumber > 10000 || !Number.isInteger(partNumber) || partNumber < 1) {
+        return callback(errors.InvalidArgument);
+    }
+    // We pad the partNumbers so that the parts will be sorted
+    // in numerical order
+    const paddedPartNumber = `000000${partNumber}`.substr(-5);
+    // Note that keys in the query object retain their case, so
+    // request.query.uploadId must be called with that exact
+    // capitalization
+    const uploadId = request.query.uploadId;
+
+    const valPutParams = {
+        authInfo,
+        bucketName: destBucketName,
+        objectKey: destObjectKey,
+        requestType: 'objectPut',
+        log,
+    };
+
+    // For validating the request at the MPU, the params are the same
+    // as validating for the destination bucket except additionally need
+    // the uploadId and splitter.
+    // Also, requestType is 'putPart or complete'
+    const valMPUParams = Object.assign({
+        uploadId,
+        splitter: constants.splitter,
+    }, valPutParams);
+    valMPUParams.requestType = 'putPart or complete';
+
+    const dataStoreContext = {
+        bucketName: destBucketName,
+        owner: authInfo.getCanonicalID(),
+        namespace: request.namespace,
+    };
+
+    return async.waterfall([
+        function checkSourceAuthorization(next) {
+            return services.metadataValidateAuthorization(valGetParams,
+                (err, sourceBucketMD, sourceObjMD) => {
+                    if (err) {
+                        log.debug('error validating get part of request',
+                        { error: err });
+                        return next(err);
+                    }
+                    if (!sourceObjMD) {
+                        log.debug('no source object', { sourceObject });
+                        return next(errors.NoSuchKey);
+                    }
+                    const headerValResult =
+                        validateHeaders(sourceObjMD, request.headers);
+                    if (headerValResult.error) {
+                        return next(errors.PreconditionFailed);
+                    }
+                    let dataLocator;
+                    // If 0 byte object just set dataLocator to empty array
+                    if (!sourceObjMD.location) {
+                        dataLocator = [];
+                    } else {
+                        // To provide for backwards compatibility before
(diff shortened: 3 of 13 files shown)


lib/api/completeMultipartUpload.js
```diff
@@ -221,17 +213,38 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                    cryptoScheme: storedPartsAsObjects[i].sseCryptoScheme,
-                    cipheredDataKey: storedPartsAsObjects[i].sseCipheredDataKey,
-                };
 
-                dataLocations.push(partRetrievalInfo);
-                calculatedSize += partSize;
+                // If part was put by a regular put part rather than a copy
+                // it is always one location.  With a put part copy, could
+                // be multiple locations so loop over array of locations.
+                for (let j = 0; j < storedPartsAsObjects[i].locations.length;
```
alexandre-merle: We can't manage backward compatbility by checking ```storedPartsAsObjects[i].location``` existence ? :)

lib/api/multipartDelete.js
```diff
@@ -72,11 +72,12 @@ function multipartDelete(authInfo, request, log, callback) {
...
             // The locations were sent to metadata as an array
             // under partLocations.  Pull the partLocations.
-            const locations = storedParts.map(item => item.value.partLocations);
+            let locations = storedParts.map(item => item.value.partLocations);
             if (locations.length === 0) {
                 return next(null, mpuBucket, storedParts);
             }
-            // If have locations, flatten the array
+            // flatten the array
+            locations = [].concat.apply([], locations);
```
alexandre-merle: ```[].concat(locations)``` should be enought no ? :)

lib/api/completeMultipartUpload.js
LaurenSpiegel: I spoke with @vrancurel about this.  It's unlikely someone has an open MPU from that far back so it would be a good time to make this change rather than complicating the code.

lib/api/multipartDelete.js
LaurenSpiegel: no.  see the section about apply and built in functions -- http://devdocs.io/javascript/global_objects/function/apply

lib/api/completeMultipartUpload.js
```diff
@@ -142,21 +142,13 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                // property use that for the actual dataStoreName.
-                // If not, default to sproxyd
-                dataStoreName: item.value.partLocations.dataStoreName ?
-                    item.value.partLocations.dataStoreName : 'sproxyd',
-                sseAlgorithm: item.value.partLocations.sseAlgorithm,
-                sseMasterKeyId: item.value.partLocations.sseMasterKeyId,
-                sseCryptoScheme: item.value.partLocations.sseCryptoScheme,
-                sseCipheredDataKey: item.value.partLocations.sseCipheredDataKey,
+
+                // BREAKING CHANGE -- NO LONGER SUPPORT PART DATA STRUCTURE
```
michael-zapata: Since you assessed that no-one has that issue, can we get rid of the comment here? The commit message can give enough context by itself.

lib/api/completeMultipartUpload.js
michael-zapata: Why not use a `forEach()` ? A `return` instruction acts as a `continue` in the `for` loop.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+
+/**
+ * PUT Part Copy during a multipart upload.
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with
+ * requester's info
+ * @param {request} request - request object given by router,
+ *                            includes normalized headers
+ * @param {string} sourceBucket - name of source bucket for object copy
+ * @param {string} sourceObject - name of source object for object copy
+ * @param {object} log - the log request
```
rahulreddy: the request logger

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+
+    // For validating the request at the MPU, the params are the same
+    // as validating for the destination bucket except additionally need
+    // the uploadId and splitter.
+    // Also, requestType is 'putPart or complete'
+    const valMPUParams = Object.assign({
+        uploadId,
+        splitter: constants.splitter,
+    }, valPutParams);
+    valMPUParams.requestType = 'putPart or complete';
```
rahulreddy: I know that this is being used by complete MPU as well, we could address this in a different PR but we should make this all upper-cased or all lower-cased, otherwise it's easy to mess up the case.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                        return next(err);
+                    }
+                    if (!sourceObjMD) {
+                        log.debug('no source object', { sourceObject });
+                        return next(errors.NoSuchKey);
+                    }
+                    const headerValResult =
+                        validateHeaders(sourceObjMD, request.headers);
+                    if (headerValResult.error) {
+                        return next(errors.PreconditionFailed);
```
rahulreddy: Let's log the validation error here along with the headers.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                        const range = parseRange(request
+                            .headers['x-amz-copy-source-range'],
+                            sourceSize);
+                        // If have a data model before version 2, cannot
+                        // support get range copy (do not have size
+                        // stored with data locations)
+                        if ((range && dataLocator.length >= 1) &&
+                            (dataLocator[0].start === undefined
+                            || dataLocator[0].size === undefined)) {
+                            return callback(errors.NotImplemented);
```
rahulreddy: We should introduce an Arsenal error for this case, something like `errors.LegacyObject`. We could lump all incompatible actions due to breaking changes into this error with a custom error description. Also we would need a log here for debugging.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                            });
+                        stream.pipe(hashedStream);
+                        const numberPartSize =
+                            Number.parseInt(part.size, 10);
+                        if (serverSideEncryption) {
+                            return kms.createCipherBundle(
+                                serverSideEncryption,
+                                log, (err, cipherBundle) => {
+                                    if (err) {
+                                        log.debug('error getting cipherBundle');
```
rahulreddy: log the error object as well.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                                        return cb(errors.InternalError);
+                                    }
+                                    return data.put(cipherBundle, hashedStream,
+                                        numberPartSize, dataStoreContext, log,
+                                        (error, partRetrievalInfo) => {
+                                            if (error) {
+                                                return cb(error);
```
rahulreddy: log the error with error object and method name.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                                                    .cryptoScheme,
+                                                sseCipheredDataKey: cipherBundle
+                                                    .cipheredDataKey,
+                                                sseAlgorithm: cipherBundle
+                                                    .algorithm,
+                                                sseMasterKeyId: cipherBundle
+                                                    .masterKeyId,
+                                            };
+                                            locations.push(partResult);
+                                            return cb(null);
```
rahulreddy: return cb()

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                                        });
+                                });
+                        }
+                        // Copied object is not encrypted so just put it
+                        // without a cipherBundle
+                        return data.put(null, hashedStream, numberPartSize,
+                        dataStoreContext, log, (error, partRetrievalInfo) => {
+                            if (error) {
+                                return cb(error);
```
rahulreddy: same, log error and method here

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                            }
+                            const partResult = {
+                                key: partRetrievalInfo.key,
+                                dataStoreName: partRetrievalInfo.dataStoreName,
+                                size: part.size,
+                            };
+                            locations.push(partResult);
+                            return cb(null);
```
rahulreddy: return cb()

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                        });
+                    });
+                }, err => {
+                    if (err) {
+                        log.debug('error transferring data from source');
```
rahulreddy: log the error along with the method

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                contentMD5: totalHash,
+                size: copyObjectSize,
+                uploadId,
+                splitter: constants.splitter,
+                lastModified,
+            };
+            return services.metadataStorePart(mpuBucket.getName(),
+                locations, metaStoreParams, log, err => {
+                    if (err) {
+                        log.debug('error storing new metadata');
```
rahulreddy: log method and error object here

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,356 @@
...
+                            logger.newRequestLoggerFromSerializedUids(
+                                log.getSerializedUids()));
+                    }
+                    return next(null, totalHash, lastModified,
+                        serverSideEncryption);
+                });
+        },
+    ], (err, totalHash, lastModified, serverSideEncryption) => {
+        if (err) {
+            return callback(err);
```
rahulreddy: trace log here as this would give us an overall context of the error.

lib/services.js
```diff
@@ -585,48 +585,39 @@ export default {
...
+        const { partNumber, contentMD5, size, uploadId, lastModified, splitter }
+            = metaStoreParams;
+        const dateModified = typeof lastModified === 'string' ?
+            lastModified : new Date().toJSON();
         assert.strictEqual(typeof splitter, 'string');
-        // TODO: Determine splitter that will not appear in
-        // any of these items.  This is GH Issue#218
-        // 1) UploadId's are UUID version 4
-        // 2) Part Number will be a stringified number between 1 and 10000
         const partKey = `${uploadId}${splitter}${partNumber}`;
```
rahulreddy: This needs unification in the future (like a shared function) to make sure we generate the same format of key.

lib/utilities/RelayMD5Sum.js
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * This function will call the done callback with the updated hash
+     *
+     * @param {function} callback(err)
+     * @return {undefined}
+     */
+    _flush(callback) {
+        this.done(this.hash);
+        callback(null);
```
rahulreddy: callback()

lib/utilities/RelayMD5Sum.js
```diff
@@ -0,0 +1,50 @@
+const Transform = require('stream').Transform;
```
rahulreddy: Use ES6 style imports here

lib/utilities/RelayMD5Sum.js
```diff
@@ -0,0 +1,50 @@
...
+     * @return {undefined}
+     */
+    _flush(callback) {
+        this.done(this.hash);
+        callback(null);
+    }
+
+}
+
+module.exports = RelayMD5Sum;
```
rahulreddy: Use ES6 export

lib/api/completeMultipartUpload.js
LaurenSpiegel: I get the no-loop-func linting error.  I don't think the forEach adds much to the readability anyway and wouldn't want to increment size in a map.

lib/api/completeMultipartUpload.js
LaurenSpiegel: Yes.  Move forward!

lib/services.js
LaurenSpiegel: good idea for another day

lib/api/objectPutCopyPart.js
LaurenSpiegel: I opened an issue to do this generally in one place.

LaurenSpiegel:
@ironman-machine up for this?

rahulreddy:
👍

LaurenSpiegel:
@ironman-machine you know you want to

LaurenSpiegel:
@ironman-machine up for this?

LaurenSpiegel:
end to end passed

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,370 @@
...
+                                    }
+                                    return data.put(cipherBundle, hashedStream,
+                                        numberPartSize, dataStoreContext, log,
+                                        (error, partRetrievalInfo) => {
+                                            if (error) {
+                                                log.debug('error putting ' +
+                                                'encrypted part', { error });
+                                                return cb(error);
+                                            }
+                                            const partResult = {
```
michael-zapata: Aren't we right-drifting too much there?

lib/api/objectPutCopyPart.js
michael-zapata: An even better solution would be using numbers, as you would use `enum`s in other languages, since integer comparison is always faster than a string comparison.

lib/api/objectPutPart.js
```diff
@@ -63,7 +74,7 @@ export default function objectPutPart(authInfo, request, log, cb) {
     if (partNumber > 10000) {
         return cb(errors.TooManyParts);
     }
-    if (!Number.isInteger(partNumber)) {
+    if (!Number.isInteger(partNumber) || partNumber < 1) {
```
alexandre-merle: Its a silent fix here, and this should go against rel/1.1 actually

lib/utilities/RelayMD5Sum.js
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * @constructor
+     * @param {Hash} starterHash - hash from prior stream (if any) to be updated
+     * with new stream here
+     * @param {function} done Callback - This callback is called when
+     * the hash computation is finished, this function need to be synchronous
+     * and being call before the end of the stream object
+     */
+    constructor(starterHash, done) {
```
alexandre-merle: starterHash is the only difference with the class MD5Sum, can we just adapt this one ?

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,370 @@
...
+    valMPUParams.requestType = 'putPart or complete';
+
+    const dataStoreContext = {
+        bucketName: destBucketName,
+        owner: authInfo.getCanonicalID(),
+        namespace: request.namespace,
+    };
+
+    return async.waterfall([
+        function checkSourceAuthorization(next) {
```
LaurenSpiegel: I moved the range parsing and location setting out to a helper function since all synchronous.  Otherwise, I find it easier to follow the flow when I can see the functions together.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,370 @@
...
+        function checkDestAuth(dataLocator, copyObjectSize, next) {
+            return services.metadataValidateAuthorization(valPutParams,
+                (err, destBucketMD) => {
+                    if (err) {
+                        log.debug('error validating authorization for ' +
+                        'destination bucket',
+                        { error: err });
+                        return next(err);
+                    }
+                    const flag = destBucketMD.hasDeletedFlag()
```
LaurenSpiegel: If the source bucket has an object to copy, it does not have a deleted flag.

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,370 @@
...
+            let totalHash;
+            const locations = [];
+             // dataLocator is an array.  need to get and put all parts
+             // in order so can get the ETag of full object
+            return async.forEachOfSeries(dataLocator,
+                // eslint-disable-next-line prefer-arrow-callback
+                function copyPart(part, index, cb) {
+                    return data.get(part, log, (err, stream) => {
+                        if (err) {
+                            return cb(err);
```
LaurenSpiegel: sure

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,370 @@
...
+                                size: part.size,
+                            };
+                            locations.push(partResult);
+                            return cb();
+                        });
+                    });
+                }, err => {
+                    if (err) {
+                        log.debug('error transferring data from source',
+                        { error: err, method: goGetData });
```
LaurenSpiegel: good catch

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,370 @@
...
+        function getExistingPartInfo(locations, totalHash,
+            copyObjectSize, mpuBucket, serverSideEncryption, next) {
+            const partKey =
+                `${uploadId}${constants.splitter}${paddedPartNumber}`;
+            metadata.getObjectMD(mpuBucket.getName(), partKey, log,
+                (err, result) => {
+                    // If there is nothing being overwritten just move on
+                    if (err && !err.NoSuchKey) {
+                        log.debug('error getting current part (if any)',
+                        { errors: err });
```
LaurenSpiegel: thanks

lib/api/objectPutCopyPart.js
```diff
@@ -0,0 +1,370 @@
...
+                size: copyObjectSize,
+                uploadId,
+                splitter: constants.splitter,
+                lastModified,
+            };
+            return services.metadataStorePart(mpuBucket.getName(),
+                locations, metaStoreParams, log, err => {
+                    if (err) {
+                        log.debug('error storing new metadata',
+                        { error: err, method: storeNewPartMetadata });
```
LaurenSpiegel: yes

lib/api/objectPutPart.js
LaurenSpiegel: I was viewing it as a feature actually (it's in its own commit)  :)
I'll drop it here and do a separate pr against 6.2.

lib/utilities/RelayMD5Sum.js
LaurenSpiegel: We also don't digest in the relay class so the flush function is different.  Also, the md5sum class is now actually no longer taking a callback but instead emitting an event when the hash is complete since I was getting a race condition for small objects.  To make sure we do not have the same problem here, I am getting rid of the relay flush function altogether and calling the done callback with my new hash before asking for a new chunk.

LaurenSpiegel:
@ironman-machine check it

LaurenSpiegel:
end to end passed.

PR #182, opened 2016-08-31
Add Configuring section

nicolas2bert:
We had a lot of issues opened because of hostname configuration.
I think it would be wise to add a section inside our READ.ME

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,37 @@ npm run mem_backend &
 npm run ft_test
 ```
 
+## Configuration
+
+If you want to specify an endpoint (other than localhost),
+you need to add it to your config.json:
+
+```json
+"regions": {
+
+     "localregion": ["localhost"],
+     "specifiedregion": ["myhostname.com"]
+},
+```
+
+Note that our S3server supports both:
+
+- path-style: http://myhostname.com/mybucket
+- hosted-style: http://mybucket.myhostname.com
+
+However, hosted-style requests will not hit the server if you are
+using an ip address for your host.
+So, make sure you are using path-style requests in that case.
+For instance, if you are using the AWS SDK for JavaScript,
+you would instantiate your client like this:
+
+```js
+const s3 = new aws.S3({
+   endpoint: 'http://127.0.0.1:8000',
+   s3ForcePathStyle: true,
+});
+```
+
 ## s3cmd versions
 
 If using s3cmd as a client to S3 be aware that v4 signature format


DavidPineauScality:
maybe it would be interesting to specify that the S3 server may not accept requests as expected without this configuration bit ?

README.md
```diff
@@ -90,6 +90,24 @@ npm run mem_backend &
 npm run ft_test
 ```
 
+## Configuring
```
LaurenSpiegel: Configuration

README.md
```diff
@@ -90,6 +90,24 @@ npm run mem_backend &
...
+
+If you want to specify a specific endpoint (different from localhost),
```
LaurenSpiegel: If you want to specify an endpoint (other than localhost),

README.md
```diff
@@ -90,6 +90,24 @@ npm run mem_backend &
...
+```shell
+"regions": {
+
+     ...
+     "localregion": ["localhost"],
+     "specifiedregion": ["myhostname.com"]
+},
+```
+
+Our S3server support both:
```
LaurenSpiegel: Note that our S3server supports both:

README.md
```diff
@@ -90,6 +90,24 @@ npm run mem_backend &
...
+path-style: http://myhostname.com/mybucket
+hosted-style: http://mybucket.myhostname.com
+
```
LaurenSpiegel: However, note that hosted-style requests will not hit the server if you are using an ip address for your host.  So, make sure you are using path-style requests in that case.  For instance, if you are using the AWS SDK for JavaScript you would instantiate your client like this:

``javascript
const s3 = new aws.S3({
    endpoint: 'http://localhost:8000',
   s3ForcePathStyle: true,
 });
``

LaurenSpiegel:
👍

README.md
```diff
@@ -90,6 +90,37 @@ npm run mem_backend &
...
+path-style: http://myhostname.com/mybucket and
+hosted-style: http://mybucket.myhostname.com
+
+However, hosted-style requests will not hit the server if you are
+using an ip address for your host.
+So, make sure you are using path-style requests in that case.
+For instance, if you are using the AWS SDK for JavaScript,
+you would instantiate your client like this:
+
+```shell
```
michael-zapata: This would be `es6` rather than `shell`.

README.md
```diff
@@ -90,6 +90,37 @@ npm run mem_backend &
 npm run ft_test
 ```
 
+## Configuration
+
+If you want to specify an endpoint (other than localhost),
+you need to add it to your config.json:
+
+```shell
```
michael-zapata: Shouldn't it be `json`? (You'd need some editions to make your example valid JSON though).

README.md
```diff
@@ -90,6 +90,37 @@ npm run mem_backend &
...
+
+     ...
+     "localregion": ["localhost"],
+     "specifiedregion": ["myhostname.com"]
+},
+```
+
+Note that our S3server supports both:
+path-style: http://myhostname.com/mybucket and
+hosted-style: http://mybucket.myhostname.com
```
michael-zapata: host-style*

nicolas2bert: I think both are ok, no?

PR #185, opened 2016-08-31
Fix/acl from query

LaurenSpiegel:
Fixes https://github.com/scality/S3/issues/184

### DIFF
--- a/lib/api/objectPut.js
+++ b/lib/api/objectPut.js
@@ -33,13 +33,20 @@ function _storeIt(bucketName, objectKey, objMD, authInfo, canonicalID,
         owner: canonicalID,
         namespace: request.namespace,
     };
+    // If the request was made with a pre-signed url, the x-amz-acl 'header'
+    // might be in the query string rather than the actual headers so include
+    // it here
+    const headers = request.headers;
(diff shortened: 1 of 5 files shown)


LaurenSpiegel:
@ironman-machine check it

lib/api/objectPut.js
```diff
@@ -33,13 +33,17 @@ function _storeIt(bucketName, objectKey, objMD, authInfo, canonicalID,
         owner: canonicalID,
         namespace: request.namespace,
     };
+    // If the request was made with a pre-signed url, the x-amz-acl 'header'
+    // might be in the query string rather than the actual headers so include
+    // the query params here to be parsed for any acl setting.
+    const headersPlusQuery = Object.assign(request.headers, request.query);
```
rahulreddy:
```js
const headers = Object.assign(request.headers, request.query);
```
This way you can use the shorthand in metadataStoreParams
Also, I just realized this. I think we should filter the fields that we want instead of taking everything from query (like just x-amz, content-type etc)

LaurenSpiegel:
@ironman-machine once more

nicolas2bert:
👍

rahulreddy:
👍

PR #187, opened 2016-09-01
BF: Pass prev. size to utapi on object overwrite

rahulreddy:
Utapi needs both previous and new content-length of the object that is being overwritten to calculate storage utilized and bandwidth for the bucket


LaurenSpiegel:
👍

nicolas2bert:
👍

rahulreddy:
@ironman-machine test please

rahulreddy:
End to end passed!

PR #189, opened 2016-09-02
Ft/config utapi

LaurenSpiegel:
This pr enables authentication and authorization for utapi.
It also creates a cli tool to list bucket metrics for utapi (using auth).

### DIFF
--- a/bin/list_bucket_metrics.js
+++ b/bin/list_bucket_metrics.js
@@ -1,0 +1,5 @@
+#!/usr/bin/env node
+'use strict'; // eslint-disable-line strict
+
+require('babel-core/register');
+require('../lib/utapi/utilities.js').listBucketMetrics();
--- a/lib/utapi/utapi.js
+++ b/lib/utapi/utapi.js
@@ -1,6 +1,6 @@
 
 const utapiServer = require('utapi').UtapiServer;
-const _config = require('./Config').default;
+const _config = require('../Config').default;
 
 // start utapi server
 export default function main() {
--- a/lib/utapi/utilities.js
+++ b/lib/utapi/utilities.js
@@ -1,0 +1,117 @@
+import { auth } from 'arsenal';
+import commander from 'commander';
+import http from 'http';
+import https from 'https';
+import { logger } from '../utilities/logger';
+
+function _listBucketMetrics(host,
+                            port,
+                            buckets,
+                            timeRange,
+                            accessKey,
+                            secretKey,
+                            verbose,
+                            ssl) {
+    const options = {
+        host,
+        port,
+        method: 'POST',
+        path: '/buckets?Action=ListMetrics',
+        headers: {
+            'content-type': 'application/json',
+            'cache-control': 'no-cache',
+        },
+        rejectUnauthorized: false,
+    };
+    const transport = ssl ? https : http;
+    const request = transport.request(options, response => {
+        if (verbose) {
+            logger.info('response status code', {
+                statusCode: response.statusCode,
+            });
+            logger.info('response headers', { headers: response.headers });
+        }
+        const body = [];
+        response.setEncoding('utf8');
+        response.on('data', chunk => body.push(chunk));
+        response.on('end', () => {
+            const responseBody = JSON.parse(body.join(''));
+            if (response.statusCode >= 200 && response.statusCode < 300) {
+                // eslint-disable-next-line no-console
+                console.log(responseBody);
+                process.exit(0);
+            } else {
+                logger.error('request failed with HTTP Status ', {
+                    statusCode: response.statusCode,
+                    body: responseBody,
+                });
+                process.exit(1);
+            }
+        });
+    });
+    auth.generateV4Headers(request, '', accessKey, secretKey, 's3');
+    if (verbose) {
+        logger.info('request headers', { headers: request._headers });
+    }
+    request.write(JSON.stringify({ buckets, timeRange }));
+    request.end();
+}
+
+/**
+ * This function is used as a binary to send a request to utapi server
+ * to list bucket metrics
+ *
+ * @return {undefined}
+ */
+export function listBucketMetrics() {
+    commander
+        .version('0.0.1')
+        .option('-a, --access-key <accessKey>', 'Access key id')
+        .option('-k, --secret-key <secretKey>', 'Secret access key')
+        .option('-b, --buckets <buckets>', 'Name of bucket(s)' +
+        'with a comma separator if more than one')
+        .option('-s, --start <start>', 'Start of time range')
+        .option('-e --end <end>', 'End of time range')
+        .option('-h, --host <host>', 'Host of the server')
+        .option('-p, --port <port>', 'Port of the server')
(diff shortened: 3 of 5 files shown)


lib/utapi/utapi.js
```diff
@@ -0,0 +1,17 @@
+
+const utapiServer = require('utapi').UtapiServer;
+const _config = require('../Config').default;
+
+// start utapi server
+export default function main() {
+    if (_config.utapi) {
+        let fullConfig = Object.assign({}, _config.utapi);
+        if (_config.vaultd) {
+            fullConfig = Object.assign(fullConfig, { vaultd: _config.vaultd });
```
rahulreddy:
```js
Object.assign(fullConfig, { vaultd: _config.vaultd });
```
should be sufficient

lib/utapi/utapi.js
```diff
@@ -0,0 +1,17 @@
...
+        }
+        if (_config.https) {
+            fullConfig = Object.assign(fullConfig, { https: _config.https });
```
rahulreddy: same here
```js
Object.assign(fullConfig, { https: _config.https });
``` 
should be sufficient

lib/utapi/utilities.js
```diff
@@ -0,0 +1,114 @@
...
+        .option('-v, --verbose')
+        .parse(process.argv);
+
+    const { host, port, accessKey, secretKey, start, end,
+        buckets, verbose, ssl } =
+        commander;
+
+    if (!host || !port || !accessKey || !secretKey || !buckets
+        || !start) {
+        logger.error('missing parameter');
```
rahulreddy: Can we log all the params (except secretKey) to indicate which one is missing.
For secretKey, you could use a different if clause with error message `missing secret key` and not log the secret key.

LaurenSpiegel:
@ironman-machine if you don't mind
SCALITY_VAULT_BRANCH=ft/utapiAuth

LaurenSpiegel:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/2976

nicolas2bert:
👍

rahulreddy:
👍

LaurenSpiegel:
SCALITY_INTEGRATION_BRANCH=dev/V4Headers
SCALITY_VAULT_BRANCH=ft/utapiAuth

LaurenSpiegel:
@ironman-machine try this with arsenal master in the integration branch
SCALITY_INTEGRATION_BRANCH=dev/V4Headers
SCALITY_VAULT_BRANCH=ft/utapiAuth

LaurenSpiegel:
@ironman-machine check this out
SCALITY_INTEGRATION_BRANCH=dev/V4Headers
SCALITY_VAULT_BRANCH=ft/utapiAuth
SCALITY_METADATA_BRANCH=rel/1.1

bin/list_bucket_metrics.js
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
DavidPineauScality: wut ? What's a shebang doing in a js file ? (same for the following shell command, btw)
If it's documentation to tell how to run the file, I'd rather see it as a proper comment explaining things clearly.
Also, `#!/usr/bin/env node` should be fine...

lib/utapi/utilities.js
```diff
@@ -0,0 +1,117 @@
...
+        .option('-k, --secret-key <secretKey>', 'Secret access key')
+        .option('-b, --buckets <buckets>', 'Name of bucket(s)' +
+        'with a comma separator if more than one')
+        .option('-s, --start <start>', 'Start of time range')
+        .option('-e --end <end>', 'End of time range')
+        .option('-h, --host <host>', 'Host of the server')
+        .option('-p, --port <port>', 'Port of the server')
+        .option('--ssl', 'Enable ssl')
+        .option('-v, --verbose')
+        .parse(process.argv);
```
DavidPineauScality: This is making things confusing for someone that starts by reading the bin script (using argv somewhere inside the code that isn't the entrypoint file)

lib/utapi/utilities.js
```diff
@@ -0,0 +1,117 @@
...
+
+    const { host, port, accessKey, secretKey, start, end,
+        buckets, verbose, ssl } =
+        commander;
+    const requiredOptions = { host, port, accessKey, secretKey, buckets,
+        start };
+    Object.keys(requiredOptions).forEach(option => {
+        if (!requiredOptions[option]) {
+            logger.error(`missing required option: ${option}`);
```
DavidPineauScality: If the aim is to use the script as a CLI tool, I'm wondering whether using werelogs is the best strategy (although not invalid). any opinion ?

lib/utapi/utilities.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+ * This function is used as a binary to send a request to utapi server
+ * to list bucket metrics
+ *
+ * @return {undefined}
+ */
+export function listBucketMetrics() {
+    commander
```
DavidPineauScality: Why not leave this to the bin/ script ?

LaurenSpiegel: leave what to the script?  I'm using a js script using commander.

lib/utapi/utilities.js
LaurenSpiegel: I was using the same format we have for the bucket encryption tool.

lib/utapi/utilities.js
DavidPineauScality: I just meant to have all the commander-related stuff in the bin/ script rather than in an utilities.js
I personally do not think that this kind of argument management code has its place in a set of utilities, but I might just be oldschool.

lib/utapi/utilities.js
LaurenSpiegel: I don't feel strongly one way or the other.  We could remove it later if we want this more independent (same with the bucket encryption tool).

lib/utapi/utilities.js
DavidPineauScality: Well, I didn't review that one, so.... I couldn't make the same comment :)

LaurenSpiegel: I'd prefer that they be in the same format so that once someone has reviewed one, they can tell how the other works.

lib/utapi/utilities.js
DavidPineauScality: I think it's mostly about what it's used for, and how we want to show the output. For instance, if it's something automated, it's fine to have parseable output (json would be a good choice then). If not, maybe we'd lean towards a more classical approach. Not an issue anyways, just wanted to know if there was a choice there or not.

LaurenSpiegel:
@ironman-machine friday fun
SCALITY_INTEGRATION_BRANCH=dev/V4Headers
SCALITY_VAULT_BRANCH=ft/utapiAuth

LaurenSpiegel:
@ironman-machine me again
SCALITY_INTEGRATION_BRANCH=dev/V4Headers
SCALITY_VAULT_BRANCH=ft/utapiAuth

PR #190, opened 2016-09-02
UKS key for DPA [to review only]

lamphamsy:
The PR is for review only.  It's waiting for #147 

Added a module to generate UKS keys using for DPA.
There are 4 sub-modules:
1. keygen.obj generates a key for an object
2. keygen.frag generates a key for a given fragment and an object's key
3. keygen.all generates all keys of `k + m` fragments for a given object's key
4. keygen.ec extracts erasure codes' parameters (backend id, k, m) from a key

UKS key schema facilitate operations:
- PUT: a key is generate for each object. That key is sent back to S3 once the
  object is successfully uploaded (and protected by erasure codes)
  Each fragment of the object is identified by an other key that is generated
  based on the object's key and its index in erasure codes. Fragment keys are not
  stored as they can be regenerated when the object's key is known.
- GET: given an object's key, keys of its fragments are generated in order to
  retrieve them from data storage backend.
- DELETE: given an object's key, keys of its fragments are generated in order to
  delete them from data storage backend.


lamphamsy:
Closed due to #293

PR #192, opened 2016-09-02
FIX: call pushMetricCreateBucket when createbucket using POST

nicolas2bert:
When creating bucket using a POST request: we call `pushMetricCreateBucket` from Utapi client.


rahulreddy:
👍

MathieuCassagne:
LGTM :+1:

PR #193, opened 2016-09-05
Repair object [to review only]

lamphamsy:
Repair process of uploaded objects are scheduled via cron
A repair of an object is triggered every `number` of `unit`s where:
- `unit` could be `day`, `week`, `month`
- `number` is frequency for repair
  Repair job starts at 2am + minutes, seconds of uploading time.

Repair job is assigned for every uploaded object. When an object is deleted, its
scheduled repair job is consequently clear.
In the case of un-recoverable object, its repair job is also clear.


lib/data/dpa/dpClient.js
```diff
@@ -68,6 +70,14 @@ const placementPolicy = {
 };
 
 const CHUNK_SIZE = 64 * 1024; // 64KB
+
+// default params for repair objects
+const repairParams = {
+    schedule: '00 00 02 * * 0',   // 02h00 Sunday, every week
```
ThibaultRiviere: You can remove the secs it's optional :)

lib/data/dpa/dpClient.js
```diff
@@ -228,6 +241,24 @@ class DpClient {
...
         }
 
+        function done(key, cb) {
+            // start a cron job for repair object
+            const job = new CronJob({
+                cronTime: this.cronParams.schedule,
+                onTick: () => {
+                    this.repair(key, err => {
+                        if (err) {
+                            process.stderr.write(`repair ${key}: ${err}\n`);
```
ThibaultRiviere: why ?

lib/data/dpa/dpClient.js
```diff
@@ -253,7 +284,7 @@ class DpClient {
                     }
                     donesNb += keysNb;
                     if (donesNb === codeLen) {
-                        return callback(null, key);
+                        return done.bind(this)(key, callback);
```
ThibaultRiviere: We should avoid the bind function for performances reasons .

lib/data/dpa/dpClient.js
```diff
@@ -429,12 +461,154 @@ class DpClient {
...
+                        return undefined;
+                    }, fragKey);
+                });
+                return undefined;
+            });
+            return undefined;
+        }
+
+        keys.forEach((keyFrag, fragIdx) => {
+            this.checkAvailFrag.bind(this)(keyFrag, log, (err, val) => {
```
ThibaultRiviere: I think you don't need the bind here.

lib/data/dpa/dpClient.js
```diff
@@ -429,12 +461,154 @@ class DpClient {
...
+                if (err) {
+                    lostFragsIds.push(fragIdx);
+                } else {
+                    fragments.push(val);
+                }
+                if (lostFragsIds.length + fragments.length === codeLen) {
+                    if (lostFragsIds.length > 0) {
+                        return repairFrags.bind(this)(lostFragsIds, fragments,
```
ThibaultRiviere: Same here, you can define repairFlags as lambda for avoiding the bind .

lib/data/dpa/dpClient.js
```diff
@@ -429,12 +461,154 @@ class DpClient {
...
+        function getFrag(key, reqUids, retriesNb, cb) {
+            log.debug('Checking availability fragment',
+                { key, retry: retriesNb });
+            client.get(key, null, reqUids, (err, val) => {
+                if (err) {
+                    const newRetriesNb = retriesNb + 1;
+                    if (newRetriesNb > maxRetriesNb) {
+                        return cb(err);
+                    }
+                    setTimeout(getFrag, interval, key, reqUids, newRetriesNb,
```
ThibaultRiviere: Be careful here, you will loose your this, i saw it's not use but because later when you call this function you are binding the this. So maybe you can bind the this here if needed or remove the bind on line 610

lib/data/dpa/dpClient.js
```diff
@@ -429,12 +461,154 @@ class DpClient {
...
+                                callback);
+                    return undefined;
+                }
+                const buf = [];
+                val.on('data', buffer => {
```
ThibaultRiviere: You should also listen the event error .

lib/data/dpa/dpClient.js
```diff
@@ -429,12 +461,154 @@ class DpClient {
...
+                });
+                val.on('end', () => {
+                    const frag = Buffer.concat(buf);
+                    return cb(null, frag);
+                });
+                return undefined;
+            });
+        }
+
+        return getFrag.bind(this)(key, key, 0, callback);
```
ThibaultRiviere: I think the second parameters is : ```log.getSerializeUids()```

lib/data/dpa/dpClient.js
ThibaultRiviere: i think the ```newRetriesNb``` will override the callback right ?

lib/data/dpa/dpClient.js
```diff
@@ -429,12 +461,154 @@ class DpClient {
...
+
+        assert.strictEqual(typeof key, 'string');
+
+        function getFrag(key, reqUids, retriesNb, cb) {
+            log.debug('Checking availability fragment',
+                { key, retry: retriesNb });
+            client.get(key, null, reqUids, (err, val) => {
+                if (err) {
+                    const newRetriesNb = retriesNb + 1;
+                    if (newRetriesNb > maxRetriesNb) {
```
ThibaultRiviere: This is not important but you begin to 0, and then check if it's more than maxRetriesNb witch is 3.
So the first fail you will increase to 1, then to 2, and again to 3.
3 is not more than 3 so you will try it again, and then return the cb with the error, 
At this end it's 4 shots.

lamphamsy:
@ThibaultRiviere : thanks so much

The code is updated

ThibaultRiviere:
@lamphamsy  my pleasure :+1:

lib/data/dpa/dpClient.js
```diff
@@ -172,39 +230,39 @@ class DpClient {
         let fragCursor = 0;
         let dataLen = 0;
 
-        // stream a buffer
-        function streamFrag(frag, fragIdx, cb) {
+        // store a fragment
+        const streamFrag = (frag, fragIdx, cb) => {
```
michael-zapata: Careful with functions created inside the scope of another function. It's an optimisation killer, because it means creating the `Function` Object on every call, instead of having it available in the global context at all times.

lib/data/dpa/dpClient.js
```diff
@@ -438,15 +516,169 @@ class DpClient {
...
                 }
                 return undefined;
             });
         });
     }
+
+    /**
+     * This trigger a repair of an object
+     * @param {String} key - The object's key
+     * @param {Callback} callback - callback
```
michael-zapata: Would it be possible to document the callback arguments?

lamphamsy:
Closed

PR #199, opened 2016-09-07
fix: pass null for push metrics on object put

### LINKED ISSUE #198: Utapi: PUT object does not increment number of objects
On Object PUT the number objects should be incremented on the push metrics call.

rahulreddy:
pushMetrics call for object PUT uses the content-length to calculate
incoming bytes, number of object counters. A null would indicate that
the object is created new and any number 0 or greater than 0 indicates
an existing object which would not alter the count of number of objects
in the bucket.

fixes #198


nicolas2bert:
👍  It works as expected !!

LaurenSpiegel:
👍

PR #204, opened 2016-09-13
Ft/batch policy eval

LaurenSpiegel:
Send an array of request contexts to Vault to do authorization checks as a batch.

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -1,6 +1,6 @@
 import querystring from 'querystring';
 
-import { auth, policies, errors } from 'arsenal';
+import { auth, errors } from 'arsenal';
 
 import bucketDelete from './bucketDelete';
 import bucketGet from './bucketGet';
@@ -21,14 +21,17 @@ import objectHead from './objectHead';
 import objectPut from './objectPut';
 import objectPutACL from './objectPutACL';
 import objectPutPart from './objectPutPart';
+import prepareRequestContexts from
+    './apiUtils/authorization/prepareRequestContexts';
 import serviceGet from './serviceGet';
 import vault from '../auth/vault';
 
-const RequestContext = policies.RequestContext;
 auth.setAuthHandler(vault);
 
 const api = {
     callApiMethod(apiMethod, request, log, callback, locationConstraint) {
+        let sourceBucket;
+        let sourceObject;
         if (apiMethod === 'objectCopy') {
             let source =
                 querystring.unescape(request.headers['x-amz-copy-source']);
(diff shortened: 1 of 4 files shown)


LaurenSpiegel:
@ironman-machine batch time
SCALITY_VAULT_BRANCH=ft/batchPolicyEval

michael-zapata:
Could we have some tests to go along with it?

LaurenSpiegel:
I added tests to integration.  We are not doing in memory policy evaluation.

lib/api/api.js
```diff
@@ -21,19 +21,22 @@ import objectHead from './objectHead';
...
 const api = {
     callApiMethod(apiMethod, request, log, callback, locationConstraint) {
+        let sourceBucket;
+        let sourceObject;
         if (apiMethod === 'objectCopy') {
             let source =
                 querystring.unescape(request.headers['x-amz-copy-source']);
-            // If client sends the source bucket/object with a leading /,
-            // remove it
+                // If client sends the source bucket/object with a leading /,
```
michael-zapata: Considering this applies to the code below it, why not leave the indentation as is?

lib/api/api.js
```diff
@@ -42,47 +45,37 @@ const api = {
...
+            authorizationResults) => {
             if (err) {
                 log.trace('authentication error', { error: err });
                 return callback(err);
             }
+            // TODO: When implement batch delete, check apiMethod
+            // and handle 'Deny' on key by key basis
+            if (authorizationResults) {
+                for (let i = 0; i < authorizationResults.length; i ++) {
+                    if (authorizationResults[i].verdict === 'Deny') {
```
michael-zapata: Why use a string? Can't a boolean work just as well while limiting typos?

lib/api/api.js
```diff
@@ -42,47 +45,37 @@ const api = {
...
+        return auth.doAuth(request, log, (err, userInfo,
+            authorizationResults) => {
             if (err) {
                 log.trace('authentication error', { error: err });
                 return callback(err);
             }
+            // TODO: When implement batch delete, check apiMethod
+            // and handle 'Deny' on key by key basis
+            if (authorizationResults) {
+                for (let i = 0; i < authorizationResults.length; i ++) {
```
MathieuCassagne: You can remove the space in ```i ++``` :)

lib/api/api.js
LaurenSpiegel: fixed

lib/api/api.js
LaurenSpiegel: changed to isAllowed

lib/api/api.js
LaurenSpiegel: changed

LaurenSpiegel:
@ironman-machine get it on
SCALITY_VAULT_BRANCH=ft/batchPolicyEval
SCALITY_INTEGRATION_BRANCH=test/multipleRequestContexts

lib/api/api.js
```diff
@@ -42,47 +45,37 @@ const api = {
...
             if (err) {
                 log.trace('authentication error', { error: err });
                 return callback(err);
             }
+            // TODO: When implement batch delete, check apiMethod
+            // and handle 'Deny' on key by key basis
+            if (authorizationResults) {
+                for (let i = 0; i < authorizationResults.length; i++) {
+                    if (!authorizationResults[i].isAllowed) {
+                        log.trace('authorization denial from Vault');
```
rahulreddy: It would be helpful to log which resource was denied authorization, do we have the info here?

LaurenSpiegel: For now only object copy has more than one requestContext and the resource is the same for both checks.  I have added to the TODO note for batch delete to add that log.

LaurenSpiegel:
End to end passed!

rahulreddy:
👍

PR #207, opened 2016-09-14
Data file backend: fix bug for put operation

### LINKED ISSUE #206: Data file backend: bug in put operation
Source file: https://github.com/scality/S3/blob/master/lib/data/file/backend.js#L35

It's possible that stream ends (#L67) before the `fs.write` operation finishes (#L52). Therefore, if `fs.write` fails, there are double return callbacks.

lamphamsy:
Fix #206 
It's possible that stream ends before the `fs.write` operation finishes.
Double checks to make sure that only one callback returns.

### DIFF
--- a/lib/data/file/backend.js
+++ b/lib/data/file/backend.js
@@ -37,40 +37,39 @@ export const backend = {
         // Consider making async
         const key = crypto.randomBytes(20).toString('hex');
         const filePath = getFilePath(key);
-
         request.pause();
-
         fs.open(filePath, 'wx', (err, fd) => {
             if (err) {
                 log.error('error opening filePath', { error: err });
                 return callback(errors.InternalError);
             }
+            const fileStream = fs.createWriteStream(filePath, { fd });
+
             request.resume();
-            return request.on('data', data => {
-                // Disable data events as we need to wait for fs.write callback
-                request.pause();
-                return fs.write(fd, data, 0, data.length,
-                    err => {
-                        if (err) {
-                            log.error('error writing data', { error: err });
-                            fs.close(fd);
-                            return callback(errors.InternalError);
-                        }
-                        return request.resume(); // Allow data events again
-                    });
-            })
-            .on('error', err => {
-                log.error('error streaming data from request', { error: err });
-                fs.close(fd);
+            request.pipe(fileStream, { end: false }).on('error', err => {
+                log.error('error streaming data from request on write',
+                    { error: err });
                 return callback(errors.InternalError);
-            })
-            .on('end', () => {
-                log.debug('finished writing data', { key });
-                fs.fsync(fd, () => {
-                    fs.close(fd);
+            });
+            request.on('error', err => {
+                log.error('error streaming data from request on read',
+                    { error: err });
+                // close fileStream
+                return fs.close(fd, () => callback(errors.InternalError));
+            }).on('end', () => {
+                fs.fsync(fd, err => {
+                    fileStream.end();
+                    if (err) {
+                        log.error('error streaming data from request on fsync',
+                            { error: err });
+                        return callback(errors.InternalError);
+                    }
+                    log.debug('finished writing data', { key });
                     return callback(null, key);
                 });
+                return undefined;
             });
+            return undefined;
         });
     },
 


lib/data/file/backend.js
```diff
@@ -65,11 +74,15 @@ export const backend = {
...
             })
             .on('end', () => {
-                log.debug('finished writing data', { key });
-                fs.fsync(fd, () => {
-                    fs.close(fd);
-                    return callback(null, key);
-                });
+                streamEnd = true;
+                // Only in flowing mode when fs.write finished
+                if (request._readableState.flowing) {
```
LaurenSpiegel: I don't think we should be relying on the internals of the readable stream to know if our last write is finished.  Instead, why don't we have a variable writeFinished initially set to false and it will only be set to true in the fs.write callback.  So when end is called on the readable stream, we only close the file and call the callback if the writeFinished is set to true.  If it is false, we keep checking back after some interval until it is finished and then close and call the callback.  @GiorgioRegni , does that make sense to you?

ThibaultRiviere: By the way why not use the ```finished``` event ?

LaurenSpiegel: @ThibaultRiviere, isn't the finish event for a writable stream and the request here is acting as a readable? @RachedBenMustapha, pointed out that with my suggestion if we have more than one chunk being finished up, we could still end early.  He suggests that we push each chunk into an array on each data event and push null into the array when end is called on the readable.  Then in the on.("data") block, we pop off of the end of the array and write it if it is a chunk or close the file and call the callback if the end of the array contains null.

ThibaultRiviere: @LaurenSpiegel yes you right and this is why i was thinking about it . I was thinking that the purpose was to wait the ```end``` of the readable stream .
So instead of putting a variable waiting for the end of the readable stream who i think doesn't mean that all datas have been called at the next call of ```on.('data')```.   Even with the ```resume``` witch is asynchronous.
Why not just listen ```finish``` on the writable stream and ```sync``` when all the data have been write ?

LaurenSpiegel: As @ThibaultRiviere and I discussed, we are not writing to a writeStream.  We could use fs.createWriteStream instead of just writing to the file and then we would have a stream to pipe to.

lamphamsy: @LaurenSpiegel : `fs.createWriteStream ` is cool

michael-zapata:
Since this is a fix, shouldn't this be targeting a release branch first?

LaurenSpiegel:
Since it's a fix only to the file backend, I don't think it is relevant for the release.

lib/data/file/backend.js
```diff
@@ -38,39 +38,16 @@ export const backend = {
...
-                    });
-            })
-            .on('error', err => {
-                log.error('error streaming data from request', { error: err });
-                fs.close(fd);
-                return callback(errors.InternalError);
-            })
-            .on('end', () => {
-                log.debug('finished writing data', { key });
-                fs.fsync(fd, () => {
```
LaurenSpiegel: Don't we still want to call fsync?

lamphamsy:
The code is updated

lib/data/file/backend.js
lamphamsy: It should be used. Thanks

lib/data/file/backend.js
```diff
@@ -37,40 +37,38 @@ export const backend = {
...
-                    fs.close(fd);
-                    return callback(null, key);
+            });
+            request.on('error', err => {
+                log.error('error streaming data from request on read',
+                    { error: err });
+                // close fileStream
+                return fs.close(fd, () => callback(errors.InternalError));
+            }).on('end', () => {
```
ThibaultRiviere: The next part should be after the ```finish``` event of the writable stream . 
And i don't understand why not ending the writableStream at the end ?
```js
request.pipe(fileStream).on('error', err => {
    // log your error and all 
}).on('finish', () => {
       fs.fsync(fd, err => {
             if (err) {
                 log.error('error streaming data from request on fsync',
                     { error: err });
                 return callback(errors.InternalError);
             }
             log.debug('finished writing data', { key });
             return fs.close(fd, () => callback(null, key));
        });
});
```

lamphamsy: Yes for ending the writableStream
No for `finish` event because at that moment, `fd` is no longer available

PR #221, opened 2016-09-22
fix: error response for object level encryption

### LINKED ISSUE #165: Should return error when encryption header sent on request (other than bucket level encryption)
We don't currently support object level encryption so an error should be returned if these headers are sent.

rahulreddy:
fix #165

### DIFF
--- a/lib/routes/routePUT.js
+++ b/lib/routes/routePUT.js
@@ -6,6 +6,14 @@ import routesUtils from './routesUtils';
 import utils from '../utils';
 import pushMetrics from '../utilities/pushMetrics';
 import constants from '../../constants';
+const encryptionHeaders = [
+    'x-amz-server-side-encryption',
+    'x-amz-server-side-encryption-customer-algorithm',
+    'x-amz-server-side-encryption-aws-kms-key-id',
+    'x-amz-server-side-encryption-context',
+    'x-amz-server-side-encryption-customer-key',
+    'x-amz-server-side-encryption-customer-key-md5',
+];
 
 export default function routePUT(request, response, log, utapi) {
     log.debug('routing request', { method: 'routePUT' });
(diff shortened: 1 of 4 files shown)


lib/routes/routePUT.js
```diff
@@ -122,6 +122,21 @@ export default function routePUT(request, response, log, utapi) {
             log.end().addDefaultFields({
                 contentLength: request.headers['content-length'],
             });
+
+            // object level encryption
+            const encryptHeaders = [
```
LaurenSpiegel: maybe move this to constants or at least outside of the actual route function.

lib/routes/routePUT.js
```diff
@@ -122,6 +122,21 @@ export default function routePUT(request, response, log, utapi) {
...
+                'x-amz-server-side-encryption',
+                'x-amz-server-side-encryption-customer-algorithm',
+                'x-amz-server-side-encryption-aws-kms-key-id',
+                'x-amz-server-side-encryption-context',
+                'x-amz-server-side-encryption-customer-key',
+                'x-amz-server-side-encryption-customer-key-md5',
+            ];
+            if (encryptHeaders.some(i => request.headers[i] !== undefined)) {
```
LaurenSpiegel: move this to top of else to put part and object copy are impacted as well

rahulreddy: I moved it to the top so that put part and object copy get the benefit as well.

lib/routes/routePUT.js
rahulreddy: moved it outside of routes function.

rahulreddy:
Added a commit which
 - skips copyObject test which uses encryption option
 - adds a test for object put part

nicolas2bert:
👍

LaurenSpiegel:
@ironman-machine check it

LaurenSpiegel:
@ironman-machine taking a break?

LaurenSpiegel:
@ironman-machine revived?

LaurenSpiegel:
@ironman-machine ready to play nicely now?

LaurenSpiegel:
@ironman-machine get it on

rahulreddy:
@ironman-machine again

PR #223, opened 2016-09-22
Kinetic backend

lamphamsy:
Added kinetic backend into S3 with put/get/delete operations
It support currently object's size <= 1MB
@AntoninCoulibaly : I developed this from your branch (https://github.com/scality/S3/tree/dev/ft/kineticBackend)

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -263,6 +263,15 @@ class Config {
             dataPath,
             metadataPath,
         };
+
+        if (config.kinetic && Array.isArray(config.kinetic)
+            && process.env.S3SPROXYD === 'kinetic') {
+            const tab = config.kinetic[0].split(':');
+            this.kinetic = {
+                host: tab[0],
+                port: tab[1],
+            };
+        }
         return config;
     }
 }
--- a/lib/data/kinetic/backend.js
+++ b/lib/data/kinetic/backend.js
@@ -1,0 +1,166 @@
+import crypto from 'crypto';
+import kinetic from 'kineticlib';
+import stream from 'stream';
+import net from 'net';
+
+import { errors } from 'arsenal';
+
+import config from '../../Config';
+
+let sequence = 0;
+
+const kDrives = {
+    port: config.kinetic ? config.kinetic.port : 8123,
+    host: config.kinetic ? config.kinetic.host : 'localhost',
+};
+
+function propError(pdu) {
+    const statusCode = pdu.getStatusCode();
+    if (statusCode !== kinetic.errors.SUCCESS) {
+        if (statusCode === kinetic.errors.NOT_FOUND) {
+            return errors.ObjNotFound;
+        }
+        return pdu.getErrorMessage();
+    }
+    return undefined;
+}
+
+function newSocket(callback) {
+    const socket = new net.Socket({ allowHalfOpen: false }).pause();
+    socket.setKeepAlive(true);
+    socket.unref();
+    return socket.connect(kDrives, () => {
+        kinetic.streamToPDU(socket, (err, pdu) => {
+            if (err) {
+                return callback(err);
+            }
+            const err1 = propError(pdu);
+            return callback(err1, socket);
+        });
+    });
+}
+
+function putKinetic(socket, value, options, callback) {
+    const key = crypto.randomBytes(20);
+    const obj = Buffer.concat(value);
+    const tag = crypto.createHmac('sha1', 'asdfasdf').update(obj).digest();
+    const pdu = new kinetic.PutPDU(0, key, obj.length, tag, options);
+    if (sequence < Number.MAX_VALUE) {
+        ++sequence;
+    } else {
+        sequence = 0;
+    }
+
+    const header = pdu.read();
+    const len = header.length + obj.length;
+
+    socket.write(Buffer.concat([header, obj], len), err => {
+        if (err) {
+            return callback(err);
+        }
+        return kinetic.streamToPDU(socket, (err, pdu) => {
+            const err1 = propError(pdu);
+            socket.destroy();
+            return callback(err || err1, key);
+        });
+    });
+}
+
+function getKinetic(socket, pdu, callback) {
+    let count = 0;
+    const chunkSize = pdu.getChunkSize();
+    const value = [];
+    const err = propError(pdu);
+    socket.resume();
+    socket.on('data', chunk => {
+        value.push(chunk);
+        count += chunk.length;
+        if (count === chunkSize) {
+            socket.end();
+        }
+    }).on('end', () => {
+        socket.destroy();
+        return callback(err, new stream.Readable({
+            read() {
+                this.push(Buffer.concat(value));
+                this.push(null);
+            },
+        }));
+    });
+}
+
+function deleteKinetic(socket, key, callback) {
+    const pdu = new kinetic.DeletePDU(0, key);
+    if (sequence < Number.MAX_VALUE) {
+        ++sequence;
+    } else {
+        sequence = 0;
+    }
+    socket.write(pdu.read(), err => {
+        if (err) {
+            return callback(err);
+        }
+        return kinetic.streamToPDU(socket, (err, pdu) => {
+            if (err) {
+                return callback(err);
+            }
+            const err1 = propError(pdu);
+            socket.destroy();
+            return callback(err || err1);
+        });
+    });
+}
+
+const backend = {
+    put: function putK(request, size, keyContext, reqUids, callback) {
+        const value = [];
+        request.on('data', data => {
+            value.push(data);
+        }).on('end', err => {
+            if (err) {
+                return callback(err);
+            }
+            return newSocket((err, socket) => {
+                const options = {
+                    synchronization: 'WRITEBACK', // FLUSH
+                };
+                putKinetic(socket, value, options, callback);
+            });
+        });
+    },
+
+    get: function getK(key, range, reqUids, callback) {
+        newSocket((err, sock) => {
+            if (err) {
+                return callback(err);
+            }
+            const pdu = new kinetic.GetPDU(0, new Buffer(key.data));
+            if (sequence < Number.MAX_VALUE) {
+                ++sequence;
+            } else {
+                sequence = 0;
+            }
+            const header = pdu.read();
+            return sock.write(header, err => {
+                if (err) {
+                    return callback(err);
+                }
+                return kinetic.streamToPDU(sock, (err, pdu) => {
+                    getKinetic(sock, pdu, callback);
+                });
+            });
+        });
+    },
+
+    delete: function delK(keyValue, reqUids, callback) {
+        const key = Buffer.from(keyValue);
+        newSocket((err, socket) => {
+            if (err) {
+                return callback(err);
+            }
+            return deleteKinetic(socket, key, callback);
+        });
+    },
+};
+
+export default backend;
--- a/config.json
+++ b/config.json
@@ -27,6 +27,7 @@
         "host": "localhost",
         "port": 8500
     },
+    "kinetic": ["localhost:8123"],
     "clusters": 10,
     "log": {
         "logLevel": "info",
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -1,11 +1,12 @@
 import async from 'async';
 import { errors } from 'arsenal';
+import assert from 'assert';
 import Sproxy from 'sproxydclient';
 import file from './file/backend';
+import kinetic from './kinetic/backend';
 import inMemory from './in_memory/backend';
 import config from '../Config';
 import MD5Sum from '../utilities/MD5Sum';
-import assert from 'assert';
 import kms from '../kms/wrapper';
 
 let client;
@@ -17,6 +18,9 @@ if (config.backends.data === 'mem') {
 } else if (config.backends.data === 'file') {
     client = file;
     implName = 'file';
+} else if (config.backends.data === 'kinetic') {
+    client = kinetic;
+    implName = 'kinetic';
 } else if (config.backends.data === 'scality') {
     client = new Sproxy({
         bootstrap: config.sproxyd.bootstrap,
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "bucketclient": "scality/bucketclient",
     "commander": "^2.9.0",
     "ioctl": "^2.0.0",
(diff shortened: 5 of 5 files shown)


lib/Config.js
```diff
@@ -263,6 +263,13 @@ class Config {
             dataPath,
             metadataPath,
         };
+
+        if (config.kinetic && Array.isArray(config.kinetic)
+           && process.env.S3DATA === 'kinetic') {
+            this.kinetic = {};
+            this.kinetic.host = config.kinetic[0].split(':')[0];
```
ThibaultRiviere: Should we trigger an error if the config kinetic doesn't exist ?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,179 @@
...
+
+export const ds = [];
+
+const kDrives = {
+    port: config.kinetic ? config.kinetic.port : 8123,
+    host: config.kinetic ? config.kinetic.host : 'localhost',
+};
+
+function newSocket() {
+    const socket = new net.Socket({ allowHalfOpen: false });
```
ThibaultRiviere: why the ```allowHalfOpen```?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,179 @@
...
+        // This message contains only status of response. Process 2nd message
+        return decodeRecMsg(socket, cb);
+    });
+}
+
+const backend = {
+    put: function x(request, size, keyContext, reqUids, callback) {
+        const value = [];
+        request.on('data', data => {
+            value.push(data);
```
ThibaultRiviere: If you wait the event ```end``` that mean you keep all the data in memory?

lib/Config.js
```diff
@@ -234,8 +234,8 @@ class Config {
             }
             this.authData = authData;
         }
-        if (process.env.S3SPROXYD) {
-            data = process.env.S3SPROXYD;
+        if (process.env.S3DATA) {
+            data = process.env.S3DATA;
```
AntoninCoulibaly: Can you put this change in another commit ?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,179 @@
+import crypto from 'crypto';
+import kinetic from 'kineticlib';
+import stream from 'stream';
+import net from 'net';
+
+import { errors } from 'arsenal';
+
+import config from '../../Config';
```
AntoninCoulibaly: Is it possible to have some logs ?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,179 @@
...
+
+let sequence = 1;
+
+export const ds = [];
```
AntoninCoulibaly: It is not used right ?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,179 @@
...
+        socket.write(pdu.read(), err => {
+            if (err) {
+                return callback(err);
+            }
+            let statusCode;
+            socket.on('data', chunk => {
+                const recPDU = new kinetic.PDU(chunk);
+                statusCode = recPDU.getStatusCode();
+                socket.end();
+            }).on('end', err => {
```
AntoninCoulibaly: I think it could be great to have smaller functions

lib/Config.js
```diff
@@ -263,6 +263,13 @@ class Config {
             dataPath,
             metadataPath,
         };
+
+        if (config.kinetic && Array.isArray(config.kinetic)
+           && process.env.S3DATA === 'kinetic') {
```
GuillaumeGomez: The "&&" should be preceded by one more space.

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,155 @@
...
+    port: config.kinetic ? config.kinetic.port : 8123,
+    host: config.kinetic ? config.kinetic.host : 'localhost',
+};
+
+function propError(pdu) {
+    const statusCode = pdu.getStatusCode();
+    let err;
+
+    if (statusCode !== kinetic.errors.SUCCESS) {
+        if (statusCode === kinetic.errors.NOT_FOUND) {
```
GuillaumeGomez: I would have wrote this as follow:

```js
        if (statusCode === kinetic.errors.NOT_FOUND) {
            return errors.ObjNotFound;
        }
        return kinetic.getErrorName(statusCode);
```

And so the `err` variable becomes useless.

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,155 @@
...
+            err = errors.ObjNotFound;
+        } else {
+            err = kinetic.getErrorName(statusCode);
+        }
+        return err;
+    }
+}
```
GuillaumeGomez: I think a return statement is missing here.

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,156 @@
...
+    const pdu = new kinetic.PutPDU(sequence, key, obj.length, tag, options);
+    ++sequence;
+
+    socket.write(Buffer.concat([pdu.read(), obj]), err => {
+        if (err) {
+            return callback(err);
+        }
+        kinetic.streamToPDU(socket, (err, pdu) => {
+            const err1 = propError(pdu);
+            socket.destroy();
```
lamphamsy: Should it be better with `socket.end()` instead of destroying the socket?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,156 @@
...
+
+function getKinetic(socket, pdu, callback) {
+    let count = 0;
+    const value = [];
+    const err = propError(pdu);
+    socket.resume();
+    socket.on('data', chunk => {
+        value.push(chunk);
+        count += chunk.length;
+        if (count === pdu.getChunkSize()) {
```
lamphamsy: object size should be calculated once before here

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,156 @@
...
+    socket.write(pdu.read(), err => {
+        if (err) {
+            return callback(err);
+        }
+        kinetic.streamToPDU(socket, (err, pdu) => {
+            if (err) {
+                return callback(err);
+            }
+            const err1 = propError(pdu);
+            socket.destroy();
```
lamphamsy: same here: `socket.end()`?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,156 @@
...
+    const socket = new net.Socket({ allowHalfOpen: false }).pause();
+    socket.setKeepAlive(true);
+    socket.unref();
+    return socket.connect(kDrives.port, kDrives.host, () => {
+        kinetic.streamToPDU(socket, (err, pdu) => {
+            if (err) {
+                return callback(err);
+            }
+            const err1 = propError(pdu);
+            return callback(err1, socket, pdu.getChunkSize());
```
lamphamsy: 3rd arg is not used

lib/Config.js
```diff
@@ -263,6 +263,13 @@ class Config {
             dataPath,
             metadataPath,
         };
+
+        if (config.kinetic && Array.isArray(config.kinetic)
+            && process.env.S3DATA === 'kinetic') {
+            this.kinetic = {};
```
MathieuCassagne: You can write it like that :)
```
const tab = config.kinetic[0].split(':');
this.kinetic = {
    host: tab[0],
    port: tab[1],
};
```
You avoid one split, and you do not add property to the object.

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,157 @@
...
+import net from 'net';
+
+import { errors } from 'arsenal';
+
+import config from '../../Config';
+
+let sequence = 1;
+
+const kDrives = {
+    port: config.kinetic ? config.kinetic.port : 8123,
```
MathieuCassagne: You do the ternaire twice, maybe you can do it only once.

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,157 @@
...
+        return pdu.getErrorMessage();
+    }
+    return undefined;
+}
+
+function newSocket(callback) {
+    const socket = new net.Socket({ allowHalfOpen: false }).pause();
+    socket.setKeepAlive(true);
+    socket.unref();
+    return socket.connect(kDrives.port, kDrives.host, () => {
```
MathieuCassagne: Here you can just pass the kDrives object to connect as first argument.
https://nodejs.org/dist/latest-v4.x/docs/api/net.html#net_socket_connect_options_connectlistener

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,157 @@
...
+}
+
+function putKinetic(socket, value, options, callback) {
+    const key = crypto.randomBytes(20);
+    const obj = Buffer.concat(value);
+    const tag = crypto.createHmac("sha1", "asdfasdf").update(obj).digest();
+    const pdu = new kinetic.PutPDU(sequence, key, obj.length, tag, options);
+    ++sequence;
+
+    socket.write(Buffer.concat([pdu.read(), obj]), err => {
```
MathieuCassagne: Can you know the size of ```pdu.read()``` in advance, and pass the resultin length to Buffer.concat ? The doc says that it will be faster if you do so, it'll require one less loop.
https://nodejs.org/dist/latest-v4.x/docs/api/buffer.html#buffer_class_method_buffer_concat_list_totallength

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,157 @@
...
+        return undefined;
+    });
+}
+
+function getKinetic(socket, pdu, callback) {
+    let count = 0;
+    const chunkSize = pdu.getChunkSize();
+    const value = [];
+    const err = propError(pdu);
+    socket.resume();
```
MathieuCassagne: You do not need to resume the socket here, since you put a ```socket.on('data')``` the socket will pass in the flowing mode automatically.
https://nodejs.org/dist/latest-v4.x/docs/api/stream.html#stream_class_stream_readable

AntoninCoulibaly: The socket need to be resume because it is intentionally paused before, so we need to resume it for getting data.

AntoninCoulibaly:
@GuillaumeGomez @MathieuCassagne  comments addressed!

lib/data/kinetic/backend.js
MathieuCassagne: I might misunderstood the docs, nevermind :)

lib/Config.js
```diff
@@ -234,8 +234,8 @@ class Config {
             }
             this.authData = authData;
         }
-        if (process.env.S3SPROXYD) {
-            data = process.env.S3SPROXYD;
+        if (process.env.S3DATA) {
```
michael-zapata: Careful with that change. I think our internal process uses `S3SPROXYD` so you will need to update there before merging here.

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,159 @@
...
+        });
+    });
+}
+
+function putKinetic(socket, value, options, callback) {
+    const key = crypto.randomBytes(20);
+    const obj = Buffer.concat(value);
+    const tag = crypto.createHmac('sha1', 'asdfasdf').update(obj).digest();
+    const pdu = new kinetic.PutPDU(sequence, key, obj.length, tag, options);
+    ++sequence;
```
michael-zapata: This will overflow at some point, do we really want that? (That goes for all occurrences)

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,159 @@
...
+    const header = pdu.read();
+    const len = header.length + obj.length;
+
+    socket.write(Buffer.concat([header, obj], len), err => {
+        if (err) {
+            return callback(err);
+        }
+        kinetic.streamToPDU(socket, (err, pdu) => {
+            const err1 = propError(pdu);
+            socket.destroy();
```
michael-zapata: Performance concern: Can't we keep sockets alive instead of destroying them systematically?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,159 @@
...
+        });
+        return undefined;
+    });
+}
+
+const backend = {
+    put: function putK(request, size, keyContext, reqUids, callback) {
+        const value = [];
+        request.on('data', data => {
+            value.push(data);
```
michael-zapata: Are we really putting all the data in memory before writing it into the socket?

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,159 @@
...
+    socket.write(Buffer.concat([header, obj], len), err => {
+        if (err) {
+            return callback(err);
+        }
+        kinetic.streamToPDU(socket, (err, pdu) => {
+            const err1 = propError(pdu);
+            socket.destroy();
+            return callback(err || err1, key);
+        });
+        return undefined;
```
michael-zapata: Instead of returning `undefined`, you can return `kinetic.streamToPDU()` instead, that'll avoid a few lines in every function.

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,159 @@
...
+    });
+}
+
+function getKinetic(socket, pdu, callback) {
```
michael-zapata: This looks like we are putting everything in memory before even trying to send some data back. Could we avoid that?

lib/data/kinetic/backend.js
AntoninCoulibaly: Yes that is what we wrote in the commit message that for now we are not handling more than 1mB files we are currently on it.
Our idea was to split into multiple 1MB files.

lib/Config.js
AntoninCoulibaly: I will remove this change to do it in a different PR.

lib/data/kinetic/backend.js
AntoninCoulibaly: Not destroying the socket seems to have same performances but seems less stable

AntoninCoulibaly: So I investigated and it seems to come from StreamToPDU that using posix-read. So I guess the kineticlib will to have some changes. I think it could be better to have this changes in another PR.

lib/data/kinetic/backend.js
lamphamsy: We need it but it should be limited by `Number.MAX_VALUE`

lamphamsy:
LGTM 👍

vrancurel:
duplicate of #288

PR #224, opened 2016-09-22
Data placement

lamphamsy:
Fragments placement algorithm
An object is encoded into (k + m) fragments which are stored in data backend storage. Supposing the storage is structured in layers, e.g. Rack, Server, Drive. The storage structure is represented as a
topology.

An important parameter of each layer is the `replacement` flag specifying whether multiple fragments of an object can be stored in a same component.

Components on the last layer are considered as leaf. Fragments will be stored in these components. Components on upper layers do not stored data. They are called intermediate components.

Each component is assigned by its weight that represents its storage capacity. Leaf component has its own weight as its capacity. Intermediate component's weight is sum of all its children's weight.

Finding locations for (k + m) fragments is choosing (k + m) leaf components respecting following constraints:
- Replacement flag on each layer
- Weight of components, i.e. probability that a component of large weight is higher than that of smaller weight.

Location of each fragment is found by using a deterministic algorithm that uses the `Dispersion` part of the fragment's key as its input. In each layer, a binary sequence extracted from the dispersion part will be used to find an appropriate component respecting the above constraints.


init.js
```diff
@@ -65,19 +67,74 @@ if (os.type() === 'Linux' && os.endianness() === 'LE' && ioctl) {
...
+            // If already exists, move on
+            if (err && err.errno !== -17) {
+                return next(err);
+            }
+            if (topo[key].leaf) {
+                return next();
+            }
+            return createDirsTopo(topo[key], path, next);
+        });
+    }, err => callback(err));
```
MathieuCassagne: Just ```}, callback);``` is enough :)

init.js
```diff
@@ -65,19 +67,74 @@ if (os.type() === 'Linux' && os.endianness() === 'LE' && ioctl) {
...
+         logger.info('Init complete.  Go forth and store data.');
+     });
+} else {
+    // if topology does not exist, create it
+    // otherwise, import it
+    const file = `${__dirname}/${constants.topoFile}.json`;
+    let topology;
+    try {
+        const fileStat = fs.statSync(file);
+        if (fileStat.isFile()) {
```
MathieuCassagne: This check is already done in ```fs.readFileSync()``` and throw EISDIR if you try to read a directory.

lib/data/dpa/dp.js
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+    });
+    return noError;
+}
+
+function getLocations(topo, ids, keys) {
+    const res = ids.map(() => [dataPath]);
+    const indices = Array.from({ length: ids.length }, (v, k) => k);
+    const flag = bFS(topo.MD, ids, topo, 0, indices, res);
```
MathieuCassagne: You do not use the flag after, you can just put the function call in the condition.

lib/data/dpa/dpClient.js
```diff
@@ -716,11 +744,21 @@ class DpClient {
...
         };
 
         return getFrag(key, log.getSerializedUids(), 0, callback);
     }
+
+    getLocations(keys) {
+        const ids = keys.map(id => parseInt(id.slice(0, 6), 16).toString(2));
+        ids.forEach((id, idx) => {
+            while (ids[idx].length < 24) {
+                ids[idx] = `0${ids[idx]}`;
```
MathieuCassagne: You can replace the while with that, but maybe check performances before :)
```
ids[idx] = `${'0'.repeat(24)}${ids[idx]}`.slice(-24);
```

Or if you're sur that ```ids[idx]``` is less than 24char you can do:
```
ids[idx] = `${'0'.repeat(24 - ids[idx].length)}${ids[idx]}`;
```
Which get better performances I think.

lamphamsy:
@MathieuCassagne : thanks for your comments

lib/data/dpa/dpClient.js
```diff
@@ -231,6 +243,14 @@ class DpClient {
         let fragCursor = 0;
         let dataLen = 0;
 
+        // get locations for all fragments
+        const locations = this.getLocations(keys);
+        if (!locations) {
+            const err = 'Cannot find locations to store fragments';
```
MathieuCassagne: Maybe add a new Arsenal error in the future ?

lib/data/dpa/topology.js
```diff
@@ -84,16 +80,19 @@ function updateWeight(obj) {
 // generate weight distribution
 function genWeightDistr(metadata, obj, depth) {
     if (!obj.leaf && obj.constructor === Object) {
+        const max = Object.keys(obj).map(val =>
+            obj[val].maxFragsNb).filter(nb => nb !== undefined);
```
MathieuCassagne: Maybe you should invert the map and the filter function. I mean if you reduce the size of the array first, the second loop will be faster.

lib/data/dpa/topology.js
```diff
@@ -84,16 +80,19 @@ function updateWeight(obj) {
...
             //         id !== undefined),
             ids: Object.keys(obj).filter(key =>
                 obj[key].constructor === Object),
             pdf: [],
             cdf: [],
-            max: Object.keys(obj).map(val =>
-                obj[val].maxFragsNb).filter(nb => nb !== undefined),
         }];
+        if (max.length > 0) {
+            obj.wdistr.max = max;                       // eslint-disable-line
```
MathieuCassagne: Maybe you can set ```obj.wdistr.max``` to ```0``` when you declare ```obj.wdistr```, so you do not add a property to the object just after.

lamphamsy:
Closed due to #293

PR #225, opened 2016-09-23
Fix warnings and set warnings as hard errors

GuillaumeGomez:


rahulreddy:
Thanks for the effort @GuillaumeGomez. While I welcome the changes, I think we should actually fix those warnings than disabling them line-by-line.

GuillaumeGomez:
Tree types of lints are more complicated to fix:

* consistent-return: in lot of cases, the return is done in a callback.
* no-param-reassign: this lint is a bit stupid from my point of view (and dangerous). If you have to copy a type to be able to modify it, you have to face some useless data copies, performance killer in other words.
* prefer-arrow-callback: this one is a bit special: I kept it only in tests to keep code as easily readable as possible. I don't mind "fixing" the code but you'll lose test readability.

GuillaumeGomez:
I removed all `non-consistent-return` flags.

rahulreddy:
We went through this cycle before. Basically, if you don't like a linting rule open an issue in Guidelines and we can arrive at a consensus whether to make an exception or not. The whole point of the linter is to not worry about personal opinions like these coming in the way of writing code. We all agreed to abide by the Airbnb's style guide with a few exceptions.

- no-param-reassign: I have a few refactors coming in for routing I will take care of it step-by-step.
- consistent-return: Just adding return will do the job
- prefer-arrow-callback: Readability can be achieved by adding a comment above the function.

You can remove the max-warnings rule, linter disabling comments for now and I will take care of them with a few refactors.

rahulreddy:
Thanks for the changes. One last thing is removing // eslint-disable-next-line no-param-reassign
and max-warnings rule. I will handle that param reassign through refactors.

DavidPineauScality:
@rahulreddy Personally, I'd rather keep these comments, and the max-warnings=0 so that we can finally be **really** warned (through a CI failure) whenever we're adding new linter warnings.

The fact that we've slowly added warnings without even caring for them is not new, and is still nonsense. As such, I'd rather see some "not so good" comments to disable hard-to-address warnings for now, and allow enabling the switch to all errors. You can still address this through refactoring later on, tracking all the eslint-based comments, and remove them slowly as you seem to be willing to. At least, we won't be adding new warnings that we'll have to catch up with later.

Does this sound ok to you ?

GuillaumeGomez:
I rebased, however I don't intend to remove the `max-warnings=0` and the eslint comments. It'll actually make your task easier when you'll update the code and avoid any new regression.

rahulreddy:
@DavidPineauScality Thanks for the explanation. I was just worried that we were sweeping the current warnings under the rug.  I am ok with the comments, will address them with refactors later.
@GuillaumeGomez Thanks. I am ok with the max-warnings option now.

GuillaumeGomez:
cc @LaurenSpiegel

LaurenSpiegel:
@ironman-machine thoughts?

PR #229, opened 2016-09-26
Simplify topology dependency

lamphamsy:
1. Simplify topology
   Topology contains metadata property that is defined in `constant.js`.
   Each component contains three properties:
   `ids` is an array of its children's id
   `pdf` is probability distribution function of its children weight
   `cdf` is cumulative distribution function of its children weight
2. Relax topology dependency
   Only put operation require a topology.
   Get/delete/repair operations do not need topology.

Object's key is a concatenation of its fragments' location and its UKS key
Example: `${UKS_KEY}:${frag_loc1},${frag_loc2},...`


lib/data/dpa/dp.js
```diff
@@ -101,14 +99,14 @@ function bFS(meta, ids, obj, depth, indices, res) {
     return noError;
 }
 
-function getLocations(topo, ids, keys) {
-    const res = ids.map(() => [dataPath]);
+function getLocations(topo, ids) {
+    const res = ids.map(() => []);
```
MathieuCassagne: Maybe this for is more explicit.
```
const res = new Array(ids.length).fill([]);
```

lamphamsy:
Closed due to #293

PR #238, opened 2016-09-30
FT: Fix crypto encoding

alexandre-merle:
in Node v4, default crypto encoding is binary, this
PR specify this encoding to avoid any change in the future
(node v6 in mind)

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -64,5 +64,6 @@ export default {
     minimumAllowedPartSize: 5242880,
 
     // hex digest of sha256 hash of empty string:
-    emptyStringHash: crypto.createHash('sha256').update('').digest('hex'),
+    emptyStringHash: crypto.createHash('sha256')
+        .update('', 'binary').digest('hex'),
 };
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -251,7 +251,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             const binaryString = bufferedHex.toString('binary');
             // Get the md5 of the binary string
             const md5Hash = crypto.createHash('md5');
-            md5Hash.update(binaryString);
+            md5Hash.update(binaryString, 'binary');
             // Get the hex digest of the md5
             let aggregateETag = md5Hash.digest('hex');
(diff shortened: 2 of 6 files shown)


alexandre-merle:
@rahulreddy i know but the doc is actually wrong with this

```
> crypto.createHmac('sha256', 'test').update('intègre').digest()
<Buffer e4 6d ae 75 7b c0 e4 56 b2 20 0e e6 a3 0b b6 1e e1 65 69 f8 5c 62 f9 36 20 ee 83 c4 3d 36 19 16>
> crypto.createHmac('sha256', 'test').update('intègre', 'utf8').digest();
<Buffer 9c 83 39 85 10 97 cd 10 83 fd b4 ab f0 9c 87 f1 1e 95 d7 0a 7e 89 92 0f 2c ee c5 88 a3 07 92 8f>
> crypto.createHmac('sha256', 'test').update('intègre', 'binary').digest();
<Buffer e4 6d ae 75 7b c0 e4 56 b2 20 0e e6 a3 0b b6 1e e1 65 69 f8 5c 62 f9 36 20 ee 83 c4 3d 36 19 16>
>
```

I basically just set the encoding explicit without changing the encoding we use, if its an issue, let's handle this after :)

rahulreddy:
@alexandre-merle 😃 

Your example only proves that the docs are inconsistent about the default encoding that is applied, not the accuracy of the result.

hmac sha256 hex digest for intègre with secret key test is `9c8339851097cd1083fdb4abf09c87f11e95d70a7e89920f2ceec588a307928f` (you can try online tools or use python) and you would only get that result if you specify the encoding as `utf8`.The other thing is you cannot enforce encoding as in there is no conversion happening. So, if you specify binary it just assumes that the data you supplied is binary encoded string.

Now the tests pass because sha256 digest is the same for non-accented characters regardless of what encoding you specify.

alexandre-merle:
@ironman-machine go for it

alexandre-merle:
end to end passed @LaurenSpiegel :)

LaurenSpiegel:
In light of https://github.com/scality/S3/issues/249, merging this now.

PR #242, opened 2016-10-02
Data placement with multiple data backends

lamphamsy:
Data and parity fragments of an object can be stored in different data
backends. For now, `memory` and `file` backends are supported. There are
4 possibilities to assign to each backend:
- `none`: this backend does not contain fragment
- `data`: this backend contains only data fragments
- `parity`: this backend contains only parity fragments
- `both`: this backend contains both data and parity fragments

Locations of fragments will be concatenated to the object's key. Each
location is prefixed by a character: `m` for `memory` backend, `f` for
`file` backend.

Example of object's key ${UKS_KEY}:m${frag_loc1},f${frag_loc2},...


lamphamsy:
Closed due to #293

PR #250, opened 2016-10-05
Dockerfile for mem backend

nicolas2bert:
We are starting adding some kind of versioning for our S3 Docker image.

On Docker, we are now pushing four tags every time we generate a new S3 image:

- **file backend** : `latest` and `HASHCOMMIT`
- **in-memory backend**: `mem-latest` and `mem-HASHCOMMIT`

I started doing it here: https://hub.docker.com/r/scality/s3server/tags/
I will add a _Supported tags and respective Dockerfile links_ inside our [scality/s3server Docker repo](https://hub.docker.com/r/scality/s3server/)

We will use [DockerfileMem](https://github.com/scality/S3/blob/dev/Dockerfile-mem/DockerfileMem) for our in-memory backend S3 image.
`docker build -t 's3mem' -f 'DockerfileMem' ./`

### DIFF
--- a/DockerfileMem
+++ b/DockerfileMem
@@ -1,0 +1,11 @@
+FROM node:4
+MAINTAINER Giorgio Regni <gr@scality.com>
+
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
+RUN npm install
+ENV S3BACKEND mem
+CMD [ "npm", "start" ]
+
+EXPOSE 8000


LaurenSpiegel:
where is the dockerfile for the file backend then?

DockerfileMem
```diff
@@ -0,0 +1,11 @@
+FROM node:4
+MAINTAINER Giorgio Regni <gr@scality.com>
+
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
+RUN npm install
+ENV S3BACKEND mem
```
LaurenSpiegel: does this work without an '='?

nicolas2bert: Yes it does, I tested it.

PR #253, opened 2016-10-07
BF: Log errors or return errors from MD

### LINKED ISSUE #252: Hidden errors
I have identified a couple of methods where the errors from MD are being consumed without logging or sometimes not being returned as errors at all.
For example
https://github.com/scality/S3/blob/rel/6.2/lib/services.js#L688
https://github.com/scality/S3/blob/rel/6.2/lib/services.js#L651
https://github.com/scality/S3/blob/rel/6.2/lib/services.js#L688
https://github.com/scality/S3/blob/rel/6.2/lib/services.js#L480

bennettbuchanan:
Fix #252
* If a method is not logging an error from metadata, then log it
* If an error object is not being handled, then log it and return
  the callback, passing the error object

### DIFF
--- a/lib/services.js
+++ b/lib/services.js
@@ -450,6 +450,7 @@ export default {
             metadata.putObjectMD(bucketName, longMPUIdentifier,
                 multipartObjectMD, log, (err) => {
                     if (err) {
+                        log.error('error from metadata', { error: err });
                         return cb(err);
                     }
                     return cb();
@@ -478,6 +479,7 @@ export default {
         const mpuBucketName = `${constants.mpuBucketPrefix}${bucketName}`;
         metadata.getBucket(mpuBucketName, log, (err, mpuBucket) => {
             if (err) {
+                log.error('error from metadata', { error: err });
                 return cb(errors.NoSuchUpload);
             }
             let splitter = constants.splitter;
@@ -496,6 +498,7 @@ export default {
                 searchArgs.marker, searchArgs.delimiter, searchArgs.maxKeys,
                 log, (err, response) => {
                     if (err) {
+                        log.error('error from metadata', { error: err });
                         return cb(err);
                     }
                     if (response.Contents.length !== 1) {
@@ -629,6 +632,7 @@ export default {
         }
         metadata.putObjectMD(mpuBucketName, partKey, omVal, log, err => {
             if (err) {
+                log.error('error from metadata', { error: err });
                 return cb(err);
             }
             return cb(null);
@@ -649,6 +653,10 @@ export default {
         assert.strictEqual(typeof listingParams.splitter, 'string');
 
         metadata.getBucket(MPUbucketName, log, (err, bucket) => {
+            if (err) {
+                log.error('error from metadata', { error: err });
+                return cb(err);
+            }
             if (bucket === undefined) {
                 return cb(null, {
                     IsTruncated: false,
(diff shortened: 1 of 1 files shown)


lib/services.js
```diff
@@ -450,6 +450,7 @@ export default {
             metadata.putObjectMD(bucketName, longMPUIdentifier,
                 multipartObjectMD, log, (err) => {
                     if (err) {
+                        log.error('error from metadata', { error: err });
```
DavidPineauScality: The reason nothing was logged here is due to this log entry: https://github.com/scality/S3/blob/master/lib/metadata/wrapper.js#L74

Maybe bumping it to "error" instead of "warn" would be enough ?

lib/services.js
```diff
@@ -478,6 +479,7 @@ export default {
         const mpuBucketName = `${constants.mpuBucketPrefix}${bucketName}`;
         metadata.getBucket(mpuBucketName, log, (err, mpuBucket) => {
             if (err) {
+                log.error('error from metadata', { error: err });
```
DavidPineauScality: Here is the original log: https://github.com/scality/S3/blob/master/lib/metadata/wrapper.js#L50
I think that in this case, we're right to log an error now, as this bucket SHOULD exist and be accessible, due to the fact that we're already validating a multipart upload. But then, maybe we should log any error != 404 in the lower layer (wrapper or even lower)

rahulreddy: Back in the days we reasoned this as  - since we don't have enough error context in lower layers warn/trace/debug were chosen (this needs to be unified) and then we would log error in the api layer by filtering out the error codes which we think are safe.
If we just check for 404 there may be cases where 404 is not even expected like for example, putobjectMD call.

lib/services.js
bennettbuchanan: I will remove this "error" log and others that follow the same logic, but not bump any "warn" to "error" in the lower layers because perhaps there are reasons for not logging "error" there. For now, I will note that certain "warn" may need to be updated to "error".

lib/services.js
bennettbuchanan: For the method `metadata.getBucket` I will log an "error" if the `err` object is returned.

lib/services.js
DavidPineauScality: That's what I mean, I want to make your `perhaps` into a certainty :)
Also, this will help you learn S3's code by digging into it anyways

DavidPineauScality: I know this may sound like nitpicking, but are we sure we won't be missing error cases this way ? If yes, then fine with me.

lib/services.js
DavidPineauScality: Given our current... difficulties.. at investigating issues, I do not believe reducing the number of error logs in the lower layers to be a very good choice. Most errors will be always interpreted as errors, except very specific cases that we should Know and control.

These specific cases, I'm back to the 404 on bucket get, will have various interpretations depending on the context of the request. Those should be logged in the upper API layer. But most of the errors we encounter do not have this ambiguity, and should always be logged.

Worded another way, errors should always be logged at the very moment where the error is understood as a `500 internal error` issue. Which means a 404 within a validateMultipart should indeed log error. but a 500 in any context should always log error too.

That's my (and not only mine to be honest) point and hope for this PR which might lead to a slightly wider scope than probably originally thought.

rahulreddy:
@DavidPineauScality Do you mind finalizing the review so that we can merge this PR?

rahulreddy:
Thanks, I created an issue to track the discussion about log levels.

PR #254, opened 2016-10-08
CLEAN UP: second "if" never been accessed

nicolas2bert:
The second `if`has never been accessed.

_CASES:_ 
- `content-length`  undefined  && `x-amz-decoded-content-length` undefined  : **errors.MissingContentLength**(411)
- `content-length` is not an integer or is negatif: node `http` module filter the request (ie: request never reach our S3 server)
- `x-amz-decoded-content-length` is not an integer or is negatif **errors.BadRequest** (400)

### DIFF
--- a/lib/routes/routePUT.js
+++ b/lib/routes/routePUT.js
@@ -139,12 +139,14 @@ export default function routePUT(request, response, log, utapi) {
                     additionalHeaders);
             });
         } else {
-            if (Number.isNaN(request.parsedContentLength)) {
+            if (request.headers['content-length'] === undefined &&
+            request.headers['x-amz-decoded-content-length'] === undefined) {
                 return routesUtils.responseNoBody(errors.MissingContentLength,
                     null, response, 411, log);
             }
-            if (Number.isNaN(request.parsedContentLength)) {
-                return routesUtils.responseNoBody(errors.InvalidArgument,
+            if (Number.isNaN(request.parsedContentLength) ||
+            request.parsedContentLength < 0) {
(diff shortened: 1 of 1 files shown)


lib/routes/routePUT.js
```diff
@@ -143,10 +143,6 @@ export default function routePUT(request, response, log, utapi) {
                 return routesUtils.responseNoBody(errors.MissingContentLength,
                     null, response, 411, log);
             }
-            if (Number.isNaN(request.parsedContentLength)) {
-                return routesUtils.responseNoBody(errors.InvalidArgument,
```
DavidPineauScality: Looking at the error message, I'd understand that the first `if`was actually the erroneous one, and should check for the presence of the ContentLength header. Do we really want to remove this difference between invalid Argument and Missing Content length ?

LaurenSpiegel: I think the first one is actually what AWS returns.  @nicolas2bert to confirm.

nicolas2bert: @LaurenSpiegel @DavidPineauScality I compared our errors to the Amazon ones, I updated our code to be as close as possible.

LaurenSpiegel:
Looks good.  I would combine the commits to one commit with a message like: "Compat: Check for incorrect content length"

DavidPineauScality:
Looks this was thought hard about (in terms of error reporting and compat), so I'll approve this once @LaurenSpiegel 's last comment is addressed, since it would make for a clearer commit msg :)

PR #257, opened 2016-10-10
bucket versioning #2 : versioning path to MetaData

ttvi:
Step 1: S3 to support versioning for buckets.
Step 2 (this PR): S3 to provide routes for interacting with Metadata.
Step 3: S3 to provide versioning support for each operation.
Step 4: MetaData to support versioning.

### DIFF
--- a/lib/metadata/acl.js
+++ b/lib/metadata/acl.js
@@ -12,11 +12,12 @@ const acl = {
         metadata.updateBucket(bucket.getName(), bucket, log, cb);
     },
 
-    addObjectACL(bucket, objectKey, objectMD, addACLParams, log, cb) {
+    addObjectACL(bucket, objectKey, objectMD, addACLParams, log, cb, params) {
         log.trace('updating object acl in metadata');
         // eslint-disable-next-line no-param-reassign
         objectMD.acl = addACLParams;
-        metadata.putObjectMD(bucket.getName(), objectKey, objectMD, log, cb);
+        metadata.putObjectMD(bucket.getName(), objectKey, objectMD, log, cb,
+                params);
     },
 
     parseAclFromHeaders(params, cb) {
--- a/lib/metadata/bucketclient/backend.js
+++ b/lib/metadata/bucketclient/backend.js
@@ -1,15 +1,23 @@
 import assert from 'assert';
 
 import bucketclient from 'bucketclient';
+import { errors } from 'arsenal';
 import BucketInfo from '../BucketInfo';
 import { logger } from '../../utilities/logger';
 import config from '../../Config';
+
+const METADATA_DEFAULT_VERSION = 1;
 
 class BucketClientInterface {
     constructor() {
         assert(config.bucketd.bootstrap.length > 0,
                'bucketd bootstrap list is empty');
         const { bootstrap, log } = config.bucketd;
+        this.metadataVersion = METADATA_DEFAULT_VERSION; // 6.3.0 or earlier
+        this.gotMetadataInfo = false; // got metadata information ?
+                                      //     go ahead sending requests : wait
+        // TODO remove when getURIComponents has been fixed
+        this.extendedVersion = false;
         if (config.https) {
             const { key, cert, ca } = config.https;
             logger.info('bucketclient configuration', {
@@ -27,14 +35,46 @@ class BucketClientInterface {
             });
             this.client = new bucketclient.RESTClient(bootstrap, log);
         }
+        this.init();
+    }
+
+    init() {
+        // TODO add a specific getMetadataInformation function in bucketclient
+        const beginPath = '/default/metadataInformation/foo';
+        const log = logger.newRequestLogger();
+        this.client.request('GET', beginPath, log, null, null, (err, data) => {
+            if (!err) {
+                try {
+                    const info = JSON.parse(data);
+                    this.metadataVersion =
+                        Number.parseInt(info.metadataVersion, 10);
+                    this.extendedVersion =
+                        this.metadataVersion > METADATA_DEFAULT_VERSION;
+                } catch (error) {
+                    log.error('error parsing metadata information', { error });
+                }
+            } else if (!err.RouteNotFound) {
+                // retry on InternalError and DBAPINotReady
+                return setTimeout(() => { this.init(); }, 0);
+            }
+            this.gotMetadataInfo = true;
+            return null;
+        });
     }
 
     createBucket(bucketName, bucketMD, log, cb) {
+        if (this.gotMetadataInfo === false) {
+            return cb(errors.ServiceUnavailable);
+        }
         this.client.createBucket(bucketName, log.getSerializedUids(),
                                  bucketMD.serialize(), cb);
+        return null;
     }
 
     getBucketAttributes(bucketName, log, cb) {
+        if (this.gotMetadataInfo === false) {
+            return cb(errors.ServiceUnavailable);
+        }
         this.client.getBucketAttributes(bucketName, log.getSerializedUids(),
             (err, data) => {
                 if (err) {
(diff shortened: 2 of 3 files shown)


ttvi:
@ironman-machine check it please

ttvi:
@ironman-machine rebased

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,71 @@
...
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketPutVersioning',
+        log,
+    };
+
+    return waterfall([
+        next => services.metadataValidateAuthorization(metadataValParams,
+            (err, bucket) => next(err, bucket)), // to ignore null object
```
MathieuCassagne: Can be write like:
```
next => services.metadataValidateAuthorization(metadataValParams, next),
```

lib/api/bucketPutVersioning.js
```diff
@@ -0,0 +1,71 @@
...
+            // TODO all metadata updates of bucket should be using CAS
+            metadata.updateBucket(bucket.getName(), bucket, log, next);
+        },
+    ], err => {
+        if (err) {
+            log.trace('error processing request', { error: err,
+                method: 'bucketPutVersioning' });
+            return callback(err);
+        }
+        return callback(err, 'Versioning Set');
```
MathieuCassagne: ```err`` is always undefined or null here.

ttvi:
passing end-to-end here http://ci.ironmann.io/gh/scality/Integration/5042

lib/routes/routePUT.js
```diff
@@ -33,6 +36,40 @@ export default function routePUT(request, response, log, utapi) {
                     return routesUtils.responseNoBody(err, null, response, 200,
                         log);
                 });
+            } else if (request.query.versioning !== undefined) {
+                if (request.post === '') {
```
MathieuCassagne: Maybe checking the length property is faster than a string comparaison.

lib/api/bucketPutVersioning.js
ttvi: No. Because the `callback` of `metadataValidateAuthorization` takes three arguments: `err`, `bucket`, `objectMD`. It will pass `objectMD` to the next waterfall without being used there.

lib/api/bucketPutVersioning.js
ttvi: Doesn't matter.

lib/routes/routePUT.js
ttvi: I checked 10^9 times each approach. Same result.

ttvi:
Please review the last commit only.

ttvi:
@ironman-machine check this out.

ttvi:
@ironman-machine again

lib/metadata/acl.js
```diff
@@ -12,11 +12,12 @@ const acl = {
         metadata.updateBucket(bucket.getName(), bucket, log, cb);
     },
 
-    addObjectACL(bucket, objectKey, objectMD, addACLParams, log, cb) {
+    addObjectACL(bucket, objectKey, objectMD, addACLParams, log, cb, params) {
```
LaurenSpiegel: would it be possible to keep the cb as the last argument throughout this pr?  so params would come after addACLParmas

lib/metadata/bucketclient/backend.js
```diff
@@ -1,15 +1,23 @@
...
 class BucketClientInterface {
     constructor() {
         assert(config.bucketd.bootstrap.length > 0,
                'bucketd bootstrap list is empty');
         const { bootstrap, log } = config.bucketd;
+        this.metadataVersion = METADATA_DEFAULT_VERSION; // 6.3.0 or earlier
+        this.gotMetadataInfo = false; // got metadata information ?
+                                      //     go ahead sending requests : wait
+        // TODO remove when getURIComponents has been fixed
+        this.extendedVersion = false;
```
LaurenSpiegel: what is extendedVersion?

lib/metadata/bucketclient/backend.js
```diff
@@ -27,75 +35,144 @@ class BucketClientInterface {
             });
             this.client = new bucketclient.RESTClient(bootstrap, log);
         }
+        this.init();
+    }
+
+    init() {
+        // get metadata information
+        const beginPath = '/default/metadataInformation/foo';
```
LaurenSpiegel: will this path always be used?  It probably makes more sense to put the raw get request in bucketclient and have a wrapper function here.

lib/metadata/bucketclient/backend.js
```diff
@@ -27,75 +35,144 @@ class BucketClientInterface {
...
+            if (!err) {
+                try {
+                    const info = JSON.parse(data);
+                    this.metadataVersion = Number(info.metadataVersion);
+                    this.extendedVersion =
+                        this.metadataVersion > METADATA_DEFAULT_VERSION;
+                } catch (error) {
+                    log.error('error parsing metadata information', { error });
+                }
+            } else if (!err.RouteNotFound) {
```
LaurenSpiegel: what if err.RouteNotFound?

lib/metadata/bucketclient/backend.js
```diff
@@ -27,75 +35,144 @@ class BucketClientInterface {
...
+            }
+            this.gotMetadataInfo = true;
+            return null;
+        });
     }
 
     createBucket(bucketName, bucketMD, log, cb) {
+        if (this.gotMetadataInfo === false) {
+            // not really DBAPINotReady, but cannot make request anyway
+            return cb(errors.DBAPINotReady);
```
LaurenSpiegel: we are not sending internal errors like this out to the external client.  So, this should be errors.InternalError with a log so we can tell what is going on.

lib/metadata/bucketclient/backend.js
```diff
@@ -27,75 +35,144 @@ class BucketClientInterface {
...
+        }
         this.client.createBucket(bucketName, log.getSerializedUids(),
                                  bucketMD.serialize(), cb);
+        return null;
     }
 
     getBucketAttributes(bucketName, log, cb) {
+        if (this.gotMetadataInfo === false) {
+            return cb(errors.DBAPINotReady);
```
LaurenSpiegel: same here and below.

lib/metadata/bucketclient/backend.js
```diff
@@ -27,75 +35,144 @@ class BucketClientInterface {
...
+
+    init() {
+        // get metadata information
+        const beginPath = '/default/metadataInformation/foo';
+        const log = logger.newRequestLogger();
+        this.client.request('GET', beginPath, log, null, null, (err, data) => {
+            if (!err) {
+                try {
+                    const info = JSON.parse(data);
+                    this.metadataVersion = Number(info.metadataVersion);
```
LaurenSpiegel: Number.parseInt(info.metadataVersion, 10)

lib/metadata/bucketclient/backend.js
```diff
@@ -27,75 +35,144 @@ class BucketClientInterface {
...
+                return setTimeout(() => { this.init(); }, 0);
+            }
+            this.gotMetadataInfo = true;
+            return null;
+        });
     }
 
     createBucket(bucketName, bucketMD, log, cb) {
+        if (this.gotMetadataInfo === false) {
+            // not really DBAPINotReady, but cannot make request anyway
```
LaurenSpiegel: why "not really DABAPINotReady"?

lib/metadata/acl.js
ttvi: Yes, we have mentioned this in the previous step and I suggested to make it final step because changing the api involves searching for all references and modifying them in legacy code which is error prone.

lib/metadata/bucketclient/backend.js
ttvi: It's a flag saying whether the metadata deployment is supporting versioning or not (now metadata's getURIComponents is incompatible with versioning, see https://github.com/scality/MetaData/pull/706).

lib/metadata/bucketclient/backend.js
ttvi: I agree. Just that we know the shortcut.

lib/metadata/bucketclient/backend.js
ttvi: It will go to the next step updating the `gotMetadataInfo` without changing the `metadataVersion`. In this case it means metadata does not support the `metadataInformation` query thus does not support versioning.

lib/metadata/bucketclient/backend.js
ttvi: It can be `InternalError` including those http such as `ECONNRESET`, `ECONNREFUSE`, ... or those internal of metadata + the actual `DBAPINotReady`.
We're sending back `DBAPINotReady` or something similar saying that we cannot reach metadata now.

lib/metadata/bucketclient/backend.js
ttvi: I think `ServiceUnavailable` is a better fit.

lib/metadata/bucketclient/backend.js
ttvi: updated.

lib/metadata/bucketclient/backend.js
ttvi: damn, oops, done :-1: .

PR #261, opened 2016-10-10
Move config parsing to encrypt bucket function.

LaurenSpiegel:
If do not move this parsing, cannot run AWS_ON_AIR tests.

The encrypted bucket function is only meant for the S3 server.
However, by having the parsing done outside of the createEncryptedBucket function,
the process will crash before running tests with AWS_ON_AIR=true.


rahulreddy:
👍

PR #272, opened 2016-10-14
stress test for bucket creation/deletion

ttvi:
It's a functional test to check if the behavior of metadata is correct in the case of recycling buckets.
To verify to some extend https://github.com/scality/MetaData/issues/707.

### DIFF
--- a/tests/functional/aws-node-sdk/test/bucket/testBucketStress.js
+++ b/tests/functional/aws-node-sdk/test/bucket/testBucketStress.js
@@ -1,0 +1,41 @@
+import { S3 } from 'aws-sdk';
+import { times, timesSeries, waterfall } from 'async';
+
+import getConfig from '../support/config';
+
+const bucket = `stress-test-bucket-${Date.now()}`;
+const text = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890';
+const objectCount = 100;
+const loopCount = 10;
+
+function putObjects(s3, loopId, cb) {
+    times(objectCount, (i, next) => {
+        const params = { Bucket: bucket, Key: `foo${loopId}_${i}`, Body: text };
+        s3.putObject(params, next);
+    }, cb);
+}
+
+function deleteObjects(s3, loopId, cb) {
+    times(objectCount, (i, next) => {
+        const params = { Bucket: bucket, Key: `foo${loopId}_${i}` };
+        s3.deleteObject(params, next);
+    }, cb);
+}
+
+describe('aws-node-sdk stress test bucket', function testSuite() {
+    this.timeout(120000);
+    let s3;
+    before(() => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+    });
+
+    it('createBucket-putObject-deleteObject-deleteBucket loop', done =>
+        timesSeries(loopCount, (loopId, next) => waterfall([
+            next => s3.createBucket({ Bucket: bucket }, err => next(err)),
+            next => putObjects(s3, loopId, err => next(err)),
+            next => deleteObjects(s3, loopId, err => next(err)),
+            next => s3.deleteBucket({ Bucket: bucket }, err => next(err)),
+        ], err => next(err)), done)
+    );
+});


ttvi:
@ironman-machine

tests/functional/aws-node-sdk/test/bucket/testBucketStress.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+describe('aws-node-sdk stress test bucket', function testSuite() {
+    this.timeout(120000);
+    let s3;
+    // setup test
+    before(done => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+        setTimeout(done, 1000); // wait for bucketclient to query metadata
```
LaurenSpiegel: I'm unclear why we need this.  If we need this here, don't we need this in all tests?  If we need a delay before running tests because of the new metadata wrapper init, we should put that delay in the circle.yml and in integration.

tests/functional/aws-node-sdk/test/bucket/testBucketStress.js
```diff
@@ -0,0 +1,55 @@
...
+const objectCount = 100;
+const loopCount = 10;
+
+function putObjects(s3, cb) {
+    let count = 0;
+    times(objectCount, (i, next) => {
+        const params = { Bucket: bucket, Key: `foo${i}`, Body: text };
+        s3.putObject(params, error => {
+            if (error || ++count === objectCount) {
+                cb(error);
```
LaurenSpiegel: If the count is off but error is null, I don't think we will see anything here.  The cb will just be called with null so the error will be hidden (I suppose until we try to delete it but it will be hard to figure out what is going on at that point).  I think you should create the error object to make sure it exists if the count is off and log a message as well (just by using process.stdout.write).

tests/functional/aws-node-sdk/test/bucket/testBucketStress.js
```diff
@@ -0,0 +1,55 @@
...
+    }, () => {});
+}
+
+function deleteObjects(s3, cb) {
+    let count = 0;
+    times(objectCount, (i, next) => {
+        const params = { Bucket: bucket, Key: `foo${i}` };
+        s3.deleteObject(params, error => {
+            if (error || ++count === objectCount) {
+                cb(error);
```
LaurenSpiegel: same as above.

tests/functional/aws-node-sdk/test/bucket/testBucketStress.js
ttvi: Yeah I was confused when setting this up. It's the delay of S3 itself (https://github.com/scality/S3/pull/257) not the S3 api that we are instantiating here so there should be no `ServiceUnavailable` because S3 server has already been started.

ttvi: I mean it will be removed.

ttvi: And don't forget to take a look at https://github.com/scality/S3/pull/257.

tests/functional/aws-node-sdk/test/bucket/testBucketStress.js
LaurenSpiegel: Ah, I misread this.  I thought you were trying to catch a counting error.  You don't have to do all of this work with async.times.  It will keep track of where you are in the loop for you.  You can just call next(error) and then your last argument to the times function is the cb which gets called immediately if there is an error or at the end when you are done.

ttvi: It was me who misused `async.times`. Thanks Lauren for pointing this out.

rahulreddy:
Nice 😃 Wouldn't this make more sense to be in Integration?

LaurenSpiegel:
@rahulreddy, why in integration?  by having it here, we benefit from it testing all backends and it will be included in integration by virtue of integration picking up s3 tests.

rahulreddy:
Oh right, Integration runs S3 functional tests, nvm.

ttvi:
@ironman-machine to make sure

PR #274, opened 2016-10-14
Bf/kms config

ploki:


DavidPineauScality:
@ironman-machine Please validate that without setting this option, nothing's broken by the config :)

PR #275, opened 2016-10-14
Make routes for healthcheck with tests

dora-korpar:
This introduces routes for simple healthchecks that will be implemented

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -35,6 +35,29 @@ function checkUnsuportedRoutes(req, res, log) {
     return routesUtils.responseXMLBody(errors.MethodNotAllowed, null, res, log);
 }
 
+// current function utility is minimal, but will be expanded
+export function isHealthy() {
+    return true;
+}
+
+export function setHealthCheckResponse(req, res, log) {
+    if (isHealthy()) {
+        if (req.method === 'GET' || req.method === 'POST') {
+            res.writeHead(200, { 'Content-Type': 'application/json' });
+            res.write('{}');
+            log.debug('healthcheck response',
+                { httpMethod: req.method, httpCode: res.statusCode });
+        } else {
+            res.writeHead(400, { 'Content-Type': 'application/json' });
+            res.write('{}');
+        }
+    } else {
+        res.writeHead(500, { 'Content-Type': 'application/json' });
+        res.write('{}');
+    }
+    return res.statusCode;
+}
+
 export default function routes(req, res, logger) {
     const clientInfo = {
         clientIP: req.socket.remoteAddress,
@@ -47,6 +70,10 @@ export default function routes(req, res, logger) {
     log.info('received request', clientInfo);
 
     log.end().addDefaultFields(clientInfo);
+
+    if (req.url === '/_/healthcheck') {
+        return setHealthCheckResponse(req, res, log);
+    }
 
     try {
         utils.normalizeRequest(req);
--- a/tests/functional/healthchecks/master.json
+++ b/tests/functional/healthchecks/master.json
@@ -1,0 +1,6 @@
+{
+    "tests": {
+        "files": [ "/test" ],
+        "on": "aggressor"
+    }
+}
--- a/tests/functional/healthchecks/test/checkRoutes.js
+++ b/tests/functional/healthchecks/test/checkRoutes.js
@@ -1,0 +1,53 @@
+const assert = require('assert');
+const http = require('http');
+const https = require('https');
+const conf = require('../../config.json');
+
+const transportStr = conf.transport;
+const transport = transportStr === 'http' ? http : https;
+const options = {
+    host: conf.ipAddress,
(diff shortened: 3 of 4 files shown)


dora-korpar:
@ironman-machine go

tests/functional/healthchecks/master.json
```diff
@@ -0,0 +1,6 @@
+{
+    "tests": {
+        "files": [ "checkRoutes.js" ],
```
LaurenSpiegel: I assume we will have more files than this so probably makes sense to add a test directory, put checkRoutes.js in it and reference the test directory here.

tests/functional/healthchecks/master.json
```diff
@@ -0,0 +1,6 @@
...
+        "on": "aggressor"
+    }
+}
```
rahulreddy: Can you add a new line here?

lib/routes.js
```diff
@@ -35,6 +35,29 @@ function checkUnsuportedRoutes(req, res, log) {
...
+
+export function setHealthCheckResponse(req, res, log) {
+    if (isHealthy()) {
+        if (req.method === 'GET' || req.method === 'POST') {
+            res.writeHead(200, { 'Content-Type': 'application/json' });
+            res.write('{}');
+            log.debug('healthcheck response',
+                { httpMethod: req.method, httpCode: res.statusCode });
+        } else {
+            res.writeHead(400, { 'Content-Type': 'application/json' });
```
rahulreddy: I think this should be a 404 as 400 would indicate that a route exists but the request payload is incorrect.

rahulreddy: Ok, I am not firm on that :)

dora-korpar:
@ironman-machine let's go

lib/routes.js
dora-korpar: @rahulreddy Added the extra line to master.json! And it looks like we're keeping the 400

rahulreddy:
Ah yes, I forgot about the https options. The tests on end to end failed because it needs the self-signed cert stuff. You can look at the usage here https://github.com/scality/S3/blob/c162d00a07f887b91d2c4ca5b68d69094237e238/tests/functional/aws-node-sdk/test/support/config.js#L11

DavidPineauScality:
Same comment as the Vault PR for healthCheck:
I see that you are providing a few things that must have been decided/designed somehow (health check, error codes, etc.) Is there any way we can have access to these design/decisions, so that we can take them into account in order to review your PR ? Otherwise, we could be missing important meaning to understand this PR's aim, and properly review it.

tests/functional/healthchecks/test/checkRoutes.js
```diff
@@ -0,0 +1,54 @@
...
+
+function makeChecker(expectedStatus, done) {
+    return res => {
+        checkResult(expectedStatus, res);
+        done();
+    };
+}
+
+describe('Healthcheck routes', () => {
+    it('should return 200 OK on GET request', done => {
```
michael-zapata: Nit: all the `it`s look awfully similar, could we avoid duplication ?

dora-korpar:
@ironman-machine work work

tests/functional/healthchecks/test/checkRoutes.js
dora-korpar: Just for clarification, do you mean the it statements themselves or the code in them?

DavidPineauScality:
For information, relatively to the issue of the healthCheck tests not passing in end-to-end; Quick problem analysis:
- And end-to-end test was launched from here and passed, but the logs show no sign of running the healthCheck tests.
- An update of this PR happened between the moment the end-to-end tests were run and the PR was merged
- Thus: no end-to-end were run with the last update of this PR

Since the first attempt at end-to-end testing succeeded, and the output was not checked, I'm guessing that the update was trusted by default (whereas it should not have been), and we ended up merging something that broke the end-to-end tests.

Please remember that updating a PR before merging it (specifically on S3) means that we are not testing properly the new code and tests in an end-to-end scenario. This is even more true if we're adding functional tests in S3.

I know that this process is suboptimal, and not always obvious. That's exactly why we're currently working on improvements of our infrastructure and bot, to make it a full fledged merge bot, that will automate such steps.

rahulreddy:
Not sure what happened yesterday for the tests to fail after merging but I am able to see the in the output that healthcheck tests were run and succeeded. The reason you might not have see anything in the artifacts is because the log for healthcheck is debug and not info. We should change that to info for visibility. And I am assuming the change after the tests are run is squashing of the commits.

DavidPineauScality:
I was talking about the test's output (mocha), not about the logs from S3 :)

We should have seen the healthCheck tests there, but it wasn't originally. Thus I guessed that they were somehow not run (bad compliance of the tests with Integration's way of running them, probably?), and ended up being run post-merge, since the PR was updated between the moment the succeeded tests were ran and the merge.

PR #276, opened 2016-10-15
CLEANUP: Remove xml module usage in bucketGetAcl api

### LINKED ISSUE #269: Remove xml module usage in bucketGetAcl api
Instead of using xml dependency, just use string concatenation for building XML.

bennettbuchanan:
Fix #269
* Add functions in aclUtils.js to generate the XML DOM

### DIFF
--- a/lib/utilities/aclUtils.js
+++ b/lib/utilities/aclUtils.js
@@ -2,6 +2,7 @@ import { parseString } from 'xml2js';
 
 import { errors } from 'arsenal';
 import constants from '../../constants';
+import escapeForXML from './escapeForXML';
 
 const aclUtils = {};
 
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@ironman-machine go

lib/utilities/aclUtils.js
```diff
@@ -233,4 +234,72 @@ aclUtils.sortHeaderGrants =
...
+ */
+aclUtils.createXmlDom = obj => {
+    if (!Array.isArray(obj)) {
+        const key = Object.keys(obj)[0];
+        aclUtils.createEl(`<${key}>`, obj[key], `</${key}>`);
+    } else {
+        obj.forEach(item => {
+            const key = Object.keys(item)[0];
+
+            if (key === 'Grantee') {
```
alexandre-merle: why creating a specific case where we can handle attributes for other keys ?

lib/utilities/aclUtils.js
```diff
@@ -233,4 +234,72 @@ aclUtils.sortHeaderGrants =
...
+                const attrArr = item[key][0][Object.keys(item[key][0])];
+                const attr = [];
+
+                Object.keys(attrArr).forEach(item => {
+                    attr.push(` ${item}="${attrArr[item]}"`);
+                });
+
+                aclUtils.createEl(`<${key + attr.join('')}>`, item[key],
+                                  `</${key}>`);
+            } else if (key !== '_attr') {
```
alexandre-merle: what about the _attr key ? we do nothing on this case ?

bennettbuchanan:
We have decided that we do not want to build an object with `aclUtils.constructGetACLsJson` to then use it to build the XML DOM. Thus we are going in a different direction, which will be demonstrated in a new pull request referencing this issue.

PR #280, opened 2016-10-17
COMPAT: check for empty content-md5 header

nicolas2bert:
Fixes https://github.com/scality/S3/issues/281

When content-md5 header is empty: return InvalidDigest error
**On [ceph/s3-tests](https://github.com/ceph/s3-tests), these changes fix**:

- `FAIL: s3tests.functional.test_headers.test_object_create_bad_md5_empty` **should return InvalidDigest error**
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_md5_bad`  **should return BadDigest error**
when `content-md5` header mismatches the hash which has been generated.

### DIFF
--- a/lib/routes/routePUT.js
+++ b/lib/routes/routePUT.js
@@ -120,6 +120,14 @@ export default function routePUT(request, response, log, utapi) {
         // PUT object copy
         // if content-md5 is not present in the headers, try to
         // parse content-md5 from meta headers
+
+        if (request.headers['content-md5'] === '') {
+            log.debug('empty content-md5 header', {
+                method: 'routePUT',
+            });
+            return routesUtils
+            .responseNoBody(errors.InvalidDigest, null, response, 200, log);
+        }
         if (request.headers['content-md5']) {
             request.contentMD5 = request.headers['content-md5'];
         } else {
--- a/lib/services.js
+++ b/lib/services.js
@@ -179,16 +179,49 @@ export default {
                              { dataRetrievalInfo });
                      hashedStream.on('hashed', () => {
                          log.trace('hashed event emitted');
+                         const checkHashMatchMD5Res = this.checkHashMatchMD5(
+                           request.contentMD5, hashedStream.completedHash, log);
+                         if (!checkHashMatchMD5Res) {
+                             log.trace('contentMD5 does not match, deleting' +
+                             ' data.');
+                             data.batchDelete(dataRetrievalInfo, log);
+                             return cb(errors.BadDigest);
+                         }
                          return cb(null, objectMetadata, dataRetrievalInfo,
                              hashedStream.completedHash);
                      });
                      if (hashedStream.completedHash) {
+                         const checkHashMatchMD5Res = this.checkHashMatchMD5(
+                           request.contentMD5, hashedStream.completedHash, log);
+                         if (!checkHashMatchMD5Res) {
+                             log.trace('contentMD5 does not match, deleting' +
+                             ' data.');
+                             data.batchDelete(dataRetrievalInfo, log);
+                             return cb(errors.BadDigest);
+                         }
                          hashedStream.removeAllListeners('hashed');
                          return cb(null, objectMetadata, dataRetrievalInfo,
                              hashedStream.completedHash);
                      }
                      return undefined;
                  });
+    },
+
+   /**
+    * Check that hashedStream.completedHash matches header contentMd5.
+    * @param {object} contentMD5 - content-md5 header
+    * @param {string} completedHash - hashed stream once completed
+    * @param {RequestLogger} log - the current request logger
+    * @return {boolean} - true if contentMD5 matches or is undefined,
+    * false otherwise
+    */
+    checkHashMatchMD5(contentMD5, completedHash, log) {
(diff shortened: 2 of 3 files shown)


lib/services.js
```diff
@@ -191,6 +201,29 @@ export default {
...
+   /**
+    * Check that hashedStream.completedHash matches header contentMd5.
+    * @param {object} request - request linked to the stream containine the data
+    * @param {string} hashedStream - hased stream once completed
+    * @param {object} dataRetrievalInfo - dataStore: backend stored key
+    * {key, dataStoreName (mem/file)}
+    * @param {RequestLogger} log - the current request logger
+    * @return {object} errors
+    */
+    checkHashMatchMD5(request, hashedStream, dataRetrievalInfo, log) {
```
alexandre-merle: I may be wrong about this, but its not related to the issue

lib/routes/routePUT.js
```diff
@@ -120,6 +120,14 @@ export default function routePUT(request, response, log, utapi) {
         // PUT object copy
         // if content-md5 is not present in the headers, try to
         // parse content-md5 from meta headers
+
+        if (request.headers['content-md5'] === '') {
```
alexandre-merle: its apply only if content-md5 is set and its an empty string, so i guess its ok. But do we have other case where we should return InvalidDigest ? like a malformed md5 ?

nicolas2bert: Yes, sorry @alexandre-merle, we talked about that we Lauren but I did not update the status.
@alexandre-merle you are right, there are more cases like: 
- unreadable MD5 `'Content-MD5': '\x07'`
- invalid MD5 `'Content-MD5':'YWJyYWNhZGFicmE='`

lib/services.js
nicolas2bert: I updated the issue, thanks

lib/services.js
```diff
@@ -191,6 +201,29 @@ export default {
...
+        if (request.contentMD5 && hashedStream.completedHash &&
+        request.contentMD5 !== hashedStream.completedHash) {
+            // delete data if does not match.
+            data.batchDelete(dataRetrievalInfo, log);
```
rahulreddy: This should go in the method `data.put` where `checkHashMatchMD5` is being called for separation of concerns.

lib/services.js
```diff
@@ -179,10 +179,26 @@ export default {
...
+                             return cb(checkHashMatchMD5Res);
+                         }
                          return cb(null, objectMetadata, dataRetrievalInfo,
                              hashedStream.completedHash);
                      });
                      if (hashedStream.completedHash) {
+                         const checkHashMatchMD5Res = this.checkHashMatchMD5(
+                            request, hashedStream, log);
+                         if (checkHashMatchMD5Res instanceof Error) {
+                             log.trace('delete data if checkHashMatchMD5Res ' +
```
nicolas2bert: thanks done

lib/services.js
```diff
@@ -191,6 +207,26 @@ export default {
...
+   /**
+    * Check that hashedStream.completedHash matches header contentMd5.
+    * @param {object} request - request linked to the stream containine the data
+    * @param {string} hashedStream - hased stream once completed
+    * {key, dataStoreName (mem/file)}
+    * @param {RequestLogger} log - the current request logger
+    * @return {object} errors
+    */
+    checkHashMatchMD5(request, hashedStream, log) {
+        if (request.contentMD5 && hashedStream.completedHash &&
```
LaurenSpiegel: why would we not have a request.contentMD5 or hashedStream.completedHash at this point?
1) I believe we are checking request[content-md5] and setting the value to request.contentMD5 in the route so we should have it?
2) if we don't have a hashedStream.completedHash at this point we should return an errors.InternalError.

lib/services.js
```diff
@@ -191,6 +207,26 @@ export default {
...
+        request.contentMD5 !== hashedStream.completedHash) {
+            log.debug('contentMD5 and completedHash does not match',
+            { method: 'checkHashMatchMD5',
+            completedHash: hashedStream.completedHash,
+            contentMD5: request.contentMD5 });
+            return errors.BadDigest;
+        }
+        return {};
```
LaurenSpiegel: i think it would make sense for this function to return true if there is a match and false if not.  Then you can set the errors.BadDigest if you get false.

lib/services.js
LaurenSpiegel: 3) I would send into this function the request.contentMD5 as an argument and the hashedStream.completedHash as an argument (it just needs these properties).  I would check to make sure hashedStream.completedHash exists before calling this function.

lib/services.js
nicolas2bert: I did that in case we would have to add more checks and so return more errors.

lib/services.js
```diff
@@ -191,6 +207,23 @@ export default {
                  });
     },
 
+   /**
+    * Check that hashedStream.completedHash matches header contentMd5.
+    * @param {object} contentMD5 - content-md5 header
+    * @param {string} completedHash - hased stream once completed
```
LaurenSpiegel: typo in 'hased'

nicolas2bert:
@ironman-machine please

PR #283, opened 2016-10-17
BF: Store / return AWS compatible headers

### LINKED ISSUE #282: Store / return optional headers for compatibility with AWS S3 REST API
Cache-Control, Content-Disposition, Content-Encoding, and Expires are optional headers that should be returned in a objectGet and objectHead response if the user provides them when creating or copying an object. These are currently omitted in responses.

Refer to:
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html

electrachong:
Related to issue fix #282
Optional headers Cache-Control, Content-Disposition, Content-Encoding, and Expires are missing in objectGet and objectHead response.
* Stores header values if specified in objectPut or objectCopy request.
* Returns stored values in objectGet and objectHead responses.
* Will follow-up in separate PR to address storing header values for multipartUpload

### DIFF
--- a/lib/services.js
+++ b/lib/services.js
@@ -218,6 +218,22 @@ export default {
             authInfo.getAccountDisplayName();
         // This should be object creator's canonical ID.
         omVal['owner-id'] = authInfo.getCanonicalID();
+        // TODO: Find a way to store these headers for multipart uploads.
+        // params.headers is undefined when called from
+        // completeMultipartUploads because the latter does not pass
+        // headers as part of the params object
+        if (params.headers) {
+            omVal['cache-control'] = params.headers['cache-control'];
+            omVal['content-disposition'] =
+                params.headers['content-disposition'];
+            if (params.headers['content-encoding']) {
+                omVal['content-encoding'] =
+                    params.headers['content-encoding']
+                    .replace(/aws-chunked,?/, '');
+            }
+            omVal.expires = params.headers.expires;
+        }
+
         omVal['content-length'] = size;
         omVal['content-type'] = contentType;
         // Sending in last modified date in object put copy since need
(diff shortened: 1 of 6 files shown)


electrachong:
@ironman-machine go for it

lib/services.js
```diff
@@ -218,6 +218,19 @@ export default {
             authInfo.getAccountDisplayName();
         // This should be object creator's canonical ID.
         omVal['owner-id'] = authInfo.getCanonicalID();
+        // TODO: Find a way to store these headers for multipart uploads.
+        // params.headers is undefined when called from
+        // completeMultipartUploads because the latter does not pass
+        // headers as part of the params object
```
LaurenSpiegel: We will probably decide to not use the headers to store this but rather send in the values directly as params.  For now, however, this is good to add the tests and get us compliant and I will help you with the follow up that will include MPU.  Note that for content-encoding we need to modify the value to remove 'aws-chunked' if there.  In streaming v4 auth, this value is sent in but AWS says they do not store it along with any other content encoding value.  So, please check whether there is 'aws-chunked' in the value and remove it.  See http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html and search for "content-encoding".

electrachong: I revised so that if 'aws-chunked' is sent as part of the Content-Encoding header it's removed. Let me know if it looks OK. To make sure, we don't use this value for our own logic in S3, correct?

LaurenSpiegel: correct. and should not be sent in the response of a get/head.

electrachong:
@ironman-machine go :)

rahulreddy:
@ironman-machine looks like the last build errored out not because of this PR

PR #284, opened 2016-10-18
Temporarily skip healthcheck tests

RachedBenMustapha:
Will be re-enabled as soon as the tests are updated.


vrancurel:
Merging because we are blocked to make progress. Apparently the fact that we pushed new images this morning Paris time caused the tests not to work any more.

DavidPineauScality:
Explanation of the issue there: https://github.com/scality/S3/pull/275

PR #286, opened 2016-10-18
Fix for end-to-end failure

dora-korpar:
This commit is to fix the e2e failure that came up yesterday post merge

### DIFF
--- a/tests/functional/healthchecks/test/checkRoutes.js
+++ b/tests/functional/healthchecks/test/checkRoutes.js
@@ -2,6 +2,7 @@ const assert = require('assert');
 const http = require('http');
 const https = require('https');
 const conf = require('../../config.json');
+const fs = require('fs');
 
 const transportStr = conf.transport;
 const transport = transportStr === 'http' ? http : https;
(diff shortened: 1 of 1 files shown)


dora-korpar:
@ironman-machine gogogo

dora-korpar:
@ironman-machine let's be sure

rahulreddy:
The tests are also skipped at the moment, they need to be re-enabled with this PR.

DavidPineauScality:
Indeed, aggreed with @rahulreddy; the end to end test run has no meaning unless we unskip the health check tests here (since they were what actually failed).
As such, we'll need a new run after they are unskipped, to confirm whether the tests were properly fixed.

Also, I'll support @alexandre-merle 's comment, as he was one of the few that really did a huge work on the https for it to work properly in testing; functional as well as end-to-end for all projects.

dora-korpar:
@ironman-machine get it

dora-korpar:
@ironman-machine cmon Jarvis

dora-korpar:
@ironman-machine you da man, Jarvis

alexandre-merle:
Thanks :)

tests/functional/healthchecks/test/checkRoutes.js
```diff
@@ -2,6 +2,7 @@ const assert = require('assert');
 const http = require('http');
 const https = require('https');
 const conf = require('../../config.json');
```
rahulreddy: `Config.js` need to be used instead of  using `config.json` directly. This would enable the tests to pickup any config made available through  the env `S3_CONFIG_FILE`

PR #287, opened 2016-10-18
CLEANUP: Remove xml module usage in objectGetAcl api

### LINKED ISSUE #285: Remove XML module usage in objectGetACl api

bennettbuchanan:
Fix #285
* Generate the XML DOM with a function instead of constructing
  a JSON object with `aclUtils.constructGetACLsJson` and then
  using the xml module


bennettbuchanan:
@ironman-machine go

bennettbuchanan:
This function can apply across both `objectGetACL` and `bucketGetACL` APIs. Thus it will be moved to aclUtils.js. The change will be opened in a new pull request.

PR #289, opened 2016-10-18
CLEANUP: Remove xml module usage in objectGetAcl api

### LINKED ISSUE #285: Remove XML module usage in objectGetACl api

bennettbuchanan:
Fix #285
* Generate the XML DOM with the function `convertToXml()` instead
  of constructing a JSON object with `aclUtils.constructGetACLsJson`
  and then using the xml module

### DIFF
--- a/lib/utilities/aclUtils.js
+++ b/lib/utilities/aclUtils.js
@@ -2,6 +2,7 @@ import { parseString } from 'xml2js';
 
 import { errors } from 'arsenal';
 import constants from '../../constants';
+import escapeForXML from '../utilities/escapeForXML';
 
 const aclUtils = {};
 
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@ironman-machine go

lib/utilities/aclUtils.js
```diff
@@ -233,4 +234,72 @@ aclUtils.sortHeaderGrants =
...
+ * @param {object} grantInfo - The `grantInfo` object defined in
+ * `objectGetACL()`
+ * @return {string} xml.join('') - The XML DOM string
+ */
+aclUtils.convertToXml = grantInfo => {
+    const { grants, ownerInfo } = grantInfo;
+    const xml = [];
+
+    // Escape any special XML characters
+    for (const key in grants) {
```
rahulreddy: You should use `Object.keys(grants).forEach` over `for(var x in )..` for consistency and potential performance issues https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#5-for-in

lib/utilities/aclUtils.js
```diff
@@ -233,4 +234,72 @@ aclUtils.sortHeaderGrants =
...
+        if (grants.hasOwnProperty(key)) {
+            grants[key] = escapeForXML(grants[key]);
+        }
+    }
+    for (const key in ownerInfo) {
```
rahulreddy: Same here. Prefer usage of Object.keys(grants).forEach over for(var x in )...

lib/utilities/aclUtils.js
```diff
@@ -233,4 +234,72 @@ aclUtils.sortHeaderGrants =
...
+    xml.push('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>',
+             '<AccessControlPolicy>',
+             '<Owner>',
+             `<ID>${ownerInfo.ID}</ID>`,
+             `<DisplayName>${ownerInfo.displayName}</DisplayName>`,
+             '</Owner>',
+             '<AccessControlList>'
+    );
+
+    // For each object in `grants`, push the relevant XML elements to `xml`
```
rahulreddy: This comment is redundant. Ideally, comments should explain `why` of the code than the `what`.

lib/utilities/aclUtils.js
bennettbuchanan: Addressed in [this commit](https://github.com/scality/S3/pull/289/commits/14c48295550a9b8598514a5d5440efdbe17421d6).

lib/utilities/aclUtils.js
bennettbuchanan: Addressed in [this commit](https://github.com/scality/S3/pull/289/commits/14c48295550a9b8598514a5d5440efdbe17421d6).

bennettbuchanan:
@ironman-machine go

lib/utilities/aclUtils.js
```diff
@@ -233,4 +234,58 @@ aclUtils.sortHeaderGrants =
...
+ */
+aclUtils.convertToXml = grantInfo => {
+    const { grants, ownerInfo } = grantInfo;
+    const xml = [];
+
+    xml.push('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>',
+             '<AccessControlPolicy>',
+             '<Owner>',
+             `<ID>${ownerInfo.ID}</ID>`,
+             `<DisplayName>${ownerInfo.displayName}</DisplayName>`,
```
LaurenSpiegel: why not escaped?

lib/utilities/aclUtils.js
```diff
@@ -233,4 +234,58 @@ aclUtils.sortHeaderGrants =
...
+            );
+        } else if (grant.URI) {
+            xml.push('<Grantee xmlns:xsi="http://www.w3.org/2001/' +
+                        'XMLSchema-instance" xsi:type="Group">',
+                     `<URI>${escapeForXML(grant.URI)}</URI>`
+            );
+        }
+
+        if (grant.displayName) {
+            xml.push(`<DisplayName>${grant.displayName}</DisplayName>`);
```
LaurenSpiegel: why not escaped?

lib/utilities/aclUtils.js
bennettbuchanan: I was incorrect here. Display names should indeed be able to handle special characters. I will update this.

lib/utilities/aclUtils.js
rahulreddy: Good call. Vault's display names follow the regex `/^[\w+=,.@ -]+$/` and `@` should be escaped in XML.

lib/utilities/aclUtils.js
bennettbuchanan: In S3/conf/authdata.json I had changed the account name from "Bart" to something like "Bart<Simpson". This change did not throw any error and the special character "<" was escaped accordingly. This is why I figured `displayName` needed escaping. 

[`escapeForXML()`](https://github.com/scality/S3/blob/e09c9d666f958e6bef3eb257f084acaed071ca3d/lib/utilities/escapeForXML.js#L5) will not escape "@", but that seems to be okay for XML as per [W3C specs](https://www.w3.org/TR/xml11/#sec-predefined-ent).

rahulreddy: My bad, I misread the escape list.

bennettbuchanan:
@ironman-machine

PR #296, opened 2016-10-20
CLEANUP: Add range tests with large end position

### LINKED ISSUE #270: [test case] GET RANGE with large end end position
customer request:

NOTE: This does not work with RS2 and needs to work with S3. 

Specific test starts by uploading a 2890 byte object.  Then, retrieve object with a RANGE specified.  Each of these should return the last 91 bytes

- range (2800-) 
- range (2800-INTEGER.max)
- range (2800-LONG.max)

With RS2, the LONG.max causes the RS2 connector to return the first 91 bytes of the object, not the last.  This is wrong.

bennettbuchanan:
Fix #270

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/getRange.js
+++ b/tests/functional/aws-node-sdk/test/object/getRange.js
@@ -1,0 +1,72 @@
+import assert from 'assert';
+import withV4 from '../support/withV4';
+import BucketUtility from '../../lib/utility/bucket-util';
+
+const bucketName = 'bucket-for-range-test';
+const objName = 'largerput';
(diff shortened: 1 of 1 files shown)


bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

LaurenSpiegel:
End to end is failing due to parsing of an s3curl test unrelated to this PR.  We are fixing this s3curl test, but that should not get in the way of merging this.

PR #297, opened 2016-10-20
BF: MPU stores/returns AWS-compat headers

electrachong:
Follow-up to PR #283, issue #282
Modify completeMultipartUpload so that Cache-Control, Content-Disposition, Content-Encoding, and Expires are stored in MD from shadow bucket for a get response. Additionally, refactor so that these headers are referenced more cleanly (by individual properties in the params object) instead of depending upon the whole request.headers object being passed as part of the metadataStoreParams / params object in metadataStoreObject.


electrachong:
@ironman-machine let's try

electrachong:
@ironman-machine let's try again

electrachong:
@ironman-machine

lib/api/objectPut.js
```diff
@@ -45,10 +44,19 @@ function _storeIt(bucketName, objectKey, objMD, authInfo, canonicalID,
...
         size,
-        contentType,
+        contentType: request.headers['content-type'],
+        cacheControl: request.headers['cache-control'],
+        contentDisposition: request.headers['content-disposition'],
+        expires: request.headers.expires,
         headers,
         log,
     };
+    // Remove V4 streaming value 'aws-chunked' if present in content-encoding to
```
LaurenSpiegel: I would extract this into a utility function since it is used 3 times.  Also, it would probably be helpful to put this link in the comments so we have the reference: http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

lib/api/objectCopy.js
```diff
@@ -55,6 +55,12 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
     const userMetadata = whichMetadata === 'COPY' ?
         utils.getMetaHeaders(sourceObjMD) :
         utils.getMetaHeaders(headers);
+
+    // If COPY, pull AWS-compatible headers from source object
+    // Otherwise, pull AWS-compative headers from request headers
```
LaurenSpiegel: typo here in compatible.  In any case, instead of compatHeadersSource I would call these headersToStoreSource.  Everything we do is compatible 😄

lib/api/objectPut.js
electrachong: OK. Is it all right that I put the comment and reference on the utility function definition and removed it elsewhere, or should we keep the comments when calling the method?

LaurenSpiegel: Having the comment in one place is good.  However, could you put the utility function in this file:
https://github.com/scality/S3/tree/master/lib/api/apiUtils/object

lib/api/apiUtils/object/removeAWSChunked.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Checks for V4 streaming value 'aws-chunked' and removes it if present in
+ * Content-Encoding to be compatible with AWS behavior. See:
+ * http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html
+ * @param {string} sourceHeader - Content-Encoding header from request headers
+ * @return {string} new value with 'aws-chunked,' substrings removed
```
bennettbuchanan: This seems to only match the first substring 'aws-chunked,'. 

It sounds like the intention here is to match all occurrences of the substring because in the function documentation of the return value 'substrings' is plural. Maybe it is actually only necessary to match the first instance of 'aws-chunked,' though. Otherwise, to match all substrings perhaps you could use `sourceHeader.replace(/aws-chunked,/g, '');`

electrachong:
squashed commits, running @ironman-machine one last time

rahulreddy:
@ironman-machine again

PR #299, opened 2016-10-20
COMPAT: create bucket with empty name

### LINKED ISSUE #298: COMPAT: create bucket with empty name: errors incompatible with AWS
When creating bucket with empty name: should return `405 MethodNotAllowed`

On **ceph/s3-tests**, these changes fix:
`ERROR: s3tests.functional.test_s3.test_bucket_create_naming_bad_short_empty`

nicolas2bert:
FIXES #298

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -104,6 +104,13 @@ export default function routes(req, res, logger) {
         bucketName: req.bucketName,
         objectKey: req.objectKey,
     });
+    // if empty name and request not a list Buckets
+    if (!req.bucketName &&
+      !(req.method.toUpperCase() === 'GET' && !req.objectKey)) {
+        log.warn('empty bucket name', { method: 'routes' });
+        return routesUtils.responseXMLBody(errors.MethodNotAllowed,
+            undefined, res, log);
+    }
 
     if (req.bucketName !== undefined &&
         utils.isValidBucketName(req.bucketName) === false) {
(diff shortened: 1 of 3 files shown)


lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -136,7 +136,10 @@ export function cleanUpBucket(bucketMD, canonicalID, log, callback) {
 export function createBucket(authInfo, bucketName, headers,
     locationConstraint, log, cb) {
     log.trace('Creating bucket');
-    assert.strictEqual(typeof bucketName, 'string');
+    if (typeof bucketName !== 'string') {
```
rahulreddy: That assertion was for type assertions of function parameters (I was not a fan of that). Anyways, shouldn't you be checking `if(bucketName === '')`?

nicolas2bert: If bucketName is empty (not undefined) it will jumb here: https://github.com/scality/S3/blob/b246509b83f45529131f3abbc1b961ec85feb123/lib/routes.js#L91

nicolas2bert:
Thanks @rahulreddy, I also had to check that the request was not` listBuckets` because this one is the only one which allows undefined bucketName.

nicolas2bert:
@ironman-machine please

LaurenSpiegel:
end to end passed here: http://ci.ironmann.io/gh/scality/Integration/5821

PR #301, opened 2016-10-21
Ft/batch delete

LaurenSpiegel:
This PR implements the multi-object delete API.

This closes https://github.com/scality/S3/issues/83

This depends on:
https://github.com/scality/utapi/pull/64
https://github.com/scality/Arsenal/pull/184
https://github.com/scality/vaultclient/pull/106
https://github.com/scality/Vault/pull/612
https://github.com/scality/Integration/pull/284

End to end passed here: http://ci.ironmann.io/gh/scality/Integration/5586

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -1,0 +1,351 @@
+import crypto from 'crypto';
+
+import async from 'async';
+import { auth, errors } from 'arsenal';
+import { parseString } from 'xml2js';
+
+import escapeForXML from '../utilities/escapeForXML';
+import bucketShield from './apiUtils/bucket/bucketShield';
+import metadata from '../metadata/wrapper';
+import services from '../services';
+import vault from '../auth/vault';
+import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
+
+
+/*
+   Format of xml request:
+   <Delete>
+       <Quiet>true</Quiet>
+       <Object>
+            <Key>Key</Key>
+            <VersionId>VersionId</VersionId>
+       </Object>
+       <Object>
+            <Key>Key</Key>
+       </Object>
+       ...
+   </Delete>
+   */
+
+
+  /*
+  Format of xml response:
+  <?xml version="1.0" encoding="UTF-8"?>
+  <DeleteResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+    <Deleted>
+      <Key>sample1.txt</Key>
+    </Deleted>
+    <Error>
+      <Key>sample2.txt</Key>
+      <Code>AccessDenied</Code>
+      <Message>Access Denied</Message>
+    </Error>
+  </DeleteResult>
+   */
+
+/**
+* formats xml for response
+* @param {boolean} quietSetting - true if xml should just include error list
+* and false if should include deleted list and error list
+* @param {object []} errorResults - list of error result objects with each
+* object containing -- key: objectName, error: arsenal error
+* @param {string []} deleted - list of object keys deleted
+* @return {string} xml string
+*/
+function _formatXML(quietSetting, errorResults, deleted) {
+    let errorXML = [];
+    errorResults.forEach(errorObj => {
+        errorXML.push(
+        '<Error>',
+        '<Key>', escapeForXML(errorObj.key), '</Key>',
+        '<Code>', errorObj.error.message, '</Code>',
+        '<Message>', errorObj.error.description, '</Message>',
+        '</Error>'
+        );
+    });
+    errorXML = errorXML.join('');
+    const xml = [
+        '<DeleteResult ',
+        'xmlns="http://s3.amazonaws.com/doc/2006-03-01/">',
+        // placeholder in array for deleted list if verbose response
+        '',
+        errorXML,
+        '</DeleteResult>',
+    ];
+    if (quietSetting) {
+        // just return errors
+        return xml.join('');
+    }
+    const deletedXML = [];
+    deleted.forEach(objKey => {
+        deletedXML.push(
+            '<Deleted>',
+            '<Key>', escapeForXML(objKey), '</Key>',
+            '</Deleted>'
+        );
+    });
+    xml[2] = deletedXML.join('');
+    return xml.join('');
+}
+
+function _parseXml(xmlToParse, next) {
+    return parseString(xmlToParse, (err, result) => {
+        if (err || !result || !result.Delete) {
+            return next(errors.MalformedXML);
+        }
+        const json = result.Delete;
+        // not quiet is the default if nothing specified
+        const quietSetting = json.Quiet && json.Quiet[0] === 'true';
+        // format of json is {"Object":[{"Key":["test1"]},{"Key":["test2"]}]}
+        const objects = json.Object.map(item => item.Key[0]);
+        return next(null, quietSetting, objects);
+    });
+}
+
+/**
+* gets object metadata and deletes object
+* @param {string} bucketName - bucketName
+* @param {boolean} quietSetting - true if xml should just include error list
+* and false if should include deleted list and error list
+* @param {object []} errorResults - list of error result objects with each
+* object containing -- key: objectName, error: arsenal error
+* @param {string []} inPlay - list of object keys still in play
+* @param {object} log - logger object
+* @param {function} next - callback to next step in waterfall
+* @return {undefined}
+* @callback called with (err, quietSetting, errorResults, numOfObjects,
+* successfullyDeleted, totalContentLengthDeleted)
+*/
+export function getObjMetadataAndDelete(bucketName, quietSetting,
+    errorResults, inPlay, log, next) {
+    const successfullyDeleted = [];
+    let totalContentLengthDeleted = 0;
+    let numOfObjects = 0;
+    // for obj deletes, no need to check acl's at object level
+    // (authority is at the bucket level for obj deletes)
+
+    // doing 5 requests at a time. note that the data wrapper
+    // will do 5 parallel requests to data backend to delete parts
+    return async.forEachLimit(inPlay, 5, (key, moveOn) => {
+        metadata.getObjectMD(bucketName, key, log, (err, objMD) => {
+            // if general error from metadata return error
+            if (err && !err.NoSuchKey) {
+                log.error('error getting object MD', { error: err, key });
+                errorResults.push({
+                    key,
+                    error: err,
+                });
+                return moveOn();
+            }
+            // if particular key does not exist, AWS returns success
+            // for key so add to successfullyDeleted list and move on
+            if (err && err.NoSuchKey) {
+                successfullyDeleted.push(key);
+                return moveOn();
+            }
+            return services.deleteObject(bucketName, objMD, key, log,
+                err => {
+                    if (err) {
+                        log.error('error deleting object', { error: err, key });
+                        errorResults.push({
+                            key,
+                            error: err,
+                        });
+                        return moveOn();
+                    }
+                    if (objMD['content-length']) {
+                        totalContentLengthDeleted +=
+                            objMD['content-length'];
+                    }
+                    numOfObjects++;
+                    successfullyDeleted.push(key);
+                    return moveOn();
+                });
+        });
+    },
+    // end of forEach func
+    err => {
+        log.trace('finished deleting objects', { numOfObjects });
+        return next(err, quietSetting, errorResults, numOfObjects,
+            successfullyDeleted, totalContentLengthDeleted);
+    });
+}
+
+/**
+ * multiObjectDelete - Delete multiple objects
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http.IncomingMessage as modified by
+ * lib/utils.normalizeRequest and routes/routePOST.js
+ * @param {object} request.headers - request headers
+ * @param {object} request.query - query from request
+ * @param {string} request.post - concatenation of request body
+ * @param {string} request.bucketName - parsed bucketName
+ * @param {string} request.socket.remoteAddress - requester IP
+ * @param {boolean} request.connection.encrypted - whether request was encrypted
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+export default
+function multiObjectDelete(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'multiObjectDelete' });
+    if (!request.post) {
+        return callback(errors.MissingRequestBodyError);
+    }
+    const md5 = crypto.createHash('md5')
+        .update(request.post, 'utf8').digest('base64');
+    if (md5 !== request.headers['content-md5']) {
+        return callback(errors.BadDigest);
+    }
+
+    const bucketName = request.bucketName;
+    const canonicalID = authInfo.getCanonicalID();
+
+    return async.waterfall([
+        function parseXML(next) {
+            return _parseXml(request.post, (err, quietSetting, objects) => {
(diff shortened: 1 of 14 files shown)


tests/functional/aws-node-sdk/test/object/multiObjectDelete.js
```diff
@@ -0,0 +1,257 @@
...
+    it('should batch delete 1000 objects quietly', () => {
+        const objects = createObjectsList(1000);
+        return s3.deleteObjectsAsync({
+            Bucket: bucketName,
+            Delete: {
+                Objects: objects,
+                Quiet: true,
+            },
+        }).then(res => {
+            assert.deepStrictEqual(res.Deleted.length, 0);
```
GuillaumeGomez: Why not `strictEqual`?

LaurenSpiegel:
@ironman-machine checking in
SCALITY_INTEGRATION_BRANCH=ft/batchDelete
SCALITY_VAULT_BRANCH=ft/batchDelete
SCALITY_ARSENAL_BRANCH=ft/batchDelete
SCALITY_UTAPI_BRANCH=ft/batchDelete
SCALITY_VAULTCLIENT_BRANCH=ft/batchDelete

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+            // for obj deletes, no need to check acl's at object level
+            // (authority is at the bucket level for obj deletes)
+
+            // doing 5 requests at a time. note that services.deleteObject
+            // will do 5 parallel requests to data backend to delete parts
+            return async.forEachLimit(inPlay, 5, (key, moveOn) => {
+                metadata.getObjectMD(bucketName, key, log, (err, objMD) => {
+                    // if general error from metadata return error
+                    if (err && !err.NoSuchKey) {
+                        return next(err);
```
michael-zapata: This means that in case of an error during the process, we do not continue with the process at all, and do not send back which elements we deleted/ errored out, is that something desired?

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                    }
+                    // if particular key does not exist, AWS returns success
+                    // for key so add to deleted list and move on
+                    if (err && err.NoSuchKey) {
+                        deleted.push(key);
```
michael-zapata: Since `deleted` is only created here, wouldn't it make sense to have it properly scoped in here, and use it after that by sending it through the callback?

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                if (objects.length < 1 || objects.length > 1000) {
+                    return next(errors.MalformedXML);
+                }
+                return next(null, quietSetting, objects);
+            });
+        },
+        function checkPolicies(quietSetting, objects, next) {
+            // if request from account, no need to check policies
+            if (!authInfo.isRequesterAnIAMUser()) {
+                inPlay = objects.slice();
```
michael-zapata: Since `inPlay` is only created here, wouldn't it make sense to have it properly scoped in here, and use it afterwards by sending it through the callback? That would properly scope every variable, making it easier to extract the functions for the aforementioned purpose, while not allocating it in case of a malformed XML. The same goes for `errorResults`.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+* @param {boolean} quietSetting - true if xml should just include error list
+* and false if should include deleted list and error list
+* @param {object []} errorResults - list of error result objects with each
+* object containing -- key: objectName, error: arsenal error
+* @param {string []} deleted - list of object keys deleted
+* @return {string} xml string
+*/
+function _formatXML(quietSetting, errorResults, deleted) {
+    let errorXML = [];
+    errorResults.forEach(errorObj => {
```
michael-zapata: Instead of a `.forEach()` with a single `.push()` inside, you can use a `.map()`.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+        '',
+        errorXML,
+        '</DeleteResult>',
+    ];
+    if (quietSetting) {
+        // just return errors
+        return xml.join('');
+    }
+    const deletedXML = [];
+    deleted.forEach(objKey => {
```
michael-zapata: Same here regarding the `.map()` operation.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+    const deleted = [];
+    // track keys that are still on track to be deleted
+    let inPlay = [];
+    let totalContentLengthDeleted = 0;
+
+    return async.waterfall([
+        function parseXML(next) {
+            return _parseXml(request.post, (err, quietSetting, objects) => {
+                if (err) {
+                    return next(err);
```
michael-zapata: Shouldn't this be a case of `MalformedXML` as well? Because we are basically dealing with a parsing error here, and that would not make sense as an internal error.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+            });
+        },
+        function checkPolicies(quietSetting, objects, next) {
+            // if request from account, no need to check policies
+            if (!authInfo.isRequesterAnIAMUser()) {
+                inPlay = objects.slice();
+                return next(null, quietSetting);
+            }
+
+            // TODO: once arsenal's extractParams is separated from doAuth
```
michael-zapata: I trust that an issue will be opened once this is merged so that we can track it afterwards and not forget about it.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                },
+                parameterize: {
+                    specificResource: objects,
+                },
+            };
+            return vault.checkPolicies(requestContextParams, authInfo.getArn(),
+                log, (err, authorizationResults) => {
+                    // there were no policies so received a blanket AccessDenied
+                    if (err && err.AccessDenied) {
+                        objects.forEach(key => {
```
michael-zapata: Same here regarding `.map()`.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                                error: errors.AccessDenied });
+                        });
+                        return next(null, quietSetting);
+                    }
+                    if (err) {
+                        log.trace('error checking policies', { error: err });
+                        return next(err);
+                    }
+                    if (objects.length !== authorizationResults.length) {
+                        log.error('vault did not return correct number of ' +
```
michael-zapata: Can we log the diff here? That strikes me as an important info to have here.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                        'authorization results');
+                        return next(errors.InternalError);
+                    }
+                    for (let i = 0; i < authorizationResults.length; i++) {
+                        const result = authorizationResults[i];
+                        // result is { isAllowed: true,
+                        // arn: arn:aws:s3:::bucket/object} unless not allowed
+                        // in which case no isAllowed key will be present
+                        const slashIndex = result.arn.indexOf('/');
+                        if (slashIndex === undefined) {
```
michael-zapata: `indexOf()` returns -1 in case it doesn't find the string, not `undefined`. The fact that this is not picked up by the tests means that this case is not handled in unit testing.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                // check whether bucket has transient or deleted flag
+                if (bucketShield(bucketMD, 'objectDelete')) {
+                    return next(errors.NoSuchBucket);
+                }
+                if (!isBucketAuthorized(bucketMD, 'objectDelete',
+                    canonicalID)) {
+                    log.trace('access denied due to bucket acl\'s');
+                    // if access denied at the bucket level, no access for
+                    // any of the objects so all results will be error results
+                    inPlay.forEach(key => {
```
michael-zapata: Same here again for `.map()`.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                                errorResults.push({
+                                    key,
+                                    error: err,
+                                });
+                            }
+                            if (objMD['content-length']) {
+                                totalContentLengthDeleted +=
+                                    objMD['content-length'];
+                            }
+                            deleted.push(key);
```
michael-zapata: So in case of an error during deletion, we still send it back as deleted as well? That's strange, especially since that means a duplicated entry.

lib/auth/in_memory/backend.js
```diff
@@ -139,6 +143,77 @@ const backend = {
...
+     * params that have the
+     * same value for each requestContext to be constructed in Vault
+     * @param {object} requestContextParams.paramaterize - params that have
+     * arrays as values since a requestContext needs to be constructed with
+     * each option in Vault
+     * @param {string} userArn - arn of requesting user
+     * @param {object} log - log object
+     * @param {function} cb - callback with either error or an array
+     * of authorization results
+     * @returns {callback} with either error or array of authorization results
```
michael-zapata: Technically, this returns `undefined`. Also, please check the syntax for callback definition in JSDoc.

http://usejsdoc.org/tags-callback.html

lib/auth/in_memory/backend.js
```diff
@@ -139,6 +143,77 @@ const backend = {
...
+                        isAllowed: true,
+                        arn: _buildArn(bucket,
+                        requestContextParams
+                        .constantParams.specificResource),
+                    });
+                });
+            }
+        } else if (specificResourceParams) {
+            // just object is parameterized
+            specificResourceParams.forEach(obj => {
```
michael-zapata: Same comment about `map()`.

lib/auth/in_memory/backend.js
```diff
@@ -139,6 +143,77 @@ const backend = {
...
+                        results.push({
+                            isAllowed: true,
+                            arn: _buildArn(bucket, obj),
+                        });
+                    });
+                });
+            } else {
+                // just bucket is parameterized
+                generalResourceParams.forEach(bucket => {
+                    results.push({
```
michael-zapata: Same comment about `map()`.

lib/auth/vault.js
```diff
@@ -80,7 +80,10 @@ function vaultSignatureCb(err, authInfo, log, callback, streamingV4Params) {
  */
 function authenticateV2Request(params, requestContexts, callback) {
     params.log.debug('authenticating V2 request');
-    const serializedRCsArr = requestContexts.map(rc => rc.serialize());
+    let serializedRCsArr;
+    if (requestContexts) {
```
michael-zapata: From what I understood, we always at least construct an array, even if empty. Am I mistaken?

lib/api/multiObjectDelete.js
LaurenSpiegel: It is pushing multiple items.  I could concatenate the strings but I think this looks clearer.

lib/api/multiObjectDelete.js
LaurenSpiegel: Why do you think this is returning InternalError?  The only error _parseXml returns is MalformedXML.

lib/auth/vault.js
LaurenSpiegel: We use null for multiObjectDelete (and streaming v4 auth so this change was already in the v4 function).

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,307 @@
...
+                    // for key so add to deleted list and move on
+                    if (err && err.NoSuchKey) {
+                        deleted.push(key);
+                        return moveOn();
+                    }
+                    return services.deleteObject(bucketName, objMD, key, log,
+                        err => {
+                            if (err) {
+                                const obj = {};
+                                obj[key] = err;
```
electrachong: Do you use this object `obj` anywhere? For some reason, I am missing it.

LaurenSpiegel: Good catch!  It was residue from an incomplete refactor.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,346 @@
...
+        },
+        function checkBucketMetadata(quietSetting, errorResults, inPlay, next) {
+            // if no objects in play, no need to check ACLs,
+            // just move on
+            if (inPlay.length === 0) {
+                return next(null, quietSetting, errorResults, inPlay);
+            }
+            return metadata.getBucket(bucketName, log, (err, bucketMD) => {
+                if (err) {
+                    log.trace('error retrieving bucket metadata');
```
rahulreddy: Please log the error object here.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,346 @@
...
+    let totalContentLengthDeleted = 0;
+    // for obj deletes, no need to check acl's at object level
+    // (authority is at the bucket level for obj deletes)
+
+    // doing 5 requests at a time. note that the data wrapper
+    // will do 5 parallel requests to data backend to delete parts
+    return async.forEachLimit(inPlay, 5, (key, moveOn) => {
+        metadata.getObjectMD(bucketName, key, log, (err, objMD) => {
+            // if general error from metadata return error
+            if (err && !err.NoSuchKey) {
```
LaurenSpiegel: Good call on the log.  As discussed, I opted to give the user a view of what happened rather than doing a blanket error for all objects.  Not a slam dunk choice but seems better to let them know there were some successes.

lib/api/multiObjectDelete.js
```diff
@@ -0,0 +1,346 @@
...
+            }
+            // if particular key does not exist, AWS returns success
+            // for key so add to successfullyDeleted list and move on
+            if (err && err.NoSuchKey) {
+                successfullyDeleted.push(key);
+                return moveOn();
+            }
+            return services.deleteObject(bucketName, objMD, key, log,
+                err => {
+                    if (err) {
```
LaurenSpiegel: same

lib/api/multiObjectDelete.js
rahulreddy: Ok, good for me now. We can revisit this behavior in future if we run into problems.

LaurenSpiegel:
@ironman-machine steel feeling it?
SCALITY_INTEGRATION_BRANCH=ft/batchDelete
SCALITY_VAULT_BRANCH=ft/batchDelete
SCALITY_UTAPI_BRANCH=ft/batchDelete
SCALITY_VAULTCLIENT_BRANCH=ft/batchDelete

lib/api/multiObjectDelete.js
michael-zapata: My bad, I read it as `parseString` for some reason. But since this is always this error, you could simplify the conditional below this way:
```es6
if (err || objects.length < 1 || objects.length > 1000) {
    return next(errors.MalformedXML);
}
```
That would make the intent clearer for the reader without a need to fallback to the previous function.

LaurenSpiegel:
@ironman-machine to be sure
SCALITY_INTEGRATION_BRANCH=ft/batchDelete
SCALITY_VAULT_BRANCH=ft/batchDelete
SCALITY_UTAPI_BRANCH=ft/batchDelete
SCALITY_VAULTCLIENT_BRANCH=ft/batchDelete

LaurenSpiegel:
@ironman-machine with slowed down object creation
SCALITY_INTEGRATION_BRANCH=ft/batchDelete
SCALITY_VAULT_BRANCH=ft/batchDelete
SCALITY_UTAPI_BRANCH=ft/batchDelete
SCALITY_VAULTCLIENT_BRANCH=ft/batchDelete

LaurenSpiegel:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/6130

PR #305, opened 2016-10-21
COMPAT: set ACLs with an empty AccessControlList

### LINKED ISSUE #304: COMPAT: set ACLs with an empty AccessControlList
COMPAT: set ACLs with an empty AccessControlList should not return any error.

```
<AccessControlPolicy xmlns= "http://s3.amazonaws.com/doc/2006-03-01/">
         <Owner>
                <ID>852b113e7a2f25102679df27bb0ae12b3f85be6BucketOwnerCanonicalUserID</ID>
                          <DisplayName>OwnerDisplayName</DisplayName>  
         </Owner>
         <AccessControlList></AccessControlList>
</AccessControlPolicy>
```

On **ceph/s3-tests**, these changes fix:

`ERROR: s3tests.functional.test_s3.test_bucket_acl_revoke_all`

nicolas2bert:
FIXES #304
I checked against AWS S3 and when:

- adding multiple AccessControlList section

OR

- AccessControlList section is not empty and there is no Grant section
it returns a MalformedACLError.

If AccessControlList section is empty `<AccessControlList></AccessControlList>`, AWS does not throw any error.

### DIFF
--- a/lib/utilities/aclUtils.js
+++ b/lib/utilities/aclUtils.js
@@ -103,8 +103,10 @@ aclUtils.parseAclXml = function parseAclXml(toBeParsed, log, next) {
         }
         if (!result.AccessControlPolicy
                 || !result.AccessControlPolicy.AccessControlList
-                || !result.AccessControlPolicy
-                .AccessControlList[0].Grant) {
+                || result.AccessControlPolicy.AccessControlList.length !== 1
+                || (result.AccessControlPolicy.AccessControlList[0] !== '' &&
+                  Object.keys(result.AccessControlPolicy.AccessControlList[0])
+                    .some(listKey => listKey !== 'Grant'))) {
             log.warn('invalid acl', { acl: result });
             return next(errors.MalformedACLError);
         }
--- a/tests/unit/api/bucketPutACL.js
+++ b/tests/unit/api/bucketPutACL.js
@@ -301,6 +301,105 @@ describe('putBucketACL API', () => {
         });
     });
 
+    it('should set ACLs with an empty AccessControlList section', done => {
+        const testACLRequest = {
+            bucketName,
+            namespace,
+            headers: { host: `${bucketName}.s3.amazonaws.com` },
+            post: '<AccessControlPolicy xmlns=' +
+                    '"http://s3.amazonaws.com/doc/2006-03-01/">' +
+                  '<Owner>' +
+                    '<ID>852b113e7a2f25102679df27bb0ae12b3f85be6' +
+                    'BucketOwnerCanonicalUserID</ID>' +
+                    '<DisplayName>OwnerDisplayName</DisplayName>' +
+                  '</Owner>' +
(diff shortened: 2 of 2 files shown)


lib/utilities/aclUtils.js
```diff
@@ -103,8 +103,8 @@ aclUtils.parseAclXml = function parseAclXml(toBeParsed, log, next) {
         }
         if (!result.AccessControlPolicy
                 || !result.AccessControlPolicy.AccessControlList
-                || !result.AccessControlPolicy
-                .AccessControlList[0].Grant) {
+                || (result.AccessControlPolicy.AccessControlList[0] !== '' &&
+                  !result.AccessControlPolicy.AccessControlList[0].Grant)) {
```
ttvi: Are there multiple ACLs for a single bucket/object? In that case, what if the first ACL is OK but the rest is not?

tests/unit/api/bucketPutACL.js
```diff
@@ -301,6 +301,104 @@ describe('putBucketACL API', () => {
...
+                assert.strictEqual(md.getAcl().READ.length, 0);
+                assert.strictEqual(md.getAcl().WRITE.length, 0);
+                assert.strictEqual(md.getAcl().WRITE_ACP.length, 0);
+                assert.strictEqual(md.getAcl().READ_ACP.length, 0);
+                done();
+            });
+        });
+    });
+
+    it('should set ACLs without AccessControlList section', done => {
```
ttvi: should not be able to set ACLs ...

tests/unit/api/bucketPutACL.js
```diff
@@ -301,6 +301,104 @@ describe('putBucketACL API', () => {
...
+                    '<DisplayName>OwnerDisplayName</DisplayName>' +
+                  '</Owner>' +
+                  '<AccessControlList></AccessControlList>' +
+                '</AccessControlPolicy>',
+            url: '/?acl',
+            query: { acl: '' },
+        };
+
+        bucketPutACL(authInfo, testACLRequest, log, err => {
+            assert.strictEqual(err, undefined);
```
ttvi: I think you have mentioned that AWS doesn't accept empty or multiple ACLs but it doesn't seem to be what being tested here.

ttvi: unless "empty" means `'<AccessControlList></AccessControlList>'`. But even in such case, shouldn't it return the malformxml error here?

tests/unit/api/bucketPutACL.js
nicolas2bert: Thanks

tests/unit/api/bucketPutACL.js
nicolas2bert: My bad, I rephrased it to clarify it.

nicolas2bert:
@ironman-machine please

PR #307, opened 2016-10-24
Add IPcheck to healthcheck routes

dora-korpar:


dora-korpar:
@ironman-machine gogo gadget Jarvis

dora-korpar:
End-to-end failed because the IP check worked and only localhost was allowed.
Changes made to Federation to allow aggressor to access healthcheck routes here:
https://github.com/scality/Federation/pull/660

RachedBenMustapha:
http://ci.ironmann.io/gh/scality/Integration/5638 passed (with proper Federation and Integration changes).

dora-korpar:
@ironman-machine get it

lib/routes.js
```diff
@@ -72,7 +82,10 @@ export default function routes(req, res, logger) {
     log.end().addDefaultFields(clientInfo);
 
     if (req.url === '/_/healthcheck') {
-        return setHealthCheckResponse(req, res, log);
+        if (!checkIP(clientInfo.clientIP)) {
+            return writeResponse(403, res, log);
```
rahulreddy: You should use arsenal error as we do for other routes. `AccessDenied` would be the best fit here.
```js
....
const error = errors.AccessDenied;

**Comment by RachedBenMustapha** [comment]
@rahulreddy updated

**Comment by RachedBenMustapha** [comment]
@ironman-machine
SCALITY_INTEGRATION_BRANCH=FT/healthchecks-whitelist-aggressor

**Comment by RachedBenMustapha** [comment]
Test `should list the parts of a multipart upload` is failing, but does not seem related, will merge this tomorrow after investigation.

File: lib/routes.js [hunk 2]
```diff
@@ -40,24 +40,38 @@ export function isHealthy() {
...
+}
+
+function healthcheckRouteHandler(req, res, log) {
     if (isHealthy()) {
         if (req.method === 'GET' || req.method === 'POST') {
-            res.writeHead(200, { 'Content-Type': 'application/json' });
-            res.write('{}');
-            log.debug('healthcheck response',
-                { httpMethod: req.method, httpCode: res.statusCode });
+            writeResponse(res, null, log);
```
rahulreddy: Not related to the MPU error but I just realized that writeResponse is async and is not completed before the return of this method. This may trick the test into believing that the operation is complete while its not.

lib/routes.js
```diff
@@ -72,7 +86,10 @@ export default function routes(req, res, logger) {
     log.end().addDefaultFields(clientInfo);
 
     if (req.url === '/_/healthcheck') {
-        return setHealthCheckResponse(req, res, log);
+        if (!checkIP(clientInfo.clientIP)) {
+            return writeResponse(res, errors.Forbidden, log);
```
LaurenSpiegel: will you please change this to errors.AccessDenied to be consistent with what Rahul did in the other projects and because the message that goes along with forbidden ('authentication failed') doesn't quite make sense here.  See https://github.com/scality/Arsenal/blob/master/errors/arsenalErrors.json#L346

RachedBenMustapha:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=FT/healthchecks-whitelist-aggressor

PR #310, opened 2016-10-25
COMPAT: return InvalidPart if complete MPU with missing part

### LINKED ISSUE #309: COMPAT: should return InvalidPart error if complete MPU request with a missing part
Should return `InvalidPart` error instead of `InvalidPartOrder` if the complete multipart upload request contains xml with a missing part.

nicolas2bert:
FIXES #309

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -172,17 +172,24 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             // Check list sent to make sure valid
             // If the number of the parts in the JSON does not
             // match the number of parts stored, return error msg
-            if (jsonList.Part.length !== storedPartsAsObjects.length) {
+            const partLength = jsonList.Part.length;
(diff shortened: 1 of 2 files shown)


lib/api/completeMultipartUpload.js
```diff
@@ -182,7 +182,13 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 // in numerical order
                 // return an error
                 if (Number.parseInt(part.PartNumber[0], 10) !== i + 1) {
-                    return next(errors.InvalidPartOrder);
+                    // If not in order return InvalidPartOrder
+                    if (jsonList.Part.some(partVal =>
```
LaurenSpiegel: Instead of looping over the list within the loop, can't we check that the part is just not greater than the max part and return InvalidPart if it is?

nicolas2bert: Actually, you are totally right. Don't know how I came this messy thing. Thanks.

nicolas2bert:
@ironman-machine ok

PR #320, opened 2016-10-27
CLEANUP: Remove xml module usage in bucketGetAcl api

### LINKED ISSUE #269: Remove xml module usage in bucketGetAcl api

bennettbuchanan:
Fix #269
- Use `aclUtils.convertToXml` function to generate the XML DOM
- Remove `aclUtils.constructGetACLsJson` function because XML
  is generated with the `grantInfo` object and does not need to
  be formatted for the XML module
- Remove `utils.convertToXml` because its purpose was to 1) handle
  JSON construction with `aclUtils.constructGetACLsJson` 2) return
  the XML DOM using the XML module

### DIFF
--- a/lib/api/bucketGetACL.js
+++ b/lib/api/bucketGetACL.js
@@ -1,7 +1,6 @@
 import aclUtils from '../utilities/aclUtils';
 import constants from '../../constants';
 import services from '../services';
-import utils from '../utils';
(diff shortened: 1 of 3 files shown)


bennettbuchanan:
@ironman-machine

PR #321, opened 2016-10-27
Skip test header mismatched MD5

nicolas2bert:
Until we figure out how to parse the response in the CI


nicolas2bert:
@ironman-machine go!

PR #322, opened 2016-10-27
FIX: fix s3curl parsing of http code

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine check it

Issue #312, opened 2016-10-25
REFACTOR: Swap order of delete metadata and data operations


electrachong:
Currently to delete an object, S3 will first delete its data and then the metadata. This can create situations in which an object is gone but still exists in MD ('ghosts'). We instead want to delete the metadata first, logging the object's location, and then delete the data. This may create data orphans if S3 happens to crash inbetween, but will allow us to eventually implement a garbage collector that checks the logs to remove any orphans.

DavidPineauScality:
Note that we are not keeping the logs eternally (ELK is size-capped, and each individual log is rotated + size capped). This may lead to loss of track pointing to the orphans

ttvi:
This is just a step toward a stand-alone GC. Because we can infer both deletions and data locations from MetaData log, the garbage collectors will later be able to delete the right things, not depending on S3 status. Saving deleted objects' locations by S3 is just a temporary fix at the moment until we have GC. In the end, not being able to delete some parts of data costs us some, but doesn't cost customers anything.

ttvi:
Oh and I also think it's not putting data location in ELK but can be using a normal log file on disk.

RachedBenMustapha:
This kind of temporary-but-important data would be a nice fit for redis's list structure: push keys into a set before deleting them, remove them when deletes are done and committed, and periodically check that the set is empty to either alert or cleanup. It can easily turn into a work-queue that can do massive batch-deletes out of band.

Also, it keeps the S3 pods stateless and does not rely on logs which should not be relied on for anything except human eyes.

rahulreddy:
I am not really sold on using Redis as a persistent datastore. We are using it as a persistent datastore for Utapi right now but at some point in future we are gonna have an archival strategy which will bring it back to being a temporal datastore as it should be. And storing data in a temp. area assuming that it would be there permanently would be a bad idea.

So my proposal is why not store the sproxyd keys in a hidden bucket in metadata à la `sproxyd..gc`. This way S3 (and may be metadata) can place the keys in here and some independent stateless gc process will go through this list deleting the keys.

RachedBenMustapha:
@rahulreddy these key lists are short lived and should never be archived. They are just there for the time between metadata updates and a possible inspection in case of a crash - deleted space should be reclaimed in a matter of minutes to avoid any bad surprises.

It's actually more of a fit for redis than what we currently put in redis and bothering the metadata backend with workqueue-like usage seems dangerous.

ttvi:
[...]

PR #324, opened 2016-10-27
CLEANUP: Reorder so MD is deleted before data

electrachong:
Related to issue #312
- Add optional callback to batchDelete so it can be used to delete object(s) instead of duplicating code
- Swap order so MD is deleted before data
- Remove redundant log traces and errors
- Add functional test for deleteObject to test deleting an object uploaded in parts

This at least swaps the order of deletion so MD is deleted before data, which is preferable to potential ghost MD of a deleted data object. The issue of how to persist the delete keys for confirmation of data deletion still needs to be resolved.

### DIFF
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -26,6 +26,28 @@ if (config.backends.data === 'mem') {
     implName = 'sproxyd';
 }
 
+/**
+ * _retryDelete - Attempt to delete key again if it failed previously
+ * @param {string} key - location of the object to delete
+ * @param {object} log - Werelogs request logger
+ * @param {number} count - keeps count of number of times function has been run
+ * @param {function} cb - callback
+ * @returns undefined and calls callback
+ */
+const MAX_RETRY = 2;
+
+function _retryDelete(key, log, count, cb) {
+    if (count > MAX_RETRY) {
+        return cb(errors.InternalError);
+    }
+    return client.delete(key, log.getSerializedUids(), err => {
+        if (err) {
+            return _retryDelete(key, log, count + 1, cb);
+        }
+        return cb();
+    });
+}
+
 const data = {
     put: (cipherBundle, value, valueSize, keyContext, log, cb) => {
         assert.strictEqual(typeof valueSize, 'number');
@@ -93,17 +115,21 @@ const data = {
     },
 
     delete: (objectGetInfo, log, cb) => {
+        const callback = cb || log.end;
         // If objectGetInfo.key exists the md-model-version is 2 or greater.
         // Otherwise, the objectGetInfo is just the key string.
         const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
-        log.debug('sending delete to datastore', { implName, key,
-            method: 'delete' });
-        client.delete(key, log.getSerializedUids(), err => {
+        log.debug('sending delete to datastore', {
+            implName,
+            key,
+            method: 'delete',
+        });
+        _retryDelete(key, log, 0, err => {
             if (err) {
-                log.error('error from sproxyd', { error: err });
-                return cb(errors.InternalError);
+                log.error('error deleting object from datastore',
+                    { error: err, key });
             }
-            return cb();
+            return callback(err);
         });
     },
 
@@ -111,15 +137,14 @@ const data = {
     // replace this
     batchDelete: (locations, log) => {
         async.eachLimit(locations, 5, (loc, next) => {
-            data.delete(loc, log, err => {
-                if (err) {
-                    log.error('one part of a batch delete failed',
-                    { location: loc });
-                }
-                return next();
-            });
+            data.delete(loc, log, next);
         },
-        () => {
+        err => {
+            if (err) {
+                log.error('batch delete failed', { error: err });
+            } else {
+                log.trace('batch delete successfully completed');
+            }
             log.end();
         });
     },
--- a/lib/services.js
+++ b/lib/services.js
@@ -11,6 +11,7 @@ import data from './data/wrapper';
 import { isBucketAuthorized, isObjAuthorized } from
     './api/apiUtils/authorization/aclChecks';
 import metadata from './metadata/wrapper';
+import { logger } from './utilities/logger';
 import V4Transform from './auth/streamingV4/V4Transform';
 import removeAWSChunked from './api/apiUtils/object/removeAWSChunked';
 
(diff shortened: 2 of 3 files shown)


electrachong:
@ironman-machine

lib/data/wrapper.js
```diff
@@ -96,31 +96,36 @@ const data = {
...
             }
+            log.trace('delete object successful');
             return cb();
         });
     },
 
     // It would be preferable to have an sproxyd batch delete route to
     // replace this
-    batchDelete: (locations, log) => {
+    batchDelete: (locations, log, cb) => {
```
ttvi: Not related to this PR but `batchDelete` should be `multiDelete`. The former is more about to provide the atomicity (all or nothing deleted) while the latter could be anything (0-to-all deleted). It'll be good to use the right term.

lib/data/wrapper.js
```diff
@@ -96,31 +96,36 @@ const data = {
...
-                }
-                return next();
-            });
+            data.delete(loc, log, () => next());
         },
-        () => {
+        err => {
+            if (err && cb) {
+                return cb(err);
+            } else if (!err) {
```
ttvi: What about `err && !cb`? Is this on purpose to silent this error?

tests/functional/aws-node-sdk/test/object/deleteObject.js
```diff
@@ -0,0 +1,82 @@
...
+            })
+            .catch(err => {
+                process.stdout.write('Error in after\n');
+                throw err;
+            });
+        });
+
+        it('should delete a object uploaded in parts successfully', done => {
+            s3.deleteObject({ Bucket: bucketName, Key: objectName }, err => {
+                assert.equal(err, null,
```
michael-zapata: Should be an `assert.strictEqual()`.

lib/data/wrapper.js
```diff
@@ -96,31 +96,36 @@ const data = {
         // If objectGetInfo.key exists the md-model-version is 2 or greater.
         // Otherwise, the objectGetInfo is just the key string.
         const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
-        log.debug('sending delete to datastore', { implName, key,
+        log.debug('sending delete to dataStore', { implName, key,
```
electrachong: Just trying to be consistent; I saw another error trace used dataStore (former line 385 of services.js) and opted to use the camelcase. If lowercase is preferred because it's not an entity except in concept, I can revert back to datastore :)

lib/data/wrapper.js
electrachong: I think your comment makes sense. Maybe we can open another PR to address this.

lib/data/wrapper.js
electrachong: The intention was to let data.delete() handle logging an error in case one part of the delete failed. After consulting with Rahul, I fixed my call to data.delete to use the appropriate shorthand for next(). I also added a log.error for the batchDelete and tried to clarify the code by setting a default for the callback.  I'd welcome your comments and suggestions on it. https://github.com/scality/S3/pull/324/commits/b2776fe453ffdda0e1ea4c9a9f49dd6a83402413

tests/functional/aws-node-sdk/test/object/deleteObject.js
electrachong: Thanks, fixed in https://github.com/scality/S3/pull/324/commits/16fd12d7c236d0b106cb7ba4c5cf53b41e42ae3e

lib/services.js
```diff
@@ -377,47 +377,28 @@ export default {
...
+                    // TODO: The method of persistence of sproxy delete key will
+                    // be finalized; refer Issue #312 for the discussion. In the
+                    // meantime, we at least log the location of the data we are
+                    // about to delete before attempting its deletion.
+                    log.trace('deleteObject: metadata delete OK; about to ' +
+                    'delete object in data', { location: objectMD.location });
+                    if (objectMD.location === null) {
+                        return cb();
+                    } else if (!Array.isArray(objectMD.location)) {
+                        return data.delete(objectMD.location, log, cb);
```
ttvi: I don't think passing the callback in the `data.delete` here and below is logical. At the step, we have successfully deleted metadata, and in the user point of view the object is deleted. How user will perceive the error when data deletion fails?

electrachong: Good point... this probably means I have to make the callback in data.delete optional as well, if I understand correctly. I will update with a comment after I remove these callback arguments.

rahulreddy:
@ttvi @electrachong @LaurenSpiegel 
I am thinking, may be we should hold this off until we actually have GC implemented.

Let's take an example. User requests `foo` object to be deleted, we delete metadata and error in data

1.  We respond back with error. Bart tries again but we send 404 as we have no record of metadata. **We created an orphan**.
2. We respond back with success. **We created an orphan**

Currently we don't have a GC setup to deal with orphans. Up until then we should live with this issue. Otherwise we are substituting one problem with another.

Quoting @ttvi from issue #312,
> The "correct" way is to include data locations in the delete call to metadata. This location information will be stored in metadata log for GC later to use

We can use this PR after that is implemented.

lib/data/wrapper.js
```diff
@@ -96,31 +96,35 @@ const data = {
...
-                return next();
-            });
+            data.delete(loc, log, next);
         },
-        () => {
-            log.end();
+        err => {
+            const callback = cb || function _noop() {
+                return log.end();
+            };
```
ttvi: Hmm... our linting supreme leader will not like this: making unnecessary new functions. It should better be `const callback = cb || log.end` if `log.end` doesn't take any argument or you have to use ternary in the returns below.

lib/services.js
ttvi: Since we don't really care if data deletion fails at the moment, no (or not harmful) callback in `data.delete` is necessary. Or you can make a loop retrying (with limit) deleting the stubborn parts.

ttvi:
@rahulreddy I would say that this PR targets a specific problem itself. Previously, we can have objects that exist but their data is corrupted if deletion fails. This PR solves that and introduces another non-harmful side-effect: we can have object deleted but their data not purged. This side-effect will be solved when GC is available.

For short, this PR reduces the severeness of the issue from "corrupting data" to "not purging deleted data" (users cannot see deleted data anyway). This PR still deserves some approvals.

LaurenSpiegel:
@rahulreddy, I agree with @ttvi.  We discussed this and we prefer to show the user a view of the world without ghosts even if it means that orphans might have to be cleaned up.

rahulreddy:
@ttvi @LaurenSpiegel Thanks for making it clearer. 
>. Previously, we can have objects that exist but their data is corrupted if deletion fails. 

Given that we actually ran in to a streaming issue last week, I agree with going on with this approach.

lib/data/wrapper.js
electrachong: Thanks :)

lib/services.js
electrachong: I've removed the callbacks and for now added one retry if deleting an object fails in `data.delete`.

electrachong:
@ironman-machine check it, please!

lib/data/wrapper.js
```diff
@@ -26,6 +26,22 @@ if (config.backends.data === 'mem') {
     implName = 'sproxyd';
 }
 
+/**
+ * _retryDelete - Attempt to delete key again if it failed previously
+ * @param  {string} key - location of the object to delete
+ * @param  {object} log - Werelogs request logger
+ * @returns {boolean} true or false - indicates whether delete was successful
+ */
+function _retryDelete(key, log) {
```
MathieuCassagne: Is this function working as intended ?
For what I see and understand, it's always returning ```undefined```.
I think you wanted to write something more like:

```es6
function _retryDelete(key, log) {
    log.trace('retrying delete one more time...');
    return client.delete(key, log.getSerializedUids(), err => {
        if (err) {
            return false;
        }
        return true;
    });
}
```

But this form only works when the call to ```client.delete()``` is synchronous, if it is asynchronous, as with sprocydclient I guess, you'll still always undefined.

Maybe what you want is more like
```es6
const MAX_RETRY = 2;
function _retryDelete(key, log, count, cb) {
    if (count > MAX_RETRY) {
        return cb(errors.InternalError);
    }
    log.trace('retrying delete one more time...');
    return client.delete(key, log.getSerializedUids(), err => {
        if (err) {
            return _retryDelete(key, log, ++count, cb);
        }
        return cb();
    });
}
```

And the delete function become:

```es6
    log.debug('sending delete to datastore', {
        implName, key,
        method: 'delete',
    });
    _retryDelete(key, log, 0, err => {
        if (err) {
            log.error('error deleting object from datastore', {
                error: err ,
                location: key
            });
        }
        return callback(err);
    });
}
```

With this solution you'll try MAX_RETRY times before calling the callback with an error.

electrachong: Hmm.. yeah. So you're saying because the call to client.delete() is asynch, _retryDelete will finish execution and return before client.delete() executes its callback, right? Thanks for the catch, I'm still trying to get a good handle on this topic. Your recursive solution looks good, I will update with the changes and test my previous code to make sure I understand.

electrachong:
@ironman-machine check it

lib/data/wrapper.js
MathieuCassagne: Yes, it was exactly that :)

ttvi: My problem is that I usually read only the first sentence of each comment.

lib/data/wrapper.js
```diff
@@ -26,6 +26,28 @@ if (config.backends.data === 'mem') {
     implName = 'sproxyd';
 }
 
+/**
+ * _retryDelete - Attempt to delete key again if it failed previously
+ * @param {string} key - location of the object to delete
+ * @param {object} log - Werelogs request logger
+ * @param {number} count - keeps count of number of times function has been run
+ * @param {function} cb - callback
+ * @returns {boolean} true or false - indicates whether delete was successful
```
LaurenSpiegel: returns undefined and calls callback

electrachong: thanks, fixed

LaurenSpiegel:
looks good.  Good that you said related to #312 since we don't want to close with this PR.  We still want to deal with the logging issue.  Please update that issue once this is merged.

electrachong:
@ironman-machine one more check

electrachong:
end-to-end passed, it's mergeable

PR #326, opened 2016-10-28
List of applications tested with S3 Server

nicolas2bert:


TESTED.md
```diff
@@ -0,0 +1,11 @@
...
+
+### List of applications that have been tested with S3 Server
+
+- [Cyberduck](https://cyberduck.io/?l=en)
+- [Cloud Explorer](https://github.com/scality/S3/issues/277)
+- [AWS SDK](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html)
+- [s3curl](https://github.com/rtdp/s3curl)
+- [s3cmd](http://s3tools.org/s3cmd)
+- [fog (Ruby)](http://fog.io/storage/)
+- [boto (python)](https://github.com/boto/boto)
```
DavidPineauScality: should we reference separately boto2 and boto3 ? Or do you think it's irrelevant as they should be providing the same compatibility ?

rahulreddy:
I think we should include config examples for the applications as well, as each application has it's own quirks for setup. This would make it easy for end users who are new to S3.

TESTED.md
rahulreddy: We should separate them (boto2 and boto3) and provide sample config for each of them. Both have different options for setup.

nicolas2bert: Got it, thanks guys

nicolas2bert:
@rahulreddy Yes that makes sense, I am going to do that.

TESTED.md
```diff
@@ -0,0 +1,11 @@
+#  Tested with S3
+
+### List of applications that have been tested with S3 Server
+
+- [Cyberduck](https://cyberduck.io/?l=en)
+- [Cloud Explorer](https://github.com/scality/S3/issues/277)
+- [AWS SDK](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html)
```
LaurenSpiegel: we should include both the java sdk and the javascript sdk.

PR #330, opened 2016-10-28
make usage of kms.destroyBucketKey on bucket deletion

### LINKED ISSUE #328: make usage of kms.destroyBucketKey

ploki:
fix #328

### DIFF
--- a/lib/api/apiUtils/bucket/bucketDeletion.js
+++ b/lib/api/apiUtils/bucket/bucketDeletion.js
@@ -7,6 +7,7 @@ import { logger } from '../../../utilities/logger';
 import constants from '../../../../constants';
 import createKeyForUserBucket from './createKeyForUserBucket';
 import metadata from '../../../metadata/wrapper';
+import kms from '../../../kms/wrapper';
 
 const usersBucket = constants.usersBucket;
 const oldUsersBucket = constants.oldUsersBucket;
@@ -171,7 +172,16 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
         }
         return metadata.deleteBucket(bucketName, log, err => {
             log.trace('deleting bucket from metadata');
-            return cb(err);
+            if (err) {
+                return cb(err);
+            }
+            const serverSideEncryption = bucketMD.getServerSideEncryption();
+            if (serverSideEncryption &&
+                serverSideEncryption.algorithm === 'AES256') {
+                const masterKeyId = serverSideEncryption.masterKeyId;
+                return kms.destroyBucketKey(masterKeyId, log, cb);
+            }
+            return cb();
         });
     });
 }
--- a/lib/kms/wrapper.js
+++ b/lib/kms/wrapper.js
@@ -111,6 +111,25 @@ export default class KMS {
         }
     }
 
+    /**
+     *
+     * @param {string} bucketKeyId - the Id of the bucket key
+     * @param {object} log - logger object
+     * @param {function} cb - callback
+     * @returns {undefined}
+     * @callback called with (err)
+     */
+    static destroyBucketKey(bucketKeyId, log, cb) {
+        log.debug('deleting bucket key', { bucketKeyId });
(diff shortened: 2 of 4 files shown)


lib/api/apiUtils/bucket/bucketDeletion.js
```diff
@@ -151,12 +152,23 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
...
+            }
+            return next();
+        },
         function addDeleteFlagStep(next) {
             log.trace('adding deleted attribute to bucket attributes');
             // Remove transient flag if any so never have both transient
             // and deleted flags.
             bucketMD.removeTransientFlag();
             bucketMD.addDeletedFlag();
+            bucketMD.setServerSideEncryption(undefined);
```
MathieuCassagne: You can just call the function withou passing ```undefined```.
Only cosmetic :)

lib/api/apiUtils/bucket/bucketDeletion.js
```diff
@@ -151,12 +152,23 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
                     return _deleteMPUbucket(bucketName, log, next);
                 });
         },
+        function destroyBucketKey(next) {
```
LaurenSpiegel: I think this step should only be done after the actual bucket is destroyed.  If we destroy the key at this step and the bucket is not actually destroyed, then if we "invisibly recreate" the flagged bucket on an object put for instance, we will be missing the key.

lib/kms/wrapper.js
```diff
@@ -111,6 +111,25 @@ export default class KMS {
...
+     * @param {object} log - logger object
+     * @param {function} cb - callback
+     * @returns {undefined}
+     * @callback called with (err)
+     */
+    static destroyBucketKey(bucketKeyId, log, cb) {
+        log.debug('deleting bucket key', { bucketKeyId });
+        client.destroyBucketKey(bucketKeyId, log, err => {
+            if (err) {
+                log.warn('error from kms', { implName, error: err });
```
rahulreddy: This should be error level to dump the logs in case of error. Also please add method to the log.

lib/api/apiUtils/bucket/bucketDeletion.js
```diff
@@ -151,12 +152,23 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
...
+            const serverSideEncryption = bucketMD.getServerSideEncryption();
+            if (serverSideEncryption) {
+                if (serverSideEncryption.algorithm === 'AES256') {
```
rahulreddy:
```js
if (serverSideEncryption && serverSideEncryption.algorithm === 'AES256') {
...

lib/api/apiUtils/bucket/bucketDeletion.js
rahulreddy: Are object PUTs allowed on buckets with delete flags?

Apart from that I agree with deleting the key after the bucket is deleted or marked for deletion.

ploki: @LaurenSpiegel are you sure the bucket may be silently recreated on object PUT? the only usage of BucketInfo::removeDeletedFlag I saw is made from bucketCreation.js.

BTW you're right, the bucket encryption key must last as long as the flagged bucket exists in the metadata backend

LaurenSpiegel:
@rahulreddy and @ploki we recreate the deleted flag bucket on an object put (and on an acl bucket change) so we should definitely only delete the key at the end.  https://github.com/scality/S3/blob/master/lib/api/objectPut.js#L176

ploki:
Ok, thanks for the explanation. Code has been modified to delete the bucket encryption key as a last step

lib/kms/wrapper.js
rahulreddy: We can address this later on a different PR.

ploki:
@ironman-machine would you mind if I ask for an end to end test ?

PR #332, opened 2016-10-28
create bucket keys after checking bucket existence

### LINKED ISSUE #314: KMS: Generate encryption key each time we try to create a bucket
Actually, try to create a bucket already existing lead to a fail from kms, because the key already exist. This is done because here https://github.com/scality/S3/blob/master/lib/api/apiUtils/bucket/bucketCreation.js#L159 we try to create in parrallel both bucket and bucket encryption information, and we do not have a check for already existing key. That being said, create the encryption key before know if the bucket exist is not something we should do :) Same for an bucket creation error, we should delete the related key against kms

ploki:
fix #314

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -122,6 +122,29 @@ export function cleanUpBucket(bucketMD, canonicalID, log, callback) {
 }
 
 /**
+ * Manage the server side encryption on bucket creation, as a side effect
+ * a bucket key is created in the kms
+ * @param {string} bucketName - name of bucket
+ * @param {object} headers - request headers
+ * @param {function} log - Werelogs logger
+ * @param {function} cb - called on completion
+ * @returns {undefined}
+ * @callback called with (err, sseInfo: object)
+ */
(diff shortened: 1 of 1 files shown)


lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -202,15 +213,19 @@ export function createBucket(authInfo, bucketName, headers,
             return cb(errors.BucketAlreadyExists);
         }
         const newBucketMD = results.prepareNewBucketMD;
-        const serverSideEncryption = results.serverSideEncryption;
-        if (serverSideEncryption) {
-            newBucketMD.setServerSideEncryption(serverSideEncryption);
-        }
         if (existingBucketMD === 'NoBucketYet') {
             log.trace('new bucket without flags; adding transient label');
-            newBucketMD.addTransientFlag();
```
LaurenSpiegel: I'd prefer to keep the log about the flag right before adding the flag so please either move the log down to below or move the newBucketMD.addTransientFlag() back to up here.

ploki: right!

PR #333, opened 2016-10-28
BF: Fix data retrieval for listMultipartUploads API

bennettbuchanan:
- `services.getMultipartUploadListing` in listMultipartUploads.js is
  passing an `undefined` DisplayName for the owner of the MPU and
  an `undefined` StorageClass for an upload. This results in the
  following JSON response for the listMultipartUploads API, respectively:
  
  ``` javascript
  "Owner": { "DisplayName": "undefined", ... }
  "Uploads": [{ ... "StorageClass": "undefined", ... }]
  ```
- These changes in getMultipartUploadListing.js fixes the
  issues stated in the above bullet point by retrieving the proper
  values for the keys `DisplayName` and `StorageClass`

### DIFF
--- a/lib/metadata/in_memory/getMultipartUploadListing.js
+++ b/lib/metadata/in_memory/getMultipartUploadListing.js
@@ -83,8 +83,8 @@ export default function getMultipartUploadListing(bucket, params, callback) {
             initiatorID: storedMD.initiator.ID,
             initiatorDisplayName: storedMD.initiator.DisplayName,
             ownerID: storedMD['owner-id'],
-            onwerDisplayName: storedMD['owner-display-name'],
-            storageClass: storedMD.storageClass,
+            ownerDisplayName: storedMD['owner-display-name'],
+            storageClass: storedMD['x-amz-storage-class'],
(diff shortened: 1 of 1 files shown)


bennettbuchanan:
@ironman-machine

lib/metadata/in_memory/getMultipartUploadListing.js
```diff
@@ -83,8 +83,8 @@ export default function getMultipartUploadListing(bucket, params, callback) {
             initiatorID: storedMD.initiator.ID,
             initiatorDisplayName: storedMD.initiator.DisplayName,
             ownerID: storedMD['owner-id'],
-            onwerDisplayName: storedMD['owner-display-name'],
```
electrachong: Nice catch

bennettbuchanan:
I have a test ready for the expected listMultipartUploads object. However, before being able to compare the object, there are other `undefined` values in the listMultipartUploadListing API that need to be handled (which are due to how the XML is being created). I have opened a new pull request (#345) that encapsulates this larger scope and contains the associated test.

PR #344, opened 2016-10-31
Allow running "lambda" predicates in response to `putObject`

austo:
Hi @GiorgioRegni and @LaurenSpiegel,

Thanks for your earlier responses. This PR is an initial "predicate registry" that handles `ObjectCreated:Put`. User code can reject a PUT operation (`Precondition Failed`) and optionally modify the body of the uploaded item. I can think of several more improvements, but I wanted to get your feedback before taking it too far.

I've tried to follow the project's coding conventions as closely as possible. Coverage is 87.39% for the predicate functionality. Please let me know if I've missed anything.

If this is something you feel would benefit the project, I would be happy to work on the multipart upload component.

Also, please note, the reverted changes to `conf/authdata.json` do not pose a security risk, as those credentials do not link to any AWS account.

Thanks!

Austin


DavidPineauScality:
Hello @austo 
To comply with our contribution Guidelines, you might want to rebase your work on top of scality's master branch, rather than merge between branches. Do no hesitate to ask if you're not confortable/familiar with this command of git; the team will be able to help you.

austo:
@DavidPineauScality 
It should be all set now, thanks for the tip!

DavidPineauScality:
Indeed, thanks :)

lib/api/objectPut.js
```diff
@@ -40,86 +41,107 @@ function _storeIt(bucketName, objectKey, objMD, authInfo, canonicalID,
...
-        size,
-        contentType: request.headers['content-type'],
-        cacheControl: request.headers['cache-control'],
-        contentDisposition: request.headers['content-disposition'],
-        contentEncoding:
-            removeAWSChunked(request.headers['content-encoding']),
-        expires: request.headers.expires,
-        headers,
+
+    registry.run({
```
LaurenSpiegel: Also, for puts where streaming v4 auth is used, we'll have to move where the parsing of the chunks is being done.  It's currently happening later in here https://github.com/scality/S3/blob/master/lib/services.js#L158 but could be moved so it is done before we run a predicate function.

lib/predicate/registry.js
```diff
@@ -0,0 +1,367 @@
...
+
+const eventInfoSpec = {
+    // TODO: expand to conver desired event types
+    eventName: /^ObjectCreated:Put$/,
+    // TODO: what is the real bucket naming spec?
+    bucket: /^[\w\-\.]+$/,
+    // TODO: make a regexp out of the prefix and cache it
+    prefix(v) {
+        if (!v) {
+            this.prefix = '';
```
austo: `this` is the context object passed to validate, not `eventInfoSpec`, we use prefix for the registry key and we don't want an NPE [here](https://github.com/austo/S3/blob/master/lib/Config.js#L309)

lib/predicate/registry.js
```diff
@@ -0,0 +1,367 @@
...
+                    },
+                    arn: runParams.bucketArn,
+                },
+                object: {
+                    key: runParams.key,
+                    size: runParams.size,
+                    eTag: runParams.etag,
+                    contentType: runParams.contentType,
+                    metadata: runParams.metadata,
+                    data: buf,
```
austo: I was trying to keep the API simple, but your streaming suggestion sounds better

lib/predicate/registry.js
```diff
@@ -0,0 +1,367 @@
+import path from 'path';
+import domain from 'domain';
+import stream from 'stream';
+import { errors } from 'arsenal';
+import validate from './validate';
+
+const PredicateError = require('swerrs').extend({
```
austo: Fair point. The `swerrs` library just helps aggregate errors over the course of multiple operations, but it's optional; `validate` can be used with a regular array.

lib/predicate/registry.js
```diff
@@ -0,0 +1,367 @@
...
+        return callback(ve.push(`No predicate exists for "${key}"`));
+    }
+    return callback(null, fn);
+}
+
+function runWrapped(context, callback) {
+    const { passthrough, event, fn } = context;
+    // NOTE: Domains are used to prevent exceptions from
+    // killing the worker process (which is not possible with
+    // promises). See tests for details.
```
austo: I don't like it either, but if the user predicate throws without a domain wrapping it, the exception will bubble up to the implicit domain created by the `cluster` module and the process will terminate. Maybe that's not a problem as long as the client's connection isn't dropped.

lib/predicate/registry.js
```diff
@@ -0,0 +1,367 @@
...
+function runPredicate(runParams, fn, callback) {
+    const { request } = runParams;
+    const chunks = [];
+    const passthrough = new stream.PassThrough();
+    passthrough.headers = request.headers; // TODO: do we need anything else??
+    passthrough.parsedContentLength = request.parsedContentLength;
+
+    request.on('data', c => {
+        chunks.push(c);
+        passthrough.write(c);
```
austo: You got me :) Let's improve this!

lib/predicate/registry.js
```diff
@@ -0,0 +1,367 @@
...
+    });
+    request.on('error', err => {
+        passthrough.end();
+        return callback(err);
+    });
+    request.on('end', () => {
+        passthrough.end();
+        return runWrapped({
+            passthrough,
+            event: makePredicateEvent(runParams, Buffer.concat(chunks)),
```
austo: So the user's `data` api would be a stream? I was trying to keep this super simple for end users, but I agree, _if_ they want the data (and they might not care since a key use case for this functionality might be metadata validation), they can should use a stream.

lib/api/objectPut.js
austo: That's great. I don't like that it's in `objectPut.js` at all, but if I hadn't started there I would probably never have gotten this PR submitted :)

austo:
Hi @LaurenSpiegel
Thanks for taking the time to review this. I definitely think the registry hook should be moved out of the object method to somewhere more universal. I also understand your concerns about dataless vs. streaming vs. transform predicates and agree that this differentiation should be part of the api. My own inclination would be to keep the directory hierarchy flat and specify all this in the `eventInfo` object to keep things flexible for users, but that's more of an implementation detail — project conventions are more important.

It would be great to setup an online chat to discuss this more next week!

Thanks again,
Austin

LaurenSpiegel:
@austo, I'm Scality S3 on skype.  Let me know if there is a time this week that works for you to chat.  I'm California time but pretty flexible.  Thanks!

austo:
@LaurenSpiegel Contact request sent.

austo:
Hi @LaurenSpiegel 

I've finally made the changes we discussed last week. Domains are gone(!) and I've moved the predicate invocation to [api.js](https://github.com/austo/S3/blob/master/lib/api/api.js#L73). All updates to the S3 object data now take place through the event passed to the predicate; s3.object.body is a [duplex stream](https://github.com/austo/S3/blob/master/lib/predicate/requestWrapper.js#L44) that clients can read from and write to, and the object body data is only placed in memory if the client reads but [doesn't write back](https://github.com/austo/S3/blob/master/lib/predicate/requestWrapper.js#L92). Clients can also update metadata, content-type, and key information by writing to the event fields, which will be [scraped](https://github.com/austo/S3/blob/master/lib/predicate/registry.js#L114) after the predicate returns.

A good place to get a feel for the API is [here](https://github.com/austo/S3/blob/master/tests/unit/predicate/simpleHandler.js#L10) and [here](https://github.com/austo/S3/blob/master/tests/unit/predicate/registry.js#L159).

Please let me know your thoughts!

lib/predicate/registry.js
```diff
@@ -0,0 +1,353 @@
...
+        out.fn = fn;
+    } catch (e) {
+        out.err = e;
+    }
+    return out;
+}
+
+function getPredicate(registry, runParams, callback) {
+    const errs =
+        validate(runParamsSpec, runParams, runParams);
```
austo: I understand why this would be confusing... It's just saying that we want to validate `runParams` against `runParamsSpec` and we want every `runParamsSpec` member function to be run in the context of `runParams`, letting us do [this](https://github.com/austo/S3/blob/master/lib/predicate/registry.js#L32).

lib/predicate/registry.js
```diff
@@ -0,0 +1,353 @@
...
+
+    // Linear search - for now we don't anticipate
+    // having too many prefixes for the same bucket.
+    Object.keys(preds).forEach(p => {
+        const plen = p.length;
+        if (plen > klen) {
+            return;
+        }
+        if (plen > maxp && k.indexOf(p) === 0) {
+            maxp = plen;
```
austo: Yes, forgot to add that comment.

lib/predicate/validate.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    switch (specType) {
+    case 'array':
+        candidate.forEach((v, i) => {
+            validateAllFields(spec[0], v, `${prefix}.${i}`, errors, ctx);
+        });
+        return;
+    case 'object':
+    case 'error':
```
austo: `Object.prototype.toString.call(new Error())` (the input to the switch statement)  returns `[object Error]`, which is then lowercased. Fallthrough is intentional.

lib/predicate/requestWrapper.js
```diff
@@ -0,0 +1,114 @@
+import stream from 'stream';
+import crypto from 'crypto';
+
+const reqSym = Symbol('req');
+const bufSym = Symbol('buf');
+const outputSym = Symbol('output');
+
```
austo: I blew it here. Thanks for testing this one :) Turns out the file I was using was too small to trigger a second `read()` event. I've updated [requestWrapper](https://github.com/austo/S3/blob/master/lib/predicate/requestWrapper.js#L68) to be more idiomatic and have tested the change against your `changeContent` code along with a few local tests I didn't include in the PR.

austo:
Hi @LaurenSpiegel 

Thanks for reviewing this again! I've updated the code to fix the upload bug you were seeing and have incorporated all of your suggestions except dismantling `validate` (if it's a showstopper I'll do it, but I think having the spec as a kind of "living type reference" is helpful, especially in JavaScript).

Regarding the "where do I put my imports?" question, I think the best answer is a `package.json` (and `node_modules`) in the user's predicate folder. Eventually we'd probably want to enforce this kind of sandboxing (`put` throws if there's no `package.json` in the parent folder), but if this PR lands I can write some examples to help users with best practices.

Anyway, thanks again!

LaurenSpiegel:
@austo, thanks for making the changes and incorporating gets!  I'm still getting an error when I try to use a streaming predicate.  This is my testing branch: https://github.com/scality/S3/commit/9be9a35c44a722dbf217d48412d5a1a5947b0aa2

Any ideas?

austo:
Hi @LaurenSpiegel, I ran your `changecontent`  predicate using the node.js `aws-sdk` with a 400kb upload file without seeing any issues. I tested both uploading a stream using the `aws-sdk` `s3.upload` method and uploading the same file as a buffer using `s3.putObject`; both times I was able to pull the entire "flipped" file back down. I didn't see any errors in the logs. What error were you seeing?

LaurenSpiegel:
When I put the object with s3cmd, I am printing out this error in my changecontent predicate: [Error: Stream yields empty buffer]

When I try to put the object with the aws node sdk, I am getting a failed predicate error.  At first, it was because I was using localhost as an endpoint and the validation requires an ip address (how come?) and once I changed that I can't tell why it is not validating and returning a failed predicate error.

austo:
@LaurenSpiegel sure, the IP address requirement is easy to remove (the lambda event includes it & I didn't realized `parsedHost` wasn't always an IP address).

Do you have a link to your aws-sdk test code? I still can't reproduce the issue, so maybe there's some difference there...

Silly question, but is it possible you're not uploading a JPEG image? That would cause `gm` to fail.

LaurenSpiegel:
@austo, I'm trying to upload a JPEG though I've also tried with just text files to see if it will at least run the predicate.  Are you able to make it work with s3cmd?  Here is a link to a gist with the node sdk code I tried (different commented out versions) https://gist.github.com/LaurenSpiegel/9b38352173e079d1824ef89bb745586e

austo:
It's V4 auth. I have to figure out how to recalculate the header.

LaurenSpiegel:
Hey @austo, I hope you enjoyed the holidays. Any luck with the v4 auth?  Do you mean on v4 auth with the node sdk no content-md5 is sent?  We do have a utility class that calculates the md5: https://github.com/scality/S3/blob/master/lib/utilities/MD5Sum.js

austo:
Thanks @LaurenSpiegel, hope your holiday was good too! Unfortunately I've had to shelve this for the moment because of a Feb. 1 deadline @ work. I think the v4 auth stuff will be pretty straightforward once I can focus on it (and thanks for the tip).
I realize a dangling PR is kind of a bummer, so if there's a way to "freeze" this until I can get back to it, that would be great.

LaurenSpiegel:
Thanks, @austo, I look forward to seeing the changes in February.  It's a cool feature and would love to see your great work integrated into the project 😄

PR #345, opened 2016-11-01
COMPAT: Handle `undefined` values for listMultipartUploads API

bennettbuchanan:
This commit fixes issues that were causing listMultipartUploads
to contain 'undefined' as XML content:
* `services.getMultipartUploadListing` in listMultipartUploads.js is
  passing an `undefined` DisplayName for the owner of the MPU and
  an `undefined` StorageClass for an upload. This results in the
  following JSON response for the listMultipartUploads API,
  respectively:

  ```javascript
  "Owner": { "DisplayName": "undefined", ... }
  "Uploads": [{ ... "StorageClass": "undefined", ... }]

  ```
* Changes in getMultipartUploadListing.js fixes the
  issues stated in the above bullet point by retrieving the proper
  values for the keys `DisplayName` and `StorageClass`
* For values that could be `undefined` in the listMultipartUploads
  API, use empty XML elements if the value is, in fact, `undefined`
* Add tests that check for previously `undefined` values, and remove
  legacy 'list ongoing multipart uploads' tests which would
  be redundant with this commit

### DIFF
--- a/lib/api/listMultipartUploads.js
+++ b/lib/api/listMultipartUploads.js
@@ -81,14 +81,32 @@ const _convertToXml = xmlParams => {
     xml.push('<?xml version="1.0" encoding="UTF-8"?>',
              '<ListMultipartUploadsResult ' +
                 'xmlns="http://s3.amazonaws.com/doc/2006-03-01/">',
-             `<Bucket>${xmlParams.bucketName}</Bucket>`,
-             `<KeyMarker>${escapeForXML(xmlParams.keyMarker)}</KeyMarker>`,
-             `<UploadIdMarker>${xmlParams.uploadIdMarker}</UploadIdMarker>`,
-             `<NextKeyMarker>${escapeForXML(l.NextKeyMarker)}</NextKeyMarker>`,
-             `<NextUploadIdMarker>${l.NextUploadIdMarker}</NextUploadIdMarker>`,
-             `<Delimiter>${escapeForXML(l.Delimiter)}</Delimiter>`,
-             `<Prefix>${escapeForXML(xmlParams.prefix)}</Prefix>`,
-             `<MaxUploads>${l.MaxKeys}</MaxUploads>`,
+             `<Bucket>${xmlParams.bucketName}</Bucket>`
+    );
+
+    // For certain XML elements, if it is `undefined`, AWS returns either an
+    // empty tag or does not include it. Hence the `optional` key in the params.
+    const params = [
+        { tag: 'KeyMarker', value: escapeForXML(xmlParams.keyMarker) },
+        { tag: 'UploadIdMarker', value: xmlParams.uploadIdMarker },
+        { tag: 'NextKeyMarker', value: escapeForXML(l.NextKeyMarker),
+            optional: true },
+        { tag: 'NextUploadIdMarker', value: l.NextUploadIdMarker,
+            optional: true },
+        { tag: 'Delimiter', value: escapeForXML(l.Delimiter), optional: true },
+        { tag: 'Prefix', value: escapeForXML(xmlParams.prefix),
+            optional: true },
+    ];
+
+    params.forEach(param => {
+        if (param.value) {
+            xml.push(`<${param.tag}>${param.value}</${param.tag}>`);
+        } else if (!param.optional) {
+            xml.push(`<${param.tag}></${param.tag}>`);
+        }
+    });
+
+    xml.push(`<MaxUploads>${l.MaxKeys}</MaxUploads>`,
              `<IsTruncated>${l.IsTruncated}</IsTruncated>`
     );
 
--- a/tests/functional/aws-node-sdk/test/object/mpu.js
+++ b/tests/functional/aws-node-sdk/test/object/mpu.js
@@ -1,0 +1,83 @@
+import assert from 'assert';
+
+import withV4 from '../support/withV4';
+import BucketUtility from '../../lib/utility/bucket-util';
+
+const bucket = 'object-test-mpu';
+const objectKey = 'toAbort&<>"\'';
+
+const throwErr = (str, err) => {
+    process.stdout.write(`${str}: ${err}\n`);
+    throw err;
+};
+
+const checkValues = (res, uploadId, displayName, cb) => {
+    const prefix = res.Prefix ? res.Prefix : undefined;
+    const delimeter = res.Delimiter ? res.Delimiter : undefined;
+    assert.deepStrictEqual(res.KeyMarker, '');
+    assert.deepStrictEqual(res.UploadIdMarker, '');
+    assert.deepStrictEqual(res.Prefix, prefix);
+    assert.deepStrictEqual(res.Delimiter, delimeter);
+    assert.deepStrictEqual(res.Uploads.length, 1);
+    assert.deepStrictEqual(res.Uploads[0].UploadId, uploadId);
+    assert.deepStrictEqual(res.Uploads[0].StorageClass, 'STANDARD');
+    assert.deepStrictEqual(res.Uploads[0].Owner.DisplayName, displayName);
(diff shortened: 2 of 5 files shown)


lib/api/listMultipartUploads.js
```diff
@@ -81,17 +81,28 @@ const _convertToXml = xmlParams => {
...
+        { tag: 'NextKeyMarker', value: escapeForXML(l.NextKeyMarker) },
+        { tag: 'NextUploadIdMarker', value: l.NextUploadIdMarker },
+        { tag: 'Delimiter', value: escapeForXML(l.Delimiter) },
+        { tag: 'Prefix', value: escapeForXML(xmlParams.Prefix) },
+        { tag: 'MaxUploads', value: l.MaxKeys },
+        { tag: 'IsTruncated', value: l.IsTruncated },
+    ];
+
+    params.forEach(param => {
+        if (typeof param.value !== 'undefined') {
```
rahulreddy: You can get away with comparing undefined directly
```js
if (param.value !== undefined) {

File: lib/api/listMultipartUploads.js [hunk 2]
```diff
@@ -111,11 +122,15 @@ const _convertToXml = xmlParams => {
...
                  `<DisplayName>${escapeForXML(val.Owner.DisplayName)}` +
                     '</DisplayName>',
-                 '</Owner>',
-                 `<StorageClass>${val.StorageClass}</StorageClass>`,
-                 `<Initiated>${val.Initiated}</Initiated>`,
-                 '</Upload>'
+                 '</Owner>'
         );
+
+        const storageClass = typeof val.StorageClass === 'undefined' ? '' :
```
rahulreddy: Same here. You can check for undefined directly
```js
const storageClass = val.StorageClass === undefined ? '' :
    `<StorageClass>${val.StorageClass}</StorageClass>`;

File: tests/functional/aws-node-sdk/test/object/mpu.js [hunk 3]
```diff
@@ -0,0 +1,137 @@
...
+        let s3;
+        let uploadId;
+        let displayName;
+        let userId;
+
+        beforeEach(() => {
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
+
+            process.stdout.write('Creating bucket');
```
rahulreddy: Let's get rid of these logs. We should be writing on console only when needed.

tests/functional/aws-node-sdk/test/object/mpu.js
```diff
@@ -0,0 +1,137 @@
...
+            .then(() => bucketUtil.getOwner())
+            .then(res => {
+                /* In this case, the owner of the bucket will also be the MPU
+                 * upload owner. We need these values for object comparison.
+                 */
+                displayName = res.DisplayName;
+                userId = res.ID;
+            })
+            .then(() => {
+                process.stdout.write('Creating MPU');
```
rahulreddy: Same here, let's remove the log.

lib/api/listMultipartUploads.js
```diff
@@ -81,17 +81,28 @@ const _convertToXml = xmlParams => {
...
+        { tag: 'NextKeyMarker', value: escapeForXML(l.NextKeyMarker) },
+        { tag: 'NextUploadIdMarker', value: l.NextUploadIdMarker },
+        { tag: 'Delimiter', value: escapeForXML(l.Delimiter) },
+        { tag: 'Prefix', value: escapeForXML(xmlParams.Prefix) },
+        { tag: 'MaxUploads', value: l.MaxKeys },
+        { tag: 'IsTruncated', value: l.IsTruncated },
+    ];
+
+    params.forEach(param => {
+        if (param.value !== undefined) {
```
alexandre-merle: you should check for null here as well, to cover any case :)

lib/services.js
```diff
@@ -476,6 +476,7 @@ export default {
             params.headers['content-type'];
         multipartObjectMD.expires =
             params.headers.expires;
+        multipartObjectMD.storageClass = params.storageClass;
```
alexandre-merle: it is not the same as 'x-amz-storage-class' field ? it should be better to get the 'x-amz-storage-class' already store in metadata and if needed, map it into 'storageClass' field when we get metadata, in that way, any older values store will be used :)

bennettbuchanan: This change was to accommodate an `undefined` value when using the file backend. @rahulreddy and I have traced this down to [this line](https://github.com/scality/Arsenal/blob/master/lib/algos/list/MPU.js#L58) in Arsenal. Here the value should be `tmp['x-amz-storage-class']`. I will make that change to handle the file backend, and a change to handle in-memory value in a new commit. Thanks!

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine try it with correct arsenal branch
SCALITY_ARSENAL_BRANCH=dev/BF/handle_undefined_value_MultipartUploads

LaurenSpiegel:
please open this against rel/6.2.1

bennettbuchanan:
@ironman-machine after rebase
SCALITY_ARSENAL_BRANCH=dev/BF/handle_undefined_value_MultipartUploads

bennettbuchanan:
@ironman-machine Thanks Jarvis.

bennettbuchanan:
@ironman-machine

lib/api/listMultipartUploads.js
```diff
@@ -81,17 +81,28 @@ const _convertToXml = xmlParams => {
...
-             `<Prefix>${escapeForXML(xmlParams.prefix)}</Prefix>`,
-             `<MaxUploads>${l.MaxKeys}</MaxUploads>`,
-             `<IsTruncated>${l.IsTruncated}</IsTruncated>`
+             `<Bucket>${xmlParams.bucketName}</Bucket>`
     );
 
+    const params = [
+        { tag: 'KeyMarker', value: escapeForXML(xmlParams.keyMarker) },
+        { tag: 'UploadIdMarker', value: xmlParams.uploadIdMarker },
+        { tag: 'NextKeyMarker', value: escapeForXML(l.NextKeyMarker) },
```
LaurenSpiegel: Should we be returning any xml tags for this or just nothing?  It appears on a regular listing we should not return NextMarker so I wonder if this is the same.

lib/api/listMultipartUploads.js
```diff
@@ -81,17 +81,28 @@ const _convertToXml = xmlParams => {
...
+        { tag: 'NextUploadIdMarker', value: l.NextUploadIdMarker },
```
LaurenSpiegel: Same as above for NextKeyMarker.

bennettbuchanan: `NextUploadIdMarker` is included in the XML response for this API from AWS.

bennettbuchanan:
I have pushed a commit the makes the XML tags `NextKeyMarker`, `NextUploadIdMarker`, `Delimiter`, `Prefix` optional. In the sense that if they are `undefined`, the XML response will not include them.

bennettbuchanan:
@ironman-machine after rebase onto master

bennettbuchanan:
@ironman-machine after rebase onto master
SCALITY_ARSENAL_BRANCH=dev/BF/handle_undefined_value_MultipartUploads

bennettbuchanan:
@ironman-machine after rebase and new commit
SCALITY_ARSENAL_BRANCH=dev/BF/handle_undefined_value_MultipartUploads

bennettbuchanan:
@ironman-machine 
SCALITY_ARSENAL_BRANCH=dev/BF/handle_undefined_value_MultipartUploads

bennettbuchanan:
@ironman-machine after arsenal branch merge

bennettbuchanan:
end-to-end passed

PR #346, opened 2016-11-01
CLEANUP: Remove xml module usage from bucketGetVersioning api

bennettbuchanan:
* Remove the dependency on xml module in bucketGetVersioning api
* Remove the xml module from package.json as it is no longer used
  in S3

### DIFF
--- a/lib/api/bucketGetVersioning.js
+++ b/lib/api/bucketGetVersioning.js
@@ -1,4 +1,3 @@
-import xml from 'xml';
 import services from '../services';
 
 //	Sample XML response:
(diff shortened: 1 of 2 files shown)


lib/api/bucketGetVersioning.js
```diff
@@ -15,23 +14,25 @@ import services from '../services';
...
+
+    xml.push('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>',
+             '<VersioningConfiguration ' +
+                'xmlns="http://s3.amazonaws.com/doc/2006-03-01/">'
+    );
+
     if (versioningConfiguration && versioningConfiguration.Status) {
-        constructedJSON.VersioningConfiguration.push(
-            { Status: versioningConfiguration.Status });
+        xml.push(`<Status>${versioningConfiguration.Status}</Status>`);
```
LaurenSpiegel: I assume there is no need to escape this?

lib/api/bucketGetVersioning.js
```diff
@@ -15,23 +14,25 @@ import services from '../services';
...
     }
+
     if (versioningConfiguration && versioningConfiguration.MfaDelete) {
-        constructedJSON.VersioningConfiguration.push(
-            { MfaDelete: versioningConfiguration.MfaDelete });
+        xml.push(`<MfaDelete>${versioningConfiguration.MfaDelete}</MfaDelete>`);
```
LaurenSpiegel: same

lib/api/bucketGetVersioning.js
bennettbuchanan: Yes, it appears that the only possible values of `versioningConfiguration.Status` and  `versioningConfiguration.MfaDelete` are `undefined`, `'Enabled'`, and `'Suspended'`.
See: https://github.com/scality/S3/blob/master/lib/metadata/BucketInfo.js#L63

Thus, no need for escaping.

bennettbuchanan:
@ironman-machine

PR #357, opened 2016-11-04
FIX: Signing key binary

alexandre-merle:
The binary digest for the signing key make
s3 under node v6 to generate wrong signature.
This pull request fix that.


alexandre-merle:
@ironman-machine go

PR #362, opened 2016-11-04
Address additional line breaks in streaming v4

LaurenSpiegel:
Oracle sbt appears to be adding extra line breaks between the metadata portions and data portions of the chunks in streaming v4 auth.  This PR removes the extra line breaks if any.


LaurenSpiegel:
@ironman-machine any thoughts on the matter?

LaurenSpiegel:
end to end passed

PR #375, opened 2016-11-10
FIX: Vault returning internal error

alexandre-merle:
Sending logger to Vault, to log error in
request and send back internal error to
S3

Fixes https://github.com/scality/S3/issues/374


alexandre-merle:
@ironman-machine let's go :)

alexandre-merle:
@ironman-machine better like this

SCALITY_VAULTCLIENT_BRANCH=fix/returned-internal-error

PR #377, opened 2016-11-10
BF: Fix off by one range

bennettbuchanan:
* When the byte range is half the object size from a MPU, an error
  is thrown by fs.ReadStream (`Error: start must be <= end`). This
  causes a `NetworkingError: socket hang up` error in file backend.
  This commit fixes that and modifies the test to ensure that the
  range is handled properly.

### DIFF
--- a/tests/functional/aws-node-sdk/test/legacy/tests.js
+++ b/tests/functional/aws-node-sdk/test/legacy/tests.js
@@ -307,9 +307,11 @@ describe('aws-node-sdk test suite as registered user', function testSuite() {
         },
         { it: 'should get a range from the second part of an object ' +
             'put by multipart upload',
-            range: 'bytes=5242881-5242890',
+            // The completed MPU byte count starts at 0, so the first part ends
+            // at byte 5242879 and the second part begins at byte 5242880
+            range: 'bytes=5242880-5242889',
             contentLength: '10',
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine test after rebase onto rel/6.2.1 branch

tests/functional/aws-node-sdk/test/legacy/tests.js
```diff
@@ -307,9 +307,9 @@ describe('aws-node-sdk test suite as registered user', function testSuite() {
         },
         { it: 'should get a range from the second part of an object ' +
             'put by multipart upload',
-            range: 'bytes=5242881-5242890',
+            range: 'bytes=5242880-5242889',
```
LaurenSpiegel: Please add note that byte count starts at 0 so first part ends at 5242879 and second part starts at 5242880

bennettbuchanan:
@ironman-machine thanks, Jarvis.

bennettbuchanan:
@ironman-machine go after rebase and target branch switch, please.

nicolas2bert:
S3server Docker images has been updated.
https://hub.docker.com/r/scality/s3server/

PR #382, opened 2016-11-11
COMPAT: Set 'Accept-Range' header for all ranges in getObject API

### LINKED ISSUE #378: COMPAT: Invalid `Range` parameter in getObject API does not return `AcceptRanges` key
When specifying an invalid `Range` parameter with `getObject` API, the returned object should include `AcceptRanges: 'bytes'`. This is a compatibility issue.

bennettbuchanan:
Fixes #378

If `Range` parameter is given to getObject API, the 'Accept-Range'
  header should be set to 'bytes' regardless whether or not the
  given range is valid. Prior to this change, [this test](https://github.com/scality/S3/blob/master/tests/functional/aws-node-sdk/test/legacy/tests.js#L418)
  would not return an object with an `AcceptRanges` key-value pair.

### DIFF
--- a/lib/api/objectGet.js
+++ b/lib/api/objectGet.js
@@ -50,12 +50,12 @@ function objectGet(authInfo, request, log, callback) {
             maxContentLength =
                 parseInt(responseMetaHeaders['Content-Length'], 10);
             range = parseRange(request.headers.range, maxContentLength);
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine after rebase

PR #383, opened 2016-11-11
Add backend healthcheck functionality

dora-korpar:
Creates route that will be called by load balancers, skips backends that don't have a `check` function. Includes the first implementation of a check for sproxyd for each backend option.
Depends on sproxydclient PR https://github.com/scality/sproxydclient/pull/92

### DIFF
--- a/lib/data/file/backend.js
+++ b/lib/data/file/backend.js
@@ -108,6 +108,11 @@ export const backend = {
             return callback();
         });
     },
+
+    healthcheck: (log, callback) => {
+        process.nextTick(() =>
+            callback(null, { statusCode: 200, statusMessage: 'OK' }));
+    },
 };
 
 export default backend;
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -153,6 +153,25 @@ const data = {
         client = newClient;
         return client;
     },
+
+    checkHealth: (log, cb) => {
+        client.healthcheck(log, (err, result) => {
+            const respBody = {};
+            if (err) {
+                log.error(`error from ${implName}`, { error: err });
+                respBody[implName] = {
+                    error: err,
+                };
+                // respBody also returned so error is written to response
+                return cb(err, respBody);
+            }
+            respBody[implName] = {
+                code: result.statusCode,
+                message: result.statusMessage,
+            };
+            return cb(null, respBody);
+        });
+    },
 };
 
 export default data;
--- a/lib/utilities/healthcheckHandler.js
+++ b/lib/utilities/healthcheckHandler.js
@@ -1,0 +1,98 @@
+import { errors, ipCheck } from 'arsenal';
+import _config from '../Config';
+import data from '../data/wrapper';
+import vault from '../auth/vault';
+import metadata from '../metadata/wrapper';
+import async from 'async';
+
+// current function utility is minimal, but will be expanded
+export function isHealthy() {
+    return true;
+}
+
+function writeResponse(res, error, log, results, cb) {
+    let statusCode = 200;
+    if (error) {
+        // error.code can be a string (such as ECONNREFUSED)
+        if (Number.isInteger(error.code)) {
+            statusCode = error.code;
+        } else {
+            statusCode = 500;
+        }
+    }
+    // If we're here then s3 is considered ok
+    const initial = { S3: { code: 200, message: 'OK' } };
+    const merged = results.reduce(
+        (prev, value) => Object.assign(prev, value), initial);
+
+    res.writeHead(statusCode, { 'Content-Type': 'application/json' });
+    res.write(JSON.stringify(merged));
+    res.end(() => {
+        cb(error, merged);
+    });
+}
+
+function clientCheck(req, res, log, cb) {
+    const clients = [
+        data,
+        metadata,
+        vault,
+    ];
+    const clientTasks = [];
+    clients.forEach(client => {
+        if (typeof client.checkHealth === 'function') {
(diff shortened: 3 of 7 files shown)


dora-korpar:
@ironman-machine do it, jarvis

dora-korpar:
@ironman-machine again plz

dora-korpar:
@ironman-machine attack

dora-korpar:
@ironman-machine one mo time

dora-korpar:
There was a problem with the http request path being passed to sproxydclient, which was fixed in https://github.com/scality/sproxydclient/pull/93

dora-korpar:
@ironman-machine blast off

lib/utilities/healthcheckHandler.js
```diff
@@ -0,0 +1,92 @@
...
+    if (error) {
+        // error.code can be a string (such as ECONNREFUSED)
+        if (Number.isInteger(error.code)) {
+            statusCode = error.code;
+        } else {
+            statusCode = 500;
+        }
+    }
+    // If we're here then s3 is considered ok
+    const initial = { S3: { code: 200, message: 'OK' } };
```
rahulreddy: If you can move this logic to clientCheck, this method can become generic enough that it can be reused. For example, my upcoming PR is adding additional stats to /_/healthcheck route. I could re-use this method.

lib/utilities/healthcheckHandler.js
```diff
@@ -0,0 +1,92 @@
...
+function checkIP(clientIP) {
+    return ipCheck.ipMatchCidrList(
+        _config.healthChecks.allowFrom, clientIP);
+}
+
+export default function healthcheckHandler(clientIP, deep, req, res, log) {
+    function healthcheckEndHandler(err, results) {
+        writeResponse(res, err, log, results, (error, body) => {
+            if (err) {
+                log.end().warn('error', {
```
rahulreddy: Please change the error message to be more descriptive.
Note: You can return early here and avoid the else block

lib/data/wrapper.js
```diff
@@ -153,6 +153,24 @@ const data = {
         client = newClient;
         return client;
     },
+
+    checkHealth: (log, done) => {
```
rahulreddy: Just in the file.  It doesn't have to be called done as it's just a variable.

lib/utilities/healthcheckHandler.js
dora-korpar: The initial object is meant to be used for only the /_/healthcheck route

lib/data/file/backend.js
```diff
@@ -108,6 +108,11 @@ export const backend = {
             return callback();
         });
     },
+
+    healthcheck: (log, done) => {
```
dora-korpar: done!

dora-korpar:
@ironman-machine holy moly let's go

lib/utilities/healthcheckHandler.js
dora-korpar: Did we end on leaving it as is?

rahulreddy: You can leave as is if you want to, my upcoming PR will update the response to contain stats.

lib/utilities/healthcheckHandler.js
```diff
@@ -0,0 +1,98 @@
...
+* @param {object} clientIP - IP address of client
+* @param {boolean} deep - true if healthcheck will check backends
+* @param {object} req - http request object
+* @param {object} res - http response object
+* @param {object} log - werelogs logger instance
+*/
+export default function healthcheckHandler(clientIP, deep, req, res, log) {
+    function healthcheckEndHandler(err, results) {
+        writeResponse(res, err, log, results, (error, body) => {
+            if (err) {
```
bennettbuchanan: Should this `err` not instead be `error` from the callback?

dora-korpar:
@ironman-machine work it

lib/data/wrapper.js
```diff
@@ -153,6 +153,25 @@ const data = {
         client = newClient;
         return client;
     },
+
+    checkHealth: (log, cb) => {
+        client.healthcheck(log, (err, result) => {
+            const respBody = {};
+            if (err) {
+                log.error('error from sproxyd', { error: err });
```
LaurenSpiegel: I like how the response is generic as to implementation.  Let's make the log that way too.  So, instead of 'error from sproxyd', it should be error from implName.

dora-korpar: donezo!

lib/data/wrapper.js
```diff
@@ -153,6 +153,24 @@ const data = {
         client = newClient;
         return client;
     },
+
+    checkHealth: (log, done) => {
+        client.healthcheck(log, (err, result) => {
+            const respBody = {};
+            if (err) {
+                log.warn('error from sproxyd', { error: err });
```
RachedBenMustapha: I asked specifically for warn here, because the healthcheck route will be called every few seconds by the load balancer and/or monitoring agent, so it would end up spamming logs.

rahulreddy: Gotcha, I agree.

dora-korpar:
@ironman-machine seriously

dora-korpar:
CI was hanging because of bug fixed here: https://github.com/scality/Federation/pull/720

rahulreddy:
@ironman-machine now that the problem has been fixed

dora-korpar:
@ironman-machine gogogogo

PR #384, opened 2016-11-11
Make Docker image even lighter

michel-kraemer:
I was happy to see that you made the Docker image lighter in 244a5f12cbc5fd3193d1a14a3b3be9ca160ec7b9.

I was able to reduce the image size even further.

Here are the commands I used to check the image size before and after each change:

    docker build -f DockerfileMem -t scality/s3server:mem-latest .
    docker save scality/s3server:mem-latest | pxz > 01.tar.xz && du -h 01.tar.xz

## Results

**master**

    119M    01.tar.xz

**After 'rm -rf /var/lib/apt/lists/*'**

    110M	01.tar.xz

**After 'npm cache clear':**

     95M	01.tar.xz

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,9 +6,13 @@ WORKDIR /usr/src/app
 COPY . /usr/src/app
 
 RUN apt-get update \
-    && apt-get install -y python git build-essential \
-    && npm install \
-    && apt-get autoremove -y python build-essential
+    && apt-get install -y python git build-essential --no-install-recommends \
+    && npm install --production \
+    && apt-get autoremove -y python build-essential \
+    && rm -rf /var/lib/apt/lists/* \
+    && npm cache clear \
+    && rm -rf ~/.node-gyp \
+    && rm -rf /tmp/npm-*
 
 VOLUME ["/usr/src/app/localData","/usr/src/app/localMetadata"]
 
--- a/DockerfileMem
+++ b/DockerfileMem
@@ -6,9 +6,13 @@ WORKDIR /usr/src/app
 COPY . /usr/src/app
 
 RUN apt-get update \
-    && apt-get install -y python git build-essential \
-    && npm install \
-    && apt-get autoremove -y python build-essential
+    && apt-get install -y python git build-essential --no-install-recommends \
+    && npm install --production \
+    && apt-get autoremove -y python build-essential \
+    && rm -rf /var/lib/apt/lists/* \
+    && npm cache clear \
+    && rm -rf ~/.node-gyp \
+    && rm -rf /tmp/npm-*
 
 ENV S3BACKEND mem
 


nicolas2bert:
Hey @michel-kraemer 
Thanks for the contribution.
We are looking at it and will get back to you soon.

nicolas2bert:
When clearing npm cache, why not removing ~/.npm as well?
`rm -rf ~/.npm`

michel-kraemer:
Because it's the same:

    $ docker run -it scality/s3server:mem-latest bash
    $ du -sh ~/.npm
    44M	/root/.npm
    $ npm cache clear
    $ du -sh ~/.npm
    du: cannot access '/root/.npm': No such file or directory

michel-kraemer:
You could delete `~/.node-gyp/` too, by the way, but it's only 3MB.

michel-kraemer:
Some else that just came to my mind: you may save another 4 MB if you create a `.dockerignore` file and put `.git` into it.

michel-kraemer:
I updated the PR to save another 20 MB.

**After 'rm -rf ~/.node-gyp' and adding '.dockerignore'**

     90M	01.tar.xz

**After 'rm -rf /tmp/npm-*'**

     75M	01.tar.xz

I think the only way to make it even smaller now is to use an Alpine base image.

alexandre-merle:
Hey :)

Sounds cool, did you try with ```apt-get install -y python git build-essential --no-install-recommends``` and ```npm install --production``` ? :)

michel-kraemer:
Good call, @alexandre-merle!

**After '--no-install-recommends'**

     71M	01.tar.xz

**After 'npm install --production'**

     66M	01.tar.xz

**After 'apt-get autoremove git'**

     57M	01.tar.xz

I'll update the PR. Please note that sizes on hub.docker.com might differ slightly because they probably use don't use `pxz` to compress the image or call it with different settings.

Is git necessary in production?

DavidPineauScality:
@michel-kraemer Well, git isn't necessarty per-se, but can be useful to confirm which git version is bundled, easing investigations and bug reproduction. I'm not completely against removing it, but keeping it can be useful for such situations.

michael-zapata:
Well, if the `.git`directory isn't bundled in the image, keeping git itself doesn't make a lot of sense. It may be time to care about the 
`package.json` version.

DavidPineauScality:
sure, git and .git go together, whichever solution is adopted :) And I  agree that the version of the package.json needs more care than ever, now that the s3server is gaining in popularity.

michel-kraemer:
IMHO Most of the time you don't need to debug, so git will just waste space. You can easily do `apt-get update && apt-get install -y git` whenever you need it. But that's just my very, very humble opinion.

If you want to know which Git SHA the build was based on we can put it into a file like this:

    git rev-parse HEAD > .gitsha

This could be accomplished by removing the `.dockerignore` file again and adding this command after `apt-get install`. We would need to do `rm -rf .git` too of course to save the space again.

These are just my two cents. Basically, I'm fine with whatever you decide. You find the numbers for everything I tested above. Feel free to include and throw away whatever you see fit. :smile:

DavidPineauScality:
Your suggestion is the most reasonable and answers the various concerns, thanks.

michel-kraemer:
I just tested what happens if I remove the `.dockerignore` file again. The image size increases to 61 MB. If I remove the `.git` directory inside the `Dockerfile` the image size shrinks to 60 MB. This is still 3 MB larger than what I was able to achieve with the `.dockerfile` but this is nitpicking, probably negligible - at least you would be able to get access the Git repo if necessary.

Let me know if I should update the PR or not.

nicolas2bert:
Hey @michel-kraemer, 
For now, we want to keep git for @DavidPineauScality reasons.
 
Also - we have a signing requirement section in our guidelines: https://github.com/scality/Guidelines/blob/master/CONTRIBUTING.md#committing-guidelines
so please update and sign your commits.

Thank you.

michel-kraemer:
@nicolas2bert OK. I updated the PR accordingly and signed the commit.

michel-kraemer:
Let me know if there is anything else I can do to help.

Michel

nicolas2bert:
Hey @michel-kraemer - Can you please drop your last commit and then rebase your branch from master?

michel-kraemer:
Sorry, my bad. I just push the "Update branch" button here on GitHub for the very first time in my life. 😄 I thought it would rebase the branch, but created a merge commit instead. Sorry. I fixed this now.

nicolas2bert:
No worries - Can you please (again) rebase your branch from master so CI tests will pass.

michel-kraemer:
I rebased the branch. I think both checks have passed but GitHub still displays the first one as 'running'.

PR #391, opened 2016-11-14
FIX: Remove useless object creation

alexandre-merle:
There is no need to create a BucketInfo from
another BucketInfo, putting the fromObj in
the in_memory backend fix that.

### DIFF
--- a/lib/metadata/in_memory/backend.js
+++ b/lib/metadata/in_memory/backend.js
@@ -3,6 +3,7 @@ import { errors } from 'arsenal';
 import { markerFilter, prefixFilter } from './bucket_utilities';
 import { ListBucketResult } from './ListBucketResult';
 import getMultipartUploadListing from './getMultipartUploadListing';
+import BucketInfo from '../BucketInfo';
 import { metadata } from './metadata';
 
 const defaultMaxKeys = 1000;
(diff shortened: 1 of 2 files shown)


rahulreddy:
@ironman-machine ...

lib/metadata/in_memory/backend.js
```diff
@@ -41,7 +42,8 @@ const metastore = {
             if (!metadata.buckets.has(bucketName)) {
                 return cb(errors.NoSuchBucket);
             }
-            return cb(null, metadata.buckets.get(bucketName));
+            return cb(null, BucketInfo.fromObj(
```
LaurenSpiegel: why are we returning an object for the bucket if we call getBucketAttributes but a string if we are returning an object and a bucket?  I believe this had to do with some double stringifying in metadata which should be fixed.

rahulreddy: By doing `git blame` I saw that this was added as a backward compatibility for the splitter. Are we not going to support the old splitter usage at all?

alexandre-merle: @LaurenSpiegel the bucketinfo fromobj do nothing in the case of the bucketclient backend actually, because it take an BucketInfo already, and create a new BucketInfo from this one :)
@rahulreddy it should not be relate to old splitter compatibility, i will think about it but basically here, i just move the pparsing of BucketInfo in the memory backend :)

alexandre-merle: @rahulreddy i take a look at the backward PR, if you look, the previous state was to directly do new BucketInfo(), i just replace it by the function fromObj, so its not related to the backward compatibility :)

PR #402, opened 2016-11-15
COMPAT: Return an error for an invalid encoding in getBucket API

bennettbuchanan:
If an encoding type is given (i.e., it is not `undefined`)
and the encoding type is not 'url', then an `InvalidArgument`
error is returned with a message 'Invalid Encoding Method 
specified in Request'.


bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine after rebase onto rel/6.2.1

bennettbuchanan:
end-to-end passed

bennettbuchanan:
@ironman-machine again, please.

PR #410, opened 2016-11-16
FIX: Log caught errors properly

LaurenSpiegel:
Fixes https://github.com/scality/IronMan-S3/issues/411#event-861971864


LaurenSpiegel:
@ironman-machine why not

PR #413, opened 2016-11-17
fix: parsing for kms tests

### LINKED ISSUE #412: kms tests fail when Werelogs output is changed

rahulreddy:
The output parsing was heavily reliant on the order of fields, so a new change in Werelogs
would break the tool. The parsing has been updated to accommodate future changes.
Fix #412

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/copyPart.js
+++ b/tests/functional/aws-node-sdk/test/object/copyPart.js
@@ -18,6 +18,16 @@ const content = 'I am the best content ever';
 const otherAccountBucketUtility = new BucketUtility('lisa', {});
 const otherAccountS3 = otherAccountBucketUtility.s3;
 
+function safeJSONParse(s) {
+    let res;
+    try {
+        res = JSON.parse(s);
+    } catch (e) {
+        return e;
+    }
+    return res;
+}
+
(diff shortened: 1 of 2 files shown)


rahulreddy:
@ironman-machine

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -53,7 +53,7 @@ function createEncryptedBucket(bucketParams, cb) {
     const child = childProcess.spawn(args[0], args)
     .on('exit', () => {
         const hasSucceed = body.join('').split('\n').find(item =>
-            item.startsWith('{"name":"S3","statusCode":200'));
+            item.indexOf('"statusCode":200')) !== -1;
```
ploki: why don't you parse the json output instead of doing string matching?

rahulreddy: I could.

rahulreddy:
Updated the parsing to parse into JSON. I haven't moved them into a common utility as they have different setup and I did not want to get into a refactor zone.

rahulreddy:
@ironman-machine

rahulreddy:
@ironman-machine ...

PR #421, opened 2016-11-19
FIX: Surround response.setHeader with try/catch

### LINKED ISSUE #420: BF: Throw error if non-UTF-8 header values
getObject throws an error when header has non-UTF-8 value.

nicolas2bert:
FIXES #420


nicolas2bert:
@ironman-machine please

PR #422, opened 2016-11-21
Wait for positive backend healthchecks on startup

RachedBenMustapha:
This change makes sure that all backends are working properly before listening for requests on the service port, allowing load balancers to try another server for write requests without introducing non-idempotent retries, because ECONNREFUSED is a clean failure.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -4,6 +4,7 @@ import cluster from 'cluster';
 import arsenal from 'arsenal';
 
 import { logger } from './utilities/logger';
+import { clientCheck } from './utilities/healthcheckHandler';
 import _config from './Config';
 import routes from './routes';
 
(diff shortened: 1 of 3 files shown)


RachedBenMustapha:
@ironman-machine

lib/server.js
```diff
@@ -133,6 +151,6 @@ export default function main() {
         });
     } else {
         const server = new S3Server(cluster.worker);
-        server.startup();
+        server.initiateStartup(logger.newRequestLogger());
```
rahulreddy: Since this is not on request, I recommend using the module logger here.

rahulreddy:
I think it's good to have some MAX_RETRIES eventually, like 10/20/100. This could be an enhancement for later once we know some safe numbers.

lib/server.js
RachedBenMustapha: No this will initiate a series of requests to backends, that explicitly expect a request logger.

rahulreddy: Ah, true. Thanks.

LaurenSpiegel:
@ironman-machine feeling healthy?

PR #424, opened 2016-11-22
fix: pass healthchecks config to Utapi

### LINKED ISSUE #373: Utapi is not receiving healthcheck config
Healthcheck config added in S3's config.json needs to be passed over to Utapi Server as well.

rahulreddy:
fix #373


rahulreddy:
@ironman-machine

PR #426, opened 2016-11-22
Adding an "in production" section

nicolas2bert:


PRODUCTION.md
```diff
@@ -0,0 +1,55 @@
+# Scality S3server in production with Docker
```
nicolas2bert: Using S3 for CI testing is to force hostname and credential using env variables.

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+
+When you start the Scality S3server image, you can adjust the configuration of
+the Scality S3server instance by passing one or more environment variables
+on the docker run command line.
+
+## Environment Variables
+
+### HOST_NAME
+
+This variable specifies an host name:
```
LaurenSpiegel: an s/b a

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+
+`docker run -d --name s3server -p 8000:8000 -e HOST_NAME=new.host.com
+scality/s3server`
+
+### ACCESS_KEY and SECRET_KEY
+
+These variables specify a Scality S3server user credentials named Docker:
```
LaurenSpiegel: These variables specify authentication credentials for an account named "Docker".  You can set credentials for many accounts by editing conf/authdata.json (see below for further info), but if you just want to specify one set of your own, you can use these environment variables.

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+
+`docker run -d --name s3server -p 8000:8000 -e ACCESS_KEY=newAccessKey -e
+SECRET_KEY=newSecretKey scality/s3server`
+
+### LOG_LEVEL
+
+This variable allows you to change the log level: info, debug or trace:
```
LaurenSpiegel: The default is info.  Debug will give you more detailed logs and trace will give you the most detailed.

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+`docker run -d --name s3server -p 8000:8000 -e LOG_LEVEL=trace scality/s3server`
+
+
+# Using Scality S3server in production with Docker
+
+## Using Docker Volume in production
+
+S3Server runs with a file backend by default.
+
+So, by default, the data are stored inside your S3server Docker container.
```
LaurenSpiegel: are s/b is

PRODUCTION.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Using Docker Volume
+
+in production
+
+S3Server runs with a file backend by default.
+
+So, by default, the data are stored inside your S3server Docker container.
+
+However, you **MUST** use Docker volumes to host the data and the metadata of
```
LaurenSpiegel: However, if you want your data and metadata to persist, you **MUST** use Docker volumes to host your data and metadata outside your s3server Docker container.  Otherwise, the data and metadata will be destroyed when you erase the container.

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+the S3server container outside your S3server Docker container.
+Otherwise, data are destroyed after erasing the container.
+
+`docker run -­v $(pwd)/data:/usr/src/app/localData -­v
+$(pwd)/metadata:/usr/src/app/localMetadata -p 8000:8000 ­-d scality/s3server`
```
LaurenSpiegel: explain what this mapping is

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Adding/ modifying/ deleting accounts/ users credentials
+
+1. Create locally a customized `conf/authdata.json`.
+
+2. Use [Docker Volume](https://docs.docker.com/engine/tutorials/dockervolumes/)
+
+to override the default one through a docker file mapping.
```
LaurenSpiegel: to override the default conf/authdata.json through a docker file mapping.  For example:

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+
+`docker run -v $(pwd)/authdata.json:/usr/src/app/conf/authdata.json -p`
```
LaurenSpiegel: don't you need conf here or remove it from point 1 above?

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+`8000:8000 -d scality/s3server`
+
+## Setting a new region
+
+To specify an host name (e.g. s3.domain.name),
```
LaurenSpiegel: an s/b a

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+1. Create locally a customized `conf/authdata.json`.
+
+2. Use [Docker Volume](https://docs.docker.com/engine/tutorials/dockervolumes/)
+
+to override the default one through a docker file mapping.
+
+`docker run -v $(pwd)/authdata.json:/usr/src/app/conf/authdata.json -p`
+`8000:8000 -d scality/s3server`
+
+## Setting a new region
```
LaurenSpiegel: Specifying your own host name

PRODUCTION.md
```diff
@@ -0,0 +1,82 @@
...
+
+To specify an host name (e.g. s3.domain.name),
+you can provide your own
+[config.json](https://github.com/scality/S3/blob/master/config.json)
+using [Docker Volume](https://docs.docker.com/engine/tutorials/dockervolumes/).
+
+First, add a new region to your config.json that you created locally:
```
LaurenSpiegel: First add a new key-value pair in the regions section of your config.json.  The key in the key-value pair should be your "region" name and the value is an array containing any host name you would like to add.

PRODUCTION.md
rahulreddy: I think expanding this would have more impact. Something like
`This variable specifies a host name to serve virtual hosted-style requests`

nicolas2bert: @rahulreddy - thanks for the recommendation.
Just to be sure, we need to specify hostname for every requests (hosted/path style): https://github.com/scality/S3/blob/4236068c5dc3863362a47a0d3af7c86dd1359fd0/lib/utils.js#L34 . Is it correct?

rahulreddy: Yes. Once we have whitelisted the hostname any we can extract the bucket names from host-style requests.

nicolas2bert: Yes @rahulreddy for hosted style but it is also true for path-style since if the host name is not listed it will return `throw new Error('bad request: hostname ${host} is not in valid endpoints');`. I just wanted to be sure.

rahulreddy: You are right. I forgot that we enforce a valid host name even for path-style requests. You can ignore my comment.

LaurenSpiegel:
Do you think it makes sense to add a link to this from the general readme and call this docker readme or something that will be more obvious to a reader since it is not just about production?

PR #427, opened 2016-11-22
FIX: Prevent unlimited setting of event listeners

LaurenSpiegel:
Fixes https://github.com/scality/S3/issues/271

### DIFF
--- a/lib/metadata/bucketfile/backend.js
+++ b/lib/metadata/bucketfile/backend.js
@@ -125,6 +125,14 @@ class BucketFileInterface {
         /* We need to wait for somebody to wake us up either by
            recodone or refdecr */
         let cbDone = false;
+
+        if (this.notifier.listeners('recodone').length > 0 ||
+            this.notifier.listeners('refdecr').length > 0) {
+            setTimeout(() => {
+                this.reConnect(cb);
+            }, 50);
+            return undefined;
+        }
         this.notifier.once('recodone', () => {
             if (cbDone) {
                 return undefined;


LaurenSpiegel:
@vrancurel, this prevents the memory leak warnings.  Thoughts?

GiorgioRegni:
@vrancurel what do you think of this patch?

MathieuCassagne:
@LaurenSpiegel: What about using the ```removeAllListeners``` before putting a new one ?
https://nodejs.org/dist/latest-v4.x/docs/api/events.html#events_emitter_removealllisteners_eventname
I'm not sure what is the cleanest way :s

LaurenSpiegel:
@MathieuCassagne, I tried that any many other permutations.  If I remember correctly, if you removeAllListeners or just limit to one altogether, you end up freezing.  Rahul had a different fix that set the listeners outside the function so they wouldn't keep getting set, but that significantly impacted performance.

rahulreddy:
@MathieuCassagne Removing all listeners would have an adverse affect as there would be more events emitted and less listeners listening to them.
This https://github.com/scality/S3/commit/100226a4205daa9545d73c5dd7b3b1c2b62b6c10 would be the clean way of implementing it but it made the tests run significantly longer.

vrancurel:
The event.emit() is synchronous, causing a 'stack' overflow in case of heavy concurrency. Isn't it enough to just call event.emit() from a process.nextTick() lines 121 and 176 ?

LaurenSpiegel:
@vrancurel, I just tested inserting the process.nextTicks and it does not prevent the memory leak from the addition of the listeners.  Is my change palatable?

LaurenSpiegel:
After discussing with @vrancurel, he approves.  Merging.

PR #428, opened 2016-11-22
DOCKER: using Environment Variables

nicolas2bert:
This PR adds an entrypoint that allows user to:

- Modify Docker user credentials
- Specify an host name 
- Log level

using [Docker environment variables](https://docs.docker.com/engine/reference/run/#/env-environment-variables).

**Docker command:**
` docker run -d --name s3server -p 8000:8000 -e HOST_NAME=new.host.com -e ACCESS_KEY=newAccessKey -e SECRET_KEY=newSecretKey scality/s3server`

If you want to try these changes, you will have to build an image from this branch 
`docker build ./ -t s3` and run the container.

### DIFF
--- a/conf/authdata.json
+++ b/conf/authdata.json
@@ -28,5 +28,15 @@
             "access": "accessKey2",
             "secret": "verySecretKey2"
         }]
+    }, {
+        "name": "Docker",
+        "email": "sampleaccount3@sampling.com",
+        "arn": "aws::iam:accessKeyDocker:user/Docker",
+        "canonicalID": "sd359df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47eh3hd",
+        "shortid": "123456789012",
+        "keys": [{
+            "access": "accessKeyDocker",
+            "secret": "verySecretKeyDocker"
+        }]
     }]
 }
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# set -e stops the execution of a script if a command or pipeline has an error
+set -e
+
+if [[ "$ACCESS_KEY" && "$SECRET_KEY" ]]; then
+    sed -i "s/accessKeyDocker/$ACCESS_KEY/" ./conf/authdata.json
+    sed -i "s/verySecretKeyDocker/$SECRET_KEY/" ./conf/authdata.json
+    echo "Access key and secret key have been modified successfully"
+fi
+
+if [[ "$HOST_NAME" ]]; then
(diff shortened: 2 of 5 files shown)


docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# set -e stops the execution of a script if a command or pipeline has an error
+set -e
+
+if [[ "$ACCESS_KEY" && "$SECRET_KEY" ]]; then
+    sed -i "s/accessKeyDocker/$ACCESS_KEY/" ./conf/authdata.json
+    sed -i "s/verySecretKeyDocker/$SECRET_KEY/" ./conf/authdata.json
+    echo "Access key and secret key have been modified successfuly"
```
bennettbuchanan: Minor: typo on 'successfuly'

docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# set -e stops the execution of a script if a command or pipeline has an error
+set -e
+
+if [[ "$ACCESS_KEY" && "$SECRET_KEY" ]]; then
+    sed -i "s/accessKey1/$ACCESS_KEY/" ./conf/authdata.json
+    sed -i "s/verySecretKey1/$SECRET_KEY/" ./conf/authdata.json
+    echo "Access key and secret key have been modified successfuly"
```
bennettbuchanan: Minor: typo on 'successfuly'

conf/authdata.json
```diff
@@ -15,7 +15,7 @@
             "arn": "aws::iam:123456789013:bart",
             "keys": [{
                 "access": "USERBARTFUNACCESSKEY",
-                "secret": "verySecretKey1"
+                "secret": "verySecretKeyUser1"
```
rahulreddy: This would mean that if any user in open-source was using this credential, it will stop working. It will be nice to add this to the release notes when you push a new Docker open-source docker image.

rahulreddy: This would mean that if any user in open-source was using this credential, it will stop working. It will be nice to add this to the release notes when you push a new Docker open-source docker image.

nicolas2bert: You are right - As I created a new user called Docker, I do not even need to change it. Thanks

docker-entrypoint.sh
nicolas2bert: Thanks :)

conf/authdata.json
rahulreddy: Ah, cool. This is better.

LaurenSpiegel:
@ironman-machine just to be sure

PR #113, opened 2016-08-02
Ft/server side encryption

ploki:
serverside encryption is triggered by bucket flag for the moment.

PR #431, opened 2016-11-24
Add functions to bucket- and vaultclient for deep healthcheck

dora-korpar:
Title says it all!
Depends on vaultclient pr #113 https://github.com/scality/vaultclient/pull/113
and bucketclient pr #91 https://github.com/scality/bucketclient/pull/91

### DIFF
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -13,10 +13,10 @@ let implName;
 
 if (config.backends.data === 'mem') {
     client = inMemory;
-    implName = 'mem';
+    implName = 'dataMem';
 } else if (config.backends.data === 'file') {
     client = file;
-    implName = 'file';
+    implName = 'dataFile';
 } else if (config.backends.data === 'scality') {
     client = new Sproxy({
         bootstrap: config.sproxyd.bootstrap,
@@ -162,8 +162,9 @@ const data = {
                 respBody[implName] = {
                     error: err,
                 };
-                // respBody also returned so error is written to response
-                return cb(err, respBody);
+                // error returned as null so async parallel doesn't return
+                // before all backends are checked
+                return cb(null, respBody);
             }
             respBody[implName] = {
                 code: result.statusCode,
--- a/lib/metadata/wrapper.js
+++ b/lib/metadata/wrapper.js
@@ -147,6 +147,15 @@ const metadata = {
     switch: newClient => {
         client = newClient;
     },
+
+    checkHealth: (log, cb) => {
+        if (!client.checkHealth) {
+            const defResp = {};
+            defResp[implName] = { code: 200, message: 'OK' };
+            return cb(null, defResp);
+        }
+        return client.checkHealth(implName, log, cb);
+    },
 };
 
 export default metadata;
--- a/lib/utilities/healthcheckHandler.js
+++ b/lib/utilities/healthcheckHandler.js
@@ -45,16 +45,25 @@ export function clientCheck(log, cb) {
 }
 
 function routeHandler(deep, req, res, log, statsClient, cb) {
-    if (isHealthy()) {
-        if (req.method === 'GET' || req.method === 'POST') {
-            if (deep) {
-                return clientCheck(log, cb);
-            }
-            return statsClient.getStats(log, cb);
-        }
+    if (!isHealthy()) {
+        return cb(errors.InternalError, []);
+    }
+    if (req.method !== 'GET' && req.method !== 'POST') {
         return cb(errors.BadRequest, []);
     }
-    return cb(errors.InternalError, []);
+    if (!deep) {
+        return statsClient.getStats(log, cb);
+    }
+    let fail = false;
+    return clientCheck(log, (err, results) => {
+        // makes S3 return 500 error if any backend fails
+        const obj = results.reduce((obj, item) => Object.assign(obj, item), {});
+        fail = Object.keys(obj).some(k => obj[k].error);
+        if (fail) {
+            return cb(errors.InternalError, obj);
+        }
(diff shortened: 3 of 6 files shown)


lib/metadata/wrapper.js
```diff
@@ -147,6 +147,31 @@ const metadata = {
...
+            const respBody = {};
+            if (err) {
+                log.error(`error from ${implName}`, { error: err });
+                respBody[implName] = {
+                    error: err,
+                };
+                // respBody also returned so error is written to response
+                return cb(err, respBody);
+            }
+            const parseResult = JSON.parse(result);
```
LaurenSpiegel: the json.parse should just be happening in bucketclient/backend.js instead of twice.  Actually, if we're only going to do the healthcheck for the bucketclient backend (rather than bucketfile and mem), this function should check whether the client supports the function and return as you have above and if it does support the function, it must be bucketclient which is filtered through bucketclient/backend.js so can just pass through the object from there.

dora-korpar:
@ironman-machine look alive, Jarvis

LaurenSpiegel:
Are tests being written in integration so we know the pieces are working together?

dora-korpar:
@LaurenSpiegel yep! That's the next step

rahulreddy:
As discussed, let's change the behavior a bit to accommodate errors. This will avoid extra steps to setup a default/regular response.

1.  Add the intended behavior to Healthchecks.md doc
2. Change backend/wrapper to not set err param in the callbacks
This way we can leverage async.parallel to call all backends even though some of them may fail      and we can rely on the response body for error or success.
3. Change the logic in `healthcheckHandler.js#routeHandler` to use a "fail" flag, something like
```js
if (deep) {
    let fail = false;
    return clientCheck(log, (err, results) => {
        fail = results.some(item => item.error);
        if (fail) {
            // results contain response body' for each component even for errors
            return cb(errors.InternalError, results);
        }
        return cb(null, results);
    });
}
```

dora-korpar:
@ironman-machine jump

dora-korpar:
@ironman-machine try again

dora-korpar:
End2end was rebuilt and passed here: http://ci.ironmann.io/gh/scality/S3/2729

dora-korpar:
@ironman-machine one mo time

dora-korpar:
@ironman-machine no glitches this time, plz

lib/utilities/healthcheckHandler.js
```diff
@@ -45,16 +45,26 @@ export function clientCheck(log, cb) {
...
     }
-    return cb(errors.InternalError, []);
+    if (!deep) {
+        return statsClient.getStats(log, cb);
+    }
+    let fail = false;
+    return clientCheck(log, (err, results) => {
+        // makes S3 return 500 error if any backend fails
+        const obj = Object.assign.apply(
+            Object, [{}].concat(results));
```
LaurenSpiegel: What does having Object as the first argument of apply here do?  If you are using the approach of taking the Object.assign method and applying it to the array that follows, wouldn't null as the first argument do the job?

rahulreddy: This was my idea. It converts an array of objects into an object.
Originally the intention was to use Object.assign with spread operator 
```js
Object.assign({}, ...arr)
```
but it's not supported. So this is the "babel transpiled" version of that.

LaurenSpiegel: doesn't null as the first argument work the same (Object is not actually doing anything as the first argument)?  couldn't you do this with reduce?

rahulreddy: null works
```js
 Object.assign.apply(null, [{}].concat(results));
```
reduce works
```js
results.reduce((obj, item) => Object.assign(obj, item), {})
```
I just used the implementation per transpiling here http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=latest%2Creact%2Cstage-2&experimental=false&loose=false&spec=false&code=const%20arr%20%3D%20%5B%7B%20a%3A%201%7D%2C%20%7B%20b%3A%202%7D%2C%20%7Bc%3A%203%7D%5D%3B%0AObject.assign(%7B%7D%2C%20...arr)%0A&playground=true

LaurenSpiegel: thanks.  i vote reduce.

lib/data/wrapper.js
```diff
@@ -158,14 +158,15 @@ const data = {
         client.healthcheck(log, (err, result) => {
             const respBody = {};
             if (err) {
-                log.error(`error from ${implName}`, { error: err });
-                respBody[implName] = {
+                log.error('error from sproxyd', { error: err });
```
dora-korpar: Because 'implName' can be 'mem' or 'file', which wasn't descriptive enough in the response. If the other backends are running in mem as well, then you wouldn't be able to tell which response came from which backend.

lib/utilities/healthcheckHandler.js
rahulreddy: Ok, let's use reduce.

lib/data/wrapper.js
LaurenSpiegel: you could change 'mem' and 'file' in lines 16 and 19 above to 'dataMem' and 'dataFile'

dora-korpar: ok! I'll do that for the other backends as well, then

dora-korpar:
@ironman-machine gogogogo

dora-korpar:
@ironman-machine one last time

LaurenSpiegel:
end to end passed http://ci.ironmann.io/gh/scality/S3/2780

PR #432, opened 2016-11-29
FT: Add putBucketWebsite to API

electrachong:
Depends on
scality/Arsenal#196,
scality/utapi#67, &
scality/Integration#321

Call action for ?website query in routePUT
* Add putBucketWebsite to shielded actions
* Update BucketInfo modelVersion to 4: include website configuration info
* Parse and validate xml WebsiteConfiguration body
* Send API metrics to utapi & deliver success or err response

### DIFF
--- a/lib/api/apiUtils/bucket/bucketWebsite.js
+++ b/lib/api/apiUtils/bucket/bucketWebsite.js
@@ -1,0 +1,364 @@
+import { parseString } from 'xml2js';
+
+import { errors } from 'arsenal';
+import {
+    WebsiteConfiguration,
+} from '../../../metadata/WebsiteConfiguration';
+/*
+   Format of xml request:
+
+   <WebsiteConfiguration xmlns='http://s3.amazonaws.com/doc/2006-03-01/'>
+     <IndexDocument>
+       <Suffix>index.html</Suffix>
+     </IndexDocument>
+     <ErrorDocument>
+       <Key>Error.html</Key>
+     </ErrorDocument>
+
+     <RoutingRules>
+       <RoutingRule>
+       <Condition>
+         <KeyPrefixEquals>docs/</KeyPrefixEquals>
+       </Condition>
+       <Redirect>
+         <ReplaceKeyPrefixWith>documents/</ReplaceKeyPrefixWith>
+       </Redirect>
+       </RoutingRule>
+       ...
+    </RoutingRules>
+   </WebsiteConfiguration>
+   */
+
+
+// Key names of redirect object values to check if are valid strings
+const redirectValuesToCheck = ['HostName', 'ReplaceKeyPrefixWith',
+    'ReplaceKeyWith'];
+
+/** Helper function for validating format of parsed xml element
+* @param {array} elem - element to check
+* @return {boolean} true / false - elem meets expected format
+*/
+function _isValidElem(elem) {
+    return (Array.isArray(elem) && elem.length === 1);
+}
+
+/** Check if parsed xml element contains a specified child element
+* @param {array} parent - represents xml element to check for child element
+* @param {(string|string[])} requiredElem - name of child element(s)
+* @param {object} [options] - specify additional options
+* @param {boolean} [isList] - indicates if parent is list of children elements,
+* used only in conjunction a singular requiredElem argument
+* @param {boolean} [checkForAll] - return true only if parent element contains
+* all children elements specified in requiredElem; by default, returns true if
+* parent element contains at least one
+* @param {boolean} [validateParent] - validate format of parent element
+* @return {boolean} true / false - if parsed xml element contains child
+*/
+export function xmlContainsElem(parent, requiredElem, options) {
+    // Non-top level xml is parsed into object in the following manner.
+
+    // Example: <Parent><requiredElem>value</requiredElem>
+    //          <anotherElem>value2</anotherElem></Parent>
+    // Result: { Parent: [{ requiredElem: ["value"],
+    //            anotherElem: ["value2"] }] }
+
+    // Example for xml list:
+    //  <ParentList>
+    //      <requiredElem>value</requiredElem>
+    //      <requiredElem>value</requiredElem>
+    //      <requiredElem>value</requiredElem>
+    //  </ParentList>
+    // Result: { ParentList: [{ requiredElem: ['value', 'value', 'value'] }] }
+
+    const isList = options ? options.isList : false;
+    const checkForAll = options ? options.checkForAll : false;
+    // true by default, validateParent only designated as false when
+    // parent was validated in previous check
+    const validateParent = (options && options.validateParent !== undefined) ?
+        options.validateParent : true;
+
+    if (validateParent && !_isValidElem(parent)) {
+        return false;
+    }
+    if (Array.isArray(requiredElem)) {
+        if (checkForAll) {
+            return requiredElem.every(elem => _isValidElem(parent[0][elem]));
+        }
+        return requiredElem.some(elem => _isValidElem(parent[0][elem]));
+    }
+    if (isList) {
+        if (!Array.isArray(parent[0][requiredElem]) ||
+        parent[0][requiredElem].length === 0) {
+            return false;
+        }
+    } else {
+        return _isValidElem(parent[0][requiredElem]);
+    }
+
+    return true;
+}
+
+
+/** Validate XML, returning an error if any part is not valid
+* @param {object} parsingResult - object parsed from xml to be validated
+* @param {object[]} parsingResult.IndexDocument -
+*   Required if RedirectAllRequestsTo is not defined
+* @param {string[]} parsingResult.IndexDocument[].Suffix -
+*   Key that specifies object in bucket to serve as index
+* @param {object[]=} parsingResult.ErrorDocument - Optional
+* @param {string[]} parsingResult.ErrorDocument[].Key -
+*   Key that specifies object in bucket to serve as error document.
+* @param {object[]} parsingResult.RedirectAllRequestsTo -
+*   Contains fields to specify how to redirect all requests to bucket.
+* @param {string[]} parsingResult.RedirectAllRequestsTo[].HostName -
+*   Hostname to use when redirecting all requests.
+* @param {string[]=} parsingResult.RedirectAllRequestsTo[].Protocol -
+*   Optional, protocol to use when redirecting all request ('http' or 'https')
+* @param {object[]=} parsingResult.RoutingRules -
+*   Contains list of rules for redirecting specific requests.
+* @param {object[]} parsingResult.RoutingRules[].RoutingRule -
+*   Contains redirect and condition information for specific redirects.
+* @param {object[]} parsingResult.RoutingRules[].RoutingRule[].Redirect -
+*   Contains information for how to redirect, required for a RoutingRule.
+* @param {string[]=} parsingResult.RoutingRules[].RoutingRule[].Redirect[]
+*   .Protocol - Protocol to use for a specific redirect.
+* @param {string[]=} parsingResult.RoutingRules[].RoutingRule[].Redirect[]
+*   .HostName - Hostname to use in a specific redirect.
+* @param {string[]=} parsingResult.RoutingRules[].RoutingRule[].Redirect[]
+*   .ReplaceKeyPrefixWith - What to replace a key prefix specified in a
+*   Condition with in a redirect
+* @param {string[]=} parsingResult.RoutingRules[].RoutingRule[].Redirect[]
+*   .ReplaceKeyWith - What to replace the object key with in a redirect
+* @param {string[]=} parsingResult.RoutingRules[].RoutingRule[].Redirect[]
+*   .HttpRedirectCode - Http code (301-399) to use in redirect.
+* @param {object[]=} parsingResult.RoutingRules[].RoutingRule[].Condition
+    - Optional, contains fields for conditions to make a specific redirect
+* @param {string[]=} parsingResult.RoutingRules[].RoutingRule[]
+*   .Condition[].KeyPrefixEquals - Specify the prefix to match for a redirect
+* @param {string[]=} parsingResult.RoutingRules[].RoutingRule[]
+*   .Condition[].HttpErrorCodeReturnedEquals - Error code to match for redirect
+* @return {(Error|WebsiteConfiguration)} return WebsiteConfiguration on success;
+*   otherwise return error
+*/
+function _validateWebsiteConfigXml(parsingResult) {
+    const websiteConfig = new WebsiteConfiguration();
+    let errMsg;
+
+    function _isValidString(value) {
+        return (typeof value === 'string' && value !== '');
+    }
+
+    if (!parsingResult.IndexDocument && !parsingResult.RedirectAllRequestsTo) {
+        errMsg = 'Value for IndexDocument Suffix must be provided if ' +
+        'RedirectAllRequestsTo is empty';
+        return errors.InvalidArgument.customizeDescription(errMsg);
+    }
+
+    if (parsingResult.RedirectAllRequestsTo) {
+        const parent = parsingResult.RedirectAllRequestsTo;
+        const redirectAllObj = {};
+        if (parsingResult.IndexDocument || parsingResult.ErrorDocument ||
+        parsingResult.RoutingRules) {
+            errMsg = 'RedirectAllRequestsTo cannot be provided in ' +
+            'conjunction with other Routing Rules.';
+            return errors.InvalidRequest.customizeDescription(errMsg);
+        }
+        if (!xmlContainsElem(parent, 'HostName')) {
+            errMsg = 'RedirectAllRequestsTo not well-formed';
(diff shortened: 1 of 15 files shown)


electrachong:
@ironman-machine 1st try
SCALITY_INTEGRATION_BRANCH=ft/bucketPutWebsite
SCALITY_ARSENAL_BRANCH=ft/bucketPutWebsite
SCALITY_UTAPI_BRANCH=ft/bucketPutWebsite

electrachong:
@ironman-machine 2nd try
SCALITY_INTEGRATION_BRANCH=ft/bucketPutWebsite
SCALITY_ARSENAL_BRANCH=ft/bucketPutWebsite
SCALITY_UTAPI_BRANCH=ft/bucketPutWebsite

electrachong:
end-to-end passed here: http://ci.ironmann.io/gh/scality/Integration/6959

lib/metadata/BucketInfo.js
```diff
@@ -30,10 +31,43 @@ export default class BucketInfo {
...
+    .httpErrorCodeReturnedEquals - http err code under which to apply redirect
+    * @param {string} websiteConfiguration.routingRules[].condition
+    .keyPrefixEquals - object key name prefix when the redirect is applied
     */
     constructor(name, owner, ownerDisplayName, creationDate,
                 mdBucketModelVersion, acl, transient, deleted,
-                serverSideEncryption, versioningConfiguration) {
+                serverSideEncryption, versioningConfiguration,
+                websiteConfiguration) {
+        let websiteConfig = websiteConfiguration;
```
michael-zapata: You can just assign to `null` here, avoiding the `||` assignment at the end.

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+        websiteConfig.setIndexDocument(parent[0].Suffix[0]);
+    }
+
+    if (resultConfig.ErrorDocument) {
+        const parent = resultConfig.ErrorDocument;
+        if (!_xmlContainsElem(parent, 'Key')) {
+            errMsg = 'ErrorDocument is not well-formed';
+            return errors.MalformedXML.customizeDescription(errMsg);
+        }
+        if (typeof parent[0].Key[0] !== 'string'
```
bennettbuchanan: I see this type of check `typeof foo !== 'string' || foo === ''` occurring frequently. Perhaps you could extract it into some find of utility check function that returns a boolean.

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+                errMsg = 'RoutingRule requires Redirect, which is ' +
+                'missing or not well-formed';
+                return errors.MalformedXML.customizeDescription(errMsg);
+            }
+            // looks like AWS doesn't actually make this check:
+            if (!(_xmlContainsElem(rule.Redirect, 'Protocol')
+            || _xmlContainsElem(rule.Redirect, 'HostName')
+            || _xmlContainsElem(rule.Redirect, 'ReplaceKeyPrefixWith')
+            || _xmlContainsElem(rule.Redirect, 'ReplaceKeyWith')
+            || _xmlContainsElem(rule.Redirect, 'HttpRedirectCode'))) {
```
bennettbuchanan: Could `_xmlContainsElem` be updated to handle multiple required elements (e.g., an array of required elements)? That might simplify this kind of check.

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+                }
+                routingRule.setCondition(conditionObj);
+            }
+            websiteConfig.addRoutingRule(routingRule);
+        }
+    }
+    return websiteConfig;
+}
+
+export function parseWebsiteConfigXml(xml, log, next) {
```
bennettbuchanan: I understand that this function is used along with `async.waterfall`, but I think something like `cb` would be a  clearer parameter name in this context.

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+       </Redirect>
+       </RoutingRule>
+       ...
+    </RoutingRules>
+   </WebsiteConfiguration>
+   */
+
+
+/** Check if parsed xml element contains a specified child element
+* @param {obj} obj - represents xml element to check for child element
```
electrachong: Thanks, you're right

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+        }
+    } else if (obj[0][requiredElem].length === 0) {
+        return false;
+    }
+
+    return true;
+}
+
+/** Validate XML, returning an error if any part is not valid
+* @param {object} jsonResult - website configuration xml parsed into JSON
```
electrachong: I added the expected fields of the parameter

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+* @return {Error} on error, {WebsiteConfiguration} on success
+*/
+function _validateWebsiteConfigXml(jsonResult) {
```
electrachong: I can split them up into smaller functions correlating to xml elements. I agree it's an `if` forest. :P Can you describe what you're imagining for a state-machine? Could that be managed by an object mapping to functions? How to get around using all the `if` statements?

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+        }
+        for (let i = 0; i < parent[0].RoutingRule.length; i++) {
+            const rule = parent[0].RoutingRule[i];
+            const redirectObj = {};
+            if (!(Array.isArray(rule.Redirect) && rule.Redirect.length === 1)) {
+                errMsg = 'RoutingRule requires Redirect, which is ' +
+                'missing or not well-formed';
+                return errors.MalformedXML.customizeDescription(errMsg);
+            }
+            // looks like AWS doesn't actually make this check:
```
electrachong: I added to the comment to be more explicit about the reason.

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,275 @@
...
+            for (let j = 0; j < valuesToCheck.length; j++) {
+                const elem = valuesToCheck[j];
+                if (rule.Redirect[0][elem]) {
+                    const value = rule.Redirect[0][elem][0];
+                    if (typeof value !== 'string' || value === '') {
+                        errMsg = `Redirect ${elem} is not well-formed`;
+                        return errors.InvalidArgument
+                            .customizeDescription(errMsg);
+                    }
+                    const objKey = elem.charAt(0).toLowerCase() + elem.slice(1);
```
electrachong: OK

lib/metadata/BucketInfo.js
electrachong: You're right. I fixed it.

lib/metadata/BucketInfo.js
```diff
@@ -211,6 +263,27 @@ export default class BucketInfo {
...
+     */
+    getWebsiteConfiguration() {
+        return this._websiteConfiguration;
+    }
+    /**
+     * Set website configuration information
+     * @param {object} websiteConfiguration - configuration for bucket website
+     * @return {BucketInfo} - bucket info instance
+     */
+    setWebsiteConfiguration(websiteConfiguration) {
```
electrachong: OK, I removed the check if it's an instanceof. As for the function itself, it's used to set the websiteConfiguration of the bucket before updating the bucket in MD.

electrachong:
@michael-zapata Thanks for the review!!

lib/api/bucketPutWebsite.js
```diff
@@ -0,0 +1,60 @@
...
+            if (request.post) {
+                log.trace('parsing website configuration from request body');
+                return parseWebsiteConfigXml(request.post, log,
+                    (err, config) => next(err, config));
+            }
+            // If no request body at all
+            return next(errors.MissingRequestBodyError);
+        },
+        function validateBucketAuthorization(config, next) {
+            services.metadataValidateAuthorization(metadataValParams,
```
rahulreddy: Instead of using `services.metadataValidateAuthorization` I would recommend calling the logic needed inside it directly, as we are planning to phase out that method with refactors for other APIs.
You can look at https://github.com/scality/S3/pull/434/files for reference.

lib/metadata/BucketInfo.js
```diff
@@ -68,6 +102,22 @@ export default class BucketInfo {
...
+                routingRules } = websiteConfiguration;
+            assert(indexDocument === undefined ||
+                typeof indexDocument === 'string');
+            assert(errorDocument === undefined ||
+                typeof errorDocument === 'string');
+            assert(redirectAllRequestsTo === undefined ||
+                typeof redirectAllRequestsTo === 'object');
+            assert(routingRules === undefined ||
+                Array.isArray(routingRules));
+            websiteConfig = websiteConfiguration instanceof
```
rahulreddy: I don't see the need of fallback, asserting `websiteConfiguration instanceof WebsiteConfiguration` should be enough.

lib/metadata/ModelVersion.md
```diff
@@ -39,6 +39,14 @@ Used to determine which splitter to use ( < 2 means old splitter)
...
+
+```
+this._websiteConfiguration = websiteConfiguration || null;
+```
+
 ### Usage
 
-Used to store the server bucket encryption info
\ No newline at end of file
+Used to store the server bucket encryption info
```
rahulreddy: Used to store the ~~server bucket encryption~~ bucket website configuratio info

lib/metadata/BucketInfo.js
LaurenSpiegel: I think this is needed when we get the json string from metadata.

rahulreddy: Ah, I see. In that case, IMHO, the instance of check should happen before the assertions and use getters instead of accessing the props. directly. Otherwise are just stringifying WebsiteConfiguration instance in BucketInfo.serialize() method, which nullifies the purpose of the class.

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -30,32 +30,75 @@ import {
...
+    //  <ParentList>
+    //      <requiredElem>value</requiredElem>
+    //      <requiredElem>value</requiredElem>
+    //      <requiredElem>value</requiredElem>
+    //  </ParentList>
+    // Result: { ParentList: [{ requiredElem: ['value', 'value', 'value'] }] }
+    const isList = options ? options.isList : false;
+    const checkForAll = options ? options.checkForAll : false;
+
+    function _elemExists(elem) {
```
michael-zapata: Why not extract this function instead of allocating this closure every time ?

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -30,32 +30,75 @@ import {
...
-        if (parent[0][requiredElem].length !== 1) {
+    if (Array.isArray(requiredElem)) {
+        if (checkForAll) {
+            for (let i = 0; i < requiredElem.length; i++) {
+                if (!_elemExists(requiredElem[i])) {
+                    return false;
+                }
+            }
+        } else {
+            for (let i = 0; i < requiredElem.length; i++) {
```
michael-zapata: You can use `.some()`

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,317 @@
...
+
+/** Check if parsed xml element contains a specified child element
+* @param {array} parent - represents xml element to check for child element
+* @param {string} requiredElem - name of child element
+* @param {boolean} [isList] - indicates if parent is list of children elements
+* @return {boolean} true / false - if parsed xml element contains child
+*/
+function _xmlContainsElem(parent, requiredElem, isList) {
+    if (!Array.isArray(parent)
+    || parent.length !== 1
```
electrachong: I added a comment to make expected structure a little more clear. Might be too much detail.. I added a few unit tests, please let me know if it fits what you had in mind.

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,317 @@
...
+                redirectObj.protocol = rule.Redirect[0].Protocol[0];
+            }
+            if (rule.Redirect[0].HttpRedirectCode) {
+                const code = parseInt(rule.Redirect[0].HttpRedirectCode[0], 10);
+                if (isNaN(code)) {
+                    errMsg = 'The provided HTTP redirect code is not valid. ' +
+                    'It should be a string containing a number.';
+                    return errors.MalformedXML.customizeDescription(errMsg);
+                }
+                if (!(code > 300 && code < 400)) {
```
electrachong: I tested AWS responses a bit, they return an error if the code is not in the right range:
<Error><Code>InvalidRequest</Code><Message>The provided HTTP redirect code (403) is not valid. Valid codes are 3XX except 300.</Message> [...] </Error>

lib/metadata/ModelVersion.md
```diff
@@ -39,6 +39,14 @@ Used to determine which splitter to use ( < 2 means old splitter)
...
+
+### Properties Added
+
+```
+this._websiteConfiguration = websiteConfiguration || null;
+```
+
 ### Usage
 
-Used to store the server bucket encryption info
```
electrachong: Thanks, I added it back

lib/metadata/WebsiteConfiguration.js
```diff
@@ -0,0 +1,202 @@
...
+    /**
+    * Set the condition object
+    * @param {object} obj - object to set as condition
+    * @return {undefined};
+    */
+    setCondition(obj) {
+        if (obj) {
+            const { keyPrefixEquals, httpErrorCodeReturnedEquals } = obj;
+            this.condition = { keyPrefixEquals, httpErrorCodeReturnedEquals };
+        } else {
```
electrachong: Turns out I did use the setters to create the RoutingRules before adding them to the WebsiteConfiguration instance while parsing, but I removed them.

lib/api/apiUtils/bucket/bucketWebsite.js
electrachong: I extracted the check to a new function

lib/api/apiUtils/bucket/bucketWebsite.js
electrachong: I extended xmlContainsElem to handle multiple elements, but the logic of the function unfortunately got a bit more complicated. :sweat_smile:

lib/api/apiUtils/bucket/bucketWebsite.js
electrachong: I agree and changed it

electrachong:
@ironman-machine please check again
SCALITY_INTEGRATION_BRANCH=ft/bucketPutWebsite
SCALITY_ARSENAL_BRANCH=ft/bucketPutWebsite
SCALITY_UTAPI_BRANCH=ft/bucketPutWebsite

lib/api/bucketPutWebsite.js
electrachong: :thumbsup:

lib/metadata/BucketInfo.js
electrachong: I've refactored the class according to your suggestion

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -0,0 +1,364 @@
...
+    //      <requiredElem>value</requiredElem>
+    //      <requiredElem>value</requiredElem>
+    //  </ParentList>
+    // Result: { ParentList: [{ requiredElem: ['value', 'value', 'value'] }] }
+
+    const isList = options ? options.isList : false;
+    const checkForAll = options ? options.checkForAll : false;
+    // true by default, validateParent only designated as false when
+    // parent was validated in previous check
+    const validateParent = (options && options.validateParent) ?
```
electrachong: lol you're right :cry:

electrachong:
end-to-end passed again here: http://ci.ironmann.io/gh/scality/Integration/7228

lib/metadata/BucketInfo.js
```diff
@@ -116,10 +138,12 @@ export default class BucketInfo {
      */
     static deSerialize(stringBucket) {
         const obj = JSON.parse(stringBucket);
+        const websiteConfig =
```
LaurenSpiegel: Please change to:

`` const websiteConfig = obj.websiteConfiguration ?
 new WebsiteConfiguration(obj.websiteConfiguration) : null;``

otherwise, we get different behavior from the file backend versus in memory since deserialized is called on the file backend and this creates a website config object whereas in memory has null.

electrachong: Thanks for pointing that out!

PR #434, opened 2016-11-29
FIX: Refactor objectPutPart API

### LINKED ISSUE #153: Refactor objectPutPart API
Metadata is being called twice to get the MPU shadow bucket metadata (once to get the MPU bucket metadata with "services.getMPUBucket" and then again when validating authorization with "services.metadataValidateMultipart").

bennettbuchanan:
Fixes refactoring part of #153

* Call `metadata.getBucket` directly and perform validation checks in the API.
* Remove calls to `services.metadataValidateAuthorization`, `services.metadataValidateMultipart`, and `services.getMPUBucket`.
* Remove `doStore` function and instead call `services.dataStore` and `services.metadataStorePart` directly in API.
* Remove extraneous code after the above changes.

### DIFF
--- a/lib/api/apiUtils/object/storeObject.js
+++ b/lib/api/apiUtils/object/storeObject.js
@@ -1,0 +1,102 @@
+import { errors } from 'arsenal';
+import V4Transform from '../../../auth/streamingV4/V4Transform';
+import data from '../../../data/wrapper';
+
+/**
+ * Prepares the stream if the chunks are sent in a v4 Auth request
+ * @param {object} stream - stream containing the data
+ * @param {object | null } streamingV4Params - if v4 auth, object containing
+ * accessKey, signatureFromRequest, region, scopeDate, timestamp, and
+ * credentialScope (to be used for streaming v4 auth if applicable)
+ * @param {RequestLogger} log - the current request logger
+ * @param {function} cb - callback containing the result for V4Transform
+ * @return {object} - V4Transform object if v4 Auth request, or else the stream
+ */
+function prepareStream(stream, streamingV4Params, log, cb) {
+    if (stream.headers['x-amz-content-sha256'] ===
+        'STREAMING-AWS4-HMAC-SHA256-PAYLOAD') {
+        const v4Transform = new V4Transform(streamingV4Params, log, cb);
+        stream.pipe(v4Transform);
+        return v4Transform;
+    }
+    return stream;
+}
+
+/**
+ * Check that `hashedStream.completedHash` matches header `stream.contentMD5`
+ * and delete old data or remove 'hashed' listeners, if applicable.
+ * @param {object} stream - stream containing the data
+ * @param {object} hashedStream - instance of MD5Sum
+ * @param {object} dataRetrievalInfo - object containing the keys of stored data
+ * @param {number} dataRetrievalInfo.key - key of the stored data
+ * @param {string} dataRetrievalInfo.dataStoreName - the implName of the data
+ * @param {object} log - request logger instance
+ * @param {function} cb - callback to send error or move to next task
+ * @return {function} - calls callback with arguments:
+ * error, dataRetrievalInfo, and completedHash (if any)
+ */
+function checkHashMatchMD5(stream, hashedStream, dataRetrievalInfo, log, cb) {
+    const contentMD5 = stream.contentMD5;
+    const completedHash = hashedStream.completedHash;
+    if (contentMD5 && completedHash && contentMD5 !== completedHash) {
+        log.debug('contentMD5 and completedHash do not match', {
+            method: 'storeObject::dataStore',
+            completedHash,
+            contentMD5,
+        });
+        log.trace('contentMD5 does not match, deleting data');
+        data.batchDelete(dataRetrievalInfo, log);
+        return cb(errors.BadDigest);
+    }
+    if (completedHash) {
+        hashedStream.removeAllListeners('hashed');
+    }
+    return cb(null, dataRetrievalInfo, completedHash);
+}
+
+/**
+ * Stores object and responds back with location and storage type
+ * @param {object} objectContext - object's keyContext for sproxyd Key
+ * computation (put API)
+ * @param {object} cipherBundle - cipher bundle that encrypt the data
+ * @param {object} stream - the stream containing the data
+ * @param {number} size - data size in the stream
+ * @param {object | null } streamingV4Params - if v4 auth, object containing
+ * accessKey, signatureFromRequest, region, scopeDate, timestamp, and
+ * credentialScope (to be used for streaming v4 auth if applicable)
+ * @param {RequestLogger} log - the current stream logger
+ * @param {function} cb - callback containing result for the next task
+ * @return {undefined}
+ */
+export function dataStore(objectContext, cipherBundle, stream, size,
+    streamingV4Params, log, cb) {
+    const dataStream = prepareStream(stream, streamingV4Params, log, cb);
+    data.put(cipherBundle, dataStream, size, objectContext, log,
+        (err, dataRetrievalInfo, hashedStream) => {
+            if (err) {
+                log.error('error in datastore', {
+                    error: err,
+                });
+                return cb(err);
+            }
+            if (!dataRetrievalInfo) {
+                log.fatal('data put returned neither an error nor a key', {
+                    method: 'storeObject::dataStore',
+                });
+                return cb(errors.InternalError);
+            }
+            log.trace('dataStore: backend stored key', {
+                dataRetrievalInfo,
+            });
+            hashedStream.on('hashed', () => {
+                log.trace('hashed event emitted');
+                return checkHashMatchMD5(stream, hashedStream,
+                    dataRetrievalInfo, log, cb);
+            });
+            if (hashedStream.completedHash) {
+                return checkHashMatchMD5(stream, hashedStream,
+                    dataRetrievalInfo, log, cb);
+            }
+            return undefined;
+        });
+}
--- a/lib/api/objectPutPart.js
+++ b/lib/api/objectPutPart.js
@@ -1,57 +1,25 @@
+import assert from 'assert';
+import async from 'async';
 import { errors } from 'arsenal';
 
-import services from '../services';
 import constants from '../../constants';
 import kms from '../kms/wrapper';
+import { dataStore } from './apiUtils/object/storeObject';
+import metadata from '../metadata/wrapper';
+import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
 
-/*
- * Private function
- */
-function doStore(objectKeyContext, cipherBundle, request, log,
-                 partNumber, size, uploadId, streamingV4Params, cb) {
-    return function mdValidateMPU(err, mpuBucket) {
-        if (err) {
-            return cb(err);
-        }
-        return services.dataStore(
-            null, objectKeyContext, cipherBundle, request,
-            size, streamingV4Params, log, (err, extraArg, dataGetInfo,
-                calculatedHash) => {
-                if (err) {
-                    return cb(err);
-                }
-                let splitter = constants.splitter;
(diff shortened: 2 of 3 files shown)


bennettbuchanan:
@ironman-machine

rahulreddy:
Looks like you committed dump.rdb file by mistake. Can you remove it?

bennettbuchanan:
Thanks @rahulreddy! I've removed it.

lib/api/apiUtils/authorization/aclChecks.js
```diff
@@ -114,3 +114,43 @@ export function isObjAuthorized(bucket, objectMD, requestType, canonicalID) {
     }
     return false;
 }
+
+export function isMpuAuthorized(mpuBucket, authInfo, response, requestType) {
```
LaurenSpiegel: please add jsdoc for this function.  also, please use something clearer than "response" for the argument name here.  Maybe 'mpuOverviewObj'.

lib/api/objectPutPart.js
```diff
@@ -72,95 +36,165 @@ function doStore(objectKeyContext, cipherBundle, request, log,
...
+    const uploadId = request.query.uploadId;
+    const mpuBucketName = `${constants.mpuBucketPrefix}${bucketName}`;
+
+    // If bucket has no server-side encryption, `cipherBundle` remains `null`.
+    let cipherBundle = null;
+    let splitter = constants.splitter;
+
+    return async.waterfall([
+        // Get the bucket.
+        next => metadata.getBucket(bucketName, log, next),
```
LaurenSpiegel: I'd prefer going a little longer form here so we can log here if there is an error getting the bucket.

lib/api/objectPutPart.js
```diff
@@ -72,95 +36,165 @@ function doStore(objectKeyContext, cipherBundle, request, log,
...
+    // `request.query.uploadId` must be called with that exact capitalization.
+    const uploadId = request.query.uploadId;
+    const mpuBucketName = `${constants.mpuBucketPrefix}${bucketName}`;
+
+    // If bucket has no server-side encryption, `cipherBundle` remains `null`.
+    let cipherBundle = null;
+    let splitter = constants.splitter;
+
+    return async.waterfall([
+        // Get the bucket.
```
LaurenSpiegel: // Get the destination bucket

lib/api/objectPutPart.js
```diff
@@ -72,95 +36,165 @@ function doStore(objectKeyContext, cipherBundle, request, log,
...
+
+            return next(null, bucket);
+        },
+        // Get bucket server-side encryption, if it exists.
+        (bucket, next) => {
+            const encryption = bucket.getServerSideEncryption();
+
+            if (encryption) {
+                return kms.createCipherBundle(encryption, log, (err, res) => {
+                    cipherBundle = res;
```
LaurenSpiegel: log any error here

lib/api/objectPutPart.js
```diff
@@ -72,95 +36,165 @@ function doStore(objectKeyContext, cipherBundle, request, log,
...
                 });
-        });
+            }
+
+            return next();
+        },
+        // Get the MPU shadow bucket.
+        next => metadata.getBucket(mpuBucketName, log, (err, mpuBucket) => {
+            if (err) {
+                return next(errors.NoSuchUpload);
```
LaurenSpiegel: log error and make clear this is an error from getting the shadow mpu bucket

lib/api/objectPutPart.js
```diff
@@ -72,95 +36,165 @@ function doStore(objectKeyContext, cipherBundle, request, log,
...
+            return next(null, mpuBucket);
+        }),
+        // Check authorization of the MPU shadow bucket.
+        (mpuBucket, next) => {
+            // BACKWARD: Remove to remove the old splitter
+            if (mpuBucket.getMdBucketModelVersion() < 2) {
+                splitter = constants.oldSplitter;
+            }
+
+            const args = {
```
LaurenSpiegel: Instead of doing a listing, I think we can do a straight get object from metadata, can't we?  At some point we had more info in the key so we weren't able to do that but I think it's possible now and would be more straightforward than doing a listing to get one key.

lib/api/objectPutPart.js
```diff
@@ -72,95 +36,165 @@ function doStore(objectKeyContext, cipherBundle, request, log,
...
+                bucketName,
+                owner: canonicalID,
+                namespace: request.namespace,
+            };
+
+            return services.dataStore(null, objectKeyContext, cipherBundle,
+                request, size, streamingV4Params, log, next);
+        },
+        // Store data locations in metadata.
+        (pass, dataGetInfo, hexDigest, next) => {
```
LaurenSpiegel: pass?

lib/services.js
```diff
@@ -554,14 +554,19 @@ export default {
                         return cb(errors.NoSuchUpload);
                     }
 
+                    if (!isMpuAuthorized(mpuBucket, authInfo, response,
```
LaurenSpiegel: where is this whole function (metadtataValidateMultipart) still used?  Is it possible to still use this function in put part so we don't have duplication if this is still being used or should we get rid of this altogether?

rahulreddy: I originally extracted this method with the intention of using it. It was going to be a step in a direction to dismantle `services.metadataValidateMultipart`. I like the step @bennettbuchanan took to bring the validation logic directly into the api.

@bennettbuchanan you can remove this change, we will make changes similar to what you did for other apis which are using validatemultipart method as a different PR.

bennettbuchanan: Okay, thanks @rahulreddy

lib/api/objectPutPart.js
```diff
@@ -87,80 +56,146 @@ export default function objectPutPart(authInfo, request, streamingV4Params,
...
+                dataGetInfoArr[0].sseCipheredDataKey = cipheredDataKey;
+            }
+            const mdParams = {
+                partNumber: _getPaddedPartNumber(partNumber),
+                contentMD5: hexDigest,
+                size,
+                uploadId,
+                splitter,
+            };
+            return services.metadataStorePart(mpuBucketName, dataGetInfoArr,
```
rahulreddy: This can be extracted into this api as well. It doesn't carry any extra setup, so it should be ok.

lib/api/objectPutPart.js
```diff
@@ -87,80 +56,146 @@ export default function objectPutPart(authInfo, request, streamingV4Params,
...
-        });
+        },
+        // Store in data backend.
+        next => {
+            const objectKeyContext = {
+                bucketName,
+                owner: canonicalID,
+                namespace: request.namespace,
+            };
+            return services.dataStore(null, objectKeyContext, cipherBundle,
```
rahulreddy: After `services.metadataStorePart` this would be the last services call left in this api. My recommendation is to move it to apiUtils/object. Then we could introduce a refactor for other apis to follow the suit.

lib/api/objectPutPart.js
```diff
@@ -87,80 +60,158 @@ export default function objectPutPart(authInfo, request, streamingV4Params,
...
+    const mpuBucketName = `${constants.mpuBucketPrefix}${bucketName}`;
+    const objectKey = request.objectKey;
+    // If bucket has no server-side encryption, `cipherBundle` remains `null`.
+    let cipherBundle = null;
+    let splitter = constants.splitter;
+    return async.waterfall([
+        // Get the destination bucket.
+        next => metadata.getBucket(bucketName, log, (err, bucket) => {
+            if (err && err.NoSuchBucket) {
+                return next(errors.NoSuchUpload);
```
LaurenSpiegel: Did you test this with AWS?  If there is no bucket and you try to do an upload it responds with NoSuchUpload?

lib/api/apiUtils/object/storeObject.js
```diff
@@ -0,0 +1,81 @@
+import { errors } from 'arsenal';
```
LaurenSpiegel: this is all just moved from services with the functions there remaining?  will you please add a note in services.js that these are to be deleted once references are updated in the other api's and open an issue for that to be done?

lib/api/objectPutPart.js
```diff
@@ -1,57 +1,24 @@
...
-                        return cb(null, calculatedHash);
-                    });
-            });
-    };
+// We pad the partNumbers so that the parts will be sorted in numerical order.
+function _getPaddedPartNumber(number) {
+    return `000000${number}`.substr(-5);
+}
+
+function _getPrefixKey(splitter, objectKey, uploadId) {
```
LaurenSpiegel: I would call this _getOverviewKey

LaurenSpiegel:
So much clearer.  Just a few remaining questions/comments.  thanks!!

lib/api/objectPutPart.js
bennettbuchanan: Updated.

lib/api/apiUtils/object/storeObject.js
bennettbuchanan: I have added a TODO comment in services.js.

lib/api/objectPutPart.js
bennettbuchanan: Thanks, good catch! I have updated this to return a `NoSuchBucket` error as is the expected behavior. A `NoSuchUpload` error in this API should only be returned when the MPU bucket is not found.

lib/api/objectPutPart.js
```diff
@@ -72,13 +39,19 @@ function doStore(objectKeyContext, cipherBundle, request, log,
...
     log.debug('processing request', { method: 'objectPutPart' });
-
-    const bucketName = request.bucketName;
-    const objectKey = request.objectKey;
     const size = request.parsedContentLength;
+    // If part size is greater than 5GB, reject it.
+    // Note: Parts are supposed to be at least 5MB except for last part.
+    // However, there is no way to know whether a part is the last part since
+    // keep taking parts until get a completion request.  But can expect parts
+    // of at least 5MB until last part.  Also, we check that part sizes are
```
nicolas2bert: Can you please reword it ? :)

bennettbuchanan: Thanks for pointing this out. I think I have made it more clear.

lib/api/apiUtils/object/storeObject.js
rahulreddy: I think it has to be it's own PR for separation of concerns.

LaurenSpiegel: Please go ahead and squash, rebase and run an end to end and we can merge this then you can open the other pr.

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine after `log.trace` object deletion

bennettbuchanan:
@ironman-machine with POC

bennettbuchanan:
@ironman-machine 
CI_AGGRESSOR_FORCE_BUILD=true

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine hey Jarvis

bennettbuchanan:
end-to-end passed with a variable declaration change in storeObject.js. The `dataStore` function has been refactored accordingly and a new function `checkHashMatchMD5` added.

bennettbuchanan:
@ironman-machine

bennettbuchanan:
end-to-end rebuilt and passed here:
http://ci.ironmann.io/gh/scality/Integration/7246

bennettbuchanan:
@ironman-machine

DavidPineauScality:
@ironman-machine sorry, ultron/rel/6.3 branch wasnt created on Integration. Should be better now.

bennettbuchanan:
Hi @DavidPineauScality, thanks for checking this. 😸

bennettbuchanan:
@ironman-machine after latest rebase

bennettbuchanan:
@ironman-machine after rebase onto master

PR #435, opened 2016-11-29
COMPAT: Return expected response for `MaxUpload` and `Delimiter` params

bennettbuchanan:
Currently, the value of `MaxUploads` for listMultipartUpload API is
the number of MPU in a bucket. This is not the behavior of AWS. This
commit changes the API's behavior so the value of `MaxUploads` can be
specified by the user using the `MaxUploads` parameter. This commit also
returns the expected default of 1000 if no `MaxUploads` is defined.
If `MaxUploads` is defined as `0`, the behavior of AWS is to return an
empty `Uploads` array and set `IsTruncated` to `false`. This commit also
adds this expected behavior.

Currently, when using the file-backend, if a `Delimiter` parameter is
defined and it does not result in a `CommonPrefixes` array that is
greater than 0, `Delimiter` is set to `''`. This causes the XML response
to omit the `Delimeter` key. For compatibility with AWS, this commit
tests an associated change in Arsenal that fixes the `Delimiter` response:
https://github.com/scality/Arsenal/pull/197

A new test that directly compares the expected object has been added to 
verify these behaviors.

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/mpu.js
+++ b/tests/functional/aws-node-sdk/test/object/mpu.js
@@ -6,31 +6,80 @@ import BucketUtility from '../../lib/utility/bucket-util';
 const bucket = 'object-test-mpu';
 const objectKey = 'toAbort&<>"\'';
 
-const throwErr = (str, err) => {
-    process.stdout.write(`${str}: ${err}\n`);
-    throw err;
-};
+// Get the expected object of listMPU API.
+function getExpectedObj(res, data) {
+    // If `maxUploads` is not given as a parameter, it should default to 1000.
+    const maxUploads = data.maxUploads === undefined ? 1000 : data.maxUploads;
 
-const checkValues = (res, uploadId, displayName, cb) => {
-    const prefix = res.Prefix ? res.Prefix : undefined;
-    const delimeter = res.Delimiter ? res.Delimiter : undefined;
-    assert.deepStrictEqual(res.KeyMarker, '');
-    assert.deepStrictEqual(res.UploadIdMarker, '');
-    assert.deepStrictEqual(res.Prefix, prefix);
-    assert.deepStrictEqual(res.Delimiter, delimeter);
-    assert.deepStrictEqual(res.Uploads.length, 1);
-    assert.deepStrictEqual(res.Uploads[0].UploadId, uploadId);
-    assert.deepStrictEqual(res.Uploads[0].StorageClass, 'STANDARD');
-    assert.deepStrictEqual(res.Uploads[0].Owner.DisplayName, displayName);
-    cb();
-};
+    // If `MaxUploads` is defined as 0, `IsTruncated` is set to `false` despite
+    // the fact that there may be multipart uploads in the bucket.
+    if (maxUploads === 0) {
+        return {
+            Bucket: bucket,
+            KeyMarker: '',
+            UploadIdMarker: '',
+            MaxUploads: 0,
+            IsTruncated: false,
+            Uploads: [],
+            CommonPrefixes: [],
+        };
+    }
 
-describe('aws-node-sdk test suite of listMultipartUploads', () => {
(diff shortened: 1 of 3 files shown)


bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine with correct Arsenal branch
SCALITY_ARSENAL_BRANCH=dev/COMPAT/handle_delimter_multipartUploads

bennettbuchanan:
@ironman-machine after fix of `Resolution method is overspecified` error
SCALITY_ARSENAL_BRANCH=dev/COMPAT/handle_delimter_multipartUploads

tests/functional/aws-node-sdk/test/object/mpu.js
```diff
@@ -39,45 +75,66 @@ describe('aws-node-sdk test suite of listMultipartUploads', () => {
...
+        it('should list 0 multipart uploads when MaxUploads is 0', () =>
+            s3.listMultipartUploadsAsync({
+                Bucket: bucket,
+                MaxUploads: 0,
+            })
+            .then(res =>
+                // When `MaxUploads` is set to 0 there is no `Initiated` value.
+                // Also, `IsTruncated` is set to `false` despite the fact that
+                // there is, in fact, one multipart upload in the bucket.
+                assert.deepStrictEqual(res, {
```
LaurenSpiegel: your compare function doesn't work since no uploads?

bennettbuchanan: I had decided the response object was different enough when `MaxUploads` is 0 that it warranted its own object. It would definitely work with the compare function though, and I think that would separate concerns better. Updated! 😃

tests/functional/aws-node-sdk/test/object/mpu.js
```diff
@@ -6,31 +6,82 @@ import BucketUtility from '../../lib/utility/bucket-util';
...
+                DisplayName: displayName,
+                ID: userId,
+            },
+        }],
+        CommonPrefixes: [],
+    };
+
+    // If no `prefixVal` is given, it should not be included in the response.
+    if (!prefixVal) {
+        assert.deepStrictEqual(res.Prefix, undefined);
```
rahulreddy: For separation of concerns, please move these assertions to the `checkValues` function. This way `getExpectedObj` does what it says it does, getting/building the expected object.
Tip: Since prefix is only a string `assert.strictEqual` should be used. `deepStrictEqual` is used typically for object assertions.

tests/functional/aws-node-sdk/test/object/mpu.js
```diff
@@ -6,31 +6,82 @@ import BucketUtility from '../../lib/utility/bucket-util';
...
+        delete expectedObj.Prefix;
+    }
+
+    // If no `delimiter` is given, it should not be included in the response.
+    if (!delimiter) {
+        assert.deepStrictEqual(res.Delimiter, undefined);
```
rahulreddy: same here, please look at the note about assertions above.

tests/functional/aws-node-sdk/test/object/mpu.js
bennettbuchanan: Thanks for those points. ✅

bennettbuchanan:
@ironman-machine
SCALITY_ARSENAL_BRANCH=dev/COMPAT/handle_delimter_multipartUploads

rahulreddy:
This can be rebased and drop me arsenal dep. commit can be removed as the PR has been merged.

bennettbuchanan:
@ironman-machine after rebase and dropped dependency on Arsenal branch

bennettbuchanan:
Re-triggered end-to-end and passed [here](http://ci.ironmann.io/gh/scality/Integration/6895).

bennettbuchanan:
@ironman-machine after rebase

PR #436, opened 2016-12-02
CLEANUP: Reference S3 Server in header response

LaurenSpiegel:


michael-zapata:
Question: wouldn't this break some compatibility testing we may face?

nicolas2bert:
@michael-zapata I am going to check it.

PR #439, opened 2016-12-02
fix: flaky healthcheck test

rahulreddy:
This commit adds a delay of 100ms to ensure that stats request is made
after the Redis calls are complete, otherwise it will result in a race
condition making the test fail.


LaurenSpiegel:
@ironman-machine do it

PR #440, opened 2016-12-02
FT(kinetic): Add a Kinetic backend

AntoninCoulibaly:
This PR replace https://github.com/scality/S3/pull/288 because this took too much time to be merged.
This new one is better, it fix some bugs...

The commit history have to be reword.

### Dependencies
 - Add [kinetic js library](https://github.com/Kinetic/kinetic-js) to the  dependencies

### Config
 - Add the kinetic configuration (host and port).
 - Create a pool of socket in a new instance of [Kinetic](#new-class-kinetic) for handling parallel requests and be faster.

### New class Kinetic
 - This class is implemented for making the usage in kinetic backend easier
 - This class is has all the methods for :
  - managing all the socket
  - handle the requests (PUT, GET, DELETE)

### Backend
 - The backend is just there for using the Kinetic class 

### Use the kinetic force
 - In kinetic we can use the force option to overwrite an object for the put or the delete. So we removed put(overwrite)->delete mechanism

TODO: Find the way to know on what disk we put the data


AntoninCoulibaly:
Commit history cleaned

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+            sock: this._newSocket(),
+            _request: false,
+            _requestGet: false,
+            _events: new EventEmitter(),
+        }));
+        this._chunkTab = [];
+        this._count = 0;
+        this._lexiKey = '0';
+        this.connect();
+        return this;
```
michael-zapata: Returning `this` in a constructor is an anti-pattern. It is an internal method meant to return a new instance of the class in any case.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+     */
+    connect() {
+        this._sockets.forEach((socket, index) => {
+            socket.sock.on('connect', () => {
+                this._bindReadable(socket, index);
+            }).connect(socket.drive);
+        });
+    }
+
+    getIt() {
```
michael-zapata: What is 'it' exactly? It could use a clearer name as well as some documentation.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+        for (let i = 0; i < this._sockets.length; i++) {
+            if (this._sockets[i]._request === false
+                && this._sockets[i]._requestGet === false) {
+                return i;
+            }
+        }
+        this._sockets.push(
+            {
+                drive: { host: '127.0.0.1', port: '8123' },
```
michael-zapata: It seems a bit strange to have hardcoded values here. Am I wrong?

Also, nit: Weird indent.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+    /**
+     * send the put request to the right drive.
+     *
+     * @param {number} index - the index of the socket
+     * @param {Buffer} value - the value to be put.
+     * @param {Object} options - the options object for the put request:
+     *                     kinetic-protocol
+     *        {string} options.synchronization - synchronization mode
+     *        {boolean} options.force - option for forcing the put
+     * @param {function} callback - the callback
```
michael-zapata: I would love to see the callback arguments defined, and the same goes for every other callback.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+    /**
+     * get index of the socket.
+     *
+     * @param {string} host - the host of the socket.
+     * @param {string} port - the port of the socket.
+     * @returns {number|object} - the index of the socket or null if it does not
+     *                            exist.
+     */
+    getSocketIndex(host, port) {
+        this._sockets.forEach((socket, index) => {
```
michael-zapata: If you want to return the index, you'll need to use the `findIndex()` array method instead.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+            return errors.ObjNotFound;
+        }
+        if (statusCode !== kineticlib.errors.SUCCESS) {
+            return pdu.getErrorMessage();
+        }
+        return undefined;
+    }
+
+    /**
+     * set the sequence of the request.
```
michael-zapata: What is a sequence? The documentation could help in that regard.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+            }
+            if (index === this._sockets.length - 1) {
+                return null;
+            }
+            return undefined;
+        });
+    }
+
+    _incrString(str) {
+        this._lexiKey = (parseInt(str, 36) + 1).toString(36);
```
michael-zapata: `parseInt()` can potentially return `NaN`, is this driven by user input?

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+    /**
+     * send the get request to the right drive.
+     *
+     * @param {number} index - the index of the socket
+     * @param {Buffer} key - the value to be put.
+     * @param {number[]} range - the range to get
+     * @param {function} callback - the callback
+     * @returns {undefined}.
+     */
+    get(index, key, range, callback) {
```
michael-zapata: I'm not exactly sure I'm getting this right, but are we putting everything in memory here? It's fine for now as objects are at maximum 1Mo, but we'll circumvent that limit in the future. This warrants at least a `FIXME` or a `TODO` in there, with an associated issue.

lib/api/objectPut.js
```diff
@@ -17,7 +17,7 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
             if (err) {
                 return callback(err);
             }
-            if (dataToDelete) {
+            if (dataToDelete && dataToDelete[0].dataStoreName !== 'kinetic') {
```
michael-zapata: Please add in this commit that the kinetic backend does not handle the batch delete yet.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+            const protobufSize = header.readInt32BE(1);
+            const rawData = socket.sock.read(protobufSize);
+            if (rawData !== null) {
+                const pdu = new kineticlib.PDU(
+                    Buffer.concat([header, rawData]));
+                const err = this._propError(pdu);
+                return callback(err, pdu);
+            }
+        }
+        return undefined;
```
michael-zapata: So in case there is no header, we do not call the callback? That seems a bit stange.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    _parseRequest(err, pdu, index) {
```
michael-zapata: I'd love documentation in there.

AntoninCoulibaly:
@michael-zapata thank you for the review, I am on it

lib/data/kinetic/Kinetic.js
AntoninCoulibaly: Not at all. It is just a lil mechanism to have keys in lexicographical order

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,348 @@
...
+                    'getResponse', errors.InternalError);
+            }
+            this._sockets[index]._events.emit('getResponse', null, chunk);
+        } else {
+            this._chunkTab.push(chunk);
+        }
+    }
+
+    _endGet(index) {
+        let chunk = Buffer.allocUnsafe(0);
```
MathieuCassagne: Same comment about the buffer.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,348 @@
...
+     * send the get request to the right drive.
+     *
+     * @param {number} index - the index of the socket
+     * @param {Buffer} key - the value to be put.
+     * @param {number[]} range - the range to get
+     * @param {function} callback - the callback
+     * @returns {undefined}.
+     */
+    get(index, key, range, callback) {
+        let endValue = Buffer.allocUnsafe(0);
```
MathieuCassagne: Same about buffer.

lib/api/objectPut.js
AntoninCoulibaly: done

lib/data/kinetic/Kinetic.js
AntoninCoulibaly: TODO comment added

lib/data/kinetic/Kinetic.js
AntoninCoulibaly: yes you are write, I have to rewrite this part of code. currently commented

lib/Config.js
```diff
@@ -318,6 +318,23 @@ class Config {
             dataPath,
             metadataPath,
         };
+
+        if (config.kinetic && process.env.S3SPROXYD === 'kinetic') {
+            assert(Array.isArray(config.kinetic),
+                   'The kinetic config should be an array - ["host:port"]');
+            this.kinetic = {};
```
MathieuCassagne: Cosmetics but you can write:
```
this.kinetic = {
    drives: [],
    hosts: [],
};
```

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    _parsePDU(socket, callback) {
+        const header = socket.sock.read(HEADER_SZ);
+        if (header !== null) {
+            const protobufSize = header.readInt32BE(1);
+            const rawData = socket.sock.read(protobufSize);
+            if (rawData !== null) {
+                const pdu = new kineticlib.PDU(
+                    Buffer.concat([header, rawData]));
```
MathieuCassagne: You can pass the final length to ```Buffer.concat()``, so it doesn't have to calculate it and so is a bit faster.
```
Buffer.concat([header, rawData], header.length + rawData.length)
```

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,350 @@
...
+     *                     kinetic-protocol
+     *        {string} options.synchronization - synchronization mode
+     *        {boolean} options.force - option for forcing the put
+     * @param {function} callback - the callback
+     * @returns {undefined}.
+     */
+    put(index, value, options, callback) {
+        this._sockets[index]._request = true;
+        const socket = this._sockets[index];
+        const key = Buffer.concat([Buffer.from(
```
MathieuCassagne: Just do the ```Buffer.from()``` isn't enough ? Why concatening only one buffer ?

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,350 @@
...
+        const pdu = new kineticlib.DeletePDU(
+            socket.sequence, socket.connectionID,
+            socket.clusterVersion, key, options);
+        this.setSequence(index, socket.sequence + 1);
+        socket.sock.write(pdu.read(), err => {
+            if (err) {
+                return callback(err);
+            }
+            return this._sockets[index]._events.once(
+                'deleteResponse', err => callback(err));
```
MathieuCassagne: You can just pass the callback here :)

lib/data/kinetic/backend.js
```diff
@@ -0,0 +1,41 @@
+import config from '../../Config';
+
+const backend = {
+    put: function putK(request, size, keyContext, reqUids, callback, drive) {
```
MathieuCassagne: Maybe you can track the length of the data received, with a valueLen variable, and then pass it to Buffer.concat() :)

lib/data/kinetic/Kinetic.js
AntoninCoulibaly: you are right, you already said me that ahah

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,350 @@
...
+
+class Kinetic {
+
+    /**
+     * @constructor
+     * @param {Object.<string>[]} drives - the list of drives
+     * @returns {Kinetic} - to allow for a functional style.
+     */
+    constructor(drives) {
+        this._sockets = [];
```
MathieuCassagne: Since you creates it with a map function, you can remove this line :)

lib/data/kinetic/Kinetic.js
AntoninCoulibaly: forgot to remove it

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,350 @@
...
+     * @returns {Kinetic} - To allow for a functional style.
+     */
+    setSequence(index, sequence) {
+        if (sequence >= Number.MAX_VALUE) {
+            const temp = this._sockets[index].sock;
+            this._sockets[index].sock = this._newSocket();
+            this._sockets[index].sock.on('connect', () => {
+                this._bindReadable(this._sockets[index], index);
+            }).connect(this._sockets[index].drive);
+            temp.destroy();
```
MathieuCassagne: Can we still receive data on the "temp" socket ? If yes there is a risk of destoying it before actually received theses datas.

lib/data/kinetic/Kinetic.js
```diff
@@ -0,0 +1,350 @@
...
+     * @param {Buffer} value - the value to be put.
+     * @param {Object} options - the options object for the put request:
+     *                     kinetic-protocol
+     *        {string} options.synchronization - synchronization mode
+     *        {boolean} options.force - option for forcing the put
+     * @param {function} callback - return an error or the key of the object
+     * @returns {undefined}.
+     */
+    put(index, value, options, callback) {
+        this._sockets[index]._request = true;
```
MathieuCassagne: You can do:
```
const socket = this._sockets[index];
socket._request = true;
```
Just cosmetics :)

lib/data/kinetic/Kinetic.js
AntoninCoulibaly: normally no, because we choose a socket with no current request we are just doing this to avoid an error with the maximum value of a number

PR #446, opened 2016-12-05
BF: ListPart - when PartNumberMarker > 0 return no part

### LINKED ISSUE #445: Can't list parts from a given PartNumberMarker
I have a small file which I multipart uploaded using 7 parts of size 5MB.
When I try to list parts starting from part number 2, the response doesn't return anything (no parts and the response isn't truncated).

I attached a small C# example of how I'm trying to get the uploaded parts.
[ListParts.txt](https://github.com/scality/S3/files/631582/ListParts.txt)

If I set PartNumberMarker to null (default), then I get all the parts as expected.
However this becomes a problem when the number of uploaded parts grows beyond the 1000 limit for one ListParts operation.

nicolas2bert:
FIXES #445

The issue is related with the fact that in the [MarkerFilter function ](https://github.com/scality/S3/blob/d3ba43b87ce07a640ade2413f5f915e178a73dff/lib/metadata/in_memory/bucket_utilities.js#L1), a marker was never equals to any keys.  This is for the in memory implementation.  For the file backend and other backends, LevelDB also needs a padded part number to sort correctly.

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/listParts.js
+++ b/tests/functional/aws-node-sdk/test/object/listParts.js
@@ -1,0 +1,77 @@
+import assert from 'assert';
+
+import withV4 from '../support/withV4';
+import BucketUtility from '../../lib/utility/bucket-util';
+
+const bucket = 'bucketlistparts';
+const key = 'key';
+const bodyFirstPart = Buffer.allocUnsafe(10);
+const bodySecondPart = Buffer.allocUnsafe(20);
+
+function checkNoError(err) {
+    assert.equal(err, null,
+        `Expected success, got error ${JSON.stringify(err)}`);
+}
+
+describe('List parts', () => {
+    withV4(sigCfg => {
+        let bucketUtil;
+        let s3;
+        let uploadId;
+        let secondEtag;
+
+        beforeEach(() => {
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
+            return s3.createBucketAsync({ Bucket: bucket })
+            .then(() => s3.createMultipartUploadAsync({
+                Bucket: bucket, Key: key }))
+            .then(res => {
+                uploadId = res.UploadId;
+                return s3.uploadPartAsync({ Bucket: bucket, Key: key,
+                  PartNumber: 1, UploadId: uploadId, Body: bodyFirstPart });
+            }).then(() => s3.uploadPartAsync({ Bucket: bucket, Key: key,
+                PartNumber: 2, UploadId: uploadId, Body: bodySecondPart })
+            ).then(res => {
+                secondEtag = res.ETag;
+                return secondEtag;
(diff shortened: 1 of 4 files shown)


tests/functional/aws-node-sdk/test/object/listParts.js
```diff
@@ -0,0 +1,77 @@
...
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
+            return s3.createBucketAsync({ Bucket: bucket })
+            .then(() => s3.createMultipartUploadAsync({
+                Bucket: bucket, Key: key }))
+            .then(res => {
+                uploadId = res.UploadId;
+                const params = { Bucket: bucket, Key: key, PartNumber: 1,
+                    UploadId: uploadId, Body: bodyFirstPart };
+                return s3.uploadPartAsync(params);
```
bennettbuchanan: For the sake of consistency, you might pass the params object directly into `s3.uploadPartAsync` as you've done just below for part 2.

nicolas2bert: Yes - thanks :)

tests/functional/aws-node-sdk/test/object/listParts.js
```diff
@@ -0,0 +1,76 @@
...
+                Bucket: bucket, Key: key }))
+            .then(res => {
+                uploadId = res.UploadId;
+                return s3.uploadPartAsync({ Bucket: bucket, Key: key,
+                  PartNumber: 1, UploadId: uploadId, Body: bodyFirstPart });
+            }).then(() => s3.uploadPartAsync({ Bucket: bucket, Key: key,
+                PartNumber: 2, UploadId: uploadId, Body: bodySecondPart })
+            ).then(res => {
+                secondEtag = res.ETag;
+                return secondEtag;
```
bennettbuchanan: Shouldn't need to return `secondEtag` here 😄

lib/utapi/utilities.js
```diff
@@ -38,7 +38,6 @@ function _listBucketMetrics(host,
             const responseBody = JSON.parse(body.join(''));
             if (response.statusCode >= 200 && response.statusCode < 300) {
                 // eslint-disable-next-line no-console
-                console.log(responseBody);
```
LaurenSpiegel: please restore this.  it is intentional and does not have to do with this pr.

nicolas2bert: Ok - done.

michael-zapata: Technically, a `process.stdout.write()` makes more sense outside of a browser, and would lift the need to disable the linter. But that entirely orthogonal to this PR anyway.

LaurenSpiegel:
Let's do it against rel/6.2.1

nicolas2bert:
@ironman-machine

PR #448, opened 2016-12-06
COMPAT: add a readable headers checker

### LINKED ISSUE #447: COMPAT: unreadable header(s) return AccessDenied on AWS
On **ceph/s3-tests**, this test fails:
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_md5_unreadable`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_contenttype_unreadable`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_date_unreadable_aws4`

nicolas2bert:
FIXES #447

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -206,7 +206,7 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
 };
 
 /**
- * Validate headers per testing if readable
+ * Validate request headers to determine if readable characters
  * @param {object} headers - headers attached to the http request (lowercased)
  * @return {boolean} - returns true if headers are readable, false if not
  */
@@ -214,17 +214,15 @@ utils.checkReadableHeaders = function checkReadableHeaders(headers) {
     // check every header except the x-amz-meta- ones for AWS compatibility
     const isReadableHeaders = Object.keys(headers).filter(h =>
     !h.startsWith('x-amz-meta-')).every(headerKey => {
-        const headerVal = String(headers[headerKey]);
-        let isReadableChar = true;
+        const headerVal = headers[headerKey];
         // check that each character is readable
         for (let i = 0; i < headerVal.length; i++) {
             const ch = headerVal.charCodeAt(i);
             if (ch < 32 || ch > 127) {
-                isReadableChar = false;
-                break;
+                return false;
             }
         }
-        return isReadableChar;
+        return true;
     });
     return isReadableHeaders;
 };
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -205,6 +205,29 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
         .update(namespace + resource, 'binary').digest('hex');
 };
 
+/**
+ * Validate headers per testing if readable
+ * @param {object} headers - headers attached to the http request (lowercased)
+ * @return {boolean} - returns true if headers are readable, false if not
+ */
+utils.checkReadableHeaders = function checkReadableHeaders(headers) {
+    // check every header except the x-amz-meta- ones for AWS compatibility
+    const isReadableHeaders = Object.keys(headers).filter(h =>
+    !h.startsWith('x-amz-meta-')).every(headerKey => {
+        const headerVal = String(headers[headerKey]);
+        let isReadableChar = true;
+        // check that each character is readable
+        for (let i = 0; i < headerVal.length; i++) {
+            const ch = headerVal.charCodeAt(i);
+            if (ch < 32 || ch > 127) {
+                isReadableChar = false;
+                break;
+            }
+        }
+        return isReadableChar;
+    });
+    return isReadableHeaders;
+};
 
 /**
  * Modify http request object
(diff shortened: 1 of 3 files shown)


lib/utils.js
```diff
@@ -205,6 +205,31 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
...
+utils.checkReadableHeaders = function checkReadableHeaders(headers) {
+    // check every header except the x-amz-meta- ones for AWS compatibility
+    const isReadableHeaders = Object.keys(headers).filter(h =>
+    !h.startsWith('x-amz-meta-')).every(headerKey => {
+        const headerVal = String(headers[headerKey]);
+        let isReadableChar = true;
+        // check that each character is readable
+        for (let i = 0; i < headerVal.length; i++) {
+            const ch = headerVal.charCodeAt(i);
+            if (ch >= 32 && ch <= 127) {
```
bennettbuchanan: Could this be reduced to a single `if` condition instead of an `if...else` statement?

nicolas2bert: You are right. Thanks 👍

lib/utils.js
```diff
@@ -205,6 +205,29 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
         .update(namespace + resource, 'binary').digest('hex');
 };
 
+/**
+ * Validate headers per testing if readable
```
LaurenSpiegel: Validate request headers to determine if readable characters

lib/utils.js
```diff
@@ -205,6 +205,29 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
...
+ * @param {object} headers - headers attached to the http request (lowercased)
+ * @return {boolean} - returns true if headers are readable, false if not
+ */
+utils.checkReadableHeaders = function checkReadableHeaders(headers) {
+    // check every header except the x-amz-meta- ones for AWS compatibility
+    const isReadableHeaders = Object.keys(headers).filter(h =>
+    !h.startsWith('x-amz-meta-')).every(headerKey => {
+        const headerVal = String(headers[headerKey]);
```
LaurenSpiegel: why do you have to stringify this?

lib/utils.js
```diff
@@ -205,6 +205,29 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
...
+        let isReadableChar = true;
+        // check that each character is readable
+        for (let i = 0; i < headerVal.length; i++) {
+            const ch = headerVal.charCodeAt(i);
+            if (ch < 32 || ch > 127) {
+                isReadableChar = false;
```
LaurenSpiegel: just return false here.  no need to break or define isReadableChar

lib/utils.js
```diff
@@ -205,6 +205,29 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
...
+                break;
+            }
+        }
+        return isReadableChar;
+    });
+    return isReadableHeaders;
```
LaurenSpiegel: return true if you made it through

tests/functional/s3curl/tests.js
```diff
@@ -465,6 +465,19 @@ describe('s3curl putObject', () => {
                 });
         });
 
+    it('should not be able to put an object if content-md5 header is ' +
```
LaurenSpiegel: please add test showing an x-amz-meta header can have unreadable chars

lib/utils.js
```diff
@@ -205,6 +205,27 @@ utils.getResourceUID = function getResourceUID(namespace, resource) {
         .update(namespace + resource, 'binary').digest('hex');
 };
 
+/**
+ * Validate request headers to determine if readable characters
+ * @param {object} headers - headers attached to the http request (lowercased)
+ * @return {boolean} - returns true if headers are readable, false if not
+ */
+utils.checkReadableHeaders = function checkReadableHeaders(headers) {
```
MathieuCassagne: Are you sure the returning value is a boolean ?
For what I understand ```isReadableHeaders``` is an array of valid headerKey. Should you return ```isReadableHeader.length !== 0``` ?

nicolas2bert: @MathieuCassagne - Yep, `every` method return true if the callback function returns a truthy value for every array element; otherwise, false.

nicolas2bert:
@ironman-machine ok

lib/utils.js
MathieuCassagne: Yes, but you're using the result of every to do the filtering, right ?

nicolas2bert: `every` tests whether all elements in the array pass the test implemented by the provided function.
I get my array from `Object.keys(headers).filter(h => !h.startsWith('x-amz-meta-'))` and my test checks that each character is readable.

Is it clearer? :)   
Thanks for the review.

MathieuCassagne: Yes, I missed that the end of the filter function was after the ```!startsWith```, sorry for annoying ^^

PR #452, opened 2016-12-07
FIX: Complete MPU allows fewer parts than put

LaurenSpiegel:
A user should be able to complete an MPU with fewer
parts than were originally put.  The other parts
are just discarded.

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -170,9 +170,29 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             // If the number of the parts in the JSON does not
             // match the number of parts stored, return error msg
             const partLength = jsonList.Part.length;
-            if (partLength !== storedPartsAsObjects.length) {
+
+            // A user can put more parts than they end up including
+            // in the completed MPU but there cannot be more
+            // parts in the complete message than were already put
+            if (partLength > storedPartsAsObjects.length) {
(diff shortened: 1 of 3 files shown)


LaurenSpiegel:
@ironman-machine check it

lib/api/completeMultipartUpload.js
```diff
@@ -361,15 +383,19 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
                 });
         },
         function waterfall8(mpuBucket, mpuOverviewKey, aggregateETag,
-            storedPartsAsObjects, next) {
+            storedPartsAsObjects, extraPartLocations, next) {
             const keysToDelete = storedPartsAsObjects.map(item => item.key);
             keysToDelete.push(mpuOverviewKey);
             services.batchDeleteObjectMetadata(mpuBucket.getName(),
                 keysToDelete, log, err => next(err, aggregateETag));
+            data.batchDelete(extraPartLocations, logger
```
electrachong: minor: we could check if extraPartLocations is an empty array to avoid making an unnecessary function call & extra log entry in batchDelete with no keys to delete

LaurenSpiegel: good call.  I made the change.

LaurenSpiegel:

LaurenSpiegel:
end to end passed http://ci.ironmann.io/gh/scality/Integration/7687

PR #453, opened 2016-12-07
BF: Fix logging location of delete keys

### LINKED ISSUE #451: Logging deletes
We are calling batch delete from a number of api's (complete mpu for example).  Instead of logging in services.deleteObject the keys, we should do it in the wrapper to make sure we are not missing any.

electrachong:
Fixes #451

### DIFF
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -119,7 +119,7 @@ const data = {
         // If objectGetInfo.key exists the md-model-version is 2 or greater.
         // Otherwise, the objectGetInfo is just the key string.
         const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
-        log.debug('sending delete to datastore', {
+        log.trace('sending delete to datastore', {
             implName,
             key,
             method: 'delete',
@@ -136,6 +136,15 @@ const data = {
     // It would be preferable to have an sproxyd batch delete route to
     // replace this
     batchDelete: (locations, log) => {
+        // TODO: The method of persistence of sproxy delete key will
+        // be finalized; refer Issue #312 for the discussion. In the
+        // meantime, we at least log the location of the data we are
+        // about to delete before attempting its deletion.
+        log.trace('initiating batch delete', {
+            keys: locations,
+            implName,
+            method: 'batchDelete',
+        });
         async.eachLimit(locations, 5, (loc, next) => {
             data.delete(loc, log, next);
         },
--- a/lib/services.js
+++ b/lib/services.js
@@ -374,12 +374,7 @@ export default {
                         return cb(err);
                     }
                     cb();
-                    // TODO: The method of persistence of sproxy delete key will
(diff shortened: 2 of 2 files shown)


lib/data/wrapper.js
```diff
@@ -137,6 +132,16 @@ const data = {
     // replace this
     batchDelete: (locations, log) => {
         async.eachLimit(locations, 5, (loc, next) => {
+            // TODO: The method of persistence of sproxy delete key will
+            // be finalized; refer Issue #312 for the discussion. In the
+            // meantime, we at least log the location of the data we are
+            // about to delete before attempting its deletion.
+            log.trace('About send delete to datastore, logging location in ' +
```
bennettbuchanan: Typo: 'About send' should be 'About to send'.

lib/data/wrapper.js
```diff
@@ -137,6 +132,16 @@ const data = {
     // replace this
     batchDelete: (locations, log) => {
         async.eachLimit(locations, 5, (loc, next) => {
```
LaurenSpiegel: I would do the log at line 134 before the async.eachLimit and log the full array so that we know what is intended to be deleted before any possible crash on the multiple async calls.  So no need to log in the async.eachLimit function.

lib/data/wrapper.js
```diff
@@ -119,11 +119,6 @@ const data = {
         // If objectGetInfo.key exists the md-model-version is 2 or greater.
         // Otherwise, the objectGetInfo is just the key string.
         const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
-        log.debug('sending delete to datastore', {
```
LaurenSpiegel: Let's keep this.  So we will have one log of all the keys intended to be deleted and then this with each key as it goes.

electrachong: OK, I put it back, but as a trace

lib/data/wrapper.js
electrachong: Thanks

electrachong:
@ironman-machine please

lib/data/wrapper.js
```diff
@@ -136,6 +136,16 @@ const data = {
     // It would be preferable to have an sproxyd batch delete route to
     // replace this
     batchDelete: (locations, log) => {
+        // TODO: The method of persistence of sproxy delete key will
+        // be finalized; refer Issue #312 for the discussion. In the
+        // meantime, we at least log the location of the data we are
+        // about to delete before attempting its deletion.
+        log.trace('About to initiate batchDelete, logging object keys in ' +
```
rahulreddy: We should aim for shorter log messages, as it would be easier to grep or to look at in Kibana. 
`initiating batch delete` should be enough IMO. You can extend the comment mentioning gc to add some context.
Edit: We should be careful of logging large objects as this may contribute to negative performance.

PR #456, opened 2016-12-09
FIX: Fill buffer with zeros

nicolas2bert:
Because using `allocUnsafe` the contents of the newly created Buffer are unknown and may contain sensitive data

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/listParts.js
+++ b/tests/functional/aws-node-sdk/test/object/listParts.js
@@ -5,8 +5,8 @@ import BucketUtility from '../../lib/utility/bucket-util';
 
 const bucket = 'bucketlistparts';
 const key = 'key';
-const bodyFirstPart = Buffer.allocUnsafe(10);
-const bodySecondPart = Buffer.allocUnsafe(20);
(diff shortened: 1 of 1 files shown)


ThibaultRiviere:
Why not using safe alloc ?

LaurenSpiegel:
@ThibaultRiviere, there is a slight perf gain to doing it this way though for testing doesn't really matter.  Why do you think using safe alloc is preferable in this instance?

From the node docs:

Note that the Buffer module pre-allocates an internal Buffer instance of size Buffer.poolSize that is used as a pool for the fast allocation of new Buffer instances created using Buffer.allocUnsafe() and the deprecated new Buffer(size) constructor only when size is less than or equal to Buffer.poolSize >> 1 (floor of Buffer.poolSize divided by two).

Use of this pre-allocated internal memory pool is a key difference between calling Buffer.alloc(size, fill) vs. Buffer.allocUnsafe(size).fill(fill). Specifically, Buffer.alloc(size, fill) will never use the internal Buffer pool, while Buffer.allocUnsafe(size).fill(fill) will use the internal Buffer pool if size is less than or equal to half Buffer.poolSize. The difference is subtle but can be important when an application requires the additional performance that Buffer.allocUnsafe() provides.

PR #458, opened 2016-12-10
Make func check handling consistent across backends

dora-korpar:
The handling to check if the backend healthcheck function exists was done differently in vaultclient and bucketclient from sproxyd. This makes sproxyd follow the same format.


dora-korpar:
@ironman-machine check it

PR #465, opened 2016-12-13
fix: allow bucket names with IP addresses in them

### LINKED ISSUE #464: Bucket creation fails with InvalidBucket error if name contains IP addresses
Bucket names like
- prefix-192.168.5.4
- pre-192.168.5.4-post

should succeed.

rahulreddy:
The bucketname check loosely checked if it matches an IP address, this
failed bucketnames which includes IP with prefixes. The commit adds a strict
check so that regex matches exactly an IP address.

fix #464


rahulreddy:
Added a test for creating bucket names with IP + suffix

PR #467, opened 2016-12-14
Ft/serve website

LaurenSpiegel:
@nicolas2bert is currently writing more tests

Depends on https://github.com/scality/Arsenal/pull/203

### DIFF
--- a/lib/routes/routeWebsite.js
+++ b/lib/routes/routeWebsite.js
@@ -1,0 +1,67 @@
+import { errors } from 'arsenal';
+
+import api from '../api/api';
+import pushMetrics from '../utilities/pushMetrics';
+import routesUtils from './routesUtils';
+import statsReport500 from '../utilities/statsReport500';
+
+export default function routerWebsite(request, response, log, utapi,
+    statsClient) {
+    log.debug('routing request', { method: 'routerWebsite' });
+    // website endpoint only supports GET and HEAD and must have a bucket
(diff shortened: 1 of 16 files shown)


lib/routes/routeWebsite.js
```diff
@@ -0,0 +1,67 @@
...
+import routesUtils from './routesUtils';
+import statsReport500 from '../utilities/statsReport500';
+
+export default function routerWebsite(request, response, log, utapi,
+    statsClient) {
+    log.debug('routing request', { method: 'routerWebsite' });
+    // website endpoint only supports GET and HEAD and must have a bucket
+    // http://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html
+    if ((request.method !== 'GET' && request.method !== 'HEAD')
+        || !request.bucketName) {
```
electrachong: @nicolas2bert may be able to confirm, but I believe if there is no bucket specified in the request, AWS behavior is to redirect to aws s3 homepage, not to return MethodNotAllowed. Is it a deliberate choice for us to return MethodNotAllowed instead?

nicolas2bert: This is correct @electrachong AWS return AWS s3 homepage  - what we could do is redirecting to s3.scality.com - It won t be compatible anyway but I guess better than returning an error.

LaurenSpiegel: I did that intentionally since I didn't think it quite made sense to redirect to scality.com.

LaurenSpiegel:
closing in favor of https://github.com/scality/S3/pull/484

PR #469, opened 2016-12-14
REFACTOR: Add & use bucketOwnerAction requestType

electrachong:
There is a similar group of bucket-level actions which are only permitted if user is the bucket owner or has granted permissions. This commit refactors acl checks to explicitly check whether an action belongs to this category and adds this requestType to bucketShield. For existing api actions, it affects bucketPutVersioning, bucketGetVersioning, and bucketPutWebsite.


electrachong:
@ironman-machine

PR #470, opened 2016-12-15
Restore max limit on listing

LaurenSpiegel:
Allow a bucket listing to return a max keys number
greater than actual keys returned to mirror AWS.


LaurenSpiegel:
@ironman-machine check it

PR #474, opened 2016-12-15
Forward/rel/6.3

LaurenSpiegel:


rahulreddy:
End to end http://ci.ironmann.io/gh/scality/Integration/7553

PR #477, opened 2016-12-16
COMPAT: Mirror AWS err for bucketAlreadyOwned

LaurenSpiegel:
AWS does not return 409 for bucketAlreadyOwned
if bucket is in us-east-1 and request is to us-east-1.

This fixes https://github.com/scality/S3/issues/475

### DIFF
--- a/config.json
+++ b/config.json
@@ -36,5 +36,6 @@
     },
     "healthChecks": {
         "allowFrom": ["127.0.0.1/8", "::1"]
-    }
+    },
+    "usEastBehavior": false
 }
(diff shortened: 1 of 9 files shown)


LaurenSpiegel:
@ironman-machine check it

LaurenSpiegel:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/7564

rahulreddy:
I have commented on the Federation PR as well, how about just calling it `usEast1Behavior` since it's different for `us-east-2`

PR #478, opened 2016-12-16
Handle bad headers

LaurenSpiegel:
Translate x-amz headers on receipt and before sending out to handle translated characters.


LaurenSpiegel:
@ironman-machine go

GiorgioRegni:
Merging, necessary for customer tests tomorrow

PR #480, opened 2016-12-16
FIX: websiteGet 

nicolas2bert:
Fixing websiteGet feature.

### DIFF
--- a/lib/routes/routesUtils.js
+++ b/lib/routes/routesUtils.js
@@ -428,19 +428,30 @@ const routesUtils = {
      * @param {string} [routingInfo.prefixFromRule] - key prefix to be replaced
      * @param {string} objectKey - key name (may have been modified in
      * websiteGet api to include index document)
+     * @param {string} indexDocument - index document's name
      * @param {boolean} encrypted - whether request was https
      * @param {object} response - response object
+     * @param {string} hostHeader - host's header from headers.host
      * @param {object} log - Werelogs instance
      * @return {undefined}
      */
-    redirectRequest(routingInfo, objectKey, encrypted, response, log) {
+    redirectRequest(routingInfo, objectKey, indexDocument, encrypted, response,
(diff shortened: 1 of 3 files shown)


lib/routes/routesUtils.js
```diff
@@ -428,19 +428,30 @@ const routesUtils = {
...
         const redirectProtocol = protocol || encrypted ? 'https' : 'http';
         const redirectCode = httpRedirectCode || 302;
+        const redirectHostName = hostName || hostHeader;
 
-        let redirectKey = objectKey === undefined ? '' : objectKey;
+        let redirectKey;
+        if (objectKey === undefined) {
+            redirectKey = '';
+        } else if (indexDocument) {
+            redirectKey = objectKey.replace(indexDocument, '');
```
electrachong: Does replacing the index key with an empty string fix some tests? Which tests were not passing with the previous behavior?

lib/routes/routesUtils.js
```diff
@@ -428,19 +428,30 @@ const routesUtils = {
...
 
         const redirectProtocol = protocol || encrypted ? 'https' : 'http';
         const redirectCode = httpRedirectCode || 302;
+        const redirectHostName = hostName || hostHeader;
 
-        let redirectKey = objectKey === undefined ? '' : objectKey;
+        let redirectKey;
+        if (objectKey === undefined) {
+            redirectKey = '';
+        } else if (indexDocument) {
```
electrachong: missing indentation and should be `else if` right?

nicolas2bert: no worries - thanks for the review :)

electrachong:
Opened a new PR https://github.com/scality/S3/pull/483 with some modifications so that we do not pass modified object keys to the redirect functions in the first place (this way we do not need to pass indexDocument through several functions just to remove it again). Closing this pull request.

PR #482, opened 2016-12-19
Ft/ipv6 support

dora-korpar:
Adds support to request made to the S3 Server over IPv6

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -37,7 +37,7 @@ class S3Server {
     /*
      * This starts the http server.
      */
-    startup() {
+    startup(port) {
         // Todo: http.globalAgent.maxSockets, http.globalAgent.maxFreeSockets
         if (_config.https) {
             this.server = https.createServer({
@@ -67,13 +67,13 @@ class S3Server {
         }
         this.server.on('listening', () => {
             const addr = this.server.address() || {
-                address: '0.0.0.0',
-                port: _config.port,
+                address: '[::]',
+                port,
             };
             logger.info('server started', { address: addr.address,
                 port: addr.port, pid: process.pid });
         });
-        this.server.listen(_config.port);
+        this.server.listen(port);
     }
 
     /*
@@ -110,7 +110,7 @@ class S3Server {
                 log.info('initial health check succeeded', {
                     healthStatus: results,
                 });
-                this.startup();
+                this.startup(_config.port);
             }
         });
     }
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,5 +1,6 @@
 import url from 'url';
 import crypto from 'crypto';
+import { ipCheck } from 'arsenal';
 
 import config from './Config';
 import constants from '../constants';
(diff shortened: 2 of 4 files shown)


dora-korpar:
@ironman-machine gogo gadget ci

lib/server.js
```diff
@@ -110,7 +110,9 @@ class S3Server {
                 log.info('initial health check succeeded', {
                     healthStatus: results,
                 });
-                this.startup();
+                // creates dual-stack endpoints for IPv4 and IPv6
+                this.startup('0.0.0.0', 8000);
```
rahulreddy: We would still want this to be configurable, so that Federation can change the ports if needed. Having something like

```js
this.startup('0.0.0.0', _config.port);
this.startup('::', _config.ipv6Port);
```
would help.

lib/server.js
```diff
@@ -110,7 +110,9 @@ class S3Server {
...
+                this.startup('::1', 8001);
```
rahulreddy: Since we want to listen on all local IPV6 addresses, this should be `::` and not `::1`. Can you confirm if that works?

lib/server.js
```diff
@@ -37,7 +37,7 @@ class S3Server {
     /*
      * This starts the http server.
      */
-    startup() {
+    startup(ipAddr, portNum) {
```
rahulreddy: Minor: If you just call it `port`, you can use shorthand to create the object.

dora-korpar: Done :)

lib/server.js
dora-korpar: and done!

dora-korpar:
@ironman-machine again plz

lib/utils.js
```diff
@@ -18,14 +18,14 @@ utils.getBucketNameFromHost = function getBucketNameFromHost(request) {
     if (headers === undefined || headers.host === undefined) {
         throw new Error('bad request: no host in headers');
     }
-
-    const host = headers.host.split(':')[0];
+    const sliceLength = headers.host.lastIndexOf(':') > 0 ?
+        headers.host.lastIndexOf(':') : headers.host.length;
+    const host = headers.host.slice(0, sliceLength);
```
LaurenSpiegel: Will you please add some unit tests for getBucketNameFromHost with ipv6?

lib/utils.js
```diff
@@ -18,14 +18,14 @@ utils.getBucketNameFromHost = function getBucketNameFromHost(request) {
...
 
     // If host is an IP address, it's path-style
     if (/^[0-9.]+$/.test(host)) {
         return undefined;
     }
-
```
LaurenSpiegel: Do we have to check whether host is an ipv6 address as well as ipV4 on line 26?

dora-korpar: That regex should also catch IPv6 addresses!

LaurenSpiegel: nifty!

dora-korpar: Oh shoot, unless they're condensed with colons

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go2

dora-korpar:
Passed after a couple rebuilds:
http://ci.ironmann.io/gh/scality/Integration/7910

dora-korpar:
@ironman-machine happy holidays

electrachong:
happy holidays @dora-korpar! :D

dora-korpar:
@ironman-machine happy new year

dora-korpar:
@ironman-machine get to it

dora-korpar:
@ironman-machine happy morning

dora-korpar:
@ironman-machine hup hup

PR #486, opened 2016-12-22
FIX: Handle split line breaks in metadata chunks

LaurenSpiegel:
Fixes https://github.com/scality/S3/issues/487


LaurenSpiegel:
@ironman-machine check it

PR #488, opened 2016-12-22
Add default contentLength field in log

dora-korpar:
Addresses issue #485

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -82,6 +82,8 @@ export default function routes(req, res, logger) {
     log.addDefaultFields({
         bucketName: req.bucketName,
         objectKey: req.objectKey,
+        bytesReceived: req.parsedContentLength || 0,
+        bodyLength: parseInt(req.headers['content-length'], 10) || 0,
     });
     // if empty name and request not a list Buckets
     if (!req.bucketName &&
--- a/lib/routes/routesUtils.js
+++ b/lib/routes/routesUtils.js
@@ -59,7 +59,11 @@ function okHeaderResponse(headers, response, httpCode, log) {
  * @return {object} response - response object with additional headers
  */
 function okXMLResponse(xml, response, log, additionalHeaders) {
+    const bytesSent = Buffer.byteLength(xml);
     log.trace('sending success xml response');
+    log.addDefaultFields({
+        bytesSent,
+    });
     setCommonResponseHeaders(additionalHeaders, response, log);
     response.writeHead(200, { 'Content-type': 'application/xml' });
     log.debug('response http code', { httpCode: 200 });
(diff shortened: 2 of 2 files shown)


dora-korpar:
@ironman-machine blast off

dora-korpar:
@ironman-machine Let's go, Jarvis

lib/routes.js
```diff
@@ -82,6 +82,8 @@ export default function routes(req, res, logger) {
     log.addDefaultFields({
         bucketName: req.bucketName,
         objectKey: req.objectKey,
+        bytesReceived: req.parsedContentLength,
```
LaurenSpiegel: should be || 0 here too so not sending NAN on a get, for instance.

lib/utils.js
```diff
@@ -226,11 +226,10 @@ utils.normalizeRequest = function normalizeRequest(request) {
     // For streaming v4 auth, the total body content length
     // without the chunk metadata is sent as
     // the x-amz-decoded-content-length
-    const contentLength = request.headers['x-amz-decoded-content-length'] ?
-        request.headers['x-amz-decoded-content-length'] :
+    request.contentLength = request.headers['x-amz-decoded-content-length'] ||
```
LaurenSpiegel: I wouldn't add a request.contentLength to the request object.  I would leave this file alone and when you are pulling the body size, I would use request.headers['content-length']

lib/routes.js
```diff
@@ -82,6 +82,8 @@ export default function routes(req, res, logger) {
...
+        bodyLength: req.contentLength || 0,
```
LaurenSpiegel: use request.headers['content-length'] here

lib/routes/routesUtils.js
```diff
@@ -59,9 +59,14 @@ function okHeaderResponse(headers, response, httpCode, log) {
...
 function okXMLResponse(xml, response, log, additionalHeaders) {
+    const bytesSent = Buffer.byteLength(xml);
     log.trace('sending success xml response');
+    log.addDefaultFields({
+        bytesSent,
+    });
     setCommonResponseHeaders(additionalHeaders, response, log);
-    response.writeHead(200, { 'Content-type': 'application/xml' });
+    response.writeHead(200, { 'Content-type': 'application/xml',
+        'Content-length': bytesSent });
```
LaurenSpiegel: let's actually not add the content length header since aws returns transfer encoding (that's why your test was failing)

lib/routes/routesUtils.js
```diff
@@ -92,9 +97,15 @@ function errorXMLResponse(errCode, response, log) {
...
+    const xmlStr = xml.join('');
+    const bytesSent = Buffer.byteLength(xmlStr);
+    log.addDefaultFields({
+        bytesSent,
+    });
     setCommonResponseHeaders(null, response, log);
-    response.writeHead(errCode.code, { 'Content-type': 'application/xml' });
-    return response.end(xml.join(''), 'utf8', () => {
+    response.writeHead(errCode.code, { 'Content-type': 'application/xml',
+        'Content-length': bytesSent });
```
LaurenSpiegel: same here on content-length.  let's not send the header.

dora-korpar:
@ironman-machine check it

PR #497, opened 2016-12-27
FT: Implement put bucket CORS api

electrachong:
Reference: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTcors.html

Depends on scality/Arsenal#211

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCors.js
+++ b/lib/api/apiUtils/bucket/bucketCors.js
@@ -1,0 +1,320 @@
+import { parseString } from 'xml2js';
+
+import { errors } from 'arsenal';
+
+/*
+    Format of xml request:
+
+    <CORSConfiguration>
+        <CORSRule>
+            <AllowedOrigin>http://www.example.com</AllowedOrigin>
+            <AllowedMethod>PUT</AllowedMethod>
+            <AllowedMethod>POST</AllowedMethod>
+            <AllowedMethod>DELETE</AllowedMethod>
+            <AllowedHeader>*</AllowedHeader>
+            <MaxAgeSeconds>3000</MaxAgeSec>
+            <ExposeHeader>x-amz-server-side-encryption</ExposeHeader>
+        </CORSRule>
+        <CORSRule>
+            <AllowedOrigin>*</AllowedOrigin>
+            <AllowedMethod>GET</AllowedMethod>
+            <AllowedHeader>*</AllowedHeader>
+            <MaxAgeSeconds>3000</MaxAgeSeconds>
+        </CORSRule>
+    </CORSConfiguration>
+*/
+
+const customizedErrs = {
+    numberRules: 'The number of CORS rules should not exceed allowed limit ' +
+    'of 100 rules.',
+    originAndMethodExist: 'Each CORSRule must identify at least one origin ' +
+    'and one method.',
+};
+
+// Helper validation methods
+export const _validator = {
+    /** _validator.validateNumberWildcards - check if string has multiple
+    *   wildcards
+    @param {string} string - string to check for multiple wildcards
+    @return {boolean} - whether more than one wildcard in string
+    */
+    validateNumberWildcards(string) {
+        const firstIndex = string.indexOf('*');
+        if (firstIndex !== -1) {
+            return (string.indexOf('*', firstIndex + 1) === -1);
+        }
+        return true;
+    },
+    /** _validator.validateID - check value of optional ID
+    * @param {string[]} id - array containing id string
+    * @return {(Error|true|undefined)} - Arsenal error on failure, true on
+    *   success, undefined if ID does not exist
+    */
+    validateID(id) {
+        if (!id) {
+            return undefined; // to indicate ID does not exist
+        }
+        if (!Array.isArray(id) || id.length !== 1
+        || typeof id[0] !== 'string') {
+            return errors.MalformedXML;
+        }
+        if (id[0] === '') {
+            return undefined;
+        }
+        return true;
+    },
+    /** _validator.validateMaxAgeSeconds - check value of optional MaxAgeSeconds
+    * @param {string[]} seconds - array containing number string
+    * @return {(Error|parsedValue|undefined)} - Arsenal error on failure, parsed
+    *   value if valid, undefined if MaxAgeSeconds does not exist
+    */
+    validateMaxAgeSeconds(seconds) {
+        if (!seconds) {
+            return undefined;
+        }
+        if (!Array.isArray(seconds) || seconds.length !== 1) {
+            return errors.MalformedXML;
+        }
+        if (seconds[0] === '') {
+            return undefined;
+        }
+        const parsedValue = parseInt(seconds[0], 10);
+        const errMsg = `MaxAgeSeconds "${seconds[0]}" is not a valid value.`;
+        if (isNaN(parsedValue) || parsedValue < 0) {
+            return errors.MalformedXML.customizeDescription(errMsg);
+        }
+        return parsedValue;
+    },
+    /** _validator.validateNumberRules - return if number of rules exceeds 100
+    * @param {number} length - array containing number string
+    * @return {(Error|true)} - Arsenal error on failure, true on success
+    */
+    validateNumberRules(length) {
+        if (length > 100) {
+            return errors.InvalidRequest
+            .customizeDescription(customizedErrs.numberRules);
+        }
+        return true;
+    },
+    /** _validator.validateOriginAndMethodExist
+    * @param {string[]} allowedMethods - array of AllowedMethod's
+    * @param {string[]} allowedOrigins - array of AllowedOrigin's
+    * @return {(Error|true)} - Arsenal error on failure, true on success
+    */
+    validateOriginAndMethodExist(allowedMethods, allowedOrigins) {
+        if (allowedOrigins && allowedMethods &&
+        Array.isArray(allowedOrigins) &&
+        Array.isArray(allowedMethods) &&
+        allowedOrigins.length > 0 &&
+        allowedMethods.length > 0) {
+            return true;
+        }
+        return errors.MalformedXML
+            .customizeDescription(customizedErrs.originAndMethodExist);
+    },
+    /** _validator.validateMethods - check values of AllowedMethod's
+    * @param {string[]} methods - array of AllowedMethod's
+    * @return {(Error|true)} - Arsenal error on failure, true on success
+    */
+    validateMethods(methods) {
+        let invalidMethod;
+        function isValidMethod(method) {
+            const acceptedValues = ['GET', 'PUT', 'HEAD', 'POST', 'DELETE'];
+            if (acceptedValues.indexOf(method) !== -1) {
+                return true;
+            }
+            invalidMethod = method;
+            return false;
+        }
+        if (!methods.every(isValidMethod)) {
+            const errMsg = 'Found unsupported HTTP method in CORS config. ' +
+            `Unsupported method is "${invalidMethod}"`;
+            return errors.InvalidRequest.customizeDescription(errMsg);
+        }
+        return true;
+    },
+    /** _validator.validateAllowedOriginsOrHeaders - check values
+    * @param {string[]} elementArr - array of elements to check
+    * @param {string} typeElement - type of element being checked
+    * @return {(Error|true)} - Arsenal error on failure, true on success
+    */
+    validateAllowedOriginsOrHeaders(elementArr, typeElement) {
+        for (let i = 0; i < elementArr.length; i++) {
+            const element = elementArr[i];
+            if (typeof element !== 'string' || element === '') {
+                return errors.MalformedXML;
+            }
+            if (!this.validateNumberWildcards(element)) {
+                const errMsg = `${typeElement} "${element}" can not have ` +
+                'more than one wildcard.';
+                return errors.InvalidRequest.customizeDescription(errMsg);
+            }
+        }
+        return true;
+    },
+    /** _validator.validateAllowedHeaders - check values of AllowedHeader's
+    * @param {string[]} headers - array of AllowedHeader's
+    * @return {(Error|true|undefined)} - Arsenal error on failure, true if
+    *   valid, undefined if optional AllowedHeader's do not exist
+    */
+    validateAllowedHeaders(headers) {
+        if (!headers) {
+            return undefined; // to indicate AllowedHeaders do not exist
+        }
+        if (!Array.isArray(headers) || headers.length === 0) {
+            return errors.MalformedXML;
+        }
+        const result =
+            this.validateAllowedOriginsOrHeaders(headers, 'AllowedHeader');
+        if (result instanceof Error) {
+            return result;
+        }
+        return true;
+    },
+    /** _validator.validateExposeHeaders - check values of ExposeHeader's
+    * @param {string[]} headers - array of ExposeHeader's
+    * @return {(Error|true|undefined)} - Arsenal error on failure, true if
+    *   valid, undefined if optional ExposeHeader's do not exist
+    */
+    validateExposeHeaders(headers) {
+        if (!headers) {
+            return undefined; // indicate ExposeHeaders do not exist
+        }
+        if (!Array.isArray(headers) || headers.length === 0) {
+            return errors.MalformedXML;
+        }
+        for (let i = 0; i < headers.length; i++) {
+            const header = headers[i];
+            if (typeof header !== 'string') {
+                return errors.MalformedXML;
+            }
+            if (header.indexOf('*') !== -1) {
+                const errMsg = `ExposeHeader ${header} contains a wildcard. ` +
+                'Wildcards are currently not supported for ExposeHeader.';
+                return errors.InvalidRequest.customizeDescription(errMsg);
+            }
+            if (!/^[A-Za-z0-9-]*$/.test(header)) {
+                const errMsg = `ExposeHeader ${header} contains invalid ` +
+                'character.';
+                return errors.InvalidRequest.customizeDescription(errMsg);
+            }
+        }
+        return true;
+    },
+};
+
+/** _validateCorsXml - Validate XML, returning an error if any part is not valid
+* @param {object[]} rules - CORSRule collection parsed from xml to be validated
+* @param {string[]} [rules[].ID] - optional id to identify rule
+* @param {string[]} rules[].AllowedMethod - methods allowed for CORS
+* @param {string[]} rules[].AllowedOrigin - origins allowed for CORS
+* @param {string[]} [rules[].AllowedHeader] - headers allowed in an OPTIONS
+* request via the Access-Control-Request-Headers header
+* @param {string[]} [rules[].MaxAgeSeconds] - seconds browsers should cache
+* OPTIONS response
+* @param {string[]} [rules[].ExposeHeader] - headers exposed to applications
+* @return {(Error|object)} - return cors object on success; error on failure
+*/
+function _validateCorsXml(rules) {
+    const cors = [];
+    let result;
+
+    if (rules.length > 100) {
+        return errors.InvalidRequest
+            .customizeDescription(customizedErrs.numberRules);
+    }
+    for (let i = 0; i < rules.length; i++) {
+        const rule = rules[i];
+        const corsRule = {};
+
+        result = _validator.validateOriginAndMethodExist(rule.AllowedMethod,
+            rule.AllowedOrigin);
+        if (result instanceof Error) {
+            return result;
+        }
+
+        result = _validator.validateMethods(rule.AllowedMethod);
+        if (result instanceof Error) {
+            return result;
+        }
+        corsRule.allowedMethods = rule.AllowedMethod;
+
+        result = _validator.validateAllowedOriginsOrHeaders(rule.AllowedOrigin,
+            'AllowedOrigin');
+        if (result instanceof Error) {
+            return result;
+        }
+        corsRule.allowedOrigins = rule.AllowedOrigin;
+
+        result = _validator.validateID(rule.ID);
+        if (result instanceof Error) {
+            return result;
+        } else if (result) {
+            corsRule.id = rule.ID[0];
+        }
+
+        result = _validator.validateAllowedHeaders(rule.AllowedHeader);
+        if (result instanceof Error) {
+            return result;
+        } else if (result) {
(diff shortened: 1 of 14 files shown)


electrachong:
@ironman-machine test

lib/api/bucketPutCors.js
```diff
@@ -0,0 +1,84 @@
...
+                if (bucketShield(bucket, requestType)) {
+                    return next(errors.NoSuchBucket);
+                }
+                log.trace('found bucket in metadata');
+                return next(null, bucket, rules);
+            });
+        },
+        function validateBucketAuthorization(bucket, rules, next) {
+            if (!isBucketAuthorized(bucket, requestType, canonicalID)) {
+                log.debug('access denied for user on bucket', {
```
LaurenSpiegel: access denied for account on bucket

lib/metadata/BucketInfo.js
```diff
@@ -2,7 +2,7 @@ import assert from 'assert';
 import { WebsiteConfiguration } from './WebsiteConfiguration';
 
 // WHEN UPDATING THIS NUMBER, UPDATE MODELVERSION.MD CHANGELOG
-const modelVersion = 4;
+const modelVersion = 5;
```
LaurenSpiegel: if you rebase, this shouldn't show as a change?

lib/metadata/BucketInfo.js
```diff
@@ -259,6 +274,31 @@ export default class BucketInfo {
...
+    getCors() {
+        return this._cors;
+    }
+    /**
+     * Set cors resource
+     * @param {object[]} rules - collection of CORS rules
+     * @param {string} [rules.id] - optional id to identify rule
+     * @param {string[]} rules[].allowedMethods - methods allowed for CORS
+     * @param {string[]} rules[].allowedOrigins - origins allowed for CORS
+     * @param {string[]} [rules[].allowedHeaders] - headers allowed in a OPTIONS
```
LaurenSpiegel: an OPTIONs

lib/routes/routePUT.js
```diff
@@ -90,6 +90,14 @@ export default function routePUT(request, response, log, statsClient) {
                     return routesUtils.responseNoBody(err, null, response, 200,
                         log);
                 });
+            } else if (request.query.cors !== undefined) {
+                api.callApiMethod('bucketPutCors', request, log, err => {
+                    statsReport500(err, statsClient);
+                    // TODO: pushMetrics(err, log, utapi, 'bucketPutCors',
```
LaurenSpiegel: actually this should be in the api now and I believe because of the refactor, it can be added without first updating utapi

package.json
```diff
@@ -19,7 +19,7 @@
   },
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
-    "arsenal": "scality/Arsenal",
+    "arsenal": "scality/Arsenal#ft/corsAPI",
```
LaurenSpiegel: don't need this anymore :)

tests/unit/api/bucketPutCors.js
```diff
@@ -0,0 +1,139 @@
...
+    allowedHeaders: ['*'],
+    maxAgeSeconds: 3000,
+    exposeHeaders: ['x-amz-server-side-encryption'] },
+  { allowedMethods: ['GET'],
+    allowedOrigins: ['*'],
+    allowedHeaders: ['*'],
+    maxAgeSeconds: 3000 },
+];
+
+function _getPutCorsRequest(xml) {
```
LaurenSpiegel: createPutCorsRequest?

tests/unit/api/bucketPutCors.js
```diff
@@ -0,0 +1,139 @@
...
+        post: xml,
+    };
+    if (xml) {
+        request.headers['content-md5'] = crypto.createHash('md5')
+            .update(request.post, 'utf8').digest('base64');
+    }
+    return request;
+}
+
+function _getCorsXml(arrayRules) {
```
LaurenSpiegel: constructCorsXml

tests/unit/api/bucketPutCors.js
```diff
@@ -0,0 +1,139 @@
...
+    maxAgeSeconds: 3000 },
+];
+
+function _getPutCorsRequest(xml) {
+    const request = {
+        bucketName,
+        headers: {
+            host: `${bucketName}.s3.amazonaws.com`,
+        },
+        url: '/?website',
```
LaurenSpiegel: why website?

tests/functional/s3cmd/tests.js
```diff
@@ -691,22 +691,14 @@ describe('s3cmd put, get and delete object with spaces ' +
...
     it('should find that policy has a value of none', done => {
         checkRawOutput(['info', `s3://${bucket}`], 'policy', 'none',
         foundIt => {
             assert(foundIt);
             done();
         });
     });
-
-
-    it('should find that cors has a value of none', done => {
```
LaurenSpiegel: might as well test that our implementation of put cors is compatible with s3cmd.  can skip until get is implemented if needed.

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+        <CORSRule>
+            <AllowedOrigin>*</AllowedOrigin>
+            <AllowedMethod>GET</AllowedMethod>
+            <AllowedHeader>*</AllowedHeader>
+            <MaxAgeSeconds>3000</MaxAgeSeconds>
+        </CORSRule>
+    </CORSConfiguration>
+*/
+
+const errs = {
```
electrachong: :heavy_check_mark:

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+    numberRules: 'The number of CORS rules should not exceed allowed limit ' +
+    'of 100 rules.',
+    originAndMethodExist: 'Each CORSRule must identify at least one origin ' +
+    'and one method.',
+};
+
+const _utility = {
```
electrachong: :heavy_check_mark:

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+        }
+        if (seconds[0] === '') {
+            return undefined;
+        }
+        const parsedValue = parseInt(seconds[0], 10);
+        const errMsg = `MaxAgeSeconds "${seconds[0]}" is not a valid value.`;
+        if (isNaN(parsedValue) || parsedValue < 0) {
+            return errors.MalformedXML.customizeDescription(errMsg);
+        }
+        return parsedValue;
```
electrachong: I tried to make them a little more consistent now so that all return 
- true if valid and validated argument exists, 
- undefined if argument does not exist but valid because optional, and 
- Error if invalid. 

The one exception is maxAgeSeconds because I parse the value in the validation function and figure I might as well return it instead of reparsing the int in the caller function.

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+        allowedMethods.length > 0) {
+            return undefined;
+        }
+        return errors.MalformedXML
+            .customizeDescription(errs.originAndMethodExist);
+    },
+    validateMethods(methods) {
+        let invalidMethod;
+        function isValidMethod(method) {
+            if (method !== 'GET' && method !== 'PUT' && method !== 'HEAD' &&
```
electrachong: :heavy_check_mark:

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+            const errMsg = 'Found unsupported HTTP method in CORS config. ' +
+            `Unsupported method is "${invalidMethod}"`;
+            return errors.InvalidRequest.customizeDescription(errMsg);
+        }
+        return undefined;
+    },
+    validateOrigins(origins) {
+        for (let i = 0; i < origins.length; i++) {
+            const origin = origins[i];
+            const errMsg = `AllowedOrigin "${origin}" can not have ` +
```
electrachong: :heavy_check_mark:

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+            'more than one wildcard.';
+            if (typeof origin !== 'string' || origin === '') {
+                return errors.MalformedXML;
+            }
+            const numberWildcards = (origin.match(/\*/g) || []).length;
```
electrachong: I made a utility function using indexOf but have yet to experience the fun of testing to see which is faster. (I assume the utility function but it will be exciting to witness).

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+            return errors.MalformedXML;
+        }
+        for (let i = 0; i < headers.length; i++) {
+            const header = headers[i];
+            if (typeof header !== 'string') {
+                return errors.MalformedXML;
+            }
+            if (header.indexOf('*') !== -1) {
+                const errMsg = `ExposeHeader ${header} contains a wildcard. ` +
+                'Wildcards are currently not supported for ExposeHeader.';
```
electrachong: I copied the error message from AWS, so maybe they are thinking one day it will be supported.
Should I remove the 'currently' to reduce confusion?

LaurenSpiegel: if that's what AWS does, leave it as is

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+        return true; // indicate ExposeHeaders exists
+    },
+};
+
+/** Validate XML, returning an error if any part is not valid
+* @param {object[]} rules - CORSRule collection parsed from xml to be validated
+* @param {string[]} [rules[].ID] - optional id to identify rule
+* @param {string[]} rules[].AllowedMethod - methods allowed for CORS
+* @param {string[]} rules[].AllowedOrigin - origins allowed for CORS
+* @param {string[]} [rules[].AllowedHeader] - headers allowed in a OPTIONS
```
electrachong: ✔️

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+* request via the Access-Control-Request-Headers header
+* @param {string[]} [rules[].MaxAgeSeconds] - seconds browsers should cache
+* OPTIONS response
+* @param {string[]} [rules[].ExposeHeader] - headers expose to applications
```
electrachong: ✔️

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+* @return {(Error|object)} - return cors object on success; error on failure
```
electrachong: ✔️

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+        if (validationRes instanceof Error) {
+            log.debug('xml validation failed', {
+                error: validationRes,
+                method: '_validateCorsXml',
+                xml,
+            });
+            return cb(validationRes);
+        }
+        // if no error, validation returns cors object
+        log.trace('cors configuration', { validationRes });
```
electrachong: ✔️

lib/api/apiUtils/bucket/bucketWebsite.js
```diff
@@ -47,12 +47,12 @@ function _isValidElem(elem) {
 * @param {array} parent - represents xml element to check for child element
 * @param {(string|string[])} requiredElem - name of child element(s)
 * @param {object} [options] - specify additional options
-* @param {boolean} [isList] - indicates if parent is list of children elements,
-* used only in conjunction a singular requiredElem argument
-* @param {boolean} [checkForAll] - return true only if parent element contains
-* all children elements specified in requiredElem; by default, returns true if
-* parent element contains at least one
-* @param {boolean} [validateParent] - validate format of parent element
+* @param {boolean} [options.isList] - indicates if parent is list of children
```
electrachong: I moved this out, I think the changes found the way in by accident.

lib/api/bucketPutCors.js
electrachong: ✔️

lib/metadata/BucketInfo.js
electrachong: Hum...

lib/metadata/BucketInfo.js
electrachong: ✔️

lib/routes/routePUT.js
electrachong: I actually don't think we can add the pushMetric call without modifying Utapi first, I think we still need to add the action to UtapiClient method map here: 
https://github.com/scality/utapi/blob/e79a6798ecb7c363c0423e7c3d66a0fbb5a92e7d/src/lib/UtapiClient.js
Maybe the refactoring will make it so we don't have to update Integration and Utapi at the same time though? Can @bennettbuchanan can confirm what's correct?

package.json
electrachong: ✔️

tests/unit/api/bucketPutCors.js
electrachong: ✔️

tests/unit/api/bucketPutCors.js
electrachong: Thanks for catching the error (result of copy-pasting)

tests/unit/api/bucketPutCors.js
electrachong: ✔️

tests/functional/s3cmd/tests.js
electrachong: This test actually failed when I tested the PR initially, otherwise I wouldn't have known about it.

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,278 @@
...
+                return errors.MalformedXML;
+            }
+            const numberWildcards = (origin.match(/\*/g) || []).length;
+            if (numberWildcards > 1) {
+                return errors.InvalidRequest.customizeDescription(errMsg);
+            }
+        }
+        return undefined;
+    },
+    validateAllowedHeaders(headers) {
```
electrachong: Do you think we should check for invalid symbols (anything non-alphanumeric or a dash) in `AllowedHeader`s? AWS performs this check for `ExposeHeader`s (the message they return is actually `AllowedHeader [value] contains invalid character.` which makes me suspect they intended to check). That's the only reason why I might not convert validateOrigins into a utility check I can also call in `validateAllowedHeaders`, because I'd have to iterate through the list of AllowedHeaders a second time.

On the other hand, the only way I can think to perform that check is to use a regex test (like I did for `ExposeHeader`s) and since AWS doesn't actually perform the check, maybe we don't need to.

electrachong: You can see the changes I made here per our discussion and let me know if you think it's OK:
https://github.com/scality/S3/pull/497/commits/c80d5f7aa0e68ce2149d207da9d192ad0257c8b8?diff=split

electrachong:
@ironman-machine try it

electrachong:
end-to-end passed

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,320 @@
...
+        if (seconds[0] === '') {
+            return undefined;
+        }
+        const parsedValue = parseInt(seconds[0], 10);
+        const errMsg = `MaxAgeSeconds "${seconds[0]}" is not a valid value.`;
+        if (isNaN(parsedValue) || parsedValue < 0) {
+            return errors.MalformedXML.customizeDescription(errMsg);
+        }
+        return parsedValue; // since we had to parse the value already
+        // return the value so it can be saved into corsRule object
```
LaurenSpiegel: I'd get rid of extra comment here.  jsdoc above the function covers it.

lib/api/apiUtils/bucket/bucketCors.js
```diff
@@ -0,0 +1,320 @@
...
+        cors.push(corsRule);
+    }
+    return cors;
+}
+
+/** parseCorsXml - Parse and validate xml body, returning cors object on success
+* @param {string} xml - xml body to parse and validate
+* @param {object} log - Werelogs logger
+* @param {function} cb - callback to server
+* @return {(Error|object)} - return cors object on success; error on failure
```
LaurenSpiegel: actually returns call of callback with arguments err and object

lib/metadata/BucketInfo.js
LaurenSpiegel: It looks like this got messed up in the forward port since the md document already says version 5.  Will you please do a separate cleanup pr that just updates this number?

tests/functional/s3cmd/tests.js
```diff
@@ -691,22 +691,14 @@ describe('s3cmd put, get and delete object with spaces ' +
 });
 
 describe('s3cmd info', () => {
-// test that CORS and POLICY are returned as 'none'
+// test that POLICY are returned as 'none'
```
LaurenSpiegel: POLICY is returned as 'none'

tests/functional/s3cmd/tests.js
LaurenSpiegel: I mean we should keep the s3cmd test for CORS.  You might have to skip it until the get CORS is merged but we shouldn't delete the test.  Instead convert it to show that we get the actual cors value.

tests/unit/api/bucketPutCors.js
```diff
@@ -0,0 +1,261 @@
...
+import bucketPut from '../../../lib/api/bucketPut';
+import bucketPutCors from '../../../lib/api/bucketPutCors';
+import {
+    _validator,
+    parseCorsXml,
+} from '../../../lib/api/apiUtils/bucket/bucketCors';
+import { cleanup,
+    DummyRequestLogger,
+    makeAuthInfo }
+from '../helpers';
```
bennettbuchanan: Why the line breaks here?

tests/unit/api/bucketPutCors.js
```diff
@@ -0,0 +1,261 @@
...
+
+    it('should return MalformedXML if body greater than 64KB', done => {
+        const body = Buffer.alloc(65537); // 64 * 1024 = 65536 bytes
+        const testBucketPutCorsRequest = _createPutCorsRequest(body);
+        _testPutBucketCors(authInfo, testBucketPutCorsRequest,
+            log, 'MalformedXML', done);
+    });
+
+    it('should return InvalidRequest if more than one MaxAgeSeconds', done => {
+        const testCors = sampleCors.slice();
```
bennettbuchanan: I don't understand the purpose of this

bennettbuchanan: Is that used to create a copy?

electrachong: Yep, similar to Object.assign() for objects

bennettbuchanan: Perhpas use that method instead because this will alter the original object `sampleCors[0]`

lib/metadata/BucketInfo.js
electrachong: Thanks, I guess we did miss it. Here: https://github.com/scality/S3/pull/506

tests/functional/s3cmd/tests.js
electrachong: Got it

tests/unit/api/bucketPutCors.js
electrachong: hmm, I see. Thanks for the tip.

lib/api/apiUtils/bucket/bucketCors.js
electrachong: ✔️

lib/api/apiUtils/bucket/bucketCors.js
electrachong: ✔️ thanks for catching it

tests/functional/s3cmd/tests.js
electrachong: ✔️

tests/unit/api/bucketPutCors.js
electrachong: I removed them; is it better to you?

tests/unit/api/bucketPutCors.js
electrachong: I updated with changes per your advice

bennettbuchanan: Ah, nice! Why not pass the copy in as you had done before, and keep `_constructCorsXml` the same? Maybe you are planning to reuse it for other customized objects? It's very minor, but that would obviate the step for those other tests that don't need to create a copy.

electrachong: Yep, I get your point. I mostly did that because I wanted to use the same helper functions and test cors object to write a unit test for deleteBucketCors, where I had to put a cors configuration first. :)

electrachong: https://github.com/scality/S3/pull/507/files#diff-bd471b7107f311a89fabfd48e44e5594

electrachong:

tests/unit/api/bucketPutCors.js
bennettbuchanan: I see it's a class now—got it. LGTM.

tests/unit/helpers.js
```diff
@@ -230,6 +230,81 @@ export class DummyRequestLogger {
...
+    }
+
+    constructXml() {
+        const xml = [];
+        xml.push('<CORSConfiguration>');
+        this._cors.forEach(rule => {
+            xml.push('<CORSRule>');
+            ['allowedMethods', 'allowedOrigins', 'allowedHeaders',
+            'exposeHeaders', 'maxAgeSeconds']
+            .forEach(elementArr => {
```
bennettbuchanan: Minor: `elementArr` should probably be named something else as it is not an array.

electrachong:
@ironman-machine you good?

tests/unit/helpers.js
electrachong: OK!

electrachong:
@LaurenSpiegel end-to-end passed, PR is ready for merge

PR #499, opened 2016-12-28
FIX: Allow for key called slash

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine slash it

LaurenSpiegel:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/8017

PR #507, opened 2017-01-04
FT: Add bucketDeleteCors to API

electrachong:
Depends on scality/S3#497 (putBucketCors)

### DIFF
--- a/lib/api/bucketDeleteCors.js
+++ b/lib/api/bucketDeleteCors.js
@@ -1,0 +1,61 @@
+import { errors } from 'arsenal';
+
+import bucketShield from './apiUtils/bucket/bucketShield';
+import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
+import metadata from '../metadata/wrapper';
+// TODO: import { pushMetric } from '../utapi/utilities';
+
+const requestType = 'bucketOwnerAction';
+
+/**
+ * Bucket Delete CORS - Delete bucket cors configuration
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+export default function bucketDeleteCors(authInfo, request, log, callback) {
+    const bucketName = request.bucketName;
+    const canonicalID = authInfo.getCanonicalID();
+
+    return metadata.getBucket(bucketName, log, (err, bucket) => {
+        if (err) {
+            log.debug('metadata getbucket failed', { error: err });
+            return callback(err);
+        }
+        if (bucketShield(bucket, requestType)) {
+            return callback(errors.NoSuchBucket);
+        }
+        log.trace('found bucket in metadata');
+
+        if (!isBucketAuthorized(bucket, requestType, canonicalID)) {
+            log.debug('access denied for user on bucket', {
+                requestType,
+                method: 'bucketDeleteCors',
+            });
+            return callback(errors.AccessDenied);
+        }
+
+        const cors = bucket.getCors();
+        if (!cors) {
+            log.trace('no existing cors configuration', {
+                method: 'bucketDeleteCors',
+            });
+            // TODO: add 'deleteBucketCors' to Utapi Client action map
+            // pushMetric('deleteBucketCors', log, { bucket: bucketName });
+            return callback();
+        }
+
+        log.trace('deleting cors configuration in metadata');
+        bucket.setCors(null);
+        return metadata.updateBucket(bucketName, bucket, log, err => {
+            if (err) {
+                return callback(err);
+            }
+            // TODO: pushMetric('deleteBucketCors', log,
+            // { bucket: bucketName });
+            return callback();
+        });
+    });
+}
--- a/tests/functional/aws-node-sdk/test/bucket/deleteCors.js
+++ b/tests/functional/aws-node-sdk/test/bucket/deleteCors.js
@@ -1,0 +1,91 @@
+import assert from 'assert';
+
+import withV4 from '../support/withV4';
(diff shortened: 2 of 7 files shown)


lib/api/bucketDeleteCors.js
```diff
@@ -0,0 +1,52 @@
...
+                method: 'bucketDeleteCors',
+            });
+            return callback(errors.AccessDenied);
+        }
+
+        const cors = bucket.getCors();
+        if (!cors) {
+            log.trace('no existing cors configuration', {
+                method: 'bucketDeleteCors',
+            });
```
LaurenSpiegel: I assume AWS does not return an error if no configuration already there?

tests/functional/aws-node-sdk/test/bucket/deleteCors.js
```diff
@@ -0,0 +1,79 @@
...
+            describe('with existing cors configuration', () => {
+                beforeEach(done => {
+                    s3.putBucketCors({ Bucket: bucketName,
+                    CORSConfiguration: sampleCors }, done);
+                });
+
+                it('should delete bucket configuration successfully', done => {
+                    s3.deleteBucketCors({ Bucket: bucketName }, err => {
+                        assert.strictEqual(err, null,
+                            `Found unexpected err ${err}`);
```
LaurenSpiegel: please add a follow up get to confirm it is not there

lib/api/bucketDeleteCors.js
electrachong: Yep, it's one of the functional tests as well (passes with AWS_ON_AIR=true).

tests/functional/aws-node-sdk/test/bucket/deleteCors.js
electrachong: I added a follow-up

electrachong:
@ironman-machine try it

electrachong:
@ironman-machine again, what's up? >.<

electrachong:
end-to-end passed

lib/api/bucketDeleteCors.js
```diff
@@ -0,0 +1,52 @@
+import { errors } from 'arsenal';
+
+import bucketShield from './apiUtils/bucket/bucketShield';
+import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
+import metadata from '../metadata/wrapper';
+// import { pushMetric } from '../utapi/utilities';
```
ploki: maybe you should remove this disabled code

lib/api/bucketDeleteCors.js
```diff
@@ -0,0 +1,52 @@
...
+
+const requestType = 'bucketOwnerAction';
+
+export default function bucketDeleteCors(authInfo, request, log, callback) {
```
ploki: a bit of documentation on this exported function would be in order

tests/functional/aws-node-sdk/test/bucket/deleteCors.js
```diff
@@ -0,0 +1,87 @@
...
+                            assert.strictEqual(err.code,
+                                'NoSuchCORSConfiguration');
+                            assert.strictEqual(err.statusCode, 404);
+                            return done();
+                        });
+                    });
+                });
+
+                // Skip if AWS because AWS Node SDK raises CredentialsError
+                // before letting the request hit the API
```
ploki: how the AWS Node SDK knows about this without hitting the API?

What's the difference between CredentialsError and 403?

Isn't it a difference in the order of evaluation of error cases between AWS and this implementation?

electrachong: I assumed it was a sdk error because CredentialsError is unfamiliar and we usually return AccessDenied, but thanks for probing so I would confirm in more detail. This is the error returned:

`{ [SharedIniFileCredentialsProviderFailure: Missing credentials in config]
  message: 'Missing credentials in config',
  code: 'CredentialsError',
  time: Tue Jan 10 2017 15:43:12 GMT-0800 (PST),
  originalError: 
   { message: 'Could not load credentials from SharedIniFileCredentials',
     code: 'CredentialsError',
     time: Tue Jan 10 2017 15:43:12 GMT-0800 (PST),
     originalError: 
      { message: 'Profile lisa not found in /Users/electra/.aws/scality',
        code: 'SharedIniFileCredentialsProviderFailure',
        time: Tue Jan 10 2017 15:43:12 GMT-0800 (PST) } } }`

The node sdk relies on a config file to pull authentication data; when testing S3 Server we've configured to pull this data from conf/authdata.json in the s3 repo. When testing with AWS_ON_AIR it pulls from ~/.aws/scality, which should contain matching profiles with real AWS credentials for the tests to pass. 

In order for this test to always pass with AWS_ON_AIR we would have to assume everyone who runs the tests has updated their local ~/aws/scality file to contain a second set of credentials in addition to the default under a 'lisa' profile. I will add a comment to explain this in case someone does want to run the test but think it's safer to keep `itSkipIfAWS` setting by default.

lib/api/bucketDeleteCors.js
electrachong: Thanks for pointing that out, I missed that

tests/functional/aws-node-sdk/test/bucket/deleteCors.js
ploki: OK, thanks for the explanation

lib/api/bucketDeleteCors.js
electrachong: The code is commented out because I will need to update Utapi and Integration both to make this function call, and there's several cors api that will need similar pull requests. It's easier to merge this first and then make joint prs to address for all cors api in S3 (+ Integration & Utapi). So it's commented out for now, but is intended to be uncommented later. :)

PR #518, opened 2017-01-10
FT: Add OPTIONS route for CORS response

electrachong:
Depends on scality/S3#507 (deleteBucketCors and its prerequisite putBucketCors)

Reference: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTOPTIONSobject.html

Thanks @nicolas2bert for the bulk of the tests!

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -16,6 +16,7 @@ import bucketPutACL from './bucketPutACL';
 import bucketPutCors from './bucketPutCors';
 import bucketPutVersioning from './bucketPutVersioning';
 import bucketPutWebsite from './bucketPutWebsite';
+import corsPreflight from './corsPreflight';
 import completeMultipartUpload from './completeMultipartUpload';
 import initiateMultipartUpload from './initiateMultipartUpload';
 import listMultipartUploads from './listMultipartUploads';
@@ -42,8 +43,9 @@ auth.setHandler(vault);
 
 const api = {
     callApiMethod(apiMethod, request, log, callback, locationConstraint) {
-        // no need to check auth on website requests
-        if (apiMethod === 'websiteGet' || apiMethod === 'websiteHead') {
+        // no need to check auth on website or cors preflight requests
+        if (apiMethod === 'websiteGet' || apiMethod === 'websiteHead' ||
+        apiMethod === 'corsPreflight') {
             return this[apiMethod](request, log, callback);
         }
         let sourceBucket;
@@ -110,6 +112,7 @@ const api = {
     bucketPutCors,
     bucketPutVersioning,
     bucketPutWebsite,
+    corsPreflight,
     completeMultipartUpload,
     initiateMultipartUpload,
     listMultipartUploads,
--- a/lib/api/apiUtils/object/corsResponse.js
+++ b/lib/api/apiUtils/object/corsResponse.js
@@ -1,0 +1,127 @@
+/** _matchesValue - compare two values to determine if they match
+* @param {string} allowedValue - an allowed value in a CORS rule;
+*    may contain wildcards
+* @param {string} value - value from CORS request
+* @return {boolean} - true/false
+*/
+function _matchesValue(allowedValue, value) {
+    const wildcardIndex = allowedValue.indexOf('*');
+    // If no wildcards, simply return whether strings are equal
+    if (wildcardIndex === -1) {
+        return allowedValue === value;
+    }
+    // Otherwise make sure substrings match expected substrings before
+    // and after the wildcard
+    const beginValue = allowedValue.substring(0, wildcardIndex);
+    const endValue = allowedValue.substring(wildcardIndex + 1);
+    return (value.startsWith(beginValue) && value.endsWith(endValue));
+}
+
+/** _matchesOneOf - check if header matches any AllowedHeaders of a rule
+* @param {string[]} allowedHeaders - headers allowed in CORS rule
+* @param {string} header - header from CORS request
+* @return {boolean} - true/false
+*/
+function _matchesOneOf(allowedHeaders, header) {
+    return allowedHeaders.some(allowedHeader =>
+        // AllowedHeaders may have been stored with uppercase letters
+        // during putBucketCors; ignore case when searching for match
+        _matchesValue(allowedHeader.toLowerCase(), header));
+}
+
+/** _headersMatchRule - check if headers match AllowedHeaders of rule
+* @param {string[]} headers - the value of the 'Access-Control-Allow-Headers'
+*   in an OPTIONS request and actual request header keys in any other request
+* @param {string[]} allowedHeaders - AllowedHeaders of a CORS rule
+* @return {boolean} - true/false
+*/
+function _headersMatchRule(headers, allowedHeaders) {
+    if (!allowedHeaders) {
+        return false;
+    }
+    if (!headers.every(header => _matchesOneOf(allowedHeaders, header))) {
+        return false;
+    }
+    return true;
+}
+
+/** _findCorsRule - Return first matching rule in cors rules that permits
+*   CORS request
+* @param {object[]} rules - array of rules
+* @param {string} [rules.id] - optional id to identify rule
+* @param {string[]} rules[].allowedMethods - methods allowed for CORS
+* @param {string[]} rules[].allowedOrigins - origins allowed for CORS
+* @param {string[]} [rules[].allowedHeaders] - headers allowed in an
+*   OPTIONS request via the Access-Control-Request-Headers header
+* @param {number} [rules[].maxAgeSeconds] - seconds browsers should cache
+*   OPTIONS response
+* @param {string[]} [rules[].exposeHeaders] - headers to expose to external
+*   applications
+* @param {string} origin - origin of CORS request
+* @param {string} method - 'Access-Control-Request-Method' header value in
+*   an OPTIONS request and the actual method in any other request
+* @param {string[]} [headers] - the value of the 'Access-Control-Allow-Headers'
+*   in an OPTIONS request and actual request header keys in any other request
+* @return {(null|object)} - matching rule if found; null if no match
+*/
+export function findCorsRule(rules, origin, method, headers) {
+    return rules.find(rule => {
+        if (rule.allowedMethods.indexOf(method) === -1) {
+            return false;
+        } else if (!rule.allowedOrigins.some(allowedOrigin =>
+            _matchesValue(allowedOrigin, origin))) {
+            return false;
+        } else if (headers &&
+            !_headersMatchRule(headers, rule.allowedHeaders)) {
+            return false;
+        }
+        return true;
+    });
+}
+
+/** _gatherResHeaders - Collect headers to return in response
+* @param {object} rule - array of rules
+* @param {string} [rule.id] - optional id to identify rule
+* @param {string[]} rule[].allowedMethods - methods allowed for CORS
+* @param {string[]} rule[].allowedOrigins - origins allowed for CORS
+* @param {string[]} [rule[].allowedHeaders] - headers allowed in an
+*   OPTIONS request via the Access-Control-Request-Headers header
+* @param {number} [rule[].maxAgeSeconds] - seconds browsers should cache
+*   OPTIONS response
+* @param {string[]} [rule[].exposeHeaders] - headers to expose to external
+*   applications
+* @param {string} origin - origin of CORS request
+* @param {string} method - 'Access-Control-Request-Method' header value in
+*   an OPTIONS request and the actual method in any other request
+* @param {string[]} [headers] - the value of the 'Access-Control-Allow-Headers'
+*   in an OPTIONS request and actual request header keys in any other request
+* @return {object} resHeaders - headers to include in response
+*/
+export function gatherCorsResHeaders(rule, origin, method, headers) {
+    const resHeaders = {
+        'access-control-max-age': rule.maxAgeSeconds,
+        'access-control-allow-methods': rule.allowedMethods.join(', '),
+        'content-length': '0',
+        'date': new Date().toUTCString(),
+        'vary':
+        'Origin, Access-Control-Request-Headers, Access-Control-Request-Method',
+    };
+    // send back '*' if any origin allowed; otherwise send back
+    // request Origin value
+    if (rule.allowedOrigins.indexOf('*') > -1) {
+        resHeaders['access-control-allow-origin'] = '*';
+    } else {
+        resHeaders['access-control-allow-origin'] = origin;
+        // can only set allow credentials to true if not returning wildcard
+        // for 'access-control-allow-origin'
+        resHeaders['access-control-allow-credentials'] = true;
+    }
+    if (headers) {
+        resHeaders['access-control-allow-headers'] = headers.join(', ');
+    }
+    if (rule.exposeHeaders) {
+        resHeaders['access-control-expose-headers'] =
+            rule.exposeHeaders.join(', ');
+    }
+    return resHeaders;
+}
--- a/lib/api/corsPreflight.js
+++ b/lib/api/corsPreflight.js
@@ -1,0 +1,82 @@
+import { errors } from 'arsenal';
+
+import metadata from '../metadata/wrapper';
+import bucketShield from './apiUtils/bucket/bucketShield';
+import { findCorsRule,
+gatherCorsResHeaders } from './apiUtils/object/corsResponse';
+// import { pushMetric } from '../utapi/utilities';
+
+const requestType = 'objectGet';
+
+const customizedErrs = {
+    corsNotEnabled: 'CORSResponse: CORS is not enabled for this bucket.',
+    notAllowed: 'CORSResponse: This CORS request is not allowed. ' +
+    'This is usually because the evalution of Origin, request method / ' +
+    'Access-Control-Request-Method or Access-Control-Request-Headers ' +
+    'are not whitelisted by the resource\'s CORS spec.',
+};
+
+/** corsPreflight - handle preflight CORS requests
+* @param  {object} request - http request object
+* @param  {function} log - Werelogs request logger
+* @param  {function} callback - callback to respond to http request
+*  with either error code or 200 response
+* @return {undefined}
+*/
+export default function corsPreflight(request, log, callback) {
+    log.debug('processing request', { method: 'corsPreflight' });
+
+    const bucketName = request.bucketName;
+    const corsOrigin = request.headers.origin;
+    const corsMethod = request.headers['access-control-request-method'];
+    const corsHeaders = request.headers['access-control-request-headers'] ?
+        request.headers['access-control-request-headers'].replace(/ /g, '')
+            .split(',').reduce((resultArr, value) => {
+                // remove empty values and convert values to lowercase
+                if (value !== '') {
+                    resultArr.push(value.toLowerCase());
+                }
+                return resultArr;
+            }, []) : null;
+
+    return metadata.getBucket(bucketName, log, (err, bucket) => {
+        if (err) {
+            log.debug('metadata getbucket failed', { error: err });
+            return callback(err);
+        }
+        if (bucketShield(bucket, requestType)) {
+            return callback(errors.NoSuchBucket);
+        }
+        log.trace('found bucket in metadata');
+
+        const corsRules = bucket.getCors();
+        if (!corsRules) {
+            const err = errors.AccessForbidden
+                .customizeDescription(customizedErrs.corsNotEnabled);
+            log.trace('no existing cors configuration', {
+                error: err,
+                method: 'corsPreflight',
+            });
+            return callback(err);
+        }
+
+        log.trace('finding cors rule');
+        const corsRule = findCorsRule(corsRules, corsOrigin, corsMethod,
+            corsHeaders);
+
+        if (!corsRule) {
+            const err = errors.AccessForbidden
+                .customizeDescription(customizedErrs.notAllowed);
+            log.trace('no matching cors rule', {
+                error: err,
+                method: 'corsPreflight',
+            });
+            return callback(err);
+        }
+
+        const resHeaders = gatherCorsResHeaders(corsRule, corsOrigin,
+            corsMethod, corsHeaders);
+        // TODO: pushMetric('corsPreflight', log, { bucket: bucketName });
+        return callback(null, resHeaders);
+    });
+}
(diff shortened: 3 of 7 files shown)


electrachong:
@ironman-machine why not, give it a go :)

electrachong:
@ironman-machine retry

electrachong:
@ironman-machine

electrachong:
@ironman-machine try it

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,994 @@
...
+        setTimeout((done, err) => {
+            callback(err);
+        }, 5000);
+    } else {
+        callback(err);
+    }
+}
+
+function methodRequest(method, headers, code, headersResponse, done,
+objectKey) {
```
LaurenSpiegel: why not have done cb as last argument?

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,994 @@
...
+        headers,
+        path: objectKey ? `/${bucket}/${objectKey}` : `/${bucket}`,
+    };
+    const req = transport.request(options, res => {
+        const body = [];
+        res.on('data', chunk => {
+            body.push(chunk);
+        });
+        res.on('error', err => {
+            process.stdout.write('err on post response');
```
LaurenSpiegel: carryover?

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,994 @@
...
+            return done(err);
+        });
+        res.on('end', () => {
+            const total = body.join('');
+            if (code) {
+                const message = code === 200 ? '' : `<Code>${code}</Code>`;
+                assert(total.indexOf(message) > -1, `Expected ${message}`);
```
LaurenSpiegel: is it too difficult to just match the total response?

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,994 @@
...
+                    assert.deepEqual(res.headers[key], headersResponse[key],
+                      `error header: ${key}`);
+                });
+            }
+            return done();
+        });
+    });
+
+    req.on('error', err => {
+        process.stdout.write('err from post request');
```
LaurenSpiegel: carryover?

lib/api/api.js
```diff
@@ -42,8 +43,9 @@ auth.setHandler(vault);
 
 const api = {
     callApiMethod(apiMethod, request, log, callback, locationConstraint) {
-        // no need to check auth on website requests
-        if (apiMethod === 'websiteGet' || apiMethod === 'websiteHead') {
+        // no need to check auth on website or cors preflight requests
+        if (apiMethod === 'websiteGet' || apiMethod === 'websiteHead' ||
+        apiMethod === 'corsGetResponse') {
```
electrachong: thanks for bringing this up to me.. it does look like AWS ignores bad credentials:
https://github.com/scality/S3/pull/518/commits/7ec42654715611ac98a65de431185ff5dcfb9062

lib/api/api.js
```diff
@@ -16,6 +16,7 @@ import bucketPutACL from './bucketPutACL';
 import bucketPutCors from './bucketPutCors';
 import bucketPutVersioning from './bucketPutVersioning';
 import bucketPutWebsite from './bucketPutWebsite';
+import corsGetResponse from './corsGetResponse';
```
electrachong: ✔️

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,127 @@
...
+* @param {string} value - value from CORS request
+* @return {boolean} - true/false
+*/
+function _matchesOneOf(allowedValues, value) {
+    return allowedValues.some(allowedValue =>
+        _matchesValue(allowedValue, value));
+}
+
+/** _headersMatchRule - check if headers match AllowedHeaders of rule
+* @param {string[]} headers - expected headers of CORS request
```
electrachong: yes, but these functions will also apply to actual cors requests where those are the real headers so I modified the jsdoc comments accordingly in this commit: https://github.com/scality/S3/pull/518/commits/e16776d8df7bfde04ea25471a09cd9929b69bd9c

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,127 @@
...
+* @param {string[]} ruleAllowedHeaders - AllowedHeaders of a CORS rule
+* @return {boolean} - true/false
+*/
+function _headersMatchRule(headers, ruleAllowedHeaders) {
+    if (!ruleAllowedHeaders) {
+        return false;
+    }
+    // AllowedHeaders may have been stored with uppercase letters
+    // during putBucketCors; ignore case when searching for match
```
electrachong: This confused me too but they preserve case in putting / getting the cors resource. I can add some tests to substantiate this.

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,127 @@
...
+* @param {string[]} [headers] - expected headers of CORS request
+* @return {object} resHeaders - headers to include in response
+*/
+export function gatherCorsResHeaders(rule, origin, method, headers) {
+    const resHeaders = {
+        'access-control-max-age': rule.maxAgeSeconds,
+        'access-control-allow-methods': rule.allowedMethods.join(', '),
+        'content-length': '0',
+        'date': new Date().toUTCString(),
+        'vary':
```
electrachong: As far as I can tell, AWS always includes the same three values: `Origin, Access-Control-Request-Method, Access-Control-Request-Headers` as long as the response is not 403 Forbidden. This is even if rule Origin is `'*'` and even if the request includes no values for Access-Control-Request-Headers.

The only thing MDN mentions regarding the Vary header for CORS:

> If the server specifies an origin host rather than "*", then it must also include Origin in the Vary response header to indicate to clients that server responses will differ based on the value of the Origin request header.

But it does seem to make sense since these are the three headers AWS uses to determine whether to send a 200 OK response or 403 AccessForbidden response to the preflight requests.

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,127 @@
...
+    };
+    // send back '*' if any origin allowed; otherwise send back
+    // request Origin value
+    if (rule.allowedOrigins.indexOf('*') > -1) {
+        resHeaders['access-control-allow-origin'] = '*';
+    } else {
+        resHeaders['access-control-allow-origin'] = origin;
+        // can only set allow credentials to true if not returning wildcard
+        // for 'access-control-allow-origin'
+        resHeaders['access-control-allow-credentials'] = true;
```
electrachong: Yes, I noticed AWS seemed to include this header as long as the allowed origin was not wildcarded (MDN mentions you cannot use wildcarding when responding to a credentialed request).

The only reason I can think for why they do this if they also do not handle credentialed browser requests is that they would rather ignore the credentials and allow those requests than to return a response without the `access-control-allow-credentials` header, which will cause the browser to ignore the response.

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
electrachong: Nicholas developed the tests first and I wanted to add some tests with object keys. I didn't want to add `undefined` before the callback to all the rest of the tests. But since it's convention to have the callback at the end, I decided to make the parameters a single object. https://github.com/scality/S3/pull/518/commits/8eebc1f9699dadad0eac10e1a07a57c8f7842b60

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
electrachong: ✔️

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
electrachong: ✔️

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
electrachong: I think it's the relevant portion that we check for in our other tests (statusCode, code). This is an example of the "total" response:
`<?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessForbidden</Code><Message>CORSResponse: This CORS request is not allowed. This is usually because the evalution of Origin, request method / Access-Control-Request-Method or Access-Control-Request-Headers are not whitelisted by the resource's CORS spec.</Message><Resource></Resource><RequestId>73101446f9bcbc5b15ce</RequestId></Error>`

We can also check the `<Message>` element which happens to be the same at the moment, but that could change. The RequestId will also differ.

lib/api/apiUtils/object/corsResponse.js
electrachong: https://github.com/scality/S3/pull/521

lib/api/apiUtils/object/corsResponse.js
LaurenSpiegel: how about -- the value of the 'Access-Control-Allow-Headers' in an options request and the actual request header keys in any other request'

lib/api/apiUtils/object/corsResponse.js
LaurenSpiegel: Makes sense:  see the rfc -- https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

lib/api/apiUtils/object/corsResponse.js
LaurenSpiegel: ok.  I assume as far as we can tell the behavior is the same whether or not the browser sends the credential header (other than responding back with the access-control-allow-credentials header).

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
LaurenSpiegel: ok

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+                assert(total.indexOf(message) > -1, `Expected ${message}`);
+                assert.deepEqual(res.statusCode, statusCode[code],
+                `status code expected: ${statusCode[code]}`);
+            }
+            if (headersResponse) {
+                Object.keys(headersResponse).forEach(key => {
+                    assert.deepEqual(res.headers[key], headersResponse[key],
+                      `error header: ${key}`);
+                });
+            } else {
```
LaurenSpiegel: add comment: if no headersResponse provided, we should not have these headers in the request

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+            _waitForAWS(done, err);
+        });
+    });
+    afterEach(done => {
+        s3.deleteBucket({ Bucket: bucket }, err => {
+            _waitForAWS(done, err);
+        });
+    });
+
+    it('should allow GET without CORS', done => {
```
LaurenSpiegel: should allow GET on bucket without cors configuration even if origin header sent

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+        const headers = {
+            Origin: 'http://www.allowedwebsite.com',
+        };
+        methodRequest({ method: 'GET', headers, code: 200,
+        headersResponse: null }, done);
+    });
+    it('should allow HEAD without CORS', done => {
```
LaurenSpiegel: should allow HEAD request on bucket without cors configuration even if origin header sent

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+        const corsParams = {
+            Bucket: bucket,
+            CORSConfiguration: {
+                CORSRules: [
+                    {
+                        AllowedMethods: [
+                            'PUT', 'POST', 'DELETE', 'GET',
+                        ],
+                        AllowedOrigins: [
+                            'http://www.allowedwebsite.com',
```
LaurenSpiegel: might as well use a variable for http://www.allowedwebsite.com since it is used so much

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+        };
+        beforeEach(done => {
+            s3.putBucketCors(corsParams, done);
+        });
+
+        afterEach(done => {
+            s3.deleteBucketCors({ Bucket: bucket }, done);
+        });
+
+        it('send OPTIONS request with allowed origin', done => {
```
LaurenSpiegel: should respond to OPTIONS request from allowed origin with access control headers

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+                'access-control-allow-origin': 'http://www.allowedwebsite.com',
+                'access-control-allow-methods': 'PUT, POST, DELETE, GET',
+                'access-control-allow-credentials': 'true',
+                'vary': 'Origin, Access-Control-Request-Headers, ' +
+                    'Access-Control-Request-Method',
+            };
+            methodRequest({ method: 'OPTIONS', headers, code: 200,
+            headersResponse }, done);
+        });
+        it('send OPTIONS request with not allowed request headers', done => {
```
LaurenSpiegel: should respond with AccessForbidden error to OPTIONS request from allowed origin but with not allowed Access-Control-Request-Headers

LaurenSpiegel: Please update it statements below accordingly.  A person looking this should be able to understand the general purpose of the test.

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+        });
+        it('send OPTIONS request with not allowed origin', done => {
+            const headers = {
+                'Origin': 'http://www.forbiddenwebsite.com',
+                'Access-Control-Request-Method': 'GET',
+            };
+            methodRequest({ method: 'OPTIONS', headers, code: 'AccessForbidden',
+            headersResponse: null }, done);
+        });
+        it('send OPTIONS request with listed method "POST"', done => {
```
LaurenSpiegel: the post, put and delete tests look like they are the same.  perhaps this is a good place for a limited forEach test to just loop through those 3 cases.

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+                'access-control-allow-credentials': 'true',
+                'vary': 'Origin, Access-Control-Request-Headers, ' +
+                    'Access-Control-Request-Method',
+            };
+            methodRequest({ method: 'OPTIONS', headers, code: 200,
+            headersResponse }, done);
+        });
+    });
+
+    describe('CORS allows method GET and allows one origin', () => {
```
LaurenSpiegel: is this different from the tests below only because of the allowed method in the cors config?  please clarify

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+            const headers = {
+                'Origin': 'http://www.allowedwebsite.com',
+                'Access-Control-Request-Method': 'GET',
+            };
+            methodRequest({ method: 'OPTIONS', headers, code: 'AccessForbidden',
+            headersResponse: null }, done);
+        });
+    });
+
+    describe('CORS allows method PUT and allows all origins', () => {
```
LaurenSpiegel: if this is exactly the same as the describe above other than the method, it would probably be clearer to use a variable and combine the code (so a short forEach)

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
```diff
@@ -0,0 +1,1083 @@
...
+        };
+        beforeEach(done => {
+            s3.putBucketCors(corsParams, done);
+        });
+
+        afterEach(done => {
+            s3.deleteBucketCors({ Bucket: bucket }, done);
+        });
+
+        it('send OPTIONS request with fake auth credentials', done => {
```
LaurenSpiegel: should send CORS response headers even if request has fake auth credentials

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,130 @@
...
+*   applications
+* @param {string} origin - origin of CORS request
+* @param {string} method - Access-Control-Request-Method value or the method of
+    a CORS request
+* @param {string[]} [headers] - Access-Control-Request-Headers values or any
+    headers that are sent in a CORS request
+* @return {(null|object)} - matching rule if found; null if no match
+*/
+export function findCorsRule(rules, origin, method, headers) {
+    for (let i = 0; i < rules.length; i++) {
```
bennettbuchanan: `.find` method might be nice here.

lib/api/corsPreflight.js
```diff
@@ -0,0 +1,77 @@
...
+*/
+export default function corsPreflight(request, log, callback) {
+    log.debug('processing request', { method: 'corsPreflight' });
+
+    const bucketName = request.bucketName;
+    const corsOrigin = request.headers.origin;
+    const corsMethod = request.headers['access-control-request-method'];
+    const corsHeaders = request.headers['access-control-request-headers'] ?
+        request.headers['access-control-request-headers'].replace(/ /g, '')
+            .split(',').filter(value => value !== '')
```
bennettbuchanan: Could avoid constructing one array by checking for `''` in the `map` method.

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
electrachong: ✔️

tests/functional/aws-node-sdk/test/bucket/getCorsResponse.js
electrachong: ✔️

lib/api/apiUtils/object/corsResponse.js
electrachong: ✔️

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,130 @@
+/** _matchesValue - compare two values to determine if they match
+* @param {string} allowedValue - an allowed value in a CORS rule;
+    may contain wildcards
+* @param {string} value - value from CORS request
+* @return {boolean} - true/false
+*/
+function _matchesValue(allowedValue, value) {
+    const wildcardIndex = allowedValue.indexOf('*');
```
electrachong: No, there is only one asterisk permitted per value and that should have been validated when the cors rules were processed in putBucketCors. :)

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,130 @@
+/** _matchesValue - compare two values to determine if they match
+* @param {string} allowedValue - an allowed value in a CORS rule;
+    may contain wildcards
```
electrachong: ✔️

electrachong: Thanks Bennett, I fixed this

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -0,0 +1,130 @@
...
+* @param {string[]} ruleAllowedHeaders - AllowedHeaders of a CORS rule
+* @return {boolean} - true/false
+*/
+function _headersMatchRule(headers, ruleAllowedHeaders) {
+    if (!ruleAllowedHeaders) {
+        return false;
+    }
+    // AllowedHeaders may have been stored with uppercase letters
+    // during putBucketCors; ignore case when searching for match
+    const allowedHeaders = ruleAllowedHeaders.map(allowedHeader =>
```
electrachong: This is a good point, so I gave it a shot in this commit: https://github.com/scality/S3/pull/518/commits/c607e660ba4c65d353282b31b4297106f9d18a55

Note: I was also using that function to check if the origin matched any of the AllowedOrigin's as well since the code was the same. But I created the function mainly so that `_headersMatchRule` would be more readable than nesting a .some() condition in an .every(). Since we are making the `_matchesOneOf` function specific to the allowed headers, I substituted the raw code back for the allowed origins check.

lib/api/corsPreflight.js
electrachong: Yes, but we also happen to use this value for the response headers @ line 72(ish):
`        const resHeaders = gatherCorsResHeaders(corsRule, corsOrigin,
             corsMethod, corsHeaders);`

Turns out AWS removes the empty values between commas when returning the headers.

bennettbuchanan: Ah, by checking, I guess I meant more like filtering. As in, you could just not return `''` values in `.map`. It's minor though.

electrachong: ah, thanks. As discussed it doesn't seem we can do this with .map() but using reduce() to avoid the extra array construction could work. I tried this here: https://github.com/scality/S3/pull/518/commits/18c6f5fcb7d0ad3b518a8e2324ef7ae8c9b539f5

lib/api/apiUtils/object/corsResponse.js
electrachong: thanks for the suggestion: https://github.com/scality/S3/pull/518/commits/bdb7774c28ad1940b94f08554859c3c611a74d30

electrachong:
@ironman-machine

electrachong:
@ironman-machine again

electrachong:
end-to-end passed

lib/api/apiUtils/object/corsResponse.js
electrachong: Generally the access-control-allow-credentials header affects requests with a 'Cookie' header. So I added a test with a Cookie header, and the response header/body look the same as it would be for a request without the header. (Even if the access-control-allow-origin is '*' & access-control-allow-credential is omitted)

electrachong:
@ironman-machine one more time to be sure

electrachong:
@LaurenSpiegel end-to-end passed and ci checks finished

PR #522, opened 2017-01-13
Add listenOn field to config

dora-korpar:
Adds a new field to the config file that allows users to explicitly set a host:port for the s3 server to listen on.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1,6 +1,7 @@
 import assert from 'assert';
 import fs from 'fs';
 import path from 'path';
+import { ipCheck } from 'arsenal';
 
 import authDataChecker from './auth/in_memory/checker';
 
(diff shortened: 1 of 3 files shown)


LaurenSpiegel:
As discussed, it would be great if the light-local config was updated to use this option so we could have it tested.

lib/Config.js
```diff
@@ -43,6 +43,14 @@ class Config {
             this.port = config.port;
         }
 
+        this.listenOn = [];
+        if (config.listenOn !== undefined) {
+            assert(Array.isArray(config.listenOn)
+                && config.listenOn.every(e => typeof e === 'string'),
```
bennettbuchanan: What about an array element of `''`? I don't know why someone would add that, but I suppose it's possible. It looks like the server would then try to listen on `''`.

bennettbuchanan: Maybe it doesn't matter though since there are many possibilities of accidental string IP addresses.

rahulreddy: The best part this the container would be in "restarting" state with the worker crashing and restarting in a loop. So it would be apparent when they mess up.

lib/Config.js
```diff
@@ -48,7 +49,22 @@ class Config {
...
+                // if address is IPv6 format, it includes brackets
+                // that have to be removed from the final IP address
+                const ipAddress = item.indexOf(']') > 0 ?
+                    item.substr(1, lastColon - 2) :
+                    item.substr(0, lastColon);
+                // the port should not include the colon
+                const port = item.substr(lastColon + 1);
+                // parseIp returns as empty object if the address is invalid
+                assert(Object.keys(ipCheck.parseIp(ipAddress)).length !== 0,
+                    'bad config: listenOn IP address must be valid');
```
bennettbuchanan: That's slick 👍

PR #525, opened 2017-01-14
tests: fix incorrect object listing tests

rahulreddy:
NextMarker is only available when Delimiter is specified in the params


rahulreddy:
@ironman-machine try

PR #531, opened 2017-01-19
[package](BUMP): bump deps version to rel/6.4

ThibaultRiviere:


DavidPineauScality:
Merging without E2E as this change has no impact over these tests due to its nature.
Exceptionally merging with less reviews than usually required, given the simplicity and importance of this PR. Let's go fast with this :)

PR #539, opened 2017-01-20
FT: Add recent metrics listing option

bennettbuchanan:
Adds the option to list recent metrics using the Utapi route with action `ListRecentMetrics`. This will return the latest metrics of between `Date.now()` and the second most recent fifteen minute interval (e.g., if it is 6:31 when the request is made, then list metrics starting from 6:15). 

If `start` or `end` options are provided along with the `recent` option, they will be ignored and metrics of the "recent" time range will be returned.

Depends on https://github.com/scality/utapi/pull/114.

### DIFF
--- a/lib/utapi/utilities.js
+++ b/lib/utapi/utilities.js
@@ -16,12 +16,14 @@ function _listMetrics(host,
                       accessKey,
                       secretKey,
                       verbose,
+                      recent,
                       ssl) {
+    const listAction = recent ? 'ListRecentMetrics' : 'ListMetrics';
     const options = {
         host,
         port,
         method: 'POST',
-        path: `/${metric}?Action=ListMetrics`,
+        path: `/${metric}?Action=${listAction}`,
         headers: {
             'content-type': 'application/json',
             'cache-control': 'no-cache',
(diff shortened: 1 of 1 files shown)


bennettbuchanan:
Yes, the route should be tested with Integration—I will be adding those. Also, tests for this command line tool are probably in order there.

lib/utapi/utilities.js
```diff
@@ -103,9 +102,10 @@ export function listMetrics(metricType) {
         .option('-p, --port <port>', 'Port of the server')
         .option('--ssl', 'Enable ssl')
         .option('-v, --verbose')
+        .option('-r, --recent', 'List metrics of the latest 15 minute interval')
```
LaurenSpiegel: this is the last full 15 minute interval?  not from 2:15 to current if it is 2:20 for instance?  please clarify

rahulreddy: It should be the nearest 15 min interval, ie 2:15 - current if it is 2:20.
@bennettbuchanan Can you update the wording to reflect _starting from the nearest 15 minute interval to the current time_.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

lib/utapi/utilities.js
bennettbuchanan: Updated to: 'List metrics including the previous and current 15 minute interval'

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed.

PR #541, opened 2017-01-20
FT: Implement CORS headers in API responses

electrachong:
Part of S3 CORS implementation.
If Origin header designates request as a CORS request and bucket has CORS configuration, should respond with `access-control-allow-origin`, `access-control-allow-methods`, `access-control-allow-credentials` and `vary` headers if the request matches a CORS rule.

Because the cors headers require checking the cors configuration of the bucket, this involved refactoring the api methods to return an extra parameter `corsHeaders` to be added to the response meta headers, even in case of errors. 

As noted, we opted to avoid making extra calls to the metadata backend (departure from AWS compatibility) and instead gather the headers once existing metadata calls to retrieve bucketInfo are completed, generally for bucket authorization.

### DIFF
--- a/lib/api/apiUtils/object/corsResponse.js
+++ b/lib/api/apiUtils/object/corsResponse.js
@@ -30,8 +30,8 @@ function _matchesOneOf(allowedHeaders, header) {
 }
 
 /** _headersMatchRule - check if headers match AllowedHeaders of rule
-* @param {string[]} headers - the value of the 'Access-Control-Allow-Headers'
-*   in an OPTIONS request and actual request header keys in any other request
+* @param {string[]} headers - the value of the 'Access-Control-Request-Headers'
+*   in an OPTIONS request
 * @param {string[]} allowedHeaders - AllowedHeaders of a CORS rule
 * @return {boolean} - true/false
 */
@@ -58,10 +58,10 @@ function _headersMatchRule(headers, allowedHeaders) {
 * @param {string[]} [rules[].exposeHeaders] - headers to expose to external
 *   applications
 * @param {string} origin - origin of CORS request
-* @param {string} method - 'Access-Control-Request-Method' header value in
+* @param {string} method - Access-Control-Request-Method header value in
 *   an OPTIONS request and the actual method in any other request
-* @param {string[]} [headers] - the value of the 'Access-Control-Allow-Headers'
-*   in an OPTIONS request and actual request header keys in any other request
+* @param {string[]} [headers] - Access-Control-Request-Headers header value
+*   in a preflight CORS request
 * @return {(null|object)} - matching rule if found; null if no match
 */
 export function findCorsRule(rules, origin, method, headers) {
@@ -91,18 +91,19 @@ export function findCorsRule(rules, origin, method, headers) {
 * @param {string[]} [rule[].exposeHeaders] - headers to expose to external
 *   applications
 * @param {string} origin - origin of CORS request
-* @param {string} method - 'Access-Control-Request-Method' header value in
+* @param {string} method - Access-Control-Request-Method header value in
 *   an OPTIONS request and the actual method in any other request
-* @param {string[]} [headers] - the value of the 'Access-Control-Allow-Headers'
-*   in an OPTIONS request and actual request header keys in any other request
+* @param {string[]} [headers] - Access-Control-Request-Headers header value
+*   in a preflight CORS request
+* @param {boolean} [isPreflight] - indicates if cors headers are being gathered
+*   for a CORS preflight request
 * @return {object} resHeaders - headers to include in response
 */
-export function gatherCorsResHeaders(rule, origin, method, headers) {
+export function generateCorsResHeaders(rule, origin, method, headers,
+isPreflight) {
     const resHeaders = {
         'access-control-max-age': rule.maxAgeSeconds,
         'access-control-allow-methods': rule.allowedMethods.join(', '),
-        'content-length': '0',
-        'date': new Date().toUTCString(),
         'vary':
         'Origin, Access-Control-Request-Headers, Access-Control-Request-Method',
     };
@@ -122,6 +123,10 @@ export function gatherCorsResHeaders(rule, origin, method, headers) {
     if (rule.exposeHeaders) {
         resHeaders['access-control-expose-headers'] =
             rule.exposeHeaders.join(', ');
+    }
+    if (isPreflight) {
+        resHeaders['content-length'] = '0';
+        resHeaders.date = new Date().toUTCString();
     }
     return resHeaders;
 }
--- a/lib/api/bucketDeleteCors.js
+++ b/lib/api/bucketDeleteCors.js
@@ -1,6 +1,7 @@
 import { errors } from 'arsenal';
 
 import bucketShield from './apiUtils/bucket/bucketShield';
+import collectCorsHeaders from '../utilities/collectCorsHeaders';
 import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
 import metadata from '../metadata/wrapper';
 // TODO: import { pushMetric } from '../utapi/utilities';
@@ -20,6 +21,8 @@ export default function bucketDeleteCors(authInfo, request, log, callback) {
     const canonicalID = authInfo.getCanonicalID();
 
     return metadata.getBucket(bucketName, log, (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
         if (err) {
             log.debug('metadata getbucket failed', { error: err });
             return callback(err);
@@ -34,7 +37,7 @@ export default function bucketDeleteCors(authInfo, request, log, callback) {
                 requestType,
                 method: 'bucketDeleteCors',
             });
-            return callback(errors.AccessDenied);
+            return callback(errors.AccessDenied, corsHeaders);
         }
 
         const cors = bucket.getCors();
@@ -44,18 +47,18 @@ export default function bucketDeleteCors(authInfo, request, log, callback) {
             });
             // TODO: add 'deleteBucketCors' to Utapi Client action map
             // pushMetric('deleteBucketCors', log, { bucket: bucketName });
-            return callback();
+            return callback(null, corsHeaders);
         }
 
         log.trace('deleting cors configuration in metadata');
         bucket.setCors(null);
         return metadata.updateBucket(bucketName, bucket, log, err => {
             if (err) {
-                return callback(err);
+                return callback(err, corsHeaders);
             }
             // TODO: pushMetric('deleteBucketCors', log,
             // { bucket: bucketName });
-            return callback();
+            return callback(err, corsHeaders);
         });
     });
 }
--- a/lib/api/bucketDeleteWebsite.js
+++ b/lib/api/bucketDeleteWebsite.js
@@ -1,6 +1,7 @@
 import { errors } from 'arsenal';
 
 import bucketShield from './apiUtils/bucket/bucketShield';
+import collectCorsHeaders from '../utilities/collectCorsHeaders';
 import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
 import metadata from '../metadata/wrapper';
 import { pushMetric } from '../utapi/utilities';
@@ -12,6 +13,8 @@ export default function bucketDeleteWebsite(authInfo, request, log, callback) {
     const canonicalID = authInfo.getCanonicalID();
 
     return metadata.getBucket(bucketName, log, (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
         if (err) {
             log.debug('metadata getbucket failed', { error: err });
             return callback(err);
@@ -26,7 +29,7 @@ export default function bucketDeleteWebsite(authInfo, request, log, callback) {
                 requestType,
                 method: 'bucketDeleteWebsite',
             });
-            return callback(errors.AccessDenied);
+            return callback(errors.AccessDenied, corsHeaders);
         }
 
         const websiteConfig = bucket.getWebsiteConfiguration();
@@ -38,20 +41,20 @@ export default function bucketDeleteWebsite(authInfo, request, log, callback) {
                 authInfo,
                 bucket: bucketName,
             });
-            return callback();
+            return callback(null, corsHeaders);
         }
 
         log.trace('deleting website configuration in metadata');
         bucket.setWebsiteConfiguration(null);
         return metadata.updateBucket(bucketName, bucket, log, err => {
             if (err) {
-                return callback(err);
+                return callback(err, corsHeaders);
             }
             pushMetric('deleteBucketWebsite', log, {
                 authInfo,
                 bucket: bucketName,
             });
-            return callback();
+            return callback(null, corsHeaders);
         });
     });
 }
--- a/lib/api/bucketPutACL.js
+++ b/lib/api/bucketPutACL.js
@@ -4,6 +4,7 @@ import async from 'async';
 import acl from '../metadata/acl';
 import aclUtils from '../utilities/aclUtils';
 import { cleanUpBucket } from './apiUtils/bucket/bucketCreation';
+import collectCorsHeaders from '../utilities/collectCorsHeaders';
 import constants from '../../constants';
 import services from '../services';
 import vault from '../auth/vault';
@@ -103,16 +104,16 @@ export default function bucketPutACL(authInfo, request, log, callback) {
     return async.waterfall([
         function waterfall1(next) {
             services.metadataValidateAuthorization(metadataValParams,
-                (err, bucket) => {
-                    if (err) {
-                        log.trace('request authorization failed', {
-                            error: err,
-                            method: 'services.metadataValidateAuthorization',
-                        });
-                        return next(err);
-                    }
-                    return next(null, bucket);
-                });
+            (err, bucket) => {
+                if (err) {
+                    log.trace('request authorization failed', {
+                        error: err,
+                        method: 'services.metadataValidateAuthorization',
+                    });
+                    return next(err, bucket);
+                }
+                return next(null, bucket);
+            });
         },
         function waterfall2(bucket, next) {
             // If not setting acl through headers, parse body
@@ -128,7 +129,7 @@ export default function bucketPutACL(authInfo, request, log, callback) {
                         (err, jsonGrants) => next(err, bucket, jsonGrants));
                 }
                 // If no ACLs sent with request at all
-                return next(errors.MalformedXML);
+                return next(errors.MalformedXML, bucket);
             }
             /**
             * If acl set in headers (including canned acl) pass bucket and
@@ -180,7 +181,7 @@ export default function bucketPutACL(authInfo, request, log, callback) {
                         if (possibleGroups.indexOf(grantee.URI[0]) < 0) {
                             log.trace('invalid user group',
                                  { userGroup: grantee.URI[0] });
-                            return next(errors.InvalidArgument);
+                            return next(errors.InvalidArgument, bucket);
                         }
                         return usersIdentifiedByGroup.push({
                             identifier: grantee.URI[0],
@@ -209,7 +210,7 @@ export default function bucketPutACL(authInfo, request, log, callback) {
                     if (possibleGroups.indexOf(userGroup) < 0) {
                         log.trace('invalid user group', { userGroup,
                             method: 'bucketPutACL' });
-                        return next(errors.InvalidArgument);
+                        return next(errors.InvalidArgument, bucket);
                     }
                 }
                 /** TODO: Consider whether want to verify with Vault
@@ -230,7 +231,7 @@ export default function bucketPutACL(authInfo, request, log, callback) {
                         id,
                         method: 'bucketPutACL',
                     });
-                    return callback(errors.InvalidArgument);
+                    return callback(errors.InvalidArgument, bucket);
                 }
             }
 
@@ -243,7 +244,7 @@ export default function bucketPutACL(authInfo, request, log, callback) {
                         if (err) {
                             log.trace('error looking up canonical ids', {
                                 error: err, method: 'vault.getCanonicalIDs' });
-                            return next(err);
+                            return next(err, bucket);
                         }
                         const reconstructedUsersIdentifiedByEmail = aclUtils
                             .reconstructUsersIdentifiedByEmail(results,
@@ -272,21 +273,25 @@ export default function bucketPutACL(authInfo, request, log, callback) {
             if (bucket.hasTransientFlag() || bucket.hasDeletedFlag()) {
                 log.trace('transient or deleted flag so cleaning up bucket');
                 bucket.setFullAcl(addACLParams);
-                return cleanUpBucket(bucket, canonicalID, log, next);
+                return cleanUpBucket(bucket, canonicalID, log, err =>
+                    next(err, bucket));
             }
             // If no bucket flags, just add acl's to bucket metadata
-            return acl.addACL(bucket, addACLParams, log, next);
+            return acl.addACL(bucket, addACLParams, log, err =>
+                next(err, bucket));
         },
-    ], err => {
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
         if (err) {
             log.trace('error processing request', { error: err,
                 method: 'bucketPutACL' });
-            return callback(err);
+        } else {
+            pushMetric('putBucketAcl', log, {
+                authInfo,
+                bucket: bucketName,
+            });
         }
-        pushMetric('putBucketAcl', log, {
-            authInfo,
-            bucket: bucketName,
-        });
-        return callback(err, 'ACL Set');
+        return callback(err, corsHeaders);
     });
 }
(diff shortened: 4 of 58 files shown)


electrachong:
@ironman-machine

electrachong:
end to end passed here: http://ci.ironmann.io/gh/scality/Integration/8874

nicolas2bert:
@LaurenSpiegel - the 2 ceph cors tests that were failing pass now against this branch.
Good job @electrachong !

electrachong:
@ironman-machine please

electrachong:
end-to-end passed

lib/api/objectCopy.js
```diff
@@ -140,7 +142,26 @@ function objectCopy(authInfo, request, sourceBucket,
...
     return async.waterfall([
-        function checkSourceAuthorization(next) {
+        function checkDestAuth(next) {
+            return services.metadataValidateAuthorization(valPutParams,
+                (err, destBucketMD, destObjMD) => {
+                    corsHeaders = collectCorsHeaders(request, destBucketMD);
+                    if (err) {
+                        log.debug('error validating put part of request',
+                        { error: err });
+                        return next(err);
```
electrachong: think I want to return cors headers here too

lib/api/bucketDeleteWebsite.js
```diff
@@ -12,9 +13,10 @@ export default function bucketDeleteWebsite(authInfo, request, log, callback) {
...
 
     return metadata.getBucket(bucketName, log, (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request, bucket);
         if (err) {
             log.debug('metadata getbucket failed', { error: err });
-            return callback(err);
+            return callback(err, corsHeaders);
         }
         if (bucketShield(bucket, requestType)) {
             return callback(errors.NoSuchBucket);
```
dora-korpar: How come corsHeaders isn't returned here?

lib/api/initiateMultipartUpload.js
```diff
@@ -144,31 +146,32 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
                         });
-                        return callback(err, xml);
+                        return callback(err, xml, corsHeaders);
                     });
             });
     }
 
     services.metadataValidateAuthorization(metadataValParams,
         (err, destinationBucket) => {
+            corsHeaders = collectCorsHeaders(request, destinationBucket);
```
dora-korpar: couldn't corsHeaders be declared as a const here?

lib/api/objectPut.js
```diff
@@ -119,7 +120,7 @@ function _storeIt(bucketName, objectKey, objMD, authInfo, canonicalID,
                                 newByteLength: size,
                                 oldByteLength: prevContentLen,
                             });
-                            return callback(null, contentMD5, prevContentLen);
```
dora-korpar: Since prevContentLen has been removed from both the callbacks, is it even necessary to calculate anymore (line 79)? Looks like it's just being logged in utapi

tests/functional/aws-node-sdk/test/object/corsHeaders.js
```diff
@@ -0,0 +1,756 @@
...
+                    ],
+                },
+            };
+            beforeEach(done => {
+                s3.putBucketCors(corsParams, done);
+            });
+
+            afterEach(done => {
+                s3.deleteBucketCors({ Bucket: bucket }, done);
+            });
```
dora-korpar: Would it be worth it to try to group the 3 or 4 tests that use this same beforeEach/afterEach setup?

lib/api/initiateMultipartUpload.js
electrachong: main reason I didn't is because I didn't want to modify the helper function `_storetheMPObject` to add the `corsHeaders` as another param, although looking at it now I definitely could and might as well. I'll update when I've pushed the change.

lib/api/objectPut.js
electrachong: I think Utapi uses this for some metrics, right @bennettbuchanan? So we still want to calculate it. Originally it was passed to the callbacks for Utapi because utapi logging used to happen in the routes. Then when the refactoring happened to put the pushMetric calls in the api methods, we forgot to remove it from the callbacks.

bennettbuchanan: UtapiClient uses the value of `prevContentLen` to calculate the storage utilized metric if the operation is overwriting an object. Should be `null` if not overwriting.

lib/api/bucketDeleteWebsite.js
electrachong: BucketShield is used to shield buckets from requests when they've been flagged for deletion or 'transient' (in process of creation). In those scenarios we return a `NoSuchBucket` error to the user so I think it's a little odd to return cors headers for a bucket that isn't supposed to exist from the user's perspective. But we technically could, so I'm open to other arguments.

tests/functional/aws-node-sdk/test/object/corsHeaders.js
electrachong: I would place the `beforeEach` in the parent `describe` but the cors configuration (`corsParams`) being put in the `putBucketCors` changes with each test group. I can at least put the `afterEach` in the parent `describe`.

Another possibility is doing a `forEach` statement. I'm not sure how readable it would be to make one that would work for the whole group but the last two tests are pretty similar.

electrachong: I also think I'll make a helper function to generate the `corsParams` because they do take up a lot of lines.

lib/api/bucketDeleteWebsite.js
nicolas2bert: What headers does real AWS return @electrachong ?

lib/api/objectCopy.js
electrachong: nvm (handled in async callback for all errors)

lib/api/initiateMultipartUpload.js
electrachong: https://github.com/scality/S3/pull/541/commits/34ba4c99567cb65dfea9ef56dbc3fac08ec29420

lib/api/bucketDeleteWebsite.js
electrachong: In case of `NoSuchBucket` they don't return the CORS headers.

lib/api/initiateMultipartUpload.js
```diff
@@ -144,31 +145,32 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
                 });
-                return callback(err);
+                return callback(err, null, corsHeaders);
             }
             if (destinationBucket.hasDeletedFlag() &&
                 accountCanonicalID !== destinationBucket.getOwner()) {
                 log.trace('deleted flag on bucket and request ' +
                     'from non-owner account');
-                return callback(errors.NoSuchBucket);
+                return callback(errors.NoSuchBucket, null, corsHeaders);
```
nicolas2bert: No corsHeaders needed here, right?

tests/functional/aws-node-sdk/test/object/corsHeaders.js
electrachong: helper function for corsParams: https://github.com/scality/S3/pull/541/commits/3aa0e7e09ab3c3fcbc5c58855b104e3deb9ad695?diff=split
forEach for last two tests:
https://github.com/scality/S3/pull/541/commits/81b3d0e328bd6276d862d831de2e38e21431670e?diff=split

tests/functional/aws-node-sdk/test/object/corsHeaders.js
```diff
@@ -0,0 +1,756 @@
...
+            checkNoCorsHeaders: false,
+        },
+    },
+];
+
+// AWS seems to take a bit long so sometimes by the time we send the request
+// the bucket has not yet been created or the bucket has been deleted.
+function _waitForAWS(callback, err) {
+    if (process.env.AWS_ON_AIR) {
+        setTimeout(() => callback(err), 1000);
```
electrachong: problem is AWS responds to request but bucket may still not be finished deleting in background

actually with this much timeout they (some of them) still fail on `AWS_ON_AIR` when I test locally :C I could increase it but I think the person testing can opt to do that if they want to

lib/api/initiateMultipartUpload.js
electrachong: thank you, you are right

electrachong:
@ironman-machine

electrachong:
@ironman-machine

electrachong:
end-to-end passing again

lib/api/bucketPutACL.js
```diff
@@ -281,12 +285,15 @@ export default function bucketPutACL(authInfo, request, log, callback) {
         if (err) {
             log.trace('error processing request', { error: err,
                 method: 'bucketPutACL' });
-            return callback(err);
+            if (err === errors.NoSuchBucket) {
+                return callback(err);
+            }
```
bennettbuchanan: I assume there is a reason you don't want `corsHeaders` when the error is NoSuchBucket. 

You might then just return the callback here with `corsHeaders` as the second argument rather than using an else-statement for the `pushMetric` call.

lib/api/bucketPutACL.js
```diff
@@ -100,19 +101,22 @@ export default function bucketPutACL(authInfo, request, log, callback) {
         aclUtils.parseGrant(request.headers
             ['x-amz-grant-full-control'], 'FULL_CONTROL');
 
+    let corsHeaders;
```
bennettbuchanan: Can you get this value at the end of the waterfall, rather than using the outer scope?

electrachong: Updated here: https://github.com/scality/S3/pull/541/commits/eed031365b423d65ce9657f9b14797cd28beaf0d

lib/api/bucketPutACL.js
electrachong: Since I pass the bucket through the waterfall with https://github.com/scality/S3/pull/541/commits/eed031365b423d65ce9657f9b14797cd28beaf0d, I don't need the check for NoSuchBucket err any more (just need to not pass the bucket when the error is NoSuchBucket and there won't be any corsHeaders).

electrachong:
@ironman-machine

lib/api/bucketPutACL.js
bennettbuchanan: 😄

electrachong:
end-to-end is running here: http://ci.ironmann.io/gh/scality/Integration/9041

lib/api/objectCopy.js
```diff
@@ -140,28 +141,46 @@ function objectCopy(authInfo, request, sourceBucket,
     }
 
     return async.waterfall([
-        function checkSourceAuthorization(next) {
+        function checkDestAuth(next) {
+            return services.metadataValidateAuthorization(valPutParams,
```
rahulreddy: Same here, why is the check moved to be the first step?

Edit: I see that you need destBucketMD for the error callbacks. I would suggest on calling the metadataGetBucket directly and process the checks in `checkDestAuth` by removing services.metadataValidateAuthorization directly.

lib/api/objectPutCopyPart.js
```diff
@@ -77,60 +77,59 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
     };
 
     return async.waterfall([
-        function checkSourceAuthorization(next) {
+        function checkDestAuth(next) {
+            return services.metadataValidateAuthorization(valPutParams,
```
rahulreddy: Same here. I would recommend calling `metadata.getBucket` and use the bucket directly in `checkSourceAuthorization` step by bringing the logic directly from `services.metadataValidateAuthorization`.

lib/routes/routeGET.js
```diff
@@ -17,55 +17,67 @@ export default function routerGET(request, response, log, statsClient) {
...
-            });
+            api.callApiMethod('listParts', request, log,
+                (err, xml, corsHeaders) => {
+                    statsReport500(err, statsClient);
+                    return routesUtils.responseXMLBody(err, xml, response, log,
+                        corsHeaders);
+                });
         } else {
             // GET object
             api.callApiMethod('objectGet', request, log, (err, dataGetInfo,
```
rahulreddy: No corsHeaders for objectGet?

lib/routes/routePUT.js
```diff
@@ -137,18 +140,23 @@ export default function routePUT(request, response, log, statsClient) {
                         log.trace('location constraint',
                             { locationConstraint });
                         return api.callApiMethod('bucketPut', request, log,
-                        err => {
+                        (err, corsHeaders) => {
                             statsReport500(err, statsClient);
-                            return routesUtils.responseNoBody(err, location,
+                            const resHeaders = corsHeaders ?
+                                Object.assign(location, corsHeaders) : location;
```
rahulreddy: This would mutate location object. I recommend
```js
 Object.assign({}, location, corsHeaders)
```

lib/routes/routePUT.js
```diff
@@ -137,18 +140,23 @@ export default function routePUT(request, response, log, statsClient) {
...
-                return api.callApiMethod('bucketPut', request, log, err => {
-                    statsReport500(err, statsClient);
-                    return routesUtils.responseNoBody(err, location, response,
-                      200, log);
-                });
+                return api.callApiMethod('bucketPut', request, log,
+                    (err, corsHeaders) => {
+                        statsReport500(err, statsClient);
+                        const resHeaders = corsHeaders ?
+                            Object.assign(location, corsHeaders) : location;
```
rahulreddy: Same here.
```js
Object.assign({}, location, corsHeaders)
```

lib/utilities/collectCorsHeaders.js
```diff
@@ -0,0 +1,34 @@
+import { findCorsRule,
+generateCorsResHeaders } from '../api/apiUtils/object/corsResponse.js';
+
+/**
+ * collectCorsHeaders - gather any relevant CORS headers
+ * @param {object} request - http request object
+ * @param {BucketInfo} bucket - instance of BucketInfo class
+ * @return {object} - object containing CORS headers
+ */
+export default function collectCorsHeaders(request, bucket) {
```
rahulreddy: I would suggest sending `httpOrigin`, `httpMethod` as params to this method instead of the entire request object. This would make it easy on us in the future when we remove the references to request object.

Edit: Sorry, I know that this would be a pain to update its references everywhere but it's a good coding pattern.

lib/utilities/collectCorsHeaders.js
```diff
@@ -0,0 +1,34 @@
...
+    // NOTE: Because collecting CORS headers requires making a call to
+    // metadata to retrieve the bucket's CORS configuration, we opt not to
+    // return the CORS headers if the request encounters an error before
+    // the api method retrieves the bucket from metadata (an example
+    // being if a request is not properly authenticated). This is a slight
+    // deviation from AWS compatibility, but rebuffs additional backend calls,
+    // anticipating that the preflight OPTIONS route will serve most client
+    // needs regarding CORS.
+    if (!origin || !bucket) {
+        return null;
```
rahulreddy: To keep the return consistent, you should return and empty object.

lib/api/completeMultipartUpload.js
```diff
@@ -114,29 +115,56 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+            const metadataValParams = {
+                objectKey,
+                authInfo,
+                bucketName,
+                // Required permissions for this action
+                // at the destinationBucket level are same as objectPut
+                requestType: 'objectPut',
+                log,
+            };
+            services.metadataValidateAuthorization(metadataValParams,
```
electrachong: We were checking the auth, but the order was switched around.
(This way, if we encounter error with auth for source dest, we would have the bucket for dest already to gather cors headers from).
I checked with Lauren about this and she seemed to think it would be fine.

lib/api/completeMultipartUpload.js
```diff
@@ -114,29 +115,56 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                (err, destinationBucket, objMD) => {
+                    if (err) {
+                        return next(err, null, destinationBucket);
```
electrachong: 👍

lib/api/completeMultipartUpload.js
```diff
@@ -114,29 +115,56 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                    }
+                    return next(null, destinationBucket, objMD);
```
electrachong: 👍 updated

lib/api/listParts.js
```diff
@@ -141,11 +148,13 @@ export default function listParts(authInfo, request, log, callback) {
             };
             services.getSomeMPUparts(getPartsParams, (err, storedParts) => {
                 if (err) {
-                    return next(err);
+                    return next(err, null, destBucket);
                 }
-                return next(null, mpuBucket, storedParts, mpuOverviewArray);
+                return next(null, mpuBucket, storedParts, mpuOverviewArray,
```
electrachong: 👍 updated

lib/api/objectCopy.js
electrachong: A comment Lauren made when I discussed this with her is that the other checks are all synchronous, and it's a bit odd to leave them hanging there in the waterfall when no further asynch call needs to be made.

The concern seems to be if there was a particular order to the checks for a reason, but there isn't much evidence for that reason. We don't have any functional tests that show the order matters at least. The error is similar for both source and destination auth failure (both result in AccessDenied).

lib/routes/routeGET.js
electrachong: No worries, there are corsHeaders -- they are added to the existing resMetaHeaders param in the api method: https://github.com/scality/S3/pull/541/files/2a17b1f986738169f32dd80e2f737fdfe921dbc3#diff-dfcd28083d70734f8867f501e8deed28

lib/routes/routePUT.js
electrachong: OK, thanks for pointing that out (updated)

lib/routes/routePUT.js
electrachong: 👍

lib/utilities/collectCorsHeaders.js
electrachong: OK, it wasn't too bad (updated)

lib/utilities/collectCorsHeaders.js
electrachong: I have updated it but it seems to have turned up an undetected issue:
in some cases, I was returning `responseMetaHeaders` including `corsHeaders` in case of error when we only want to return `corsHeaders`. This caused one of the legacy tests for objectGet to fail. I'm working on fixing it.

electrachong:
@rahulreddy if you approve of the changes as is, let me know so I can squash my changes & let the ci run :)

lib/utilities/collectCorsHeaders.js
electrachong: Fixed it: https://github.com/scality/S3/pull/541/commits/520adf18e24f57adee09138434d57b9b2a9910b5

electrachong:
@ironman-machine do you spontaneously work yet?

electrachong:
end to end running here: http://ci.ironmann.io/gh/scality/Integration/9087

lib/api/bucketDeleteCors.js
```diff
@@ -44,18 +47,18 @@ export default function bucketDeleteCors(authInfo, request, log, callback) {
...
-            return callback();
+            return callback(null, corsHeaders);
         }
 
         log.trace('deleting cors configuration in metadata');
         bucket.setCors(null);
         return metadata.updateBucket(bucketName, bucket, log, err => {
             if (err) {
-                return callback(err);
+                return callback(err, corsHeaders);
```
LaurenSpiegel: is this what aws does?  sort of odd that you return the cors headers based on the configuration you are deleting

lib/api/bucketPutCors.js
```diff
@@ -58,30 +59,34 @@ export default function bucketPutCors(authInfo, request, log, callback) {
...
-        function updateBucketMetadata(bucket, rules, next) {
+        function updateBucketMetadata(bucket, rules, corsHeaders, next) {
             log.trace('updating bucket cors rules in metadata');
             bucket.setCors(rules);
-            metadata.updateBucket(bucketName, bucket, log, next);
+            metadata.updateBucket(bucketName, bucket, log, err =>
+                next(err, corsHeaders));
         },
-    ], err => {
+    ], (err, corsHeaders) => {
```
LaurenSpiegel: should this be based on the old cors config or the new?

lib/utilities/collectCorsHeaders.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {BucketInfo} bucket - instance of BucketInfo class
+ * @return {object} - object containing CORS headers
+ */
+export default function collectCorsHeaders(origin, httpMethod, bucket) {
+    // NOTE: Because collecting CORS headers requires making a call to
+    // metadata to retrieve the bucket's CORS configuration, we opt not to
+    // return the CORS headers if the request encounters an error before
+    // the api method retrieves the bucket from metadata (an example
+    // being if a request is not properly authenticated). This is a slight
+    // deviation from AWS compatibility, but rebuffs additional backend calls,
```
LaurenSpiegel: but has the benefit of avoiding additional backend calls for an invalid request. Also, we anticipate that the preflight OPTIONS route will serve most client needs regarding CORS.

tests/unit/api/bucketHead.js
```diff
@@ -37,15 +37,4 @@ describe('bucketHead API', () => {
             });
         });
     });
-
-    it('should return a success message if ' +
```
LaurenSpiegel: why deleted?

lib/api/bucketDeleteCors.js
electrachong: Yep. It seems strange at first but if you think about it, the cors headers are the green light for CORS requests. If a cors config currently exists, it makes sense to permit a CORS request that matches to delete the config (it hasn't been deleted yet). Once the config is deleted, the same request would no longer be permitted.

lib/api/bucketPutCors.js
electrachong: old cors config
Functional test creates new bucket but request still returns cors headers of previous bucket cors config (against aws)

electrachong: .. Let me test for a minute because my memory seems different from what I coded / commented.

electrachong: Returns headers based on old cors config

lib/api/bucketDeleteCors.js
LaurenSpiegel: ok.  just want to make sure we tested.

lib/api/bucketPutCors.js
LaurenSpiegel: ok.  do we have a test?

lib/utilities/collectCorsHeaders.js
electrachong: ✔️ updated

lib/api/apiUtils/object/corsResponse.js
```diff
@@ -91,18 +91,19 @@ export function findCorsRule(rules, origin, method, headers) {
...
 *   applications
 * @param {string} origin - origin of CORS request
-* @param {string} method - 'Access-Control-Request-Method' header value in
+* @param {string} method - Access-Control-Request-Method header value in
 *   an OPTIONS request and the actual method in any other request
-* @param {string[]} [headers] - the value of the 'Access-Control-Allow-Headers'
-*   in an OPTIONS request and actual request header keys in any other request
+* @param {string[]} [headers] - Access-Control-Request-Headers header value
+*   in a preflight CORS request
+* @param {boolean} [isPreflight] - indicate if cors headers are being gathered
```
electrachong: ✔️ updated

lib/api/bucketPutCors.js
electrachong: yes --> params for putbucketcors test have a different cors config (only 'put' in allowedmethods), we assert value of headers reflects the original cors config (which allows all the methods)
https://github.com/scality/S3/pull/541/files#diff-02149df1857c884500ca831ab83b1734 line 173ish

tests/unit/api/bucketHead.js
electrachong: it asserts result, a string I removed as a param for the callback in routes, and replaced with corsHeaders. I think the string was included as a param specifically for this test?
Anyway I thought for some reason there was another test that covered success but I don't see it; so I can replace this test with one that asserts no error if that's sufficient.

LaurenSpiegel: yes, please

electrachong:
another end-to-end: http://ci.ironmann.io/gh/scality/Integration/9154

electrachong:
another end-to-end: http://ci.ironmann.io/gh/scality/Integration/9156

tests/unit/api/bucketHead.js
electrachong: updated

electrachong:
end-to-end passed

lib/api/completeMultipartUpload.js
```diff
@@ -145,7 +139,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             if (request.post) {
                 return parseXml(request.post, (err, jsonList) => {
                     if (err) {
-                        return next(err, null, destBucket);
+                        return next(err, destBucket, null);
```
rahulreddy: Minor, ignoring null should work.
```js
next(err, destBucket)
```

electrachong: OK, I updated that

Issue #544, opened 2017-01-22
pid of a dying worker process needs to be logged


msegura:
when an s3 worker dies and is restarted, the following log lines are generated:
```

s3-0.log.1:{"name":"S3","time":1485071547861,"workerId":18,"level":"error","message":"worker disconnected. making sure exits","hostname":"vnappdxobjs302.gecis.io","pid":117}
s3-0.log.1:{"name":"S3","time":1485071547862,"workerId":18,"level":"error","message":"worker exited.","hostname":"vnappdxobjs302.gecis.io","pid":117}
s3-0.log.1:{"name":"S3","time":1485071548284,"workerId":27,"level":"error","message":"new worker forked","hostname":"vnappdxobjs302.gecis.io","pid":117}
s3-0.log.1:{"name":"S3","time":1485071551836,"bootstrap":["10.64.50.106"],"https":false,"level":"info","message":"bucketclient configuration","hostname":"vnappdxobjs302.gecis.io","pid":1162}
s3-0.log.1:{"name":"S3","time":1485071553151,"host":"10.64.50.106","port":8500,"https":false,"level":"info","message":"vaultclient configuration","hostname":"vnappdxobjs302.gecis.io","pid":1162}
s3-0.log.1:{"name":"S3","time":1485071555396,"https":false,"level":"info","message":"Https server configuration","hostname":"vnappdxobjs302.gecis.io","pid":1162}
```

This is good. It would be tremendously useful if the pid of the worker that died was logged as well (we log the pid in the other log lines, not the workerId, hence it is hard to track what the worker was doing prior its premature death)

LaurenSpiegel:
Fixed by #545

PR #545, opened 2017-01-22
FIX: Worker pid

alexandre-merle:
Log worker pid to help worker investigations

Fixes https://github.com/scality/S3/issues/544


alexandre-merle:
@ironman-machine try

ThibaultRiviere:
See: #544

PR #554, opened 2017-01-24
Forward rel/6.2.5 to rel/6.3

bennettbuchanan:


tests/functional/aws-node-sdk/test/bucket/putACL.js
```diff
@@ -1,9 +1,64 @@
 import assert from 'assert';
+import { S3 } from 'aws-sdk';
 
 import withV4 from '../support/withV4';
 import BucketUtility from '../../lib/utility/bucket-util';
+import getConfig from '../support/config';
 
+const bucket = `bucketacl-bucket-${Date.now()}`;
 const bucketName = 'putbucketaclfttest';
```
electrachong: is it intentional to use different bucket names for the different tests or could you use the same var for both?

bennettbuchanan: The intention was to preserve as much of the two commits as possible. It's likely this file should be refactored.

rahulreddy:
Closing this in favor of #567

PR #560, opened 2017-01-25
fix: increase max post body size

### LINKED ISSUE #559: Increase MAX_POST_LENGTH value
MAX_POST_LENGTH = 512KB was introduced in rel/6.2.5 and we have a test in rel/6.3 which uploads 10000 parts, the complete multipart upload request fails as the XML clocks around 900KB. The limit needs to be increased to at least 1MB.

rahulreddy:
fixes #559

### DIFF
--- a/lib/routes/routePOST.js
+++ b/lib/routes/routePOST.js
@@ -4,7 +4,7 @@ import api from '../api/api';
 import routesUtils from './routesUtils';
 import pushMetrics from '../utilities/pushMetrics';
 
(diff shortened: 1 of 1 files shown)


rahulreddy:
@ironman-machine try

bennettbuchanan:
Might `MAX_POST_LENGTH` in routePUT.js be updated also?
https://github.com/scality/S3/blob/rel/6.2.5/lib/routes/routePUT.js#L20

rahulreddy:
That should be ok, given that ACL request XML' are pretty small.

PR #564, opened 2017-01-26
README: Use s3api instead of s3 for awscli

Ge0:
Hey guys,

Just went through the README to bootstrap a s3 server on my machine and I found it easier to use `s3api` subcommand from the `aws` CLI too instead of `s3`. Reason is that with `s3` one has to specify a region and I haven't found any workout to deal with it, since it is not describe in the `README.md`.

What do you thinks?

Cheers,

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,10 +186,16 @@ aws_access_key_id = accessKey1
 aws_secret_access_key = verySecretKey1
 ```
 
+Create a bucket:
+
+```shell
+aws s3api create-buket --endpoint-url=http://localhost:8000 --bucket mybucket
+```
+
 See all buckets:
 
 ```shell
-aws s3 ls --endpoint-url=http://localhost:8000
+aws s3api list-buckets --endpoint-url=http://localhost:8000 --output=text
 ```
 
 #### [s3cmd](http://s3tools.org/s3cmd)
--- a/README.md
+++ b/README.md
@@ -186,16 +186,24 @@ aws_access_key_id = accessKey1
 aws_secret_access_key = verySecretKey1
 ```
 
+Also, `~/.aws/config` on Linux, OS X, or Unix or
+`C:\Users\USERNAME\.aws\config` on Windows
+
+```shell
+[default]
+region = us-east-1
+```
+
 Create a bucket:
 
 ```shell
-aws s3api create-buket --endpoint-url=http://localhost:8000 --bucket mybucket
+aws s3 --endpoint-url=http://localhost:8000 mb s3://mybucket
 ```
 
 See all buckets:
 
 ```shell
-aws s3api list-buckets --endpoint-url=http://localhost:8000 --output=text
+aws s3 --endpoint-url=http://localhost:8000 ls
 ```
 
 #### [s3cmd](http://s3tools.org/s3cmd)


electrachong:
Here is how you can configure a default location:
http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-quick-configuration

I do like s3api, I feel it is more robust when you want to execute a variety of actions. `s3` is more for when you want to execute some common actions quickly, which also suits this scenario. Rather than picking one or the other, why not list them both and let the reader choose? We can make sure to include the link for how to configure region by default for the `s3` service.

nicolas2bert:
Hey @Ge0 , since s3 and s3api share the same aws configuration, I was wondering what was your issue concerning the region?

README.md
```diff
@@ -186,10 +186,16 @@ aws_access_key_id = accessKey1
...
+Create a bucket:
+
+```shell
+aws s3api create-buket --endpoint-url=http://localhost:8000 --bucket mybucket
+```
+
 See all buckets:
 
 ```shell
-aws s3 ls --endpoint-url=http://localhost:8000
```
nicolas2bert: why not keeping the s3 one?

Ge0:
Hello folks,

@nicolas2bert when I use the `s3` subcommand from `awscli`, I actually get this issue:

```shell
$ aws --profile=scality --endpoint-url=http://localhost:8000 s3 mb s3://mybucket             
make_bucket failed: s3://mybucket Parameter validation failed:                          
Invalid type for parameter CreateBucketConfiguration.LocationConstraint, value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```

Fyi:

```shell
$ cat ~/.aws/credentials

[scality]
aws_access_key_id = accessKey1
aws_secret_access_key = verySecretKey1
$ cat ~/.aws/config     

[scality]
region = localregion
```

Whereas it works with the subcommand `s3api` from the awscli tool.

Despite the link provided by @electrachong I didn't come up with a solution.

Any ideas?

README.md
Ge0: We could actually, but since I only managed to make thinks working out with `s3api` I wanted to remain homogeneous.

nicolas2bert:
Hey @Ge0 

If you want to configure an additional profile - you have to add entries to the config and credentials files:

**~/.aws/credentials**
```
[scality]
aws_access_key_id=accessKey1
aws_secret_access_key=verySecretKey1
```

**~/.aws/config**
```
[profile scality]
region=us-east-1
output=json
```

Ge0:
Works like a charm now! ;) But how was I supposed to know that I should have set `us-east-1` as a region and `profile scality` as a key?

nicolas2bert:
`us-east-1` is the default region - but with our s3 server you can specify any regions us-west-2... (since there is no real location).
About the profile section there is a **Named Profiles** section here: http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html
I am glad it worked out :)

Ge0:
So is my patch likely to be merged or do you need me to add additional content in the README.me file?

nicolas2bert:
Hey @Ge0 - we really appreciate your comments and we are going to improve the AWS-CLI part of our `README.md`. 

Thanks again :)

PR #570, opened 2017-01-27
FIX: update completempu test for 1 MB limit

electrachong:
Needed to update this test now that MAX_POST_SIZE for POST route was updated in https://github.com/scality/S3/pull/560 to 1 MB.

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/compluteMpu.js
+++ b/tests/functional/aws-node-sdk/test/object/compluteMpu.js
@@ -7,8 +7,8 @@ const bucket = `bucketcompletempu-bucket-${Date.now()}`;
 
 const parts = [];
 
-// results in body of 589824 bytes
-for (let i = 0; i < 100000; i ++) {
+// results in body of 1049793 bytes (1 MB = 1048576)
+for (let i = 0; i < 15600; i ++) {
     // an mpu of this many parts would not be allowed.
     // testing here to make sure we are not sent excess xml
     parts.push({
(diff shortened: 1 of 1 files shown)


tests/functional/aws-node-sdk/test/object/compluteMpu.js
```diff
@@ -7,8 +7,8 @@ const bucket = `bucketcompletempu-bucket-${Date.now()}`;
 
 const parts = [];
 
-// results in body of 589824 bytes
-for (let i = 0; i < 100000; i ++) {
+// results in body of 1049793 bytes (1 MB = 1048576)
+for (let i = 0; i < 15600; i ++) {
```
dora-korpar: how come this number changed?

electrachong: I think it makes sense to test a post size bigger than the limit but close to it.
Originally the number of parts being uploaded was 100,000 but this resulted in a post body size that was a lot larger than the limit, so even when we increased the limit the test was still passing.

electrachong:
@ironman-machine

rahulreddy:
@ironman-machine try

rahulreddy:
Thanks @electrachong !

rahulreddy:
Manually triggered end to end here http://ci.ironmann.io/gh/scality/Integration/9000#build-parameters

rahulreddy:
End to end passed.

PR #574, opened 2017-01-28
Fwdport/6.4 master

rahulreddy:


rahulreddy:
end to end http://ci.ironmann.io/gh/scality/Integration/9046#build-parameters

PR #575, opened 2017-01-30
Ft/multiple backend config

dora-korpar:
Restructures S3 config to allow possibility of using multiple backends

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,24 +1,68 @@
 {
     "port": 8000,
     "listenOn": [],
-    "regions": {
-        "ap-northeast-1": ["s3.ap-northeast-1.amazonaws.com"],
-        "ap-southeast-1": ["s3.ap-southeast-1.amazonaws.com"],
-        "ap-southeast-2": ["s3.ap-southeast-2.amazonaws.com"],
-        "eu-central-1": ["s3.eu-central-1.amazonaws.com",
-                         "s3.eu.central-1.amazonaws.com"],
-        "eu-west-1": ["s3.eu-west-1.amazonaws.com"],
-        "sa-east-1": ["s3.sa-east-1.amazonaws.com"],
-        "us-east-1": ["s3.amazonaws.com",
-                      "s3-external-1.amazonaws.com",
-                      "s3.us-east-1.amazonaws.com"],
-        "us-west-1": ["s3.us-west-1.amazonaws.com"],
-        "us-west-2": ["s3-us-west-2.amazonaws.com"],
-        "us-gov-west-1": ["s3-us-gov-west-1.amazonaws.com",
-                          "s3-fips-us-gov-west-1.amazonaws.com"],
-        "localregion": ["localhost"],
-        "test-region": ["s3.scality.test"],
-        "docker-region": ["s3.docker.test"]
+    "locationConstraints": {
+        "aws-us-east-1": {
+            "type": "aws_s3",
+            "legacyAwsBehavior": true,
+            "details": {
+                "awsEndpoint": "s3.amazonaws.com",
+                "bucketName": "premadebucket",
+                "credentialsProfile": "default"
+            }
+        },
+        "scality-us-east-1": {
+            "type": "scality",
+            "legacyAwsBehavior": true,
+            "details": {
+                "connector": {
+                    "sproxyd": {
+                        "bootstrap": ["localhost:8181"]
+                    }
+                }
+            }
+        },
+        "scality-us-west-1": {
+            "type": "scality",
+            "legacyAwsBehavior": false,
+            "details": {
+                "connector": {
+                    "sproxyd": {
+                        "bootstrap": ["localhost:8182"]
+                    }
+                }
+            }
+        },
+        "us-east-1": {
+            "type": "file",
+            "legacyAwsBehavior": true,
+            "details": {
+            }
+        },
+        "virtual-user-metadata": {
+            "type": "scality_user_metadata",
+            "legacyAwsBehavior": false,
+            "details": {
+            }
+        },
+        "file": {
+            "type": "file",
+            "legacyAwsBehavior": false,
+            "details": {
+            }
+        },
+        "mem": {
+            "type": "mem",
+            "legacyAwsBehavior": false,
+            "details": {
+            }
+        }
+    },
+    "restEndpoints": {
+        "127.0.0.1": "file",
+        "s3.docker.test": "scality-us-west-1",
+        "127.0.0.2": "aws-us-east-1",
+        "s3.amazonaws.com": "aws-us-east-1"
     },
     "websiteEndpoints": ["s3-website-us-east-1.amazonaws.com",
                         "s3-website.us-east-2.amazonaws.com",
@@ -34,9 +78,6 @@
                         "s3-website-sa-east-1.amazonaws.com",
                         "s3-website.localhost",
                         "s3-website.scality.test"],
-    "sproxyd": {
-        "bootstrap": ["localhost:8181"]
-    },
     "bucketd": {
         "bootstrap": ["localhost"]
     },
@@ -51,6 +92,5 @@
     },
     "healthChecks": {
         "allowFrom": ["127.0.0.1/8", "::1"]
-    },
-    "usEastBehavior": false
+    }
 }
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -9,6 +9,59 @@ import authDataChecker from './auth/in_memory/checker';
 const defaultHealthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
 
 const defaultLocalCache = { host: '127.0.0.1', port: 6379 };
+
+export function sproxydAssert(configSproxyd) {
+    let sproxydField;
+    if (configSproxyd.bootstrap !== undefined) {
+        assert(Array.isArray(configSproxyd.bootstrap)
+            && configSproxyd.bootstrap
+                .every(e => typeof e === 'string'),
+            'bad config: sproxyd.bootstrap must be a list of strings');
+        assert(configSproxyd.bootstrap.length > 0,
+                'bad config: sproxyd bootstrap list is empty');
+        sproxydField = 'bootstrap';
+    }
+    if (configSproxyd.chordCos !== undefined) {
+        assert(typeof configSproxyd.chordCos === 'string',
+            'bad config: sproxyd.chordCos must be a string');
+        assert(configSproxyd.chordCos.match(/^[0-9a-fA-F]{2}$/),
+            'bad config: sproxyd.chordCos must be a 2hex-chars string');
+        sproxydField = 'chordCos';
+    }
+    return sproxydField;
+}
+
+export function locationConstraintAssert(locationConstraints) {
+    assert(typeof locationConstraints === 'object',
+        'bad config: locationConstraints must be an object');
+    Object.keys(locationConstraints).forEach(l => {
+        assert(typeof locationConstraints[l] === 'object',
+            'bad config: locationConstraints[region] must be an object');
+        assert(typeof locationConstraints[l].type === 'string',
+            'bad config: locationConstraints[region].type is ' +
+            'mandatory and must be a string');
+        assert(typeof locationConstraints[l].legacyAwsBehavior
+            === 'boolean',
+            'bad config: locationConstraints[region]' +
+            '.legacyAwsBehavior is mandatory and must be a boolean');
+        assert(typeof locationConstraints[l].details
+            === 'object',
+            'bad config: locationConstraints[region].details is ' +
+            'mandatory and must be an object');
+        const details = locationConstraints[l].details;
+        const stringFields = [
+            'awsEndpoint',
+            'bucketName',
+            'credentialsProfile',
+        ];
+        stringFields.forEach(field => {
+            if (details[field] !== undefined) {
+                assert(typeof details[field] === 'string',
+                    `bad config: ${field} must be a string`);
+            }
+        });
+    });
+}
 /**
  * Reads from a config file and returns the content as a config object
  */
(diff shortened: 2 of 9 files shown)


dora-korpar:
@ironman-machine gogo

lib/Config.js
```diff
@@ -9,6 +9,28 @@ import authDataChecker from './auth/in_memory/checker';
...
+
+function sproxydAssert(configSproxyd, cb) {
+    if (configSproxyd.bootstrap !== undefined) {
+        assert(Array.isArray(configSproxyd.bootstrap)
+            && configSproxyd.bootstrap
+                .every(e => typeof e === 'string'),
+            'bad config: sproxyd.bootstrap must be a list of strings');
+        assert(configSproxyd.bootstrap.length > 0,
+                'sproxyd bootstrap list is empty');
+        // thisSproxyd.bootstrap = configSproxyd.bootstrap;
```
LaurenSpiegel: ?

lib/Config.js
```diff
@@ -9,6 +9,28 @@ import authDataChecker from './auth/in_memory/checker';
 const defaultHealthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
 
 const defaultLocalCache = { host: '127.0.0.1', port: 6379 };
+
+function sproxydAssert(configSproxyd, cb) {
```
LaurenSpiegel: why a cb if all synchronous?

lib/Config.js
```diff
@@ -9,6 +9,28 @@ import authDataChecker from './auth/in_memory/checker';
...
+                'sproxyd bootstrap list is empty');
+        // thisSproxyd.bootstrap = configSproxyd.bootstrap;
+        cb('bootstrap');
+    }
+    if (configSproxyd.chordCos !== undefined) {
+        assert(typeof configSproxyd.chordCos === 'string',
+            'bad config: sproxyd.chordCos must be a string');
+        assert(configSproxyd.chordCos.match(/^[0-9a-fA-F]{2}$/),
+            'bad config: sproxyd.chordCos must be a 2hex-chars string');
+        // thisSproxyd.chordCos =
```
LaurenSpiegel: ?

lib/Config.js
```diff
@@ -67,13 +89,76 @@ class Config {
...
         }
 
-        assert(typeof config.regions === 'object',
-               'bad config: the list of regions is mandatory');
-        assert(Object.keys(config.regions).every(
-               r => typeof r === 'string' && config.regions[r] instanceof Array
-               && config.regions[r].every(e => typeof e === 'string')),
-               'bad config: regions must be a set of {region: [endpoints]}');
-        this.regions = config.regions;
+        // legacy
```
LaurenSpiegel: we should also check to make sure they are not mixing legacy with the new set up

lib/Config.js
```diff
@@ -95,25 +180,17 @@ class Config {
             assert(typeof config.usEastBehavior === 'boolean');
             this.usEastBehavior = config.usEastBehavior;
         }
+        // legacy
```
LaurenSpiegel: we should check to make sure not mixing legacy and new format here as well

LaurenSpiegel: we should check to make sure not mixing legacy and new format here as well

LaurenSpiegel: we should check to make sure not mixing legacy and new format here as well

lib/Config.js
```diff
@@ -329,8 +406,19 @@ class Config {
             }
             this.authData = authData;
         }
-        if (process.env.S3SPROXYD) {
```
LaurenSpiegel: don't we still need this for legacy?  does federation use this?

lib/utils.js
```diff
@@ -62,28 +62,17 @@ utils.getBucketNameFromHost = function getBucketNameFromHost(request) {
...
-/**
  * Get all valid endpoints, according to our configuration
  *
  * @returns {string[]} - list of valid endpoints
  */
 utils.getAllEndpoints = function getAllEndpoints() {
-    return Object.keys(config.regions)
-        .map(r => config.regions[r])
-        .reduce((a, b) => a.concat(b));
+    if (config.regions !== undefined) {
```
LaurenSpiegel: add note that this is legacy

tests/unit/utils.js
```diff
@@ -78,38 +72,14 @@ describe('utils.getBucketNameFromHost', () => {
...
-    });
-
-    it('should return regions from config', () => {
-        const allRegions = utils.getAllRegions();
-
-        assert(allRegions.indexOf('localregion') >= 0);
-    });
-});
-
 describe('utils.getAllEndpoints', () => {
```
LaurenSpiegel: please add tests that test the legacy and that test the new config (you'll have to set up the config based on what you are testing)

lib/Config.js
```diff
@@ -9,6 +9,25 @@ import authDataChecker from './auth/in_memory/checker';
...
 const defaultLocalCache = { host: '127.0.0.1', port: 6379 };
+
+function sproxydAssert(configSproxyd, cb) {
+    if (configSproxyd.bootstrap !== undefined) {
+        assert(Array.isArray(configSproxyd.bootstrap)
+            && configSproxyd.bootstrap
+                .every(e => typeof e === 'string'),
+            'bad config: sproxyd.bootstrap must be a list of strings');
+        assert(configSproxyd.bootstrap.length > 0,
+                'sproxyd bootstrap list is empty');
```
bennettbuchanan: Perhaps add "bad config:" here as well and use property syntax as in the other assertion statements.

lib/Config.js
```diff
@@ -67,13 +86,76 @@ class Config {
...
+                'bad config: regions must be a set of {region: [endpoints]}');
+            this.regions = config.regions;
+        }
+
+        this.locationConstraints = {};
+        if (config.locationConstraints !== undefined) {
+            assert(typeof config.locationConstraints === 'object',
+                'bad config: locationConstraints must be an object');
+            Object.keys(config.locationConstraints).forEach(l => {
+                assert(typeof config.locationConstraints[l].type === 'string' &&
```
bennettbuchanan: Can this be simplified to `typeof l.type === 'string'`? (And throughout these checks.)

lib/Config.js
```diff
@@ -67,13 +86,76 @@ class Config {
...
+                    config.locationConstraints[l].type;
+                const info = config.locationConstraints[l].information;
+                const stringFields = [
+                    'region',
+                    'awsEndpoint',
+                    'bucketName',
+                    'credentialsProfile',
+                ];
+                stringFields.forEach(field => {
+                    if (info.field !== undefined) {
```
bennettbuchanan: If these are optional, could you instead loop through the existing properties of `info.field`?

lib/Config.js
dora-korpar: Object.keys makes an array of strings of all the keys, so just checking l.type would always be a string. To actually check the value in the object, I have to do it the long way :/

bennettbuchanan: Maybe creating a variable with the value of `config.locationConstraints[l]` and then using it throughout would save some lines.

bennettbuchanan: It's minor though 😄

lib/Config.js
dora-korpar: I actually think doing it this way is more concise because otherwise I'd have to generate a new array containing all the fields for every locationConstraint, and then loop through that, right?

lib/Config.js
dora-korpar: good catch 😄

lib/Config.js
bennettbuchanan: Sorry, properties of `info`.

config.json
```diff
@@ -1,24 +1,59 @@
...
-                      "s3-external-1.amazonaws.com",
-                      "s3.us-east-1.amazonaws.com"],
-        "us-west-1": ["s3.us-west-1.amazonaws.com"],
-        "us-west-2": ["s3-us-west-2.amazonaws.com"],
-        "us-gov-west-1": ["s3-us-gov-west-1.amazonaws.com",
-                          "s3-fips-us-gov-west-1.amazonaws.com"],
-        "localregion": ["localhost"],
-        "test-region": ["s3.scality.test"],
-        "docker-region": ["s3.docker.test"]
+    "locationConstraints": {
```
nicolas2bert: Since AWS JS SDK send a request with locationConstraint us-east-1 if no locationConstraint provided.
Should we add another section named us-east-1? :
```
"us-east-1": {
             "type": "aws_s3",
             "information": {
                 "region": "us-east-1",
                 }
             }
         },
```

LaurenSpiegel:
We will have to update the readme to describe the new config but since this is backwards compatible that can wait until we actually have the multiple backend pr's merged.  Please open an issue to update the readme later.

nicolas2bert:
However, we do not need to make any change in https://github.com/scality/S3/blob/master/docker-entrypoint.sh. 
- credentials
- hostname
- log level
still work after these config changes.

LaurenSpiegel:
@nicolas2bert, thanks. We should consider not building docker off of master while we are in a config "transient" state since it might be confusing to people who look at the config.json.  Just build off of rel/6.4.  What do you think?

nicolas2bert:
@LaurenSpiegel  Agreed - we should wait until the hole multi-backend feature is merged.

LaurenSpiegel:
A couple of comments from @GiorgioRegni:

1) change scality_s3 to just scality
2) change "information" to "details"
3) only keep region for real aws configs.  take out in all others.
4) add at the level of type and information, a new entry called "legacyBehavior" and can be set to true or false.

nicolas2bert:
5. get rid of `"usEastBehavior": false/true`

Why are we not calling 4. new entry `"usEastBehavior"` instead of `"legacyBehavior"` to keep it consistent?

LaurenSpiegel:
@nicolas2bert, agree that we should remove the old usEastBehavior config line as it will be replaced with this new part of the locationConstraint config.  I think legacyAwsBehavior is a better descriptor but if others think usEastBehavior is clearer I could be swayed.

config.json
```diff
@@ -1,24 +1,69 @@
...
+        "aws-us-east-1": {
+            "type": "aws_s3",
+            "legacyAwsBehavior": true,
+            "details": {
+                "region": "us-east-1",
```
nicolas2bert: @dora-korpar @LaurenSpiegel - I guess this `region` is for the location of the bucket "premadebucket" in the real AWS.
So it will be possible to have` "legacyAwsBehavior": true,` with `"region": "us-east-1",`
Right?
_Suggestion:_ Won't it be clearer to have `legacyAwsBehavior` in the `details` block with all the bucket's informations.

LaurenSpiegel: 1) it is possible to have legacyAwsBehavior true and us-east-1.  we are controlling the behavior (errors, etc.).  the data is just being stored on AWS in the background.

2) in fact, we could get rid of region altogether in the aws_s3 type.  if we have the endpoint that is enough.  we just need to know where to call.

3) let's keep legacyAwsBehavior at the current level (same as type and details).  the details are really about the details to hit the right data endpoint.  legacyAwsBehavior is about behavior.

nicolas2bert: Ok - it is clear - So, as you said (and to avoid confusion) we can get rid of the regions

tests/unit/legacyTestConfig.json
```diff
@@ -0,0 +1,56 @@
+{
```
LaurenSpiegel: Let's put this into a subdir (it's possible there will be more of these as the project evolves).  perhaps, call it testConfigs

lib/Config.js
```diff
@@ -9,6 +9,27 @@ import authDataChecker from './auth/in_memory/checker';
 const defaultHealthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
 
 const defaultLocalCache = { host: '127.0.0.1', port: 6379 };
+
+function sproxydAssert(configSproxyd) {
```
LaurenSpiegel: please add tests for this function

lib/Config.js
```diff
@@ -67,13 +88,83 @@ class Config {
...
+            assert(typeof config.regions === 'object',
+                'bad config: regions must be an object');
+            assert(Object.keys(config.regions).every(
+                r => typeof r === 'string' && config.regions[r] instanceof Array
+                && config.regions[r].every(e => typeof e === 'string')),
+                'bad config: regions must be a set of {region: [endpoints]}');
+            this.regions = config.regions;
+        }
+
+        if (config.locationConstraints !== undefined) {
```
LaurenSpiegel: please add tests for this logic.  if it is easier to abstract into a function in order to test, do that.

dora-korpar:
@ironman-machine go

dora-korpar:
e2e here:
http://ci.ironmann.io/gh/scality/Integration/9516

dora-korpar:
new e2e: http://ci.ironmann.io/gh/scality/Integration/9518

dora-korpar:
This one's gonna pass:
http://ci.ironmann.io/gh/scality/Integration/9537

config.json
```diff
@@ -1,24 +1,68 @@
...
+        },
+        "mem": {
+            "type": "mem",
+            "legacyAwsBehavior": false,
+            "details": {
+            }
+        }
+    },
+    "restEndpoints": {
+        "127.0.0.1": "scality-us-east-1",
```
LaurenSpiegel: let's change this to "file" so the default localhost backend will be file.

tests/unit/testConfigs/locConstraintAssert.js
```diff
@@ -0,0 +1,118 @@
+import assert from 'assert';
+import { locationConstraintAssert } from '../../../lib/Config';
+
+function generateConfig(testKey) {
```
LaurenSpiegel: as discussed, create a default config and then here use object assign to create a new object and switch out the item you want to change.

dora-korpar: After more discussion, creating a LocationConstraints class and passing it objects mimicking the config with item to change

dora-korpar:
Passed: http://ci.ironmann.io/gh/scality/Integration/9546

PR #576, opened 2017-01-30
FIX: Exit server when connections are drained

### LINKED ISSUE #550: TERM signal handler should wait
The signal handler for TERM exits the process just after calling `server.close()` without waiting for in-flight requests to be processed.

This prevents graceful shutdown after handling of in-flight requests and causes errors on the client side. Doing `process.exit` in the callback given to `server.close()` would ensure that no in-flight request is interrupted, making it easier to do seamless upgrades or container recycling on certain events (such as when a certain memory threshold is crossed).

RachedBenMustapha:
This allows seamless exits/restarts where already-connected clients
won't get any errors for in-flight requests.

Fix #550

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -84,8 +84,7 @@ class S3Server {
      */
     cleanUp() {
         logger.info('server shutting down');
-        this.server.close();
-        process.exit(0);
+        this.server.close(() => process.exit(0));
     }
 
     caughtExceptionShutdown() {


RachedBenMustapha:
@ironman-machine try

LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/9101

PR #578, opened 2017-01-31
FIX: Kill worker less zealously

LaurenSpiegel:
Only set one timeout to check worker killed.
Check that worker.suicide not already
set before killing.

Should help prevent seeing errors like those described here: https://github.com/scality/S3/issues/546

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -95,12 +95,6 @@ class S3Server {
         // Will close all servers, cause disconnect event on master and kill
         // worker process with 'SIGTERM'.
         this.worker.kill();
-        const killTimer = setTimeout(() => {
-            if (!this.worker.isDead()) {
-                this.worker.kill('SIGKILL');
-            }
-        }, 2000);
-        killTimer.unref();
     }
 }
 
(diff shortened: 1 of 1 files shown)


LaurenSpiegel:
@ironman-machine please

LaurenSpiegel:
http://ci.ironmann.io/gh/scality/Integration/9182

LaurenSpiegel:
end to end passed http://ci.ironmann.io/gh/scality/Integration/9182

lib/server.js
```diff
@@ -135,7 +129,10 @@ export default function main() {
                 workerPid: worker.process.pid,
             });
             setTimeout(() => {
-                if (!worker.isDead()) {
+                // TODO: When upgrading to Node v6,
+                // instead of !worker.suicide,
+                // use !worker.exitedAfterDisconnect
```
ttvi: Why using `!worker.exitedAfterDisconnect` in node v6? There is still IPC issue: http://ci.ironmann.io/gh/scality/Integration/10757.

LaurenSpiegel: worker.suicide was deprecated -- https://nodejs.org/dist/latest-v6.x/docs/api/cluster.html#cluster_worker_suicide

This was not intended to definitively fix the IPC issue.  There is likely a race condition in the node code, but we should only end up in this situation and see this error if there was another problem causing a crash.

PR #579, opened 2017-02-01
FT: log uploadId for initiateMPU

### LINKED ISSUE #569: Add uploadId to the log on a successful Initiate MPU
Should target `6.2.5`

electrachong:
FIXES #569

### DIFF
--- a/lib/api/initiateMultipartUpload.js
+++ b/lib/api/initiateMultipartUpload.js
@@ -141,7 +141,8 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
                             });
                             return callback(err);
                         }
-                        log.trace('successfully initiated mpu', { uploadId });
+                        log.addDefaultFields({ uploadId });
+                        log.trace('successfully initiated mpu');
                         return callback(null, xml);
                     });
             });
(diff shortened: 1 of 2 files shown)


lib/api/initiateMultipartUpload.js
```diff
@@ -134,7 +134,16 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
                     cipherBundle, metadataStoreParams,
-                    log, err => callback(err, xml));
+                    log, err => {
+                        if (err) {
+                            log.trace('error storing multipart object', {
+                                error: err,
+                            });
+                            return callback(err);
+                        }
+                        log.trace('successfully initiated mpu', { uploadId });
```
electrachong: we only want to log on success, right?

electrachong:
@ironman-machine

electrachong:
end-to-end: http://ci.ironmann.io/gh/scality/Integration/9169

lib/api/initiateMultipartUpload.js
rahulreddy: Yes, we only want to log on success but we need to add this as a default field so that it shows up in the logs after this step. Something like
```js
log.addDefaultFields({ uploadId });
```
This call adds it as a field for all the log entries which are going to be printed after this step.
Sorry, I should have mentioned this in the issue.

electrachong: is https://github.com/scality/S3/pull/579/commits/356d8b2dddb854d54d8a3a26e2511a9a854b1b8d better?

electrachong:
new end-to-end: http://ci.ironmann.io/gh/scality/Integration/9179

PR #580, opened 2017-02-01
FIX: abortMpu return 404 only if usEastBehavior

### LINKED ISSUE #568: Abort MPU returns 404 for all regions when Upload Id doesn't exist

electrachong:
FIX #568

Change response so returns 204 instead of 404 when UploadID does not exist, unless usEastBehavior is set to true.

Note: Since changes to save locationConstraint in metadata have not been introduced yet in this branch, we do not check region and rely on usEastBehavior alone.

### DIFF
--- a/config.json
+++ b/config.json
@@ -34,5 +34,6 @@
     },
     "healthChecks": {
         "allowFrom": ["127.0.0.1/8", "::1"]
-    }
+    },
+    "usEastBehavior": false
 }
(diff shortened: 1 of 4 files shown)


electrachong:
end-to-end: http://ci.ironmann.io/gh/scality/Integration/9192

electrachong:
Closed in favor of https://github.com/scality/S3/pull/589

PR #581, opened 2017-02-02
REFACTO: Metadata object using class

alexandre-merle:
Using a class to compute regular object
metadata.

Only manage the Put here, not the Get


alexandre-merle:
Retargetting to master

PR #582, opened 2017-02-02
REFACTOR: Metadata object using class

alexandre-merle:
Using a class to compute regular object
metadata

### DIFF
--- a/lib/metadata/models/object/ObjectMD.js
+++ b/lib/metadata/models/object/ObjectMD.js
@@ -1,0 +1,611 @@
+
+// Version 2 changes the format of the data location property
+const modelVersion = 2;
+
+/**
+ * Class to manage metadata object for regular s3 objects (instead of
+ * mpuPart metadata for example)
+ */
+module.exports = class ObjectMD {
+
+    /**
+     * @constructor
+     *
+     * @param {number} version - Version of the metadata model
+     */
+    constructor(version) {
+        const now = new Date().toJSON();
+        this._data = {
+            'md-model-version': version || modelVersion,
+            'owner-display-name': '',
+            'owner-id': '',
+            'cache-control': '',
+            'content-disposition': '',
+            'content-encoding': '',
+            'expires': '',
+            'content-length': 0,
+            'content-type': '',
+            'last-modified': now,
+            'content-md5': '',
+            // simple/no version. will expand once object versioning is
+            // introduced
+            'x-amz-version-id': 'null',
+            'x-amz-server-version-id': '',
+            // TODO: Handle this as a utility function for all object puts
+            // similar to normalizing request but after checkAuth so
+            // string to sign is not impacted.  This is GH Issue#89.
+            'x-amz-storage-class': 'STANDARD',
+            'x-amz-server-side-encryption': '',
+            'x-amz-server-side-encryption-aws-kms-key-id': '',
+            'x-amz-server-side-encryption-customer-algorithm': '',
+            'x-amz-website-redirect-location': '',
+            'acl': {
+                Canned: 'private',
+                FULL_CONTROL: [],
+                WRITE_ACP: [],
+                READ: [],
+                READ_ACP: [],
+            },
+            'key': '',
+            'location': [],
+            'isNull': '',
+            'nullVersionId': '',
+            'isDeleteMarker': '',
+            'versionId': undefined, // If no versionId, it should be undefined
+            'tags': {},
+        };
+    }
+
+    /**
+     * Returns metadata model version
+     *
+     * @return {number} Metadata model version
+     */
+    getModelVersion() {
+        return this._data['md-model-version'];
+    }
+
+    /**
+     * Set owner display name
+     *
+     * @param {string} displayName - Owner display name
+     * @return {ObjectMD} itself
+     */
+    setOwnerDisplayName(displayName) {
+        this._data['owner-display-name'] = displayName;
+        return this;
+    }
+
+    /**
+     * Returns owner display name
+     *
+     * @return {string} Onwer display name
+     */
+    getOwnerDisplayName() {
+        return this._data['owner-display-name'];
+    }
+
+    /**
+     * Set owner id
+     *
+     * @param {string} id - Owner id
+     * @return {ObjectMD} itself
+     */
+    setOwnerId(id) {
+        this._data['owner-id'] = id;
+        return this;
+    }
+
+    /**
+     * Returns owner id
+     *
+     * @return {string} owner id
+     */
+    getOwnerId() {
+        return this._data['owner-id'];
+    }
+
+    /**
+     * Set cache control
+     *
+     * @param {string} cacheControl - Cache control
+     * @return {ObjectMD} itself
+     */
+    setCacheControl(cacheControl) {
+        this._data['cache-control'] = cacheControl;
+        return this;
+    }
+
+    /**
+     * Returns cache control
+     *
+     * @return {string} Cache control
+     */
+    getCacheControl() {
+        return this._data['cache-control'];
+    }
+
+    /**
+     * Set content disposition
+     *
+     * @param {string} contentDisposition - Content disposition
+     * @return {ObjectMD} itself
+     */
+    setContentDisposition(contentDisposition) {
+        this._data['content-disposition'] = contentDisposition;
+        return this;
+    }
+
+    /**
+     * Returns content disposition
+     *
+     * @return {string} Content disposition
+     */
+    getContentDisposition() {
+        return this._data['content-disposition'];
+    }
+
+    /**
+     * Set content encoding
+     *
+     * @param {string} contentEncoding - Content encoding
+     * @return {ObjectMD} itself
+     */
+    setContentEncoding(contentEncoding) {
+        this._data['content-encoding'] = contentEncoding;
+        return this;
+    }
+
+    /**
+     * Returns content encoding
+     *
+     * @return {string} Content encoding
+     */
+    getContentEncoding() {
+        return this._data['content-encoding'];
+    }
+
+    /**
+     * Set expiration date
+     *
+     * @param {string} expires - Expiration date
+     * @return {ObjectMD} itself
+     */
+    setExpires(expires) {
+        this._data.expires = expires;
+        return this;
+    }
+
+    /**
+     * Returns expiration date
+     *
+     * @return {string} Expiration date
+     */
+    getExpires() {
+        return this._data.expires;
+    }
+
+    /**
+     * Set content length
+     *
+     * @param {number} contentLength - Content length
+     * @return {ObjectMD} itself
+     */
+    setContentLength(contentLength) {
+        this._data['content-length'] = contentLength;
+        return this;
+    }
+
+    /**
+     * Returns content length
+     *
+     * @return {number} Content length
+     */
+    getContentLength() {
+        return this._data['content-length'];
+    }
+
+    /**
+     * Set content type
+     *
+     * @param {string} contentType - Content type
+     * @return {ObjectMD} itself
+     */
+    setContentType(contentType) {
+        this._data['content-type'] = contentType;
+        return this;
+    }
+
+    /**
+     * Returns content type
+     *
+     * @return {string} Content type
+     */
+    getContentType() {
+        return this._data['content-type'];
+    }
+
+    /**
+     * Set last modified date
+     *
+     * @param {string} lastModified - Last modified date
+     * @return {ObjectMD} itself
+     */
+    setLastModified(lastModified) {
+        this._data['last-modified'] = lastModified;
+        return this;
+    }
+
+    /**
+     * Returns last modified date
+     *
+     * @return {string} Last modified date
+     */
+    getLastModified() {
+        return this._data['last-modified'];
+    }
+
+    /**
+     * Set content md5 hash
+     *
+     * @param {string} contentMd5 - Content md5 hash
+     * @return {ObjectMD} itself
+     */
+    setContentMd5(contentMd5) {
+        this._data['content-md5'] = contentMd5;
+        return this;
+    }
+
+    /**
+     * Returns content md5 hash
+     *
+     * @return {string} content md5 hash
+     */
+    getContentMd5() {
+        return this._data['content-md5'];
+    }
+
+    /**
+     * Set version id
+     *
+     * @param {string} versionId - Version id
+     * @return {ObjectMD} itself
+     */
+    setAmzVersionId(versionId) {
+        this._data['x-amz-version-id'] = versionId;
+        return this;
+    }
+
+    /**
+     * Returns version id
+     *
+     * @return {string} Version id
+     */
+    getAmzVersionId() {
+        return this._data['x-amz-version-id'];
+    }
+
+    /**
+     * Set server version id
+     *
+     * @param {string} versionId - server version id
+     * @return {ObjectMD} itself
+     */
+    setAmzServerVersionId(versionId) {
+        this._data['x-amz-server-version-id'] = versionId;
+        return this;
+    }
+
+    /**
+     * Returns server version id
+     *
+     * @return {string} server version id
+     */
+    getAmzServerVersionId() {
+        return this._data['x-amz-server-version-id'];
+    }
+
+    /**
+     * Set storage class
+     *
+     * @param {string} storageClass - Storage class
+     * @return {ObjectMD} itself
+     */
+    setAmzStorageClass(storageClass) {
+        this._data['x-amz-storage-class'] = storageClass;
+        return this;
+    }
+
+    /**
+     * Returns storage class
+     *
+     * @return {string} Storage class
+     */
+    getAmzStorageClass() {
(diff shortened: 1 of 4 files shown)


lib/services.js
```diff
@@ -252,8 +225,11 @@ export default {
                     log.warn('error parsing acl', { error: err });
                     return cb(err);
                 }
-                omVal.acl = parsedACL;
-                metadata.putObjectMD(bucketName, objectKey, omVal, log, err => {
+                md.setAcl(parsedACL);
+                log.trace('object metadata', {
+                    omVal: md.getData(),
```
electrachong: I have heard it's better not to store large objects in the logs to minimize performance and storage impact, not sure if object MD qualifies but it's something to consider

lib/services.js
```diff
@@ -264,7 +240,10 @@ export default {
                 return undefined;
             });
         } else {
-            metadata.putObjectMD(bucketName, objectKey, omVal, log, err => {
+            log.trace('object metadata', {
+                omVal: md.getData(),
```
electrachong: same note here

lib/metadata/models/object/object.js
```diff
@@ -0,0 +1,465 @@
...
+        Object.keys(data).forEach(key => { this.data[key] = data[key]; });
+        return this;
+    }
+
+    /**
+     * Returns metadata object
+     *
+     * @return {object} metadata object
+     */
+    getData() {
```
electrachong: minor: since we have a public method for accessing the property, you could name the 'data' property '_data' to follow convention, similar to what we do in BucketInfo class

lib/services.js
alexandre-merle: you're right :) i have replace what exists before, but we can consider to remove this :) Here its a trace log, so it will be log only if there is an error, no opinion on this one :)

lib/metadata/models/object/object.js
alexandre-merle: Yes, its better :)

alexandre-merle:
@electrachong yes, here there is only the put of the metadata for regular object, then i have to do the other metdata mpuObject, mpuPart, and then the get will comes later :)

alexandre-merle:
@electrachong updated, unit tests added, we can think in the futur to add more specific check in the setter of the metadata :)

lib/services.js
ThibaultRiviere: I think @electrachong is right, even if it's only trace, the logger does some operations like object copy right ?

tests/unit/metadata/models/object.js
```diff
@@ -0,0 +1,78 @@
...
+import ObjectMD from '../../../../lib/metadata/models/object/object';
+
+describe('Metadata regular object', () => {
+    let md = null;
+
+    beforeEach(() => {
+        md = new ObjectMD();
+    });
+
+    [
```
electrachong: Maybe add an explanatory comment:
`// In order: data property, value to set/get, default value`

tests/unit/metadata/models/object.js
```diff
@@ -0,0 +1,78 @@
+import assert from 'assert';
+import ObjectMD from '../../../../lib/metadata/models/object/object';
+
+describe('Metadata regular object', () => {
```
electrachong: `describe('ObjectMD class setters/getters', () => {`

tests/unit/metadata/models/object.js
```diff
@@ -0,0 +1,78 @@
...
+            FULL_CONTROL: ['id'],
+            WRITE_ACP: ['id'],
+            READ: ['id'],
+            READ_ACP: ['id'],
+        }],
+        ['Key', null, ''],
+        ['Key', 'key'],
+        ['Location', null, []],
+        ['Location', ['location1']],
+    ].forEach(test => {
```
electrachong: even better:
create some variable names and the test will be a lot easier to read:
```
const property = test[0];
const testValue = test[1];
const defaultValue =  test[2]
```

do you agree? :smile:

lib/services.js
alexandre-merle: im okay to remove it, but its not my goal, since it can be useful for s3 debugging @LaurenSpiegel ?

tests/unit/metadata/models/object.js
alexandre-merle: i can do that ^^

tests/unit/metadata/models/object.js
alexandre-merle: take more space on the 80 chars :p joking, i can do that as well ^^

alexandre-merle:
@electrachong updated :)

lib/metadata/models/object/object.js
```diff
@@ -0,0 +1,465 @@
...
+        return this._data.location;
+    }
+
+    /**
+     * Set custom headers
+     *
+     * @param {object} data - Headers
+     * @return {ObjectMD} itself
+     */
+    setHeaders(data) {
```
LaurenSpiegel: we should do some sanity check to make sure these are all setting x-amz-meta headers.  I do not want to unintentionally overwrite another property like the location.

also, let's call this setUserMetadata

lib/services.js
LaurenSpiegel: Let's just keep the log for the acl's.  Everything else should be pretty straightforward.

tests/unit/metadata/models/object.js
electrachong: okay, thank you :)

electrachong: but why is it you only test default value for many of them? Do you not want to test all the setter methods, or did I misunderstand?

LaurenSpiegel:
@alexandre-merle, please update so can be merged

lib/metadata/models/object/object.js
```diff
@@ -0,0 +1,465 @@
...
+     *
+     * @param {object} acl - Access control list
+     * @param {string} acl.Canned -
+     * @param {string[]} acl.FULL_CONTROL -
+     * @param {string[]} acl.WRITE_ACP -
+     * @param {string[]} acl.READ -
+     * @param {string[]} acl.READ_ACP -
+     * @return {ObjectMD} itself
+     */
+    setAcl(acl) {
```
rahulreddy: This needs a class to manage the ACL properties as well. Not in this PR, but definitely needs one.

lib/metadata/models/object/object.js
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    /**
+     * Set location
+     *
+     * @param {string[]} location - location
+     * @return {ObjectMD} itself
+     */
+    setLocation(location) {
+        this._data.location = location;
```
rahulreddy: Should clone object to handle array references here.

alexandre-merle:
Updated :)

lib/metadata/models/object/object.js
```diff
@@ -0,0 +1,465 @@
+
+// Version 2 changes the format of the location property
+const modelVersion = 2;
+
+/**
+ * Class to manage metadata object for regular s3 objects
```
LaurenSpiegel: please update the comment so it is clear you are distinguishing between the metadata for an mpu part

lib/metadata/models/object/object.js
```diff
@@ -0,0 +1,472 @@
...
+     */
+    setLocation(location) {
+        const newLocation = location ? [] : undefined;
+        if (location) {
+            const len = location.length;
+            for (let i = 0; i < len; ++i) {
+                newLocation.push(location[i]);
+            }
+        }
+        this._data.location = location;
```
LaurenSpiegel: I think you meant newLocation here.  In any case, each location is an object (other than the legacy version where each location is just a string) so you are just sending references to the the objects into a new array.  I would leave this as it was without the newLocation and just use this._data.location = location

lib/metadata/models/object/object.js
LaurenSpiegel: not done?

LaurenSpiegel:
@alexandre-merle, it looks like you missed some earlier comments here?

lib/metadata/models/object/object.js
rahulreddy: I guess this is a problem for some other time.

alexandre-merle:
@LaurenSpiegel indeed yes, i just missed some comments x)

alexandre-merle:
updated :)

lib/metadata/models/object/object.js
LaurenSpiegel: how is this different from setUserMetadata?  what is the difference?

bennettbuchanan:
Update: Taking over this branch so we can use this way of storing objectMD for bucket replication. 😃

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

lib/metadata/models/object/ObjectMD.js
```diff
@@ -0,0 +1,613 @@
...
+
+    /**
+     * Set custom meta headers
+     *
+     * @param {object} data - Meta headers
+     * @return {ObjectMD} itself
+     */
+    setUserMetadata(data) {
+        Object.keys(data).forEach(key => {
+            if (key.startsWith('x-amz-meta')) {
```
jonathan-gramain: minor: shouldn't the prefix checked against `x-amz-meta-` instead? (with trailing `-`) not sure it really matters but it looks more correct - could suddenly matter e.g. if amazon decided to use some `x-amz-metacool-header` to annoy devs :rofl:

lib/metadata/models/object/ObjectMD.js
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    /**
+     * Set custom headers
+     *
+     * @param {object} data - Headers
+     * @return {ObjectMD} itself
+     */
+    setHeaders(data) {
+        Object.keys(data).forEach(key => {
```
jonathan-gramain: May be done as `Object.assign(this._data, data)`?

lib/metadata/models/object/ObjectMD.js
```diff
@@ -0,0 +1,613 @@
...
+     * @return {object} tags object
+     */
+    getTags() {
+        return this._data.tags;
+    }
+
+    /**
+     * Set custom meta headers
+     *
+     * @param {object} data - Meta headers
```
jonathan-gramain: I suggest renaming `data` to `metaHeaders`

lib/metadata/models/object/ObjectMD.js
```diff
@@ -0,0 +1,613 @@
...
+        return this;
+    }
+
+    /**
+     * Set custom headers
+     *
+     * @param {object} data - Headers
+     * @return {ObjectMD} itself
+     */
+    setHeaders(data) {
```
jonathan-gramain: I suggest renaming `data` to `headers` or `customHeaders`

lib/services.js
```diff
@@ -95,54 +96,36 @@ const services = {
...
-        // Version 2 changes the format of the location property
-        omVal['md-model-version'] = 2;
+        const md = new ObjectMD();
         // If an IAM user uploads a resource,
         // the owner should be the parent account.
         // http://docs.aws.amazon.com/AmazonS3/
         // latest/dev/access-control-overview.html
-        omVal['owner-display-name'] =
-            authInfo.getAccountDisplayName();
+        md.setOwnerDisplayName(authInfo.getAccountDisplayName());
```
jonathan-gramain: Why not chaining this call too

lib/services.js
```diff
@@ -95,54 +96,36 @@ const services = {
...
-
-        omVal.acl = {
-            Canned: 'private',
-            FULL_CONTROL: [],
-            WRITE_ACP: [],
-            READ: [],
-            READ_ACP: [],
-        };
-
+        if (headers && headers['x-amz-website-redirect-location']) {
```
jonathan-gramain: Saving the website redirect location seems to have been introduced in this commit, maybe deserve a special mention or separate commit (if not separate issue)?

lib/metadata/models/object/ObjectMD.js
```diff
@@ -0,0 +1,613 @@
...
+        });
+        return this;
+    }
+
+    /**
+     * Returns metadata object
+     *
+     * @return {object} metadata object
+     */
+    getData() {
```
jonathan-gramain: I find the name `getData` confusing, it can be taken for the actual object data. Could be name `getValue` instead for example, or `asJSObject` to point the fact that we're returning the native JS object.

tests/unit/metadata/models/object.js
```diff
@@ -0,0 +1,93 @@
...
+        const defaultValue = test[2];
+        const testName = testValue === null ? 'get default' : 'get/set';
+        it(`${testName}: ${property}`, () => {
+            if (testValue !== null) {
+                md[`set${property}`](testValue);
+            }
+            const value = md[`get${property}`]();
+            if ((testValue !== null && typeof testValue === 'object') ||
+                typeof defaultValue === 'object') {
+                assert.deepStrictEqual(value, testValue || defaultValue);
```
jonathan-gramain: Shouldn't you branch the test here depending on if `testValue !== null`? Just to handle the potential case where `testValue === 0`.

lib/services.js
bennettbuchanan: I guess this ordering is just left over. I've added it to the chain. 😄

lib/metadata/models/object/ObjectMD.js
bennettbuchanan: That's true, as there are no conditional checks here.

lib/metadata/models/object/ObjectMD.js
bennettbuchanan: Updated.

lib/metadata/models/object/ObjectMD.js
bennettbuchanan: Good point. Updated.

lib/metadata/models/object/ObjectMD.js
bennettbuchanan: Updated.

lib/metadata/models/object/ObjectMD.js
```diff
@@ -0,0 +1,613 @@
...
+     */
+    setVersionId(versionId) {
+        this._data.versionId = versionId;
+        return this;
+    }
+
+    /**
+     * Get metadata versionId value
+     *
+     * @return {boolean} The object versionId
```
bennettbuchanan: Good catch. Thanks!

tests/unit/metadata/models/object.js
bennettbuchanan: Good catch. Updated.

lib/services.js
bennettbuchanan: Looks like this objectMD value was originally handled by this loop: https://github.com/scality/S3/pull/582/files#diff-304c00e406caafeefbc9c0a7d27086d8L188. I've added the check to instead include the conditional logic in the setUserMetadata method.

lib/metadata/models/object/ObjectMD.js
bennettbuchanan: Updated.

lib/metadata/models/object/ObjectMD.js
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    /**
+     * Set location
+     *
+     * @param {string[]} location - location
+     * @return {ObjectMD} itself
+     */
+    setLocation(location) {
+        this._data.location = JSON.parse(JSON.stringify(location));
```
bennettbuchanan: This was to clone the object. I guess we should be cloning any object properties though.

jonathan-gramain: An `Object.assign()` instead would not work? (I assume it's because we need a deep copy here)

bennettbuchanan:
@ironman-machine try

lib/services.js
bennettbuchanan: After conversation, keeping this as is.

lib/metadata/models/object/object.js
bennettbuchanan: Keeping this as is due to performance considerations, and since locations should not be modified outside of this class.

lib/metadata/models/object/ObjectMD.js
```diff
@@ -595,7 +595,7 @@ module.exports = class ObjectMD {
      * @param {object} headers - Headers
      * @return {ObjectMD} itself
      */
-    setHeaders(headers) {
+    overrideMetadataValues(headers) {
```
LaurenSpiegel: please update description of function above to: overrideMetadataValues (used for complete MPU and object copy)

bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed on retry: http://ci.ironmann.io/gh/scality/Integration/13243.

PR #585, opened 2017-02-02
cleanup: send query params to arsenal auth method

rahulreddy:


rahulreddy:
End to end http://ci.ironmann.io/gh/scality/Integration/9232

lib/utapi/utilities.js
```diff
@@ -49,7 +49,11 @@ function _listBucketMetrics(host,
             }
         });
     });
-    auth.client.generateV4Headers(request, '', accessKey, secretKey, 's3');
+    // TODO: cleanup with refactor of generateV4Headers
+    request.path = '/buckets';
+    auth.client.generateV4Headers(request, { Action: 'ListMetrics' },
+        accessKey, secretKey, 's3');
+    request.path = '/buckets?Action=ListMetrics';
```
electrachong: sorry it is not obvious to me : why do you change the request path after setting it before generating the headers?

rahulreddy: So, the problem was that `auth.client.generateV4Headers` expects `request.path` to not have any query params in it, but we cannot make a HTTP request to Utapi without those query params.
We would be refactoring the Arsenal method soon, so for now, we remove the query params before generating the headers and put them back for making the request.

electrachong: ok

DavidPineauScality:
Build finally succeeded (after a few flaky failures and retries) at https://ci.ironmann.io/gh/scality/Integration/9251#build-parameters
Merging.

PR #587, opened 2017-02-03
Forward rel/6.3 to rel/6.4

DavidPineauScality:


LaurenSpiegel:
closing since we will first forward port from 6.2.5

PR #589, opened 2017-02-03
FIX: abortMpu returns 204 if no such uploadId

### LINKED ISSUE #568: Abort MPU returns 404 for all regions when Upload Id doesn't exist

electrachong:
FIX #568

### DIFF
--- a/lib/api/multipartDelete.js
+++ b/lib/api/multipartDelete.js
@@ -1,4 +1,5 @@
 import async from 'async';
+import { errors } from 'arsenal';
 
 import constants from '../../constants';
 import data from '../data/wrapper';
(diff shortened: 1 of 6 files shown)


tests/unit/api/multipartDelete.js
```diff
@@ -0,0 +1,73 @@
...
+                query: `uploadId=${testUploadId}`,
+            };
+            next(null, deleteMpuRequest);
+        },
+        (deleteMpuRequest, next) =>
+            multipartDelete(authInfo, deleteMpuRequest, log, next),
+    ], callback);
+}
+
+describe.only('Multipart Delete API', () => {
```
LaurenSpiegel: nix the only

tests/unit/api/multipartDelete.js
```diff
@@ -0,0 +1,73 @@
...
+    beforeEach(() => {
+        cleanup();
+    });
+    afterEach(() => {
+        cleanup();
+    });
+
+    it('should return 204 if uploadId does not exist', done => {
```
LaurenSpiegel: should return 204 if uploadId does not exist on multipart abort call

lib/api/multipartDelete.js
```diff
@@ -57,7 +58,16 @@ function multipartDelete(authInfo, request, log, callback) {
                 });
         },
         function checkMPUval(next) {
-            services.metadataValidateMultipart(metadataValParams, next);
+            services.metadataValidateMultipart(metadataValParams,
+                (err, mpuBucket, mpuOverviewArray) => {
+                    if (err === errors.NoSuchUpload) {
+                        log.trace('did not find valid mpu with uploadId' +
+                        `${uploadId}`, { method: 'multipartDelete' });
+                        // return 404 if usEastBehavior is enabled in config
```
LaurenSpiegel: TODO in rel/6.3 and later: return 404 if usEastBehavior is enabled in config

tests/unit/api/multipartDelete.js
electrachong: ok

lib/api/multipartDelete.js
electrachong: ok

tests/unit/api/multipartDelete.js
electrachong: 👍

electrachong:
@ironman-machine

electrachong:
end-to-end: http://ci.ironmann.io/gh/scality/Integration/9286

PR #600, opened 2017-02-08
listenOn: Disable IP parsing to allow the usage of FQDNs

DavidPineauScality:
This patch removes the check of the IP's format, since we want to be able to use FQDNs instead.

Related to our ongoing multi-ports fixes.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1,7 +1,6 @@
 import assert from 'assert';
 import fs from 'fs';
 import path from 'path';
-import { ipCheck } from 'arsenal';
 
 import authDataChecker from './auth/in_memory/checker';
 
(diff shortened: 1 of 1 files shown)


LaurenSpiegel:
Is there an issue that goes with this?

DavidPineauScality:
Not at the moment, since it was a "on-the-fly-fix" (meaning I made it through experimentation with @xaf-scality) to put the finishing touches to the rc4

LaurenSpiegel:
@DavidPineauScality, if we really need to be able to support FQDNs, let's just get rid of the assertion without the added comment lines.  If the listenOn address is incorrect, it seems this will be discovered rather quickly without the assertion. Do you agree? Also, there is a linting error since you are no longer using ipCheck but it is defined.

DavidPineauScality:
Well, more than "really need", it's often easier to remember machines through their fqdns, and as such we tend to configure them when available.
I'm fine with removing the bit of code altogether, along with the comment (which was mostly there for doc purposes)

DavidPineauScality:
rebased + fixed (or should be)

PR #601, opened 2017-02-08
Forward rel/6.3 to rel6/4

nicolas2bert:
Forward rel/6.3 to rel6/4
@electrachong they were two conflicts.
Feel free to review :)


nicolas2bert:
E2E: http://ci.ironmann.io/gh/scality/Integration/9443

PR #602, opened 2017-02-08
Fix/disable large mpu retry

electrachong:
Review https://github.com/scality/S3/commit/2284dd9e7b605586b9eb1eb2b246f0f575ccfde7 only, the code change accidentally made its way into the previous commit.

FIXES scality/Integration#449

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/bigMpu.js
+++ b/tests/functional/aws-node-sdk/test/object/bigMpu.js
@@ -36,9 +36,6 @@ describe('large mpu', function tester() {
     before(done => {
         const config = getConfig('default', { signatureVersion: 'v4' });
         s3 = new S3(config);
-        // disable retries so sdk will not resend request because complete mpu
-        // is taking long, causing test to fail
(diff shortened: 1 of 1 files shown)


electrachong:
end-to-end: http://ci.ironmann.io/gh/scality/Integration/9444

electrachong:
@bennettbuchanan The commit actually disables retries, so it should not fail due to a retry.
However, the test may still fail if it takes too long to respond and hangs up the socket. I added a comment to reference the current issue. Let me know if it's clear to you.

bennettbuchanan:
Ah, I had assumed that if it got to the point of needing to retry the test, then it means the test is failed instead of letting it retry repeatedly. 

It sounds like it would be clear that the request was taking too long.

electrachong:
@bennettbuchanan Oh, OK, I think I get what you mean about the original comment being ambiguous. The retry was actually causing the test to fail because the first complete mpu request would delete some stored parts by the time the second request is sent by aws-node-sdk, and that's why we disabled it. I clarified the retry may cause the test to fail.

PR #603, opened 2017-02-08
FT: respond NotImplemented to unsupported queries

### LINKED ISSUE #568: Abort MPU returns 404 for all regions when Upload Id doesn't exist

electrachong:
Depends on https://github.com/scality/Integration/pull/451

FIXES #568


nicolas2bert:
@electrachong - Is #568 fixed by this PR?

lib/routes.js
```diff
@@ -22,6 +22,24 @@ const routeMap = {
     OPTIONS: routeOPTIONS,
 };
 
+// Queries supported by AWS that we do not currently support.
+const unsupportedQueries = [
```
nicolas2bert: Suggestion: Why not exporting `unsupportedQueries` and then importing it into the different file when needed.

electrachong:
@nicolas2bert OK, thanks for the catch

lib/routes.js
electrachong: OK, I moved it to routeUtils :)

electrachong:
here is an end-to-end: http://ci.ironmann.io/gh/scality/Integration/9452

lib/routes/routesUtils.js
```diff
@@ -541,4 +541,22 @@ const routesUtils = {
     },
 };
 
+// Queries supported by AWS that we do not currently support.
```
nicolas2bert: _Suggestion_: adding `unsupportedQueries` to `routesUtils`:
```
const routesUtils = {
    // Queries supported by AWS that we do not currently support.
    unsupportedQueries: [
        'accelerate',
        'analytics',
        'inventory',
        'lifecycle',
        'logging',
        ...
    ],
```

electrachong: I don't want to mostly because it would mean the unsupportedQueries object would then be imported wherever else routeUtils is used, e.g. all the routes. It's not a big object but it is unnecessary given that the only place we use it so far is in routes.js. What do you think?

lib/routes.js
```diff
@@ -35,8 +36,9 @@ const statsClient = new StatsClient(localCacheClient, STATS_INTERVAL,
     STATS_EXPIRY);
 
 function checkUnsuportedRoutes(req, res, log) {
-    if (req.query.policy !== undefined ||
-        req.query.tagging !== undefined) {
+    const unsupportedRequestQuery = Object.keys(req.query)
```
rahulreddy: Since this would happen on every request, it would be better if we test the url vs looping through the properties of query object (as it is expensive). It's a common attack vector for public facing apps.
Here's what I would suggest
```js
const reqQueryStr = url.parse(req.url).query;
unsupportedQueries.some(q => reqQueryStr.includes(q));
```
Since we already parse the url once in https://github.com/scality/S3/blob/master/lib/utils.js#L252, to avoid parsing multiple times you could
```js
const parsedUrl = url.parse(request.url);
request.queryStr = parsedUrl.query;
request.query = querystring.parse(parsedUrl.query);
```

tests/functional/aws-node-sdk/lib/utility/makeRequest.js
```diff
@@ -0,0 +1,87 @@
...
+ * @param {string} params.objectKey - object key name
+ * @param {function} callback - with error and response parameters
+ * @return {undefined} - and call callback
+ */
+export default function makeRequest(params, callback) {
+    const { hostname, port, method, queryObj, headers, bucket, objectKey }
+        = params;
+    const options = {
+        hostname: hostname || ipAddress, // default to where S3 Server is hosted
+        port: port || 8000, // default to port 8000
```
rahulreddy: Ideally we should not write any fallbacks in tests. It would add some confusion.

tests/functional/aws-node-sdk/test/object/unsupportedQueries.js
```diff
@@ -0,0 +1,39 @@
+import assert from 'assert';
+import makeRequest from '../../lib/utility/makeRequest';
+
+// Queries supported by AWS that we do not currently support.
+const unsupportedQueries = [
```
rahulreddy: You should define this list somewhere, utils, for example and import it everywhere you need it. This would avoid changing things in two place when we have to update.

tests/functional/aws-node-sdk/lib/utility/makeRequest.js
```diff
@@ -0,0 +1,87 @@
+import http from 'http';
```
LaurenSpiegel: I like this utility but it's not really related to the aws-node-sdk.  Let's make this a separate directory called something like raw-node.  You will also have to create a new script for testing in the package.json, update circle.yml here and update Integration to call the new script.

tests/functional/aws-node-sdk/test/object/unsupportedQueries.js
LaurenSpiegel: agree. easy to import from routeUtils or this could be moved to constants.js and we can update it as we support more.

lib/routes/routesUtils.js
electrachong: I ended up moving this to constants.js per Lauren's suggestion

tests/functional/aws-node-sdk/lib/utility/makeRequest.js
electrachong: OK, I hope it makes sense I added a new function `makeS3Request` [here](https://github.com/scality/S3/pull/603/commits/9b0593f9f7c541414902cf4cf11c7315a97c1f54#diff-811e22af2f7d03b99d66636c51d1ad5e) since I think we will be making the majority of the requests against S3, and I don't think it makes sense to have to specify the hostname and port for it all the time. If we need to send to another host we can still access the base `makeRequest` function and specify the hostname.

tests/functional/aws-node-sdk/test/object/unsupportedQueries.js
electrachong: 👍 Yeah, that was the intent behind moving it to routeUtils, I forgot to modify the test file. I switched the location to constants.js though.

lib/routes.js
electrachong: do you think it makes sense to add this commit to avoid parsing the url again here, too?: 
https://github.com/scality/S3/pull/603/commits/87d91876ccb8809e08faa8cb519ee29df9b5a046

tests/functional/aws-node-sdk/lib/utility/makeRequest.js
electrachong: it looks like I might not need to update Integration, it just runs ft_test which runs all the tests in the S3 directory recursively? 
https://github.com/scality/Integration/blob/master/genericStaas/env/run_s3.bash
I'll run an end-to-end and see if my tests are run (after I fix some tests)

LaurenSpiegel: Excellent!  Even better!

lib/routes.js
rahulreddy: Yes, that's the duplication I was mentioning about. You can parse the url once in there and use it everywhere.

electrachong:
end-to-end: http://ci.ironmann.io/gh/scality/Integration/9473

electrachong:
new end-to-end: http://ci.ironmann.io/gh/scality/Integration/9476

electrachong:
another one: http://ci.ironmann.io/gh/scality/Integration/9478

electrachong:
end-to-end: http://ci.ironmann.io/gh/scality/Integration/9555

electrachong:
end-to-end and ci runs passed

electrachong:
@rahulreddy @nicolas2bert  when you have a moment let me know if your requested changes were addressed & I can squash my commits if no further changes are needed :)

lib/utils.js
```diff
@@ -304,4 +310,32 @@ utils.mapHeaders = function mapHeaders(headers, addHeaders) {
...
 };
 
+utils.isUnsupportedQuery = function isUnsupportedQuery(queryStr) {
+    if (!queryStr) {
+        return false;
+    }
+    // Reject unsupported query fields (e.g. ?accelerate), but allow
+    // the same strings as values (e.g. ?test=accelerate) and queries like
+    // '?acceleratefoobar' or '&foobaraccelerate'
+    return constants.unsupportedQueries.some(q => {
```
LaurenSpiegel: why not just use Object.keys(query) (using the parsed query object)?

tests/functional/s3cmd/tests.js
```diff
@@ -111,6 +111,27 @@ function checkRawOutput(args, lineFinder, testString, cb) {
     });
 }
 
+// Test stderr against expected output
+function checkErrOutput(args, lineFinder, testString, cb) {
```
LaurenSpiegel: instead of creating a new function, perhaps add an argument to the checkRawOutput function above that tells it to check stderr instead.

tests/functional/s3cmd/tests.js
```diff
@@ -283,7 +304,11 @@ describe('s3cmd put and get bucket ACLs', function aclBuck() {
         exec(['setacl', `s3://${bucket}`, '--acl-public'], done);
     });
 
-    it('should get canned ACL that was set', done => {
+    // TODO: Unskip when getBucketLocation has been implemented
+    // We currently return NotImplemented to ?location queries, so `s3cmd info`
+    // will not print info for canned ACL (but once we support those queries,
+    // this test should work again)
+    it.skip('should get canned ACL that was set', done => {
```
LaurenSpiegel: please make sure we have a functional test of this behavior at least with another client.

tests/functional/s3cmd/tests.js
```diff
@@ -296,7 +321,11 @@ describe('s3cmd put and get bucket ACLs', function aclBuck() {
         `--acl-grant=write:${emailAccount}`], done);
     });
 
-    it('should get specific ACL that was set', done => {
+    // TODO: Unskip when getBucketLocation has been implemented
+    // We currently return NotImplemented to ?location queries, so `s3cmd info`
+    // will not print info for specific ACL (but once we support those queries,
+    // this test should work again)
+    it.skip('should get specific ACL that was set', done => {
```
LaurenSpiegel: same

lib/routes.js
```diff
@@ -35,8 +35,8 @@ const statsClient = new StatsClient(localCacheClient, STATS_INTERVAL,
     STATS_EXPIRY);
 
 function checkUnsuportedRoutes(req, res, log) {
-    if (req.query.policy !== undefined ||
-        req.query.tagging !== undefined) {
+    if (utils.isUnsupportedQuery(req.queryStr)) {
```
LaurenSpiegel: will you handle the filter for the post routes that do not contain our two supported queries in a separate pr?  That's fine just want to make sure it gets picked up.

lib/utils.js
electrachong: Rahul's comment was that it's more expensive to iterate through the keys of an object than to check if a string contains the values, and it's an actual concern since we can receive many requests with thousands of queries.

tests/functional/s3cmd/tests.js
electrachong: OK, will do

lib/routes.js
electrachong: I looked at [routePOST](https://github.com/scality/S3/blob/master/lib/routes/routePOST.js) and saw that it effectively does this check already (if none of the three supported queries are defined, it returns NotImplemented). But do you think `routes.js` is a better place to filter out those requests?

LaurenSpiegel: If it's covered, that's fine

lib/utils.js
LaurenSpiegel: The query object would have a few items whereas the string itself would have more items and (I assume) you are doing a number of string loops here to check the beginning and end.  I don't know the current optimization state of Object.keys.  You could run a quick perf test on your machine to see the difference, if any.  @rahulreddy, thoughts?

rahulreddy: This is the reasoning I was going with.
constants.unsupportedQueries is definitive (as in we control it's length, hence we control the loop length).
`String. startsWith`, `String.charAt`, `String.endsWith` access the string at locations as if its an array, so that shouldn't be of any any concern. `String.includes`, this I am curious but I don't think it will be a huge deal, as it would be an equivalent of `indexOf`. So overall this approach should be ok IMO.

The other way, `Object.keys(query)`, now if we get a request with 400 query params in the url (for example), it would be an expensive operation to extract the keys and loop through them, we are giving control of the loop length to the user which is prone to attacks.

LaurenSpiegel: The query size is at least limited by node.  I think it's worth doing a couple tests considering this function will be run on all requests with queries.

electrachong: I don't mind running a few performance tests to get some experience with that since I haven't run performance tests before. As long as merging this PR is not urgent, which I don't think it is at the moment.

rahulreddy: Yes a perf test would be definitive. Btw HTTP_MAX_HEADER_SIZE is 80KB which apparently includes url, query and headers, beyond which the parser rejects the query. Is this the limit you were mentioning? It's way too big to not be protected against.

tests/functional/s3cmd/tests.js
electrachong: OK. We have some s3curl tests for putting and getting canned as well as specific ACLs:
https://github.com/scality/S3/blob/master/tests/functional/s3curl/tests.js#L285

tests/functional/s3cmd/tests.js
electrachong: https://github.com/scality/S3/blob/master/tests/functional/s3curl/tests.js#L285

lib/utils.js
electrachong: I extracted the functions and ran some timing tests locally to try and compare the performance between checking the querystr and the query object. It seems checking values for the query object is more performant.

with 1000 time measurements per median, using a querystr of 400 queries (e.g. `testquery0&testquery1&[...]`:
1. fxn using querystr took (medians) [ 127385, 127844, 125469, 125785, 125371 ] nanoseconds, with an average of 126370.8
2. fxn using queryobj took [ 64499, 75429, 68070, 77557, 74643 ] nanoseconds, with an average of 72039.6

with 100,000 time measurements per median:
1. fxn using querystr took [ 126757, 125313, 126057, 126333, 124928 ] nanoseconds, with an average of 125877.6
2. fxn using queryobj took [ 74821, 68282, 72476, 68157, 74749 ] nanoseconds, with an average of 71697

I also tried to compare the two codebases using the apache benchmarking tool against the server running the respective code, but testing with 10,000 requests (100 concurrent requests) and querystring of length 20, 100, and 400 didn't seem to reveal conclusive differences. Since more testing is required to get something conclusive with `ab`, we'll go with the results from extracting the functions. Query object wins :1st_place_medal:

rahulreddy: I concede :expressionless:.

lib/utils.js
```diff
@@ -304,4 +309,9 @@ utils.mapHeaders = function mapHeaders(headers, addHeaders) {
     return headers;
 };
 
+utils.isUnsupportedQuery = function isUnsupportedQuery(queryObj) {
```
LaurenSpiegel: add jsdoc please

tests/functional/raw-node/test/routes/unsupportedQuries.js
```diff
@@ -0,0 +1,34 @@
...
+            makeS3Request({ method: 'GET', queryObj, bucket, objectKey },
+            err => {
+                assert.strictEqual(err.code, 'NotImplemented');
+                assert.strictEqual(err.statusCode, 501);
+                done();
+            });
+        });
+    });
+
+    itSkipIfAWS('should accept blacklisted query field as a value ' +
```
LaurenSpiegel: should accept blacklisted query key as a query value to a query key that is not on the blacklist

electrachong:
@ironman-machine

tests/functional/s3cmd/configure.bash
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 apt-get install -y -q  python-dateutil python-magic
-wget http://launchpadlibrarian.net/222422124/s3cmd_1.6.0-2_all.deb
-dpkg -i s3cmd*.deb
+wget https://launchpad.net/ubuntu/+archive/primary/+files/s3cmd_1.6.1.orig.tar.gz
```
rahulreddy: ~~This can be updated to use the `pip install` command.~~
I think this file is legacy and can be safely removed. We have to do the same for other test dirs. as well (may/may not be in this PR).

electrachong:
@ironman-machine

electrachong:
@ironman-machine

tests/functional/s3cmd/tests.js
```diff
@@ -296,9 +310,13 @@ describe('s3cmd put and get bucket ACLs', function aclBuck() {
         `--acl-grant=write:${emailAccount}`], done);
     });
 
-    it('should get specific ACL that was set', done => {
+    // TODO: Unskip when getBucketLocation has been implemented
```
nicolas2bert: @LaurenSpiegel @electrachong Do you want to merge getBucketLocation implementation: https://github.com/scality/S3/pull/535 first?

electrachong:
end-to-end passed here: http://ci.ironmann.io/gh/scality/Integration/9789#build-parameters

electrachong:
@ironman-machine try
SCALITY_INTEGRATION_BRANCH=ft/update-s3cmd-ver

electrachong:
@ironman-machine try
SCALITY_INTEGRATION_BRANCH=ft/update-s3cmd-ver
CI_AGGRESSOR_FORCE_BUILD=1

electrachong:
end-to-end passed

PR #604, opened 2017-02-09
Forward rel/6.3 to rel/6.4

nicolas2bert:
Forwarding rel/6.3 to rel/6.4


nicolas2bert:
http://ci.ironmann.io/gh/scality/Integration/9471

PR #609, opened 2017-02-14
BF: removing .only in unit tests

nicolas2bert:
Removing `.only` in unit tests.


nicolas2bert:
E2E: http://ci.ironmann.io/gh/scality/Integration/9551

PR #611, opened 2017-02-17
FIX: Include part size sum in Utapi object

bennettbuchanan:
Get sum of part sizes from existing part objects to pass to utapiClient.

### DIFF
--- a/lib/api/multipartDelete.js
+++ b/lib/api/multipartDelete.js
@@ -107,6 +107,9 @@ function multipartDelete(authInfo, request, log, callback) {
             // Reconstruct mpuOverviewKey
             const mpuOverviewKey =
                 `overview${splitter}${objectKey}${splitter}${uploadId}`;
+            // Get the sum of all part sizes to include in pushMetric object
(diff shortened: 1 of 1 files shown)


lib/api/multipartDelete.js
```diff
@@ -117,6 +120,7 @@ function multipartDelete(authInfo, request, log, callback) {
                     pushMetric('abortMultipartUpload', log, {
                         authInfo,
                         bucket: bucketName,
+                        byteLength: partSizeSum,
```
LaurenSpiegel: Is this turned into a negative number somewhere?

bennettbuchanan: No. UtapiClient will use the value to decrement the `storageUtilized` metric. Similar to how DELETE operations are handled. 😄

LaurenSpiegel: Great.  Can you get this ready to merge?

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
e2e passed

PR #612, opened 2017-02-17
FT: Add UtapiReplay

bennettbuchanan:
* Add npm script to initiate UtapiReplay instance and start the replay job.
* Update Config.js to handle UtapiReplay configuration options.

Depends on https://github.com/scality/utapi/pull/109.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -141,8 +141,25 @@ class Config {
             }
         }
 
-        this.utapi = { component: 's3' };
+        if (process.env.ENABLE_LOCAL_CACHE) {
+            this.localCache = defaultLocalCache;
+        }
+        if (config.localCache) {
+            assert(typeof config.localCache === 'object',
+                'config: invalid local cache configuration. localCache must ' +
+                'be an object');
+            assert(typeof config.localCache.host === 'string',
+                'config: invalid host for localCache. host must be a string');
+            assert(typeof config.localCache.port === 'number',
+                'config: invalid port for localCache. port must be a number');
+            this.localCache = {
+                host: config.localCache.host,
+                port: config.localCache.port,
+            };
+        }
+
         if (config.utapi) {
+            this.utapi = { component: 's3' };
             if (config.utapi.port) {
                 assert(Number.isInteger(config.utapi.port)
                     && config.utapi.port > 0,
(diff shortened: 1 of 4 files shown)


lib/Config.js
```diff
@@ -141,6 +141,23 @@ class Config {
             }
         }
 
+        if (process.env.ENABLE_LOCAL_CACHE) {
```
bennettbuchanan: We move the section defining `this.localCache` since utapi uses the value.

lib/Config.js
```diff
@@ -183,13 +202,22 @@ class Config {
                     this.utapi.redis.host = config.utapi.redis.host;
                     this.utapi.redis.port = config.utapi.redis.port;
                 }
+            } else {
+                // `redis` is a required property of UtapiClient
+                config.utapi.redis = this.localCache;
```
rahulreddy: You should throw an assertion here, if `config.utapi` is set, it should have a redis section. It's a bad idea to use localcache for fallback.

lib/Config.js
```diff
@@ -155,41 +172,51 @@ class Config {
                     'bad config: utapi workers must be a positive integer');
                 this.utapi.workers = config.utapi.workers;
             }
-            if (config.utapi.redis) {
-                if (config.utapi.redis.sentinels) {
-                    this.utapi.redis = { sentinels: [], name: null };
+            // `localCache` is not part of the utapi configuration property.
+            assert(config.localCache, 'missing required property of utapi ' +
```
rahulreddy: Since we are expecting localCache from S3' config and not in the Utapi section
`config.localcache: missing required property of utapi: localCache` .
The comment above should be updated as `Utapi uses the same localCache config defined for S3 to avoid config duplication`

lib/Config.js
```diff
@@ -155,41 +172,51 @@ class Config {
...
+                this.utapi.redis.host = config.utapi.redis.host;
+                this.utapi.redis.port = config.utapi.redis.port;
             }
             if (config.utapi.metrics) {
                 this.utapi.metrics = config.utapi.metrics;
             }
             if (config.utapi.component) {
                 this.utapi.component = config.utapi.component;
             }
+            if (config.utapi.replaySchedule) {
```
rahulreddy: Add a comment here describing that the value is crontab style with an example. Also add `(optional)` to the comment suggesting that this is an optional config.

lib/Config.js
```diff
@@ -155,41 +172,51 @@ class Config {
...
+                this.utapi.replaySchedule = config.utapi.replaySchedule;
+            }
+            if (config.utapi.batchSize) {
+                this.utapi.batchSize = config.utapi.batchSize;
```
rahulreddy: Same here, add `(optional)` to the comment suggesting that this is an optional config. Provide an example value. Also, needs an assertion checking that the value is a positive integer > 0.

lib/utapi/utapiReplay.js
```diff
@@ -0,0 +1,12 @@
+const UtapiReplay = require('utapi').UtapiReplay;
+const _config = require('../Config').default;
+
+// start utapi server
+export default function main() {
+    let config;
```
rahulreddy: We no longer need this to be a conditional since config for utapi is either defined or not. If it's not defined and replay is started, Utapi will throw an assertion.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
e2e passed

PR #614, opened 2017-02-22
FIX: remove auth credential check from websiteHead/Get

### LINKED ISSUE #553: COMPAT: should not check vault credentials for get/head website requests
UPDATE: @nicolas2bert fixed the following behavior in https://github.com/scality/S3/issues/573 but we realized his refactor involved checking auth credentials, when we have not tested AWS behavior for website endpoint requests with credentials. Testing reveals we should not be making checking authentication credential and should only check if objects and buckets have public-* acls.

-----------------------------------------------

While testing cors headers responses for website endpoints, I found AWS responds with the following to requests for an object that does not exist (on bucket with public-read-write ACLs).

```
<html>
<head><title>404 Not Found</title></head>
<body>
<h1>404 Not Found</h1>
<ul>
<li>Code: NoSuchKey</li>
<li>Message: The specified key does not exist.</li>
<li>Key: test</li>
<li>RequestId: BC20E1D1388DFB25</li>
<li>HostId: NX+GRaKR27CCl/kiMa4iiF/KDApIrzT299WVHVgEhuG8o2X94FIfE6R/C+MFYSpiFvZRiufWxf8=</li>
</ul>
<hr/>
</body>
</html>
```

S3 current response:
```
<html><head><title>403 Forbidden</title></head><body><h1>403 Forbidden</h1><ul><li id='code'>Code: AccessDenied</li><li id='message'>Message: Access Denied</li><li>BucketName: bucketwebsitetester</li><li>RequestId: 8d95b84118e1695f47aa</li></ul><hr></body></html>
```

I remember we believed this was the appropriate response when testing/developing the website endpoint responses before, so will want to investigate the reason for that belief and if we have any other tests to corroborate a 403 AccessDenied response in certain scenarios (e.g. different ACLs).

electrachong:
FIXES #553

Also includes commit for refactoring website tests and clean-up in preparation for adding auth credentials to the website test requests. Zombie was a neat tool for visiting websites as if in a browser, but since it added unnecessary wait time to the redirect tests and we can check what we need in the response using the node `http`/`https` request module, I removed it.

### DIFF
--- a/tests/functional/aws-node-sdk/lib/utility/website-util.js
+++ b/tests/functional/aws-node-sdk/lib/utility/website-util.js
@@ -6,17 +6,72 @@ import url from 'url';
 
 import makeRequest from '../../../raw-node/utils/makeRequest';
 
-function _makeWebsiteRequest(urlstring, method, callback) {
+const bucketName = process.env.AWS_ON_AIR ? 'awsbucketwebsitetester' :
+    'bucketwebsitetester';
+let awsCredentials;
+
+function _parseConfigValue(string, fileSlice) {
+    const line = fileSlice.find(element => element.indexOf(string) > -1);
+    const keyValue = line.replace(/ /g, '').split('=');
+    // the element after the '=' should be the value
+    return keyValue[1];
+}
+
+function _retrieveAWSCredentials(profile) {
+    const filename = path.join(process.env.HOME, '/.aws/scality');
+    let file;
+
+    try {
+        file = fs.readFileSync(filename, 'utf8');
+    } catch (e) {
+        const msg = `AWS credential file does not exist: ${filename}`;
+        throw new Error(msg);
+    }
+
+    const fileContents = file.split('\n');
+    const profileIndex = file.indexOf(`[${profile}]`);
+    if (profileIndex > -1) {
+        const accessKey = _parseConfigValue('aws_access_key_id',
+            fileContents.slice(profileIndex));
+        const secretKey = _parseConfigValue('aws_secret_access_key',
+            fileContents.slice(profileIndex));
+        return { accessKey, secretKey };
+    }
+    const msg = `Profile ${profile} does not exist in AWS credential file`;
+    throw new Error(msg);
+}
+
+if (process.env.AWS_ON_AIR) {
+    awsCredentials = _retrieveAWSCredentials('default');
+}
+
+function _makeWebsiteRequest(auth, method, urlstring, callback) {
+    let authCredentials;
+    if (auth === 'valid credentials') {
+        authCredentials = awsCredentials || {
+            accessKey: 'accessKey1',
+            secretKey: 'verySecretKey1',
+        };
+    } else if (auth === 'invalid credentials') {
+        authCredentials = {
+            accessKey: 'fakeKey1',
+            secretKey: 'fakeSecretKey1',
+        };
+    } else if (auth !== undefined) {
+        throw new Error(`Unsupported auth type ${auth}`);
+    }
     const { hostname, port, path } = url.parse(urlstring);
-    makeRequest({ hostname, port, method, path }, callback);
+    makeRequest({ hostname, port, method, path, authCredentials }, callback);
 }
 
 function _assertResponseHtml(response, elemtag, content) {
     if (elemtag === 'ul') {
-        const startIndex = response.indexOf('<ul>');
+        const startingTag = '<ul>';
+        const startIndex = response.indexOf(startingTag);
         const endIndex = response.indexOf('</ul>');
         assert(startIndex > -1 && endIndex > -1, 'Did not find ul element');
-        const ulElem = response.slice(startIndex + 4, endIndex);
+        const ulElem = response.slice(startIndex + startingTag.length,
+            endIndex);
         content.forEach(item => {
             _assertResponseHtml(ulElem, 'li', item);
         });
@@ -32,14 +87,29 @@ function _assertContainsHtml(responseBody) {
         responseBody.includes('</html>'), 'Did not find html tags');
 }
 
-function _assertResponseHtml404(method, response, type, bucketName) {
+function _assertResponseHtml404(method, response, type) {
     assert.strictEqual(response.statusCode, 404);
     if (method === 'HEAD') {
-        if (type !== '404-no-such-bucket'
-        && type !== '404-no-such-website-configuration'
-        && type !== '404-not-found') {
-            throw new Error('This 404 error is not checked in ' +
-            'checkHTML()');
+        if (type === '404-no-such-bucket') {
+            assert.strictEqual(response.headers['x-amz-error-code'],
+            'NoSuchBucket');
+            // Need arsenal fixed to remove period at the end
+            // so compatible with aws
+            assert.strictEqual(response.headers['x-amz-error-message'],
+            'The specified bucket does not exist.');
+        } else if (type === '404-no-such-website-configuration') {
+            assert.strictEqual(response.headers['x-amz-error-code'],
+            'NoSuchWebsiteConfiguration');
+            assert.strictEqual(response.headers['x-amz-error-message'],
+            'The specified bucket does not have a website configuration');
+        } else if (type === '404-not-found') {
+            assert.strictEqual(response.headers['x-amz-error-code'],
+            'NoSuchKey');
+            assert.strictEqual(response.headers['x-amz-error-message'],
+            'The specified key does not exist.');
+        } else {
+            throw new Error(`'${type}' is not a recognized 404 ` +
+            'error checked in the WebsiteConfigTester.checkHTML function');
         }
         // don't need to check HTML for head requests
         return;
@@ -66,8 +136,8 @@ function _assertResponseHtml404(method, response, type, bucketName) {
             'Message: The specified key does not exist.',
         ]);
     } else {
-        throw new Error('This 404 error is not checked in ' +
-        'checkHTML()');
+        throw new Error(`'${type}' is not a recognized 404 ` +
+        'error checked in the WebsiteConfigTester.checkHTML function');
     }
 }
 
@@ -80,8 +150,8 @@ function _assertResponseHtml403(method, response, type) {
             assert.strictEqual(response.headers['x-amz-error-message'],
             'Access Denied');
         } else if (type !== '403-retrieve-error-document') {
-            throw new Error('This 403 error is not checked in ' +
-            'checkHTML()');
+            throw new Error(`'${type}' is not a recognized 403 ` +
+            'error checked in the WebsiteConfigTester.checkHTML function');
         }
     } else {
         _assertContainsHtml(response.body);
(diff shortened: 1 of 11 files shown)


tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -1,6 +1,206 @@
...
+            'website configuration',
+            `BucketName: ${bucketName}`,
+        ]);
+    } else if (type === '404-not-found') {
+        _assertResponseHtml(response.body, 'ul', [
+            'Code: NoSuchKey',
+            'Message: The specified key does not exist.',
+        ]);
+    } else {
+        throw new Error('This 404 error is not checked in ' +
```
LaurenSpiegel: would it be accurate to say 'This 404 error is not a recognized 404 error in the checkHTML function.'?

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -1,6 +1,206 @@
...
+
+function _assertResponseHtml404(method, response, type, bucketName) {
+    assert.strictEqual(response.statusCode, 404);
+    if (method === 'HEAD') {
+        if (type !== '404-no-such-bucket'
+        && type !== '404-no-such-website-configuration'
+        && type !== '404-not-found') {
+            throw new Error('This 404 error is not checked in ' +
+            'checkHTML()');
+        }
```
LaurenSpiegel: assert the error response headers

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -1,6 +1,206 @@
...
+        _assertResponseHtml(response.body, 'h1', '403 Forbidden');
+        _assertResponseHtml(response.body, 'ul', [
+            'Code: AccessDenied',
+            'Message: Access Denied',
+        ]);
+        if (type === '403-retrieve-error-document') {
+            _assertResponseHtml(response.body, 'h3',
+            'An Error Occurred While Attempting to ' +
+            'Retrieve a Custom Error Document');
+            const startIndex = response.body.indexOf('</h3>') + 5;
```
LaurenSpiegel: why the closing header tag plus 5?

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -36,91 +236,61 @@ export class WebsiteConfigTester {
...
-                    'Message: The specified bucket does not have a website ' +
-                    'configuration');
-                    browser.assert.text('ul:first-of-type > li:nth-child(3)',
-                    `BucketName: ${bucketName}`);
-                } else if (type === '404-not-found') {
-                    browser.assert.text('ul:first-of-type > li:first-child',
-                    'Code: NoSuchKey');
-                    browser.assert.text('ul:first-of-type > li:nth-child(2)',
-                    'Message: The specified key does not exist.');
+    static checkHTML(auth, method, url, type, redirectUrl, bucketName,
```
LaurenSpiegel: please add a jsdoc for this (especially to explain what type is)

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -36,91 +236,61 @@ export class WebsiteConfigTester {
...
+                        return this.checkHTML(null, method,
+                        res.headers.location, 'redirected-user', null, null,
+                        callback);
+                    }
+                } else if (type === 'index-user') {
+                    _assertResponseHtmlIndexUser(res);
                 } else {
-                    throw new Error('This 404 error is not checked in ' +
-                    'checkHTML()');
+                    throw new Error('This is not checked in checkHTML()');
```
LaurenSpiegel: This type is not recognized by the checkHTML function

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -36,91 +236,61 @@ export class WebsiteConfigTester {
...
+     * @param {function} cb - callback to end test
+     * @return {undefined}
+     */
+    static makeHeadRequest(auth, url, expectedStatusCode, expectedHeaders, cb) {
+        _makeWebsiteRequest(auth, 'HEAD', url, (err, res) => {
+            // body should be empty
+            assert.deepStrictEqual(res.body, '');
+            assert.strictEqual(res.statusCode, expectedStatusCode);
+            const headers = Object.keys(expectedHeaders);
+            for (let i = 0; i < headers.length; i++) {
```
LaurenSpiegel: are we guaranteed that the headers are always returned in the same order?

tests/functional/aws-node-sdk/test/object/websiteGet.js
```diff
@@ -504,20 +419,17 @@ describe('User visits bucket website endpoint', () => {
                 };
                 const redirect = {
                     HostName: 'www.google.com',
-                    ReplaceKeyPrefixWith: '/about',
+                    ReplaceKeyPrefixWith: 'about/',
```
LaurenSpiegel: is this change aws compatible?

tests/functional/aws-node-sdk/test/object/websiteGetWithACL.js
```diff
@@ -132,14 +129,30 @@ describe('User visits bucket website endpoint with ACL', () => {
...
+                            // Give some time for AWS to finish deleting
+                            // object and buckets before starting next test
+                            setTimeout(() => done(err), 10000);
+                        } else {
+                            done(err);
+                        }
+                    });
                 });
 
                 it(`${test.it}`, done => {
```
LaurenSpiegel: with no auth credentials sent?

tests/functional/aws-node-sdk/test/object/websiteHeadWithACL.js
```diff
@@ -190,8 +157,23 @@ describe('Head request on bucket website endpoint with ACL', () => {
 
                 it(`${test.it}`, done => {
```
LaurenSpiegel: with no credentials?

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -1,6 +1,206 @@
...
+            secretKey: 'verySecretKey1',
+        };
+    } else if (auth === 'invalidAuth') {
+        authCredentials = {
+            accessKey: 'fakeKey1',
+            secretKey: 'fakeSecretKey1',
+        };
+    } else if (!auth) {
+        authCredentials = undefined;
+    } else {
```
electrachong: only I don't want to throw err if auth is `null` :) only if auth is anything besides `null`, `invalidAuth` or `validAuth`

tests/functional/aws-node-sdk/lib/utility/website-util.js
electrachong: basically just asserting there is a `ul` element after the end of the `h3` element (for this error there are two ul elements, one before the `h3` and one after)
the plus 5 doesn't make much difference, it just means create the slice after the `</h3>` -- creating the slice starting from the `indexOf('</h3>')` would create the slice including `</h3>`

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -1,6 +1,206 @@
...
+    const { hostname, port, path } = url.parse(urlstring);
+    makeRequest({ hostname, port, method, path, authCredentials }, callback);
+}
+
+function _assertResponseHtml(response, elemtag, content) {
+    if (elemtag === 'ul') {
+        const startIndex = response.indexOf('<ul>');
+        const endIndex = response.indexOf('</ul>');
+        assert(startIndex > -1 && endIndex > -1, 'Did not find ul element');
+        const ulElem = response.slice(startIndex + 4, endIndex);
```
nicolas2bert: _Suggestion_: 
```
const startingTag = '<ul>'
const startingTagLength = startingTag.length
```
So, we understand where this "4" comes from and also if we change the tag we won't need to change its length.

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -1,6 +1,206 @@
...
+        _assertResponseHtml(response.body, 'h1', '403 Forbidden');
+        _assertResponseHtml(response.body, 'ul', [
+            'Code: AccessDenied',
+            'Message: Access Denied',
+        ]);
+        if (type === '403-retrieve-error-document') {
+            _assertResponseHtml(response.body, 'h3',
+            'An Error Occurred While Attempting to ' +
+            'Retrieve a Custom Error Document');
+            const startIndex = response.body.indexOf('</h3>') + 5;
```
nicolas2bert: Same suggestion about the length here.

tests/functional/aws-node-sdk/lib/utility/website-util.js
```diff
@@ -1,6 +1,206 @@
...
+
+function _assertContainsHtml(responseBody) {
+    assert(responseBody.startsWith('<html>') &&
+        responseBody.includes('</html>'), 'Did not find html tags');
+}
+
+function _assertResponseHtml404(method, response, type, bucketName) {
+    assert.strictEqual(response.statusCode, 404);
+    if (method === 'HEAD') {
+        if (type !== '404-no-such-bucket'
```
nicolas2bert: Why is this condition only inside the `if (method === 'HEAD')`?
Should not we check it for all the methods?

tests/functional/aws-node-sdk/lib/utility/website-util.js
electrachong: the function has been updated with new params and a jsdoc

tests/functional/aws-node-sdk/lib/utility/website-util.js
electrachong: this for loop is actually iterating through the header keys and comparing the value of header in result versus expected headers. I updated it to use `.forEach` since the for loop was a bit misleading.

tests/functional/aws-node-sdk/lib/utility/website-util.js
electrachong: updated to use your suggestion :)

tests/functional/aws-node-sdk/lib/utility/website-util.js
electrachong: I've updated the test helper function to use a `params` object

tests/functional/aws-node-sdk/lib/utility/website-util.js
electrachong: Sure, thank you, I updated it

tests/functional/aws-node-sdk/test/object/websiteGet.js
electrachong: yes -- think we had a typo before as it was redirecting to `http://google.com//about` (and it was passing the test because google then redirects to `https://www.google.com/about/`)

electrachong:
I guess I never ran an end-to-end but it might not work right now if I try :C
@ironman-machine

electrachong:
@ironman-machine

electrachong:
@ironman-machine

electrachong:
@ironman-machine

electrachong:
@ironman-machine

electrachong:
end-to-end passed, commit were squashed

PR #617, opened 2017-02-23
FIX: Request with hostname localhost returns InvalidURI

### LINKED ISSUE #616: Sending request with hostname localhost returns InvalidURI
Sending request with hostname `localhost` returns InvalidURI error response.

Should add `restEndpoints` (./config.json) `"localhost": "file"`

Why are we not removing `"127.0.0.1": "file"` in restEndpoints (./config.json) since we check for ip here: https://github.com/scality/S3/blob/master/lib/utils.js#L33?

Because we still want to use the endpoint 127.0.0.1 to determine the locationConstraint: https://github.com/scality/S3/blob/011b3f93ef97dae601dc441a1a03d021091989cc/lib/api/bucketPut.js#L53

nicolas2bert:
FIXES #616

### DIFF
--- a/config.json
+++ b/config.json
@@ -59,6 +59,7 @@
         }
     },
     "restEndpoints": {
+        "localhost": "file",
         "127.0.0.1": "file",
         "s3.docker.test": "scality-us-west-1",
         "127.0.0.2": "aws-us-east-1",


nicolas2bert:
E2E http://ci.ironmann.io/gh/scality/Integration/9941

PR #621, opened 2017-02-28
rf: metadata params for versioning S3C-44

electrachong:
This PR is in preparation for refactoring the S3 api methods for versioning.
It should be able to be merged immediately.

- reorder metadata wrapper params to precede callback
- pass callback params directly from bucketclient (when versioning is merged in MD, this will include versionId for affected methods)

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -323,7 +323,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 `overview${splitter}${objectKey}${splitter}${uploadId}`;
 
             return metadata.getObjectMD(mpuBucket.getName(), mpuOverviewKey,
-                log, (err, storedMetadata) => {
+                {}, log, (err, storedMetadata) => {
                     if (err) {
                         return next(err, destBucket);
                     }
--- a/lib/metadata/wrapper.js
+++ b/lib/metadata/wrapper.js
@@ -67,54 +67,61 @@ const metadata = {
         });
     },
 
-    putObjectMD: (bucketName, objName, objVal, log, cb, params) => {
+    putObjectMD: (bucketName, objName, objVal, params, log, cb) => {
         log.debug('putting object in metdata');
-        client.putObject(bucketName, objName, objVal, log, err => {
+        client.putObject(bucketName, objName, objVal, params, log,
+        (err, data) => {
             if (err) {
                 log.debug('error from metadata', { implName, error: err });
                 return cb(err);
             }
-            log.debug('object successfully put in metadata');
-            return cb(err);
-        }, params);
+            if (data) {
+                log.debug('object version successfully put in metadata',
+                { version: data });
+            } else {
+                log.debug('object successfully put in metadata');
+            }
+            return cb(err, data);
+        });
     },
 
-    getBucketAndObjectMD: (bucketName, objName, log, cb, params) => {
+    getBucketAndObjectMD: (bucketName, objName, params, log, cb) => {
         log.debug('getting bucket and object from metadata',
                   { database: bucketName, object: objName });
-        client.getBucketAndObject(bucketName, objName, log, (err, data) => {
-            if (err) {
-                log.debug('error from metadata', { implName, err });
-                return cb(err);
-            }
-            log.debug('bucket and object retrieved from metadata',
-                      { database: bucketName, object: objName });
-            return cb(err, data);
-        }, params);
+        client.getBucketAndObject(bucketName, objName, params, log,
+            (err, data) => {
+                if (err) {
+                    log.debug('error from metadata', { implName, err });
+                    return cb(err);
+                }
+                log.debug('bucket and object retrieved from metadata',
+                { database: bucketName, object: objName });
+                return cb(err, data);
+            });
     },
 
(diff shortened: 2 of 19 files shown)


electrachong:
@ironman-machine

ttvi:
Please also remove `this.extendedVersion`, `this.metadataVersion`, and `METADATA_DEFAULT_VERSION` in `/bucketclient/backend.js`.

electrachong:
@ttvi I removed them, but this implies we don't have any concerns for backwards compatibility, right?

ttvi:
We don't have any compatibility issue.

lib/api/completeMultipartUpload.js
```diff
@@ -323,10 +323,12 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 `overview${splitter}${objectKey}${splitter}${uploadId}`;
 
             return metadata.getObjectMD(mpuBucket.getName(), mpuOverviewKey,
-                log, (err, storedMetadata) => {
+                {}, log, (err, storedMetadata) => {
                     if (err) {
+                        log.warn('error from metadata', { err });
                         return next(err, destBucket);
                     }
+                    log.debug('object retrieved from metadata');
```
LaurenSpiegel: mpu overview object retrieved from metadata

lib/metadata/wrapper.js
```diff
@@ -67,54 +67,25 @@ const metadata = {
         });
     },
 
-    putObjectMD: (bucketName, objName, objVal, log, cb, params) => {
+    putObjectMD: (bucketName, objName, objVal, params, log, cb) => {
         log.debug('putting object in metdata');
-        client.putObject(bucketName, objName, objVal, log, err => {
-            if (err) {
```
LaurenSpiegel: I would prefer keeping the error logs in the wrapper so we are certain we are logging them.

lib/metadata/wrapper.js
```diff
@@ -67,54 +67,25 @@ const metadata = {
...
+        client.putObject(bucketName, objName, objVal, params, log, cb);
     },
 
-    getBucketAndObjectMD: (bucketName, objName, log, cb, params) => {
+    getBucketAndObjectMD: (bucketName, objName, params, log, cb) => {
         log.debug('getting bucket and object from metadata',
                   { database: bucketName, object: objName });
-        client.getBucketAndObject(bucketName, objName, log, (err, data) => {
-            if (err) {
-                log.debug('error from metadata', { implName, err });
```
LaurenSpiegel: same

lib/metadata/wrapper.js
```diff
@@ -67,54 +67,25 @@ const metadata = {
...
-            return cb(err, data);
-        }, params);
+        client.getBucketAndObject(bucketName, objName, params, log, cb);
     },
 
-    getObjectMD: (bucketName, objName, log, cb, params) => {
+    getObjectMD: (bucketName, objName, params, log, cb) => {
         log.debug('getting object from metadata');
-        client.getObject(bucketName, objName, log, (err, data) => {
-            if (err) {
```
LaurenSpiegel: same

lib/metadata/wrapper.js
```diff
@@ -67,54 +67,25 @@ const metadata = {
...
-        }, params);
+        client.getObject(bucketName, objName, params, log, cb);
     },
 
-    deleteObjectMD: (bucketName, objName, log, cb, params) => {
+    deleteObjectMD: (bucketName, objName, params, log, cb) => {
         log.debug('deleting object from metadata');
-        client.deleteObject(bucketName, objName, log, err => {
-            if (err) {
-                log.warn('error from metadata', { implName, err });
```
LaurenSpiegel: same

lib/metadata/wrapper.js
ttvi: By using the callback `cb` directly we can extend the number of parameters (not limited to `err` and `data`) in the future when MetaData will not inject additional "metadata" (such as `versionId`) directly into the value but return it using another channel.
Also I think it's a good idea not keeping the logging here because, first, this has been logged at bucketclient, and second, different operations may interpret the same error differently, such as `ObjNotFound` for `objectGet` and `objectPut`; these operations should have their own logging.

electrachong: These are good arguments, but :
- if we log here, we can be sure all errors are logged. Although the errors should be logged by the calling operation,  it's an easy mistake to omit the logging at that level.
- Although bucketclient logs all the errors, it's helpful from a debugging perspective if these errors are transparent when viewed from S3 logs. That's why it can be a pain when the logs are omitted by accident.
-  OK to have multiple layers of logging.

I don't have too strong an opinion on this one, but it could be worth the trouble to have to update the callback parameters when they are extended.

LaurenSpiegel: We're downgrading the warn log level to debug so there won't be confusing error logs in production at this level.  

I agree with @electrachong and think we should be mindful of what arguments we are getting back from bucketclient metadata (and the other metadata backends should provide the same arguments) so updating the arguments here would be good so we are clear on what is expected from any metadata backend.

electrachong:
@ironman-machine

electrachong:
@ironman-machine

electrachong:
end-to-end passed

PR #623, opened 2017-03-01
FT: Set one pair of accessKey/secretKey with env variables

nicolas2bert:


lib/auth/in_memory/builder.js
```diff
@@ -0,0 +1,21 @@
+/** build simple authdata with only one account
+ * @param {string} accessKey - account's accessKey
+ * @param {string} secretKey - account's secretKey
+ * @return {object} authdata - authdata with account's accessKey and secretKey
+ */
+export function buildAuthDataAccount(accessKey, secretKey) {
+    return {
+        accounts: [{
+            name: 'CustomAccount',
+            email: 'customaccount1@sampling.com',
```
LaurenSpiegel: maybe call it 'customaccount1@setbyenv.com'

lib/Config.js
```diff
@@ -435,7 +436,11 @@ class Config {
             if (process.env.S3AUTH_CONFIG) {
                 authfile = process.env.S3AUTH_CONFIG;
             }
-            const authData = require(authfile);
+            let authData = require(authfile);
+            if (process.env.ACCESS_KEY && process.env.SECRET_KEY) {
```
rahulreddy: Suggestion, why not use `SCALITY_ACCESS_KEY_ID` and `SCALITY_SECRET_ACCESS_KEY` which would be similar to `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` used by AWS. When we have multiple backends we can use both sets to work with AWS/SCALITY.

nicolas2bert: Makes sense! @rahulreddy

DOCKER.md
```diff
@@ -25,19 +25,19 @@ you and your users can direct s3 server requests to example.com.
...
+#### SCALITY_ACCESS_KEY_ID and SCALITY_SECRET_ACCESS_KEY
 
 These variables specify authentication credentials for an account
-named "Docker".
+named "CustomAccount".
 
 You can set credentials for many accounts by editing `conf/authdata.json`
 (see below for further info),
 but if you just want to specify one set of your own,
 you can use these environment variables.
```
LaurenSpiegel: Please add here as well -- Note that if you set these environment variables, the conf/authdata.json will be ignored.

README.md
```diff
@@ -71,6 +71,22 @@ This starts an S3 server on port 8000.
 The default access key is accessKey1 with
 a secret key of verySecretKey1.
 
+## Run it with environment variables
```
LaurenSpiegel: I would do this as a separate section (not a run it) with a heading such as "Setting your own access key and secret key pairs"

Then I would say there are 2 options for setting your own, the env variables or the authdata.json file.

DOCKER.md
```diff
@@ -25,19 +25,19 @@ you and your users can direct s3 server requests to example.com.
 docker run -d --name s3server -p 8000:8000 -e HOST_NAME=new.host.com scality/s3server
 ```
 
-#### ACCESS_KEY and SECRET_KEY
```
rahulreddy: You can add a note at the bottom that the old `ACCESS_KEY` and `SECRET_KEY` environment variables are now deprecated.

nicolas2bert: Good idea - Done :)

DOCKER.md
nicolas2bert: Done :)

nicolas2bert:
@ironman-machine

nicolas2bert:
E2E passed http://ci.ironmann.io/gh/scality/Integration/10174#build-parameters

README.md
```diff
@@ -71,6 +71,23 @@ This starts an S3 server on port 8000.
 The default access key is accessKey1 with
 a secret key of verySecretKey1.
 
+## Setting your own access key and secret key pairs
+
+You can set credentials for many accounts by editing `conf/authdata.json` but if
+you want to specify one set of your own credential, you can use
```
bennettbuchanan: ~credential~ credentials

lib/Config.js
```diff
@@ -469,7 +470,13 @@ class Config {
             if (process.env.S3AUTH_CONFIG) {
                 authfile = process.env.S3AUTH_CONFIG;
             }
-            const authData = require(authfile);
+            let authData = require(authfile);
```
bennettbuchanan: If `authfile` is invalid JSON, this would throw an error. Maybe it's okay, but it could be helpful to check that first.

nicolas2bert: I guess we are ok throwing an error if `authfile` is invalid JSON, @LaurenSpiegel can confirm.

LaurenSpiegel: yes, let them know early that there is a problem

PR #624, opened 2017-03-02
Change warn logs to debug logs

LaurenSpiegel:


lib/kms/common.js
```diff
@@ -124,7 +124,7 @@ export default class Common {
             cryptoScheme, dataKey, log,
             (err, derivedKey, derivedIV) => {
                 if (err) {
-                    log.warn('key derivation failed', { error: err });
+                    log.debug('key derivation failed', { error: err });
```
ploki: Errors are not expected from the kms during normal operation. No data from the user may trigger an error at this stage.
Since the purpose is to get rid of overzealous warning. I think these messages don't fall in this category.
On the other hand, the callee function has already logged a message and the lower level interface is the exact replicate of this one. So it's safe to proceed this way. (this statement only applies to this file)

LaurenSpiegel:
@ironman-machine go

PR #626, opened 2017-03-03
Fix logic in parsing of sproxyd config and update tests

dora-korpar:
Previously, the parsing logic assumed that sproxyd would have either a bootstrap list or a chordCos value. This PR allows for both to be used

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -10,7 +10,7 @@ const defaultHealthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
 const defaultLocalCache = { host: '127.0.0.1', port: 6379 };
 
 export function sproxydAssert(configSproxyd) {
-    let sproxydField;
+    const sproxydFields = [];
     if (configSproxyd.bootstrap !== undefined) {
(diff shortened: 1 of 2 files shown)


dora-korpar:
@ironman-machine gogogo

lib/Config.js
```diff
@@ -141,17 +141,20 @@ class Config {
...
-                                    connector.sproxyd[field], 16);
-                        } else {
+                                .connector.sproxyd.bootstrap =
+                                lc[l].details.connector.sproxyd.bootstrap;
+                        }
+                        if (fields.indexOf('chordCos') > -1) {
                             this.locationConstraints[l].details
-                                .connector.sproxyd[field] =
-                                lc[l].details.connector.sproxyd[field];
+                                .connector.sproxyd[fields] =
```
jonathan-gramain: Is this right? I would expect instead `.connector.sproxyd.chordCos`. From my tests, that works if the array has only one element, but does not as soon as it has 2+.

dora-korpar: Interesting, it seems to work either way for me, but I agree it makes more sense to specify `.chordCos`. Thanks for pointing it out!

PR #627, opened 2017-03-03
S3C-92 GET wont report the size in the correct field

nicolas2bert:
- `bodyLength: req.parsedContentLength || 0,` because for streaming v4 auth, the total body content length without the chunk metadata is sent as the `x-amz-decoded-content-length` 
- Getting rid of `contentLength: request.parsedContentLength,` in routePut.js since contentLength === bodyLength === req.parsedContentLength

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -80,8 +80,8 @@ export default function routes(req, res, logger) {
     log.addDefaultFields({
         bucketName: req.bucketName,
         objectKey: req.objectKey,
-        bytesReceived: req.parsedContentLength || 0,
-        bodyLength: parseInt(req.headers['content-length'], 10) || 0,
+        bytesReceived: req.socket.bytesRead || 0,
+        bytesSent: req.socket.bytesWritten || 0,
     });
     // if empty name and request not a list Buckets
     if (!req.bucketName &&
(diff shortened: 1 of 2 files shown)


lib/routes.js
```diff
@@ -80,8 +80,9 @@ export default function routes(req, res, logger) {
     log.addDefaultFields({
         bucketName: req.bucketName,
         objectKey: req.objectKey,
-        bytesReceived: req.parsedContentLength || 0,
-        bodyLength: parseInt(req.headers['content-length'], 10) || 0,
+        bytesReceived: req.socket.bytesWritten || 0,
```
rahulreddy: This should be other way around 
```
 bytesReceived: req.socket.bytesRead || 0
```
```
bytesSent: req.socket.bytesWritten || 0
```

LaurenSpiegel:
I just tested and req.socket.bytesRead with keep alive will aggregate a number of requests which is not what we want. It does not look like there is an easy way to do this in node.  We need to reconsider whether to continue with this.

rahulreddy:
There is no easy way to do this. Even if we calculate the headers manually, we will be off by a few bytes due to random stuff like incorrect number of new line characters etc. I don't see a node.js solution for this.

PR #629, opened 2017-03-03
FT: Separate location config

nicolas2bert:


lib/Config.js
```diff
@@ -71,19 +71,78 @@ class Config {
...
 
+    _getLocationConfig() {
+        let locationConfig;
+        // check exist file
+        try {
+            fs.statSync(this.locationConfigPath);
+        } catch (err) {
+            if (err) {
+                // location file does not exist
+                // old config
```
LaurenSpiegel: instead of 'old config', I would say 'legacy config format where only one config file'

lib/Config.js
```diff
@@ -71,19 +71,78 @@ class Config {
...
+              { encoding: 'utf-8' });
+            locationConfig = JSON.parse(data);
+        } catch (err) {
+            throw new Error(`could not parse location config file:
+              ${err.message}`);
+        }
+
+        if (locationConfig !== undefined) {
+            this.locationConstraints = {};
+            const lc = locationConfig;
```
LaurenSpiegel: why redefine it as lc?

lib/Config.js
```diff
@@ -488,7 +517,6 @@ class Config {
             dataPath,
             metadataPath,
         };
-        return config;
```
LaurenSpiegel: why no longer return?

nicolas2bert: I had no idea why we did return something here. 
I removed it because of a lint return issue.

lib/Config.js
```diff
@@ -71,19 +71,78 @@ class Config {
          * It can be overridden using the S3_CONFIG_FILE environment var.
```
LaurenSpiegel: let's expand this comment due to the additional file

lib/Config.js
LaurenSpiegel: ok

lib/Config.js
```diff
@@ -69,21 +69,85 @@ class Config {
...
+        try {
+            const data = fs.readFileSync(this.locationConfigPath,
+              { encoding: 'utf-8' });
+            locationConfig = JSON.parse(data);
+        } catch (err) {
+            throw new Error(`could not parse location config file:
+              ${err.message}`);
+        }
+
+        if (locationConfig !== undefined) {
```
LaurenSpiegel: If it's the legacy set up, there wouldn't be a separate file so this wouldn't get defined.

lib/Config.js
```diff
@@ -69,21 +69,83 @@ class Config {
...
+        try {
+            const data = fs.readFileSync(this.locationConfigPath,
+              { encoding: 'utf-8' });
+            locationConfig = JSON.parse(data);
+        } catch (err) {
+            throw new Error(`could not parse location config file:
+              ${err.message}`);
+        }
+
+        this.locationConstraints = {};
```
nicolas2bert: @LaurenSpiegel @bennettbuchanan  Checking if locationConfig is not needed here because 
- if location config path is not valid => return https://github.com/scality/S3/pull/629/files#diff-9280903abc69f09eff9ab36070b6d751R101
- if can not parse it => throw error

lib/Config.js
bennettbuchanan: I see, I was thinking this would be caught by the `try...catch`.

nicolas2bert:
@ironman-machine ok

nicolas2bert:
@ironman-machine try
SCALITY_FEDERATION_BRANCH=ft/separateconfig

nicolas2bert:
@ironman-machine try
SCALITY_FEDERATION_BRANCH=ft/separateconfig

nicolas2bert:
@ironman-machine ok

nicolas2bert:
@ironman-machine should work

nicolas2bert:
Changes:

- I replaced `const locationConstraint = config.locationConstraints ? 'aws-us-east-1' :
'us-east-1';` to `const locationConstraint = 'us-east-1';` since `us-east-1` is in the locationConfig.json

- For the skip config, instead of checking if `conf.locationConstraints` is defined we check for `conf.regions` : `const describeSkipIfOldConfig = conf.regions ? describe.skip : describe;`
Indeed, if the region config section is defined it will mean that the old config has been used.

nicolas2bert:
@ironman-machine last one after squash

PR #631, opened 2017-03-04
FT: Support Expect header field

### LINKED ISSUE #510: 100 Continue Handling per RFC

bennettbuchanan:
Add a listener for the 'checkContinue' event. 

Send a `HTTP/1.1 100 Continue` message to the client if requester is

- from a public user
- request is V4 streaming authentication
- when a V2/V4(non-streaming) request is authenticated by Arsenal.

Fix #510

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -1,5 +1,4 @@
 import querystring from 'querystring';
-
 import { auth, errors } from 'arsenal';
 
 import bucketDelete from './bucketDelete';
@@ -42,10 +41,10 @@ import websiteHead from './websiteHead';
 
 auth.setHandler(vault);
 
+/* eslint-disable no-param-reassign */
 const api = {
-    callApiMethod(apiMethod, request, log, callback, locationConstraint) {
-        // no need to check auth on website serving (head or get)
-        // or cors preflight requests
+    callApiMethod(apiMethod, request, response, log, callback) {
+        // no need to check auth on website or cors preflight requests
         if (apiMethod === 'websiteGet' || apiMethod === 'websiteHead' ||
         apiMethod === 'corsPreflight') {
             return this[apiMethod](request, log, callback);
@@ -68,8 +67,8 @@ const api = {
             sourceBucket = source.slice(0, slashSeparator);
             sourceObject = source.slice(slashSeparator + 1);
         }
-        const requestContexts = prepareRequestContexts(apiMethod,
-            request, locationConstraint, sourceBucket, sourceObject);
+        const requestContexts = prepareRequestContexts(apiMethod, request,
+            sourceBucket, sourceObject);
         return auth.server.doAuth(request, log, (err, userInfo,
             authorizationResults, streamingV4Params) => {
             if (err) {
@@ -84,20 +83,47 @@ const api = {
                     }
                 }
             }
-            if (apiMethod === 'bucketPut') {
-                return bucketPut(userInfo, request, locationConstraint,
-                    log, callback);
-            }
-            if (apiMethod === 'objectCopy' ||
-                apiMethod === 'objectPutCopyPart') {
-                return this[apiMethod](userInfo, request, sourceBucket,
-                    sourceObject, log, callback);
-            }
             if (apiMethod === 'objectPut' || apiMethod === 'objectPutPart') {
                 return this[apiMethod](userInfo, request, streamingV4Params,
                     log, callback);
             }
-            return this[apiMethod](userInfo, request, log, callback);
+            const MAX_POST_LENGTH = request.method.toUpperCase() === 'POST' ?
+                1024 * 1024 : 1024 * 1024 / 2; // 1 MB or 512 KB
+            const post = [];
+            let postLength = 0;
+            request.on('data', chunk => {
+                postLength += chunk.length;
+                // Sanity check on post length
+                if (postLength <= MAX_POST_LENGTH) {
+                    post.push(chunk);
+                }
+                return undefined;
+            });
+
+            request.on('error', err => {
+                log.trace('error receiving request', {
+                    error: err,
+                });
+                return callback(errors.InternalError);
+            });
+
+            request.on('end', () => {
+                if (postLength > MAX_POST_LENGTH) {
+                    log.error('body length is too long for request type',
+                        { postLength });
+                    return callback(errors.InvalidRequest);
+                }
+                // Convert array of post buffers into one string
+                request.post = Buffer.concat(post, postLength).toString();
+
+                if (apiMethod === 'objectCopy' ||
+                    apiMethod === 'objectPutCopyPart') {
+                    return this[apiMethod](userInfo, request, sourceBucket,
+                        sourceObject, log, callback);
+                }
+                return this[apiMethod](userInfo, request, log, callback);
+            });
+            return undefined;
         }, 's3', requestContexts);
     },
     bucketDelete,
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -38,6 +38,7 @@ import serviceGet from './serviceGet';
 import vault from '../auth/vault';
 import websiteGet from './websiteGet';
 import websiteHead from './websiteHead';
+import writeContinue from '../utilities/writeContinue';
 
 auth.setHandler(vault);
 
@@ -83,6 +84,8 @@ const api = {
                     }
                 }
             }
+            // issue 100 Continue to the client
+            writeContinue(request, response);
             if (apiMethod === 'objectPut' || apiMethod === 'objectPutPart') {
                 return this[apiMethod](userInfo, request, streamingV4Params,
                     log, callback);
--- a/lib/api/apiUtils/authorization/prepareRequestContexts.js
+++ b/lib/api/apiUtils/authorization/prepareRequestContexts.js
@@ -6,19 +6,22 @@ const RequestContext = policies.RequestContext;
  * Prepares the requestContexts array to send to Vault for authorization
  * @param {string} apiMethod - api being called
  * @param {object} request - request object
- * @param {string} locationConstraint - locationConstraint if bucket put
- * operation
  * @param {string} sourceBucket - name of sourceBucket if copy request
  * @param {string} sourceObject - name of sourceObject if copy request
  * @return {RequestContext []} array of requestContexts
  */
-export default function prepareRequestContexts(apiMethod, request,
-    locationConstraint, sourceBucket, sourceObject) {
+export default function prepareRequestContexts(apiMethod, request, sourceBucket,
+    sourceObject) {
     // if multiObjectDelete request, we want to authenticate
     // before parsing the post body and creating multiple requestContexts
     // so send null as requestContexts to Vault to avoid authorization
     // checks at this point
-    if (apiMethod === 'multiObjectDelete') {
+    //
+    // If bucketPut request, we want to do the authorization check in the API
+    // itself (once we parse the locationConstraint from the xml body) so send
+    // null as the requestContext to Vault so it will only do an authentication
+    // check.
+    if (apiMethod === 'multiObjectDelete' || apiMethod === 'bucketPut') {
         return null;
     }
     const requestContexts = [];
@@ -26,17 +29,17 @@ export default function prepareRequestContexts(apiMethod, request,
         const getRequestContext = new RequestContext(request.headers,
             request.query, sourceBucket, sourceObject,
             request.socket.remoteAddress, request.connection.encrypted,
-            'objectGet', 's3', locationConstraint);
+            'objectGet', 's3');
         const putRequestContext = new RequestContext(request.headers,
             request.query, request.bucketName, request.objectKey,
             request.socket.remoteAddress, request.connection.encrypted,
-            'objectPut', 's3', locationConstraint);
+            'objectPut', 's3');
         requestContexts.push(getRequestContext, putRequestContext);
     } else {
         const requestContext = new RequestContext(request.headers,
             request.query, request.bucketName, request.objectKey,
             request.socket.remoteAddress, request.connection.encrypted,
-            apiMethod, 's3', locationConstraint);
+            apiMethod, 's3');
         requestContexts.push(requestContext);
     }
     return requestContexts;
--- a/lib/api/bucketPut.js
+++ b/lib/api/bucketPut.js
@@ -1,12 +1,12 @@
-import { errors } from 'arsenal';
-
+import { auth, errors } from 'arsenal';
+import vault from '../auth/vault';
+import { parseString } from 'xml2js';
+import { waterfall } from 'async';
 import { createBucket } from './apiUtils/bucket/bucketCreation';
 import collectCorsHeaders from '../utilities/collectCorsHeaders';
 import config from '../Config';
 import aclUtils from '../utilities/aclUtils';
 import { pushMetric } from '../utapi/utilities';
-
-let locationConstraintChecked;
 
 /*
    Format of xml request:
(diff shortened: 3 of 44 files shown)


lib/utilities/writeContinue.js
```diff
@@ -0,0 +1,32 @@
...
+    if (authHeader) {
+        authV2 = authHeader.startsWith('AWS ');
+        authV4 = authHeader.startsWith('AWS4');
+    }
+    const isStreaming = headers['content-type'] === 'application/octet-stream';
+    // The Expect field-value is case-insensitive. Thus, just check for the
+    // Expect header's presence and a valid field-value.
+    if (headers.expect && headers.expect.toLowerCase() === '100-continue') {
+        if (authInfo) {
+            if (authInfo.isRequesterPublicUser()) {
```
rahulreddy: Minor: You can leverage `return undefined;` instead of if...else. It would look more readable.

tests/functional/aws-node-sdk/test/object/100-continue.js
```diff
@@ -0,0 +1,123 @@
...
+            req.end(body);
+        }
+    }
+
+    sendsBodyOnContinue(cb) {
+        const options = this.getRequestOptions();
+        const req = http.request(options);
+        // At this point we have only sent the header.
+        assert(req.output.length === 1);
+        const headerLen = req.output[0].length;
```
rahulreddy: Using request object internals is a bad idea as it is prone to changes. Can you explore other options?

tests/functional/aws-node-sdk/test/object/100-continue.js
```diff
@@ -0,0 +1,123 @@
...
+        beforeEach(() => {
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
+            const params = {
+                Bucket: bucket,
+                Key: key,
+            };
+            const signedUrl = s3.getSignedUrl('putObject', params);
+            const { path } = url.parse(signedUrl);
+            ContinueRequest = new ContinueRequestHandler(path);
```
rahulreddy: Class instantiations should have first char as lowercase `continueRequest`. Usually linter would throw an error for this.

lib/routes/routePUT.js
```diff
@@ -38,6 +39,8 @@ export default function routePUT(request, response, log, statsClient) {
         request.post = '';
         const post = [];
         let postLength = 0;
+        // Send continue message for streaming V4 requests, if applicable.
+        writeContinue(request, response);
```
nicolas2bert: _Suggestion_: If I got it right - why not removing `writeContinue(...)` here and add `putBucket` and `putBucketACL` to the list `lib/utilities/writeContinue.js`? It might be clearer.

lib/routes/routePOST.js
```diff
@@ -2,13 +2,16 @@ import { errors } from 'arsenal';
...
+import writeContinue from '../utilities/writeContinue';
 
 const MAX_POST_LENGTH = 1024 * 1024; // 1 MB
 
 /* eslint-disable no-param-reassign */
 export default function routePOST(request, response, log) {
     log.debug('routing request', { method: 'routePOST' });
 
+    // Send continue message, if applicable, for streaming V4 requests.
+    writeContinue(request, response);
```
nicolas2bert: _Same suggestion_ here with `initiateMultipartUpload`, `completeMultipartUpload` and `multiObjectDelete`

lib/routes/routePUT.js
bennettbuchanan: If the `writeContinue` method is not called at this point, the check for `MAX_POST_LENGTH` will never be made since the body will not be sent.

https://github.com/scality/S3/blob/6a3115b9ae52238b77cfe5dbfac0a8e393336c2f/lib/routes/routePUT.js#L54

lib/routes/routePOST.js
bennettbuchanan: Same as above, but for `POST` operations.

lib/api/api.js
```diff
@@ -82,6 +84,9 @@ const api = {
                     }
                 }
             }
+            // issue 100 Continue to the client
+            writeContinue(request, response, apiMethod);
+
```
LaurenSpiegel: I think you need a writeContinue call at around line 50 for the apiMethods that don't require a check with vault.

lib/routes/routePOST.js
LaurenSpiegel: There would only be a streaming v4 auth request on a put object or put object part.  Do we need this here?

lib/routes/routePUT.js
LaurenSpiegel: If this is only for streaming v4, can you put it in the route if condition for just when there is an object key (so put object) or put object part?

lib/utilities/writeContinue.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {undefined}
+ */
+export default function writeContinue(req, res, apiMethod) {
+    const { headers } = req;
+    // The Expect field-value is case-insensitive. Thus, just check for the
+    // Expect header's presence and a valid field-value.
+    if (headers.expect && headers.expect.toLowerCase() === '100-continue') {
+        if (apiMethod) {
+            if (apiMethod === 'objectCopy' ||
+                apiMethod === 'objectPutCopyPart' ||
```
LaurenSpiegel: why are objectCopy and objectPutCopyPart included here?  Is this meant for the streaming v4 cases?  You would only have streaming v4 on objectPut and objectPutPart.  There is no body in copy requests.

tests/functional/aws-node-sdk/test/object/100-continue.js
```diff
@@ -0,0 +1,123 @@
...
+            const expected = body.length + headerLen;
+            // Has the entire body been sent?
+            assert.strictEqual(req.socket.bytesWritten, expected);
+            return cb();
+        });
+        req.on('error', err => cb(err));
+    }
+}
+
+describe('PUT public object with 100-continue header', () => {
```
LaurenSpiegel: It would be good to test the streaming v4 cases to make sure it works.  You can mess with either the java or the fog tests to add a 100-continue header and make sure the tests still pass.

lib/routes/routePUT.js
nicolas2bert: Got it! Makes sense, thanks

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine 
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
E2E passed

lib/api/api.js
bennettbuchanan: As discussed, this only applies to POST and POST requests.

lib/utilities/writeContinue.js
bennettbuchanan: This method has been updated.

tests/functional/aws-node-sdk/test/object/100-continue.js
bennettbuchanan: Added the Expect 100-continue header in the java putObject tests.

tests/functional/aws-node-sdk/test/object/100-continue.js
bennettbuchanan: Updated.

tests/functional/aws-node-sdk/test/object/100-continue.js
bennettbuchanan: I was unable to find a method that works exactly for this.

lib/utilities/writeContinue.js
bennettbuchanan: This method has been updated.

lib/utilities/writeContinue.js
```diff
@@ -0,0 +1,32 @@
...
+export default function writeContinue(req, res, authInfo) {
+    const { headers, query } = req;
+    let authV2 = query.Signature;
+    let authV4 = query['X-Amz-Algorithm'];
+    const authHeader = headers.authorization;
+    if (authHeader) {
+        authV2 = authHeader.startsWith('AWS ');
+        authV4 = authHeader.startsWith('AWS4');
+    }
+    const isStreaming = headers['content-type'] === 'application/octet-stream';
```
bennettbuchanan: This method has been updated.

lib/api/api.js
```diff
@@ -82,6 +84,8 @@ const api = {
                     }
                 }
             }
+            writeContinue(request, response, userInfo);
```
bennettbuchanan: Updated.

lib/api/api.js
```diff
@@ -82,20 +84,50 @@ const api = {
...
+                    post.push(chunk);
+                }
+                return undefined;
+            });
+
+            request.on('error', err => {
+                log.trace('error receiving request', {
+                    error: err,
+                });
+                return undefined;
```
LaurenSpiegel: I think we want to call the callback back to the routes here with an internal error.

lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -6,37 +6,38 @@ const RequestContext = policies.RequestContext;
...
+export default function prepareRequestContexts(apiMethod, request, sourceBucket,
+    sourceObject) {
     // if multiObjectDelete request, we want to authenticate
     // before parsing the post body and creating multiple requestContexts
     // so send null as requestContexts to Vault to avoid authorization
     // checks at this point
-    if (apiMethod === 'multiObjectDelete') {
+    //
+    // If bucketPut request, we want to authenticate in the API itself because
+    // we only have access to the locationConstraint at that point.
```
LaurenSpiegel: If bucketPut request , we want to do the authorization check in the API itself (once we parse the locationConstraint from the xml body) so send null as the
requestContext to Vault so it will only do an authentication check.

tests/unit/api/bucketGetLocation.js
```diff
@@ -59,11 +68,12 @@ describe('getBucketLocation API', () => {
             });
         });
     });
-    [undefined, 'us-east-1'].forEach(location => {
+    ['us-east-1'].forEach(location => {
```
bennettbuchanan: Not sure if `undefined` should be tested since we are no longer giving the value as a parameter.

rahulreddy:
It's better to separate the refactor and 100-continue feature into separate commits. It can still be in the same PR, just logical separation of commits makes it easier to track the feature implementation.

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

lib/api/bucketPut.js
```diff
@@ -39,36 +56,77 @@ export default function bucketPut(authInfo, request, locationConstraint, log,
...
+                        authType: authParams.params.data.signatureVersion,
+                        signatureAge: authParams.params.data.signatureAge,
+                    },
+                };
+                return vault.checkPolicies(requestContextParams,
+                    authInfo.getArn(), log, (err, authorizationResults) => {
+                        if (err) {
+                            return next(err);
+                        }
+                        if (authorizationResults[0].isAllowed === false) {
```
LaurenSpiegel: 1) just to be careful, let's reverse this and do if authorizationResults[0].isAllowed !== true
2) let's add a log saying that the authorization check failed

bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
E2E passed here: http://ci.ironmann.io/gh/scality/Integration/10802.
The dependent PR https://github.com/scality/Vault/pull/722 will need to be merged first.

lib/api/bucketPut.js
```diff
@@ -39,36 +56,79 @@ export default function bucketPut(authInfo, request, locationConstraint, log,
...
-                authInfo,
-                bucket: bucketName,
-            });
-            return callback(null, corsHeaders);
-        });
+            return next(null, locationConstraint);
+        },
+        // AWS JS SDK sends a request with locationConstraint us-east-1 if no
+        // locationConstraint provided.
+        (locationConstraint, next) => {
```
rahulreddy: This need not be another waterfall step as there is no async call here.

lib/api/bucketPutVersioning.js
```diff
@@ -17,6 +18,36 @@ import services from '../services';
     x-amz-mfa: [SerialNumber] [TokenCode]
  */
 
+function _parseXML(request, log, cb) {
+    if (request.post === '') {
+        log.debug('request xml is missing');
+        return cb(errors.MalformedXML);
```
rahulreddy: Use process.nextTick here as well.

lib/api/bucketPutVersioning.js
```diff
@@ -17,6 +18,36 @@ import services from '../services';
     x-amz-mfa: [SerialNumber] [TokenCode]
  */
 
+function _parseXML(request, log, cb) {
+    if (request.post === '') {
```
rahulreddy: `if (!request.post)` check should be sufficient here. Also, need to confirm if `MalformedXML` is the correct error.

lib/api/bucketPut.js
```diff
@@ -39,36 +56,79 @@ export default function bucketPut(authInfo, request, locationConstraint, log,
...
+                if (err) {
+                    return next(err, corsHeaders);
+                }
+                pushMetric('createBucket', log, {
+                    authInfo,
+                    bucket: bucketName,
+                });
+                return next(null, corsHeaders);
+            }),
+    ], (err, corsHeaders) => callback(err, corsHeaders));
```
rahulreddy: `(err, corsHeaders) => callback(err, corsHeaders)` need not be expanded, sending callback as param is sufficient.

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

lib/api/bucketPutVersioning.js
bennettbuchanan: This logic replicates pre-existing code from routePUT.js. See
https://github.com/scality/S3/pull/631/files#diff-25f86c99d9d8badaa3509e13e8cc8edbL70

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
@ironman-machine
SCALITY_VAULT_BRANCH=dev/allow-empty-string-specificRequest
try

bennettbuchanan:
E2E Passed

PR #634, opened 2017-03-06
FIX: Increase bucket listing limit to 10,000

### LINKED ISSUE #633: Truncated listing of buckets per users
The S3 API does not support a continuation mechanism to list the buckets per user. But AFAIK, it does not enforce a specific limit.

The current implementation in services.js suggests that a maximum of 10000 buckets is supported:

        // Note that since maxKeys on a listObject request is 10,000,
        // this request will retrieve up to 10,000 bucket names for a user.

Actually, I believe the comment is wrong and the actual limit is only 1000 (corresponding to the maxKeys default value), which is annoying since users may have thousands of buckets.

Please consider increasing this limit to 10000 or even more.

LaurenSpiegel:
Fixes #633

### DIFF
--- a/tests/functional/aws-node-sdk/test/service/get.js
+++ b/tests/functional/aws-node-sdk/test/service/get.js
@@ -1,6 +1,7 @@
 import assert from 'assert';
 import tv4 from 'tv4';
 import Promise from 'bluebird';
+import async from 'async';
 import { S3 } from 'aws-sdk';
 
 import BucketUtility from '../../lib/utility/bucket-util';
@@ -8,7 +9,13 @@ import getConfig from '../support/config';
 import withV4 from '../support/withV4';
 import svcSchema from '../../schema/service';
 
-describe('GET Service - AWS.S3.listBuckets', () => {
+const describeFn = process.env.AWS_ON_AIR
+    ? describe.skip
+    : describe;
+
+describeFn('GET Service - AWS.S3.listBuckets', function getService() {
+    this.timeout(600000);
+
     describe('When user is unauthorized', () => {
         let s3;
         let config;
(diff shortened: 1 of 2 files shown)


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
@ironman-machine with only one before statement

LaurenSpiegel:
@ironman-machine testy...

LaurenSpiegel:
@ThibaultRiviere, yes.  I'm playing with 1001 right now.  once that works, I can increase to 10,000.

ThibaultRiviere:
@LaurenSpiegel Perfect thanks :)

LaurenSpiegel:
@ironman-machine without retries

tests/functional/aws-node-sdk/test/service/get.js
```diff
@@ -1,14 +1,16 @@
...
+import async from 'async';
 import { S3 } from 'aws-sdk';
 
 import BucketUtility from '../../lib/utility/bucket-util';
 import getConfig from '../support/config';
 import withV4 from '../support/withV4';
 import svcSchema from '../../schema/service';
 
-describe('GET Service - AWS.S3.listBuckets', () => {
+describe.only('GET Service - AWS.S3.listBuckets', function getService() {
```
LaurenSpiegel: just for testing

LaurenSpiegel:
@ironman-machine with slower deletes

LaurenSpiegel:
@ironman-machine with 10000 buckets

LaurenSpiegel:
@ironman-machine with concurrency test

LaurenSpiegel:
@ironman-machine just to see again

LaurenSpiegel:
@ironman-machine once more

PR #635, opened 2017-03-07
Add report handler

RachedBenMustapha:
Returns uuid, server version, system (cpu/mem/disk usage) stats, and config at the moment, minus authentication keys. The report endpoint is protected by an ephemeral token that changes on
restart and is only available to the S3 Server's children as an environment variable.

Needs scality/Arsenal#302


RachedBenMustapha:
@ironman-machine 
SCALITY_INTEGRATION_BRANCH=ft/s3-report-token

RachedBenMustapha:
@ironman-machine

RachedBenMustapha:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/s3-report-token

circle.yml
```diff
@@ -104,6 +105,10 @@ test:
             > $CIRCLE_ARTIFACTS/server_file_rawnode.txt
         & bash wait_for_local_port.bash 8000 40
         && npm run ft_node
+    - S3BACKEND=file S3VAULT=mem npm start
```
LaurenSpiegel: Is there any value to having the reporting work with the mem backend?

tests/functional/report/checkRoutes.js
```diff
@@ -0,0 +1,94 @@
...
+const https = require('https');
+const fs = require('fs');
+
+const conf = require('../config.json');
+
+const transportStr = conf.transport;
+const transport = transportStr === 'http' ? http : https;
+
+function makeAgent() {
+    if (transportStr === 'https') {
```
LaurenSpiegel: This currently would never be the case (only in the end to end). Do you think we should be testing this in just s3?

tests/functional/report/checkRoutes.js
```diff
@@ -0,0 +1,94 @@
...
+        headers: { 'x-scal-report-token': token },
+    };
+}
+
+function queryReport(done, resultCheck) {
+    const req = transport.request(options(), res => {
+        if (res.statusCode !== 200) {
+            return done(new Error(`non-200 status ${res.statusCode}`));
+        }
+
```
LaurenSpiegel: what about checking res.on('error')?

circle.yml
RachedBenMustapha: I don't think there's much value in monitoring a system that forgets data when it's restarted :)

tests/functional/report/checkRoutes.js
RachedBenMustapha: Good idea, will do

tests/functional/report/checkRoutes.js
RachedBenMustapha: You're right, I only realized this after I noticed that the end to end tests were only using scality's backends which won't benefit from this and the tests don't need to run there. I'll remove this.

LaurenSpiegel:
@ironman-machine r+

PR #636, opened 2017-03-07
FIX: Limit bucket deletion blocking for mpu

LaurenSpiegel:
Only block the deletion of a bucket if there is an
overview object in the mpu shadow bucket. If there are
orphan part objects, proceed with the deletion.


LaurenSpiegel:
@ironman-machine check it

LaurenSpiegel:
end to end re-running here http://ci.ironmann.io/gh/scality/Integration/10157
(first run failed due to utapi)

rahulreddy:
I added the failed build to the issue on Jira. We may increase the timeout to reduce utapi flakiness (current 1 sec. wait is not enough)

LaurenSpiegel:
end to end passed https://github.com/scality/S3/pull/634

PR #637, opened 2017-03-08
Add mutiple env variable option

dora-korpar:
This PR creates a new option for the S3DATA env variable of 'multiple.'
When set, it allows the user to set the storage location of an object by setting the 'x-amz-meta-scal-location-constraint' header on a put request.
Current supported locations are mem, file, and sproxyd.


dora-korpar:
@ironman-machine check it!

dora-korpar:
@ironman-machine backendInfo fixed

dora-korpar:
@ironman-machine now

tests/functional/aws-node-sdk/test/bucket/getLocation.js
```diff
@@ -70,27 +70,12 @@ describeSkipAWS('GET bucket location ', () => {
             });
         });
 
-        describe('without location configuration', () => {
```
dora-korpar: Removed this test for redundancy because it's done more dynamically in multipleBackend/put.js

dora-korpar:
@ironman-machine go

jonathan-gramain:
Please try to split into smaller commits with well-defined scope next time if that is possible without too much pain, to help the reviewer ;)

lib/data/locationConstraintParser.js
```diff
@@ -0,0 +1,41 @@
...
+        Object.keys(config.locationConstraints).forEach(location => {
+            const locationObj = config.locationConstraints[location];
+            if (locationObj.type === 'mem') {
+                clients[location] = inMemory;
+            }
+            if (locationObj.type === 'file') {
+                clients[location] = file;
+            }
+            if (locationObj.type === 'scality'
+            && Object.keys(locationObj.details.connector).
```
LaurenSpiegel: I think we can just check whether locationObj.details.connector.sproxyd is defined

lib/data/locationConstraintParser.js
```diff
@@ -0,0 +1,41 @@
+import file from './file/backend';
+import inMemory from './in_memory/backend';
+import Sproxy from 'sproxydclient';
+
+import config from '../Config';
+
+const clients = {};
```
LaurenSpiegel: define this in the function so not working by side effect

lib/data/multipleBackendGateway.js
```diff
@@ -0,0 +1,68 @@
...
+
+    delete: (objectGetInfo, reqUids, callback) => {
+        const client = clients[objectGetInfo.dataStoreName];
+        if (client.clientType === 'scality') {
+            return client.delete(objectGetInfo.key, reqUids, callback);
+        }
+        return client.delete(objectGetInfo, reqUids, callback);
+    },
+
+    // checkHealth:
```
LaurenSpiegel: this needs to be done as well but could be in a subsequent pr.

lib/data/wrapper.js
```diff
@@ -61,61 +67,81 @@ const data = {
...
+                };
+                return cb(null, dataRetrievalInfo, hashedStream);
+            });
     },
 
     get: (objectGetInfo, log, cb) => {
         // If objectGetInfo.key exists the md-model-version is 2 or greater.
         // Otherwise, the objectGetInfo is just the key string.
         const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
+        // const implNameGetInfo = (implName === 'sproxyd') ?
```
LaurenSpiegel: why is this commented out?  does not seem needed?

lib/api/objectPutPart.js
```diff
@@ -118,36 +121,37 @@ export default function objectPutPart(authInfo, request, streamingV4Params, log,
...
+                    return next(null, destinationBucket, res);
                 });
             }
             // The bucket does not have server-side encryption, so pass `null`
-            return next(null, null, bucket);
+            return next(null, destinationBucket, null);
         },
         // Get the MPU shadow bucket.
-        (cipherBundle, bucket, next) => metadata.getBucket(mpuBucketName, log,
+        (destinationBucket, cipherBundle, next) =>
```
jonathan-gramain: This kind of call chains with quite a few order-dependent arguments would benefit from passing them as object and using deconstruction to get them in callees (this is more or a general cleanup that can be the topic of another PR, but it would have helped here I think especially when moving arguments around).

lib/data/wrapper.js
```diff
@@ -61,61 +67,81 @@ const data = {
...
+                    dataStoreName: implName,
+                };
+                return cb(null, dataRetrievalInfo, hashedStream);
+            });
     },
 
     get: (objectGetInfo, log, cb) => {
         // If objectGetInfo.key exists the md-model-version is 2 or greater.
         // Otherwise, the objectGetInfo is just the key string.
         const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
```
LaurenSpiegel: does not seem like you need this since passing along objectGetInfo below?

lib/data/wrapper.js
```diff
@@ -61,61 +67,81 @@ const data = {
...
+                            return cb(null, decipherBundle.decipher);
+                        });
+                }
+                return cb(null, stream);
+            });
     },
 
     delete: (objectGetInfo, log, cb) => {
         const callback = cb || log.end;
+        // const implNameGetInfo = (implName === 'sproxyd') ?
```
LaurenSpiegel: again, why here?

lib/data/wrapper.js
```diff
@@ -170,6 +196,7 @@ const data = {
             return cb(null, defResp);
         }
         return client.healthcheck(log, (err, result) => {
+            // NEED TO UPDATE FOR POSSIBILITY OF MULTIPLE HEALTHCHECKS HERE!!
```
LaurenSpiegel: this is part of the todo for the healthcheck in the gateway.  let's have a todo here and please handle in the subsequent pr.

package.json
```diff
@@ -33,7 +34,7 @@
     "multilevel": "^7.3.0",
     "node-uuid": "^1.4.3",
     "ready-set-stream": "1.0.7",
-    "sproxydclient": "scality/sproxydclient",
+    "sproxydclient": "scality/sproxydclient#ft/multipleBackendGet",
```
LaurenSpiegel: do we still need this?

package.json
```diff
@@ -19,6 +19,7 @@
   },
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
+    "aws-sdk": "^2.2.11",
```
LaurenSpiegel: this is for the aws backend, not this pr.

package.json
```diff
@@ -44,7 +45,6 @@
     "ioctl": "2.0.0"
   },
   "devDependencies": {
-    "aws-sdk": "^2.2.11",
```
LaurenSpiegel: restore this as just a dev dependency until the aws backend pr

lib/data/locationConstraintParser.js
```diff
@@ -0,0 +1,41 @@
...
+
+export default function parseLC() {
```
jonathan-gramain: This function does not seem to parse anything (except some JSON internally) but to initialize a global table of clients, so the name seems a bit off-topic. (same comment for where it's used)

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
```diff
@@ -0,0 +1,64 @@
...
+            .catch(err => {
+                process.stdout.write(`Error deleting bucket: ${err}\n`);
+                throw err;
+            });
+        });
+
+        it('should delete object from mem', done => {
+            s3.deleteObject({ Bucket: bucket, Key: memObject }, err => {
+                assert.strictEqual(err, null,
+                    `Expected success, got error ${JSON.stringify(err)}`);
```
LaurenSpiegel: let's throw in a get to make sure it is not still there

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
```diff
@@ -0,0 +1,64 @@
...
+                done();
+            });
+        });
+        it('should delete object from file', done => {
+            s3.deleteObject({ Bucket: bucket, Key: fileObject }, err => {
+                assert.strictEqual(err, null,
+                    `Expected success, got error ${JSON.stringify(err)}`);
```
LaurenSpiegel: same

lib/data/multipleBackendGateway.js
```diff
@@ -0,0 +1,68 @@
...
+                callback(errors.InternalError);
+            });
+        }
+        return client.put(stream, size, keyContext,
+            reqUids, (err, key) => {
+                if (err) {
+                    const log = createLogger(reqUids);
+                    log.error('error from datastore',
+                             { error: err, implName: client.clientType });
+                    return callback(errors.InternalError);
```
jonathan-gramain: I would rather return the original error here (and I doubt logging here is useful since it shall already be done in the actual backend). Actually I wonder what is the reason to not just call the backend function directly (as it's done for the other API calls get and delete).

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
```diff
@@ -0,0 +1,64 @@
...
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
+            return s3.createBucketAsync({ Bucket: bucket })
+            .catch(err => {
+                process.stdout.write(`Error creating bucket: ${err}\n`);
+                throw err;
+            })
+            .then(() => {
+                process.stdout.write('Putting object to mem\n');
+                const params = { Bucket: bucket, Key: memObject,
```
LaurenSpiegel: If you're not sending a body, you're really using the 0 byte object path and there is nothing to delete.  we should test both paths (with and without a data body)

tests/functional/aws-node-sdk/test/multipleBackend/get.js
```diff
@@ -0,0 +1,89 @@
...
+                    assert.strictEqual(err.code, 'NoSuchKey');
+                    done();
+                });
+            });
+
+        describe('with objects in all available backends ' +
+            '(mem/file)', () => {
+            before(() => {
+                process.stdout.write('Putting object to mem');
+                return s3.putObjectAsync({ Bucket: bucket, Key: memObject,
```
LaurenSpiegel: we should be testing with a 0 byte object and with some data.

tests/functional/aws-node-sdk/test/multipleBackend/get.js
```diff
@@ -0,0 +1,89 @@
...
+                })
+                .catch(err => {
+                    process.stdout.write(`Error putting objects: ${err}\n`);
+                    throw err;
+                });
+            });
+            it('should get an object from mem', done => {
+                s3.getObject({ Bucket: bucket, Key: memObject }, err => {
+                    assert.equal(err, null, 'Expected success but got error ' +
+                        `${err}`);
```
LaurenSpiegel: let's make sure the body is what we expect

tests/functional/aws-node-sdk/test/multipleBackend/get.js
```diff
@@ -0,0 +1,89 @@
...
+                });
+                done();
+            });
+            it('should get an object from file', done => {
+                s3.getObject({ Bucket: bucket, Key: fileObject }, err => {
+                    assert.equal(err, null, 'Expected success but got error ' +
+                        `${err}`);
+                });
```
LaurenSpiegel: same.  let's make sure we got the body back intact.

lib/data/multipleBackendGateway.js
```diff
@@ -0,0 +1,68 @@
...
+                    key,
+                    dataStoreName: controllingLocationConstraint,
+                };
+                return callback(null, dataRetrievalInfo);
+            });
+    },
+
+    get: (objectGetInfo, range, reqUids, callback) => {
+        const client = clients[objectGetInfo.dataStoreName];
+        if (client.clientType === 'scality') {
```
jonathan-gramain: I wonder why is the "scality" backend a special case, the adaptor code `const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;` added in file and memory backends may be inserted in "scality" sproxydclient backend too right, or is it because we need backward compat with that other repo? I think sproxydclient is bundled with the release so I'm not sure we need such retro-compat if that's the idea here.

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,181 @@
...
+            done => {
+                s3.putObject({ Bucket: '', Key: key }, err => {
+                    assert.notEqual(err, null,
+                        'Expected failure but got success');
+                    assert.strictEqual(err.code, 'MethodNotAllowed');
+                    done();
+                });
+            });
+
+        it('should return an error to put request without a valid key name',
```
LaurenSpiegel: I'd remove this test.  It's really just making a put bucket call since there is no key.  You would get a bucketAlreadyOwnedByYou error if not in us-east-1 and the bucket is already owned by you and otherwise the bucket would actually be put.

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,181 @@
...
+                    assert.notEqual(err, null, 'Expected failure but got ' +
+                        'success');
+                    assert.strictEqual(err.code, 'InvalidArgument');
+                    done();
+                });
+            });
+
+            it('should put an object to mem', done => {
+                const params = { Bucket: bucket, Key: key,
+                    Body: body,
```
LaurenSpiegel: we should test 0 byte object code path as well

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,181 @@
...
+                process.stdout.write('Deleting bucket\n');
+                return bucketUtil.deleteOne(bucket);
+            })
+            .catch(err => {
+                process.stdout.write(`Error in afterEach: ${err}\n`);
+                throw err;
+            });
+        });
+
+        it('should put an object mem with no location header', done => {
```
LaurenSpiegel: to mem

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,181 @@
...
+            bucketUtil = new BucketUtility('default', sigCfg);
```
LaurenSpiegel: why is all of this in the actual test rather than a before statement?

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,181 @@
...
+        after(() => {
+            process.stdout.write('Deleting bucket\n');
+            return bucketUtil.deleteOne(bucket)
+            .catch(err => {
+                process.stdout.write(`Error in after: ${err}\n`);
+                throw err;
+            });
+        });
+
+        it('should create bucket in corresponding backend', done => {
```
LaurenSpiegel: shouldn't the test really be about putting an object once the bucket is created without a locationConstraint?  I think you need to take the next step and put the object.

tests/functional/aws-node-sdk/test/bucket/getLocation.js
LaurenSpiegel: the test in put should really be here since it is testing the getLocation api call.  in the test in put, you should take the next step and put an object.  see comment in that file.

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,181 @@
...
+                const host = request.service.endpoint.hostname;
+                const endpoint = config.restEndpoints[host];
+                s3.getBucketLocation({ Bucket: bucket }, (err, data) => {
+                    assert.strictEqual(err, null, 'Expected succes, ' +
+                        `got error ${JSON.stringify(err)}`);
+                    assert.strictEqual(data.LocationConstraint, endpoint);
+                    done();
+                });
+            });
+        });
```
LaurenSpiegel: we should also test mpu.  instead of doing "manual mpu's" like in some of our tests, you could use the node aws sdk upload call with a proper size object.

tests/multipleBackend/objectPut.js
```diff
@@ -0,0 +1,148 @@
...
+import bucketPut from '../../lib/api/bucketPut';
+import objectPut from '../../lib/api/objectPut';
+import DummyRequest from '../unit/DummyRequest';
+
+const log = new DummyRequestLogger();
+const canonicalID = 'accessKey1';
+const authInfo = makeAuthInfo(canonicalID);
+const namespace = 'default';
+const bucketName = 'bucketname';
+const postBody = Buffer.from('I am a body', 'utf8');
```
LaurenSpiegel: I would just call this 'body' since it's the data being put on a put request rather than a post request.

tests/multipleBackend/objectPut.js
```diff
@@ -0,0 +1,148 @@
...
+        const testPutObjectRequest = new DummyRequest(objPutParams,
+            postBody);
+        bucketPut(authInfo, bucketPutRequest, locationConstraint,
+            log, () => {
+                objectPut(authInfo, testPutObjectRequest, undefined, log,
+                    (err, result) => {
+                        assert.strictEqual(err, null, 'Error putting object ' +
+                            `${err}`);
+                        assert.strictEqual(result, correctMD5);
+                        assert.deepStrictEqual(ds, []);
```
LaurenSpiegel: put a comment stating that since it is not in mem it must be file since that is the only possibility for the unit tests

tests/multipleBackend/objectPutCopyPart.js
```diff
@@ -0,0 +1,322 @@
...
+const namespace = 'default';
+
+const bucketName = 'superbucket9999999';
+const sourceObjName = 'supersourceobject';
+const destObjName = 'copycatobject';
+const fileLocationConstraint = 'file';
+const memLocationConstraint = 'mem';
+const mpuBucket = `${constants.mpuBucketPrefix}${bucketName}`;
+
+const postBody = Buffer.from('I am a body', 'utf8');
```
LaurenSpiegel: same.  call this 'body'

tests/multipleBackend/objectPutPart.js
```diff
@@ -0,0 +1,327 @@
...
+        };
+        async.waterfall([
+            next => bucketPut(authInfo, bucketPutRequest,
+                locationConstraint, log, next),
+            (corsHeaders, next) => initiateMultipartUpload(authInfo,
+                initiateRequest, log, next),
+            (result, corsHeaders, next) => {
+                const mpuKeys = metadata.keyMaps.get(mpuBucket);
+                assert.strictEqual(mpuKeys.size, 1);
+                // assert(mpuKeys.keys().next().value
```
LaurenSpiegel: why commented out?

tests/multipleBackend/objectPutPart.js
```diff
@@ -0,0 +1,327 @@
...
+            bucketName,
+            namespace,
+            objectName,
+            headers: { host: `${bucketName}.s3.amazonaws.com` },
+            url: `/${objectName}?uploads`,
+        };
+    });
+
+    it('should upload a part to file based on mpu location', done => {
+        locationConstraint = 'mem';
```
LaurenSpiegel: you could create a function to do these tests with the parameters you are changing between tests to avoid the code duplication

tests/unit/api/multiObjectDelete.js
```diff
@@ -85,7 +85,7 @@ describe('getObjMetadataAndDelete function for multiObjectDelete', () => {
                     // eslint-disable-next-line
                     assert.deepStrictEqual(ds, [ , , ]);
                     done();
-                }, 20);
+                }, 500);
```
LaurenSpiegel: there was this noticeable of a difference?  need to change comment in line 82

init.js
```diff
@@ -62,7 +62,8 @@ function printUUID(metadataPath) {
     logger.info(`This deployment's identifier is ${uuidValue}`);
 }
 
-if (config.backends.data !== 'file' && config.backends.metadata !== 'file') {
+if (config.backends.data !== 'file' && config.backends.data !== 'multiple'
+    || config.backends.metadata === 'scality') {
```
dora-korpar: Wouldn't we still want to init if, for example, data = file/multiple and metadata = mem/scality?
Which would make it just:
(config.backends.data !== 'file' && config.backends.data !== 'multiple' && config.backends.metadata !== 'file')

lib/api/bucketPut.js
```diff
@@ -41,7 +41,7 @@ export default function bucketPut(authInfo, request, locationConstraint, log,
     }
     // - AWS JS SDK sends a request with locationConstraint us-east-1
     // if no locationConstraint provided.
-    if (locationConstraint && config.locationConstraints &&
+    if (locationConstraint && !config.regions && config.locationConstraints &&
```
dora-korpar: That check had to be added because now that we have 2 configs, it's possible to have the locationConfig and the legacy version of the regular config. And if locationConfig is present, regardless of the version of the regular config, config.locationConstraints will be set.

I think there are 2 possible ways to fix this. One would be to have some logic in Config.js that would prevent config.locationConstraint from being set if the regular config was the legacy version. The other is to change the logic in conditions like this to rely on !config.region or config.restEndpoints as the indicator of legacy or new config instead of config.locationConstraint.

lib/data/wrapper.js
dora-korpar: I've uncommented both of these lines. They're needed because sproxyd expects a key string instead of an object

lib/data/multipleBackendGateway.js
dora-korpar: I've got a branch going :)

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -0,0 +1,144 @@
...
+     * validate requestEndpoint against config
+     * if there is a mismatch between the request endpoint and what is in
+     * the config, this could cause a problem for setting the backend location
+     * for storing data
+     * @param {string} requestEndpoint - request endpoint
+     * @param {object} log - werelogs logger
+     * @return {boolean} - true if valid, false if not
+     */
+    static isValidRequestEndpoint(requestEndpoint, log) {
+        if (utils.getAllEndpoints().indexOf(requestEndpoint) < 0) {
```
LaurenSpiegel: this should have been Object.keys(config.restEndpoints).

We want to make sure that the requestEndpoint matches an entry in the dictionary of restEndpoints

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -0,0 +1,144 @@
...
+            log.trace('requestEndpoint does not match config restEndpoints',
+                { requestEndpoint });
+            return false;
+        }
+        if (!config.regions && config.locationConstraints &&
+            Object.keys(config.locationConstraints).indexOf(config
```
LaurenSpiegel: @jonathan-gramain, thank you for catching!  the use of utils.getAllEndpoints() above was a mistake (probably mine)

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -0,0 +1,144 @@
...
+    /**
+    * Return locationConstraint that should be used with put request
+    * Order of priority is:
+    * (1) objectLocationConstraint,
+    * (2) bucketLocationConstraint,
+    * (3) default locationConstraint for requestEndpoint
+    * @return {string} locationConstraint;
+    */
+    getControllingLocationConstraint() {
+        if (this._objectLocationConstraint) {
```
LaurenSpiegel: fair point.  let's use the getter throughout the function.

lib/api/bucketPut.js
LaurenSpiegel: If we have both the legacy property of config.regions **and** the new property of locationConstraint, we should throw an error when parsing the config and the server shouldn't start.  Same thing if we have locationConstraints but are missing restEndpoints.

LaurenSpiegel: @jonathan-gramain, this is more about checking to make sure we do not have the legacy config.

lib/api/apiUtils/object/BackendInfo.js
dora-korpar: utils.getAllEndpoints() actually returns Object.keys(config.restEndpoints) for the new config, and it works with the old config too, which is why I used it :)

lib/api/bucketPut.js
LaurenSpiegel: And as just discussed, if we have the env variable "multiple" set we should make sure we have both locationConstraints and restEndpoints and otherwise throw an error at the config parsing stage.

lib/api/objectPutCopyPart.js
```diff
@@ -122,29 +127,86 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                         copyLocator.copyObjectSize);
                 });
         },
-        function checkMPUBucketAuth(dataLocator, destBucketMD,
```
LaurenSpiegel: The services.metadataValidateMultipart function tries to do too much and then doesn't do something we need to do here now-- get from the mpu overview object the controlling location for the data for the mpu.  We are still using the services.metadataValidateMultipart in a few other api's where it fits the need though it could certainly be refactored at some point.

lib/data/multipleBackendGateway.js
LaurenSpiegel: The error might not be an arsenal error once we are using a real AWS backend since the client will be the node aws sdk but we can deal with that when we add the aws backend.

lib/data/multipleBackendGateway.js
LaurenSpiegel: yes, we could add that to sproxyd client.  metadata uses sproxyd client as well for backups so we were trying to not mess with it but that change would not break anything.

lib/Config.js
```diff
@@ -179,8 +187,7 @@ class Config {
```
LaurenSpiegel: let's do another check to make sure there is at least regions or restEndpoints

lib/utilities/locationConstraintCheck.js
```diff
@@ -0,0 +1,35 @@
+import { errors } from 'arsenal';
+
+import config from '../Config';
+import { BackendInfo } from '../api/apiUtils/object/BackendInfo';
+import constants from '../../constants';
+
+export default function locationConstraintCheck(request, metaHeaders,
```
LaurenSpiegel: this function should not take a callback.  It is only doing synchronous actions.  I would make it a synchronous function that returns either backendInfo or an arsenal error.  If it returns an error, then from the api you call your cb with that error.

lib/utilities/locationConstraintCheck.js
```diff
@@ -0,0 +1,35 @@
+import { errors } from 'arsenal';
+
+import config from '../Config';
+import { BackendInfo } from '../api/apiUtils/object/BackendInfo';
```
LaurenSpiegel: I would move this locationConstraintCheck file to api/apiUtils/object as well.  It's used only on object operations, correct?  We've been trying to split our utilities so they are somewhat easy to find.

lib/api/initiateMultipartUpload.js
```diff
@@ -136,30 +139,47 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
-                if (MPUbucket.getMdBucketModelVersion() < 2) {
-                    metadataStoreParams.splitter = constants.oldSplitter;
-                }
-                services.metadataStoreMPObject(MPUbucket.getName(),
-                    cipherBundle, metadataStoreParams,
-                    log, err => {
-                        if (err) {
-                            log.trace('error storing multipart object', {
-                                error: err,
+        return locationConstraintCheck(request, null, destinationBucket, log,
```
LaurenSpiegel: as noted below, this function should be used synchronously.

lib/utilities/locationConstraintCheck.js
dora-korpar: agreed!

lib/api/initiateMultipartUpload.js
```diff
@@ -136,30 +139,47 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
+        (err, backendInfo) => {
+            if (err) {
+                return process.nextTick(() => {
+                    callback(err.customizeDescription('Location Constraint ' +
+                    'Info is invalid.'));
+                });
+            }
+            if (config.regions) {
+                metadataStoreParams.controllingLocationConstraint =
+                   backendInfo;
```
LaurenSpiegel: can you just return the locationConstraint so you can treat these two assignments the same?

lib/api/initiateMultipartUpload.js
```diff
@@ -136,30 +139,47 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
+        return locationConstraintCheck(request, null, destinationBucket, log,
+        (err, backendInfo) => {
+            if (err) {
+                return process.nextTick(() => {
+                    callback(err.customizeDescription('Location Constraint ' +
+                    'Info is invalid.'));
+                });
+            }
+            if (config.regions) {
+                metadataStoreParams.controllingLocationConstraint =
```
LaurenSpiegel: how are we modifying metadataStoreParams here?  we shouldn't be modifying the greater scope.  if we need to change it, we should send it into this function as an argument.

lib/api/objectDelete.js
```diff
@@ -54,8 +54,9 @@ export default function objectDelete(authInfo, request, log, cb) {
                     contentLength: objMD['content-length'],
                 });
             }
-            return services.deleteObject(bucketName, objMD, objectKey, log,
-                err => {
+
+            return services.deleteObject(bucketName, objMD, objectKey,
```
LaurenSpiegel: this looks like we are just changing a line break.  let's revert so that no changes appear in this file.

dora-korpar:
@ironman-machine let's go

lib/api/objectDelete.js
```diff
@@ -54,8 +56,21 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+            const bucketLocationConstraint = bucket.getLocationConstraint();
+            const requestEndpoint = request.parsedHost;
+            if (!BackendInfo.areValidBackendParameters(objectLocationConstraint,
+                bucketLocationConstraint, requestEndpoint, log)) {
+                return process.nextTick(() => {
+                    cb(errors.InvalidArgument
+                        .customizeDescription('Location Constraint Info is ' +
+                        'invalid.'));
+                });
+            }
```
LaurenSpiegel: hi, any thoughts on this?  I'm unclear why we need this check on a delete.

lib/api/apiUtils/object/locationConstraintCheck.js
```diff
@@ -0,0 +1,46 @@
+import { errors } from 'arsenal';
+
+import config from '../../../Config';
+import { BackendInfo } from './BackendInfo';
+import constants from '../../../../constants';
+
+export default function locationConstraintCheck(request, metaHeaders,
```
LaurenSpiegel: please add jsdocs

lib/api/objectDelete.js
dora-korpar: It's been removed :)

LaurenSpiegel: cool.  Not sure why it came up in my review.  Github had a historical moment.  Other than the cors stuff, this PR is looking good to me 😃

lib/api/objectPut.js
```diff
@@ -78,18 +81,29 @@ function _storeIt(bucketName, objectKey, objMD, authInfo, canonicalID,
     // 0 or > 0 - existing object with content-length 0 or greater than 0
     const prevContentLen = objMD && objMD['content-length'] !== undefined ?
         objMD['content-length'] : null;
+
+    const backendInfoObj = locationConstraintCheck(request, null, bucket, log);
+    if (backendInfoObj.err) {
+        return process.nextTick(() => {
```
nicolas2bert: Question: Why do we use ` process.nextTick()` here?

lib/Config.js
```diff
@@ -39,6 +41,9 @@ export function locationConstraintAssert(locationConstraints) {
         assert(typeof locationConstraints[l].type === 'string',
             'bad config: locationConstraints[region].type is ' +
             'mandatory and must be a string');
+        assert(supportedBackends.indexOf(locationConstraints[l].type) > -1,
```
bennettbuchanan: I had encountered this before: the `includes` array method did not work with our previous node version 4. However it should be good now that we're using node v6.9.5.

dora-korpar: Ah yeah, I was wondering about that. That's why we use `.indexOf` everywhere else in the code. Maybe cleaning that up could be its own separate PR

dora-korpar:
@ironman-machine do it

lib/api/objectPut.js
dora-korpar: As discussed with Lauren, this makes the synchronous check work asynchronously for consistency in the file :)

dora-korpar:
@ironman-machine last one

dora-korpar:
@ironman-machine really last one

dora-korpar:
@ironman-machine try
SCALITY_FEDERATION_BRANCH=ft/addnewconfig
SCALITY_INTEGRATION_BRANCH=ft/addnewconfig

dora-korpar:
@ironman-machine try
SCALITY_FEDERATION_BRANCH=ft/addnewconfig
SCALITY_INTEGRATION_BRANCH=ft/addnewconfig

lib/Config.js
```diff
@@ -86,58 +91,51 @@ class Config {
...
         let locationConfig;
-        // check exist file
-        try {
-            fs.statSync(this.locationConfigPath);
-        } catch (err) {
-            if (err) {
-                // location config path is invalid
-                // legacy config format where only one config file
-                return;
+        if (this.restEndpoints) {
```
LaurenSpiegel: I don't think this should be in a conditional anymore.  They must have the locationConfig

lib/Config.js
```diff
@@ -488,16 +472,15 @@ class Config {
...
-                    'must be set'
-                );
-            }
             data = process.env.S3DATA;
         }
-        if (process.env.S3SPROXYD) {
-            data = process.env.S3SPROXYD;
+        if (data === 'scality' || data === 'multiple') {
+            data = 'multiple';
+            assert(this.locationConstraints !== undefined &&
```
LaurenSpiegel: don't you do this check regardless of the data value?  they should have locationConstraints and restEndpoints no matter what

config.json
```diff
@@ -5,8 +5,8 @@
         "localhost": "file",
         "127.0.0.1": "file",
         "s3.docker.test": "scality-us-west-1",
-        "127.0.0.2": "aws-us-east-1",
-        "s3.amazonaws.com": "aws-us-east-1"
+        "127.0.0.2": "scality-us-east-1",
+        "s3.amazonaws.com": "scality-us-east-1"
```
LaurenSpiegel: good point.  @dora-korpar ?

dora-korpar: Sounds good to me

lib/api/objectCopy.js
```diff
@@ -217,10 +219,21 @@ function objectCopy(authInfo, request, sourceBucket,
...
             }
-             // dataLocator is an array.  need to get and put all parts
-             // For now, copy 1 part at a time. Could increase the second
-             // argument here to increase the number of parts
-             // copied at once.
+
+            const backendInfoObj = locationConstraintCheck(request,
+                storeMetadataParams.metaHeaders, destBucketMD, log);
+            if (backendInfoObj.err) {
+                return process.nextTick(() => {
```
LaurenSpiegel: I would get rid of the process.nextTick() here since we're already inside async.waterfall which is controlling the error callback behavior behind the scenes.

lib/api/objectPut.js
LaurenSpiegel: let's get rid of it here though since the function is already returning an err above directly.  at some point, we could do a clean up pr to make this all consistent.

tests/functional/aws-node-sdk/test/multipleBackend/get.js
```diff
@@ -0,0 +1,109 @@
...
+            done => {
+                s3.getObject({ Bucket: bucket, Key: 'nope' }, err => {
+                    assert.notEqual(err, null,
+                        'Expected failure but got success');
+                    assert.strictEqual(err.code, 'NoSuchKey');
+                    done();
+                });
+            });
+
+        describe('with objects in all available backends ' +
```
nicolas2bert: skipIfE2E :)

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,237 @@
...
+                    s3.getObject({ Bucket: bucket, Key: key }, (err, res) => {
+                        assert.strictEqual(err, null, 'Expected success, ' +
+                        `got error ${err}`);
+                        assert.strictEqual(res.ETag, `"${correctMD5}"`);
+                        done();
+                    });
+                });
+            });
+
+            it('should put a 0-byte object to mem', done => {
```
nicolas2bert: skipIfE2E

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,237 @@
...
+                        assert.strictEqual(res.ETag, `"${correctMD5}"`);
+                        done();
+                    });
+                });
+            });
+        });
+    });
+});
+
+describe('MultipleBackend put object based on bucket location', () => {
```
nicolas2bert: skipIfE2E

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -0,0 +1,237 @@
...
+                return bucketUtil.deleteOne(bucket);
+            })
+            .catch(err => {
+                process.stdout.write(`Error in afterEach: ${err}\n`);
+                throw err;
+            });
+        });
+
+        // SKIP because not mem location constraint in E2E.
+        itSkipIfE2E('should put an object to mem with no location header',
```
nicolas2bert: this skip would be useless if you skip the describe.

tests/functional/aws-node-sdk/test/bucket/put.js
```diff
@@ -47,19 +44,17 @@ describe('PUT Bucket - AWS.S3.createBucket', () => {
             bucketUtil = new BucketUtility('default', sigCfg);
         });
 
-        // Why describeSkipIfOldConfig?
-        // AWS returns 404 - NoSuchUpload in us-east-1. This behavior
-        // can be toggled to be compatible with AWS by enabling
-        // usEastBehavior in the config.
-        describeSkipIfOldConfig('create bucket twice', () => {
+        describe('create bucket twice', () => {
             beforeEach(done => bucketUtil.s3.createBucket({ Bucket:
```
nicolas2bert: add location us-east-1 in the `before each`:
```
beforeEach(done => bucketUtil.s3.createBucket({ Bucket: bucketName, 
    CreateBucketConfiguration: {
         LocationConstraint: 'us-east-1',
     },
 }, done));
```

dora-korpar:
@ironman-machine try
SCALITY_FEDERATION_BRANCH=ft/addnewconfig
SCALITY_INTEGRATION_BRANCH=ft/addnewconfig

dora-korpar:
@ironman-machine try
SCALITY_FEDERATION_BRANCH=ft/addnewconfig
SCALITY_INTEGRATION_BRANCH=ft/addnewconfig

PR #642, opened 2017-03-13
DEV: Use fog-aws metagem

bennettbuchanan:
Prevent wait in CircleCI that causes a timeout
```
[...]
fog's executable "fog" conflicts with fog-digitalocean
Overwrite the executable? [yN]  
command gem install fog took more than 10 minutes since last output
```

Currently all fog providers are getting separated into metagems to lower the load time and dependency count. Thus we should be using `fog-aws` instead of requiring the full fog collection to avoid unnecessary dependencies. (See Dependency Notice section of https://github.com/fog/fog.)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -25,7 +25,7 @@ dependencies:
     - wget http://launchpadlibrarian.net/222422124/s3cmd_1.6.0-2_all.deb
     - sudo dpkg -i s3cmd*.deb
     # fog and ruby testing dependencies
-    - gem install fog
+    - gem install fog-aws
     - gem install mime-types
     - gem install rspec
     - gem install json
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -315,7 +315,7 @@ resp = s3.list_buckets
 #### [fog](http://fog.io/storage/)
```
bennettbuchanan: This link does not use `require fog/aws`. Perhaps the README should stay as is, to reflect the examples on that page?

bennettbuchanan:
@ironman-machine

README.md
LaurenSpiegel: does requiring regular fog still work just slower?  if so, keep this as is.

bennettbuchanan: Indeed, it does. Updated to keep the README as is.

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=dev/use-aws-fog-gem
try

PR #648, opened 2017-03-14
S3C-35 Refactoring of bucketfile

jonathan-gramain:
The changes allow bucketfile to use a new API in Arsenal to
    communicate with a remote leveldb database, containing sublevels for
    bucket storage. Metadata is still stored on a local levelDB server,
    but it should now be easy to move the storage logic in a new daemon
    running on a remote server, and it should be robust.
    
    Instead of relying on the existing implementation of multilevel, it
    uses client/server wrappers around a new level-net communication
    protocol and API in Arsenal based on socket.io to exchange messages.
    
    It shall be compatible with the existing metadata since it still uses
    the core sublevel module for the storage logic, only the RPC procotol
    has changed.
    
    Test done:
    
     - put a few 100s of files in different S3 subdirectories
     - list directories and subdirectories
     - get/delete files
     - multi-part upload
     - introduce random connection errors (tcpkill) to check robustness
       and automatic reconnection

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -269,6 +269,20 @@ class Config {
             }
         }
 
+        if (config.metadataDaemon) {
+            this.metadataDaemon = {};
+            assert.strictEqual(
+                typeof config.metadataDaemon.host, 'string',
+                'bad config: metadata daemon host must be a string');
+            this.metadataDaemon.host = config.metadataDaemon.host;
+
+            assert(Number.isInteger(config.metadataDaemon.port)
+                   && config.metadataDaemon.port > 0,
+                   'bad config: metadata daemon port must be a ' +
+                   'positive integer');
+            this.metadataDaemon.port = config.metadataDaemon.port;
+        }
+
         if (process.env.ENABLE_LOCAL_CACHE) {
             this.localCache = defaultLocalCache;
         }
--- a/lib/metadata/bucketfile/backend.js
+++ b/lib/metadata/bucketfile/backend.js
@@ -1,12 +1,5 @@
-import net from 'net';
-import fs from 'fs';
 import cluster from 'cluster';
-import events from 'events';
-import assert from 'assert';
 
-import level from 'level';
-import multilevel from 'multilevel';
-import sublevel from 'level-sublevel';
 import arsenal from 'arsenal';
 
 import { logger } from '../../utilities/logger';
@@ -15,12 +8,8 @@ import constants from '../../../constants';
 import config from '../../Config';
 
 const errors = arsenal.errors;
+const MetadataClient = arsenal.storage.metadata.client;
 
-const METADATA_PORT = 9990;
-const METADATA_PATH = `${config.filePaths.metadataPath}/`;
-const MANIFEST_JSON = 'manifest.json';
-const MANIFEST_JSON_TMP = 'manifest.json.tmp';
-const ROOT_DB = 'rootDB';
 const METASTORE = '__metastore';
 const OPTIONS = { sync: true };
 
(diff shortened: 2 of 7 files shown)


lib/metadata/bucketfile/backend.js
```diff
@@ -29,152 +21,41 @@ class BucketFileInterface {
...
+              log: config.log });
+            this.mdServer.startServer();
+        }
+        this.mdClient = new MetadataClient(
+            { metadataHost: 'localhost',
+              metadataPort: METADATA_PORT,
+              log: config.log });
+        this.mdDB = this.mdClient.openDB();
+        this.metastore = this.mdDB.openSub(METASTORE);
+        if (cluster.isMaster) {
```
LaurenSpiegel: why have the conditional again?

lib/metadata/bucketfile/backend.js
```diff
@@ -29,152 +21,41 @@ class BucketFileInterface {
...
-            cbDone = true;
-            process.nextTick(() => {
-                this.reConnect(cb);
+        this.metastore.put(
+            constants.usersBucket,
+            usersBucketAttr.serialize(), err => {
+                if (err) {
+                    this.logger.error('error writing usersBucket ' +
+                                      'attributes to metadata',
+                                      { error: err });
```
LaurenSpiegel: If this error is due to something other than the bucket already being there, we probably do not want to even start up so we should throw an error.

lib/metadata/bucketfile/backend.js
```diff
@@ -184,121 +65,74 @@ class BucketFileInterface {
...
-        this.ref();
-        this.loadDBIfExistsNoRef(bucketName, log, (err, db, attr) => {
+        this.getBucketAttributes(bucketName, log, (err, attr) => {
             if (err) {
-                this.unRef();
                 return cb(err);
             }
-            // we hold a ref here
-            return cb(err, db, attr);
+            const db = this.mdDB.openSub(bucketName);
```
LaurenSpiegel: why no try/catch around this one but there is one around the openSub command in create bucket?

lib/metadata/bucketfile/backend.js
```diff
@@ -29,152 +21,41 @@ class BucketFileInterface {
...
+              metadataPort: METADATA_PORT,
+              log: config.log });
+            this.mdServer.startServer();
+        }
+        this.mdClient = new MetadataClient(
+            { metadataHost: 'localhost',
+              metadataPort: METADATA_PORT,
+              log: config.log });
+        this.mdDB = this.mdClient.openDB();
+        this.metastore = this.mdDB.openSub(METASTORE);
```
LaurenSpiegel: please add comment that metastore is for storing bucket attributes

lib/metadata/bucketfile/backend.js
jonathan-gramain: Correct me if I'm wrong, the conditional is still necessary because we did not change the architecture yet, the server is still spawned locally on the same container, so it has to be unique, and one way of doing it (that was there before) is to check that we're the master process in the cluster.

The proper separation of the server in a separate daemon will be the topic of another PR :)

LaurenSpiegel: as discussed, i was just commenting on the fact that the same condition is on line 23

jonathan-gramain: Now I remember why there are two `isMaster` checks done, it's because the setup has to be done in three phases:

1. start the metadata server, so that it can accept clients
2. create a metadata client and connect to the server
3. setup the basic dbs on the storage (namely the usersBucket db), which is done through the client (previously it was done on the local db, but it would not work with a remote db).

So steps 1 and 3 are done exclusively on the master.

lib/metadata/bucketfile/backend.js
jonathan-gramain: Even better, we can actually remove this code! (since it has no side-effects).

The other try/catch is not actually necessary since openSub does not throw, but I did it as a good practice to ensure we call the callback in case errors are thrown by some code below.

jonathan-gramain: well, not this one, the other with the try/catch :)

jonathan-gramain:
Depends on Arsenal branch ft/S3C-35-metadataServer-2

lib/Config.js
```diff
@@ -243,6 +243,23 @@ class Config {
             }
         }
 
+        this.metadataDaemon = {};
```
rahulreddy: It's better to add defaults here since host and port are required for the file backend to start.
```js
this.metadataDaemon = { host: 'localhost', port: 9990 }
```

lib/Config.js
```diff
@@ -243,6 +243,23 @@ class Config {
...
+        if (config.metadataDaemon) {
+            if (config.metadataDaemon.host !== undefined) {
```
rahulreddy: Either you would create the defaults or ensure that both host and port are defined here. The if conditionals allow one of them to be defined and the other undefined.

package.json
```diff
@@ -41,7 +38,6 @@
     "xml2js": "~0.4.16"
   },
   "optionalDependencies": {
```
rahulreddy: You can remove this entire section of optionalDependencies.

lib/metadata/bucketfile/backend.js
```diff
@@ -29,152 +19,42 @@ class BucketFileInterface {
     constructor() {
         this.logger = logger;
         if (cluster.isMaster) {
-            this.startServer();
-        } else {
-            this.refcnt = 0;
-            this.waitreco = 0;
-            this.realReConnect();
-            this.notifier = new events.EventEmitter();
+            this.mdServer = new MetadataServer(
```
ttvi: I think you should be able to start the metadata server in `init.js`; this would make metadata, data, and S3 more separated and it's easier to deploy more S3 to use the same existing metadata and data backends.

lib/Config.js
```diff
@@ -269,6 +269,28 @@ class Config {
...
+                assert(Number.isInteger(config.metadataDaemon.port)
+                       && config.metadataDaemon.port > 0,
+                       'bad config: metadata daemon port must be a ' +
+                       'positive integer');
+                this.metadataDaemon.port = config.metadataDaemon.port;
+            } else {
+                assert(config.metadataDaemon.host === undefined &&
+                       config.metadataDaemon.port === undefined,
+                       'bad config: you must specify both host and ' +
+                       'port of metadata daemon or neither of them');
```
ttvi: I don't think you need this assertion and the condition `if` above because the assertions of `host` and `port` in that `if` does the job of the whole block. Just need to move all the assignment to after the assertions.

jonathan-gramain: Not sure I understand what you mean. This assertion ensures that host and port are *both* null, which is not guaranteed by the above test. Please correct me if I miss the point of your comment.

ttvi: That means you allow the case when either of `host` and `port` is valid and the other is not? What I commented applies to the situation where we want both values to be valid.

jonathan-gramain: The error cases are:

- If `host` is valid and `port` is not or vice-versa, one of the first asserts will raise.
- If you provide `host` but not `port` or vice-versa, the last assert will raise.

I think all asserts here are necessary and sufficient to catch those errors.

lib/metadata/bucketfile/backend.js
```diff
@@ -429,31 +260,38 @@ class BucketFileInterface {
...
-                })
-                .on('end', () => {
-                    if (!cbDone) {
-                        this.unRef();
-                        cbDone = true;
-                        const data = extension.result();
-                        cb(null, data);
-                    }
-                });
+            db.createReadStream(requestParams, (err, stream) => {
```
ttvi: I think it's more standard here to use `createReadStream` as a synchronous call since we catch the events `error` of the result stream.

lib/metadata/bucketfile/backend.js
```diff
@@ -15,166 +8,45 @@ import constants from '../../../constants';
...
 const errors = arsenal.errors;
+const MetadataClient = arsenal.storage.metadata.client;
 
-const METADATA_PORT = 9990;
-const METADATA_PATH = `${config.filePaths.metadataPath}/`;
-const MANIFEST_JSON = 'manifest.json';
-const MANIFEST_JSON_TMP = 'manifest.json.tmp';
-const ROOT_DB = 'rootDB';
 const METASTORE = '__metastore';
 const OPTIONS = { sync: true };
```
ttvi: Do you think this should be controlled on the server side?

lib/metadata/bucketfile/backend.js
ttvi: Unless the error is from socket.io.

lib/metadata/bucketfile/backend.js
jonathan-gramain: the "sync: true" options? maybe that could be a hard-coded option on the server indeed

jonathan-gramain: let's keep this for the versioning changes then

lib/metadata/bucketfile/backend.js
jonathan-gramain: Keep for later: https://scality.atlassian.net/browse/S3C-175

lib/metadata/bucketfile/backend.js
ttvi: I think this should be on the server side to be inline with what we have in metadata now. We use ann environment variable for SYNC.

ttvi: Yes, let's save it for versioning.

mdserver.js
```diff
@@ -0,0 +1,14 @@
...
+require('babel-core/register');
+
+const config = require('./lib/Config.js').default;
+const MetadataServer = require('arsenal').storage.metadata.server;
+
+if (config.backends.metadata === 'file') {
+    const mdServer = new MetadataServer(
+        { metadataPath: config.filePaths.metadataPath,
+          metadataPort: config.metadataDaemon.port,
+          log: config.log });
```
ttvi: This `config.log` is only the logging configuration, not a logger, unless that's the purpose.

ttvi: NVM, it seems to be the purpose.

jonathan-gramain:
@rahulreddy I think we should be good to merge when the tests are done.

LaurenSpiegel:
@ironman-machine just to make sure

ttvi:
I don't think we the bucketfile backend for integration.

LaurenSpiegel:
We don't.  just running the end to end to be safe and good practice.

PR #649, opened 2017-03-17
Tiduster/no root

Tiduster:
Hi !
Just some thoughts about the Dockerfile.

First, git is installed and not removed. Maybe this is intentional ? But I couldn't find any use case for this package.
I also added the --purge option. The docker image is reduced from 369MB to 312MB.

Then, as we move to production usage, we didn't want Scality to run as root user.
This is just a proposition, there is a lot of different ways to do that.
On our infrastructure, scality run as user scality, and we didn't see any side effect.

Best regards,

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -6,6 +6,7 @@
   * [Using Docker Volume in production](#using-docker-volume-in-production)
   * [Adding modifying or deleting accounts or users credentials](#adding-modifying-or-deleting-accounts-or-users-credentials)
   * [Specifying your own host name](#specifying-your-own-host-name)
+  * [Running as an unprivileged user](#running-as-an-unprivileged-user)
 
 ## For continuous integration with Docker
 
@@ -154,3 +155,26 @@ docker run -v $(pwd)/config.json:/usr/src/app/config.json -p 8000:8000 -d scalit
 
 Your local `config.json` file will override the default one through a docker
 file mapping.
+
+### Running as an unprivileged user
+
+S3 Server runs as root by default.
+
+You can change that by modifing the dockerfile and specifying a user before the entrypoint.
+
+The user needs to exist within the container,
+and own the folder **/usr/src/app** for Scality S3 Server to run properly.
+
+For instance, you can modify these lines in the dockerfile:
+
+```shell
+...
+&& groupadd -r -g 1001 scality \
+&& useradd -u 1001 -g 1001 -d /usr/src/app -r scality \
+&& chown -R scality:scality /usr/src/app
+
+...
+
+USER scality
+ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]
+```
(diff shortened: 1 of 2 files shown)


GiorgioRegni:
Thanks, indeed I don't think there's any reason to run as root so that's something the team will be looking at.

nicolas2bert:
@Tiduster - First, thanks for this PR.

About git, it isn't necessary per-se, but can be useful to confirm which git version is bundled, easing investigations and bug reproduction.  
Since most of the time we don't need to debug, git just waste space and we can easily do `apt-get update && apt-get install -y git`, I am ok getting rid of it.

nicolas2bert:
Thanks for the PR. 

Regarding changing the user to scality, it will pose an issue for someone who was already running the container as root -- the volumes won't be accessible by the new scality user.  Also, setting permissions can vary based on the system being used so we rather leave that up to the person setting up the system. For example, while playing with this we noticed that on a mac, changing the user does not have an impact on the stored volume since docker for mac is mapping a user.  We will add a comment in our docs stating that the dockerfile can be changed to specify a user if someone wishes to run as non-root.  

If you update your PR to just contain the git removal and sign your commit, we will go ahead and merge.  

Thanks

Tiduster:
> **Regarding changing the user to scality, it will pose an issue for someone who was already running the container as root**
Absolutely

> **We will add a comment in our docs**
I write a proposition, if you are interested.

Regards

DOCKER.md
```diff
@@ -148,3 +149,26 @@ docker run -v $(pwd)/config.json:/usr/src/app/config.json -p 8000:8000 -d scalit
...
+You can change that by modifing the dockerfile and specifying a user before the entrypoint.
+
+The user needs to exist within the container,
+and own the folder **/usr/src/app** for Scality S3 Server to run properly.
+
+For instance, you can modify these lines in the dockerfile:
+
+```shell
+...
+&& groupadd -r -g 1666 scality \
```
nicolas2bert: Why not starting with user id and group id 1000?

nicolas2bert:
@Tiduster please, 
- squash your 2 commits and 
- sign your commit (we have a signing requirement section in our guidelines: https://github.com/scality/Guidelines/blob/master/CONTRIBUTING.md#committing-guidelines), 
- then rebase your branch from master 
- we will merge your PR.  

We really appreciate your contribution.

Regards.

DOCKER.md
Tiduster: I could be wrong, but I think that the node user already use this number.

I can use any number you like.

nicolas2bert: You are right, let's use 1001 so - @RachedBenMustapha agree?

Tiduster:
uid/gid change, squash, signature and rebase done

Tiduster:
I am sorry, are you waiting for me to do something ?

nicolas2bert:
Can you please rebase it from master (again) and @LaurenSpiegel will review it and merge it?

Thank you again @Tiduster

Tiduster:
done

PR #651, opened 2017-03-21
s3 versioning api [S3C-44]

electrachong:
Pending entries:
- [x] tests: to complete versioning `objectPutACL` and `objectGetACL`
- [x] tests: to do versioning `objectPutCopyPart`

initial clean up progress:
- [x] objectPut
- [x] complete mpu
- [x] objectPutAcl
- [x] objectGetAcl
- [x] objectCopy
- [x] objectCopyPart
(still need to address any comments reviewers have for these)

upcoming:
- [x] objectDelete
- [x] multiObjectDelete
- [x] objectHead
- [x] objectGet
- [x] bucketGet (looks clean, did not refactor)
- [x] bucketDelete (looks clean, did not refactor)


lib/api/multiObjectDelete.js
```diff
@@ -50,18 +53,27 @@ import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
...
         errorXML.push(
         '<Error>',
-        '<Key>', escapeForXML(errorObj.key), '</Key>',
-        '<Code>', errorObj.error.message, '</Code>',
-        '<Message>', errorObj.error.description, '</Message>',
+        '<Key>', escapeForXML(errorObj.entry.key), '</Key>',
+        '<Code>', errorObj.error.message, '</Code>');
+        if (errorObj.entry.versionId) {
+            const version = errorObj.entry.versionId === 'null' ?
+                'null' : escapeForXML(errorObj.entry.versionId);
```
LaurenSpiegel: is this encrypted already?

lib/api/multiObjectDelete.js
```diff
@@ -79,34 +91,94 @@ function _formatXML(quietSetting, errorResults, deleted) {
         return xml.join('');
     }
     const deletedXML = [];
-    deleted.forEach(objKey => {
+    deleted.forEach(version => {
+        // TODO include isDeletingDeleteMarker in the result
```
LaurenSpiegel: ?

lib/api/multiObjectDelete.js
```diff
@@ -79,34 +91,94 @@ function _formatXML(quietSetting, errorResults, deleted) {
...
+        const isDeleteMarker = !!version.result;
```
LaurenSpiegel: how does this work?

lib/api/multiObjectDelete.js
```diff
@@ -79,34 +91,94 @@ function _formatXML(quietSetting, errorResults, deleted) {
...
+        // ]}
+        const objects = [];
+        const itemErrors = [];
+        for (let i = 0; i < json.Object.length; i++) {
+            const item = json.Object[i];
+            if (!item.Key) {
+                return next(errors.MalformedXML);
+            }
+            const object = { key: item.Key[0] };
+            // TODO check aws behaviour, maybe returning InvalidArgument
```
LaurenSpiegel: result?

lib/api/multiObjectDelete.js
```diff
@@ -79,34 +91,94 @@ function _formatXML(quietSetting, errorResults, deleted) {
...
+        }
+        deletedXML.push('</Deleted>');
     });
     xml[2] = deletedXML.join('');
     return xml.join('');
 }
 
 function _parseXml(xmlToParse, next) {
     return parseString(xmlToParse, (err, result) => {
+        let itemError = null;
```
LaurenSpiegel: why defining out here?

lib/api/multiObjectDelete.js
```diff
@@ -128,48 +200,93 @@ export function getObjMetadataAndDelete(bucketName, quietSetting,
 
     // doing 5 requests at a time. note that the data wrapper
     // will do 5 parallel requests to data backend to delete parts
-    return async.forEachLimit(inPlay, 5, (key, moveOn) => {
-        metadata.getObjectMD(bucketName, key, {}, log, (err, objMD) => {
+    return async.forEachLimit(inPlay, 5, (entry, moveOn) => {
+        const opts = {
+            versionId: (entry.versionId === 'null' ? '' : entry.versionId),
```
LaurenSpiegel: why this translation?

lib/api/multiObjectDelete.js
```diff
@@ -128,48 +200,93 @@ export function getObjMetadataAndDelete(bucketName, quietSetting,
...
-                        log.error('error deleting object', { error: err, key });
-                        errorResults.push({
-                            key,
-                            error: err,
-                        });
+            objMD = obj;
+            let deleted = false;
+            return async.waterfall([
+                callback => {
+                    // get the latest version or a version other than 'null'
```
LaurenSpiegel: please clarify the objective of these sequence of checks and what each means

lib/api/multiObjectDelete.js
```diff
@@ -128,48 +200,93 @@ export function getObjMetadataAndDelete(bucketName, quietSetting,
...
+                (options, callback) => {
+                    if (options && options.deleteData) {
+                        deleted = true;
+                        return services.deleteObject(bucketName, objMD,
+                                entry.key, options, log, callback);
                     }
-                    numOfObjects++;
-                    successfullyDeleted.push(key);
+                    request.isDeleteMarker = true; // eslint-disable-line
+                    // TODO need authInfo and canonicalID
```
LaurenSpiegel: ?

lib/api/multiObjectDelete.js
```diff
@@ -128,48 +200,93 @@ export function getObjMetadataAndDelete(bucketName, quietSetting,
...
+                    return createAndStoreObject(bucketName, bucket, entry.key,
```
LaurenSpiegel: add comment as to why this is being done

lib/api/multiObjectDelete.js
```diff
@@ -205,17 +322,28 @@ function multiObjectDelete(authInfo, request, log, callback) {
...
     return async.waterfall([
         function parseXML(next) {
-            return _parseXml(request.post, (err, quietSetting, objects) => {
-                if (err || objects.length < 1 || objects.length > 1000) {
-                    return next(errors.MalformedXML);
-                }
-                return next(null, quietSetting, objects);
-            });
+            return _parseXml(request.post,
+                (err, quietSetting, objects, itemErrors) => {
```
LaurenSpiegel: add comment as to what itemErrors are - improperly formed version ID's in requests to delete specific versions of objects?

lib/api/multiObjectDelete.js
```diff
@@ -246,16 +374,16 @@ function multiObjectDelete(authInfo, request, log, callback) {
                     signatureAge: authParams.params.data.signatureAge,
                 },
                 parameterize: {
-                    specificResource: objects,
+                    specificResource: objects.map(entry => entry.key),
```
LaurenSpiegel: if there is a version ID it should be sent so it can be checked in a policy check that includes a version id condition.

ttvi:
@ironman-machine try
SCALITY_METADATA_BRANCH="ft/vsp"
SCALITY_ARSENAL_BRANCH="ft/vsp"
SCALITY_INTEGRATION_BRANCH="fix/awsversion"
DEFAULT_BRANCH="master"

ttvi:
Apparently aws-sdk version 2.2.28 does not support versioning ACL
operations (error "UnexpectedParameter: Unexpected key 'VersionId'
found in params"), upgrading integration's aws-sdk version to 2.28.0
solves the issue but this seems to make the existing tests for bucket
cors fail. Put on hold until external issues are fixed.

constants.js
```diff
@@ -68,23 +68,22 @@ export default {
...
-        'notification',
-        'policy',
-        'replication',
-        'requestPayment',
-        'restore',
-        'tagging',
-        'torrent',
-        'versions',
-    ],
+    unsupportedQueries: {
```
electrachong: We have a commit that gathers all the unrelated fixes and optimizations prior to versioning (currently 4c7ee0d , named `clean: some optimization and dev dependencies`) - I moved this change to there.

lib/api/api.js
```diff
@@ -40,6 +41,8 @@ import websiteGet from './websiteGet';
 import websiteHead from './websiteHead';
 import writeContinue from '../utilities/writeContinue';
 
+const VID = versioning.VersionID;
```
electrachong: see: https://github.com/scality/S3/pull/651/commits/70634bd1e4fe8ef848c7b5bf7287e5acc2afb259

lib/api/api.js
```diff
@@ -67,6 +72,17 @@ const api = {
             // Pull the source bucket and source object separated by /
             sourceBucket = source.slice(0, slashSeparator);
             sourceObject = source.slice(slashSeparator + 1);
+            sourceVersionId = query ?
+                querystring.parse(query).versionId : undefined;
+            sourceVersionId = sourceVersionId || undefined;
```
electrachong: see: https://github.com/scality/S3/pull/651/commits/70634bd1e4fe8ef848c7b5bf7287e5acc2afb259

electrachong:
@ironman-machine
S3_INTEGRATION_BRANCH=ft/awsversion
S3_METADATA_BRANCH=ft/vsp
S3_ARSENAL_BRANCH=ft/vsp

electrachong:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/awsversion
SCALITY_METADATA_BRANCH=ft/vsp
SCALITY_ARSENAL_BRANCH=ft/vsp

electrachong:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=fix/awsversion
SCALITY_METADATA_BRANCH=ft/vsp
SCALITY_ARSENAL_BRANCH=ft/vsp

lib/api/api.js
```diff
@@ -67,6 +72,17 @@ const api = {
...
+            if (sourceVersionId && sourceVersionId !== 'null') {
+                try {
+                    sourceVersionId = VID.decrypt(sourceVersionId);
+                } catch (exception) {
+                    return callback(errors.InvalidArgument
+                        .customizeDescription('Invalid version id specified'));
+                }
+            }
```
electrachong: It looks like we'll have to update Arsenal to update the request contexts.
But I also see we already pass the request headers, which contain the version id, to Arsenal. Do we still want to add another parameter?
Do we currently use these condition keys, or we're just adding them to request contexts / Arsenal for future use?

LaurenSpiegel: I don't think you have to update the requestContexts for version id since we already parse it from the headers.  @nicolas2bert is adding the additional actions to arsenal for versioning (for instance, s3:GetObjectVersion).  What i'm getting at is that when you build the request context for the "get" part of the object copy, you have to send in the version id on the source since that would not be in the regular place in the header -- it would be in the source header.

electrachong:
(end to end passed)

lib/api/api.js
electrachong: That's tricky. Do you mean you want to send the copy source version id in the regular (`x-amz-version-id`) place in the headers? Something like this in lib/api/apiUtils/authorization/prepareRequestContexts.js:

```
export default function prepareRequestContexts(apiMethod, request, sourceBucket,
    sourceObject, sourceVersionId) {
[...]
    if (apiMethod === 'objectCopy' || apiMethod === 'objectPutCopyPart') {
       const requestHeaders = Object.assign{{}, request.headers, { 'x-amz-version-id': sourceVersionId });
        const getRequestContext = new RequestContext(request.headers,
            request.query, sourceBucket, sourceObject,
            request.socket.remoteAddress, request.connection.encrypted,
            'objectGet', 's3');
[...]
```

nicolas2bert: @electrachong @LaurenSpiegel - Please look at my 2 small changes in S3 and Arsenal - https://github.com/scality/S3/commit/a829fcfa3fdcc2cc580576d2cd88e9f466522404
https://github.com/scality/Arsenal/commit/d88282075161fd3d05368a0208e08a93a42e7405

lib/api/objectPut.js
```diff
@@ -211,74 +194,50 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
...
+            },
+            (cipherBundle, callback) => createAndStoreObject(bucketName,
+                bucket, objectKey, objMD, authInfo, canonicalID, cipherBundle,
+                request, streamingV4Params, log, callback),
+        ], (err, res) => {
+            if (err) {
+                return callback(err, null, corsHeaders);
+            }
+            const newByteLength = request.parsedContentLength;
+            const oldByteLength = objMD ? objMD['content-length'] : null;
```
bennettbuchanan: If it's a new object, Utapi expects oldByteLength to be null. That means we need to check that `objMD['content-length']` is not `undefined`. See https://github.com/scality/S3/blob/rel/6.4/lib/api/objectPut.js#L79

lib/api/objectDelete.js
```diff
@@ -24,48 +28,124 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                method: 'metadataValidateAuthorization' });
+        } else if (deleteOptions === undefined) {
+            // TODO metric for delete marker
+            if (res.versionId) {
+                corsHeaders['x-amz-delete-marker'] = true;
+                corsHeaders['x-amz-version-id'] = res.versionId === 'null' ?
+                    res.versionId : VID.encrypt(res.versionId);
+            }
+            pushMetric('putObject', log, { authInfo, bucket: bucketName,
+                newByteLength: 0, oldByteLength: 0 });
```
bennettbuchanan: I'm not sure I follow why we increment the putObject operation.

lib/api/apiUtils/bucket/bucketDeletion.js
```diff
@@ -111,15 +111,19 @@ export function deleteBucket(bucketMD, bucketName, canonicalID, log, cb) {
 
     return async.waterfall([
         function checkForObjectsStep(next) {
-            return metadata.listObject(bucketName, { maxKeys: 1 }, log,
-                (err, objectsListRes) => {
+            const params = { maxKeys: 1, listingType: 'DelimiterVersions' };
```
ttvi: No. We're using the `DelimiterVersions` extension and `metadata` here to do low-level listing. This will always result in a list of versions and delete markers, regardless of the versioning configuration of the bucket, which is S3's notation.

ttvi:
I don't like the "review" mode of github. Adding individual comments is better for following and answering from the conversation page as it connects questions/answers/comments in a concise manner.

circle.yml
```diff
@@ -96,7 +96,7 @@ test:
     - S3BACKEND=file S3VAULT=mem MPU_TESTING=yes npm start
             > $CIRCLE_ARTIFACTS/server_file_awssdk.txt
         & bash wait_for_local_port.bash 8000 40
-        && npm run ft_awssdk
+        && VERSIONING=no npm run ft_awssdk
```
ttvi: Commented.

lib/api/multiObjectDelete.js
alexandre-merle: yes :)

lib/api/multiObjectDelete.js
alexandre-merle: to have the error and be able to do the if else at the end

lib/api/multiObjectDelete.js
alexandre-merle: its because metadata does not know about the null version, so we have to get the master version to know where the null version is :)

lib/api/multiObjectDelete.js
alexandre-merle: Hm true, but this need to be handle in a subsequent pull request, we already have huge enought pull requests :)

lib/api/objectDelete.js
```diff
@@ -24,48 +28,124 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+            const versioning = bucketMD.getVersioningConfiguration();
+            // TODO adhere to the behaviour of AWS: create a delete marker
+            // even if the deleting object does not exist
             if (!objMD) {
-                return cb(errors.NoSuchKey, corsHeaders);
+                if (versioning) {
+                    if (reqVersionId) {
+                        corsHeaders['x-amz-version-id'] =
+                            VID.encrypt(reqVersionId);
+                        return cb(null, corsHeaders, null);
```
alexandre-merle: nope, removing this :)

lib/api/objectDelete.js
```diff
@@ -24,48 +28,124 @@ export default function objectDelete(authInfo, request, log, cb) {
...
-            if (objMD['content-length']) {
-                log.end().addDefaultFields({
-                    contentLength: objMD['content-length'],
-                });
+            // putting a new delete marker
+            if (bucketMD.hasDeletedFlag() &&
+                    canonicalID !== bucketMD.getOwner()) {
+                log.trace('deleted flag on bucket and request ' +
+                        'from non-owner account');
+                return callback(errors.NoSuchBucket);
```
alexandre-merle: Since the bucket have deleting flag, and the requester is not the bucket owner, does that make sense to return AccessDenied ? since the bucket is supposed to be deleted already. We copy this logic from the objectPut :)

lib/api/objectDelete.js
```diff
@@ -24,48 +28,124 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+            }
+            request.isDeleteMarker = true; // eslint-disable-line
+            return createAndStoreObject(bucketName, bucketMD,
+                objectKey, objectMD, authInfo, canonicalID, null, request,
+                null, log, callback);
+        },
+    ], (err, res) => {
+        if (err) {
+            log.debug('error processing request', { error: err,
+                method: 'metadataValidateAuthorization' });
```
alexandre-merle: mistake, should be objectDelete :)

lib/api/objectDelete.js
alexandre-merle: because we put a new delete marker :)

alexandre-merle: can you check with @rahulreddy which behavior we should have here, putting the old length ?

lib/api/multiObjectDelete.js
alexandre-merle: done :)

lib/api/multiObjectDelete.js
alexandre-merle: done :)

lib/api/multiObjectDelete.js
alexandre-merle: removed, old todo :)

lib/api/multiObjectDelete.js
alexandre-merle: i clarify a bit, let me know if its okay :)

ttvi:
@electrachong @LaurenSpiegel I saw a lot of cleanups. Since this is a large enough PR and some other in MetaData and Arsenal are waiting for this to be merge, could you hold this refactoring off, finish the feature, and make another cleanup PR just after this.

lib/api/apiUtils/bucket/bucketDeletion.js
LaurenSpiegel: so, this will return all versions even the null version if versioning has not been enabled?  please add comment in the code

lib/api/objectPut.js
```diff
@@ -58,31 +64,24 @@ function _storeIt(bucket, objectKey, objMD, authInfo, canonicalID, cipherBundle,
...
         cacheControl: request.headers['cache-control'],
         contentDisposition: request.headers['content-disposition'],
-        contentEncoding:
-            removeAWSChunked(request.headers['content-encoding']),
-        expires: request.headers.expires,
-        headers,
-        log,
+        contentEncoding: removeAWSChunked(request.headers['content-encoding']),
+        expires: request.headers.expires, headers, log,
+        isDeleteMarker: request.isDeleteMarker,
```
LaurenSpiegel: why on request?

lib/api/objectPut.js
```diff
@@ -91,77 +90,68 @@ function _storeIt(bucket, objectKey, objMD, authInfo, canonicalID, cipherBundle,
...
-                if (err) {
-                    return callback(err);
-                }
-                pushMetric('putObject', log, {
-                    authInfo,
-                    bucket: bucketName,
-                    newByteLength: size,
-                    oldByteLength: prevContentLen,
+    // null - new object
+    // 0 or > 0 - existing object with content-length 0 or greater than 0
```
LaurenSpiegel: what do these comments mean?  correct place?

lib/api/objectPut.js
```diff
@@ -91,77 +90,68 @@ function _storeIt(bucket, objectKey, objMD, authInfo, canonicalID, cipherBundle,
...
+            const dataGetInfoArr = [{ key, size, start: 0, dataStoreName }];
+            if (cipherBundle) {
+                dataGetInfoArr[0].cryptoScheme = cipherBundle.cryptoScheme;
+                dataGetInfoArr[0].cipheredDataKey =
+                    cipherBundle.cipheredDataKey;
+            }
+            metadataStoreParams.contentMD5 = calculatedHash;
+            return next(null, dataGetInfoArr);
+        },
+        function getVersioningInfo(infoArr, next) {
```
LaurenSpiegel: why not just keep passing dataGetInfoArr along until used?

lib/api/objectPut.js
```diff
@@ -91,77 +90,68 @@ function _storeIt(bucket, objectKey, objMD, authInfo, canonicalID, cipherBundle,
...
+            dataGetInfoArr = infoArr;
+            return services.versioningPreprocessing(bucketName, bucketMD,
+                metadataStoreParams.objectKey, objMD, reqVersionId, log, next);
```
LaurenSpiegel: instead of doing next here, let's have a cb so can pass dataGetInfoArr and so it's clearer what is getting sent to the next step.

lib/api/objectPut.js
```diff
@@ -91,77 +90,68 @@ function _storeIt(bucket, objectKey, objMD, authInfo, canonicalID, cipherBundle,
...
+        },
+        function createDeleteArray(options, next) {
+            if (!options.deleteNullVersionData) {
+                return next(null, options);
+            }
```
LaurenSpiegel: please add comment stating that if options.deleteNullVersionData is true, we need to get the location info of the null version.  this will only apply when versioning has not been enabled or has been disabled.

lib/api/objectPut.js
```diff
@@ -190,95 +179,71 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
...
+                            serverSideEncryption, log, next);
+                }
+                return next(null, null);
+            },
+            function objectCreateAndStore(cipherBundle, next) {
+                return createAndStoreObject(bucketName,
+                bucket, objectKey, objMD, authInfo, canonicalID, cipherBundle,
+                request, streamingV4Params, log, next);
+            },
+        ], (err, res) => {
```
LaurenSpiegel: can we call this resInfo or something similar since it's not actually the request object?

lib/api/objectPut.js
```diff
@@ -190,95 +179,71 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
...
+            if (err) {
+                return callback(err, responseHeaders);
+            }
+            const newByteLength = request.parsedContentLength;
+            // Utapi expects null or a number for oldByteLength
+            const oldByteLength = objMD && objMD['content-length']
```
LaurenSpiegel: shouldn't this be the nullObjMd now?

lib/api/objectPut.js
```diff
@@ -13,23 +14,29 @@ import { logger } from '../utilities/logger';
...
+import metadata from '../metadata/wrapper';
+
+const versionIdUtils = versioning.VersionID;
+
 
 function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
     metadataStoreParams, dataToDelete, deleteLog, callback) {
     services.metadataStoreObject(bucketName, dataGetInfo,
-        cipherBundle, metadataStoreParams, (err, contentMD5) => {
+        cipherBundle, metadataStoreParams, (err, res) => {
```
LaurenSpiegel: let's call this something more descriptive than res especially since it's not the actual response object

lib/api/multiObjectDelete.js
LaurenSpiegel: then, please create an issue in jira so it doesn't get forgotten

lib/api/completeMultipartUpload.js
```diff
@@ -182,7 +184,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             // they are returning the MD5 of the parts' MD5s so that is
             // what we have done here. We:
             // 1) concatenate the hex version of the
-            // individual ETags
+            // invidual ETags
```
LaurenSpiegel: restore prior spelling

lib/api/completeMultipartUpload.js
```diff
@@ -363,66 +365,88 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                dataLocations, pseudoCipherBundle, metaStoreParams,
+                (err, res) => {
                     if (err) {
                         return next(err, destinationBucket);
                     }
-                    if (objMD && objMD.location) {
+                    const generatedVersionId = res ? res.versionId : undefined;
+                    // in cases where completing mpu overwrites a previous
+                    // null version or versioning is not enabled, need to delete
+                    // pre-existing data
```
LaurenSpiegel: I think this should be in cases where versioning has never been unable or is suspended

lib/api/objectPut.js
electrachong: is `result` OK?

lib/api/objectPut.js
electrachong: changed createAndStoreObject to accept a new parameter 'isDeleteMarker' so we don't have to modify the request object in delete / multiobjectdelete apis

lib/api/objectPut.js
electrachong: they were left by accident in the refactor, they are supposed to explain the pushmetrics oldByteLength. I moved them to the right place

lib/api/objectPut.js
electrachong: OK, changed

lib/api/objectPut.js
electrachong: OK

lib/api/objectPut.js
electrachong: OK
Actually when we delete the null version data is a little more complicated, so hope the comment helps explain: https://github.com/scality/S3/pull/651/commits/7456e668e263341e95e905bed998ced16ac8c78a#diff-e45defa8c8c53634533cf165cc9232ceR123

Normally as long as the null version is the latest version, it can be deleted just by using the master version's object md location property. We only have to get the null version's location when there is a null version that is not the latest version (can happen when we store the null version data when re-enabling versioning or enabling it for the first time).

lib/api/objectPut.js
electrachong: as discussed, since the object or version we would have overwritten (if we overwrite any) is the master version, this is the correct objMD to use to calculate the old byte length here.  I added a comment to help

lib/services.js
```diff
@@ -737,4 +769,131 @@ export default {
...
+                return callback(null, options);
+            }
+            // master is versioned and is not a null version
+            const nullVersionId = objMD.nullVersionId;
+            if (vstat === 'Suspended') {
+                // versioning is suspended, overwrite the existing version
+                options.versionId = '';
+                options.isNull = true;
+                options.deleteNullVersionData = true;
+                if (nullVersionId === undefined) {
```
electrachong: @ttvi FYI: I think we need to bump this if condition up one line (we should not attempt to delete null version data if there is no null version)

lib/api/completeMultipartUpload.js
electrachong: fixed

lib/api/completeMultipartUpload.js
electrachong: clarified

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
...
+            }
+            // versioning is enabled, create a new version
+            options.versioning = true;
+            if (objMD) {
+                // store master version in a new key
+                const versionId = mstIsNull ? mstVersionId : VID_INF;
+                objMD.versionId = versionId; // eslint-disable-line
+                objMD.isNull = true; // eslint-disable-line
+                options.nullVersionId = versionId;
+                return metadata.putObjectMD(bucketName, objectKey, objMD,
```
LaurenSpiegel: please add a log line for the error in the callback so we know that the error is from storing the null mater version in a new key in metadata.

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
...
+            // versioning is suspended, overwrite the existing version
+            options.versionId = '';
+            options.isNull = true;
+            if (nullVersionId === undefined) {
+                return callback(null, options);
+            }
+            options.deleteNullVersionData = true;
+            return metadata.deleteObjectMD(bucketName, objectKey,
+                { versionId: nullVersionId }, log,
+                err => callback(err, options));
```
LaurenSpiegel: please add log line to error in callback so we know that the error is from deleting an old null version

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
...
+                options.nullVersionId = versionId;
+                return metadata.putObjectMD(bucketName, objectKey, objMD,
+                        { versionId }, log, err => callback(err, options));
+            }
+            return callback(null, options);
+        }
+        // master is versioned and is not a null version
+        const nullVersionId = objMD.nullVersionId;
+        if (vstat === 'Suspended') {
+            // versioning is suspended, overwrite the existing version
```
LaurenSpiegel: do you mean overwrite an existing null version?

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
...
+            options.isNull = true;
+            if (nullVersionId === undefined) {
+                return callback(null, options);
+            }
+            options.deleteNullVersionData = true;
+            return metadata.deleteObjectMD(bucketName, objectKey,
+                { versionId: nullVersionId }, log,
+                err => callback(err, options));
+        }
+        // versioning is enabled, put the new version
```
LaurenSpiegel: how does this ensure putting of the new version?  Do you mean keep the current nullVersionId saved in master and given api will put new version as new key?

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
...
+        return callback(null, options);
+    }
+    // bucket is versioning enabled
+    if (reqVersionId && reqVersionId !== 'null') {
+        // deleting a specific version
+        options.deleteData = true;
+        options.versionId = reqVersionId;
+        return callback(null, options);
+    }
+    if (reqVersionId) {
```
LaurenSpiegel: would be clearer to say `if(reqVersionId === 'null')`

lib/api/objectCopy.js
```diff
@@ -162,27 +169,29 @@ function objectCopy(authInfo, request, sourceBucket,
                 });
         },
         function checkSourceAuthorization(destBucketMD, destObjMD, next) {
-            return services.metadataValidateAuthorization(valGetParams,
+            return metadataValidateBucketAndObj(valGetParams, log,
                 (err, sourceBucketMD, sourceObjMD) => {
                     if (err) {
-                        log.debug('error validating get part of request',
```
LaurenSpiegel: please restore log

lib/api/objectCopy.js
```diff
@@ -298,37 +306,62 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                        destObjMD['content-length'] !== undefined) ?
                         destObjMD['content-length'] : null;
-                    return next(null, destBucketMD, storeMetadataParams,
+                    return next(null, result, destBucketMD, storeMetadataParams,
                         serverSideEncryption, sourceObjSize, destObjPrevSize);
                 });
         },
-    ], (err, destBucketMD, storeMetadataParams, serverSideEncryption,
-        sourceObjSize, destObjPrevSize) => {
+    ], (err, storingNewDataResult, destBucketMD, storeMetadataParams,
```
LaurenSpiegel: isn't `storingNewDataResult` really `storingNewMdResults`?

lib/api/objectPutCopyPart.js
```diff
@@ -22,21 +26,24 @@ import validateHeaders from '../utilities/validateHeaders';
...
  */
 export default
 function objectPutCopyPart(authInfo, request, sourceBucket,
-    sourceObject, log, callback) {
+    sourceObject, reqVersionId, log, callback) {
     log.debug('processing request', { method: 'objectPutCopyPart' });
     const destBucketName = request.bucketName;
     const destObjectKey = request.objectKey;
     const mpuBucketName = `${constants.mpuBucketPrefix}${destBucketName}`;
+    let sourceVerId = undefined;
```
LaurenSpiegel: please don't define this out here.  instead pass it along in the waterfall once you have the value.

lib/api/objectCopy.js
```diff
@@ -162,27 +169,29 @@ function objectCopy(authInfo, request, sourceBucket,
...
         function checkSourceAuthorization(destBucketMD, destObjMD, next) {
-            return services.metadataValidateAuthorization(valGetParams,
+            return metadataValidateBucketAndObj(valGetParams, log,
                 (err, sourceBucketMD, sourceObjMD) => {
                     if (err) {
-                        log.debug('error validating get part of request',
-                        { error: err });
-                        return next(err, destBucketMD);
+                        return next(err, null, destBucketMD);
                     }
```
LaurenSpiegel: do you have to check if source version is a delete marker?

lib/api/objectGetACL.js
```diff
@@ -60,20 +75,24 @@ export default function objectGetACL(authInfo, request, log, callback) {
...
-                log.trace('error processing request',
-                    { method: 'objectGetACL', error: err });
-                return callback(errors.NoSuchKey, null, corsHeaders);
-            }
+    return async.waterfall([
+        function validateBucketAndObj(next) {
+            return metadataValidateBucketAndObj(metadataValParams, log,
+                (err, bucket, objectMD) => {
+                    if (err) {
+                        return next(err);
```
rahulreddy: Why were the logs removed?

lib/api/objectGetACL.js
```diff
@@ -154,7 +165,7 @@ export default function objectGetACL(authInfo, request, log, callback) {
                 if (err) {
                     log.trace('error processing request',
                         { method: 'objectGetACL', error: err });
-                    return callback(err, null, corsHeaders);
+                    return callback(err, bucket);
```
rahulreddy: This calls the main callback directly, shouldn't this be calling next with appropriate callback params i.e.`next(err, bucket, xml, resVersionId)`

lib/api/objectGetACL.js
```diff
@@ -1,11 +1,15 @@
 import { errors } from 'arsenal';
+import async from 'async';
```
rahulreddy: Please move this above arsenal. We generally follow the convention core node.js modules > third party modules > scality modules > local modules.

lib/api/objectGetACL.js
electrachong: good catch, I think I just missed changing the callback name to next (since it's an error case, we just return the error and the bucket for cors headers)
I will update when I get to addressing comments

lib/api/objectDelete.js
electrachong: @alexandre-merle `bucketShield` (which is called by `metadataValidateAuthorization` as well as the newer `metadataValidateBucketAndObj` I will use) should take care of this check, no need to include here--
I will update with this in consideration

https://github.com/scality/S3/blob/master/lib/api/apiUtils/bucket/bucketShield.js (when bucketShield returns true, `metadataValidateAuthorization` and `metadataValidateBucketAndObj` will return `NoSuchBucket`)

lib/api/completeMultipartUpload.js
```diff
@@ -363,66 +366,88 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                    algorithm: destBucket.getSseAlgorithm(),
+                    masterKeyId: destBucket.getSseMasterKeyId(),
                 };
             }
-            services.metadataStoreObject(destinationBucket.getName(),
-                dataLocations, pseudoCipherBundle, metaStoreParams, err => {
+            return versioningPreprocessing(bucketName,
+                destBucket, objectKey, objMD, null, log, (err, options) => {
+                    if (err) {
+                        return next(err, destBucket);
```
rahulreddy: There should be an error log here.

tests/functional/aws-node-sdk/test/object/completeMPU.js
```diff
@@ -78,9 +50,115 @@ describe('Complete MPU', () => {
...
+                }
+                return s3.deleteBucket({ Bucket: bucket }, done);
+            });
+        });
+
+        describe('on bucket without versioning configuration', () => {
+            let uploadId;
+            let firstEtag;
+
+            beforeEach(() => s3.createBucketAsync({ Bucket: bucket })
```
rahulreddy: The setup in beforeEach is more or less the same across the tests. Can you extract this into a utility and use it across the tests?

tests/functional/aws-node-sdk/test/versioning/objectCopy.js
```diff
@@ -0,0 +1,1006 @@
...
+        let s3;
+        let etag;
+        let etagTrim;
+        let lastModified;
+        let versionId = undefined;
+        let copySource = undefined;
+
+        before(() => {
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
```
rahulreddy: It's odd to see cleanup of resources in `before()`, it should be handled in `after()`

lib/api/objectDelete.js
```diff
@@ -24,48 +30,133 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+            (err, bucketMD, objMD) => {
+                if (err) {
+                    return (err, bucketMD);
+                }
+                const versioningCfg = bucketMD.getVersioningConfiguration();
+                if (!objMD) {
+                    if (!versioningCfg) {
+                        return next(errors.NoSuchKey, bucketMD);
+                    }
+                    // TODO: currently we return do not create a delete
```
electrachong: @alexandre-merle can you confirm the comment is accurate?

lib/api/objectDelete.js
```diff
@@ -24,48 +30,133 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                    // marker if the version specified for deletion does
+                    // not exist. To adhere to AWS behavior, we should
+                    // update to create a delete marker anyway.
+                    if (reqVersionId) {
+                        const corsHeaders =
+                            collectCorsHeaders(request.headers.origin,
+                                request.method, bucketMD);
+                        return cb(null, corsHeaders);
+                    }
+                    // To adhere to AWS behavior, create a delete marker even
```
electrachong: @alexandre-merle confirm this is what we do?

lib/api/objectDelete.js
```diff
@@ -24,48 +30,133 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+        });
+        return cb(decryptVidResult);
+    }
+    const reqVersionId = decryptVidResult;
+
     const valParams = {
         authInfo,
         bucketName,
         objectKey,
+        versionId: reqVersionId || undefined,
```
electrachong: fixed: https://github.com/scality/S3/pull/651/commits/2c6834bf6d9246cc0f2f8b8838ed8bf4129a5afd
(will delete comment after squash)

lib/api/objectDelete.js
```diff
@@ -24,48 +30,133 @@ export default function objectDelete(authInfo, request, log, cb) {
...
-                    error: err,
-                    method: 'metadataValidateAuthorization',
+
+    const canonicalID = authInfo.getCanonicalID();
+    return async.waterfall([
+        function validateBucketAndObj(next) {
+            return metadataValidateBucketAndObj(valParams, log,
+            (err, bucketMD, objMD) => {
+                if (err) {
+                    return (err, bucketMD);
```
electrachong: fixed: https://github.com/scality/S3/pull/651/commits/2c6834bf6d9246cc0f2f8b8838ed8bf4129a5afd
(will delete comment after squash)

lib/api/multiObjectDelete.js
```diff
@@ -50,18 +53,27 @@ import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
 * @param {boolean} quietSetting - true if xml should just include error list
 * and false if should include deleted list and error list
 * @param {object []} errorResults - list of error result objects with each
-* object containing -- key: objectName, error: arsenal error
-* @param {string []} deleted - list of object keys deleted
+* object containing -- entry: { key, versionId }, error: arsenal error
+* @param {object []} deleted - list of object deleted, an object has the format
+*     object: { entry, result, isDeletingDeleteMarker }
+*     object.entry : above
+*     object.result: stringification of { versionId }
```
electrachong: this is changed since we don't seem to need the information inside result, just whether a delete marker was created.
before, it was the parsed but unencrypted version id (metadata returns as JSON object `{ versionId }`, we parse this in the function we get 'result' from) if a delete marker was created; undefined if an object was deleted

lib/api/multiObjectDelete.js
```diff
@@ -50,18 +53,27 @@ import { isBucketAuthorized } from './apiUtils/authorization/aclChecks';
...
+*     object.isDeletingDeleteMarker: name as comment
```
electrachong: updated

lib/api/multiObjectDelete.js
electrachong: old comment

lib/api/multiObjectDelete.js
electrachong: I changed this to be clearer.

lib/api/objectDelete.js
```diff
@@ -24,48 +30,132 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                    deleteInfo.removeDeleteMarker = true;
+                }
+                return services.deleteObject(bucketName, objectMD, objectKey,
+                    delOptions, log, (err, delResult) => next(err, bucketMD,
+                    objectMD, delResult, deleteInfo));
             }
-            const headerValResult = validateHeaders(objMD, request.headers);
-            if (headerValResult.error) {
-                return cb(headerValResult.error, corsHeaders);
+            // putting a new delete marker
```
LaurenSpiegel: isn't this only if versioning is enabled?  what about non-versioned bucket?

lib/api/objectDelete.js
```diff
@@ -24,48 +30,132 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                resHeaders['x-amz-delete-marker'] = true;
+                resHeaders['x-amz-version-id'] = result.versionId === 'null' ?
+                    result.versionId : versionIdUtils.encrypt(result.versionId);
             }
-            if (objMD['content-length']) {
-                log.end().addDefaultFields({
-                    contentLength: objMD['content-length'],
-                });
+            // TODO metric for delete marker
+            pushMetric('putObject', log, { authInfo, bucket: bucketName,
```
LaurenSpiegel: why putObject?

lib/api/multiObjectDelete.js
LaurenSpiegel: it should be defined within the scope of the for loop

lib/api/multiObjectDelete.js
LaurenSpiegel: I've added this to https://scality.atlassian.net/browse/S3C-184

lib/api/multiObjectDelete.js
```diff
@@ -118,58 +195,74 @@ function _parseXml(xmlToParse, next) {
...
+                        errorResults.push({ entry, error: err });
                         return moveOn();
                     }
-                    if (objMD['content-length']) {
-                        totalContentLengthDeleted +=
-                            objMD['content-length'];
+                    // if particular key does not exist, AWS returns success
+                    // for key so add to successfullyDeleted list and move on
+                    if (err && err.NoSuchKey) {
+                        successfullyDeleted.push({ entry });
```
LaurenSpiegel: There are comments in objectDelete.js that if versioning is enabled and there is a delete request but no object, you actually put a delete marker.  Shouldn't that be the same behavior here?

lib/api/objectHead.js
```diff
@@ -20,38 +23,60 @@ export default function objectHead(authInfo, request, log, callback) {
...
             if (!objMD) {
-                return callback(errors.NoSuchKey, corsHeaders);
+                const err = versionId ? errors.NoSuchVersion : errors.NoSuchKey;
+                return callback(err, corsHeaders);
+            }
+            const verCfg = bucket.getVersioningConfiguration();
+            if (objMD.isDeleteMarker) {
+                const responseHeaders = Object.assign({},
+                    { 'x-amz-delete-marker': true }, corsHeaders);
+                if (!versionId) {
```
LaurenSpiegel: add comment: if general delete request without versionID and the master object is a delete marker, return NoSuchKey

lib/api/objectGet.js
```diff
@@ -20,30 +23,57 @@ function objectGet(authInfo, request, log, callback) {
...
-        log,
     };
 
-    services.metadataValidateAuthorization(mdValParams, (err, bucket,
+    return metadataValidateBucketAndObj(mdValParams, log, (err, bucket,
         objMD) => {
         const corsHeaders = collectCorsHeaders(request.headers.origin,
             request.method, bucket);
         if (err) {
-            log.debug('error processing request', { error: err });
```
LaurenSpiegel: let's keep the log

lib/api/objectGet.js
```diff
@@ -20,30 +23,57 @@ function objectGet(authInfo, request, log, callback) {
...
         if (!objMD) {
-            return callback(errors.NoSuchKey, null, corsHeaders);
+            const err = versionId ? errors.NoSuchVersion : errors.NoSuchKey;
+            return callback(err, null, corsHeaders);
+        }
+        const verCfg = bucket.getVersioningConfiguration();
+        if (objMD.isDeleteMarker) {
+            const responseMetaHeaders = Object.assign({},
+                { 'x-amz-delete-marker': true }, corsHeaders);
+            if (!versionId) {
```
LaurenSpiegel: add comment: if general get request without versionId and master version is a delete marker, return NoSuchKey

lib/api/objectDelete.js
```diff
@@ -23,44 +29,132 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                    return next(null, bucketMD, objMD);
+                }
+                const headerValResult = validateHeaders(objMD,
+                    request.headers);
+                if (headerValResult.error) {
+                    return next(headerValResult.error, bucketMD);
+                }
+                if (objMD['content-length']) {
+                    log.end().addDefaultFields({
+                        contentLength: objMD['content-length'],
```
rahulreddy: `contentLength` as a log field may be confusing, `bytesDeleted` would be more appropriate. Also the check should be  `if (objMD['content-length'] !== undefined) {`

lib/api/objectDelete.js
```diff
@@ -23,44 +29,132 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                }
+                return next(null, bucketMD, objMD);
+            });
+        },
+        function getVersioningInfo(bucketMD, objectMD, next) {
+            return preprocessingVersioningDelete(bucketName,
+                bucketMD, objectKey, objectMD, reqVersionId, log,
+                (err, options) => {
+                    if (err) {
+                        log.debug('err processing versioning info',
```
rahulreddy: This should be error level because at this point we don't expect the request to fail due to user input.

lib/api/objectDelete.js
```diff
@@ -23,44 +29,132 @@ export default function objectDelete(authInfo, request, log, cb) {
...
-                    pushMetric('deleteObject', log, {
-                        authInfo,
-                        bucket: bucketName,
-                        byteLength: objMD['content-length'],
-                        numberOfObjects: 1,
-                    });
-                    return cb(null, corsHeaders);
-                });
-        });
+            pushMetric('deleteObject', log, { authInfo, bucket: bucketName,
```
rahulreddy: Except for unversioned or versioning-suspended buckets, numberOfObjects deleted should be 0.
For versioned buckets, numberOfObjects should be 0 once we set the delete marker.

lib/api/objectDelete.js
rahulreddy: This should be...

1. If the delete marker is being set for the first time
```js
pushMetric('deleteObject', log, { authInfo, bucket: bucketName,
    byteLength: 0, numberOfObjects: 1 } );
```

2. If we are creating a new delete marker and the object already has a delete marker `numberOfObjects` should be 0 (otherwise Utapi will incorrectly report the number of objects in the system).

```js
pushMetric('deleteObject', log, { authInfo, bucket: bucketName, byteLength: 0, numberOfObjects: 0 } );
```

`byteLength` is 0 in both cases since no data is actually deleted from the system.

lib/api/multiObjectDelete.js
```diff
@@ -118,58 +195,74 @@ function _parseXml(xmlToParse, next) {
...
+                        return moveOn();
```
rahulreddy: Calling `moveOn` directly would mean that the waterfall's step's `callback` never gets called, which will block the event loop eventually.

lib/api/multiObjectDelete.js
```diff
@@ -118,58 +195,74 @@ function _parseXml(xmlToParse, next) {
...
+            // for obj deletes, no need to check acl's at object level
+            // (authority is at the bucket level for obj deletes)
+            callback => metadataGetObject(bucketName, entry.key,
+                entry.versionId, log, (err, objMD) => {
+                    // if general error from metadata return error
+                    if (err && !err.NoSuchKey) {
+                        log.error('error getting object MD',
+                                { error: err, key: entry.key });
+                        errorResults.push({ entry, error: err });
                         return moveOn();
```
rahulreddy: This should be callback as well.
Edit: This can fixed in another PR since this commit is not the one introducing it.

lib/api/multiObjectDelete.js
```diff
@@ -118,58 +195,74 @@ function _parseXml(xmlToParse, next) {
...
+        ], (err, objMD, deleteInfo) => {
+            if (err) {
+                log.error('error deleting object', { error: err, entry });
+                errorResults.push({ entry, error: err });
+                return moveOn();
+            }
+            if (deleteInfo.deleted && objMD['content-length']) {
+                totalContentLengthDeleted += objMD['content-length'];
+            }
+            numOfObjects++;
```
rahulreddy: `numOfObjects` which represents number of objects deleted, should be incremented only when a delete marker is created for the first time for the object.

lib/api/multiObjectDelete.js
```diff
@@ -205,17 +298,30 @@ function multiObjectDelete(authInfo, request, log, callback) {
...
+                    }
+                    return next(null, quietSetting, objects, itemErrors);
+                });
         },
-        function checkPolicies(quietSetting, objects, next) {
+        function checkPolicies(quietSetting, objects, itemErrors, next) {
             // track the error results for any keys with
             // an error response
             const errorResults = [];
+            itemErrors.forEach(error => {
```
rahulreddy: Minor: Can you rename this to`item` instead of error to avoid confusion?

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,220 @@
...
+
+/** metadataGetObject - retrieves specified object or version from metadata
+ * @param {string} bucketName - name of bucket
+ * @param {string} objectKey - name of object key
+ * @param {string} [versionId] - version of object to retrieve
+ * @param {RequestLogger} log - request logger
+ * @param {function} cb - callback
+ * @return {undefined} - and call callback with err, bucket md and object md
+ */
+export function metadataGetObject(bucketName, objectKey, versionId, log, cb) {
```
electrachong: This one is now being used by multiObjectDelete api.

electrachong:
@ironman-machine just to check changes so far
SCALITY_METADATA_BRANCH=ft/vsp
SCALITY_ARSENAL_BRANCH=ft/vsp

lib/services.js
```diff
@@ -196,7 +199,7 @@ export default {
         omVal['last-modified'] = lastModifiedDate || new Date().toJSON();
         omVal['content-md5'] = contentMD5;
 
-        omVal['x-amz-server-version-id'] = '';
+        // omVal['x-amz-server-version-id'] = '';
```
LaurenSpiegel: why comment out?  it's pretty useless but does not seem related to object versioning

lib/services.js
```diff
@@ -215,7 +218,7 @@ export default {
         }
 
         // simple/no version. will expand once object versioning is introduced
-        omVal['x-amz-version-id'] = 'null';
+        // omVal['x-amz-version-id'] = 'null';
```
LaurenSpiegel: why not just remove along with the comment above?

lib/services.js
```diff
@@ -224,6 +227,20 @@ export default {
             READ_ACP: [],
         };
 
+        const options = {};
```
LaurenSpiegel: please add comments in the next 10 or so lines as to why these settings are being set to the options and the omVal

lib/services.js
```diff
@@ -245,79 +262,80 @@ export default {
...
+                    resourceType: 'object',
+                    acl: omVal.acl,
+                    log,
+                };
+                log.trace('parsing acl from headers');
+                acl.parseAclFromHeaders(parseAclParams, (err, parsedACL) => {
                     if (err) {
-                        log.error('error from metadata', { error: err });
-                        return cb(err);
+                        log.warn('error parsing acl', { error: err });
```
LaurenSpiegel: why warn?

lib/services.js
```diff
@@ -245,79 +262,80 @@ export default {
...
+            },
+            callback => metadata.putObjectMD(bucketName, objectKey, omVal,
+                    options, log, callback),
+        ], (err, data) => {
+            if (err) {
+                log.error('error from metadata', { error: err });
+                return cb(err);
+            }
+            log.trace('object successfully stored in metadata');
+            // for versioning only, other features need to process their data
```
LaurenSpiegel: perhaps: // if versioning is enabled, data will be returned from metadata containing a versionId which some APIs will need sent back to it

lib/metadata/in_memory/backend.js
```diff
@@ -64,12 +80,36 @@ const metastore = {
...
+                    objVal.versionId = versionId; // eslint-disable-line
+                    metadata.keyMaps.get(bucketName).set(objName, objVal);
+                    return cb(null, `{"versionId":"${objVal.versionId}"}`);
+                } else if (params && params.versionId) {
+                    objVal.versionId = params.versionId; // eslint-disable-line
+                    const mst = metadata.keyMaps.get(bucketName).get(objName);
+                    if (mst && mst.versionId === params.versionId) {
+                        metadata.keyMaps.get(bucketName).set(objName, objVal);
+                    }
+                    // eslint-disable-next-line
```
ttvi: Please avoid asking obvious questions, unless you've really tried but still couldn't understand.

lib/metadata/in_memory/backend.js
```diff
@@ -118,6 +166,29 @@ const metastore = {
                 if (!metadata.keyMaps.get(bucketName).has(objName)) {
                     return cb(errors.NoSuchKey);
                 }
+                if (params && params.versionId) {
+                    const baseKey = inc(formatVersionKey(objName, ''));
```
ttvi: @LaurenSpiegel please stop the demand for comments. A piece of code, unlike the specifications for a function or an API, speaks for itself and requires no comments for people to understand it, unless it is not self containing.

ttvi:
Guys, I will repeat this: please stop all ongoing and demands for refactoring, let's save it for later; this is a huge PR, a lot of other things are waiting for this to be merged, not mentioning refactoring may create new bugs.

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+            log.debug('metadata getbucket failed', { error: err });
+            return cb(err);
+        }
+        if (bucketShield(bucket, requestType)) {
+            log.debug('bucket is shielded from request', {
+                requestType,
+                method: 'metadataGetBucketAndObject',
+            });
+            // still return bucket for cors headers
+            return cb(errors.NoSuchBucket, bucket);
```
electrachong: you're right, will remove

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+ * @param {function} callback - callback
+ * @return {undefined} - and call callback with params err, bucket md
+ */
+export function metadataValidateBucket(params, log, callback) {
+    const { authInfo, bucketName, requestType } = params;
+    const canonicalID = authInfo.getCanonicalID();
+    return metadataGetBucket(requestType, bucketName, log, (err, bucket) => {
+        if (err) {
+            // still return bucket for cors headers
+            return callback(err, bucket);
```
electrachong: removed, it was from metadataGetBucket returning the bucket if bucket was shielded (which we will no longer due)

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+ * @param {string} objectKey - name of object key
+ * @param {RequestLogger} log - request logger
+ * @param {function} cb - callback
+ * @return {undefined} - and call callback with params err, objMD of null ver
+ */
+export function getNullVersion(objMD, bucketName, objectKey, log, cb) {
+    const options = {};
+    if (objMD.isNull || !objMD.versionId) {
+        // null version is current version
+        log.debug('found null version'); // don't want to log internal id?
```
electrachong: the log doesn't currently log the decrypted version id so I will keep it
but in general we can consider whether it would be useful to log the version id, and if so, only when it is encrypted or the 'internal' (decrypted) vid as well? it's something we can save for later; I'll remove the comment

lib/metadata/in_memory/backend.js
ttvi: I'm sorry for all the strong words. My opinion still holds.

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+        return cb(null, objMD);
+    }
+    if (objMD.nullVersionId) {
+        // the latest version is not the null version, but null version exists
+        log.debug('null version exists, get the null version');
+        options.versionId = objMD.nullVersionId;
+        return metadata.getObjectMD(bucketName, objectKey, options, log, cb);
+    }
+    log.debug('could not find a null version');
```
electrachong: null versions are created a) when putting an object before versioning is configured; b) putting an object when versioning is suspended. someone may request getting (or other actions which allow targeting a specific version) the 'null' version when neither of these situations have occurred and in that case there will be no null version in metadata to retrieve. It's similar to someone requesting any other version id that does not exist.

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+    return cb();
+}
+
+/** metadataGetBucketAndObject - retrieves bucket and specified version
+ * NOTE: This function only returns master version if 'null' version is
+ * of object is specified. Getting the actual null version must be a manual
```
electrachong: if trying to get the null version instead of any other version using this function, specified in the param `versionId`. 'Actual null version' --> instead of the master version objMD.
I'll rewrite the note to be more clear

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+                });
+                return cb(errors.NoSuchBucket);
+            }
+            if (bucketShield(bucket, requestType)) {
+                log.debug('bucket is shielded from request', {
+                    requestType,
+                    method: 'metadataGetBucketAndObject',
+                });
+                // still return bucket for cors headers
+                return cb(errors.NoSuchBucket, bucket);
```
electrachong: removed

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -0,0 +1,158 @@
...
+                (err, options) => next(err, options, infoArr));
+        },
+        function createNullVerDeleteArray(options, infoArr, next) {
+            if (!options.deleteNullVersionData) {
+                return next(null, options, infoArr);
+            }
+            // When options.deleteNullVersionData is true, need to get
+            // location info of null version for deletion. Only applies when
+            // there is pre-existing null version that is not the latest version
+            // and versioning is suspended.
```
electrachong: if null version is the latest version, options.deleteData would be true and data is deleted using objMD. We have a specific check for deleteNullVersionData because in that case we don't want to delete data of the latest version, but of the previous null version.

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -0,0 +1,158 @@
...
+                            error: err,
+                        });
+                        return next(err);
+                    }
+                    if (nullObjMD.location) {
+                        const nullDataToDelete =
+                            Array.isArray(nullObjMD.location) ?
+                            nullObjMD.location : [nullObjMD.location];
+                        dataToDelete = dataToDelete ?
+                            dataToDelete.push(nullDataToDelete) :
```
electrachong: I need to update this - nullDataToDelete (when overwriting a null version that is not the latest version) should not happen when there is dataToDelete (overwriting data from the latest version).

lib/api/objectDelete.js
ttvi: @rahulreddy can I infer from you description that: the number of objects in a bucket is equal to the number of objects that we can see through the non-versioning api? Shouldn't we take into account the number of versions as well?
I cannot find it now but I saw an AWS document saying they count each delete marker as an object with 0-byte length. In our case a delete marker basically takes up as much space in metadata as a normal object. Should we update UTAPI metrics or counting a delete marker as an object?

lib/api/apiUtils/object/versioning.js
electrachong: I think you understood it correctly. Sending `{ versioning: true }` to metadata as the params for the putObjectMD call will instruct metadata to create a new version. We save the current `nullVersionId` to put in the object metadata for the new version so we can reference it when a user requests the null version in a s3 api call.

lib/api/objectDelete.js
rahulreddy: @ttvi Interesting, I would assume the number of objects count would reflect what you would see from list objects api and not list object versions api. Let me play with S3 metrics to see how they are handling this situation.

lib/api/multiObjectDelete.js
electrachong: @alexandre-merle I think they are only encrypted if the error happens in _parseXml when decrypting, otherwise the versionId that is saved in the entry is the decrypted id. I'm thinking to save the original version ID in the entry as well so we don't have to re-encrypt ourselves.

electrachong:
latest end to end passed: http://ci.ironmann.io/gh/scality/Integration/11213

lib/api/multiObjectDelete.js
electrachong: thanks for the catch @LaurenSpiegel

lib/api/bucketPutVersioning.js
```diff
@@ -42,7 +42,7 @@ function _parseXML(request, log, cb) {
         if (mfaDelete) {
             log.debug('mfa deletion is not implemented');
             return cb(errors.NotImplemented
-                .customizedDescription('MFA Deletion is not supported yet.'));
+                .customizeDescription('MFA Deletion is not supported yet.'));
```
electrachong: moved to `clean: some optimization and dev dependencies` commit

lib/api/completeMultipartUpload.js
```diff
@@ -380,22 +387,43 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                        data.batchDelete(dataToDelete, logger
-                            .newRequestLoggerFromSerializedUids(log
-                            .getSerializedUids()));
-                    }
-                    return next(null, mpuBucket, mpuOverviewKey,
-                        aggregateETag, storedPartsAsObjects,
-                        extraPartLocations, destinationBucket);
-                });
+            let versioningOptions = undefined;
+            async.waterfall([
```
electrachong: was refactored

lib/api/completeMultipartUpload.js
```diff
@@ -380,22 +387,43 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                callback => services.versioningPreprocessing(bucketName,
+                    bucketMD, objectKey, objectMD, null, log, callback),
+                (options, callback) => {
+                    versioningOptions = options;
```
electrachong: was refactored

lib/api/completeMultipartUpload.js
```diff
@@ -380,22 +387,43 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                    // eslint-disable-next-line
+                    metaStoreParams.isNull = options.isNull;
+                    // eslint-disable-next-line
+                    metaStoreParams.nullVersionId = options.nullVersionId;
+                    services.metadataStoreObject(
+                            destinationBucket.getName(), dataLocations,
+                            pseudoCipherBundle, metaStoreParams, callback);
+                },
+            ], (err, res) => {
+                // TODO to include the versioning information in result
```
electrachong: removed, old comment

lib/api/completeMultipartUpload.js
```diff
@@ -417,6 +445,9 @@ function completeMultipartUpload(authInfo, request, log, callback) {
         if (err) {
             return callback(err, null, corsHeaders);
         }
+        if (generatedVersionId) {
+            corsHeaders['x-amz-version-id'] = VID.encrypt(generatedVersionId);
```
electrachong: renamed to res headers to be more accurate

lib/api/objectDelete.js
```diff
@@ -24,48 +28,124 @@ export default function objectDelete(authInfo, request, log, cb) {
...
-            }
+
+    const canonicalID = authInfo.getCanonicalID();
+    let bucketMD = undefined;
+    let objectMD = undefined;
+    let corsHeaders = undefined;
+    let removeDeleteMarker = false;
+    let deleteOptions = undefined;
+    return async.waterfall([
+        callback => services.metadataValidateAuthorization(valParams, callback),
```
electrachong: I refactored objectDelete and may have taken care of this, let me know if you still see any issue with the cors headers.

lib/api/objectPut.js
electrachong: updated

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
+import { errors, versioning } from 'arsenal';
+
+import metadata from '../../../metadata/wrapper';
+
+const versionIdUtils = versioning.VersionID;
+const VID_INF = versionIdUtils.VID_INF;
```
electrachong: is `nonVersionedObjId` OK?

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
...
+
+export function decryptVersionId(reqQuery) {
```
electrachong: done, https://github.com/scality/S3/pull/651/commits/88fecbdc9f708c0760be797065a58e3b4e61bcb1

lib/api/apiUtils/object/versioning.js
```diff
@@ -0,0 +1,161 @@
...
+    try {
+        versionId = versionIdUtils.decrypt(versionId);
+    } catch (exception) {
+        return errors.InvalidArgument
+            .customizeDescription('Invalid version id specified');
+    }
+    return versionId;
+}
+
+export function getVersionIdResHeader(verCfg, objectMD) {
```
electrachong: done, https://github.com/scality/S3/pull/651/commits/88fecbdc9f708c0760be797065a58e3b4e61bcb1

electrachong: done, https://github.com/scality/S3/pull/651/commits/88fecbdc9f708c0760be797065a58e3b4e61bcb1

lib/api/apiUtils/object/versioning.js
electrachong: done, https://github.com/scality/S3/pull/651/commits/88fecbdc9f708c0760be797065a58e3b4e61bcb1

lib/api/apiUtils/object/versioning.js
LaurenSpiegel: great

lib/api/apiUtils/object/versioning.js
electrachong: in a technical sense, yes; I think overwriting the master version is what is meant

lib/api/apiUtils/object/versioning.js
electrachong: done, https://github.com/scality/S3/pull/651/commits/88fecbdc9f708c0760be797065a58e3b4e61bcb1

lib/api/apiUtils/object/versioning.js
electrachong: ok

lib/api/objectCopy.js
electrachong: done, https://github.com/scality/S3/pull/651/commits/88fecbdc9f708c0760be797065a58e3b4e61bcb1

lib/api/objectCopy.js
electrachong: fixed

lib/api/objectCopy.js
electrachong: yes, you're right. I updated the api

lib/api/objectGetACL.js
electrachong: done, https://github.com/scality/S3/pull/651/commits/88fecbdc9f708c0760be797065a58e3b4e61bcb1

lib/api/objectGetACL.js
electrachong: I put them back https://github.com/scality/S3/pull/651/files#diff-fe5d22d3317694e890e517aa9c1c6566R83

lib/api/objectGetACL.js
electrachong: updated

lib/api/objectDelete.js
```diff
@@ -24,48 +30,132 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                            collectCorsHeaders(request.headers.origin,
+                                request.method, bucketMD);
+                        return cb(null, corsHeaders);
+                    }
+                    // To adhere to AWS behavior, create a delete marker even
+                    // if trying to delete an object that does not exist when
+                    // versioning has been configured
+                    return next(null, bucketMD, objMD);
+                }
+                const headerValResult = validateHeaders(objMD,
```
electrachong: ok

lib/api/multiObjectDelete.js
electrachong: fixed

lib/api/objectGet.js
electrachong: ok

lib/api/objectDelete.js
```diff
@@ -1,13 +1,19 @@
-import { errors } from 'arsenal';
+import { errors, versioning } from 'arsenal';
+import async from 'async';
```
electrachong: done

lib/api/objectDelete.js
```diff
@@ -23,44 +29,132 @@ export default function objectDelete(authInfo, request, log, cb) {
...
+                    }
+                    // TODO: currently we return do not create a delete
+                    // marker if the version specified for deletion does
+                    // not exist. To adhere to AWS behavior, we should
+                    // update to create a delete marker anyway.
+                    if (reqVersionId) {
+                        const corsHeaders =
+                            collectCorsHeaders(request.headers.origin,
+                                request.method, bucketMD);
+                        return cb(null, corsHeaders);
```
electrachong: ok, done

lib/api/objectDelete.js
electrachong: ok

lib/api/multiObjectDelete.js
electrachong: done

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+ * @param {RequestLogger} log - request logger
+ * @param {function} cb - callback
+ * @return {undefined} - and call callback with params err, objMD of null ver
+ */
+export function getNullVersion(objMD, bucketName, objectKey, log, cb) {
+    const options = {};
+    if (objMD.isNull || !objMD.versionId) {
+        // null version is current version
+        log.debug('found null version'); // don't want to log internal id?
+        return cb(null, objMD);
```
electrachong: ok

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+        log.debug('null version exists, get the null version');
+        options.versionId = objMD.nullVersionId;
+        return metadata.getObjectMD(bucketName, objectKey, options, log, cb);
+    }
+    log.debug('could not find a null version');
+    return cb();
+}
+
+/** metadataGetBucketAndObject - retrieves bucket and specified version
+ * NOTE: This function only returns master version if 'null' version is
```
electrachong: updated

lib/api/apiUtils/authorization/validation.js
```diff
@@ -0,0 +1,217 @@
...
+        versionId: versionId === 'null' ? undefined : versionId,
+    };
+    return metadata.getBucketAndObjectMD(bucketName, objectKey, options, log,
+        (err, data) => {
+            if (err) {
+                log.debug('metadata get failed', { error: err });
+                return cb(err);
+            }
+            log.debug('bucket and object retrieved from metadata',
+                { database: bucketName, object: objectKey });
```
electrachong: ok

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -0,0 +1,158 @@
...
+                return callback(err);
+            }
+            if (dataToDelete) {
+                data.batchDelete(dataToDelete, deleteLog);
+            }
+            return callback(null, result);
+        });
+}
+
+export default
```
electrachong: done

lib/services.js
electrachong: removed comment, it was unrelated

lib/services.js
electrachong: removed

lib/services.js
electrachong: ok

lib/services.js
electrachong: changed back to debug

lib/services.js
electrachong: updated

electrachong:
@ironman-machine

electrachong:
@ironman-machine 
SCALITY_METADATA_BRANCH=ft/vsp
SCALITY_ARSENAL_BRANCH=ft/vsp

electrachong:
end-to-end passed: http://ci.ironmann.io/gh/scality/Integration/11283

lib/api/objectDelete.js
rahulreddy: I couldn't find any docs which say delete marker is treated as a 0B object (they have an analogy) for analytics but a few tests on AWS show that each version is treated as an object itself.

rahulreddy:
We will handle Utapi corner cases for versioning once this is merged.

electrachong:
a last end-to-end:
@ironman-machine
SCALITY_METADATA_BRANCH=ft/vsp
SCALITY_ARSENAL_BRANCH=ft/vsp

PR #658, opened 2017-03-21
Moving S3 config to conf/

nicolas2bert:


LaurenSpiegel:
I think you have an extra commit in here by accident (the docker one). Also, this seems like it will cause problems for federation/integration.

PR #659, opened 2017-03-21
Implement healthcheck for multiple backend

dora-korpar:


dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

lib/data/multipleBackendGateway.js
```diff
@@ -62,7 +63,28 @@ const multipleBackendGateway = {
...
+            const client = clients[location];
+            if (client.clientType && client.clientType === 'scality') {
+                client.healthcheck(log, (err, res) => {
+                    if (err) {
+                        multBackendResp[location] = { error: err };
+                    } else {
+                        multBackendResp[location] = { code: res.statusCode,
+                            message: res.statusMessage };
+                    }
+                    cb();
```
bennettbuchanan: Shouldn't this `cb` be returned here so as to not move to the else clause?

lib/data/multipleBackendGateway.js
```diff
@@ -62,7 +63,28 @@ const multipleBackendGateway = {
         return client.delete(objectGetInfo, reqUids, callback);
     },
 
-    // checkHealth:
+    healthcheck: (log, callback) => {
+        const multBackendResp = {};
+        async.each(Object.keys(clients), (location, cb) => {
+            const client = clients[location];
+            if (client.clientType && client.clientType === 'scality') {
```
bennettbuchanan: Is the check for `client.clientType` necessary? Would it be possible for `client` to be `undefined`?

dora-korpar: The client.clientType check is there because not every backend gets assigned a clientType, so it would be undefined. And client should never be undefined because that would mean the locationConfig.json was empty, which will throw an error on startup :)

bennettbuchanan: Thanks. It's minor, but an undefined `client.clientType` will also not equal `'scality'`. That's what I meant 😄 .

dora-korpar: Oooh, gotcha. I think I had things set up a little differently earlier so that I needed the check, but you're right, it's redundant here. Thx!

lib/data/multipleBackendGateway.js
```diff
@@ -62,7 +63,28 @@ const multipleBackendGateway = {
...
+                client.healthcheck(log, (err, res) => {
+                    if (err) {
+                        multBackendResp[location] = { error: err };
+                    } else {
+                        multBackendResp[location] = { code: res.statusCode,
+                            message: res.statusMessage };
+                    }
+                    cb();
+                });
+            } else {
```
LaurenSpiegel: please add comment that this is the response for a file or mem backend.  as we add other possible backends, they will have to be addressed.

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

PR #661, opened 2017-03-24
FIX: return invalid location constraint error

nicolas2bert:


nicolas2bert:
@ironman-machine ok

lib/api/apiUtils/object/locationConstraintCheck.js
```diff
@@ -31,11 +31,14 @@ export default function locationConstraintCheck(request, metaHeaders,
...
 
-    if (!BackendInfo.isValidControllingBackendParam(
+    const controllingBackend = BackendInfo.controllingBackendParam(
         objectLocationConstraint, bucketLocationConstraint,
-        requestEndpoint, log)) {
+        requestEndpoint, log);
+    if (!controllingBackend.isValid) {
         backendInfoObj = {
             err: errors.InvalidArgument,
+            description: 'Invalid location constraint for the' +
```
LaurenSpiegel: why not just customize the description of the error itself and send that back and use it in the 3 api's and remove the customize description step from those api's since will already be done?

lib/api/initiateMultipartUpload.js
```diff
@@ -143,8 +143,8 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
             destinationBucket, log);
         if (backendInfoObj.err) {
             return process.nextTick(() => {
-                callback(backendInfoObj.err.customizeDescription('Location ' +
-                    'Constraint Info is invalid.'));
+                callback(backendInfoObj.err.customizeDescription(
```
LaurenSpiegel: this is also done in put and copy I believe.  In any case, the customize step here should be taken out and just do it in the utility function.

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -7,26 +7,32 @@ const log = new DummyRequestLogger();
 const dummyBackendInfo = new BackendInfo('mem', 'file', '127.0.0.1');
 
 describe('BackendInfo class', () => {
-    describe('isValidControllingBackendParam', () => {
+    describe('controllingBackendParam', () => {
         it('should return false if objectLocationConstraint is invalid', () => {
```
LaurenSpiegel: I would change to say, "should return object with applicable error if objectLocationConstraint is invalid"

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -7,26 +7,32 @@ const log = new DummyRequestLogger();
...
-            const res = BackendInfo.isValidControllingBackendParam(
+            const res = BackendInfo.controllingBackendParam(
                 'notValid', 'file', '127.0.0.1', log);
-            assert.equal(res, false);
+            assert.equal(res.isValid, false);
+            assert((res.description).indexOf('Object Location Error')
+            > -1);
         });
         it('should return false if bucketLocationConstraint is invalid', () => {
```
LaurenSpiegel: same idea for this one and the ones below since the function no longer returns false/true

tests/unit/multipleBackend/BackendInfo.js
nicolas2bert: Thanks :)

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -7,26 +7,36 @@ const log = new DummyRequestLogger();
...
+            const res = BackendInfo.controllingBackendParam(
                 'mem', 'file', 'notValid', log);
-            assert.equal(res, false);
+            assert.equal(res.isValid, false);
+            assert((res.description).indexOf('Endpoint ' +
+            'Location Error') > -1);
         });
-        it('should return true if all backend parameters are valid', () => {
-            const res = BackendInfo.isValidControllingBackendParam(
+        it('should return object with applicable if all backend ' +
```
dora-korpar: applicable _error_

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -69,27 +69,35 @@ export class BackendInfo {
...
-    static isValidControllingBackendParam(objectLocationConstraint,
+    static controllingBackendParam(objectLocationConstraint,
         bucketLocationConstraint, requestEndpoint, log) {
         if (objectLocationConstraint &&
             !BackendInfo.isValidLocationConstraint(objectLocationConstraint,
                 log)) {
             log.trace('objectLocationConstraint is invalid');
-            return false;
+            return { isValid: false, description: 'Object Location Error - ' +
+            `Your object location "${objectLocationConstraint}" is not ` +
```
LaurenSpiegel: do you think it is worth escaping for xml?

nicolas2bert: Yes - better safe than sorry :)

nicolas2bert:
@ironman-machine ok

PR #663, opened 2017-03-25
Flush write stream buffers before fsync()

### LINKED ISSUE #662: Data corruption after power loss
Hi !

This is the next step of #311.

Here is my configuration:
Docker version : 17.03.0-ce
Scality S3 Server version : commit 84a7e6f (I also test with latest docker hub image)
OS : Debian 8.7
Virtualisation system : VMWare 6.0u2
WriteCache is disabled, as I am on a virtual machine

I use 4 volumes:
```
          -v /data/localMetadata:/usr/src/app/localMetadata
          -v /data/localData:/usr/src/app/localData
          -v /data/settings/config.json:/usr/src/app/config.json
          -v /data/settings/authdata.json:/usr/src/app/conf/authdata.json
```

I was testing data integrity and I find something odd.
I use a python script to massively insert jpg file in S3 Server.

If during the test, I power off the virtual machine using VCenter (dirty electrical shutdown), some of the files that were written in the last 30 seconds before shutdown are corrupt. More precisely, they are incomplete.

I did a lot of tries, here are my observations:

1) Here is a sample of the last keys that were written before shutdown:
```
Key:1490267308.4234602 - READ ERROR : ('Connection broken: IncompleteRead(147456 bytes read, 15129 more expected)', IncompleteRead(147456 bytes read, 15129 more expected))
Key:1490267309.099464 - Value:NOT UTF-8
Key:1490267309.2076397 - READ ERROR : ('Connection broken: IncompleteRead(147456 bytes read, 2683 more expected)', IncompleteRead(147456 bytes read, 2683 more expected))
Key:1490267309.547425 - Value:NOT UTF-8
Key:1490267310.1784554 - READ ERROR : ('Connection broken: IncompleteRead(208896 bytes read, 26207 more expected)', IncompleteRead(208896 bytes read, 26207 more expected))
Key:1490267310.6997852 - READ ERROR : ('Connection broken: IncompleteRead(147456 bytes read, 22547 more expected)', IncompleteRead(147456 bytes read, 22547 more expected))
Key:1490267310.920769 - Value:NOT UTF-8
Key:1490267312.0441651 - Value:NOT UTF-8
Key:1490267312.729269 - Value:NOT UTF-8
```
Boto3 detect issue is some of them, but not all, and not the last one.
If you locate the file on disk, you see a valid picture, but the bottom is missing, and the size/md5 don't match original file.

2) Electrical shutdown is the only way I found to experience this kind of corruption.
During the test, I tried to :
- Stop the docker
- Stop the docker daemon
- kill -9 node application
- kill -9 docker process
without experiencing any corruption (that's a very good point).

4) I spoke with my storage engineers. They didn't think this issue could be related to VMWare or our arrays, but to exclude VMWare definitively, I tried on my laptop (with writecache disabled). I mount the docker the same way, use the same script, and observe the same issue.

5) If I wait 30 seconds after the last write before shutting down the virtual machine, there is no corruption.

6) Metadata are always correct and readable (that's another very good point)

After some reading, I found that you can reduce this gap by reducing the **dirty_expire_centisecs** value in sysctl. So, localData writing are async, and the Kernel flush them to disk every 30 seconds by default.

I spoke with my devs, and they say it's not necessarily an issue, just a design choice, and that I should buy an UPS :-) .
As my data is very important, and I don't care about performance, I would prefer not to lose anything, maybe with a new setting in the configuration ?
If there is nothing to correct, maybe the documentation should mention it (and the **dirty_expire_centisecs** solution) because this work very well for me (less than 2 seconds of data loss when properly tweaked).

Best regards,

jonathan-gramain:
This should fix corruption issue when shutting down the power, by
forcing fsync() to be called after all user-space buffers have been
flushed to the OS.

Credits to @RemiCardona for spotting the root cause of the issue

fixes #662

### DIFF
--- a/lib/data/file/backend.js
+++ b/lib/data/file/backend.js
@@ -43,31 +43,39 @@ export const backend = {
                 log.error('error opening filePath', { error: err });
                 return callback(errors.InternalError);
             }
-            const fileStream = fs.createWriteStream(filePath, { fd });
-
-            request.resume();
-            request.pipe(fileStream, { end: false }).on('error', err => {
-                log.error('error streaming data from request on write',
-                    { error: err });
-                return callback(errors.InternalError);
-            });
-            request.on('error', err => {
-                log.error('error streaming data from request on read',
-                    { error: err });
-                // close fileStream
-                return fs.close(fd, () => callback(errors.InternalError));
-            }).on('end', () => {
+            // disable autoClose so that we can close(fd) only after
+            // fsync() has been called
+            const fileStream = fs.createWriteStream(filePath,
+                                                    { fd,
+                                                      autoClose: false });
+            fileStream.on('finish', () => {
                 fs.fsync(fd, err => {
-                    fileStream.end();
+                    fs.close(fd);
                     if (err) {
                         log.error('error streaming data from request on fsync',
-                            { error: err });
+                                  { error: err });
                         return callback(errors.InternalError);
                     }
                     log.debug('finished writing data', { key });
                     return callback(null, key);
                 });
                 return undefined;
+            }).on('error', err => {
+                log.error('error streaming data from request on write',
+                          { error: err });
+                // destroying the write stream forces a close(fd)
+                fileStream.destroy();
+                return callback(errors.InternalError);
+            });
+
+            request.resume();
+            request.pipe(fileStream);
+            request.on('error', err => {
+                log.error('error streaming data from request on read',
+                    { error: err });
+                // destroying the write stream forces a close(fd)
+                fileStream.destroy();
+                return callback(errors.InternalError);
             });
             return undefined;
         });


lib/data/file/backend.js
```diff
@@ -43,31 +43,39 @@ export const backend = {
...
+                log.error('error streaming data from request on write',
+                          { error: err });
+                // destroying the write stream forces a close(fd)
+                fileStream.destroy();
+                return callback(errors.InternalError);
+            });
+
+            request.resume();
+            request.pipe(fileStream);
+            request.on('error', err => {
```
ttvi: Will `request` and `fileStream` ever receive an `error` concurrently?

lib/data/file/backend.js
```diff
@@ -43,31 +43,39 @@ export const backend = {
...
                         log.error('error streaming data from request on fsync',
-                            { error: err });
+                                  { error: err });
                         return callback(errors.InternalError);
                     }
                     log.debug('finished writing data', { key });
                     return callback(null, key);
                 });
                 return undefined;
+            }).on('error', err => {
```
ttvi: I don't know if `pipe` is atomic and if we need in such case to cleanup the unfinished file on error or not?

Tiduster:
Hi !
Well done @jonathan-gramain .
I build a docker image from your branch and relaunch my tests.

No more corrupted keys after dirty shutdown.

Regards,

jonathan-gramain:
Thanks for this feedback, I'm glad to hear it actually helped!

lib/data/file/backend.js
jonathan-gramain: Supposedly possible. I think we should add a flag to avoid calling the callback twice then (calling destroy() multiple times on `fileStream` is fine).

lib/data/file/backend.js
jonathan-gramain: I think we should remove the new file in such case, indeed to avoid making orphans if the input stream is interrupted. Though not related to this PR, so I will do it as part of the current refactoring.

lib/data/file/backend.js
jonathan-gramain: Same than above, it's not a regression and I want to keep this PR small and focused, so will do in the refactoring.

lib/data/file/backend.js
```diff
@@ -43,31 +43,39 @@ export const backend = {
...
+                log.error('error streaming data from request on write',
```
nicolas2bert: Don't we need to close(fd) here in the error event handler as well, since we disabled the autoClose?

jonathan-gramain: This is already done by `fileStream.destroy()` - strace can attest.

PR #666, opened 2017-03-27
S3 c 182/fix/missing error handler

LaurenSpiegel:
I am unable to add a test for the added error handler without a lot of refactoring so I am just making the fix for now. In a separate commit, I added an error handling initiate mpu test to test an error that is returned earlier but was not tested.


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
@ironman-machine now that integration is fixed...

LaurenSpiegel:
end to end passed

PR #667, opened 2017-03-27
FT: Pass requestUID to vaultclient

bennettbuchanan:
Depends on https://github.com/scality/vaultclient/pull/136.

Since vaultclient creates its own request Logger using the requestUID, we just pass the requestUID. We also allow the log and dump level to be configured based on the `log` field of Config.js.

### DIFF
--- a/lib/auth/vault.js
+++ b/lib/auth/vault.js
@@ -31,6 +31,12 @@ if (config.backends.auth === 'mem') {
         });
         client = new vaultclient.Client(host, port);
     }
+    if (config.log) {
+        client.setLoggerConfig({
+            level: config.log.logLevel,
+            dump: config.log.dumpLevel,
+        });
+    }
 }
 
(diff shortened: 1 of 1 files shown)


lib/auth/vault.js
```diff
@@ -270,7 +265,7 @@ function checkHealth(log, callback) {
         defResp[implName] = { code: 200, message: 'OK' };
         return callback(null, defResp);
     }
-    return client.healthcheck(log, (err, obj) => {
+    return client.healthcheck(log.getSerializedUids(), (err, obj) => {
```
bennettbuchanan: This change is needed for https://github.com/scality/vaultclient/pull/136 since the werelogs method `unserializeUids` requires a request UID, and not a logger instance.

bennettbuchanan:
e2e passed with dependent branch https://github.com/scality/vaultclient/pull/136.
http://ci.ironmann.io/gh/scality/Integration/11376

PR #668, opened 2017-03-27
Make sproxyd path customizable

dora-korpar:
Gives user ability to set custom sproxyd path in bootstrap array and fixes chordCos parsing.
Goes with https://github.com/scality/sproxydclient/pull/118

### DIFF
--- a/locationConfig.json
+++ b/locationConfig.json
@@ -16,7 +16,8 @@
         "details": {
             "connector": {
                 "sproxyd": {
-                    "bootstrap": ["localhost:8182"]
+                    "bootstrap": ["localhost:8182"],
+                    "path": "/proxy/arc/"
                 }
             }
         }
--- a/tests/unit/testConfigs/sproxydAssert.js
+++ b/tests/unit/testConfigs/sproxydAssert.js
@@ -1,41 +1,82 @@
 import assert from 'assert';
 import { sproxydAssert } from '../../../lib/Config';
 
-const dummySproxydConfBoot = {
-    sproxyd: {
-        bootstrap: ['localhost:8181'],
-    },
-};
-
-const dummySproxydConfChord = {
-    sproxyd: {
-        chordCos: '20',
-    },
-};
-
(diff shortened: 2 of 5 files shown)


dora-korpar:
@ironman-machine try

tests/unit/testConfigs/sproxydAssert.js
```diff
@@ -1,39 +1,65 @@
...
-
-const dummySproxydConfBoth = {
-    sproxyd: {
-        bootstrap: ['localhost:8181'],
-        chordCos: '20',
-    },
-};
-
-describe('sproxydAssert', () => {
+describe.only('sproxydAssert', () => {
```
LaurenSpiegel: nix the only

dora-korpar:
@ironman-machine try

locationConfig.json
```diff
@@ -16,7 +16,7 @@
         "details": {
             "connector": {
                 "sproxyd": {
-                    "bootstrap": ["localhost:8182"]
+                    "bootstrap": ["localhost:8182:/proxy/arc/"]
```
jonathan-gramain: I don't see where this new config element is parsed.
Otherwise, I don't think this is the right place to put the base path: it ought to be the same path for all bootstraps (otherwise bad things will happen) so it should be in its own configuration variable IMO.

bennettbuchanan: I believe its parsed in this method: https://github.com/scality/sproxydclient/blob/master/lib/sproxyd.js#L45

dora-korpar: It's actually parsed in sproxydclient, here https://github.com/scality/sproxydclient/blob/master/lib/sproxyd.js#L45.
As to the other point, @LaurenSpiegel, do you have any thoughts?

dora-korpar:
@ironman-machine try

locationConfig.json
dora-korpar: As discussed, path should not be set per bootstrap, but per sproxyd. New commits reflect changes

PR #672, opened 2017-03-29
FIX: Handle account-level metrics listing

bennettbuchanan:
Query params sent to the arsenal auth method currently only handle bucket-level metrics. Since the listing binary supports account-level metrics, we also need to be able to send the correct path for account-level listings. This also prepares the route for remaining metric levels: to be introduced in https://github.com/scality/S3/pull/536


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed

PR #677, opened 2017-03-31
DOC: Change Boto3 usage to use boto3 mapping feature

arnaudlamy4scality:
Boto3's client is just a small part of the Boto3 library. With the
current implementation, it will return a dict of the S3 item. With this
one, it will use the boto3 mapping feature as well and will return an object
instead of a dict.

### DIFF
--- a/docs/CLIENTS.rst
+++ b/docs/CLIENTS.rst
@@ -236,9 +236,11 @@ Python
 `boto3 <http://boto3.readthedocs.io/en/latest/index.html>`__
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code:: python
+Client integration
 
+.. code:: python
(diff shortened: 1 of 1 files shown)


nicolas2bert:
Hey @arnaudlamy4scality -
Thank you for contributing! 

We will try to test and integrate the change as soon as we can.
In order to contribute to the project, please sign your work https://github.com/scality/Guidelines/blob/master/CONTRIBUTING.md#sign-your-work

nicolas2bert:
Boto 3 has both:
- low-level client `boto3.client('s3')`
- higher-level resource (object-oriented interface) `boto3.resource('s3')`

I think we should still keep the low-level client and add your higher-level resource contribution.

dora-korpar:
Hey @arnaudlamy4scality, I know it's been a while, but if you're still able to rebase this we can get it merged in a jiff :)

arnaudlamy4scality:
done :)

PR #679, opened 2017-04-03
bf/Handle dataFile and datatMem backend names

dora-korpar:
If a server was started with multiple backends after having put an object on a server started with file backend, the object would be unretrievable. This PR fixes that

### DIFF
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -14,10 +14,10 @@ let implName;
 
 if (config.backends.data === 'mem') {
     client = inMemory;
-    implName = 'dataMem';
+    implName = 'mem';
 } else if (config.backends.data === 'file') {
     client = file;
-    implName = 'dataFile';
+    implName = 'file';
 } else if (config.backends.data === 'multiple') {
     client = multipleBackendGateway;
     implName = 'multipleBackends';
--- a/locationConfig.json
+++ b/locationConfig.json
@@ -25,19 +25,21 @@
     "us-east-1": {
         "type": "file",
         "legacyAwsBehavior": true,
-        "details": {
-        }
+        "details": {}
     },
     "file": {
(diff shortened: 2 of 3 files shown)


dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

lib/data/locationConstraintParser.js
```diff
@@ -9,10 +9,10 @@ export default function parseLC() {
 
     Object.keys(config.locationConstraints).forEach(location => {
         const locationObj = config.locationConstraints[location];
-        if (locationObj.type === 'mem') {
+        if (locationObj.type === 'mem' || locationObj.type === 'dataMem') {
```
LaurenSpiegel: I thought we were going to leave this alone and just change the implName in the data wrapper to "mem" and "file"?

dora-korpar:
@ironman-machine try

lib/data/locationConstraintParser.js
```diff
@@ -12,7 +12,7 @@ export default function parseLC() {
         if (locationObj.type === 'mem') {
             clients[location] = inMemory;
         }
-        if (locationObj.type === 'file') {
+        if (locationObj.type === 'file' || locationObj.type === 'dataFile') {
```
LaurenSpiegel: we should get rid of this change as well

dora-korpar: Without that, objects put to file before this release won't be accessible though, right?

LaurenSpiegel: if you add both to the config, it should work, right?

dora-korpar: ooh yeah, you're right

dora-korpar:
@ironman-machine one mo time

PR #680, opened 2017-04-03
ft: add versionId to multiObjDelete requestContext [S3C-184]

electrachong:
This is to add version ids to the request contexts for multi object delete requests.
Dependency of scality/Vault#728

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -143,9 +143,7 @@ function _parseXml(xmlToParse, next) {
         //     {"Key":["test2"]}
         // ]}
         const objects = [];
-        const itemErrors = [];
         for (let i = 0; i < json.Object.length; i++) {
-            let itemError = null;
             const item = json.Object[i];
             if (!item.Key) {
                 return next(errors.MalformedXML);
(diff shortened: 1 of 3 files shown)


lib/api/multiObjectDelete.js
```diff
@@ -370,7 +370,10 @@ function multiObjectDelete(authInfo, request, log, callback) {
                     signatureAge: authParams.params.data.signatureAge,
                 },
                 parameterize: {
-                    specificResource: objects.map(entry => entry.key),
+                    specificResource: objects.map(entry => Object.assign({}, {
```
LaurenSpiegel: no need to do object.assign

electrachong:
end to end passed here: http://ci.ironmann.io/gh/scality/Integration/11688

lib/api/multiObjectDelete.js
```diff
@@ -412,11 +418,21 @@ function multiObjectDelete(authInfo, request, log, callback) {
...
                         }
+                        const key = result.arn.slice(slashIndex + 1);
+                        const versionId = result.versionId ?
+                            versionIdUtils.encode(result.versionId) : undefined;
                         if (result.isAllowed) {
-                            inPlay.push(objects[i]);
+                            inPlay.push({
+                                key,
+                                versionId,
+                                decodedVid: result.versionId,
```
LaurenSpiegel: add note as to why you save the decodedVid here

electrachong: I think this is wrong. We should be using the encoded version id sent by user for policies and so result.versionId should be the encoded version id. I need to look at this again

electrachong:
end-to-end passed here: http://ci.ironmann.io/gh/scality/Integration/11765

PR #681, opened 2017-04-04
fix: check for numeric filtering results instead of boolean

ttvi:
Fix the missed out check in #651 for the filtering results of listing in bucketfile (should check for numeric values instead of boolean).

### DIFF
--- a/lib/metadata/bucketfile/backend.js
+++ b/lib/metadata/bucketfile/backend.js
@@ -432,7 +432,7 @@ class BucketFileInterface {
             const stream = db.createReadStream(requestParams);
             stream
                 .on('data', e => {
-                    if (extension.filter(e) === false) {
+                    if (extension.filter(e) < 0) {
                         stream.emit('end');
                         stream.destroy();
                     }
--- a/tests/unit/metadata/bucketfile/backend.js
+++ b/tests/unit/metadata/bucketfile/backend.js
@@ -1,0 +1,68 @@
+const EventEmitter = require('events');
+const assert = require('assert');
+
+const extensions = require('arsenal').algorithms.list;
+
+const BucketFileInterface =
+    require('../../../../lib/metadata/bucketfile/backend').default;
+
+const KEY_LENGTH = 15;
+const KEY_COUNT = 1000;
+const MAX_KEYS = 100;
+
+const KEY_TEMPLATE = '0'.repeat(KEY_LENGTH + 1);
+
+function zpad(key) {
+    return `${KEY_TEMPLATE}${key}`.slice(-KEY_LENGTH);
+}
+
+class Reader extends EventEmitter {
+    constructor() {
+        super();
+        this.done = false;
+        this.index = 0;
+        process.nextTick(() => this.start());
+    }
+
+    start() {
+        if (this.done) {
+            return null;
+        }
+        const i = this.index++;
+        // extensions should take into account maxKeys
+        // and should not filter more than that intended value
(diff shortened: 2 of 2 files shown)


ttvi:
Expectation: the commit for the unit test should fail and the commit for the fix should not.

lib/metadata/bucketfile/backend.js
```diff
@@ -432,7 +432,7 @@ class BucketFileInterface {
             const stream = db.createReadStream(requestParams);
             stream
                 .on('data', e => {
-                    if (extension.filter(e) === false) {
+                    if (extension.filter(e) < 0) {
```
ThibaultRiviere: Shouldn't use the filter variable from arsenal ?

ttvi: Same answer from https://github.com/scality/MetaData/pull/1009/files#r107927401: no, metadata only cares if the value is `< 0`, it does not need to understand the specific values (-1, -2, ...) unless we map these values to specific actions in metadata.

tests/unit/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,73 @@
...
+    bucketfile.loadDBIfExists = (bucket, log, callback) => callback(null, db);
+    // eslint-disable-next-line no-unused-vars
+    const unRef = bucketfile.unRef;
+    bucketfile.unRef = () => {};
+    // remember to set the original functions back if having more tests
+
+    Object.keys(extensions).forEach(listingType => {
+        it(`listing max ${MAX_KEYS} keys using ${listingType}`, done => {
+            const params = { listingType, maxKeys: MAX_KEYS };
+            // assertion to check if extensions work with maxKeys is in Reader
```
jonathan-gramain: I don't see the assertion you're referring to here.

jonathan-gramain: Oh it must be the assertion in the `start()` function then.

ttvi:
@ironman-machine hello

tests/unit/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,73 @@
...
+
+    // stub db to inspect the extensions
+    const db = {
+        createReadStream: () => new Reader(),
+    };
+
+    // stub functions and components
+    const logger = { info: () => {}, debug: () => {}, error: () => {} };
+    // eslint-disable-next-line no-unused-vars
+    const loadDBIfExists = bucketfile.loadDBIfExists;
```
LaurenSpiegel: why is this defined?  To restore the function after the test?  If so, let's do that now in an after block.

tests/unit/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,73 @@
...
+    bucketfile.loadDBIfExists = (bucket, log, callback) => callback(null, db);
+    // eslint-disable-next-line no-unused-vars
+    const unRef = bucketfile.unRef;
+    bucketfile.unRef = () => {};
+    // remember to set the original functions back if having more tests
```
LaurenSpiegel: Let's just do that clean up now.

tests/unit/metadata/bucketfile/backend.js
```diff
@@ -0,0 +1,73 @@
...
+        process.nextTick(() => this.start());
+    }
+
+    start() {
+        if (this.done) {
+            return null;
+        }
+        const i = this.index++;
+        // extensions should take into account maxKeys
+        // and should not filter more than that intented value
```
LaurenSpiegel: typo -- intended

PR #684, opened 2017-04-05
rf: change encoding of version ids [S3C-184]

electrachong:
Dependent on https://github.com/scality/Arsenal/pull/242

Switched to just encoding the version ids, and hex instead of base64 to simplify authentication checks.

Functional tests were changed just to test with both v2 and v4, not just v4. The large diffs are mostly due to indentation level.


electrachong:
@ironman-machine 
SCALITY_ARSENAL_BRANCH=ft/version-encoding
try

electrachong:
@ironman-machine
SCALITY_ARSENAL_BRANCH=ft/version-encoding
try

PR #687, opened 2017-04-08
FIX: returning correct DeleteMarkerVersionId when deleteObjects

nicolas2bert:
Fixing ceph test: 
`ERROR: s3tests.functional.test_s3.test_versioning_multi_object_delete_with_marker_create`

`DeleteMarkerVersionId` should be a versionId not a boolean. 

6. Also check that multi-object delete has ‘isDeleteMarker’ in result in both cases deleting delete marker and new delete marker (according to AWS)

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -97,10 +97,9 @@ function _formatXML(quietSetting, errorResults, deleted) {
     }
     const deletedXML = [];
     deleted.forEach(version => {
-        const isDeletingDeleteMarker = version.isDeletingDeleteMarker;
+        const isDeleteMarker = version.isDeleteMarker;
+        const deleteMarkerVersionId = version.deleteMarkerVersionId;
         // if deletion resulted in new delete marker or deleting a delete marker
-        const isDeleteMarker =
-            version.newDeleteMarker || isDeletingDeleteMarker;
         deletedXML.push(
             '<Deleted>',
             '<Key>',
@@ -118,13 +117,9 @@ function _formatXML(quietSetting, errorResults, deleted) {
             deletedXML.push(
                 '<DeleteMarker>',
                 isDeleteMarker,
-                '</DeleteMarker>'
-            );
-        }
-        if (isDeletingDeleteMarker) {
-            deletedXML.push(
+                '</DeleteMarker>',
                 '<DeleteMarkerVersionId>',
-                isDeletingDeleteMarker,
+                deleteMarkerVersionId,
                 '</DeleteMarkerVersionId>'
             );
         }
@@ -253,10 +248,10 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
                 // This call will create a delete-marker
                 return createAndStoreObject(bucketName, bucket, entry.key,
                     objMD, authInfo, canonicalID, null, request,
-                    deleteInfo.newDeleteMarker, null, log, err =>
-                    callback(err, objMD, deleteInfo));
+                    deleteInfo.newDeleteMarker, null, log, (err, result) =>
+                    callback(err, objMD, deleteInfo, result.versionId));
             },
-        ], (err, objMD, deleteInfo) => {
+        ], (err, objMD, deleteInfo, versionId) => {
             if (err === skipError) {
                 return moveOn();
             } else if (err) {
@@ -271,9 +266,24 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             // TODO: update number of objects to reflect creation of
             // delete markers; Utapi must be updated to allow this for
             // 'deleteObject' metrics
-            const newDeleteMarker = deleteInfo.newDeleteMarker;
-            successfullyDeleted.push({ entry, newDeleteMarker,
-                isDeletingDeleteMarker: objMD.isDeleteMarker });
+            let isDeleteMarker;
+            let deleteMarkerVersionId;
+            // - If trying to delete an object that does not exist (if a new
+            // delete marker was created)
+            // - Or if an object exists but no version was specified
+            // return DeleteMarkerVersionId equals the versionID of the marker
+            // you just generated and DeleteMarker tag equals true
+            if (deleteInfo.newDeleteMarker) {
+                isDeleteMarker = true;
+                deleteMarkerVersionId = versionIdUtils.encode(versionId);
+            // If trying to delete a delete marker, DeleteMarkerVersionId equals
+            // deleteMarker's versionID and DeleteMarker equals true
+            } else if (objMD && objMD.isDeleteMarker) {
+                isDeleteMarker = true;
+                deleteMarkerVersionId = entry.versionId;
+            }
+            successfullyDeleted.push({ entry, isDeleteMarker,
+              deleteMarkerVersionId });
             return moveOn();
         });
     },
--- a/tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
+++ b/tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
@@ -3,6 +3,7 @@ import async from 'async';
 
 import withV4 from '../support/withV4';
 import BucketUtility from '../../lib/utility/bucket-util';
+import { removeAllVersions } from '../../lib/utility/versioning-util';
 
 const bucketName = `multi-object-delete-${Date.now()}`;
 const key = 'key';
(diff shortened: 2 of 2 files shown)


tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
```diff
@@ -138,3 +139,96 @@ testing('Multi-Object Versioning Delete Success', function success() {
...
+                return s3.deleteBucket({ Bucket: bucketName }, err => {
+                    assert.strictEqual(err, null,
+                        `Error deleting bucket: ${err}`);
+                    return done();
+                });
+            });
+        });
+
+        it('deleteing a delete marker - should send back VersionId and ' +
+        'DeleteMarkerVersionId equals to deleteVersionId', done => {
```
LaurenSpiegel: equals should be 'equal'

also, would be clearer to say 'should send back VersionId and DeleteMarkerVersionId both equal to deleteVersionId'.

tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
```diff
@@ -138,3 +139,96 @@ testing('Multi-Object Versioning Delete Success', function success() {
...
+                    assert.strictEqual(data.Deleted[0].VersionId,
+                      deleteVersionId);
+                    assert.strictEqual(data.Deleted[0].DeleteMarkerVersionId,
+                      deleteVersionId);
+                    next(err);
+                }),
+            ], err => done(err));
+        });
+
+        it('delete object that has not been created - should send back the ' +
```
LaurenSpiegel: should send back a DeleteMarkerVersionId matching the versionId stored for the object

tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
nicolas2bert: Done - thanks :)

nicolas2bert:
@ironman-machine try

lib/api/multiObjectDelete.js
```diff
@@ -271,9 +272,27 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
             // delete markers; Utapi must be updated to allow this for
             // 'deleteObject' metrics
+            let isDeleteMarker;
+            let deleteMarker;
+            // If trying to delete an object that does not exist (empty objMD)
+            // return DeleteMarkerVersionId equals the versionID generated and
+            // DeleteMarker equals true
+            if (!objMD) {
+                isDeleteMarker = true;
+                deleteMarker = versionIdUtils.encode(versionId);
```
electrachong: instead of re-encoding the version ID here, you can probably use entry.versionId (the encoded version ID provided by the user)

lib/api/multiObjectDelete.js
```diff
@@ -124,7 +125,7 @@ function _formatXML(quietSetting, errorResults, deleted) {
         if (isDeletingDeleteMarker) {
```
electrachong: we can get rid of this variable and just rely on `deletingDeleteMarker`:
```
if (deletingDeleteMarker) {
             deletedXML.push(
                  '<DeleteMarkerVersionId>',
                deletingDeleteMarker,
                  '</DeleteMarkerVersionId>'
              );
}
```
`deletingDeleteMarker` is only defined if we are deleting the delete marker, if I understand correctly.

electrachong: nvm, I looked at the code more, I guess deletingDeleteMarker is also true if we are creating a new delete marker?

lib/api/multiObjectDelete.js
```diff
@@ -271,9 +272,27 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             // TODO: update number of objects to reflect creation of
             // delete markers; Utapi must be updated to allow this for
             // 'deleteObject' metrics
+            let isDeleteMarker;
+            let deleteMarker;
+            // If trying to delete an object that does not exist (empty objMD)
+            // return DeleteMarkerVersionId equals the versionID generated and
+            // DeleteMarker equals true
+            if (!objMD) {
```
electrachong: isn't this the case not just if we create a delete marker for an object that doesn't exist, but in general if we create a delete marker? I think this is why I created the variable `deleteInfo.newDeleteMarker` and set it equal to true if we create a delete marker. In that case you can replace `!objMD` here with `deleteInfo.newDeleteMarker`.

lib/api/multiObjectDelete.js
electrachong: [AWS document](http://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html) states we should return `DeleteMarkerVersionId` whether creating or deleting a delete marker:

> If the specific delete request in the Multi-Object Delete either creates or deletes a delete marker, Amazon S3 returns this element in response with the version ID of the delete marker. When deleting an object in a bucket with versioning enabled, this value is present for the following two reasons [...]

According to the document, I think we should get rid of `isDeletingDeleteMarker` and stick this xml in the previous check, `if (isDeleteMarker)`. We should also rename `deletingDeleteMarker` to `deleteMarkerVersionId` so it's not confusing. And `isDeleteMarker` should only be true if we are passing `deletingDeleteMarker`/`deleteMarkerVersionId` so we can also replace / get rid of `isDeleteMarker` if we want and check for `if (deleteMarkerVersionId)` instead.

It was probably confusing due to the existing code -- sorry, I must have mis-interpreted the docs while refactoring and wrote this with the wrong interpretation. Please double-check and make sure I'm not wrong again by checking AWS behavior and making tests if we don't already have them.

lib/api/multiObjectDelete.js
```diff
@@ -271,9 +266,25 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             // TODO: update number of objects to reflect creation of
             // delete markers; Utapi must be updated to allow this for
             // 'deleteObject' metrics
-            const newDeleteMarker = deleteInfo.newDeleteMarker;
-            successfullyDeleted.push({ entry, newDeleteMarker,
-                isDeletingDeleteMarker: objMD.isDeleteMarker });
+            let isDeleteMarker;
+            let deleteMarkerVersionId;
+            // If trying to delete an object that does not exist (empty objMD)
```
electrachong: update the comment (`If trying to delete an object that does not exist (empty objMD)` --> `if a new delete marker was created`)

lib/api/multiObjectDelete.js
```diff
@@ -271,9 +266,25 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
+                isDeleteMarker = true;
+                deleteMarkerVersionId = versionIdUtils.encode(versionId);
+            // If trying to delete a delete marker, DeleteMarkerVersionId equals
+            // deleteMarker's versionID and DeleteMarker equals true
+            } else if (objMD && objMD.isDeleteMarker) {
+                isDeleteMarker = true;
+                deleteMarkerVersionId = entry.versionId;
+            } else {
+                isDeleteMarker = false;
+                deleteMarkerVersionId = undefined;
```
electrachong: default value of `deleteMarkerVersionId` should already be undefined and if you don't care if `isDeleteMarker` is `undefined` instead of `false`, you don't need this condition block. It's up to you though, maybe you want to be explicit.

lib/api/multiObjectDelete.js
nicolas2bert: thanks :)

tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
```diff
@@ -140,7 +140,7 @@ testing('Multi-Object Versioning Delete Success', function success() {
     });
 });
 
-testing('Multi-Object Versioning Delete - deleting delete marker',
+describe('Multi-Object Versioning Delete - deleting delete marker',
```
electrachong: if you change 'describe' to 'testing' these tests will probably not pass the ci until https://github.com/scality/S3/pull/690 is merged because versioning is not implemented for bucketfile yet

nicolas2bert: sure!

lib/api/multiObjectDelete.js
```diff
@@ -273,26 +267,22 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             // delete markers; Utapi must be updated to allow this for
             // 'deleteObject' metrics
             let isDeleteMarker;
-            let deleteMarker;
-            // If trying to delete an object that does not exist (empty objMD)
-            // return DeleteMarkerVersionId equals the versionID generated and
-            // DeleteMarker equals true
-            if (!objMD) {
+            let deleteMarkerVersionId;
+            // If trying to delete an object that does not exist (if a new
```
electrachong: we also create delete markers when an object exists but no version was specified, and versioning has been configured

nicolas2bert: Sure - I added one more test. :)

electrachong: Thank you!! :smiley:

nicolas2bert:
@ironman-machine try

PR #688, opened 2017-04-10
Ft/aws backend

dora-korpar:
Adds real AWS as possible backend.

### DIFF
--- a/lib/data/locationConstraintParser.js
+++ b/lib/data/locationConstraintParser.js
@@ -1,3 +1,4 @@
+import AWS from 'aws-sdk';
 import file from './file/backend';
 import inMemory from './in_memory/backend';
 import Sproxy from 'sproxydclient';
@@ -31,7 +32,38 @@ export default function parseLC() {
             clients[location].clientType = 'scality';
         }
         if (locationObj.type === 'aws_s3') {
-            // TODO
+            // users can either include the desired profile name from their
+            // ~/.aws/credentials file or include the accessKeyId and
+            // secretAccessKey directly in the locationConfig
+            if (locationObj.details.credentialsProfile) {
+                clients[location] = new AWS.S3({
+                    endpoint: `https://${locationObj.details.awsEndpoint}`,
+                    // Non-file stream objects are not supported
+                    // with SigV4 (node sdk)
+                    // so note that we are using the default of
+                    // signatureVersion v2
+
+                    // consider disabling
+                    debug: true,
+                    // perhaps use this rather than setting ourselves.
+                    // Not implemented yet for streams in node sdk!!!
+                    computeChecksums: true,
+                    credentials: new AWS.SharedIniFileCredentials({ profile:
+                        locationObj.details.credentialsProfile }),
+                });
+            } else {
+                clients[location] = new AWS.S3({
+                    endpoint: `https://${locationObj.details.awsEndpoint}`,
+                    debug: true,
+                    computeChecksums: true,
+                    accessKeyId: locationObj.details.credentials.accessKey,
+                    secretAccessKey: locationObj.details.credentials.secretKey,
+                });
+            }
+            clients[location].clientType = 'aws_s3';
+            clients[location].awsBucketName = locationObj.details.bucketName;
+            clients[location].bucketMatch = locationObj.details.bucketMatch;
+            clients[location].dataStoreName = location;
         }
     });
     return clients;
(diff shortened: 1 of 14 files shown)


dora-korpar:
@ironman-machine try plz

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

lib/data/multipleBackendGateway.js
```diff
@@ -18,6 +18,16 @@ function createLogger(reqUids) {
 
 const clients = parseLC(config);
 
+function _createAwsKey(requestBucketName, requestObjectKey,
+    partNumber, uploadId, bucketMatch) {
+    if (bucketMatch) {
```
nicolas2bert: Should not it be: if bucketMatch => real AWS key starting with `requestBucketName`?

lib/data/locationConstraintParser.js
```diff
@@ -31,7 +32,38 @@ export default function parseLC() {
...
+                    // consider disabling
+                    debug: true,
+                    // perhaps use this rather than setting ourselves.
+                    // Not implemented yet for streams in node sdk!!!
+                    computeChecksums: true,
+                    credentials: new AWS.SharedIniFileCredentials({ profile:
+                        locationObj.details.credentialsProfile }),
+                });
+            } else {
+                clients[location] = new AWS.S3({
```
nicolas2bert: Is there a good/simple/cool way to warn users that including the accessKeyId and
secretAccessKey directly in the locationConfig is not secure (response/ Readme)?

dora-korpar: Yeah, I'll update the readme in the coolest way :)

lib/data/multipleBackendGateway.js
```diff
@@ -77,9 +147,36 @@ const multipleBackendGateway = {
                     }
                     cb();
                 });
+            } else if (client.clientType === 'aws_s3') {
+                client.headBucket({ Bucket: client.awsBucketName },
+                err => {
+                    if (err) {
+                        multBackendResp[location] = { error: err };
```
nicolas2bert: cb() here :)

nicolas2bert: _Suggestion_ : adding a text before the error may help the user to figure out what is wrong, like you did for versioning. (AWS S3 bucket is missing)

dora-korpar: Ah good call. For the error, unless there's something wrong with AWS itself, it will return a 403 AccessDenied error, which the user will be able to see. And I didn't do versioning :)

dora-korpar: Oh, I guess it could return bucket missing too, but the user gets the entire error back. Do you think they need more information?

nicolas2bert: We return this: ```'aws-test':
   { error:
      { NotFound: null
          at Request.extractError (/Users/nicolas.humbert/Desktop/s3/node_modules/aws-sdk/lib/services/s3.js:519:35)
          at Request.callListeners (/Users/nicolas.humbert/Desktop/s3/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
          at Request.emit (/Users/nicolas.humbert/Desktop/s3/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
         ...
message: null,
        code: 'NotFound',
        region: null,
        time: 2017-04-11T22:49:02.165Z,
        requestId: '0401E2616E70626B',
        extendedRequestId: 'PWNV0HrkokrIWTTzxkKXARf0z/r8YnrMA/DgCvUnKiimBSOsYpmzwmhLHH3sCYSW9YRWGzDLTzE=',
        cfId: undefined,
        statusCode: 404,
        retryable: false,
        retryDelay: 11.879374018464993 } } }```

It may be enough but adding `AWS S3 bucket is missing` or something similar before it may make it clearer.

dora-korpar: If I do that for one possible error, then I have to do it for all of them. I think AWS's errors are well documented enough that they'll be able to figure it out

lib/data/locationConstraintParser.js
ploki: Let me gently disagree ;-) ```~/.aws/credentials``` is in plain text as well. and so, not more secure than ```locationConfig.json``` file.
BTW, these files remain safe as long as the system is not compromised and the proper permissions are used

PR #689, opened 2017-04-10
TEST: versioning objectCopy add more checks

nicolas2bert:
1. For object copy, checking that we are conform on deleting data when copying to the same bucket/objectKey and versioning is enabled.


nicolas2bert:
@ironman-machine ok

PR #686, opened 2017-04-07
S3C-158 fix metadata class names

jonathan-gramain:
Adapt to changes done in https://github.com/scality/Arsenal/pull/235

PR #692, opened 2017-04-11
S3C-158 use remote data server for file backend

jonathan-gramain:
## Description

Use the remote data server for data transfers in file backend (Arsenal PR: https://github.com/scality/Arsenal/pull/235)

### Motivation and context

For HA project we need to have the ability to run the storage in a separate daemon that can be restarted and on which multiple S3 processes can connect.

### Related issues

https://scality.atlassian.net/browse/S3C-158

unit tests and ft_test should be enough on S3 side (new unit tests exist on Arsenal)


init.js
```diff
@@ -1,55 +1,3 @@
...
-            return next(err);
-        }
-        return next();
-    });
-},
- err => {
-     assert.strictEqual(err, null, `Error creating data files ${err}`);
-     logger.info('Init complete.  Go forth and store data.');
- });
+// no global init code :)
```
LaurenSpiegel: can we get rid of this file then?  Did you retain my "Go forth and store data" in arsenal?  I don't want to disappoint our users.

lib/Config.js
```diff
@@ -268,6 +268,19 @@ class Config {
...
+            this.dataDaemon = {};
+            assert.strictEqual(typeof config.dataDaemon.host, 'string',
+                               'bad config: data daemon host must be ' +
+                               'a string');
+            this.dataDaemon.host = config.dataDaemon.host;
+
+            assert(Number.isInteger(config.dataDaemon.port)
+                   && config.dataDaemon.port > 0,
+                   'bad config: dataDaemon port must be a positive ' +
+                   'integer');
```
LaurenSpiegel: should we check to make sure it's not the same port as the s3 server and metadata server?

dataserver.js
```diff
@@ -0,0 +1,29 @@
...
+const config = require('./lib/Config.js').default;
+const logger = require('./lib/utilities/logger').logger;
+
+if (config.backends.data === 'file') {
+    /**
+     * Configure the file paths for data if using the file
+     * backend. If no path provided, uses data at the root of
+     * the S3 project directory.
+     */
+    const dataPath = process.env.S3DATAPATH ?
```
LaurenSpiegel: why do this here instead of in config.js?

lib/metadata/bucketfile/backend.js
```diff
@@ -8,7 +8,7 @@ import constants from '../../../constants';
 import config from '../../Config';
 
 const errors = arsenal.errors;
-const MetadataClient = arsenal.storage.metadata.client;
+const MetadataClient = arsenal.storage.metadata.MetadataFileClient;
```
LaurenSpiegel: why is this showing up in this pr?

mdserver.js
```diff
@@ -2,12 +2,19 @@
 require('babel-core/register');
 
 const config = require('./lib/Config.js').default;
-const MetadataServer = require('arsenal').storage.metadata.server;
+const MetadataServer =
```
LaurenSpiegel: same

package.json
```diff
@@ -67,9 +67,11 @@
     "lint_md": "mdlint $(git ls-files '*.md')",
     "mem_backend": "S3BACKEND=mem node index.js",
     "perf": "mocha --compilers js:babel-core/register tests/performance/s3standard.js",
-    "start": "node init.js && npm-run-all --parallel start_mdserver start_s3server",
+    "start": "npm-run-all --parallel start_dmd start_s3server",
     "start_mdserver": "node mdserver.js",
-    "start_s3server": "node index.js",
+    "start_dataserver": "node dataserver.js",
+    "start_s3server": "node init.js && node index.js",
```
LaurenSpiegel: can we get rid of node init.js now?

init.js
jonathan-gramain: Vinh had concerns that this file could be used by other scripts. In doubt, I kept it, if you're confident we can remove it, I'll remove it.

jonathan-gramain: And yes the message is still there in Arsenal, in the DataFileStore setup() function.

LaurenSpiegel: It should only have been used in the package.json.

lib/Config.js
jonathan-gramain: We can, if you think it's worth it.

dataserver.js
jonathan-gramain: I moved it here, because it's not an S3 parameter anymore, but a "dmd" parameter. I think it makes sense to have it in this file because this is the startup file of the dmd. If you think it's better to keep everything in Config.js, I can revert back.

lib/metadata/bucketfile/backend.js
jonathan-gramain: This is because this PR has a dependency on https://github.com/scality/S3/pull/686, so it actually contains the changes.

jonathan-gramain: Or more exactly, it depends on https://github.com/scality/Arsenal/pull/235 which itself has a two-way dependency with #686.

dataserver.js
LaurenSpiegel: I'd rather minimize the pulling of env variables throughout the code and centralize it in the config.

init.js
jonathan-gramain: Now that I think about it, it may be a wider-scale message only popped when *all* initialization is complete, not just the data storage setup. Although it probably ends up being the same now because, maybe I should put this message in another place (init.js? ^^).

package.json
jonathan-gramain: done

dataserver.js
jonathan-gramain: Now that I think about it, in order to put this back in Config.js requires mdserver.js to parse the whole config as S3 does, because we can and some will run the dmd as a process independent of S3, which means that the dmd would have to have a valid S3 config. I don't like this approach too much, I think they should stay independent. (the `config.json` of the dmd side should ideally be able to contain only what is required to launch the dmd).

Another approach is to let the S3 server pass on the data path configuration bit to the dmd initialization code, but here the issue is that it should not be S3 server's responsiblity to choose the path to store data to but the dmd, which again may be on a remote machine.

Let's just make sure that `S3DATAPATH` and `S3METADATAPATH` are documented clearly if not already the case.

LaurenSpiegel: ok

lib/Config.js
LaurenSpiegel: I can go either way.  If we get a complaint we can add it.

circle.yml
```diff
@@ -127,3 +126,9 @@ test:
             > $CIRCLE_ARTIFACTS/server_file_kms_rawnode.txt
         & bash wait_for_local_port.bash 8000 40
         && ENABLE_KMS_ENCRYPTION=true npm run ft_node
+
+    # Run S3 with multiple data backends ; run multiple_backend_test
+    - S3BACKEND=mem S3DATA=multiple npm start
```
LaurenSpiegel: instead of npm start, perhaps we should just start the dmd since we don't actually need the s3 server running for these.

lib/api/apiUtils/object/setUpCopyLocator.js
```diff
@@ -63,10 +62,8 @@ function setUpCopyLocator(sourceObjMD, rangeHeader, log) {
                 };
         }
         if (range) {
-            dataLocator =
-                routesUtils.setPartRanges(dataLocator, range);
-            copyObjectSize = Math.min(sourceSize - range[0],
-                range[1] - range[0] + 1);
+            dataLocator = routesUtils.setPartRanges(dataLocator, range);
+            copyObjectSize = range[1] - range[0] + 1;
```
LaurenSpiegel: was the math.min unnecessary?

jonathan-gramain: It was, the returned range is guaranteed to not go beyond the object size, that is provided as a parameter.

dataserver.js
rahulreddy: I would side with @LaurenSpiegel on this one. Although the dmd may reside on a different server it will still be packaged and deployed using S3/ project. What services we start on independent servers is a different story but one thing is for sure, they should be pulling config from one place. It's easier to manage and we will have no ENV vars floating all over the place. You can still use the ENV vars if you want to provide overrides.

lib/api/objectGet.js
```diff
@@ -80,68 +78,51 @@ function objectGet(authInfo, request, log, callback) {
...
             if (range) {
-                // End of range should be included so + 1
+                byteRange = range;
                 responseMetaHeaders['Content-Length'] =
-                    Math.min(maxContentLength - range[0],
-                    range[1] - range[0] + 1);
-                responseMetaHeaders['Content-Range'] = `bytes ${range[0]}-`
-                    + `${Math.min(maxContentLength - 1, range[1])}` +
-                    `/${maxContentLength}`;
+                    range[1] - range[0] + 1;
```
rahulreddy: May be leave the comment `// End of range should be included so + 1` as it gives context to the mysterious `+ 1`

dataserver.js
jonathan-gramain: @rahulreddy I think Lauren was only talking about putting the ENV var reading all in one place, not to create a new configuration value in config.json, nevertheless I second you on this one, I think it's a good idea. We can still allow the possibility of an ENV var to override the config value, which would not break the existing setups.

We don't have the feature of "pulling config from one place" when it comes to reading the exact same file from different servers, but we can duplicate the config.json on both S3 connectors and the dmd so that the config is consistent (actually the dmd would not run if not provided a valid S3 config file since it's calling the global config parsing code). For the single-server, simple deployment it makes perfect sense to have only one config.json containing all the config.

jonathan-gramain: Or we could pull the config from the same server if they are on shared storage available to both S3 and dmds, where the S3 directory where S3 and dmds are run would be the same.

lib/api/objectGet.js
jonathan-gramain: I removed it because I though that the additional info given was not worth the additional like taken in the code (as you may have guessed :p) but will add it back then if you think +1 is mysterious :)

jonathan-gramain:
@ironman-machine r+

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

rahulreddy:
End to end passed here http://ci.ironmann.io/gh/scality/Integration/12665#build-parameters

PR #693, opened 2017-04-12
CLEANUP: Remove obsolete metadataValidateAuthorization func

dora-korpar:
Versioning PR introduced new metadata validation function called metadataValidateBucketAndObj, so the old metadataValidateAuthorization function is no longer needed.

### DIFF
--- a/lib/api/bucketDelete.js
+++ b/lib/api/bucketDelete.js
@@ -2,7 +2,7 @@ import { errors } from 'arsenal';
 
 import collectCorsHeaders from '../utilities/collectCorsHeaders';
 import { deleteBucket } from './apiUtils/bucket/bucketDeletion';
-import services from '../services';
+import { metadataValidateBucket } from '../metadata/metadataUtils';
 import { pushMetric } from '../utapi/utilities';
 
 /**
@@ -28,20 +28,19 @@ export default function bucketDelete(authInfo, request, log, cb) {
         authInfo,
         bucketName,
         requestType: 'bucketDelete',
-        log,
     };
 
-    return services.metadataValidateAuthorization(metadataValParams,
+    return metadataValidateBucket(metadataValParams, log,
         (err, bucketMD) => {
             const corsHeaders = collectCorsHeaders(request.headers.origin,
                 request.method, bucketMD);
             if (err) {
                 log.debug('error processing request',
-                    { method: 'metadataValidateAuthorization', error: err });
+                    { method: 'metadataValidateBucket', error: err });
                 return cb(err, corsHeaders);
             }
             log.trace('passed checks',
-                { method: 'metadataValidateAuthorization' });
+                { method: 'metadataValidateBucket' });
             return deleteBucket(bucketMD, bucketName, authInfo.getCanonicalID(),
                 log, err => {
                     if (err) {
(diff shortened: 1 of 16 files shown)


dora-korpar:
@ironman-machine try

lib/api/bucketDelete.js
```diff
@@ -31,17 +31,17 @@ export default function bucketDelete(authInfo, request, log, cb) {
         log,
     };
 
-    return services.metadataValidateAuthorization(metadataValParams,
+    return metadataValidateBucketAndObj(metadataValParams, log,
```
electrachong: for methods where we are only validating the bucket, I thought it would make more sense to use a function that only retrieves the bucket in MD and does bucket authorization. If you agree, you can add the following two functions in `metadataUtils` (I removed them in the versioning PR since we weren't using them yet):

```
/** metadataGetBucket - retrieves bucket from metadata, returning error if
 * bucket is shielded
 * @param {string} requestType - type of request
 * @param {string} bucketName - name of bucket
 * @param {RequestLogger} log - request logger
 * @return {undefined} - and call callback with err, bucket md
 */
export function metadataGetBucket(requestType, bucketName, log, cb) {
    return metadata.getBucket(bucketName, log, (err, bucket) => {
        if (err) {
            log.debug('metadata getbucket failed', { error: err });
            return cb(err);
        }
        if (bucketShield(bucket, requestType)) {
            log.debug('bucket is shielded from request', {
                requestType,
                method: 'metadataGetBucketAndObject',
            });
            return cb(errors.NoSuchBucket);
        }
        log.trace('found bucket in metadata');
        return cb(null, bucket);
    });
}

/** metadataValidateBucket - retrieve bucket from metadata and check if user
 * is authorized to access it
 * @param {object} params - function parameters
 * @param {AuthInfo} params.authInfo - AuthInfo class instance, requester's info
 * @param {string} params.bucketName - name of bucket
 * @param {string} params.requestType - type of request
 * @param {RequestLogger} log - request logger
 * @param {function} callback - callback
 * @return {undefined} - and call callback with params err, bucket md
 */
export function metadataValidateBucket(params, log, callback) {
    const { authInfo, bucketName, requestType } = params;
    return metadataGetBucket(requestType, bucketName, log, (err, bucket) => {
        if (err) {
            return callback(err);
        }
        // still return bucket for cors headers
        if (!isBucketAuthorized(bucket, requestType, canonicalID)) {
            log.debug('access denied for user on bucket', { requestType });
            return callback(errors.AccessDenied, bucket);
        }
        return callback(null, bucket);
    });
}
```

lib/api/bucketDelete.js
```diff
@@ -31,17 +31,17 @@ export default function bucketDelete(authInfo, request, log, cb) {
         log,
```
electrachong: You can remove the log here since we have it as a separate param in the `metadataValidateBucketAndObj` function. If you don't want to remove the log from all the `metadataValParams` constants, you can change the `metadataValidateBucketAndObj` function parameters to expect it in the params object (unfortunately, you'll still have to modify the existing `metadataValidateBucketAndObj` references that were used for the versioning apis)

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine gogogo

PR #695, opened 2017-04-12
No log clutter on successful healthcheck

dora-korpar:


dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

PR #697, opened 2017-04-13
FIX: ensure null ver data is deleted on overwrite [S3C-184]

electrachong:
Depends on scality/S3#698

Normally we overwrite and clean data when creating a new version when versioning is suspended 
and the latest version is a null version. However, we must also take care to delete the data when overwriting a null version that is not the latest version.

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -68,14 +68,55 @@ function _storeNullVersionMD(bucketName, objKey, objMD, options, log, cb) {
     });
 }
 
+/** get location of data for deletion
+* @param {string} bucketName - name of bucket
+* @param {string} objKey - name of object key
+* @param {object} options - metadata options for getting object MD
+* @param {string} options.versionId - version to get from metadata
+* @param {RequestLogger} log - logger instanceof
+* @param {function} cb - callback
+* @return {undefined} - and call callback with (err, dataToDelete)
+*/
+function _getDeleteLocations(bucketName, objKey, options, log, cb) {
+    return metadata.getObjectMD(bucketName, objKey, options, log,
+        (err, versionMD) => {
+            if (err) {
+                log.debug('err from metadata getting specified version', {
+                    error: err,
+                    method: '_getDeleteLocations',
+                });
+                return cb(err);
+            }
+            if (!versionMD.location) {
+                return cb();
+            }
+            const dataToDelete = Array.isArray(versionMD.location) ?
+                versionMD.location : [versionMD.location];
+            return cb(null, dataToDelete);
+        });
+}
+
 function _deleteNullVersionMD(bucketName, objKey, options, log, cb) {
-    metadata.deleteObjectMD(bucketName, objKey, options, log, err => {
-        if (err) {
-            log.debug('error from metadata deleting null version',
-            { error: err });
-        }
-        cb(err, options);
-    });
+    // before deleting null version md, retrieve location of data to delete
+    return _getDeleteLocations(bucketName, objKey, options, log,
+        (err, nullDataToDelete) => {
+            if (err) {
+                log.warn('could not find null version metadata', {
+                    error: err,
+                    method: '_deleteNullVersionMD',
+                });
+                return cb(err);
+            }
+            return metadata.deleteObjectMD(bucketName, objKey, options, log,
+                err => {
+                    if (err) {
+                        log.warn('metadata error deleting null version',
+                        { error: err, method: '_deleteNullVersionMD' });
+                        return cb(err);
+                    }
+                    return cb(null, nullDataToDelete);
+                });
+        });
 }
 
 /** versioningPreprocessing - return versioning information for S3 to handle
@@ -88,7 +129,7 @@ function _deleteNullVersionMD(bucketName, objKey, options, log, cb) {
  * @param {RequestLogger} log - logger instance
  * @param {function} callback - callback
  * @return {undefined} and call callback with params (err, options):
- * options.deleteData - (true/undefined) whether to delete data of latest ver
+ * options.dataToDelete - (array/undefined) location of data to delete
  * options.versionId - specific versionId to overwrite in metadata
  *  ('' overwrites the master version)
  * options.versioning - (true/undefined) metadata instruction to create new ver
@@ -100,9 +141,14 @@ function _deleteNullVersionMD(bucketName, objKey, options, log, cb) {
 export function versioningPreprocessing(bucketName, bucketMD, objectKey,
     objMD, reqVersionId, log, callback) {
     const options = {};
+    let mstObjLocation;
+    if (objMD && objMD.location) {
+        mstObjLocation = Array.isArray(objMD.location) ?
+            objMD.location : [objMD.location];
+    }
     // bucket is not versioning enabled
     if (!bucketMD.getVersioningConfiguration()) {
-        options.deleteData = true;
+        options.dataToDelete = mstObjLocation;
         return callback(null, options);
     }
     // bucket is versioning enabled
@@ -117,7 +163,7 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
                 // versioning is suspended, overwrite existing master version
                 options.versionId = '';
                 options.isNull = true;
-                options.deleteData = true;
+                options.dataToDelete = mstObjLocation;
                 return callback(null, options);
             }
             // versioning is enabled, create a new version
@@ -142,25 +188,40 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
             if (nullVersionId === undefined) {
                 return callback(null, options);
             }
-            options.deleteNullVersionData = true;
             return _deleteNullVersionMD(bucketName, objectKey,
-                { versionId: nullVersionId }, log,
-                err => callback(err, options));
+                { versionId: nullVersionId }, log, (err, nullDataToDelete) => {
+                    if (err) {
+                        log.warn('unexpected error deleting null version md', {
+                            error: err,
+                            method: 'versioningPreprocessing',
+                        });
+                        if (err === errors.NoSuchKey) {
+                            // it's possible there was a concurrent request to
+                            // delete the null version, so proceed with putting
+                            // a new version
+                            return callback(null, options);
+                        }
+                        return callback(errors.InternalError);
+                    }
+                    options.dataToDelete = nullDataToDelete;
+                    return callback(err, options);
+                });
         }
         // versioning is enabled, put the new version
         options.versioning = true;
         options.nullVersionId = nullVersionId;
         return callback(null, options);
+    // NOTE: Currently we do not accept requests with PUTs to specific
+    // version ID, but this may change after implementing georeplication.
     } else if (!mstVersionId) {
         // version-specific versioning operation, master is not versioned
         if (vstat === 'Suspended' || reqVersionId === 'null') {
             // object does not exist or is not versioned (before versioning)
             options.versionId = '';
             options.isNull = true;
-            options.deleteData = true;
+            options.dataToDelete = mstObjLocation;
             return callback(null, options);
         }
-        // TODO check AWS behaviour
         return callback(errors.BadRequest);
     } else if (mstIsNull) {
         // master is versioned and is a null version
@@ -168,18 +229,26 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
             // overwrite the existing version, make new version null
             options.versionId = '';
             options.isNull = true;
-            options.deleteData = true;
+            options.dataToDelete = mstObjLocation;
             return callback(null, options);
         }
-        // TODO check AWS behaviour
-        options.versionId = reqVersionId;
-        options.deleteData = true;
-        return callback(null, options);
     }
     // master is versioned and is not a null version
     options.versionId = reqVersionId;
-    options.deleteData = true;
-    return callback(null, options);
+    return _getDeleteLocations(bucketName, objectKey,
+        { versionId: reqVersionId }, log, (err, dataToDelete) => {
+            if (err && err === errors.NoSuchKey) {
+                return callback(null, options);
+            } else if (err) {
+                log.warn('unexpected error getting delete locations', {
+                    error: err,
+                    method: 'versioningPreprocessing',
+                });
+                return callback(errors.InternalError);
+            }
+            options.dataToDelete = dataToDelete;
+            return callback(null, options);
+        });
 }
 
 /** preprocessingVersioningDelete - return versioning information for S3 to
--- a/tests/unit/api/multipartUpload.js
+++ b/tests/unit/api/multipartUpload.js
@@ -6,9 +6,12 @@ import async from 'async';
 import { parseString } from 'xml2js';
 
 import bucketPut from '../../../lib/api/bucketPut';
+import bucketPutVersioning from '../../../lib/api/bucketPutVersioning';
+import objectPut from '../../../lib/api/objectPut';
 import completeMultipartUpload from '../../../lib/api/completeMultipartUpload';
 import constants from '../../../constants';
-import { cleanup, DummyRequestLogger, makeAuthInfo } from '../helpers';
+import { cleanup, DummyRequestLogger, makeAuthInfo, versioningTestUtils }
+    from '../helpers';
 import { ds } from '../../../lib/data/in_memory/backend';
 import initiateMultipartUpload from '../../../lib/api/initiateMultipartUpload';
 import { metadata } from '../../../lib/metadata/in_memory/metadata';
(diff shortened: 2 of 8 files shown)


electrachong:
@ironman-machine 
try

lib/api/apiUtils/object/versioning.js
```diff
@@ -155,17 +186,17 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
...
             options.versionId = '';
             options.isNull = true;
-            options.deleteData = true;
+            options.dataToDelete = mstObjLocation;
             return callback(null, options);
         }
         // TODO check AWS behaviour
         options.versionId = reqVersionId;
-        options.deleteData = true;
+        options.dataToDelete = mstObjLocation;
```
electrachong: I think in this case (as well as on line 199) we may need to retrieve the data location of the specified version and delete it, not the mstObjLocation. Although at the current point, there is no way to PUT to a specific version ID. @ttvi can you confirm that sounds right?

electrachong:
previous end-to-end passed; trying a new one after changes
@ironman-machine try

electrachong:
end-to-end passed

lib/api/apiUtils/object/versioning.js
```diff
@@ -55,14 +55,46 @@ function _storeNullVersionMD(bucketName, objKey, objMD, options, log, cb) {
     });
 }
 
+function _getVersionLocation(bucketName, objKey, options, log, cb) {
```
LaurenSpiegel: please add jsdoc.  do you think this is better called, getDeleteLocations?  Is this used for anything else?

lib/api/apiUtils/object/versioning.js
```diff
@@ -55,14 +55,46 @@ function _storeNullVersionMD(bucketName, objKey, objMD, options, log, cb) {
...
+    return _getVersionLocation(bucketName, objKey, options, log,
+        (err, nullDataToDelete) => {
+            if (err) {
+                log.warning('could not find null version metadata', {
+                    error: err,
+                    method: '_deleteNullVersionMD',
+                });
+                return cb(err);
+            }
+            return metadata.deleteObjectMD(bucketName, objKey, options, log,
```
LaurenSpiegel: does this mirror what we were doing before?  getting delete locations, storing and immediately deleting the metadata?

lib/api/apiUtils/object/versioning.js
```diff
@@ -129,10 +166,14 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
             if (nullVersionId === undefined) {
                 return callback(null, options);
             }
-            options.deleteNullVersionData = true;
             return _deleteNullVersionMD(bucketName, objectKey,
-                { versionId: nullVersionId }, log,
-                err => callback(err, options));
+                { versionId: nullVersionId }, log, (err, nullDataToDelete) => {
+                    if (err === 'NoSuchVersion') {
+                        return callback(errors.InternalError);
```
LaurenSpiegel: why the translation?
please add log.

lib/api/apiUtils/object/versioning.js
```diff
@@ -144,29 +185,48 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
             // object does not exist or is not versioned (before versioning)
             options.versionId = '';
             options.isNull = true;
-            options.deleteData = true;
+            options.dataToDelete = mstObjLocation;
             return callback(null, options);
         }
         // TODO check AWS behaviour
         return callback(errors.BadRequest);
+    // NOTE: Currently we do not accept receive requests with PUTs to
```
LaurenSpiegel: delete "receive"

lib/api/apiUtils/object/versioning.js
```diff
@@ -144,29 +185,48 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
...
     }
     // master is versioned and is not a null version
     options.versionId = reqVersionId;
-    options.deleteData = true;
-    return callback(null, options);
+    return _getVersionLocation(bucketName, objectKey,
+        { versionId: reqVersionId }, log, (err, dataToDelete) => {
+            if (err && err === 'NoSuchVersion') {
+                return callback(null, options);
+            } else if (err) {
```
LaurenSpiegel: why not the actual error?
please at least log it.

tests/unit/api/objectPutVersioned.js
```diff
@@ -0,0 +1,199 @@
...
+                assert.strictEqual(ds[j], expectedValues[i]);
+            } else {
+                assert.deepStrictEqual(ds[j].value, expectedValues[i]);
+            }
+        }
+    }
+
+    it('should delete latest version when creating new null version ' +
+    'if latest version is null version', done => {
+        let i = 0;
```
LaurenSpiegel: I think these tests would be clearer if you did them without 'i'.  You have a set array in each, just use the actual index you are trying to pull.

tests/unit/api/objectPutVersioned.js
```diff
@@ -0,0 +1,199 @@
...
+                    });
+                }),
+        ], err => {
+            if (err) {
+                return done(err);
+            }
+            return done();
+        });
+    });
+});
```
LaurenSpiegel: there should also be tests for object copy and complete mpu

lib/api/apiUtils/object/versioning.js
```diff
@@ -144,29 +185,48 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
             // object does not exist or is not versioned (before versioning)
             options.versionId = '';
             options.isNull = true;
-            options.deleteData = true;
+            options.dataToDelete = mstObjLocation;
             return callback(null, options);
         }
         // TODO check AWS behaviour
```
LaurenSpiegel: can we delete this now?

lib/api/apiUtils/object/versioning.js
```diff
@@ -144,29 +185,48 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
...
         options.versionId = reqVersionId;
-        options.deleteData = true;
-        return callback(null, options);
+        return _getVersionLocation(bucketName, objectKey,
+            { versionId: reqVersionId }, log, (err, dataToDelete) => {
+                if (err && err === 'NoSuchVersion') {
+                    // no data to delete
+                    return callback(null, options);
+                } else if (err) {
+                    return callback(errors.InternalError);
```
LaurenSpiegel: why mapping?  please log

lib/api/apiUtils/object/versioning.js
```diff
@@ -129,10 +166,14 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
             if (nullVersionId === undefined) {
```
LaurenSpiegel: when suspended, do you only delete the master version if it is a null version?  please update comment above

lib/api/apiUtils/object/versioning.js
electrachong: before this fix, we were deleting metadata, getting delete locations, then storing new metadata + deleting data. This presented a problem because the metadata was deleted before getting the data delete locations from it.

lib/api/apiUtils/object/versioning.js
electrachong: when suspended, we always overwrite the master version by using the `''` option in the metadata put call. Normally this takes care of deleting/overwriting the previous null version metadata, as long as the null version is the master version.

However, if there was a previous null version in the version history and it is not the master version, we have to manually delete the null version metadata. Here we check if there is no null version before we do the manual delete.

lib/api/apiUtils/object/versioning.js
electrachong: before if metadata unexpectedly could not find the null version to update MD / delete data, we were returning NoSuchKey error. (If we keep the check, should check for 'NoSuchKey', not 'NoSuchVersion') This meant if user is trying to put a version, complete a mpu or do object copy and we get this unexpected error, they would get 'NoSuchKey', which might be confusing to the user.

The alternative is we could silence the error and put a new version anyway, which may succeed, but the previous null version may unexpectedly show up in listings if deleting the metadata failed.

electrachong: Are the logs in `_deleteNullVersionMD` enough?

lib/api/apiUtils/object/versioning.js
electrachong: I added a jsdoc and updated the name. It's only used for getting the delete locations.

lib/api/apiUtils/object/versioning.js
electrachong: I deleted it; in any case we can't check AWS behavior because AWS also does not allow PUT to specific versions afaik.

lib/api/apiUtils/object/versioning.js
electrachong: Mapped because the only error we expect is 'NoSuchKey'; if it is some other system error, we won't expose it to the user. I added another log.

lib/api/apiUtils/object/versioning.js
electrachong: It's been logged in the _getDeleteLocations function, but I added another log for clarity.

tests/unit/api/objectPutVersioned.js
electrachong: I updated to use the actual index instead of referring to `i`

tests/unit/api/objectPutVersioned.js
electrachong: I added these two tests

electrachong:
@ironman-machine try

tests/unit/api/multipartUpload.js
```diff
@@ -1626,3 +1667,114 @@ describe('Multipart Upload API', () => {
...
+                assert.strictEqual(ds[j], expectedValues[i]);
+            } else {
+                assert.deepStrictEqual(ds[j].value, expectedValues[i]);
+            }
+        }
+    }
+
+    before(done => {
+        cleanup();
+        async.series([
```
bennettbuchanan: Minor: I guess I would expect the final result object to be used with `series` since that's what seems to differentiate it from `waterfall`.

tests/unit/api/objectPut.js
```diff
@@ -22,10 +25,39 @@ const testPutBucketRequest = new DummyRequest({
...
+    'xmlns="http://s3.amazonaws.com/doc/2006-03-01/">' +
+    `<Status>${status}</Status>` +
+    '</VersioningConfiguration>';
+    request.post = xml;
+    return request;
+}
 
 const objectName = 'objectName';
 
+function _createPutObjectRequest(body) {
```
bennettbuchanan: You might make some of these functions test utilities so as to not duplicate.

tests/unit/api/objectPut.js
```diff
@@ -230,3 +262,113 @@ describe('objectPut API', () => {
...
+            callback => objectPut(authInfo, testPutObjectRequests[2],
+                undefined, log, err => {
+                    process.nextTick(() => {
+                        // old null version should be deleted
+                        objData[1] = undefined;
+                        _assertDataStoreValues(objData.slice(0, 3));
+                        callback(err);
+                    });
+                }),
+        ], err => {
```
bennettbuchanan: You could return `done` here.

tests/unit/api/multipartUpload.js
electrachong: I like to use series when I don't need the return arguments for the following call (e.g. I can just have `callback` as the parameter for each function, unlike a waterfall). In this case I'm comparing the expected result against `ds` rather than any of the results of the functions, so I don't need the array of results object.

tests/unit/api/objectPut.js
electrachong: OK, I extracted some of these into test utilities.

tests/unit/api/objectPut.js
electrachong: Thanks, that's nice

electrachong:
@ironman-machine try

tests/unit/api/multipartUpload.js
bennettbuchanan: True, that's cleaner.

lib/api/apiUtils/object/versioning.js
```diff
@@ -142,44 +188,98 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
...
                 return callback(null, options);
             }
-            options.deleteNullVersionData = true;
             return _deleteNullVersionMD(bucketName, objectKey,
-                { versionId: nullVersionId }, log,
-                err => callback(err, options));
+                { versionId: nullVersionId }, log, (err, nullDataToDelete) => {
+                    if (err) {
+                        process.stdout.write('====================\n');
```
jonathan-gramain: Is the console output useful (isn't the log enough)? Particularly since we consider `errors.NoSuchKey` to be a legit error.

lib/api/apiUtils/object/versioning.js
```diff
@@ -142,44 +188,98 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
...
         options.versionId = reqVersionId;
-        options.deleteData = true;
-        return callback(null, options);
+        return _getDeleteLocations(bucketName, objectKey,
+            { versionId: reqVersionId }, log, (err, dataToDelete) => {
+                if (err && err === errors.NoSuchKey) {
+                    // no data to delete
+                    return callback(null, options);
+                } else if (err) {
+                    process.stdout.write('====================\n');
```
jonathan-gramain: Same here and below (although `NoSuchKey` is processed earlier in this case)

lib/api/apiUtils/object/versioning.js
```diff
@@ -142,44 +188,98 @@ export function versioningPreprocessing(bucketName, bucketMD, objectKey,
...
     } else if (mstIsNull) {
         // master is versioned and is a null version
         if (reqVersionId === 'null') {
             // overwrite the existing version, make new version null
             options.versionId = '';
             options.isNull = true;
-            options.deleteData = true;
+            options.dataToDelete = mstObjLocation;
             return callback(null, options);
         }
```
jonathan-gramain: If you close the `if (mstIsNull)` block here, you can factorize the call and callback to `_getDeleteLocations`.

As a general comment, this function is getting rather long, it should probably defer some of its processing to external functions.

tests/unit/api/objectPut.js
```diff
@@ -230,3 +238,102 @@ describe('objectPut API', () => {
...
+            callback => bucketPutVersioning(authInfo, suspendVersioningRequest,
+                log, callback),
+            // creating new null version by putting obj after ver suspended
+            callback => objectPut(authInfo, testPutObjectRequests[1],
+                undefined, log, err => {
+                    // wait until next tick since mem backend executes
+                    // deletes in the next tick
+                    process.nextTick(() => {
+                        // old null version should be deleted
+                        objData[0] = undefined;
```
electrachong: Fair, I think I was slicing the arrays at first because I initially was testing with more values than just three. I updated to list the values each time instead of slicing.

> (possibly modifying assertDataStoreValues() to take data strings as input)
do you providing strings to assertDataStoreValues() instead of the data buffers?

tests/unit/api/objectCopy.js
```diff
@@ -0,0 +1,112 @@
...
+
+    it('should delete null version when creating new null version, ' +
+    'even when null version is not the latest version', done => {
+        const expectedValues = objData.slice();
+        // will have another copy of last object in datastore after objectCopy
+        expectedValues.push(objData[2]);
+        const testObjectCopyRequest = _createObjectCopyRequest(destBucketName);
+        objectCopy(authInfo, testObjectCopyRequest, sourceBucketName, objectKey,
+            undefined, log, err => {
+                assert.strictEqual(err, null, `Unexpected err: ${err}`);
```
electrachong: updated

lib/api/apiUtils/object/versioning.js
```diff
@@ -68,14 +68,55 @@ function _storeNullVersionMD(bucketName, objKey, objMD, options, log, cb) {
...
+                log.warn('could not find null version metadata', {
+                    error: err,
+                    method: '_deleteNullVersionMD',
+                });
+                return cb(err);
+            }
+            return metadata.deleteObjectMD(bucketName, objKey, options, log,
+                err => {
+                    if (err) {
+                        log.debug('metadata error deleting null version',
```
electrachong: `log.warn` maybe, I updated it

lib/api/apiUtils/object/versioning.js
electrachong: I added these in a dropme commit just to print more information in an end-to-end. A previous end-to-end failed with InternalError and I couldn't find the relevant logs (not sure what log level the end to ends run with). Unfortunately, I couldn't duplicate the error so I dropped the commit.

lib/api/apiUtils/object/versioning.js
electrachong: I removed the duplicate code. I agree with your comment about the length, but at the moment I feel there are more important refactors to prioritize. :)

electrachong:
@ironman-machine try

electrachong:
end-to-end passed

PR #700, opened 2017-04-14
fix: delete marker creation / md [S3C-184]

electrachong:
@jonathan-gramain noticed we did not reset the content length for multi-object delete requests when creating delete markers, meaning the delete marker stores content-length with the length of the xml, which is incorrect and causes the request to hang at least for file backend. It would likely also pose an issue for accurately tracking bytes created in Utapi.

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -51,7 +51,7 @@ export default
 function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         canonicalID, cipherBundle, request, isDeleteMarker, streamingV4Params,
         log, callback) {
-    const size = request.parsedContentLength;
+    const size = isDeleteMarker ? 0 : request.parsedContentLength;
 
     const websiteRedirectHeader =
         request.headers['x-amz-website-redirect-location'];
@@ -62,7 +62,8 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         return callback(err);
     }
 
-    const metaHeaders = utils.getMetaHeaders(request.headers);
+    const metaHeaders = isDeleteMarker ? [] :
+        utils.getMetaHeaders(request.headers);
     log.trace('meta headers', { metaHeaders, method: 'objectPut' });
     const objectKeyContext = {
         bucketName,
@@ -78,16 +79,24 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         headers['x-amz-acl'] = request.query['x-amz-acl'];
     }
     const metadataStoreParams = {
-        objectKey, authInfo, metaHeaders, size,
-        contentType: request.headers['content-type'],
-        cacheControl: request.headers['cache-control'],
-        contentDisposition: request.headers['content-disposition'],
-        contentEncoding: removeAWSChunked(request.headers['content-encoding']),
-        expires: request.headers.expires,
+        objectKey,
+        authInfo,
+        metaHeaders,
+        size,
         headers,
-        log,
         isDeleteMarker,
+        log,
     };
+    if (!isDeleteMarker) {
+        metadataStoreParams.contentType = request.headers['content-type'];
+        metadataStoreParams.cacheControl = request.headers['cache-control'];
+        metadataStoreParams.contentDisposition =
+            request.headers['content-disposition'];
+        metadataStoreParams.contentEncoding =
+            removeAWSChunked(request.headers['content-encoding']);
+        metadataStoreParams.expires = request.headers.expires;
+    }
+
     let dataToDelete = undefined;
     if (objMD && objMD.location) {
         dataToDelete = Array.isArray(objMD.location) ?
--- a/lib/services.js
+++ b/lib/services.js
@@ -177,12 +177,14 @@ export default {
                 omVal[val] = overrideMetadata[val];
             });
         }
+
         log.trace('object metadata', { omVal });
-        // If this is not the completion of a multipart upload
-        // parse the headers to get the ACL's if any
+        // If this is not the completion of a multipart upload or
+        // the creation of a delete marker, parse the headers to
+        // get the ACL's if any
         async.waterfall([
             callback => {
-                if (multipart) {
+                if (multipart || omVal.isDeleteMarker) {
                     return callback();
                 }
                 const parseAclParams = {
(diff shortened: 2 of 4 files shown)


electrachong:
@ironman-machine try

electrachong:
end-to-end passed: http://ci.ironmann.io/gh/scality/Integration/11881

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -77,17 +78,30 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
...
-        contentType: request.headers['content-type'],
-        cacheControl: request.headers['cache-control'],
-        contentDisposition: request.headers['content-disposition'],
-        contentEncoding: removeAWSChunked(request.headers['content-encoding']),
-        expires: request.headers.expires,
-        headers,
-        log,
-        isDeleteMarker,
-    };
+    if (isDeleteMarker) {
```
LaurenSpiegel: can't you define metadataStoreParams with all of these items and if !isDeleteMarker add the other items (rather than starting from scratch each time)?

lib/services.js
```diff
@@ -168,21 +168,25 @@ export default {
...
         // from the intiation of the multipart upload
         // (for instance, storage class)
-        Object.keys(metaHeaders).forEach(val => {
-            omVal[val] = metaHeaders[val];
-        });
-
-        if (overrideMetadata) {
-            Object.keys(overrideMetadata).forEach(val => {
-                omVal[val] = overrideMetadata[val];
+        if (!omVal.isDeleteMarker) {
```
LaurenSpiegel: is this necessary? wouldn't there just be an empty array of metaHeaders if it's a delete marker?

tests/unit/api/deleteMarker.js
```diff
@@ -0,0 +1,114 @@
...
+        assert.strictEqual(deleteMarkerMD['content-length'], 0);
+        assert.strictEqual(deleteMarkerMD.location, null);
+        assert.deepStrictEqual(deleteMarkerMD.acl, expectedAcl);
+        undefHeadersExpected.forEach(header => {
+            assert.strictEqual(deleteMarkerMD[header], undefined);
+        });
+    }
+
+    it('should create a delete marker if versioning enabled and deleting ' +
+    'object without version id', done => {
```
LaurenSpiegel: shouldn't this also be the case when deleting an object with a version id?

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -77,17 +78,30 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
...
+        metadataStoreParams = {
+            objectKey,
+            authInfo,
+            metaHeaders,
+            size,
```
jonathan-gramain: could be written as `size: 0`

tests/unit/api/deleteMarker.js
```diff
@@ -0,0 +1,114 @@
...
+            return bucketPutVersioning(authInfo, enableVersioningRequest,
+                log, done);
+        });
+    });
+
+    afterEach(() => {
+        cleanup();
+    });
+
+    function _assertDeleteMakerMd(deleteMarkerMD, deleteResHeaderVersionId) {
```
jonathan-gramain: typo `Maker` -> `Marker`

lib/api/apiUtils/object/createAndStoreObject.js
electrachong: True. Technically, I think we might be using an abundance of caution here since the headers I excluded from the construction of `metadataStoreParams` for the delete marker (cache control, expires, etc.) should not be defined in object delete / multiobjectdelete requests anyway. 

If we assume that we can just go with using the original object, but I guess it might be good to explicitly exclude them in the off chance someone includes those headers in those requests.

LaurenSpiegel: I'm ok excluding them but there is duplication here we can still get rid of.

tests/unit/api/deleteMarker.js
electrachong: by 'deleting object without version id' I mean a delete request sent without specifying a specific version id. If you specify a specific version id the delete will delete the version instead of creating a delete marker for the object. I'll try to make it clear.

For making an equivalent functional test, normally we cannot get the object MD of a delete marker using `getObject.` Should / can I still use `metadata.getObject` directly?

lib/api/apiUtils/object/createAndStoreObject.js
electrachong: Yep, I will update

tests/unit/api/deleteMarker.js
LaurenSpiegel: Since they are different processes it wouldn't work directly.  Couldn't we do a delete test that doesn't get so far under the covers or do we already have that?

electrachong: We have functional tests for creating delete markers with object delete and multi object delete already. They just don't assert the content-length sent for the PUT requests to the datastore (or other metadata properties of the delete marker) because those are behind-the-scenes.

LaurenSpiegel: couldn't you test getting the delete marker back and checking what metadata is on it?

LaurenSpiegel: As discussed, AWS returns MethodNotAllowed for get/head on delete marker so we'll just test with the unit test.

tests/unit/api/deleteMarker.js
electrachong: fixed

electrachong:
@ironman-machine

tests/unit/api/deleteMarker.js
```diff
@@ -1,10 +1,14 @@
-import { versioning } from 'arsenal';
-
+import async from 'async';
```
LaurenSpiegel: why create a deleteMarker.js file instead of mutliObjectDelete or Delete?  those are the api's.

electrachong: There are existing unit tests for multiObjectDelete / Delete that I could have placed these tests in, but I wanted to create a separate file since I knew the tests for creating the delete markers would share functions and set-up.

LaurenSpiegel: ok

LaurenSpiegel:
@ironman-machine try

LaurenSpiegel:
end to end passed here: http://ci.ironmann.io/gh/scality/Integration/12087

Issue #704, opened 2017-04-19
Unable to put Bucket as a user


nicolas2bert:
When I authenticate with a user and try to put a bucket, s3 server throws an error.

PR #705, opened 2017-04-20
FIX: put Bucket as a user

nicolas2bert:
FIXES: #704


LaurenSpiegel:
let's add a functional test like we do for multi-object delete from a user but with bucket put

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine ok

nicolas2bert:
@ironman-machine debug

nicolas2bert:
@ironman-machine ok

nicolas2bert:
@ironman-machine ok

PR #709, opened 2017-04-27
FT: putObjectTagging

nicolas2bert:
ARSENAL: https://github.com/scality/Arsenal/pull/248
UTAPI: https://github.com/scality/utapi/pull/127
INTEGRATION: https://github.com/scality/Integration/tree/ft/putobjecttagging

### DIFF
--- a/lib/api/apiUtils/object/tagging.js
+++ b/lib/api/apiUtils/object/tagging.js
@@ -1,0 +1,142 @@
+import { errors } from 'arsenal';
+import { parseString } from 'xml2js';
+
+const tagRegex = new RegExp(/[^a-zA-Z0-9 +-=._:/]/g);
+
+/*
+    Format of xml request:
+
+    <Tagging>
+      <TagSet>
+         <Tag>
+           <Key>Tag Name</Key>
+           <Value>Tag Value</Value>
+         </Tag>
+      </TagSet>
+    </Tagging>
+*/
+
+
+export const _validator = {
+    validateTagStructure: tag => tag
+        && Object.keys(tag).length === 2
+        && tag.Key && tag.Value
+        && tag.Key.length === 1 && tag.Value.length === 1
+        && tag.Key[0] && tag.Value[0]
+        && typeof tag.Key[0] === 'string' && typeof tag.Value[0] === 'string',
+
+    validateXMLStructure: result =>
+        result && Object.keys(result).length === 1 &&
+        result.Tagging &&
+        result.Tagging.TagSet &&
+        result.Tagging.TagSet.length === 1 &&
+        (
+          result.Tagging.TagSet[0] === '' ||
+          result.Tagging.TagSet[0] &&
+          Object.keys(result.Tagging.TagSet[0]).length === 1 &&
+          result.Tagging.TagSet[0].Tag &&
+          Array.isArray(result.Tagging.TagSet[0].Tag)
+        ),
+
+    validateKeyValue: (key, value) => {
+        if (key.length > 128 || key.match(tagRegex)) {
+            return errors.InvalidTag.customizeDescription('The TagKey you ' +
+              'have provided is invalid');
+        }
+        if (value.length > 256 || value.match(tagRegex)) {
+            return errors.InvalidTag.customizeDescription('The TagValue you ' +
+              'have provided is invalid');
+        }
+        return true;
+    },
+};
+/** _validateTags - Validate tags, returning an error if tags are invalid
+* @param {object[]} tags - tags parsed from xml to be validated
+* @param {string[]} tags[].Key - Name of the tag
+* @param {string[]} tags[].Value - Value of the tag
+* @return {(Error|object)} tagsResult - return object tags on success
+* { key: value}; error on failure
+*/
+function _validateTags(tags) {
+    let result;
+    const tagsResult = {};
+
+    if (tags.length === 0) {
+        return tagsResult;
+    }
+    // Maximum number of tags per resource: 10
+    if (tags.length > 10) {
+        return errors.BadRequest
+            .customizeDescription('Object tags cannot be greater than 10');
+    }
+    for (let i = 0; i < tags.length; i++) {
+        const tag = tags[i];
+
+        if (!_validator.validateTagStructure(tag)) {
+            return errors.MalformedXML;
+        }
+        const key = tag.Key[0];
+        const value = tag.Value[0];
+
+        // Allowed characters are letters, whitespace, and numbers, plus
+        // the following special characters: + - = . _ : /
+        // Maximum key length: 128 Unicode characters
+        // Maximum value length: 256 Unicode characters
+        result = _validator.validateKeyValue(key, value);
+        if (result instanceof Error) {
+            return result;
+        }
+
+        tagsResult[key] = value;
+    }
+    // not repeating keys
+    if (tags.length > Object.keys(tagsResult).length) {
+        return errors.InvalidTag.customizeDescription('Cannot provide ' +
+        'multiple Tags with the same key');
+    }
+    return tagsResult;
+}
+
+/** parseTagXml - Parse and validate xml body, returning callback with object
+* tags : { key: value}
+* @param {string} xml - xml body to parse and validate
+* @param {object} log - Werelogs logger
+* @param {function} cb - callback to server
+* @return {undefined} - calls callback with tags object on success, error on
+*   failure
+*/
+export function parseTagXml(xml, log, cb) {
+    parseString(xml, (err, result) => {
+        if (err) {
+            log.trace('xml parsing failed', {
+                error: err,
+                method: 'parseTagXml',
+            });
+            log.debug('invalid xml', { xml });
+            return cb(errors.MalformedXML);
+        }
+        if (!_validator.validateXMLStructure(result)) {
+            log.debug('xml validation failed', {
+                error: errors.MalformedXML,
+                method: '_validator.validateXMLStructure',
+                xml,
+            });
+            return cb(errors.MalformedXML);
+        }
+        // AWS does not return error if no tag
+        if (result.Tagging.TagSet[0] === '') {
+            return cb(null, []);
+        }
+        const validationRes = _validateTags(result.Tagging.TagSet[0].Tag);
+        if (validationRes instanceof Error) {
+            log.debug('tag validation failed', {
+                error: validationRes,
+                method: '_validateTags',
+                xml,
+            });
+            return cb(validationRes);
+        }
+        // if no error, validation returns tags object
+        return cb(null, validationRes);
+    });
+}
--- a/lib/api/objectPutTagging.js
+++ b/lib/api/objectPutTagging.js
@@ -1,0 +1,98 @@
+import async from 'async';
+import { errors } from 'arsenal';
+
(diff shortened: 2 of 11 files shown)


tests/functional/aws-node-sdk/lib/utility/tagging.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+    return key;
+}
+
+export const taggingTests = [
+    { tag: { key: '+- =._:/', value: '+- =._:/' },
+      it: 'should return tags if tags are valid' },
+    { tag: { key: _generateWordWithLength(129), value: 'foo' },
+      error: 'InvalidTag',
+      it: 'should return InvalidTag if key length supperior to 128' },
```
dora-korpar: 'if key length is greater than 128'

tests/functional/aws-node-sdk/lib/utility/tagging.js
```diff
@@ -0,0 +1,27 @@
...
+    { tag: { key: 'bar', value: _generateWordWithLength(257) },
+      error: 'InvalidTag',
+      it: 'should return InvalidTag if value length supperior to 256',
```
dora-korpar: same as above

tests/unit/api/objectPutTagging.js
```diff
@@ -0,0 +1,187 @@
...
+                return done(err);
+            }
+            return objectPut(authInfo, testPutObjectRequest, undefined, log,
+              done);
+        });
+    });
+
+    afterEach(() => cleanup());
+
+    it('should update a object\'s metadata with tags resource', done => {
```
dora-korpar: an object's

lib/api/apiUtils/object/tagging.js
```diff
@@ -0,0 +1,143 @@
...
+        (
+          result.Tagging.TagSet[0] === '' ||
+          result.Tagging.TagSet[0] &&
+          Object.keys(result.Tagging.TagSet[0]).length === 1 &&
+          result.Tagging.TagSet[0].Tag &&
+          Array.isArray(result.Tagging.TagSet[0].Tag)
+        ),
+
+    validateKeyValue: (key, value) => {
+        if (key.length > 128 || key.match(tagRegex)) {
```
electrachong: you don't mean `!key.match(tagRegex)`?

lib/api/apiUtils/object/tagging.js
```diff
@@ -0,0 +1,143 @@
...
+            return errors.InvalidTag.customizeDescription('The TagKey you ' +
+              'have provided is invalid');
+        }
+        if (value.length > 256 || value.match(tagRegex)) {
```
electrachong: same here?

lib/api/objectPutTagging.js
```diff
@@ -0,0 +1,86 @@
...
+          (err, bucket, objectMD) => next(err, bucket, objectMD)),
+        (bucket, objectMD, next) => {
+            log.trace('parsing tag(s)');
+            return parseTagXml(request.post, log, (err, tags) =>
+              next(err, bucket, tags, objectMD));
+        },
+        (bucket, tags, objectMD, next) => {
+            // eslint-disable-next-line no-param-reassign
+            objectMD.tags = tags;
+            const params = metadataValParams.versionId ?
```
electrachong: if the version ID specified in the request query is 'null', you will want to provide the internally-used version id of the null version here instead of 'null'. Since you validated the version in the previous step, the `objectMD` that was returned should be the null version so I think you can use `objectMD.versionId`. It would be great to have a test when all is done to make sure it works properly.

tests/unit/api/objectPutTagging.js
```diff
@@ -0,0 +1,187 @@
...
+
+    afterEach(() => cleanup());
+
+    it('should update an object\'s metadata with tags resource', done => {
+        const taggingUtil = new TaggingConfigTester();
+        const testObjectPutTaggingRequest = taggingUtil
+            .createObjectTaggingRequest('PUT', bucketName, objectName);
+        objectPutTagging(authInfo, testObjectPutTaggingRequest, log, err => {
+            if (err) {
+                process.stdout.write(`Err putting website config ${err}`);
```
electrachong: update the error message

tests/unit/api/objectPutTagging.js
```diff
@@ -0,0 +1,187 @@
...
+                return done(err);
+            }
+            return metadata.getObjectMD(bucketName, objectName, {}, log,
+            (err, objectMD) => {
+                if (err) {
+                    process.stdout.write(`Err retrieving bucket MD ${err}`);
```
electrachong: error retrieving object md

lib/api/apiUtils/object/tagging.js
electrachong: I guess your regex checks if any non-valid characters exist, you can ignore the comment

lib/api/objectPutTagging.js
```diff
@@ -0,0 +1,95 @@
...
+            log.trace('parsing tag(s)');
+            return parseTagXml(request.post, log, (err, tags) => {
+                if (err) {
+                    return next(err);
+                }
+                if (!objectMD) {
+                    const err = reqVersionId ? errors.NoSuchVersion :
+                        errors.NoSuchKey;
+                    return next(err, bucket);
+                }
```
electrachong: probably, you may need to add one more check if the version/object you got is a delete marker. It seems the error in that case is usually MethodNotAllowed, or InvalidRequest, if the version was specified in the query, or NoSuchKey if no version was specified (just made a similar fix here for objAcl apis: https://github.com/scality/S3/pull/713/files).

nicolas2bert: Fixed. 

I decided to put "putObjectTagging" versioning tests to a separate file in the test/versioning folder to make it clearer and consistent.
I tested against real AWS and both putting tag with AND without version id if version specified is a delete marker should return `MethodNotAllowed`

Thanks for the review :)

electrachong: Thanks!

lib/api/apiUtils/object/tagging.js
```diff
@@ -0,0 +1,143 @@
...
+              'have provided is invalid');
+        }
+        return true;
+    },
+};
+/** _validateTags - Validate tags, returning an error if tags are invalid
+* @param {object[]} tags - tags parsed from xml to be validated
+* @param {string[]} tags[].Key - Name of the tag
+* @param {string[]} tags[].Value - Value of the tag
+* @return {(Error|object)} - return tags object on success; error on failure
```
LaurenSpiegel: actually returning an array of objects.  please update and state what the format of the objects in the array is.

lib/api/apiUtils/object/tagging.js
```diff
@@ -0,0 +1,143 @@
...
+        keys.push(key);
+    }
+    return tagsResult;
+}
+
+/** parseTagXml - Parse and validate xml body, returning tags object on success
+* @param {string} xml - xml body to parse and validate
+* @param {object} log - Werelogs logger
+* @param {function} cb - callback to server
+* @return {undefined} - calls callback with tags object on success, error on
```
LaurenSpiegel: callback with array containing tags objects

lib/api/objectPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+export default function objectPutTagging(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'objectPutTagging' });
+
+    if (!aclUtils.checkGrantHeaderValidity(request.headers)) {
```
LaurenSpiegel: this shouldn't be here unless acl's are being sent.

lib/api/objectPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+    const bucketName = request.bucketName;
+    const objectKey = request.objectKey;
+
+    const decodedVidResult = decodeVersionId(request.query);
+    if (decodedVidResult instanceof Error) {
+        log.trace('invalid versionId query', {
+            versionId: request.query.versionId,
+            error: decodedVidResult,
+        });
+        return callback(decodedVidResult);
```
LaurenSpiegel: process.nextTick here

lib/api/objectPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    return async.waterfall([
+        next => metadataValidateBucketAndObj(metadataValParams, log,
+          (err, bucket, objectMD) => next(err, bucket, objectMD)),
+        (bucket, objectMD, next) => {
+            log.trace('parsing tag(s)');
+            return parseTagXml(request.post, log, (err, tags) => {
+                if (err) {
+                    log.trace('request authorization failed',
```
LaurenSpiegel: it would be clearer to put this log above where the failure actually happens

lib/api/objectPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+                       { method: 'objectPutTagging', error: err });
+                    return next(err);
+                }
+                if (!objectMD) {
+                    const err = reqVersionId ? errors.NoSuchVersion :
+                        errors.NoSuchKey;
+                    log.trace('error processing request',
```
LaurenSpiegel: log should say something more along the lines of no objectMd found

tests/functional/aws-node-sdk/test/object/putObjTagging.js
```diff
@@ -0,0 +1,130 @@
+import assert from 'assert';
+
+import withV4 from '../support/withV4';
+import BucketUtility from '../../lib/utility/bucket-util';
+
+const bucketName = 'testtaggingbucket';
+const objectName = 'testtaggingobject';
+import { taggingTests } from '../../lib/utility/tagging';
+
+function generateMultipleTagConfig(nbr) {
```
LaurenSpiegel: nbr?

lib/api/objectGetACL.js
```diff
@@ -87,7 +87,7 @@ export default function objectGetACL(authInfo, request, log, callback) {
                     if (!objectMD) {
                         const err = versionId ? errors.NoSuchVersion :
                             errors.NoSuchKey;
-                        log.trace('error processing request',
+                        log.trace('error no object metadata found',
```
LaurenSpiegel: wrong file

nicolas2bert:
We are now storing an object "tags" instead of an array of tag's object.
@LaurenSpiegel 
- you might want to do a quick review (just to be sure)
- we can merge as soon as http://ci.ironmann.io/gh/scality/Integration/12750 passes.

lib/api/apiUtils/object/tagging.js
```diff
@@ -0,0 +1,145 @@
...
+* @return {(Error|object)} tagsResult - return object tags on success
+* { key: value}; error on failure
+*/
+function _validateTags(tags) {
+    let result;
+    const tagsResult = {};
+    const keys = [];
+
+    if (tags.length === 0) {
+        return [];
```
LaurenSpiegel: should no longer be an array

PR #710, opened 2017-04-28
chore: upgrade ci ruby version

rahulreddy:


rahulreddy:
I may have to update the fog-aws versions and lock them to match Integration. Integration and S3 fog/gem/ruby versions ideally should be the same.

PR #712, opened 2017-05-01
ft: add versioning info to DESIGN.md

electrachong:
Adapted from the Metadata DESIGN.md document to include details of S3 API implementation.


DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+- delete the specified version identified by `versionId`
+- replace the master version with a place holder for deletion version
+- initiate a repair operation to update the value of the master version:
+  - involves listing the versions of the object and get the latest
+  version to replace the place holder delete version
+  - if no more versions exists, metadata deletes the master version
+
+##### GET
+
+-  no options: original DELETE operation, will get the master version
```
dora-korpar: original GET op :)

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+information specified. For example, in order to get the version `v1` of `foo`,
+users have to specify that version in the query, such as
+`GET /foo?versionId=v1`.
+
+A bucket can be non-versioning or versioning. The version of the objects of a
+bucket before it was versioning-enabled are called the "null" version; these
+versions are kept when a bucket is versioning enabled and are identified with
+the version identifier as "null". Turning off the versioning of a bucket would
+make all operations target the "null" versions of the objects, but would not
+delete all existing versions.
```
electrachong: It's misleading, I guess, because that changing the versioning status doesn't actually change what version the operations targets. I tried to update to be a little more accurate.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+
+### Implementation of Bucket Versioning in S3
+
+Each version of an object is stored as a separate key in metadata that
+corresponds to the data of the version in the datastore. The API component of
+S3 manages the putting and delete of version data, sends instructions to create,
```
electrachong: I changed to `put` and `delete`

DESIGN.md
electrachong: thanks for the catch! :)

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+Version-specific deletions involve a few steps. A deletion targeting the latest
+version of an object has to:
+
+- delete the specified version identified by `versionId`
+- replace the master version with a place holder for deletion version
+- initiate a repair operation to update the value of the master version:
+  - involves listing the versions of the object and get the latest
+  version to replace the place holder delete version
+  - if no more versions exists, metadata deletes the master version
+
```
LaurenSpiegel: specify that this all happens before responding to s3

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+Version-specific deletions involve a few steps. A deletion targeting the latest
+version of an object has to:
+
+- delete the specified version identified by `versionId`
+- replace the master version with a place holder for deletion version
+- initiate a repair operation to update the value of the master version:
+  - involves listing the versions of the object and get the latest
+  version to replace the place holder delete version
+  - if no more versions exists, metadata deletes the master version
```
LaurenSpiegel: meaning that the key no longer exists in metadata

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+
+The implementation of a GET operation does not change compared to the standard
+version. A standard GET without versioning information would get the master
+version of a key. A version-specific GET would retrieve the specific version
+identified by the key for that version.
+
+##### LIST
+
+For a standard LIST on a bucket, metadata iterates through the keys by using
+the separator `.` as an extra delimiter. For a listing of all versions of a
```
LaurenSpiegel: why '.' here?

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+  or remote updates in geo-replication);
+  - if the version identified by `versionId` happens to be the latest version,
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
+    the master version only (same as no options, but returns a version ID)
+
+##### DELETE
+
+-  no options: original DELETE operation, will delete the master version
+- `versionId: <versionId>` delete a specific version
```
LaurenSpiegel: mention delete marker here.  maybe move that whole section here.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+created. When performing actions upon an object whose latest version is a delete
+maker, the S3 API will return a 404 NoSuchKey error and behave as if that object
+has been deleted. To restore a previous version as the latest version of an
+object, the delete marker must be deleted, by the same process as deleting any
+other version.
+
+The metadata engine's versioning processing is agnostic to whether a specific
+version is a delete marker. The S3 API stores information about whether
+a version is a delete marker in its metadata and performs a check to retrieve
+that information.
```
LaurenSpiegel: this part about what s3 does with respect to delete markers needs to be expanded.  How s3 interacts with delete markers in different api's is complex and needs to be communicated to people who might expand the codebase in a clear manner.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+
+#### Null Version Mapping
+
+Null versions are created when versioning has not been enabled or when
+versioning is suspended. Previous null versions are overwritten by new null
```
LaurenSpiegel: has been suspended

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+versions. Within metadata, this means the null version is stored by overwriting
+the master version only, instead of overwriting the master version and
+generating a new version.
```
LaurenSpiegel: Expand to explain that the null version need not be the master version for example when versioning is enabled and new versions are put, the null version will just be another null version.

Also, when versioning is suspended and you have a null version that is not the master, I assume the non-master null version gets overwritten regardless?  Please specify.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
 
 ![Arch](res/architecture.png)
 
+## Versioning
```
electrachong: 👍

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+as GET and LIST would see the latest version of the objects. The table below
+shows such example, a `PUT foo` generates a new version `v2` and keeps `v1`;
+GET and LIST return the latest version of `foo`, which is `v2` in this case.
+
+| key | version | value |
+|-----|:-------:|:-----:|
+| foo |   v2    |   A   |
+| foo |   v1    |   B   |
+
+To work with all existing versions of the objects, S3 provides an extended
```
electrachong: 👍

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+A bucket can be non-versioning or versioning. The version of the objects of a
+bucket before it was versioning-enabled are called the "null" version; these
+versions are kept when a bucket is versioning enabled and are identified with
+the version identifier as "null". Suspending the versioning of a bucket would
+result in the overwrite of any previous null version when a new version is put,
+but would not delete all existing versions.
+
+### Implementation of Bucket Versioning in S3
+
+Each version of an object is stored as a separate key in metadata that
```
electrachong: 👍

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+version identifiers as the suffix for the keys of the object versions, and a
+special version to represent the latest version.
+
+The following is a summary of the design and details of implementation of
+versioning in S3.
+
+#### Version Id Format
+
+The version id is generated by the metadata backend, and encoded in a
+hexadecimal string format by S3. The following information is encoded in a
```
electrachong: 👍

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+version ID:
+
+`object_name separator version_id`
+where:
+
+- `object_name`: is the key of the object in metadata
+- `separator`: we use the `null` character (`0x00` or `\0`) as the separator
+  between the `object_name` and the `version_id` of a key
+- `version_id`: is the version identifier; this encodes the ordering
+  information in the texture format as metadata orders keys alphabetically
```
electrachong: It's the initial wording from the original document, I inferred it refers to the specific format for a version ID. I excluded the specific format for a version ID since I didn't sense the need for that level of detail, but since we're now including an example of the unencoded version IDs in the previous section, I'm adding it back.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+new version or overwrite, retrieve, or delete a specific version by sending
+values for special options in PUT, GET, or DELETE calls to metadata.
+
+Note: all operations (PUT and DELETE) that generate a new version of an object
+will return the `version_id` of the new version to the API.
+
+##### PUT
+
+- no options: original PUT operation, will update the master version
+- `versioning: true` create a new version of the object,
```
electrachong: They are separate options. You should only specify one.
`versioning: true` creates a new version and updates the master version (+ returns the version id of the new version)
`versionId: <versionId>` updates a specific version, and the master version if <versionId> is the latest version.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+  then update the master version with this version.
+- `versionId: <versionId>` update a specific version (for updating version's ACL
+  or remote updates in geo-replication);
+  - if the version identified by `versionId` happens to be the latest version,
```
electrachong: The version id is injected into the (meta)data stored as the value when a version is stored.
Metadata gets the master version, parses the data into JSON, and checks if its version ID matches the `versionId` being updated. If it matches, it will update both.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
```
electrachong: When versioning is suspended, to overwrite the master version only if the master version is null.

DESIGN.md
electrachong: I tried to make a distinction between what happens in Metadata and what happens in S3, so I would prefer to keep them separate. I can make a mention as suggested that all of this happens before responding to S3.

DESIGN.md
electrachong: 👍

DESIGN.md
electrachong: 👍

DESIGN.md
electrachong: I think it was used to simplify representation, instead of using the null character, similar to how `v2` is used to represent an actual version id. It's used in other examples as well. I added 'represented by'.

DESIGN.md
electrachong: 👍

DESIGN.md
electrachong: I wasn't sure what you had in mind to add, so for now I broke down the related steps in Delete Object API and Multi Object Delete and detailed the checks the other APIs must perform for delete markers.

DESIGN.md
```diff
@@ -28,6 +28,196 @@ daemon is spawned to handle future requests in its stead.
...
+##### DELETE
+
+-  no options: original DELETE operation, will delete the master version
+- `versionId: <versionId>` delete a specific version
+
+Version-specific deletions involve a few steps. A deletion targeting the latest
+version of an object has to:
+
+- delete the specified version identified by `versionId`
+- replace the master version with a place holder for deletion version
```
electrachong: I added some more detail; I'm not sure it's important to people who are modifying S3 though.

electrachong:
@LaurenSpiegel I missed your overall comment regarding the metadata attributes and the steps. I'll try to address those tomorrow if I get a chance.

docs/versioning.md
```diff
@@ -0,0 +1,463 @@
...
+API which uses the same operations of the standard API, but with versioning
+information specified. For example, in order to get the version `v1` of `foo`,
+users have to specify that version in the query, such as
+`GET /foo?versionId=v1`.
+
+A bucket can be non-versioning or versioning. Any objects placed in a bucket
+before versioning is enabled are considered to be the "null" version. When
+versioning is enabled, the "null" version will be kept and continue to be
+identified as the "null" version. However, if versioning is suspended, the
+"null" version may be overwritten with a new "null" version. All versions with
```
dora-korpar: Under what circumstances would that happen? And what would the new "null" version be overwritten with?

electrachong: When versioning is suspended, any existing null version will be overwritten when an object of the same name is put, resulting in a new null version. The new null version overwrites the existing null version. Does it seem clearer or is it still unclear?

docs/versioning.md
```diff
@@ -0,0 +1,463 @@
...
+| foo |   v1    |   A   |
+
+To work with all existing versions of an object, S3 provides an extended
+API which uses the same operations of the standard API, but with versioning
+information specified. For example, in order to get the version `v1` of `foo`,
+users have to specify that version in the query, such as
+`GET /foo?versionId=v1`.
+
+A bucket can be non-versioning or versioning. Any objects placed in a bucket
+before versioning is enabled are considered to be the "null" version. When
```
nicolas2bert: It might be my english but since "objects" is plural, I thought we could have multiple "null" versions.

docs/versioning.md
nicolas2bert: We talked about it. Maybe you want to change `overwriting` with something like  `deleting the old "null" version and create a new "null" version`?

docs/versioning.md
```diff
@@ -0,0 +1,463 @@
...
+version identifiers as the suffix for the keys of the object versions, and a
+special version to represent the latest version. An example of what these
+keys may look like in metadata, for an object `foo/bar` with three versions:
+
+| key |
+|-----|
+| foo/bar |
+| foo/bar\098506163554375999999PARIS  0.a430a1f85c6ec |
+| foo/bar\098506163554373999999PARIS  0.41b510cd0fdf8 |
+| foo/bar\098506163554373999998PARIS  0.f9b82c166f695 |
```
dora-korpar: Is one of these the special version you mentioned? If so, how can you tell? :)

docs/versioning.md
```diff
@@ -0,0 +1,463 @@
...
+- `object_name`: is the key of the object in metadata
+- `separator`: we use the `null` character (`0x00` or `\0`) as the separator
+  between the `object_name` and the `version_id` of a key
+- `version_id`: is the version identifier; this encodes the ordering
+  information in the texture format as metadata orders keys alphabetically
+
+The format of a `version_id` is: `ts` `rep_group_id` `seq_id`; `ts` is the
+combination of epoch and an increasing number, `rep_group_id` is the name of
+data center or a site (or a region AWS equivalently) used for replication later,
+and `seq_id` is supposed to be a unique value based on metadata information.
```
dora-korpar: Might help continuity if this paragraph was formatted similarly to the one above? Unless you think 2 lists in a row would be confusing

docs/versioning.md
dora-korpar: Yeah, I think it's clarified further down in the doc :)

docs/versioning.md
electrachong: The key `foo/bar` is a special version because it's not followed by the version id suffix, but I agree it's not very clear for readers at this point. I added a note to make it clearer and alluded to where it's described in more detail in the section "Master Versions"

docs/versioning.md
electrachong: I reformatted it to appeal to the parallel structure gods :P

docs/versioning.md
electrachong: I think we can change it to singular if that's less confusing. I updated it

docs/versioning.md
electrachong: It depends on the perspective you look at it. Without knowing the details of implementation and from the perspective of someone using the S3 api, we want them to perceive what happens simply as overwriting the null version. (You can access the new version with the same version ID, it just also becomes the latest version if the null version was not the latest version before)

Although behind the scenes what happens is sometimes the null version is overwritten, other times the old version is deleted and a new version is created, and we're trying to describe what happens in this document, since we're at the overview phase here and haven't gotten to details of implementation, does it make sense to stick to the current wording?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+## AWS S3 Bucket Versioning
+
+An object in a versioning enabled bucket can have multiple versions, whereas an
+object can have only a single version in a standard bucket without versioning.
+
+With the standard S3 API, any update (PUT, DELETE) to an object in a versioning
+enabled bucket would create a new version of that object. Read operations such
+as GET and LIST would see the latest version of the objects. The table below
+shows such example, a `PUT foo` generates a new version `v2` and keeps `v1`;
```
LaurenSpiegel: shows an example on a standard bucket: a `PUT foo`...

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+when there are new puts on the object.
+
+## Implementation of Bucket Versioning in S3
+
+### Roles of Metadata Engine and API Component
+
+Each version of an object is stored as a separate key in metadata. The API
+component of S3 manages the put and delete of version data, sends instructions
+to create, retrieve, overwrite or delete a version's metadata to the metadata
+backend, and tracks extra information about a version, such as whether it is a
```
LaurenSpiegel: let's be more precise about "tracks".  S3 has no state so it is not keeping any information.

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+The following is a summary of the design and implementation details of bucket
+versioning, described first for the metadata engine and then for the API.
+
+## Implementation of Bucket Versioning in Metadata
+
+### Version Id Format
+
+The version id is generated by the metadata backend, and encoded in a
+hexadecimal string format by S3. The following information is encoded before
```
LaurenSpiegel: The version id is encoded before sending a response to a request.

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+sending a response to a request. S3 also decodes the hexadecimal string
+received from a request before sending to metadata to retrieve a particular
+version.
+
+`object_name separator version_id`
```
LaurenSpiegel: what is this supposed to represent?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+where:
+
+- `object_name`: is the key of the object in metadata
+- `separator`: we use the `null` character (`0x00` or `\0`) as the separator
+  between the `object_name` and the `version_id` of a key
+- `version_id`: is the version identifier; this encodes the ordering
+  information in the texture format as metadata orders keys alphabetically
```
LaurenSpiegel: what is texture format?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+The format of a `version_id` is:
+`ts` `rep_group_id` `seq_id`
+where:
+
+- `ts`: is the combination of epoch and an increasing number
+- `rep_group_id`: is the name of deployment(s) considered one unit used for
+   replication later
```
LaurenSpiegel: I would get rid of "later"

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+- `seq_id`: is supposed to be a unique value based on metadata information.
+
+An example of a key: `foo\01234567890123456PARIS 1234.123456` indicating that
+this specific version of `foo` was the `123456`th entry created during the epoch
+`1234567890` in the replication group `PARIS` with `1234.1234566` as `seq_id`.
```
LaurenSpiegel: it would be clearer if you did not use 123456 twice in the example

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+### Master Version
+
+We store a copy of the latest version of an object's metadata using
+`object_name` as the key; this version is called the master version. The master
+versions of the objects facilitate the standard GET operation, which would
```
LaurenSpiegel: The master version of each object facilitates the standard GET and LIST operations, which would..

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+engine can handle. The process for creating new versions or delete markers has
+additional steps and is described more comprehensively in the section
+"Implementation of Bucket Versioning in API" and its subsections.
+
+#### PUT
+
+- no options: original PUT operation, will update the master version
+- `versioning: true` create a new version of the object,
+  then update the master version with this version.
+- `versionId: <versionId>` update a specific version (for updating version's ACL
```
LaurenSpiegel: or updating object tags?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+Note also that these describe only the basic CRUD operations that the metadata
+engine can handle. The process for creating new versions or delete markers has
+additional steps and is described more comprehensively in the section
+"Implementation of Bucket Versioning in API" and its subsections.
+
+#### PUT
+
+- no options: original PUT operation, will update the master version
+- `versioning: true` create a new version of the object,
```
LaurenSpiegel: what happens if you have versioning: true and you send a versionID?  please specify.

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+  or remote updates in geo-replication);
+  - if the version identified by `versionId` happens to be the latest version,
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
+    the master version only (same as no options, but returns a version ID)
+
+#### DELETE
+
+- no options: original DELETE operation, will delete the master version
+- `versionId: <versionId>` delete a specific version
```
LaurenSpiegel: so when versioning is enabled, do we always specify a versionId?  meaning the "null" version if that is all there is?  Please specify.

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+### Object Metadata Versioning Attributes
+
+To access all the information needed to properly handle all cases that may exist
+in versioned operations, the API stores certain versioning-related information
+in the metadata attributes of each version's object metadata.
+
+These are the versioning-related metadata properties:
+
+- `isNull`: whether the version being stored is a null version.
+- `nullVersionId`: the version ID of the latest null version.
```
LaurenSpiegel: please specify what you mean here -- what is the internal version id and how created when versioning wasn't enabled yet

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+Null versions are generated when an object is put before versioning is enabled
+or when versioning is suspended, and are distinguished from normal versions in
+that they are overwritten by new null versions. Therefore, only one null version
+should exist at any given time, and it can be identified in S3 requests
+and responses with the version id "null".
+
+With respect to metadata, since the null version is overwritten by subsequent
+null versions, the null version is initially stored in the master key alone,
+as opposed to being stored in the master key and a new version, like how normal
+versions are stored. The metadata engine does not differentiate between null and
```
LaurenSpiegel: I'd get rid of "like how normal versions are stored."  Seem unnecessary and a bit confusing.

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+normal versions, so S3 checks if versioning is suspended or has never been
+configured, and sets the `versionId` option to `''` in PUT calls to the metadata
+engine.
```
LaurenSpiegel: when...?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+- no options: original PUT operation, will update the master version
+- `versioning: true` create a new version of the object,
+  then update the master version with this version.
+- `versionId: <versionId>` update a specific version (for updating version's ACL
+  or remote updates in geo-replication);
+  - if the version identified by `versionId` happens to be the latest version,
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
+    the master version only (same as no options, but returns a version ID)
```
LaurenSpiegel: why do we need this versionID returned?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+(2) PUT foo (second put), versionId: `''`
+
+| key | value |
+|-----|:-----:|
+| foo (null) |   B   |
+
+However, null versions are preserved when versioning has been enabled or
+re-enabled, and new non-null versions are created. To preserve the null version
+if the null version is the master version, the the S3 API stores the current
```
LaurenSpiegel: "the" is repeated

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+Then the master version is overwritten with the new null version:
+
+(4B) PUT foo, versionId: `''`
+
+| key | value |
+|-----|:-----:|
+| foo (null) |   D   |
+| foo.v2 |   C   |
+
+Note that S3 keeps track of the internally-used version ID of the current null
```
LaurenSpiegel: please be careful of "S3 tracks".  S3 stores this information in metadata as...

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+version in the `nullVersionId` metadata attribute for each new version.
+In addition, to determine whether the null version is the latest version, the S3
+API sets the `isNull` attribute for each version. The `nullVersionId` attribute
```
LaurenSpiegel: how does setting 'isNull' help determine whether it is the latest version?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+is also used to retrieve the correct version when the version ID "null" is
+specified in certain object-level APIs, described further in the section
+"Null Version Mapping".
+
+#### Specifying Versions in APIs for Putting Versions
+
+Since S3 does not currently allow overwrite of existing version data, Put
```
LaurenSpiegel: why currently?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+Object, Complete Multipart Upload and Copy Object return `400 InvalidArgument`
```
LaurenSpiegel: Does this differ from the aws api in some way?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+### PUT Example
+
+When S3 receives a request to PUT an object:
+
+- It checks first if versioning has been configured
+- If it has not been configured, S3 proceeds to delete any pre-existing data and
+  put the new object as the master version in Metadata
+
+If it has been configured, S3 checks the following:
```
LaurenSpiegel: If versioning has been enabled at any point, S3 checks...

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+Object, Complete Multipart Upload and Copy Object return `400 InvalidArgument`
+if a specific version ID is specified in the request query, e.g. for a
+`PUT /foo?versionId=v1` request.
+
+### PUT Example
+
+When S3 receives a request to PUT an object:
+
+- It checks first if versioning has been configured
+- If it has not been configured, S3 proceeds to delete any pre-existing data and
```
LaurenSpiegel: It actually puts the new data, puts the metadata and then deletes any pre-existing data

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+- If the master version is a null version or has no version ID (put before
+    versioning was configured):
+    - store the null version metadata as a new version
+    - create a new version and overwrite the master version
+        - set `nullVersionId`: version ID returned when storing null version
+
+If versioning is enabled and the master version is not null;
+or there is no existing object metadata:
+
+- create a new version and store it
```
LaurenSpiegel: and overwrite the master version?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+    - delete previous object data
+- If master is not a null version and `nullVersionId` is defined in the object’s
+  metadata:
+    - delete the current null version metadata and data
+    - overwrite the master version with the new metadata
+
+If there is no existing object metadata, create the new null version as the
+master version.
+
+In each of the above cases, set `isNull` metadata attribute to true when when
```
LaurenSpiegel: "when when"

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+
+### Behavior of Object-Targeting APIs
+
+API methods which can target existing objects or versions, such as Get Object,
+Head Object, Get Object ACL, Put Object ACL, Copy Object and Copy Part, will
+perform the action on the latest version of an object if no version ID is
+specified in the request query or relevant request header
+(`x-amz-copy-source-version-id` for Copy Object and Copy Part APIs).
+
+One exception are the Delete Object and Multi-Object Delete APIs, which will
```
LaurenSpiegel: Two exceptions...

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+instead attempt to create delete markers, described in the following section, if
+no version ID is specified.
+
+No versioning options are necessary to retrieve the latest version from
+metadata, since the master version is stored in a key with the name of the
+object. However, when updating the latest version, such as with the Put Object
+ACL API, S3 sets the `versionId` option in the PUT call to metadata
+to the value stored in the object metadata's `versionId` attribute. This is done
+in order to update the metadata both in the master version and the version
+itself, if it is not a null version.
```
LaurenSpiegel: what if it is a null version?

docs/versioning.md
```diff
@@ -0,0 +1,488 @@
...
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
+    the master version only (same as no options, but returns a version ID)
+
+#### DELETE
+
+- no options: original DELETE operation, will delete the master version
+- `versionId: <versionId>` delete a specific version
+
+Version-specific deletions involve a few steps. A deletion targeting the latest
```
bennettbuchanan: You might define 'Version-specific' to make the term extra clear. 😄

docs/versioning.md
electrachong: updated to specify what happens (metadata ignores the request)

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: to return a version ID for null versions. I updated the information to include this

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: Updated to avoid the term

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: It's the metadata key but it wasn't clear. I updated the description.

docs/versioning.md
electrachong: I mentioned before that it was in the original document. I assumed it meant the format used for the version ID described below. Since it's not clear, I removed the term.

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: I think this is described in the "Delete Markers" section. A version ID is always specified to delete a version. For the null version, the version ID "null" would be used. No version ID would result in creating a delete marker. Let me know if it's not clear enough from that section.

docs/versioning.md
electrachong: I updated the description (was inaccurate as it was), and explained further in the section "Overwriting a Null Version That is Not Latest Version"

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: updated

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: I updated this section to explain how the isNull attribute is used to check if the null version is the latest version.

docs/versioning.md
electrachong: There's some code in the versioningPreprocessing function that deals with scenarios where we would be overwriting existing version data. I think we wrote it in case it would be needed for bucket replication, but at the moment we return 400 InvalidArgument before the request reaches versioningPreprocessing.

docs/versioning.md
electrachong: It's following AWS behavior.

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: It's possible versioning was never enabled, but always suspended from when it was first configured. I changed it to "If versioning has been configured..."

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: I updated it

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: Initially I assumed it would only update the master version. But I checked with Vinh, and it will create a new version with that versionId if it does not exist. We may need to add a check for this case in objectPutAcl and objectPutTagging.

docs/versioning.md
electrachong: We need this version ID to be returned when storing null versions because we may need to store the null version again separately, but with the same internal version ID if possible. (If there are concurrent requests for a null version, the null version should be overwritten instead of generating multiple versions.)

docs/versioning.md
LaurenSpiegel: you mean it creates an internal version number for the null version?  yes, please add tests.

docs/versioning.md
```diff
@@ -0,0 +1,566 @@
...
+
+This section describes S3 Server's support for the specific AWS S3 Bucket
+Versioning feature.
+
+## AWS S3 Bucket Versioning
+
+An object in a versioning enabled bucket can have multiple versions, whereas an
+object can have only a single version in a standard bucket without versioning.
+
+With the standard S3 API, any update (PUT, DELETE) to an object in a versioning
```
jonathan-gramain: It could be worth mentioning in this paragraph that a DELETE on a versioned bucket creates delete markers in place of deleting the data, to precise the "create a new version" for the DELETE case, as it seems the first mention of delete markers is far below in this document.

docs/versioning.md
electrachong: working on this here: https://github.com/scality/S3/pull/748

docs/versioning.md
electrachong: I updated to follow your suggestion

docs/versioning.md
```diff
@@ -0,0 +1,563 @@
...
+where:
+
+- `object_name`: is the key of the object in metadata
+- `separator`: we use the `null` character (`0x00` or `\0`) as the separator
+  between the `object_name` and the `version_id` of a key
+- `version_id`: is the version identifier; this encodes the ordering
+  information in the format described above as metadata orders keys
+  alphabetically
+
+An example of a key in metadata: `foo\01234567890000777PARIS 1234.123456`
```
LaurenSpiegel: is there a space between the rep_group_id and the seq_id?  if so, please state that above in the description of the format.

docs/versioning.md
```diff
@@ -0,0 +1,563 @@
...
+  - if the version identified by `versionId` happens to be the latest version,
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
+    the master version only (same as no options, but returns a version ID).
+    This is used for creating null versions, which are discussed further in
+    "Null Version Management."
+
+Only one option is used at a time. `versionId: <versionId>` does not have to be
+used with `versioning: true` set to work, nor should they both be set. If both
+are used at once, the metadata engine will not perform any action.
```
LaurenSpiegel: what does that mean -- "no action"? Does it return an error?  What does it return?

docs/versioning.md
LaurenSpiegel: so I would delete "currently"

docs/versioning.md
```diff
@@ -0,0 +1,563 @@
...
+- `versioning: true` create a new version of the object,
+  then update the master version with this version.
+- `versionId: <versionId>` update a specific version (for updating version's ACL
+  or tags, or remote updates in geo-replication);
+  - if the version identified by `versionId` happens to be the latest version,
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
+    the master version only (same as no options, but returns a version ID).
+    This is used for creating null versions, which are discussed further in
+    "Null Version Management."
```
LaurenSpiegel: it will overwrite the master version only - same as no options, but the master version will have a versionId property that will be set just like any other version Id but never exposed to an external user. Storing this internal only versionID enables S3 to find this version later if it is no longer the master. This option of `versionId` set to '' is used for creating null versions once versioning has been suspended, which are...

docs/versioning.md
electrachong: I think the space is actually due to padding the rg_token. The padding character for the rg_token is `' '`.

docs/versioning.md
electrachong: Updated

docs/versioning.md
electrachong: I updated; metadata returns `BadRequest`

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+  version metadata based on bucket versioning state and version metadata
+- encode version ID information to return in responses to requests, and decode
+  version IDs sent in requests
+
+The implementation of Bucket Versioning in S3 is described in this document in
+two main parts. The first section, "Implementation of Bucket Versioning in
+Metadata" describes the way versions are stored in metadata, and the metadata
+options for manipulating version metadata.
+
+The second section, "Implementation of Bucket Versioning in API" describes the
```
bennettbuchanan: Could be nice to have an internal link to these sections.

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+way the metadata options are used in the API within S3 actions to create new
+versions, update their metadata, and delete them. The management of null
+versions and creation of delete markers are also described in this section.
+
+## Implementation of Bucket Versioning in Metadata
+
+As mentioned above, each version of an object is stored as a separate key in
+metadata. We use version identifiers as the suffix for the keys of the object
+versions, and a special version (the "Master Version") to represent the latest
```
bennettbuchanan: ~"Master Version"~ master version

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+| key |
+|-----|
+| foo/bar |
+| foo/bar\098506163554375999999PARIS  0.a430a1f85c6ec |
+| foo/bar\098506163554373999999PARIS  0.41b510cd0fdf8 |
+| foo/bar\098506163554373999998PARIS  0.f9b82c166f695 |
+
+The most recent version created is represented above in the key `foo/bar` and is
+the master version. This special version is described further in the subsection
+"Master Version". The subsection that immediately follows describes the metadata
```
bennettbuchanan: This sentence "The subsection..." seems a bit redundant.

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+metadata to retrieve a particular version.
+
+The format of a `version_id` is:
+`ts` `rep_group_id` `seq_id`
+where:
+
+- `ts`: is the combination of epoch and an increasing number
+- `rep_group_id`: is the name of deployment(s) considered one unit used for
+   replication
+- `seq_id`: is supposed to be a unique value based on metadata information.
```
bennettbuchanan: ~is supposed to be~

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+An example of a key in metadata: `foo\01234567890000777PARIS 1234.123456`
+indicating that this specific version of `foo` was the `000777`th entry created
+during the epoch `1234567890` in the replication group `PARIS` with
+`1234.123456` as `seq_id`.
+
+### Master Version
+
+We store a copy of the latest version of an object's metadata using
+`object_name` as the key; this version is called the master version. The master
+versions of each object facilitates the standard GET operation, which would
```
bennettbuchanan: Sounds to me like each object has just one master version, no? If so: ~versions~ version

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+
+S3 Server sends instructions to the metadata engine about whether to create a
+new version or overwrite, retrieve, or delete a specific version by sending
+values for special options in PUT, GET, or DELETE calls to metadata. The
+metadata engine can also list versions in the database, which is used by S3
+to list object versions.
+
+These only describe the basic CRUD operations that the metadata engine can
+handle. How these options are used by the S3 API to generate and update versions
+involves more steps and is described more comprehensively in "Implementation of
```
bennettbuchanan: ~involves more steps and~

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+Bucket Versioning in API" and its subsections.
```
bennettbuchanan: ~and its subsections~ I think this is implied.

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+
+Note: all operations (PUT and DELETE) that generate a new version of an object
+will return the `version_id` of the new version to the API.
+
+#### PUT
+
+- no options: original PUT operation, will update the master version
```
bennettbuchanan: unneeded comma

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+  then update the master version with this version.
+- `versionId: <versionId>` update a specific version (for updating version's ACL
+  or tags, or remote updates in geo-replication);
+  - if the version identified by `versionId` happens to be the latest version,
+    the master version will be updated as well
+  - note that with `versionId` set to an empty string `''`, it will overwrite
+    the master version only (same as no options, but the master version will
+    have a `versionId` property set in its metadata like any other version).
+    The `versionId` will never be exposed to an external user, but setting this
+    internal-only versionID enables S3 to find this version later if it is no
```
bennettbuchanan: Some of these references to versionID are in in-line code, others aren't.

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+    null versions once versioning has been suspended, which is discussed
+    further in "Null Version Management."
+
+Only one option is used at a time. `versionId: <versionId>` does not have to be
+used with `versioning: true` set to work, nor should they both be set. If both
+are used at once, the metadata engine will return an error.
+
+To summarize the valid combinations of versioning options:
+
+- !versioning && !versionId: normal non-versioning put
```
bennettbuchanan: ~put~ PUT

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+  - if versionId === '' update master version
+
+Other cases are invalid and the metadata engine returns the error `BadRequest`.
+
+#### DELETE
+
+- no options: original DELETE operation, will delete the master version
+- `versionId: <versionId>` delete a specific version
+
+Deleting a specific version in metadata involves a few steps. A deletion
```
bennettbuchanan: ~Deleting a specific version in metadata involves a few steps.~

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+targeting the latest version of an object has to:
+
+- delete the specified version identified by `versionId`
+- replace the master version with a version that is a placeholder for deletion
+  - this version contains a special keyword, 'isPHD', to indicate the master
+  version was deleted and needs to be updated
+- initiate a repair operation to update the value of the master version:
+  - involves listing the versions of the object and get the latest
+  version to replace the place holder delete version
```
bennettbuchanan: ~place holder~ placeholder

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+  - if no more versions exists, metadata deletes the master version, removing
```
bennettbuchanan: ~exists~ exist

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+
+The implementation of a GET operation does not change compared to the standard
+version. A standard GET without versioning information would get the master
+version of a key. A version-specific GET would retrieve the specific version
+identified by the key for that version.
+
+#### LIST
+
+For a standard LIST on a bucket, metadata iterates through the keys by using
+the separator represented by `.` as an extra delimiter. For a listing of all
```
bennettbuchanan: This is restated just below, just before the example.

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+versions of a bucket, there is no change compared to the original listing
+function. Instead, the API component returns all the keys in a List Objects call
+and filters for just the keys of the master versions in a List Object Versions
+call.
+
+For example, a standard LIST operation against the keys in a table below would
+return from metadata the list of `[ foo/bar, bar, qux/quz, quz ]` using the
+separator represented by `.` as the extra delimiter.
```
bennettbuchanan: Just to be sure, you are using a dot `.` just in this design document for the sake of brevity?

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+
+To prevent issues with concurrent requests, S3 ensures the null version is
+stored with the same version ID by using `versionId` option. S3 sets the
+`versionId` option to the master version's `versionId` metadata attribute value
+during the PUT. This creates a new version with the same version ID of the
+existing null master version.
+
+The null version's `versionId` attribute may be undefined because it was
+generated before the bucket versioning was configured. In that case, a version
+ID is generated using the max timestamp and sequence values possible so that the
```
bennettbuchanan: What is the max timestamp?

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+null version will be properly ordered as the last entry in a metadata listing.
+This value ("non-versioned object id") is used in the PUT call with the
+`versionId` option.
+
+##### Case 3: Overwriting a Null Version That is Not Latest Version
+
+Note that in the case versioning is suspended and there is a null version is not
+the latest version, S3 cannot rely on the `versionId: ''` option during a PUT to
```
bennettbuchanan: Probably could reword the opening sentence.

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+| foo (null) |   D   |
+| foo.v2 |   C   |
+
+The `nullVersionId` attribute is also used to retrieve the correct version when
+the version ID "null" is specified in certain object-level APIs, described
+further in the section "Null Version Mapping".
+
+#### Specifying Versions in APIs for Putting Versions
+
+Since S3 does not allow overwrite of existing version data, Put Object,
```
bennettbuchanan: ~allow overwrite~ allow an overwrite

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+
+If versioning has not been configured for a bucket, the Delete Object and
+Multi-Object Delete APIs behave as their standard APIs.
+
+If versioning has been configured, S3 deletes object or version data only if
+a specific version ID is provided in the request query, e.g.
+`DELETE /foo?versionId=v1`.
+
+If no version ID is provided, S3 creates a delete marker by creating a 0-byte
+version with the metadata attribute `isDeleteMarker: true`.The S3 API will
```
bennettbuchanan: ~.The~ . The

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+return a `404 NoSuchKey` error in response to requests getting or heading an
+object whose latest version is a delete maker.
+
+To restore a previous version as the latest version of an object, the delete
+marker must be deleted, by the same process as deleting any other version.
```
bennettbuchanan: extra comma

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
+# Versioning
+
+This document describes S3 Server's support for the specific AWS S3 Bucket
```
electrachong: 👍

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+Versioning feature.
+
+## AWS S3 Bucket Versioning
+
+See AWS documentation on Versioning for a description of the Bucket Versioning
```
electrachong: 👍

docs/versioning.md
```diff
@@ -0,0 +1,567 @@
...
+
+### Overview of Metadata and API Component Roles
+
+Each version of an object is stored as a separate key in metadata. The S3 API
+interacts with the metadata backend to store, retrieve, and delete version
+metadata.
+
+The implementation of versioning within the metadata backend is naive. The
+metadata backend does not evaluate any information about bucket or version state
+(whether versioning is enabled or suspended, and whether a version is a null
```
electrachong: Initially the document described most of the feature, including null versions, in the beginning and throughout the document. But we decided it was a lot of irrelevant information to filter through for those who know about the versioning feature and just want to learn how it is implemented. So we removed that information and instead added the links to AWS's description of the feature.

docs/versioning.md
electrachong: I agree, I added the anchor links now.

docs/versioning.md
electrachong: I added a link so it's clearer that the quotes are referencing a specific section.

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: I think it's a helpful comma :)
It means: "This is the original PUT operation. Will update the master version"

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: Thanks, I left it here and removed it later.

docs/versioning.md
electrachong: Yes, I think it's a little clearer to read when using the `.` in the examples than putting the actual `\0` as long as it's understood the `.` is representative of the actual separator (`\0`).

docs/versioning.md
electrachong: It's the max epoch, part of the `ts` value described for the `versionId` format.

docs/versioning.md
electrachong: Tried to reword it, is it clearer?

docs/versioning.md
electrachong: 👍

docs/versioning.md
electrachong: 👍

PR #713, opened 2017-05-02
fix: add objAcl checks for delete marker [S3C-184]

electrachong:
We were not checking for delete markers when responding to put and get acl requests.

### DIFF
--- a/lib/api/objectGetACL.js
+++ b/lib/api/objectGetACL.js
@@ -91,6 +91,16 @@ export default function objectGetACL(authInfo, request, log, callback) {
                         { method: 'objectGetACL', error: err });
                         return next(err, bucket);
                     }
+                    if (objectMD.isDeleteMarker) {
+                        if (versionId) {
+                            log.trace('requested version is delete marker',
+                            { method: 'objectGetACL' });
+                            return next(errors.MethodNotAllowed);
+                        }
+                        log.trace('most recent version is delete marker',
+                        { method: 'objectGetACL' });
+                        return next(errors.NoSuchKey);
+                    }
                     return next(null, bucket, objectMD);
                 });
         },
--- a/lib/api/objectPutACL.js
+++ b/lib/api/objectPutACL.js
@@ -113,6 +113,11 @@ export default function objectPutACL(authInfo, request, log, cb) {
                             errors.NoSuchKey;
                         return next(err, bucket);
                     }
+                    if (objectMD.isDeleteMarker) {
+                        log.trace('delete marker detected',
+                        { method: 'objectPutACL' });
+                        return next(errors.MethodNotAllowed, bucket);
+                    }
                     return next(null, bucket, objectMD);
                 });
         },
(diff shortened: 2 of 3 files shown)


lib/api/objectGetACL.js
```diff
@@ -91,6 +91,14 @@ export default function objectGetACL(authInfo, request, log, callback) {
                         { method: 'objectGetACL', error: err });
                         return next(err, bucket);
                     }
+                    if (objectMD.isDeleteMarker) {
+                        log.trace('delete marker detected',
```
LaurenSpiegel: I think 'most recent version is a delete marker' is clearer for case where return NoSuchKey and a separate log stating 'requested version is a delete marker' when return MethodNotAllowed.

lib/api/objectPutACL.js
```diff
@@ -113,6 +113,11 @@ export default function objectPutACL(authInfo, request, log, cb) {
                             errors.NoSuchKey;
                         return next(err, bucket);
                     }
+                    if (objectMD.isDeleteMarker) {
+                        log.trace('delete marker detected',
```
LaurenSpiegel: So, AWS is not consistent between error messages for get and put acl?

tests/functional/aws-node-sdk/test/versioning/objectACL.js
```diff
@@ -14,14 +14,16 @@ const counter = 100;
 let bucket;
 const key = '/';
 const invalidId = 'invalidId';
-const nonExistingId = '3939393939393939393936493939393939393939756e6437';
+const nonExistingId = process.env.AWS_ON_AIR ?
+    'MhhyTHhmZ4cxSi4Y9SMe5P7UJAz7HLJ9' :
```
LaurenSpiegel: why? is this just to have the correct format for aws?  please put comment

electrachong:
@ironman-machine try

lib/api/objectGetACL.js
electrachong: 👍

lib/api/objectPutACL.js
electrachong: Yes. Although, I think it might make more sense if we return 404 NoSuchKey if someone tries to put an acl on an object whose latest version is a delete marker without specifying the version id.

tests/functional/aws-node-sdk/test/versioning/objectACL.js
electrachong: I added a comment. It's to have the correct format when running against AWS, otherwise tests for non-existing ids will get InvalidArgument instead of NoSuchKey.

lib/api/objectPutACL.js
LaurenSpiegel: we should do what AWS does but it's not pretty

electrachong:
end-to-end passed: http://ci.ironmann.io/gh/scality/Integration/12652

PR #714, opened 2017-05-03
Ft/rpc service rest api

jonathan-gramain:
# Pull request template

## Description

Add params to enable metadata HTTP server (Arsenal PR https://github.com/scality/Arsenal/pull/252)

### Motivation and context

Ease debugging by offering an optional HTTP interface that allows access to internal metadata through HTTP POST requests.

This is disabled by default and not documented for now, it can crash the daemon if requests are not properly forged, so it's not meant to be enabled in any production system, but it can help troubleshooting and development by allowing to use curl or other standard HTTP tools instead of a socket.io client.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -364,6 +364,10 @@ class Config {
             this.metadataDaemon.metadataPath =
                 process.env.S3METADATAPATH ?
                 process.env.S3METADATAPATH : `${__dirname}/../localMetadata`;
+
+            this.metadataDaemon.restEnabled =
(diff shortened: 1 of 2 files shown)


jonathan-gramain:
@ironman-machine try

jonathan-gramain:
Integration tests passed in http://ci.ironmann.io/gh/scality/Integration/13704

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

PR #716, opened 2017-05-04
rf: clean up listing types, parsing [S3C-184]

electrachong:
No longer use 'Basic' listing type.
Move JSON parsing for 'DelimiterVersions' to metadata wrapper to be consistent with what we do for other listing types.
Add some more assertions to listing tests.

Note that it seems we still use 'Delimiter' listing type for listing buckets and for internal bucket listing, such as listing objects inside a mpuBucket.

### DIFF
--- a/docs/listing.md
+++ b/docs/listing.md
@@ -1,0 +1,20 @@
+# Listing
+
+## Listing Types
+
+We use three different types of metadata listing for various operations.
+Here are the scenarios we use each for:
+
+- 'Delimiter' - when no versions are possible in the bucket since it is an
+   internally-used only bucket which is not exposed to a user. Namely,
+  1. to list objects in the "user's bucket" to respond to a GET SERVICE
+  request and
+  2. to do internal listings on an MPU shadow bucket to complete multipart
+  upload operations.
+- 'DelimiterVersion' - to list all versions in a bucket
+- 'DelimiterMaster' - to list just the master versions of objects in a bucket
+
+## Algorithms
+
+The algorithms for each listing type can be found in the open-source
+[scality/Arsenal](https://github.com/scality/Arsenal) repository, in [lib/algos/list](https://github.com/scality/Arsenal/tree/master/lib/algos/list).
--- a/lib/metadata/metadataUtils.js
+++ b/lib/metadata/metadataUtils.js
@@ -8,6 +8,56 @@ import { isBucketAuthorized,
     isObjAuthorized } from '../api/apiUtils/authorization/aclChecks';
 import bucketShield from '../api/apiUtils/bucket/bucketShield';
 
+/** _parseListEntries - parse the values returned in a listing by metadata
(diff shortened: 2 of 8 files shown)


lib/metadata/metadataUtils.js
```diff
@@ -8,6 +8,33 @@ import { isBucketAuthorized,
     isObjAuthorized } from '../api/apiUtils/authorization/aclChecks';
 import bucketShield from '../api/apiUtils/bucket/bucketShield';
 
+exports.parseListEntries = entries =>
+    entries.map(entry => {
+        const tmp = JSON.parse(entry.value);
```
bennettbuchanan: Should the parsing be in a try...catch.

tests/functional/aws-node-sdk/test/versioning/listObjectMasterVersions.js
```diff
@@ -8,6 +8,24 @@ import { removeAllVersions } from '../../lib/utility/versioning-util';
...
+        'ETag',
+        'Size',
+        'Owner',
+        'StorageClass',
+    ];
+    elements.forEach(elem => {
+        assert.notStrictEqual(entry[elem], undefined,
+            `Expected ${elem} in result but did not find it`);
+        if (elem === 'Owner') {
+            assert(entry.Owner.ID, 'Expected Owner ID but did not find it');
```
bennettbuchanan: Could an ID be `0`?

lib/metadata/metadataUtils.js
electrachong: I guess it's better to be prudent. I added another commit, can you check and see if it's OK?

electrachong:
@ironman-machine

electrachong:
@ironman-machine

docs/listing.md
```diff
@@ -0,0 +1,15 @@
+# Listing
+
+## Listing Types
+
+We use three different types of metadata listing for various operations.
+Here are the scenarios we use each for:
+
+- 'Delimiter' - when no versions are involved, such as listing buckets
+- 'DelimiterVersion' - to list all versions belonging to a bucket
```
LaurenSpiegel: all versions in a bucket

docs/listing.md
```diff
@@ -0,0 +1,15 @@
+# Listing
+
+## Listing Types
+
+We use three different types of metadata listing for various operations.
+Here are the scenarios we use each for:
+
+- 'Delimiter' - when no versions are involved, such as listing buckets
```
LaurenSpiegel: when no versions are possible in the bucket since it is a bucket only used internally and not exposed to a user. Namely, (1) to list objects in the "user's bucket" to respond to a GET SERVICE request and (2) to do internal listings on an MPU shadow bucket to complete multipart upload operations.

electrachong:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/12794

PR #717, opened 2017-05-05
FT: getObjectTagging

nicolas2bert:
UTAPI: https://github.com/scality/utapi/pull/128 **MERGED**
ARSENAL: https://github.com/scality/Arsenal/pull/254 **MERGED**
INTEGRATION: https://github.com/scality/Integration/pull/520


electrachong:
I'm sure you must be versioning's biggest fan by now with all these tests you are doing related to it :D 

Can you do two more tests, for getting tagging of delete markers with / without version ID specified? Should be very similar to the PUT tagging to delete marker tests you have so it shouldn't be difficult.

lib/api/objectGetTagging.js
```diff
@@ -0,0 +1,88 @@
...
+        });
+        return process.nextTick(() => callback(decodedVidResult));
+    }
+    const reqVersionId = decodedVidResult;
+
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        objectKey,
+        requestType: 'bucketOwnerAction',
```
LaurenSpiegel: why bucketOwner?  I would think that you would need object get acl rights? did we do bucketOwnerAction on put?  If so, that probably needs to be updated.

lib/api/objectGetTagging.js
```diff
@@ -0,0 +1,88 @@
...
+              }
+              return next(null, bucket, objectMD);
+          }),
+        (bucket, objectMD, next) => {
+            const tags = objectMD.tags;
+            const xml = convertToXml(tags);
+            next(null, bucket, xml, objectMD);
+        },
+    ], (err, bucket, xml, objectMD) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
```
LaurenSpiegel: let's not call this corsHeaders since we are also adding the version header to it.  maybe 'additionalResHeaders'?

lib/api/objectGetTagging.js
nicolas2bert: Fixed here: https://github.com/scality/S3/pull/718

PR #719, opened 2017-05-06
test: creating delete markers w/ null ver

electrachong:
AWS has this description in its [REST API](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETE.html) which is different from our own implementation: The DELETE operation removes the null version (if there is one) of an object and inserts a delete marker, which becomes the current version of the object. If there isn't a null version, Amazon S3 does not remove any objects.

These tests were run against AWS and verify despite what the description implies, that our implementation is compatible's with AWS's. The DELETE operation only removes the null version if versioning is suspended, not if it is enabled.

### DIFF
--- a/tests/functional/aws-node-sdk/test/versioning/objectDelete.js
+++ b/tests/functional/aws-node-sdk/test/versioning/objectDelete.js
@@ -4,11 +4,119 @@ import async from 'async';
 import withV4 from '../support/withV4';
 import BucketUtility from '../../lib/utility/bucket-util';
 
-import { removeAllVersions } from '../../lib/utility/versioning-util.js';
+import {
+    versioningSuspended,
+    versioningEnabled,
+    removeAllVersions,
+} from '../../lib/utility/versioning-util.js';
 
 const bucket = `versioning-bucket-${Date.now()}`;
 const key = 'anObject';
 
+function _assertNoError(err, desc) {
+    assert.strictEqual(err, null, `Unexpected err ${desc || ''}: ${err}`);
+}
+
+describe('delete marker creation in bucket with null version', () => {
+    withV4(sigCfg => {
+        const bucketUtil = new BucketUtility('default', sigCfg);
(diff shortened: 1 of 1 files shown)


electrachong:
@LaurenSpiegel updated the tests

electrachong:
@ironman-machine

electrachong:
@ironman-machine

tests/functional/aws-node-sdk/test/versioning/objectDelete.js
```diff
@@ -4,11 +4,119 @@ import async from 'async';
...
+describe('delete marker creation in bucket with null version', () => {
+    withV4(sigCfg => {
+        const bucketUtil = new BucketUtility('default', sigCfg);
+        const s3 = bucketUtil.s3;
+        const nullVersionBody = 'nullversionbody';
+
+        beforeEach(done => {
+            s3.createBucket({ Bucket: bucket }, err => {
+                if (err) {
+                    done(err);
```
nicolas2bert: return done(err) ?

tests/functional/aws-node-sdk/test/versioning/objectDelete.js
```diff
@@ -4,11 +4,119 @@ import async from 'async';
...
+                } // put null object
+                s3.putObject({
```
nicolas2bert: return s3.putObject({ ?

tests/functional/aws-node-sdk/test/versioning/objectDelete.js
electrachong: thanks 👍

tests/functional/aws-node-sdk/test/versioning/objectDelete.js
electrachong: 👍

electrachong:
end to end passed

PR #721, opened 2017-05-09
FT: Handle Utapi metrics for versioning

bennettbuchanan:
Depends on https://github.com/scality/utapi/pull/129.

Update APIs to push correct metrics for changes related to versioning.

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -180,7 +180,7 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
         bucketName, bucket, quietSetting, errorResults, inPlay, log, next) {
     const successfullyDeleted = [];
     let totalContentLengthDeleted = 0;
-    let numOfObjects = 0;
+    let numOfObjectsRemoved = 0;
     const skipError = new Error('skip');
 
     // doing 5 requests at a time. note that the data wrapper
@@ -255,12 +255,9 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
                 return moveOn();
             }
             if (deleteInfo.deleted && objMD['content-length']) {
-                numOfObjects++;
+                numOfObjectsRemoved++;
                 totalContentLengthDeleted += objMD['content-length'];
             }
-            // TODO: update number of objects to reflect creation of
-            // delete markers; Utapi must be updated to allow this for
-            // 'deleteObject' metrics
             let isDeleteMarker;
             let deleteMarkerVersionId;
             // - If trying to delete an object that does not exist (if a new
@@ -271,6 +268,9 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             if (deleteInfo.newDeleteMarker) {
                 isDeleteMarker = true;
                 deleteMarkerVersionId = versionIdUtils.encode(versionId);
+                // In this case we are putting a new object (i.e., the delete
+                // marker), so we decrement the numOfObjectsRemoved value.
+                numOfObjectsRemoved--;
             // If trying to delete a delete marker, DeleteMarkerVersionId equals
             // deleteMarker's versionID and DeleteMarker equals true
             } else if (objMD && objMD.isDeleteMarker) {
@@ -284,8 +284,8 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
     },
     // end of forEach func
     err => {
-        log.trace('finished deleting objects', { numOfObjects });
-        return next(err, quietSetting, errorResults, numOfObjects,
+        log.trace('finished deleting objects', { numOfObjectsRemoved });
+        return next(err, quietSetting, errorResults, numOfObjectsRemoved,
             successfullyDeleted, totalContentLengthDeleted, bucket);
     });
 }
@@ -474,7 +474,7 @@ function multiObjectDelete(authInfo, request, log, callback) {
                     bucketName, bucket, quietSetting, errorResults, inPlay,
                     log, next);
         },
-    ], (err, quietSetting, errorResults, numOfObjects,
+    ], (err, quietSetting, errorResults, numOfObjectsRemoved,
         successfullyDeleted, totalContentLengthDeleted, bucket) => {
         const corsHeaders = collectCorsHeaders(request.headers.origin,
             request.method, bucket);
@@ -487,7 +487,7 @@ function multiObjectDelete(authInfo, request, log, callback) {
             authInfo,
             bucket: bucketName,
             byteLength: totalContentLengthDeleted,
-            numberOfObjects: numOfObjects,
+            numberOfObjects: numOfObjectsRemoved,
         });
         return callback(null, xml, corsHeaders);
     });
--- a/lib/api/objectDelete.js
+++ b/lib/api/objectDelete.js
@@ -140,11 +140,8 @@ export default function objectDelete(authInfo, request, log, cb) {
                 resHeaders['x-amz-version-id'] = result.versionId === 'null' ?
                     result.versionId : versionIdUtils.encode(result.versionId);
             }
-            // TODO metric should be 'deleteObject', not 'putObject',
-            // but number of objects should be incremented since putting a new
-            // version. Utapi will be updated to allow this
-            pushMetric('putObject', log, { authInfo, bucket: bucketName,
-                newByteLength: 0, oldByteLength: 0 });
+            pushMetric('putDeleteMarkerObject', log, { authInfo,
+                bucket: bucketName });
         } else {
             // otherwise, if deleting a specific version or delete marker,
             // return the version id in the response headers
--- a/lib/api/objectPut.js
+++ b/lib/api/objectPut.js
@@ -97,21 +97,28 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
             // * 0 or > 0 - existing object with content-length 0 or > 0
             // objMD here is the master version that we would
             // have overwritten if there was an existing version or object
+            //
+            // TODO: Handle utapi metrics for null version overwrites.
             const oldByteLength = objMD && objMD['content-length']
                 !== undefined ? objMD['content-length'] : null;
-            pushMetric('putObject', log, { authInfo, bucket: bucketName,
-                newByteLength, oldByteLength });
             if (storingResult) {
                 // ETag's hex should always be enclosed in quotes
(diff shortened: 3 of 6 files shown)


bennettbuchanan:
@ironman-machine
SCALITY_UTAPI_BRANCH=S3C-194-handle-metrics-for-versioning
SCALITY_INTEGRATION_BRANCH=S3C-194-handle-metrics-for-versioning
try

lib/api/multiObjectDelete.js
```diff
@@ -269,6 +269,8 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             // return DeleteMarkerVersionId equals the versionID of the marker
             // you just generated and DeleteMarker tag equals true
             if (deleteInfo.newDeleteMarker) {
+                // Adding a delete marker should add one to numberOfObjects
+                numOfObjects--;
```
electrachong: Will multiObjectDelete metric accept a negative value for `numberOfObjects`? There are cases where the number of new delete markers may outnumber the number of deleted objects, resulting in a negative value.

It might also help for clarity's sake to rename this variable to `numOfObjectsRemoved`

lib/api/objectDelete.js
```diff
@@ -140,11 +140,9 @@ export default function objectDelete(authInfo, request, log, cb) {
                 resHeaders['x-amz-version-id'] = result.versionId === 'null' ?
                     result.versionId : versionIdUtils.encode(result.versionId);
             }
-            // TODO metric should be 'deleteObject', not 'putObject',
-            // but number of objects should be incremented since putting a new
-            // version. Utapi will be updated to allow this
-            pushMetric('putObject', log, { authInfo, bucket: bucketName,
-                newByteLength: 0, oldByteLength: 0 });
+            // We increment deleteObject even if deleting a non-existant version
```
electrachong: even if creating a delete marker

lib/api/multiObjectDelete.js
```diff
@@ -269,6 +269,8 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             // return DeleteMarkerVersionId equals the versionID of the marker
             // you just generated and DeleteMarker tag equals true
             if (deleteInfo.newDeleteMarker) {
+                // Adding a delete marker should add one to numberOfObjects
```
jonathan-gramain: Shouldn't the above TODO be removed now?

Overall I think the whole set of comments here should be clarified and grouped into a single explanation block. I'm not sure why we decrement `numOfObjects`, is it because we're supposed to have incremented it when we created the delete marker first? A more explicit comment would be welcome.

lib/api/objectPut.js
```diff
@@ -99,19 +99,24 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
             // have overwritten if there was an existing version or object
             const oldByteLength = objMD && objMD['content-length']
```
electrachong: There are some edge-ish cases where the old byte length which was overwritten by the new object/version is not the content length of objMD (the latest version), i.e. when a null version put when versioning suspended is overwriting a previous null version that is not the latest version.

The logic used for this situation is located in versioningPreprocessing in certain conditionals. That function probably should be refactored at some point because it's gotten a bit long, but we might be able to return old byte length from there before we delete the old null version. Maybe we can add this as a todo item?

lib/api/multiObjectDelete.js
electrachong: There is also a case when, if versioning is suspended and there is an existing null version, the new delete marker overwrites the null version, resulting in a change in the total content stored in the bucket. Is it something we need to track?

lib/api/multiObjectDelete.js
bennettbuchanan: Thanks for the feedback. Yes, with the help of @electrachong, I found that we could simply decrement `numOfObjects` for each delete marker that is put. This handles a couple cases:
* If only two delete markers are put, this value becomes `-2`, and the utapi metrics for `numberOfObjects` is incremented by two.
* If versions are deleted and delete markers are put in the same `multiObjectDelete` operation, then we end up with the correct number of objects to either increment or decrement for Utapi.

I wonder if this choice goes against the indended use of Redis commands (i.e., using the `decrby` command with a negative number to increment a value. What do you think @rahulreddy? Perhaps we could add a case in Utapi to check for the value being > 0 or < 0 and increment and decrement accordingly, for clarity's sake.

lib/api/multiObjectDelete.js
bennettbuchanan: Comment 1: Please see my response, just above, to Jonathan.
Comment 2: In this case, we would need to decrement storage utilized by the null version, if any. And it sounds like number of objects should remain unchanged (since we both delete a version, and put a delete marker).

electrachong:
I think `completeMpu` may need the same treatment as objectCopy or objectPut (no previous content length when adding a new version when versioning enabled). However, it doesn't look like we are tracking the prevContentLength for now in the case we are overwriting existing object MD // an object by the same name when completing a mpu. As discussed, you might want to look into this.

lib/api/objectPut.js
bennettbuchanan: Thanks for noting this. I've made some changes to address it.

Updated the callback here: https://github.com/scality/S3/pull/721/commits/94bbca2b4a5e1600d1c2f6adf33cd1d2978cd025.
Added a test in Integration: https://github.com/scality/Integration/pull/521/commits/a2f08ccd645f7138547ab410c93fc886432a365e

bennettbuchanan:
@ironman-machine
SCALITY_UTAPI_BRANCH=S3C-194-handle-metrics-for-versioning
SCALITY_INTEGRATION_BRANCH=S3C-194-handle-metrics-for-versioning
try

bennettbuchanan:
@ironman-machine
SCALITY_UTAPI_BRANCH=S3C-194-handle-metrics-for-versioning
SCALITY_INTEGRATION_BRANCH=S3C-194-handle-metrics-for-versioning
try

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -19,10 +19,12 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
             if (err) {
                 return callback(err);
             }
+            let nullVersionLength;
             if (dataToDelete) {
+                nullVersionLength = dataToDelete[0].size;
```
electrachong: I didn't think of this. The main caveat is that mpus are stored as an array of multiple data locations, so you may not be able to assume the entire object's content is in `dataToDelete[0].size`. If the null version was stored as a mpu, you will need to iterate through the array and add up the size of each part (the same way a mpu's content-length was calculated). Take this with a grain of salt, I don't know this first-hand, but it's what I interpreted from referencing the code.

bennettbuchanan:
@ironman-machine
SCALITY_UTAPI_BRANCH=S3C-194-handle-metrics-for-versioning
SCALITY_INTEGRATION_BRANCH=S3C-194-handle-metrics-for-versioning
try

bennettbuchanan:
e2e passed here: http://ci.ironmann.io/gh/scality/Integration/12864

lib/api/objectDelete.js
bennettbuchanan: Updated.

lib/api/apiUtils/object/createAndStoreObject.js
bennettbuchanan: As discussed: handling metrics for null version overwrites will be introduced in a later PR.

lib/api/multiObjectDelete.js
```diff
@@ -180,7 +180,7 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
         bucketName, bucket, quietSetting, errorResults, inPlay, log, next) {
     const successfullyDeleted = [];
     let totalContentLengthDeleted = 0;
-    let numOfObjects = 0;
+    let numOfObjectsRemoved = 0;
```
rahulreddy: You shouldn't have touched this file as the changes are not related to this PR.

lib/api/objectPut.js
rahulreddy: Can you add a TODO note in there mentioning the follow up. We can easily lose track of this otherwise.

lib/api/objectDelete.js
```diff
@@ -140,11 +140,9 @@ export default function objectDelete(authInfo, request, log, cb) {
                 resHeaders['x-amz-version-id'] = result.versionId === 'null' ?
                     result.versionId : versionIdUtils.encode(result.versionId);
             }
-            // TODO metric should be 'deleteObject', not 'putObject',
-            // but number of objects should be incremented since putting a new
-            // version. Utapi will be updated to allow this
-            pushMetric('putObject', log, { authInfo, bucket: bucketName,
-                newByteLength: 0, oldByteLength: 0 });
+            // We increment deleteObject even if creating a delete marker
```
LaurenSpiegel: don't you mean we increment putDeleteMarkerObject?

lib/api/multiObjectDelete.js
```diff
@@ -271,6 +268,9 @@ export function getObjMetadataAndDelete(authInfo, canonicalID, request,
             if (deleteInfo.newDeleteMarker) {
                 isDeleteMarker = true;
                 deleteMarkerVersionId = versionIdUtils.encode(versionId);
+                // In this case we are putting a new object (i.e., the delete
+                // marker), so we decrement the numOfObjectsRemoved value.
```
LaurenSpiegel: thank you for this comment 😄

lib/api/objectDelete.js
bennettbuchanan: I could add a clarifying comment about this. The method called in utapi 'putDeleteMarkerObject' actually increments the 's3:deleteObject' metric. The intention of adding this method was to encapsulate the logic of putting a delete marker (i.e., incrementing both 'deleteObject' and 'numberOfObjects' at once).

bennettbuchanan:
@ironman-machine
SCALITY_UTAPI_BRANCH=S3C-194-handle-metrics-for-versioning
SCALITY_INTEGRATION_BRANCH=S3C-194-handle-metrics-for-versioning
try

bennettbuchanan:
@ironman-machine
SCALITY_UTAPI_BRANCH=S3C-194-handle-metrics-for-versioning
SCALITY_INTEGRATION_BRANCH=S3C-194-handle-metrics-for-versioning
try

bennettbuchanan:
@ironman-machine try

PR #723, opened 2017-05-10
FIX: put object tag with empty value

nicolas2bert:
For `putObjectTagging` - the tag's value can be an empty string. 
However - tag's key can not be empty


nicolas2bert:
@ironman-machine ok

PR #725, opened 2017-05-10
fix: return versionId when delete non-exist ver [S3C-368]

electrachong:
- tests to confirm we do not create delete markers for deleting specific versions
- make sure version ID is sent back for requests trying to delete non-existing version

### DIFF
--- a/lib/api/objectDelete.js
+++ b/lib/api/objectDelete.js
@@ -12,7 +12,6 @@ import { metadataValidateBucketAndObj } from
 '../metadata/metadataUtils';
 
 const versionIdUtils = versioning.VersionID;
-const skipError = new Error('skip');
 
 /**
  * objectDelete - DELETE an object from a bucket
@@ -66,13 +65,10 @@ export default function objectDelete(authInfo, request, log, cb) {
                     // AWS does not return an error when trying to delete a
                     // specific version that does not exist. We skip to the end
                     // of the waterfall here.
-                    // TODO: Look into AWS behavior for whether to create
-                    // a delete marker if request is trying to delete a version
-                    // that does not exist.
                     if (reqVersionId) {
                         log.debug('trying to delete specific version ' +
                         ' that does not exist');
-                        return next(skipError, bucketMD);
+                        return next(errors.NoSuchVersion, bucketMD);
                     }
                     // To adhere to AWS behavior, create a delete marker even
                     // if trying to delete an object that does not exist when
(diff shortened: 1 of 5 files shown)


electrachong:
@ironman-machine

lib/api/objectDelete.js
```diff
@@ -125,12 +122,19 @@ export default function objectDelete(authInfo, request, log, cb) {
...
             }
-            log.debug('error processing request', { error: err,
-                method: 'objectDelete' });
-            return cb(err, resHeaders);
+            log.trace('received request to delete non-existing version');
+            if (reqVersionId) {
+                resHeaders['x-amz-version-id'] = reqVersionId === 'null' ?
+                    reqVersionId : versionIdUtils.encode(reqVersionId);
+            }
+            pushMetric('deleteObject', log, { authInfo, bucket: bucketName,
```
bennettbuchanan: Have we validated this is the correct behavior? In any case, we should probably add an Integration test for this case.

electrachong:
end-to-end passed

lib/api/objectDelete.js
electrachong: I think I will have to check with @rahulreddy whether we should increment the counter for the  'deleteObject' action, even though we don't delete any actual object/version in this scenario.

tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
```diff
@@ -136,6 +142,35 @@ describe('Multi-Object Versioning Delete Success', function success() {
...
+                ({ Key: obj.Key, VersionId: obj.VersionId }));
+            const prevVersion = objects[0].VersionId;
+            objects[0].VersionId = nonExistingId;
+            return s3.deleteObjectsAsync({
+                Bucket: bucketName,
+                Delete: {
+                    Objects: objects,
+                },
+            }).then(res =>
+                s3.deleteObjectAsync({
```
LaurenSpiegel: why are you doing the clean up delete in here?

lib/api/objectDelete.js
electrachong: I checked with Rahul and he said we don't increment the counter for the api if the error is NoSuchKey or NoSuchVersion. I removed the pushMetric call here.

tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
electrachong: I moved the clean-up to the after each callback.

electrachong:
@ironman-machine

electrachong:
end-to-end passed
http://ci.ironmann.io/gh/scality/Integration/13008

PR #726, opened 2017-05-11
chore: add yarn.lock file

rahulreddy:


rahulreddy:
Doesn't do well with github projects as dependencies as the commit hash is used during install. I will re-open later with proper lock file.

PR #730, opened 2017-05-16
FT: objectCopy with tags

nicolas2bert:
objectCopy with tags

Integration PR: https://github.com/scality/Integration/pull/529

### DIFF
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -36,6 +36,13 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
     if (whichMetadata !== 'COPY' && whichMetadata !== 'REPLACE') {
         return { error: errors.InvalidArgument };
     }
+    let whichTagging = headers['x-amz-tagging-directive'];
+    // Default is COPY
+    whichTagging = whichTagging === undefined ? 'COPY' : whichTagging;
+    if (whichTagging !== 'COPY' && whichTagging !== 'REPLACE') {
+        return { error: errors.InvalidArgument
+          .customizeDescription('Unknown tagging directive') };
+    }
     const overrideMetadata = {};
     if (headers['x-amz-server-side-encryption']) {
         overrideMetadata['x-amz-server-side-encryption'] =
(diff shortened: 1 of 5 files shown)


lib/api/objectCopy.js
```diff
@@ -36,6 +36,12 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
     if (whichMetadata !== 'COPY' && whichMetadata !== 'REPLACE') {
         return { error: errors.InvalidArgument };
     }
+    let whichTagging = headers['x-amz-tagging-directive'];
+    // Default is COPY
+    whichTagging = whichTagging === undefined ? 'COPY' : whichTagging;
+    if (whichTagging !== 'COPY' && whichTagging !== 'REPLACE') {
+        return { error: errors.InvalidArgument };
```
LaurenSpiegel: let's customize the error to say that the tagging directive is invalid.

lib/api/objectCopy.js
```diff
@@ -65,6 +71,17 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         utils.getMetaHeaders(sourceObjMD) :
         utils.getMetaHeaders(headers);
 
+    // If tagging directive is REPLACE, but you don't specify the
+    // x-amz-tagging in the request, the destination object will
+    // not have any tags.
```
LaurenSpiegel: I think you mean -- "If tagging directive is REPLACE but the source object does not have tags, the destination object will not have any tags."

correct?

Is there a difference for us saving an empty object versus having undefined? When a get is done, will 0 be returned for tag count versus nothing?

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -149,6 +170,64 @@ describe('Object Copy', () => {
...
+                    checkSuccessTagging(originalTagKey, originalTagValue, done);
+                });
+        });
+
+        it('should copy an object from a source bucket to a different ' +
+            'destination bucket and copy the tag set if COPY tagging ' +
+            'directive header provided', done => {
+            s3.copyObject({ Bucket: destBucketName, Key: destObjName,
+                CopySource: `${sourceBucketName}/${sourceObjName}`,
+                TaggingDirective: 'COPY' },
```
LaurenSpiegel: can we do a test where you say 'COPY' but send new tags anyway to make sure the tags are ignored or the correct error is returned?

LaurenSpiegel: let's also test when we do a copy and the origin object had no tags.

lib/api/objectCopy.js
nicolas2bert: No but I can also add: `if tagging directive is COPY but the source object does not have tags, the destination object will not have any tags.`

There is no difference between saving {} and undefined both are considered as 0 tag.
`x-amz-tagging-count` header is returned only if the count is greater than zero.

lib/api/objectCopy.js
nicolas2bert: Good idea :)

lib/api/objectCopy.js
LaurenSpiegel: ah, I think the comment would be clearer: 

// If tagging directive is REPLACE but you don't specify any
// tags in the request, the destination object will
// not have any tags.
// If tagging directive is COPY but the source object does not have tags,
// the destination object will not have any tags.

tests/functional/aws-node-sdk/test/object/objectCopy.js
nicolas2bert: done :)

nicolas2bert:
@ironman-machine ok

nicolas2bert:
E2E with its integration branch: http://ci.ironmann.io/gh/scality/Integration/13128

PR #734, opened 2017-05-17
S3C-380: Get security token

alexandre-merle:
Send security token to vault


alexandre-merle:
@ironman-machine clean try

alexandre-merle:
@ironman-machine try

PR #735, opened 2017-05-17
Add AWS backend

dora-korpar:
This PR adds PUT, GET, DELETE routes from Scality S3 to AWS S3.


tests/functional/aws-node-sdk/test/multipleBackend/delete.js
```diff
@@ -64,7 +71,18 @@ describe('Multiple backend delete', () => {
...
                 s3.getObject({ Bucket: bucket, Key: fileObject }, err => {
                     assert.strictEqual(err.code, 'NoSuchKey', 'Expected ' +
-                        'error but got sucess');
+                        'error but got success');
+                    done();
+                });
+            });
+        });
+        it('should delete object from AWS', done => {
+            s3.deleteObject({ Bucket: bucket, Key: awsObject }, err => {
```
electrachong: You don't have to define the location constraint when getting or deleting an object? What happens if you have objects put with the same names in buckets of the same name in different backends?

lib/data/wrapper.js
```diff
@@ -153,11 +153,16 @@ const data = {
...
-    batchDelete: (locations, log) => {
+    batchDelete: (locations, log, noDelete) => {
         // TODO: The method of persistence of sproxy delete key will
         // be finalized; refer Issue #312 for the discussion. In the
         // meantime, we at least log the location of the data we are
         // about to delete before attempting its deletion.
+        const skipIfBackend = ['aws-test'];
+        if (noDelete &&
+        skipIfBackend.indexOf(locations[0].dataStoreName) > -1) {
+            return;
```
electrachong: can you explain the need for this?

lib/data/locationConstraintParser.js
```diff
@@ -31,7 +32,38 @@ export default function parseLC() {
...
+                    computeChecksums: true,
+                    accessKeyId: locationObj.details.credentials.accessKey,
+                    secretAccessKey: locationObj.details.credentials.secretKey,
+                });
+            }
+            clients[location].clientType = 'aws_s3';
+            clients[location].awsBucketName = locationObj.details.bucketName;
+            clients[location].bucketMatch = locationObj.details.bucketMatch;
+            clients[location].dataStoreName = location;
         }
```
dora-korpar: yup, they're asserted in Config.js. If the config values aren't right, the server throws an error and doesn't start

lib/data/wrapper.js
dora-korpar: yeah, with our normal functionality, if an object is put with the same key as an existing object, we delete the old object and put the new one. This doesn't work with AWS though, because we end up deleting the object after the second put request is done, so there's ultimately no object on AWS

lib/data/multipleBackendGateway.js
```diff
@@ -48,23 +115,52 @@ const multipleBackendGateway = {
     },
 
     get: (objectGetInfo, range, reqUids, callback) => {
+        const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
```
dora-korpar: If objectGetInfo.key exists the md-model-version is 2 or greater.
Otherwise, the objectGetInfo is just the key string.
That's from the data wrapper file :)

lib/data/wrapper.js
electrachong: Are you sure you want to check for dataStoreName 'aws-test' instead of clientType 'aws_s3' though?

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
dora-korpar: If two objects with the same key are put in the same bucket, even if the location constraints are different, the old one will be deleted and the new one put in the new location

lib/data/wrapper.js
dora-korpar: That's going to get changed. It's just gonna take a little more finagling to get the clientType :)

dora-korpar:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/awsCredentials
try

lib/data/multipleBackendGateway.js
electrachong: Okay. So what about if you're trying to retrieve an object that was stored before md-model-version 2 (or at least before the implementation of multiple backends) for the next line? `objectGetInfo.dataStoreName` would be undefined, right? Is the multiple backends feature backwards incompatible with objects stored before multiple backend implementation anyway? (Sorry, I just have a lot of questions because I'm not familiar with this feature)

dora-korpar:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/awsCredentials
try

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -113,6 +113,22 @@ describe('MultipleBackend put object', () => {
                     });
                 });
             });
+
+            it('should put an object to AWS', done => {
```
nicolas2bert: How do you check that the data is stored in AWS?

lib/data/multipleBackendGateway.js
```diff
@@ -77,14 +173,28 @@ const multipleBackendGateway = {
                     }
                     cb();
                 });
+            } else if (client.clientType === 'aws_s3') {
+                if (deep) {
```
LaurenSpiegel: is deep a boolean?  what determines its value/what is calling it? why is this only needed for the aws backend?

lib/data/multipleBackendGateway.js
```diff
@@ -77,14 +173,28 @@ const multipleBackendGateway = {
...
                 multBackendResp[location] = { code: 200, message: 'OK' };
                 cb();
             }
         }, () => {
-            callback(null, multBackendResp);
+            if (awsArray.length > 0) {
+                const randomAWS = awsArray[Math.floor(Math.random() *
+                    awsArray.length)];
+                const checkThisOne = clients[randomAWS];
+                checkAWSHealth(checkThisOne, randomAWS, callback);
```
LaurenSpiegel: what if there are aws and other backends?  won't you just be returning the aws responses here?

if you don't have a test for healthcheck on multiple backends that involve multiple aws buckets and other backends, please add.

lib/data/wrapper.js
```diff
@@ -153,11 +153,16 @@ const data = {
 
     // It would be preferable to have an sproxyd batch delete route to
     // replace this
-    batchDelete: (locations, log) => {
+    batchDelete: (locations, log, noDelete) => {
         // TODO: The method of persistence of sproxy delete key will
         // be finalized; refer Issue #312 for the discussion. In the
         // meantime, we at least log the location of the data we are
         // about to delete before attempting its deletion.
+        const skipIfBackend = ['aws-test'];
```
LaurenSpiegel: as discussed, this needs to be updated.

lib/data/wrapper.js
```diff
@@ -207,7 +212,7 @@ const data = {
                 message: result.statusMessage,
             };
             return cb(null, respBody);
-        });
+        }, deep);
```
LaurenSpiegel: please switch the order of the arguments so you don't have to have an argument after the callback.

lib/utilities/healthcheckHandler.js
```diff
@@ -27,7 +27,7 @@ function writeResponse(res, error, log, results, cb) {
     });
 }
 
-export function clientCheck(log, cb) {
+export function clientCheck(log, cb, deep) {
```
LaurenSpiegel: please add jsdoc for this.

lib/data/multipleBackendGateway.js
```diff
@@ -48,23 +115,52 @@ const multipleBackendGateway = {
...
+        }
+        if (client.clientType === 'aws_s3') {
+            const stream = client.getObject({
+                Bucket: client.awsBucketName,
+                Key: key,
+                Range: range,
+            }).createReadStream().on('error', err => {
+                const log = createLogger(reqUids);
+                log.error('error creating stream', { err });
+                return callback(err);
```
LaurenSpiegel: please make sure there are tests where you expect an error from aws -- like no such key, access denied.

lib/api/objectPut.js
```diff
@@ -82,6 +82,10 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
                 return next(null, null);
             },
             function objectCreateAndStore(cipherBundle, next) {
+                if (objMD) {
```
LaurenSpiegel: Please add tests for not overwriting data on S3 as intended and then deleting as intended.

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -162,13 +163,14 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
...
         function storeMDAndDeleteData(options, infoArr, next) {
+            const noDelete = objMD ? objMD.noDelete : null;
             metadataStoreParams.versionId = options.versionId;
             metadataStoreParams.versioning = options.versioning;
             metadataStoreParams.isNull = options.isNull;
             metadataStoreParams.nullVersionId = options.nullVersionId;
             return _storeInMDandDeleteData(bucketName, infoArr,
                 cipherBundle, metadataStoreParams,
-                options.dataToDelete, requestLogger, next);
+                options.dataToDelete, requestLogger, noDelete, next);
```
LaurenSpiegel: Please add tests for putting objects with 0 bytes on aws to make sure the metadata is there and can get both directly from aws and through scality s3.

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
LaurenSpiegel: please add a test where you have an object foo on the aws backend and then you put foo on the mem backend.  The foo on aws should be deleted.  Also add a test the other way around -- put foo on mem and then foo on aws.

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
```diff
@@ -33,6 +34,12 @@ describe('Multiple backend delete', () => {
                     Metadata: { 'scal-location-constraint': 'file' } };
                 return s3.putObject(params);
             })
+            .then(() => {
+                process.stdout.write('Putting object to AWS\n');
+                const params = { Bucket: bucket, Key: awsObject, Body: body,
```
LaurenSpiegel: please also add tests with 0 byte objects.

tests/functional/aws-node-sdk/test/multipleBackend/get.js
```diff
@@ -106,6 +113,15 @@ describe('Multiple backend get object', () => {
                         done();
                     });
             });
+            it('should get an object from AWS', done => {
```
LaurenSpiegel: please also add test getting 0 byte object.  also, need to test that you are getting the user metadata back correctly in both this test and the 0 byte object test.

tests/functional/aws-node-sdk/test/multipleBackend/get.js
```diff
@@ -5,6 +5,7 @@ import BucketUtility from '../../lib/utility/bucket-util';
 const bucket = 'buckettestmultiplebackendget';
 const memObject = 'memobject';
 const fileObject = 'fileobject';
+const awsObject = 'awsobject';
 const emptyObject = 'emptyObject';
 const body = Buffer.from('I am a body', 'utf8');
```
LaurenSpiegel: we should also test putting a larger object.  you can create a 1GB buffer and put it.

lib/data/multipleBackendGateway.js
```diff
@@ -31,6 +70,34 @@ const multipleBackendGateway = {
...
+                       log.error('err from data backend',
+                       { err, dataStoreName: client.dataStoreName });
+                       return callback(errors.InternalError);
+                   }
+                   const dataRetrievalInfo = {
+                       key: awsKey,
+                       dataStoreName: client.dataStoreName,
+                       // because of encryption the ETag here could be
+                       // different from our metadata so let's store it
+                       dataStoreETag: data.ETag,
```
LaurenSpiegel: Let's discuss whether we should store the AWS version id that is returned.

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -182,6 +198,28 @@ describeSkipIfE2E('MultipleBackend put object based on bucket location', () => {
...
+            }, err => {
+                assert.equal(err, null, `Error creating bucket: ${err}`);
+                process.stdout.write('Putting object\n');
+                return s3.putObject(params, err => {
+                    assert.equal(err, null, 'Expected success, ' +
+                        `got error ${JSON.stringify(err)}`);
+                    s3.getObject({ Bucket: bucket, Key: key }, (err, res) => {
+                        assert.strictEqual(err, null, 'Expected success, ' +
+                        `got error ${JSON.stringify(err)}`);
+                        assert.strictEqual(res.ETag, `"${correctMD5}"`);
```
LaurenSpiegel: these tests should also all assert that the metadata is correctly there.

tests/multipleBackend/objectPutCopyPart.js
```diff
@@ -153,6 +153,13 @@ describe('ObjectCopyPutPart API with multiple backends', () => {
         });
     });
 
+    it('should copy part to AWS based on mpu location', done => {
```
LaurenSpiegel: how does this work if you haven't implemented the mpu stuff yet in this pr?

tests/multipleBackend/objectPutCopyPart.js
```diff
@@ -172,6 +179,13 @@ describe('ObjectCopyPutPart API with multiple backends', () => {
         });
     });
 
+    it('should copy part to AWS based on bucket location', done => {
```
LaurenSpiegel: same

tests/multipleBackend/objectPutPart.js
```diff
@@ -152,6 +152,13 @@ describe('objectPutPart API with multiple backends', () => {
         });
     });
 
+    it('should put a part to AWS based on mpu location', done => {
+        putPart('file', 'aws-test', null, 'localhost', () => {
```
LaurenSpiegel: same

tests/functional/aws-node-sdk/test/multipleBackend/get.js
```diff
@@ -60,7 +61,7 @@ describe('Multiple backend get object', () => {
             });
 
         describeSkipIfE2E('with objects in all available backends ' +
-            '(mem/file)', () => {
```
LaurenSpiegel: where are the tests that will run in the end to end? in integration?

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -113,6 +113,22 @@ describe('MultipleBackend put object', () => {
                     });
                 });
             });
+
```
LaurenSpiegel: we should have tests where we are overwriting objects that were already put.  both 0 byte and non-0 byte.  also, overwriting to different backends.

tests/functional/aws-node-sdk/test/multipleBackend/put.js
dora-korpar: Just after the put, I do a get object to make sure it's there :)

lib/data/multipleBackendGateway.js
```diff
@@ -31,6 +70,34 @@ const multipleBackendGateway = {
...
         }
+        // client is AWS SDK
+        if (client.clientType === 'aws_s3') {
+            const awsKey = _createAwsKey(keyContext.bucketName,
+               keyContext.objectKey, client.bucketMatch);
+            return client.putObject({
+                Bucket: client.awsBucketName,
+                Key: awsKey,
+                Body: stream,
+                Metadata: keyContext.metaHeaders,
```
dora-korpar: yup :)

lib/data/multipleBackendGateway.js
```diff
@@ -17,6 +17,45 @@ function createLogger(reqUids) {
 }
 
 const clients = parseLC(config);
+let multBackendResp = {};
```
dora-korpar: It's being used in both the checkAwsHealth function and the gateway healthcheck function

lib/data/multipleBackendGateway.js
```diff
@@ -31,6 +69,37 @@ const multipleBackendGateway = {
                 callback(errors.InternalError);
             });
         }
+        // client is AWS SDK
+        if (client.clientType === 'aws_s3') {
+            const awsKey = _createAwsKey(keyContext.bucketName,
+               keyContext.objectKey, client.bucketMatch);
+            // TODO: if object to be encrypted, use encryption
+            // on AWS
+            return client.putObject({
```
jonathan-gramain: Why not create a client wrapper for AWS, so that the code gets out of the way of multiple backend management?

lib/data/multipleBackendGateway.js
```diff
@@ -42,49 +111,106 @@ const multipleBackendGateway = {
...
+            });
+        }
         if (client.clientType === 'scality') {
-            return client.delete(objectGetInfo.key, reqUids, callback);
+            return client.delete(key, reqUids, callback);
         }
         return client.delete(objectGetInfo, reqUids, callback);
     },
 
     healthcheck: (log, callback) => {
```
jonathan-gramain: Seeing this function getting more complex, maybe time to think making `healthcheck` another client backend-specific call?

lib/data/wrapper.js
```diff
@@ -153,11 +153,19 @@ const data = {
...
         // TODO: The method of persistence of sproxy delete key will
         // be finalized; refer Issue #312 for the discussion. In the
         // meantime, we at least log the location of the data we are
         // about to delete before attempting its deletion.
+        const skipBackend = ['aws_s3'];
+        const isSkipBackend = (locations[0] && locations[0].dataStoreType) ?
+            skipBackend.indexOf(locations[0].dataStoreType) > -1 : false;
+        const isMatchingBackends = locations[0] ?
+            locations[0].dataStoreName === newObjDataStoreName : false;
+        if (requestMethod === 'PUT' && isSkipBackend && isMatchingBackends) {
```
jonathan-gramain: Why is this new check necessary? It does not come straightforward, maybe a comment would be useful. Can it make more sense to have the check supported by a backend-specific batchDelete?

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
jonathan-gramain: +1

tests/functional/aws-node-sdk/test/multipleBackend/get.js
jonathan-gramain: I would be reluctant to run such test in the CI (because of time to send all bytes to AWS and $$). Although I agree this must be tested, at least once, and may be in Integration instead or could be run as part of a non-reg test suite before a release?

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -113,6 +128,22 @@ describe('MultipleBackend put object', () => {
...
+                    Body: body,
+                    Metadata: { 'scal-location-constraint': 'aws-test' } };
+                s3.putObject(params, err => {
+                    assert.equal(err, null, 'Expected success, ' +
+                        `got error ${err}`);
+                    s3.getObject({ Bucket: bucket, Key: key }, (err, res) => {
+                        assert.strictEqual(err, null, 'Expected success, ' +
+                        `got error ${err}`);
+                        assert.strictEqual(res.ETag, `"${correctMD5}"`);
+                        done();
```
jonathan-gramain: You should also check the body in `res` against the original to make sure there was no procedural error in the storage.

tests/multipleBackend/objectPut.js
```diff
@@ -64,8 +64,8 @@ function put(bucketLoc, objLoc, requestHost, cb, errorDescription) {
     });
 }
 
-describe('objectPutAPI with multiple backends', () => {
-    beforeEach(() => {
+describe.only('objectPutAPI with multiple backends', () => {
```
jonathan-gramain: Put back your screen post-it and kick that `.only` ;)

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -113,6 +151,118 @@ describe('MultipleBackend put object', () => {
...
+                    });
+                });
+            });
+
+            itSkipIfE2E('should put objects with same key to file then AWS, ' +
+            'and object should only be present on AWS', done => {
+                const params = { Bucket: bucket, Key: key,
+                    Body: body,
+                    Metadata: { 'scal-location-constraint': 'file' } };
+                s3.putObject(params, err => {
```
jonathan-gramain: You may use `async.series` to chain the calls in a more readable way and decrease indentation level. (same for other tests of similar shape).

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -114,12 +119,15 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
...
+    const location = backendInfo.getControllingLocationConstraint();
+    const locationType = config.locationConstraints[location].type;
+    const skipBackend = ['aws_s3'];
 
     const requestLogger =
         logger.newRequestLoggerFromSerializedUids(log.getSerializedUids());
     return async.waterfall([
         function storeData(next) {
-            if (size === 0) {
+            if (size === 0 && skipBackend.indexOf(locationType) < 0) {
```
nicolas2bert: Instead of going through an array, it might be faster to create an object 
**ex**: `const skipBackend = { 'aws_s3': true };` and check for `skipBackend[locationType]`

dora-korpar:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/awsCredentials
try

README.md
```diff
@@ -207,12 +207,14 @@ sent to `myLocationConstraint` to the file backend:
 
 Each locationConstraint must include the `type`, `legacyAwsBehavior`,
 and `details` keys. `type` indicates which backend will be used
-for that region. Currently, mem, file, and scality are the
+for that region. Currently, mem, file, scality, and AWS S3 are the
```
LaurenSpiegel: I think we want to defer updating the readme until mpu and the encryption pr's are merged.  Will you please move these changes there?

lib/Config.js
```diff
@@ -70,6 +70,18 @@ function locationConstraintAssert(locationConstraints) {
                     `bad config: ${field} must be a string`);
             }
         });
+        if (details.bucketMatch !== undefined) {
+            assert(typeof details.bucketMatch === 'boolean',
+                'bad config: details.bucketMatch must be a boolean');
+        }
+        if (details.credentials !== undefined) {
```
LaurenSpiegel: what about credentialsProfile?

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -114,12 +119,15 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         });
     }
     const backendInfo = backendInfoObj.backendInfo;
+    const location = backendInfo.getControllingLocationConstraint();
+    const locationType = config.locationConstraints[location].type;
+    const skipBackend = ['aws_s3'];
```
LaurenSpiegel: let's call this dontSkipBackend and do what @nicolas2bert suggested.

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -11,16 +11,20 @@ const locationConstraintCheck = require('./locationConstraintCheck');
...
-    metadataStoreParams, dataToDelete, deleteLog, callback) {
+    metadataStoreParams, dataToDelete, deleteLog, requestMethod, callback) {
     services.metadataStoreObject(bucketName, dataGetInfo,
         cipherBundle, metadataStoreParams, (err, result) => {
             if (err) {
                 return callback(err);
             }
             if (dataToDelete) {
-                data.batchDelete(dataToDelete, deleteLog);
+                const newDataStoreName = Array.isArray(dataGetInfo) ?
```
LaurenSpiegel: isn't dataGetInfo always an array?

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -114,12 +119,17 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
...
+    /* eslint-disable camelcase */
+    const dontSkipBackend = { aws_s3: true };
+    /* eslint-enable camelcase */
 
     const requestLogger =
         logger.newRequestLoggerFromSerializedUids(log.getSerializedUids());
     return async.waterfall([
         function storeData(next) {
-            if (size === 0) {
+            if (size === 0 && dontSkipBackend.indexOf(locationType) < 0) {
```
LaurenSpiegel: this needs to be updated too...

lib/data/locationConstraintParser.js
```diff
@@ -32,7 +33,40 @@ function parseLC() {
...
         if (locationObj.type === 'aws_s3') {
-            // TODO
+            let s3Params;
+            // users can either include the desired profile name from their
+            // ~/.aws/credentials file or include the accessKeyId and
+            // secretAccessKey directly in the locationConfig
+            if (locationObj.details.credentialsProfile) {
+                s3Params = {
+                    endpoint: `https://${locationObj.details.awsEndpoint}`,
```
LaurenSpiegel: I'd really rather not repeat these items.  can you build s3Params and then just add on the credentials or accessKeyId/secretAccessKey params as applicable?

lib/data/multipleBackendGateway.js
```diff
@@ -32,6 +71,37 @@ const multipleBackendGateway = {
...
+                       const log = createLogger(reqUids);
+                       log.error('err from data backend',
+                       { error: err, dataStoreName: client.dataStoreName });
+                       return callback(errors.InternalError);
+                   }
+                   const dataRetrievalInfo = {
+                       key: awsKey,
+                       dataStoreName: client.dataStoreName,
+                       dataStoreType: client.clientType,
+                       // because of encryption the ETag here could be
```
LaurenSpiegel: make a note to update this when you do the encryption pr.

lib/data/wrapper.js
```diff
@@ -154,11 +154,24 @@ const data = {
...
+        const skipBackend = { aws_s3: true };
+        /* eslint-enable camelcase */
+        const isSkipBackend = (locations[0] && locations[0].dataStoreType) ?
+            skipBackend[locations[0].dataStoreType] : false;
+        const isMatchingBackends = locations[0] ?
+            locations[0].dataStoreName === newObjDataStoreName : false;
+        // This check is done because on a PUT request to AWS, if the object
+        // already exists on AWS, the existing object should not be deleted,
+        // which is the functionality for all other backends
+        if (requestMethod === 'PUT' && isSkipBackend && isMatchingBackends) {
```
LaurenSpiegel: please add a todo to update for mpu and object copy as well

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
```diff
@@ -5,9 +5,13 @@ const BucketUtility = require('../../lib/utility/bucket-util');
...
 const memObject = 'memObject';
 const fileObject = 'fileObject';
+const awsObject = 'awsObject';
+const emptyObject = 'emptyObject';
+const bigObject = 'bigObject';
 const body = Buffer.from('I am a body', 'utf8');
+const bigBody = new Buffer(10485760);
 
-describe('Multiple backend delete', () => {
+describe.only('Multiple backend delete', () => {
```
LaurenSpiegel: uh oh...

tests/functional/aws-node-sdk/test/multipleBackend/delete.js
```diff
@@ -33,6 +37,25 @@ describe('Multiple backend delete', () => {
...
+            .then(() => {
+                process.stdout.write('Putting 0-byte object to AWS\n');
+                const params = { Bucket: bucket, Key: emptyObject,
+                    Metadata: { 'scal-location-constraint': 'aws-test' } };
+                return s3.putObject(params);
+            })
+            .then(() => {
+                process.stdout.write('Putting large object to AWS\n');
+                const params = { Bucket: bucket, Key: bigObject,
+                    Body: bigBody,
```
LaurenSpiegel: it might prefer full figured

tests/functional/aws-node-sdk/test/support/awsConfig.js
```diff
@@ -0,0 +1,20 @@
+const AWS = require('aws-sdk');
```
LaurenSpiegel: this file seems to be a duplicate of the function in tests/functional/aws-node-sdk/test/support/config.js.  I would remove that function in that file and import your function into that file so we don't duplicate it.

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -113,6 +141,147 @@ describe('MultipleBackend put object', () => {
...
+                            Key: key }, err => {
+                                assert.strictEqual(err.code, 'NoSuchKey');
+                                done();
+                            });
+                        });
+                    });
+                });
+            });
+
+            itSkipIfNotMultiple('should put objects with same key to file ' +
```
LaurenSpiegel: can you add a test with maybe file first and then mem where you check to make sure the ds array doesn't have the mem stuff anymore?

dora-korpar:
@ironman-machine gogogo

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go now

dora-korpar:
@ironman-machine run

dora-korpar:
@ironman-machine pass

dora-korpar:
@ironman-machine pass

dora-korpar:
@ironman-machine go

lib/Config.js
dora-korpar: credentialsProfile is included in the 'stringFields' array :)

dora-korpar:
@ironman-machine one mo time

dora-korpar:
@ironman-machine go

lib/data/locationConstraintParser.js
```diff
@@ -32,7 +33,35 @@ function parseLC() {
...
+            if (locationObj.details.credentialsProfile) {
+                s3Params.credentials = new AWS.SharedIniFileCredentials({
+                    profile: locationObj.details.credentialsProfile });
+            } else {
+                s3Params.accessKeyId =
+                    locationObj.details.credentials.accessKey;
+                s3Params.secretAccessKey =
+                    locationObj.details.credentials.secretKey;
+            }
+            clients[location] = new AWS.S3(s3Params);
```
jonathan-gramain: I don't want to block this PR for this, but it seems to me that `client[location]` should be an instance of a wrapper that embeds the actual AWS handle, instead of being the AWS handle, that would avoid "polluting" the multipleBackendGateway file with special cases for aws_s3.

dora-korpar:
@ironman-machine plz go

PR #737, opened 2017-05-19
Fix: Streaming v4 auth extra chunk

LaurenSpiegel:
In streaming v4 auth, the last piece is indicated by:
0;chunk-signature=whateversig\r\n\r\n

We were only expecting:
0;chunk-signature=whateversig\r\n

So in the infrequent event when the final '\r\n' was in a separate chunk in the node transform stream, we were doing an unnecessary authentication step which was resulting in a signature mismatch error.

This PR addresses that problem, adds additional logs and cherry-picks a prior streaming v4 refactor that was applied in 6.4

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -25,7 +25,7 @@ dependencies:
     - wget http://launchpadlibrarian.net/222422124/s3cmd_1.6.0-2_all.deb
     - sudo dpkg -i s3cmd*.deb
     # fog and ruby testing dependencies
-    - gem install fog
+    - gem install fog-aws
     - gem install mime-types
     - gem install rspec
     - gem install json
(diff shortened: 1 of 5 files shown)


LaurenSpiegel:
@ironman-machine start with this one

LaurenSpiegel:
@ironman-machine try
DEFAULT_BRANCH=GA6.3.2

LaurenSpiegel:
@ironman-machine so picky
DEFAULT_BRANCH=rel/6.3

LaurenSpiegel:
@ironman-machine with fog gem update
DEFAULT_BRANCH=rel/6.3

LaurenSpiegel:
end to end passed here: http://ci.ironmann.io/gh/scality/Integration/13043

PR #741, opened 2017-05-22
TEST: version non-specified (tag and acl)

nicolas2bert:
**AWS compatibility**

AWS does not create version:
- putting ACL on a version-enabled bucket where no version id is specified
- putting object tags on a version-enabled bucket where no version id is specified
- deleting object tags on a version-enabled bucket where no version id is specified


nicolas2bert:
@ironman-machine ok

PR #743, opened 2017-05-23
FX: putObject with tag set policy

nicolas2bert:
To specify tags on an object, the requester must have `s3:PutObjectTagging` included in the list of permitted actions in their IAM policy.


nicolas2bert:
Integration:
BRANCH: https://github.com/scality/Integration/compare/fx/putobjectwithtagging?expand=1
E2E: http://ci.ironmann.io/gh/scality/Integration/13081#build-parameters

PR #744, opened 2017-05-25
Patch: Bypass double buffering for 1 location

LaurenSpiegel:
When getting a large object, we are experiencing memory issues. This is a patch to prevent the issues from occurring when an object was put without MPU.

### DIFF
--- a/lib/routes/routesUtils.js
+++ b/lib/routes/routesUtils.js
@@ -359,6 +359,25 @@ const routesUtils = {
                 httpCode: response.statusCode,
             });
         });
+        if (parsedLocations.length === 1) {
+            return data.get(parsedLocations[0], log,
+                (err, readable) => {
+                    if (err) {
+                        log.error('failed to get full object', {
+                            error: err,
+                            method: 'data.get',
+                        });
+                        return response.connection.destroy();
+                    }
+                    readable.on('error', () => {
+                        log.error('error piping data ' +
+                        'from readable to response');
+                        return response.connection.destroy();
+                    });
+                    readable.pipe(response);
+                    return undefined;
+                });
+        }
         return readySetStream(parsedLocations, data.get, response, log);
     },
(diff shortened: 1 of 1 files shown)


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
@ironman-machine try
DEFAULT_BRANCH=rel/6.4.2

rahulreddy:
@LaurenSpiegel You may need DEFAULT_BRANCH=rel/6.4

LaurenSpiegel:
@ironman-machine try
DEFAULT_BRANCH=rel/6.4

LaurenSpiegel:
@ironman-machine try
DEFAULT_BRANCH=rel/6.4
CI_AGGRESSOR_FORCE_BUILD=true

LaurenSpiegel:
End to end passed except for the fog tests due to 6.4.2 in s3 not updated to require fog/aws even though this is what the integration aggressor has in rel/6.4.  Utapi tests also failed.  @rahulreddy, can you please confirm this is just an issue with an update that is not in s3/utapi 6.4.2 but is in rel/6.4?

lib/routes/routesUtils.js
```diff
@@ -359,6 +359,25 @@ const routesUtils = {
...
+                (err, readable) => {
+                    if (err) {
+                        log.error('failed to get full object', {
+                            error: err,
+                            method: 'data.get',
+                        });
+                        return response.connection.destroy();
+                    }
+                    readable.pipe(response);
+                    readable.on('error', () => {
```
electrachong: should you put this before the pipe?

LaurenSpiegel: updated

rahulreddy:
@LaurenSpiegel Correct. User level metrics are not available in 6.4.2.

PR #745, opened 2017-05-25
S3C-413 put object ACL with s3:PutObjectAcl permission

nicolas2bert:
If `x-amz-grant-read, x-amz-grant-read-acp, and x-amz-grant-write-acp, x-amz-grant-full-control`
 or `x-amz-acl` headers are used to change the object ACLs to something other than the default, the requester must have `s3:PutObjectAcl` included in the list of permitted actions in their AWS Identity and Access Management (IAM) policy. 

Integration tests: https://github.com/scality/Integration/pull/531


nicolas2bert:
@LaurenSpiegel E2E with scality/Integration#531 passed: http://ci.ironmann.io/gh/scality/Integration/13240

PR #748, opened 2017-05-26
fix: clean up existing master null version before overwrite [S3C-444]

electrachong:
We identified a bug where we overwrite just the master null version when we have the same null version stored in master and a separate version, resulting in two null versions with different versionIds.

That situation may arise when putting object tags or acls on an existing master null version (using metadata PUT option `versionId=<version id of master version>` will both overwrite the master version + create new version with that versionId), for example, and then putting a new null version.

The solution settled upon is to do a separate clean-up delete call to delete the existing null version(s) before putting a new null version.

As @jonathan-gramain noted in a previous versioning PR, the versioningPreprocessing function was getting a bit long and nested and due for a refactor. Review by commits is advised.

~~Tests are still WIP, they were initially written to debug.~~ Tests are updated

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -1,4 +1,5 @@
 const { errors, versioning } = require('arsenal');
+const async = require('async');
 
 const metadata = require('../../../metadata/wrapper');
 const { config } = require('../../../Config');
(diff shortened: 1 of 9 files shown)


lib/api/apiUtils/object/versioning.js
```diff
@@ -133,6 +133,11 @@ function processVersioningState(reqVersionId, mst, vstat, cb) {
             options.versionId = '';
             options.isNull = true;
             options.dataToDelete = mst.objLocation;
+            // if null version exists, clean it up prior to put
+            if (mst.isNull) {
+                delOptions.versionId = mst.versionId;
+                return cb(null, options, null, delOptions);
```
jonathan-gramain: Discussed offline with @electrachong: I'm slightly concerned that this incurs a non-atomic operation where potentially old data is deleted without the new data written yet (atomicity issue), or that the old data is lost without being replaced by the new data (put failure), but this seems an orthogonal issue to this PR, the case already exists in other situations.

As this has already been discussed by mail with @ttvi , and since it affects only old data that the user wants removed anyway, I'm ok with it, but I think it should be tracked for further refinements (as I heard it requires additional metadata routes to selectively delete a version).

electrachong:
@ironman-machine

electrachong:
end-to-end passed

lib/api/apiUtils/object/versioning.js
ttvi: It's true that this is an issue of atomicity.
We can solve it using the existing BATCH operation of metadata, but we need update this operation to support versioning; we did not do it before because S3 does not use BATCH.

electrachong:
@ironman-machine

electrachong:
end-to-end retried: http://ci.ironmann.io/gh/scality/Integration/13585

PR #749, opened 2017-05-26
FIX: Only stream from one data location at a time.

LaurenSpiegel:
S3C-429
Prevent memory problems by no longer double buffering.

### DIFF
--- a/lib/routes/routesUtils.js
+++ b/lib/routes/routesUtils.js
@@ -1,6 +1,5 @@
-const { readySetStream } = require('ready-set-stream');
-
 const data = require('../data/wrapper');
+const retrieveData = require('../utilities/retrieveData');
 
 /**
  * setCommonResponseHeaders - Set HTTP response headers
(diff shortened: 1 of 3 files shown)


lib/routes/routesUtils.js
```diff
@@ -1,6 +1,5 @@
-const { readySetStream } = require('ready-set-stream');
```
GiorgioRegni: so sad...

LaurenSpiegel:
@ironman-machine try

lib/utilities/retrieveData.js
```diff
@@ -0,0 +1,34 @@
+function retrieveData(locations, dataRetrievalFn,
+    response, logger, errorHandlerFn) {
+    if (locations.length === 0) {
```
vrancurel: When does this condition occurs ?

LaurenSpiegel: I'm shifting from the array and calling the function recursively so when all is finished the array length is 0.

PR #752, opened 2017-05-31
FT: Create replication rule ID if not provided

bennettbuchanan:
The ID property of a rule in the replication configuration is optional. However if a rule ID is omitted, we should generate a unique ID to store with the bucket replication metadata. That ID is then associated with the rule and should be accessible with the bucketGetReplication API.

### DIFF
--- a/lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
+++ b/lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
@@ -1,4 +1,5 @@
 const assert = require('assert');
+const UUID = require('node-uuid');
 
 const { errors } = require('arsenal');
 
@@ -101,13 +102,13 @@ class ReplicationConfiguration {
             prefix: rule.Prefix[0],
             enabled: rule.Status[0] === 'Enabled',
         };
+        // ID is an optional property, but create one if not provided or is ''.
+        // We generate a 48-character alphanumeric, unique ID for the rule.
+        obj.id = rule.ID && rule.ID[0] !== '' ? rule.ID[0] :
+            Buffer.from(UUID.v4()).toString('base64');
         // StorageClass is an optional property.
         if (rule.Destination[0].StorageClass) {
             obj.storageClass = rule.Destination[0].StorageClass[0];
-        }
-        // ID is an optional property.
-        if (rule.ID) {
-            obj.id = rule.ID[0];
         }
         return obj;
     }
--- a/tests/unit/api/bucketPutReplication.js
+++ b/tests/unit/api/bucketPutReplication.js
@@ -5,10 +5,10 @@ const getReplicationConfiguration =
     require('../../../lib/api/apiUtils/bucket/getReplicationConfiguration');
 const replicationUtils =
     require('../../functional/aws-node-sdk/lib/utility/replication');
+const log = new DummyRequestLogger();
 
 // Check for the expected error response code and status code.
 function checkError(xml, expectedErr, cb) {
-    const log = new DummyRequestLogger();
     getReplicationConfiguration(xml, log, err => {
         if (expectedErr === null) {
             assert.strictEqual(err, null, `expected no error but got '${err}'`);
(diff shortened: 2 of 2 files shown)


lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
```diff
@@ -92,6 +93,22 @@ class ReplicationConfiguration {
...
+     * Generate a 48-character alphanumeric, unique ID for a rule
+     * @return {string} - The unique rule ID
+     */
+    _createUniqueID() {
+        // If no ID is given in the replication configuration, we create it.
+        const id = Buffer.from(UUID.v4()).toString('base64');
+        // Each ID in a list of rules must be unique. If the generated ID
+        // already exists, call recursively until a unique ID is generated.
+        if (this._configIDs.includes(id)) {
+            return this._createUniqueID();
```
rahulreddy: A recursion should have an exit condition. There is no way that `uuid` will generate non-unique id on retry but for safety I would say limit it to 5/10 tries. If it fails to generate a unique id, we will take the non-unique id as technically this id doesn't serve any functional purpose for us, yet.

tests/unit/api/bucketPutReplication.js
```diff
@@ -61,4 +77,15 @@ describe('\'getReplicationConfiguration\' function', () => {
         it(`should accept replication configuration without \'${prop}\'`,
             done => checkError(createReplicationXML(prop), null, done));
     });
+
+    it('should create a rule \'ID\' if omitted from the replication ' +
```
rahulreddy: Just use double quotes if you ever have to escape single quotes.

bennettbuchanan: Updated.

lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
bennettbuchanan: Agreed that it will probably never happen. In any case, I've limited the recursion to exit at the fifth attempt.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
```diff
@@ -92,6 +93,25 @@ class ReplicationConfiguration {
...
+     * @param {number} [n] - The retry count, if calling recursively
+     * @return {string} - The unique rule ID
+     */
+    _createUniqueID(n) {
+        // If no ID is given in the replication configuration, we create it.
+        const id = Buffer.from(UUID.v4()).toString('base64');
+        // Each ID in a list of rules must be unique. If the generated ID
+        // already exists, call recursively until a unique ID is generated or
+        // we have retried five times.
+        let retries = n || 0;
```
jonathan-gramain: Isn't UUID supposed to create supposedly unique IDs already?

All in all, I would rather remove the existence checks, it's not worth the added complexity IMO, a simple call to `UUID.v4()` without any further check should be enough.

lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
```diff
@@ -101,14 +121,13 @@ class ReplicationConfiguration {
             prefix: rule.Prefix[0],
             enabled: rule.Status[0] === 'Enabled',
         };
+        // ID is an optional property, but create one if not provided or is ''.
+        obj.id = rule.ID && rule.ID[0] !== '' ? rule.ID[0] :
```
jonathan-gramain: Is `rule.ID` an array when not undefined? In which case if it's an empty array I assume we will affect an undefined element to `obj.id` which may not be the goal here. (an empty array is truish)

bennettbuchanan: If `rule.ID` is defined, it will be an array with one element that is a string, since it comes from the  `xml2js.parseString` method.

lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
bennettbuchanan: They are random. The check was just to prevent the (very) unlikely chance that it already exists. I would be happy to remove the check though, if it is too unlikely and thus not worth it.

bennettbuchanan:
@ironman-machine try

lib/api/apiUtils/bucket/models/ReplicationConfiguration.js
bennettbuchanan: I've removed those checks.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed.

PR #754, opened 2017-05-31
Fix: EntityTooLarge error for put part

LaurenSpiegel:
Backporting range request on object copy fix.

(cherry picked from commit f4239a4743d187622b2efdf4237f18d24dbbc2de)


LaurenSpiegel:
@ironman-machine go

PR #756, opened 2017-06-01
ft: limit user metadata to 2 KB

electrachong:
http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html

AWS limits user-provided metadata to 2 KB. Now that we are implementing AWS and other backends, we should be consistent in the amount of user metadata we allow during puts.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -68,6 +68,11 @@ const constants = {
     maximumAllowedPartSize: process.env.MPU_TESTING === 'yes' ? 104857600 :
         5368709120,
 
+    // AWS states max size for user-defined metadata (x-amz-meta- headers) is
+    // 2 KB: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html
+    // In testing, AWS seems to allow up to 88 more bytes, so we do the same.
+    maximumMetaHeadersSize: 2136,
+
     // hex digest of sha256 hash of empty string:
     emptyStringHash: crypto.createHash('sha256')
         .update('', 'binary').digest('hex'),
--- a/lib/services.js
+++ b/lib/services.js
@@ -159,7 +159,7 @@ const services = {
             md.setTags(validationTagRes);
         }
 
-        // Store user provided metadata.  TODO: limit size.
+        // Store user provided metadata.
         // For multipart upload this also serves to transfer
         // over metadata originally sent with the initiation
         // of the multipart upload (for instance, ACL's).
--- a/tests/functional/aws-node-sdk/test/object/put.js
+++ b/tests/functional/aws-node-sdk/test/object/put.js
@@ -4,6 +4,8 @@ const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
 const provideRawOutput = require('../../lib/utility/provideRawOutput');
 const { taggingTests } = require('../../lib/utility/tagging');
+const genMaxSizeMetaHeaders
+    = require('../../lib/utility/genMaxSizeMetaHeaders');
 
 const bucket = 'bucket2putstuffin4324242';
 const object = 'object2putstuffin';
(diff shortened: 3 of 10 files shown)


lib/services.js
```diff
@@ -168,6 +169,14 @@ const services = {
         // will reassign some of the above values with the ones
         // from the intiation of the multipart upload
         // (for instance, storage class)
+        const validateUserMdRes = validateUserMetadata(metaHeaders);
```
LaurenSpiegel: If we add the check here will we only provide the error for an mpu on the complete step?  It should be on the initiate step.

lib/api/apiUtils/object/storeObjectMd.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {objec} metaHeaders - object containing user-provided metaheaders
+ * @return {(Error|null)} - return MetadataTooLarge or null
+ */
+function validateUserMetadata(metaHeaders) {
+    let totalLength = 0;
+    Object.keys(metaHeaders).forEach(key => {
+        totalLength += key.length;
+        totalLength += metaHeaders[key].length;
+    });
+    if (totalLength > 2136) {
```
LaurenSpiegel: please add the 2136 to the constants file instead of having here

tests/functional/aws-node-sdk/test/object/put.js
```diff
@@ -85,6 +85,28 @@ describe('PUT object', () => {
                 });
             });
 
+        it('should return error if putting object w/ > 2KB user-defined md',
```
LaurenSpiegel: let's add an mpu and object copy test as well

tests/functional/aws-node-sdk/test/object/put.js
```diff
@@ -85,6 +85,28 @@ describe('PUT object', () => {
...
+            done => {
+                const metadata = {};
+                const counter = 8;
+                // 256 * 8 = 2048 (2 KB)
+                for (let i = 0; i < counter; i++) {
+                    const key = `header${i}`;
+                    const valueLength =
+                        256 - ('x-amz-meta-'.length + key.length);
```
LaurenSpiegel: if you can use the constants number rather than hard-coding here that would be good as well

lib/api/apiUtils/object/storeObjectMd.js
bennettbuchanan: Would there be some benefit to checking this condition inside the loop and then returning immediately?

lib/services.js
bennettbuchanan: Will this method eventually handle multiple errors? If not, it might be clearer to have it return a boolean.

electrachong: I moved it from services storing md step to when the metaheaders object is defined for each of the apis.

lib/api/apiUtils/object/storeObjectMd.js
electrachong: Updated

tests/functional/aws-node-sdk/test/object/put.js
electrachong: Updated

tests/functional/aws-node-sdk/test/object/put.js
electrachong: Updated

lib/api/apiUtils/object/storeObjectMd.js
electrachong: I updated using .every instead of .forEach so the loop stops if it doesn't meet the condition.

lib/services.js
electrachong: I ended up combining this function with the existing util to get the metaheaders, so now it returns the metaHeaders if they are valid.

electrachong:
@ironman-machine

constants.js
```diff
@@ -68,6 +68,8 @@ const constants = {
     maximumAllowedPartSize: process.env.MPU_TESTING === 'yes' ? 104857600 :
         5368709120,
 
+    maximumMetaHeadersSize: 2136,
```
LaurenSpiegel: please add note referencing aws documentation and then comment about what they actually allow

PR #758, opened 2017-06-02
Add MPU initiate/complete and putPart for AWS

dora-korpar:
Adds implementation of MPU basics to AWS backend

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -5,6 +5,8 @@ const { errors, versioning } = require('arsenal');
 
 const escapeForXML = require('../utilities/escapeForXML');
 const { pushMetric } = require('../utapi/utilities');
+const { config } = require('../Config');
+const multipleBackendGateway = require('../data/multipleBackendGateway');
 
 const data = require('../data/wrapper');
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
@@ -155,19 +157,72 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             }
             return next(errors.MalformedXML, destBucket);
         },
-        function retrieveParts(destBucket, objMD, mpuBucket, jsonList, next) {
-            services.getMPUparts(mpuBucket.getName(), uploadId, log,
+        function getMPUMetadata(destBucket, objMD, mpuBucket, jsonList, next) {
+            let splitter = constants.splitter;
+            // BACKWARD: Remove to remove the old splitter
+            if (mpuBucket.getMdBucketModelVersion() < 2) {
+                splitter = constants.oldSplitter;
+            }
+            // Reconstruct mpuOverviewKey to serve
+            // as key to pull metadata originally stored when mpu initiated
+            const mpuOverviewKey =
+                `overview${splitter}${objectKey}${splitter}${uploadId}`;
+
+            return metadata.getObjectMD(mpuBucket.getName(), mpuOverviewKey,
+            {}, log, (err, storedMetadata) => {
+                if (err) {
+                    return next(err, destBucket);
+                }
+                const location = storedMetadata.controllingLocationConstraint;
+                return next(null, destBucket, objMD, mpuBucket, jsonList,
+                storedMetadata, location, mpuOverviewKey);
+            });
+        },
+        function ifMultipleBackend(destBucket, objMD, mpuBucket, jsonList,
+        storedMetadata, location, mpuOverviewKey, next) {
+            if (config.backends.data === 'multiple') {
+                return multipleBackendGateway.completeMPU(objectKey,
+                uploadId, location, jsonList, destBucket, log,
+                (err, completeObjData) => {
+                    if (err) {
+                        return next(err, destBucket);
+                    }
+                    return next(null, destBucket, objMD, mpuBucket,
+                    jsonList, storedMetadata, completeObjData, mpuOverviewKey);
+                });
+            }
+            return next(null, destBucket, objMD, mpuBucket, jsonList,
+            storedMetadata, null, mpuOverviewKey);
+        },
+        function retrieveParts(destBucket, objMD, mpuBucket, jsonList,
+        storedMetadata, completeObjData, mpuOverviewKey, next) {
+            // means MPU completion was done on third-party backend,
+            // so part info is not in our metadata and related steps
+            // should be skipped
+            if (completeObjData) {
+                return next(null, destBucket, objMD, mpuBucket, null,
+                jsonList, storedMetadata, completeObjData, mpuOverviewKey);
+            }
+            return services.getMPUparts(mpuBucket.getName(), uploadId, log,
                 (err, result) => {
                     if (err) {
                         return next(err, destBucket);
                     }
                     const storedParts = result.Contents;
                     return next(null, destBucket, objMD, mpuBucket,
-                        storedParts, jsonList);
+                        storedParts, jsonList, storedMetadata, null,
+                        mpuOverviewKey);
                 });
         },
         function processParts(destBucket, objMD, mpuBucket, storedParts,
-            jsonList, next) {
+            jsonList, storedMetadata, completeObjData, mpuOverviewKey, next) {
+            let calculatedSize = 0;
+            if (completeObjData) {
+                calculatedSize = completeObjData.contentLength;
+                return next(null, destBucket, objMD, mpuBucket,
+                    storedMetadata, completeObjData.eTag, calculatedSize,
+                    null, mpuOverviewKey, null, null, completeObjData);
+            }
             const storedPartsAsObjects = storedParts.map(item => ({
                 // In order to delete the part listing in the shadow
                 // bucket, need the full key
@@ -178,11 +233,6 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 locations: Array.isArray(item.value.partLocations) ?
                     item.value.partLocations : [item.value.partLocations],
             }));
-            let splitter = constants.splitter;
-            // BACKWARD: Remove to remove the old splitter
-            if (mpuBucket.getMdBucketModelVersion() < 2) {
-                splitter = constants.oldSplitter;
-            }
             const dataLocations = [];
             // AWS documentation is unclear on what the MD5 is that it returns
             // in the response for a complete multipart upload request.
@@ -197,7 +247,6 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             // 4) create the hex digest of the md5
             // 5) add '-' plus the number of parts at the end
             let concatETags = '';
-            let calculatedSize = 0;
             // Check list sent to make sure valid
             // If the number of the parts in the JSON does not
             // match the number of parts stored, return error msg
@@ -323,28 +372,13 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             // Add the number of parts at the end
             aggregateETag = `${aggregateETag}-${jsonList.Part.length}`;
 
-            // All is good so get the metadata stored when the mpu
-            // was initiated and store everything as a new object
-
-            // Reconstruct mpuOverviewKey to serve
-            // as key to pull metadata originally stored when mpu initiated
-            const mpuOverviewKey =
-                `overview${splitter}${objectKey}${splitter}${uploadId}`;
-
-            return metadata.getObjectMD(mpuBucket.getName(), mpuOverviewKey,
-                {}, log, (err, storedMetadata) => {
-                    if (err) {
-                        return next(err, destBucket);
-                    }
-                    return next(null, destBucket, objMD, mpuBucket,
-                        storedMetadata, aggregateETag, calculatedSize,
-                        dataLocations, mpuOverviewKey,
-                        storedPartsAsObjects, extraPartLocations);
-                });
+            return next(null, destBucket, objMD, mpuBucket, storedMetadata,
+                aggregateETag, calculatedSize, dataLocations, mpuOverviewKey,
+                storedPartsAsObjects, extraPartLocations, null);
         },
         function prepForStoring(destBucket, objMD, mpuBucket, storedMetadata,
             aggregateETag, calculatedSize, dataLocations, mpuOverviewKey,
-            storedPartsAsObjects, extraPartLocations, next) {
+            storedPartsAsObjects, extraPartLocations, completeObjData, next) {
             const metaHeaders = {};
             const keysNotNeeded =
                 ['initiator', 'partLocations', 'key',
@@ -402,13 +436,14 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                     return next(null, destBucket, dataLocations,
                         metaStoreParams, mpuBucket, mpuOverviewKey,
                         aggregateETag, storedPartsAsObjects, objMD,
-                        extraPartLocations, pseudoCipherBundle, dataToDelete);
+                        extraPartLocations, pseudoCipherBundle, dataToDelete,
+                        completeObjData);
                 });
         },
         function storeAsNewObj(destinationBucket, dataLocations,
             metaStoreParams, mpuBucket, mpuOverviewKey, aggregateETag,
             storedPartsAsObjects, objMD, extraPartLocations, pseudoCipherBundle,
-            dataToDelete, next) {
+            dataToDelete, completeObjData, next) {
             return services.metadataStoreObject(destinationBucket.getName(),
                 dataLocations, pseudoCipherBundle, metaStoreParams,
                 (err, res) => {
@@ -426,18 +461,21 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                     }
                     return next(null, mpuBucket, mpuOverviewKey, aggregateETag,
                         storedPartsAsObjects, extraPartLocations,
-                        destinationBucket, generatedVersionId);
+                        destinationBucket, generatedVersionId, completeObjData);
                 });
         },
         function deletePartsMetadata(mpuBucket, mpuOverviewKey, aggregateETag,
             storedPartsAsObjects, extraPartLocations, destinationBucket,
-            generatedVersionId, next) {
-            const keysToDelete = storedPartsAsObjects.map(item => item.key);
+            generatedVersionId, completeObjData, next) {
+            let keysToDelete = [];
+            if (storedPartsAsObjects) {
+                keysToDelete = storedPartsAsObjects.map(item => item.key);
+            }
             keysToDelete.push(mpuOverviewKey);
             services.batchDeleteObjectMetadata(mpuBucket.getName(),
                 keysToDelete, log, err => next(err, destinationBucket,
                     aggregateETag, generatedVersionId));
-            if (extraPartLocations.length > 0) {
+            if (extraPartLocations && extraPartLocations.length > 0) {
                 data.batchDelete(extraPartLocations, request.method, null,
                     logger.newRequestLoggerFromSerializedUids(log
                     .getSerializedUids()));
--- a/lib/api/initiateMultipartUpload.js
+++ b/lib/api/initiateMultipartUpload.js
@@ -13,7 +13,8 @@ const locationConstraintCheck
     = require('./apiUtils/object/locationConstraintCheck');
 const validateWebsiteHeader = require('./apiUtils/object/websiteServing')
     .validateWebsiteHeader;
-
+const { config } = require('../Config');
+const multipleBackendGateway = require('../data/multipleBackendGateway');
 
 /*
 Sample xml response:
@@ -85,8 +86,6 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
         });
         return process.nextTick(() => callback(metaHeaders));
     }
-    // Generate uniqueID without dashes so that routing not messed up
-    const uploadId = UUID.v4().replace(/-/g, '');
     // TODO: Add this as a utility function for all object put requests
     // but after authentication so that string to sign is not impacted
     // This is GH Issue#89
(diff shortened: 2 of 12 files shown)


dora-korpar:
@ironman-machine go

lib/api/completeMultipartUpload.js
```diff
@@ -402,20 +434,26 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
             return services.metadataStoreObject(destinationBucket.getName(),
                 dataLocations, pseudoCipherBundle, metaStoreParams,
                 (err, res) => {
                     if (err) {
                         return next(err, destinationBucket);
                     }
-                    const generatedVersionId = res ? res.versionId : undefined;
+                    let generatedVersionId;
+                    if (newObjData) {
+                        generatedVersionId = newObjData.VersionId || undefined;
```
LaurenSpiegel: this would be on the info returned from AWS?  what if it was a different third-party backend?

lib/api/completeMultipartUpload.js
```diff
@@ -426,18 +464,21 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                        destinationBucket, generatedVersionId, newObjData);
                 });
         },
         function deletePartsMetadata(mpuBucket, mpuOverviewKey, aggregateETag,
             storedPartsAsObjects, extraPartLocations, destinationBucket,
-            generatedVersionId, next) {
-            const keysToDelete = storedPartsAsObjects.map(item => item.key);
+            generatedVersionId, newObjData, next) {
+            let keysToDelete = [];
+            if (!newObjData) {
```
LaurenSpiegel: maybe more straightforward to check if storedPartsAsObjects?

lib/api/initiateMultipartUpload.js
```diff
@@ -182,6 +165,46 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
+            });
+        }
+        const locConstraint = backendInfoObj.controllingLC;
+        metadataStoreParams.controllingLocationConstraint = locConstraint;
+
+        let uploadId;
+        if (config.backends.data === 'multiple') {
+            return multipleBackendGateway.createMPU(objectKey, bucketName,
+            metaHeaders, websiteRedirectHeader, locConstraint, log,
+            (err, data) => {
```
LaurenSpiegel: instead of 'data' let's use something more descriptive like resFromDataBackend

lib/api/initiateMultipartUpload.js
```diff
@@ -182,6 +165,46 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
+                if (err) {
+                    return callback(err);
+                }
+                uploadId = data.UploadId;
```
LaurenSpiegel: what if it's multiple but the backend is not aws?

lib/api/objectPutCopyPart.js
```diff
@@ -259,9 +259,9 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                                         { error: err });
                                         return cb(errors.InternalError);
                                     }
-                                    return data.put(cipherBundle, hashedStream,
-                                        numberPartSize, dataStoreContext,
-                                        backendInfo, log,
+                                    return data.put(cipherBundle,
+                                        hashedStream, numberPartSize,
+                                        dataStoreContext, backendInfo, log,
```
LaurenSpiegel: these just look like line break changes in this file so please remove.  separate pr for copy part?

lib/api/objectPutPart.js
```diff
@@ -177,6 +180,31 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                         cipherBundle, splitter);
                 });
         },
+        // If data backend is real AWS, no need to store part metadata
+        (destinationBucket, objectLocationConstraint, cipherBundle,
+        splitter, next) => {
+            if (config.backends.data === 'multiple') {
+                return multipleBackendGateway.uploadPart(
+                destinationBucket, objectLocationConstraint, objectKey,
+                uploadId, partNumber, log, (err, data) => {
```
LaurenSpiegel: please change data to something more descriptive

lib/api/objectPutPart.js
```diff
@@ -177,6 +180,31 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
+                            error: err,
+                            method:
+                            'objectPutPart::multipleBackendGateway.uploadPart',
+                        });
+                        return next(err, destinationBucket);
+                    } else if (data) {
+                        // skip to end of waterfall
+                        return next(skipError, data);
+                    }
+                    return next(null, destinationBucket,
```
LaurenSpiegel: please add comment that if backend doesn't handle mpu part, continue with putting steps

lib/api/objectPutPart.js
```diff
@@ -177,6 +180,31 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                         cipherBundle, splitter);
                 });
         },
+        // If data backend is real AWS, no need to store part metadata
```
LaurenSpiegel: If data backend is backend that handles mpu (like real AWS), no need to store part info in metadata

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
...
+
+    createMPU: (key, bucket, metaHeaders, websiteRedirectHeader,
+    location, log, cb) => {
+        const client = clients[location];
+        const awsKey = _createAwsKey(bucket, key, client.bucketMatch);
+        if (client.clientType === 'aws_s3') {
+            const params = { Bucket: client.awsBucketName, Key: awsKey,
+                WebsiteRedirectLocation: websiteRedirectHeader,
+                Metadata: metaHeaders };
+            return client.createMultipartUpload(params, (err, data) => {
```
LaurenSpiegel: let's not call this 'data'.  It makes it appear like we are actually getting the data back.

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
             return callback(null, multBackendResp);
         });
     },
+
+    createMPU: (key, bucket, metaHeaders, websiteRedirectHeader,
+    location, log, cb) => {
+        const client = clients[location];
+        const awsKey = _createAwsKey(bucket, key, client.bucketMatch);
```
LaurenSpiegel: might as well only create this within the if condition below

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
...
+                if (err) {
+                    log.error('err from data backend',
+                    { error: err, dataStore: client.dataStoreName });
+                    return cb(errors.InternalError);
+                }
+                return cb(null, data);
+            });
+        }
+        const scalResponse = {};
+        scalResponse.UploadId = UUID.v4().replace(/-/g, '');
```
LaurenSpiegel: I would do this back in the api so it's more transparent what is going on.

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
...
+
+    uploadPart: (bucket, location, key, uploadId, partNumber, log, cb) => {
+        const client = clients[location];
+        if (client.clientType === 'aws_s3') {
+            const awsKey = _createAwsKey(bucket, key, client.bucketMatch);
+            const params = { Bucket: client.awsBucketName, Key: awsKey,
+                UploadId: uploadId, PartNumber: partNumber };
+            return client.uploadPart(params, (err, data) => {
+                if (err) {
+                    log.error('err from data backend',
```
LaurenSpiegel: 'err from data backend on uploadPart'

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
...
+                    dataStoreETag: data.ETag,
+                };
+                return cb(null, dataRetrievalInfo);
+            });
+        }
+        return cb();
+    },
+
+    completeMPU: (key, uploadId, location, jsonList, log, cb) => {
+        const client = clients[location];
```
LaurenSpiegel: don't you still need to recreate the awskey?

why do you use the bucket as an argument in put part but pull the bucket from the client here?  shouldn't we always pull the bucket from the client info in case there is not a bucket match?

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
...
+        const partList = jsonList.Part;
+        if (client.clientType === 'aws_s3') {
+            const bucket = client.awsBucketName;
+            const partArray = [];
+            return async.each(partList, (partObj, callback) => {
```
LaurenSpiegel: why does this have to be async?

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
...
+                return callback();
+            }, () => {
+                const mpuParams = {
+                    Bucket: bucket, Key: key, UploadId: uploadId,
+                    MultipartUpload: {
+                        Parts: partArray,
+                    },
+                };
+                return client.completeMultipartUpload(mpuParams,
+                (err, data) => {
```
LaurenSpiegel: please use more descriptive name instead of 'data'

lib/data/multipleBackendGateway.js
```diff
@@ -216,6 +218,90 @@ const multipleBackendGateway = {
...
+                    if (err) {
+                        log.error('err from data backend',
```
LaurenSpiegel: 'err from data backend on completeMPU call'

LaurenSpiegel: I would only log at error level if the error is one of the "other" errors.  For instance, if it is an InvalidPart error, this shouldn't be an error log.  more like a trace log.

lib/data/wrapper.js
```diff
@@ -49,7 +49,8 @@ function _retryDelete(objectGetInfo, log, count, cb) {
 }
 
 const data = {
-    put: (cipherBundle, value, valueSize, keyContext, backendInfo, log, cb) => {
+    put: (cipherBundle, value, valueSize, keyContext, backendInfo,
```
LaurenSpiegel: if the only changes in this file are line breaks, please remove the changes.

tests/multipleBackend/multipartUpload.js
```diff
@@ -0,0 +1,169 @@
...
+                    assert.strictEqual(json.InitiateMultipartUploadResult
+                        .Key[0], objectKey);
+                    assert(json.InitiateMultipartUploadResult.UploadId[0]);
+                    const mpuKeys = metadata.keyMaps.get(mpuBucket);
+                    assert.strictEqual(mpuKeys.size, 1);
+                    assert(mpuKeys.keys().next().value
+                        .startsWith(`overview${splitter}${objectKey}`));
+                    awsParams.UploadId =
+                        json.InitiateMultipartUploadResult.UploadId[0];
+                    s3.abortMultipartUpload(awsParams, err => {
```
LaurenSpiegel: this should be in an after step

tests/multipleBackend/multipartUpload.js
```diff
@@ -0,0 +1,169 @@
...
+
+const awsParams = { Bucket: awsBucket, Key: objectKey };
+
+describe('Multipart Upload API with AWS Backend', () => {
+    afterEach(() => {
+        cleanup();
+    });
+
+    it('should initiate a multipart upload on real AWS', done => {
+        bucketPut(authInfo, bucketPutRequest, log, () => {
```
LaurenSpiegel: this should be in a before step.

tests/multipleBackend/multipartUpload.js
```diff
@@ -0,0 +1,169 @@
...
+            log, (err, result) =>
+                next(err, result)
+            ),
+            (result, next) => {
+                parseString(result, (err, json) => {
+                    const uploadId =
+                        json.InitiateMultipartUploadResult.UploadId[0];
+                    return next(err, uploadId);
+                });
+            },
```
LaurenSpiegel: the above should be in before steps

tests/multipleBackend/multipartUpload.js
```diff
@@ -0,0 +1,169 @@
...
+                    assert(MD);
+                    return next(err);
+                });
+            },
+        ], err => {
+            assert.strictEqual(err, null, `Error completing MPU: ${err}`);
+            done();
+        });
+    });
+});
```
LaurenSpiegel: we should be testing all of the mapped errors you have in the gateway -- invalid part order, etc.

LaurenSpiegel: we should be testing the overwriting like we did with regular object put (putting to file then completing an mpu on aws, etc.)

tests/multipleBackend/objectPutCopyPart.js
```diff
@@ -158,6 +161,18 @@ describeSkipIfE2E('ObjectCopyPutPart API with multiple backends', () => {
         });
     });
 
+    it('should copy part to AWS based on mpu location', done => {
+        copyPutPart('mem', 'aws-test', null, 'localhost', uploadId => {
+            assert.strictEqual(ds.length, 2);
+            s3.abortMultipartUpload({ Bucket: 'multitester444',
```
LaurenSpiegel: this should be in after step

LaurenSpiegel: let's do a list parts on aws to check it's there.  why do we expect the ds.length to be 2 here?

tests/multipleBackend/objectPutCopyPart.js
```diff
@@ -177,6 +192,18 @@ describeSkipIfE2E('ObjectCopyPutPart API with multiple backends', () => {
         });
     });
 
+    it('should copy part to AWS based on bucket location', done => {
+        copyPutPart('aws-test', null, null, 'localhost', uploadId => {
+            assert.deepStrictEqual(ds, []);
+            s3.abortMultipartUpload({ Bucket: 'multitester444',
```
LaurenSpiegel: same comments as above

tests/multipleBackend/multipartUpload.js
LaurenSpiegel: we should test putting the same part number multiple times on real aws.

LaurenSpiegel: we should also be testing listing of mpu's and separately testing listing of parts.

lib/api/completeMultipartUpload.js
```diff
@@ -112,6 +115,12 @@ function completeMultipartUpload(authInfo, request, log, callback) {
         return callback(queryContainsVersionId);
     }
 
+    // Reconstruct mpuOverviewKey to serve
+    // as key to pull metadata originally stored when mpu initiated
+    const mpuOverviewKey =
```
dora-korpar: yeah, I fixed this in my other MPU branch but forgot to fix it here >.< thanks for catching

lib/api/completeMultipartUpload.js
```diff
@@ -155,19 +164,59 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                const location = storedMetadata.controllingLocationConstraint;
+                return next(null, destBucket, objMD, mpuBucket, jsonList,
+                storedMetadata, location);
+            });
+        },
+        function ifMultipleBackend(destBucket, objMD, mpuBucket, jsonList,
+        storedMetadata, location, next) {
+            if (config.backends.data === 'multiple') {
+                return multipleBackendGateway.completeMPU(objectKey,
+                uploadId, location, jsonList, log, (err, newObjData) => {
```
dora-korpar: It's the object returned from AWS on a complete. It includes the ETag, bucket name, key name, location url, version id, and expiration if there's any lifecycle settings on the bucket.

lib/api/completeMultipartUpload.js
```diff
@@ -155,19 +164,59 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
                     const storedParts = result.Contents;
                     return next(null, destBucket, objMD, mpuBucket,
-                        storedParts, jsonList);
+                        storedParts, jsonList, storedMetadata, null);
                 });
         },
         function processParts(destBucket, objMD, mpuBucket, storedParts,
-            jsonList, next) {
+            jsonList, storedMetadata, newObjData, next) {
+            let calculatedSize = 0;
```
dora-korpar: the calculatedSize line was just moved there from further down in the function so I could return it in the if statement. Since it was just a one-line move, I thought it would be better to pass the variable than just 'null'

lib/api/completeMultipartUpload.js
dora-korpar: yeah, I guess this might not work with other backends. I'll add some processing of the object returned from AWS in the gateway

dora-korpar:
@ironman-machine go

lib/data/external/AwsClient.js
```diff
@@ -111,6 +113,89 @@ class AwsClient {
...
+            };
+            return callback(null, dataRetrievalInfo);
+        });
+    }
+
+    completeMPU(partList, key, uploadId, log, callback) {
+        const bucket = this._awsBucketName;
+        const awsKey = this._createAwsKey(bucket, key, this._bucketMatch);
+        const partArray = [];
+        return async.each(partList, (partObj, callback) => {
```
nicolas2bert: We may not want to use `async.each` here since the iteratee function is synchronous.

lib/data/external/AwsClient.js
```diff
@@ -111,6 +113,89 @@ class AwsClient {
...
+            const mpuParams = {
+                Bucket: bucket, Key: awsKey, UploadId: uploadId,
+                MultipartUpload: {
+                    Parts: partArray,
+                },
+            };
+            return this._client.completeMultipartUpload(mpuParams,
+            (err, completeMpuResObj) => {
+                if (err) {
+                    if (err.code === 'InvalidPart') {
```
nicolas2bert: I am ok with this - We also could have create an object `const mpuError = {InvalidPart: true, InvalidPartOrder: true, EntityTooSmall: true }` and then check `if( mpuError[err.code]) { log.trace(...) ; return errors[err.code]; } so you won't have to `log.trace` every time. Not a big deal tho :)

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

lib/api/completeMultipartUpload.js
```diff
@@ -155,19 +157,71 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
                     return next(null, destBucket, objMD, mpuBucket,
-                        storedParts, jsonList);
+                        storedParts, jsonList, storedMetadata, null,
+                        mpuOverviewKey);
                 });
         },
         function processParts(destBucket, objMD, mpuBucket, storedParts,
-            jsonList, next) {
+            jsonList, storedMetadata, completeObjData, mpuOverviewKey, next) {
+            let calculatedSize = 0;
```
LaurenSpiegel: we shouldn't be storing a size of 0 for an mpu put on aws.  we will have to do a head request on the completed object and use that for the size.  we should add a test checking it.

tests/multipleBackend/objectPutPart.js
```diff
@@ -145,15 +155,67 @@ describeSkipIfE2E('objectPutPart API with multiple backends', () => {
...
             done();
         });
     });
 
+    it('should put a part to AWS based on mpu location', done => {
+        putPart('file', 'aws-test', null, 'localhost', uploadId => {
+            assert.deepStrictEqual(ds, []);
+            const params = { Bucket: awsBucket, Key: objectName,
+            UploadId: uploadId };
+            s3.listParts(params, (err, data) => {
```
LaurenSpiegel: shouldn't this be in an after step?

tests/multipleBackend/objectPutPart.js
```diff
@@ -145,15 +155,67 @@ describeSkipIfE2E('objectPutPart API with multiple backends', () => {
...
+                query: {
+                    partNumber: '1', uploadId,
+                },
+            };
+            const partReq = new DummyRequest(partReqParams, body2);
+            objectPutPart(authInfo, partReq, undefined, log, err => {
+                assert.equal(err, null, `Error putting second part: ${err}`);
+                const params = { Bucket: awsBucket, Key: objectName,
+                    UploadId: uploadId };
+                s3.listParts(params, (err, data) => {
```
LaurenSpiegel: same -- should be in an after step?

tests/multipleBackend/objectPutPart.js
```diff
@@ -167,11 +229,29 @@ describeSkipIfE2E('objectPutPart API with multiple backends', () => {
...
         });
     });
 
+    it('should put a part to AWS based on bucket location', done => {
+        putPart('aws-test', null, null, 'localhost',
+        uploadId => {
+            assert.deepStrictEqual(ds, []);
+            const params = { Bucket: awsBucket, Key: objectName,
+            UploadId: uploadId };
+            s3.listParts(params, (err, data) => {
```
LaurenSpiegel: same

tests/multipleBackend/multipartUpload.js
```diff
@@ -0,0 +1,253 @@
+const assert = require('assert');
+const AWS = require('aws-sdk');
```
LaurenSpiegel: do we have any tests where we overwrite a key that was put to file by completing an mpu with the same key name and the other way around to make sure we are cleaning up correctly?

LaurenSpiegel: could we also do a test where versioning is enabled on scality s3 and we complete an mpu to show that it creates a new version on scality s3?

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine do it

dora-korpar:
@ironman-machine

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/excludeRubyAwsTests

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

PR #759, opened 2017-06-03
FT: Add replication information to object MD

bennettbuchanan:
Add property `replicationInfo` to an object's metadata. Set the values conditionally for Backbeat to know which objects need to be replicated and whether to replicate the object data and/or metadata.

### DIFF
--- a/lib/api/apiUtils/object/replication.js
+++ b/lib/api/apiUtils/object/replication.js
@@ -1,0 +1,57 @@
+/**
+ * Check that a bucket replication rule applies for the given object key. If it
+ * does, assign replication information to the given object.
+ * @param {object} obj - The object to set replicationInfo for
+ * @param {string} objKey - The key of the object
+ * @param {object} bucketMD - The bucket metadata
+ * @param {array} content - The content type that should be replicated
+ * @return {undefined}
+ */
+function buildReplicationInfo(obj, objKey, bucketMD, content) {
+    const config = bucketMD.getReplicationConfiguration();
+    // If bucket does not have a replication configuration, do not replicate.
+    if (config) {
+        const rule = config.rules.find(rule => objKey.startsWith(rule.prefix));
+        if (rule) {
+            // eslint-disable-next-line no-param-reassign
+            obj.replicationInfo = {
+                status: 'PENDING',
+                content,
+                destination: config.destination,
+                storageClass: rule.storageClass || '',
+            };
+        }
+    }
+    return undefined;
+}
+
+/**
+ * Set the object replicationInfo to replicate data and metadata, or only
+ * metadata if the object is a delete marker
+ * @param {object} obj - The object to set replicationInfo for
+ * @param {string} objKey - The key of the object
+ * @param {object} bucketMD - The bucket metadata
+ * @param {boolean} isDeleteMarker - Whether the object is a deleteMarker
+ * @return {undefined}
+ */
+function buildReplicationInfoForObject(obj, objKey, bucketMD, isDeleteMarker) {
+    // Delete markers have no data, so we only replicate metadata.
+    const content = isDeleteMarker ? ['METADATA'] : ['DATA', 'METADATA'];
+    return buildReplicationInfo(obj, objKey, bucketMD, content);
+}
+
+/**
+ * Set the object replicationInfo to replicate only metadata
+ * @param {object} obj - The object to set replicationInfo for
+ * @param {string} objKey - The key of the object
+ * @param {object} bucketMD - The bucket metadata
+ * @return {undefined}
+ */
+function buildReplicationInfoForObjectMD(obj, objKey, bucketMD) {
+    return buildReplicationInfo(obj, objKey, bucketMD, ['METADATA']);
+}
+
+module.exports = {
+    buildReplicationInfoForObject,
+    buildReplicationInfoForObjectMD,
+};
--- a/lib/metadata/models/object/ObjectMD.js
+++ b/lib/metadata/models/object/ObjectMD.js
@@ -53,6 +53,12 @@ module.exports = class ObjectMD {
             'isDeleteMarker': '',
             'versionId': undefined, // If no versionId, it should be undefined
             'tags': {},
+            'replicationInfo': {
+                status: '',
+                content: [],
+                destination: '',
+                storageClass: '',
+            },
         };
     }
 
@@ -568,6 +574,32 @@ module.exports = class ObjectMD {
      */
     getTags() {
         return this._data.tags;
+    }
+
+    /**
+     * Set replication information
+     *
+     * @param {object} replicationInfo - replication information object
+     * @return {ObjectMD} itself
+     */
+    setReplicationInfo(replicationInfo) {
+        const { status, content, destination, storageClass } = replicationInfo;
+        this._data.replicationInfo = {
+            status,
+            content,
+            destination,
+            storageClass: storageClass || '',
+        };
+        return this;
+    }
+
+    /**
+     * Get replication information
+     *
+     * @return {object} replication object
+     */
+    getReplicationInfo() {
+        return this._data.replicationInfo;
     }
 
     /**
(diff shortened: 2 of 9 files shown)


tests/unit/api/objectReplicationMD.js
```diff
@@ -0,0 +1,295 @@
...
+    content: ['DATA', 'METADATA'],
+    destination: bucketARN,
+    storageClass: 'STANDARD',
+};
+
+const replicateMetadataOnly =
+    Object.assign({}, newReplicationMD, { content: ['METADATA'] });
+
+// Check that the object key has the expected replication information.
+function checkObjectMD(key, expected) {
```
jonathan-gramain: The name should reflect that it is checking the replication info only

tests/unit/api/objectReplicationMD.js
```diff
@@ -0,0 +1,295 @@
...
+            next => bucketPutReplication(authInfo, replicationReq, log, next),
+        ], err => done(err));
+    });
+
+    afterEach(() => cleanup());
+
+    it('should update metadata when replication config prefix applies', done =>
+        putObjectAndCheckMD(keyA, newReplicationMD, done));
+
+    it('should not update metadata when replication config prefix does not ' +
```
jonathan-gramain: Could also be worth adding tests with `${keyA}something` and `${keyB}something` as keys to check that prefix matching code is correct.

lib/utils.js
```diff
@@ -312,4 +312,76 @@ utils.isUnsupportedQuery = function isUnsupportedQuery(queryObj) {
...
+
+/**
+ * Set the object's metatdata to indicate that the object's data and metadata
+ * should be replicated. If object is delete marker, replicate only metadata.
+ * @param {object} bucketMD - The bucket metadata
+ * @param {string} objKey - The key of the object
+ * @param {object} objMD - The object metadata
+ * @return {undefined}
+ */
+utils.replicateObject = function replicateObject(bucketMD, objKey, objMD) {
```
bennettbuchanan: Updated.

lib/utils.js
```diff
@@ -312,4 +312,76 @@ utils.isUnsupportedQuery = function isUnsupportedQuery(queryObj) {
...
+
+/**
+ * Set the object's metadata to indicate that only the object's metadata should
+ * be replicated
+ * @param {object} bucketMD - The bucket metadata
+ * @param {string} objKey - The key of the object
+ * @param {object} objMD - The object metadata
+ * @return {undefined}
+ */
+utils.replicateObjectMD = function replicateObjectMD(bucketMD, objKey, objMD) {
```
bennettbuchanan: Agree. I've updated these methods.

tests/unit/api/objectReplicationMD.js
bennettbuchanan: Agree. Updated.

tests/unit/api/objectReplicationMD.js
bennettbuchanan: Good idea. I have added some tests with characters `'acb'` on either side of `${keyA}`. I think the existing test for `${keyB}` should be sufficient with this.

lib/utils.js
```diff
@@ -320,4 +320,87 @@ utils.isUnsupportedQuery = function isUnsupportedQuery(queryObj) {
...
+
+/**
+ * Check the bucket replication configuration and set replication content
+ * @param {object} bucketMD - The bucket metadata
+ * @param {string} objKey - The key of the object
+ * @param {object} objMD - The object metadata
+ * @param {array} content - The content of the object to be replicated
+ * @return {undefined}
+ */
+utils.handleReplicationInfo =
```
jonathan-gramain: Sorry to be picky (again) I find this name confusing, and this function should stay internal instead of being part of the `utils` object. Maybe just call it `setReplicationInfoHelper` since it's not fundamentally different than the two other setters and is meant to be only a helper function.

Otherwise all other changes are good to me.

bennettbuchanan: I am actually moving these files out of utils.js and into their very own apiUtils/object/replication.js file. I will update names then and indicate it is a private method.

bennettbuchanan:
@ironman-machine try

tests/unit/api/objectReplicationMD.js
```diff
@@ -0,0 +1,304 @@
+const assert = require('assert');
+const async = require('async');
+
+const { cleanup, DummyRequestLogger, makeAuthInfo, TaggingConfigTester } =
+    require('../helpers');
+const { metadata } = require('../../../lib/metadata/in_memory/metadata');
+const DummyRequest = require('../DummyRequest');
+const { bucketPut } = require('../../../lib/api/bucketPut');
+const objectDelete = require('../../../lib/api/objectDelete');
```
rahulreddy: Ideally unit tests should not rely on other APIs. Is there a way you can think of mocking the data?

lib/api/apiUtils/object/replication.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Check that a bucket replication rule applies for the given object key. If it
+ * does, assign replication information to the given object.
+ * @param {object} obj - The object to set replicationInfo for
+ * @param {string} objKey - The key of the object
+ * @param {object} bucketMD - The bucket metadata
+ * @param {boolean} isDeleteMarker - Whether the object is a deleteMarker
+ * @return {undefined}
+ */
+function assignObjectReplicationInfo(obj, objKey, bucketMD, isDeleteMarker) {
```
bennettbuchanan: I've renamed the main function `buildReplicationInfo`, but have kept the file name since I've added two wrapper functions (see next comment).

lib/api/apiUtils/object/replication.js
```diff
@@ -0,0 +1,46 @@
...
+    const config = bucketMD.getReplicationConfiguration();
+    // If bucket does not have a replication configuration, do not replicate.
+    if (config) {
+        const rule = config.rules.find(rule => objKey.startsWith(rule.prefix));
+        if (rule) {
+            // Delete markers have no data, so we only replicate metadata.
+            const content = isDeleteMarker || isDeleteMarker === undefined ?
```
bennettbuchanan: I've broken this into two wrapper functions around `buildReplicationInfo` that simplifies this logic.

lib/metadata/models/object/ObjectMD.js
```diff
@@ -571,6 +577,26 @@ module.exports = class ObjectMD {
     }
 
     /**
+     * Set replication information
+     *
+     * @param {object} replicationInfo - replication information object
+     * @return {ObjectMD} itself
+     */
+    setReplicationInfo(replicationInfo) {
```
bennettbuchanan: If `storageClass` is not included in the rule, I assume we want to keep `replicationInfo.storageClass` as `''` in the object MD.

bennettbuchanan: Updated.

tests/unit/api/objectReplicationMD.js
bennettbuchanan: I've mocked the APIs for bucketPut, bucketPutVersioning, and bucketPutReplication. The remaining APIs follow the paths to apiUtil replication methods added by this PR, so I've kept them. We could also remove them and test the apiUtil methods directly, but I'm thinking this way might be a bit more thorough.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3-REP.1
try

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3-REP.1
DEFAULT_BRANCH=master
try

bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=master
SCALITY_ARSENAL_BRANCH=ft/S3-REP.1
try

bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=master
SCALITY_ARSENAL_BRANCH=ft/S3-REP.1
try

bennettbuchanan:
E2E passed with Arsenal branch ft/S3-REP.1.

PR #760, opened 2017-06-07
FT: Add deleteBucketReplication API

bennettbuchanan:


lib/api/bucketDeleteReplication.js
```diff
@@ -0,0 +1,56 @@
...
+                error: err,
+                method: 'bucketDeleteReplication',
+            });
+            return callback(err, corsHeaders);
+        }
+        if (!bucket.getReplicationConfiguration()) {
+            log.trace('no existing replication configuration', {
+                method: 'bucketDeleteReplication',
+            });
+            pushMetric('deleteBucketReplication', log, {
```
jonathan-gramain: philosophical question: should we count the no-op deletes in metrics?

- if the goal is to count the number of user operations executed, it's likely a yes
- if to count the actual actions that change a state, it would rather not be counted

Likely not counting those would give a counter semantic closer to the semantic of `putBucketReplication` counter (except you can re-write the policy multiple times but can only delete it once so they're not directly comparable).

bennettbuchanan: I had wondered the same, and saw that other DELETE operation APIs are doing it this way.

rahulreddy: Logging all successful operations is a way of providing a sort of cloudtrail style logs to see what actions happened in an S3 account.

bennettbuchanan: Seems it hinges on a philosophical definition of 'success' @jonathan-gramain. 😸

rahulreddy: Utapi is underrated but you can technically paint a timeline of actions that happened on that S3 cluster, which IMO is awesome.

tests/functional/aws-node-sdk/test/bucket/deleteBucketReplication.js
```diff
@@ -0,0 +1,102 @@
...
+        otherAccountS3 = new BucketUtility('lisa', {}).s3;
+        return s3.createBucket({ Bucket: bucket }, done);
+    });
+
+    afterEach(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+    it('should return empty object if bucket is not versioned enabled', done =>
+        deleteReplicationAndCheckResponse(bucket, done));
+
+    it('should return empty object if bucket is version enabled but has no ' +
```
electrachong: it doesn't seem like you check if versioning is enabled in the deleteBucketReplication api, so just wondering why you have an extra test?

bennettbuchanan: You're right—this is overkill. There really only needs to be one test for a case in which the bucket does not have a replication configuration, and another for when it does. I've updated the tests. Thanks! 😸

rahulreddy:
@bennettbuchanan Rebase and squash, I will merge this.  Don't forget to remove the Arsenal dep. branch.

bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=master
try

bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=master
SCALITY_ARSENAL_BRANCH=ft/S3-REP.1
try

bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=master
SCALITY_ARSENAL_BRANCH=ft/S3-REP.1
try

bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=master
SCALITY_ARSENAL_BRANCH=ft/S3-REP.1
try

bennettbuchanan:
E2E passed with Arsenal branch ft/S3-REP.1.

PR #766, opened 2017-06-09
FIX: Allow operation if have valid loc

LaurenSpiegel:
1) If have object level location constraint
or bucket level location constraint allow
operation to continue even if restEndpoint is
not in config.

2) Set bucket location to us-east-1 if no location provided and no default for endpoint.


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
@ironman-machine go again

LaurenSpiegel:
@ironman-machine once more

LaurenSpiegel:
end to end passed

PR #768, opened 2017-06-12
Rf/extract routes to Arsenal [S3C-511]

electrachong:
Dependent on https://github.com/scality/Arsenal/pull/268/

Extracting routes for use in other S3-related projects.


lib/server.js
```diff
@@ -40,7 +63,17 @@ class S3Server {
...
         // disable nagle algorithm
         req.socket.setNoDelay();
-        routes(req, res, logger);
+        const params = {
+            api,
+            healthcheckHandler,
+            statsClient,
+            allEndpoints,
+            websiteEndpoints,
+            isValidBucketName,
```
LaurenSpiegel: not clear why we're passing this one

lib/utils.js
```diff
@@ -1,67 +1,12 @@
...
-const { ipCheck, errors } = require('arsenal');
+const { errors } = require('arsenal');
 
 const { config } = require('./Config');
 const constants = require('../constants');
 
 const utils = {};
 
 /**
- * Get bucket name from the request of a virtually hosted bucket
```
LaurenSpiegel: How did you decide what to leave in this file? Is there any unifying theme to the remaining functions? Should we put them somewhere else?

electrachong: I left `isValidBucketName` because it checks if the bucket starts with `constants.mpuBucketPrefix` which I figured would not be relevant for Mystique. I was thinking we might create an equivalent function in the other repo since there may be additional constraints for bucket name validity.

However, since it seems like the rest of the checks are applicable, we could refactor `isValidBucketName` to accept an array of strings to check if `bucketname.startsWith(...)` and move it to Arsenal in case that might be useful. Or we could create a class/function in Arsenal which accepts an array of additional functions as checks to run on `bucketName`. 👍 

`getAllEndpoints` I left because it depends on the local config. I guess I can move this function and pass the endpoints instead of the keys instead.

The rest of the functions I moved were used directly in the routes. I left any functions that were not used by the routes, so I guess most of them are used by the apis and can be moved to `apiUtils` if we want.

electrachong:
@ironman-machine 
SCALITY_ARSENAL_BRANCH=rf/extract-s3routes

electrachong:
end-to-end retried after fixing typo: http://ci.ironmann.io/gh/scality/Integration/13542

electrachong:
last end-to-end, just in case
@ironman-machine

PR #771, opened 2017-06-14
BF S3C-531 chordChos must be 1 digit string smaller than 7

msegura:
When using the scality sproxyd storage, the chordCos should be a decimal integer smaller than 7 (bigger number are erasure coding and not replication.

Currently the chordCos is supposed to be a 2 hexadecimal digits string. This PR changes that.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -24,8 +24,8 @@ function sproxydAssert(configSproxyd) {
     if (configSproxyd.chordCos !== undefined) {
         assert(typeof configSproxyd.chordCos === 'string',
             'bad config: sproxyd.chordCos must be a string');
-        assert(configSproxyd.chordCos.match(/^[0-9a-fA-F]{2}$/),
-            'bad config: sproxyd.chordCos must be a 2hex-chars string');
+        assert(configSproxyd.chordCos.match(/^[0-6]{1}$/),
+         'bad config: sproxyd.chordCos must be a digit smaller than 7');
         sproxydFields.push('chordCos');
     }
     if (configSproxyd.path !== undefined) {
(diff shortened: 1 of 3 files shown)


msegura:
@ironman-machine go

lib/Config.js
```diff
@@ -91,7 +91,7 @@ function locationConstraintAssert(locationConstraints) {
 function cosParse(chordCos) {
     // Cos number should only be first digit of config value
     const firstDigit = chordCos[0];
```
LaurenSpiegel: can we remove this now and just do a pareInt on chordCos?  Is the former second digit referenced anywhere?

tests/unit/testConfigs/sproxydAssert.js
```diff
@@ -37,17 +37,17 @@ describe('sproxydAssert', () => {
...
-    it('should throw an error if chordCos is more than 2 digits', () => {
+    it('should throw an error if chordCos is more than 1 digit', () => {
         assert.throws(() => {
             sproxydAssert(makeSproxydConf(null, '200'));
         },
-        /bad config: sproxyd.chordCos must be a 2hex-chars string/);
+        /bad config: sproxyd.chordCos must be one digit/);
     });
-    it('should throw an error if chordCos includes non-hex digit', () => {
+    it('should throw an error if chordCos includes decimal digit', () => {
```
LaurenSpiegel: maybe change to just be test to make sure it is not greater than 6?

msegura:
@ironman-machine go

lib/Config.js
msegura: ok

tests/unit/testConfigs/sproxydAssert.js
msegura: ok

msegura:
@ironman-machine go

msegura:
I think Ultron got a bit crazy (I deleted some messages it repeating). The end to end seems fine however.

msegura:
@ironman-machine go

PR #772, opened 2017-06-14
Fix multiple backend tests

dora-korpar:
This PR fixes multiple backend tests that were not being run and their socket hangups

### DIFF
--- a/lib/data/multipleBackendGateway.js
+++ b/lib/data/multipleBackendGateway.js
@@ -121,6 +121,7 @@ const multipleBackendGateway = {
     },
 
     get: (objectGetInfo, range, reqUids, callback) => {
+        const log = createLogger(reqUids);
         let key;
         let client;
         // for backwards compatibility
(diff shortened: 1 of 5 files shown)


dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

lib/data/multipleBackendGateway.js
```diff
@@ -134,13 +135,20 @@ const multipleBackendGateway = {
         if (client.clientType === 'scality') {
             return client.get(key, range, reqUids, callback);
         }
+
         if (client.clientType === 'aws_s3') {
-            const stream = client.getObject({
+            const action = client.getObject;
+            const method = action.bind(client);
+            const request = method({
```
nicolas2bert: I might be wrong, but I think we can call client.getObject directly instead of binding the client.
`const request = client.getObject({`

dora-korpar: Yeah, we did that as a work-around to be able to access the response headers

nicolas2bert: Ok - we can leave it like that here, I will test if we can get rid of it in the refactor.

PR #775, opened 2017-06-17
Skip tests involving AWS

dora-korpar:
Resolving the socket hangups from AWS has proven to be more challenging than anticipated :( To stop the blocking of unrelated PRs, we will skip the tests for now.


dora-korpar:
@ironman-machine go

PR #777, opened 2017-06-19
New docker envionment variables

RachedBenMustapha:
This change allows changing more configuration options through environment variables in the `docker-entrypoint.sh` script.

The generated image is testable at `scality/s3server:with-jq`

The primary motivation is to avoid having to generate and distribute configuration files to dynamically scheduled container environments such as swarm.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /usr/src/app
 COPY . /usr/src/app
 
 RUN apt-get update \
-    && apt-get install -y python git build-essential --no-install-recommends \
+    && apt-get install -y jq python git build-essential --no-install-recommends \
     && npm install --production \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR /usr/src/app
 COPY . /usr/src/app
 
 RUN apt-get update \
-    && apt-get install -y python git build-essential --no-install-recommends \
+    && apt-get install -y jq python git build-essential --no-install-recommends \
     && npm install --production \
     && apt-get autoremove --purge -y python git build-essential \
```
RachedBenMustapha: It's actually needed at runtime, but it's ok because the package is tiny :)

nicolas2bert: You are right, no problemo.

PR #778, opened 2017-06-19
ft: update versioning doc to reflect changes

electrachong:
Will need @nicolas2bert's help to figure out how to render the documents for preview on the readthedocs :smile:

See rendered changes here: http://s3-server.readthedocs.io/en/doc-adddoc/ARCHITECTURE/#versioning

### DIFF
--- a/docs/ARCHITECTURE.rst
+++ b/docs/ARCHITECTURE.rst
@@ -177,11 +177,13 @@ PUT
 -  no options: original PUT operation, will update the master version
 -  ``versioning: true`` create a new version of the object, then update
    the master version with this version.
--  ``versionId: <versionId>`` update a specific version (for updating
-   version's ACL or tags, or remote updates in geo-replication
-
+-  ``versionId: <versionId>`` create or update a specific version (for updating
+   version's ACL or tags, or remote updates in geo-replication)
    -  if the version identified by ``versionId`` happens to be the latest
       version, the master version will be updated as well
+   -  if the master version is not as recent as the version identified by
+      ``versionId``, as may happen with cross-region replication, the master
+      will be updated as well
    -  note that with ``versionId`` set to an empty string ``''``, it will
       overwrite the master version only (same as no options, but the master
       version will have a ``versionId`` property set in its metadata like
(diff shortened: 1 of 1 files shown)


docs/ARCHITECTURE.rst
```diff
@@ -368,7 +368,11 @@ version. S3 checks if versioning is suspended or has never been
 configured, and sets the ``versionId`` option to ``''`` in PUT calls to
 the metadata engine when creating a new null version.
 
-The tables below show the keys resulting from PUT calls to metadata if
+If the master version is a null version, S3 also sends a DELETE call to metadata
+prior to the PUT, in order to clean up any pre-existing null versions which may,
+in certain edge cases, have been stored as a separate version.[#1]_
```
bennettbuchanan: This footnote doesn't seem to be rendering properly when previewed.

jonathan-gramain:
When https://github.com/scality/Arsenal/pull/262 gets merged, it will change a bit how versioning code behaves:

The "Metadata Versioning Options" should be updated to reflect this: when `versioning` and `versionId` are both set, instead of saying that it returns an error, it could say that it's only used for internal needs (e.g. CRR). What is does is create or update a specific version + the master only if the master is older than the written version. I'm not convinced we need to explain this in the doc though.

I guess the easiest is to update the doc in this PR directly.

electrachong:
@jonathan-gramain does the update look OK for when your pr is merged?

docs/ARCHITECTURE.rst
```diff
@@ -192,10 +192,9 @@ PUT
       versions once versioning has been suspended, which is discussed in
       `"Null Version Management" <#null-version-management>`__.
 
-Only one option is used at a time. ``versionId: <versionId>`` does not
-have to be used with ``versioning: true`` set to work, nor should they
-both be set. If both are used at once, the metadata engine will return
-an error.
+In general, only one option is used at a time. When ``versionId`` and
+``versioning`` are both set, it is for internal needs (e.g. cross-region
```
LaurenSpiegel: If versionId and versioning is set, then a new version with versionId will be created and master will be updated if the versionId is later than the current master.

electrachong:
@jonathan-gramain updated again - if you can proof sometime, can you confirm you agree with the description?

jonathan-gramain:
@electrachong Looks good to me!

PR #781, opened 2017-06-20
Update how multiple backend tests are run

dora-korpar:
It all started with some mocha 'it's that should've been 'describe's, which meant some tests weren't updated in the last few PR's that should've been. This PR fixes the describes, removes an unnecessary error check, and runs the tests with mdserver.js


dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go

PR #784, opened 2017-06-21
FT: More info on location err message

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
end to end failed for unrelated reason. otherwise passed.

PR #787, opened 2017-06-22
bucketfile record log service for backbeat

jonathan-gramain:
Enable or disable record log service provided in https://github.com/scality/Arsenal/pull/253

### DIFF
--- a/config.json
+++ b/config.json
@@ -53,5 +53,9 @@
     "dataDaemon": {
         "bindAddress": "localhost",
         "port": 9991
+    },
+    "recordLog": {
+        "enabled": false,
+        "recordLogName": "s3-recordlog"
     }
 }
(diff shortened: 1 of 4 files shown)


config.json
```diff
@@ -53,5 +53,9 @@
     "dataDaemon": {
         "bindAddress": "localhost",
         "port": 9991
+    },
+    "recordLog": {
+        "enabled": false,
+        "recordLogName": "main"
```
rahulreddy: Since this is the default and a lot of people don't change defaults, it's better to give it a specific name. May be
`s3-recordLog`?

jonathan-gramain: I agree, will rename it.

jonathan-gramain:
@ironman-machine try

PR #788, opened 2017-06-23
Rf/unsupported check

electrachong:
Dependent on https://github.com/scality/Arsenal/pull/278

Moving check for unsupported queries and headers from routes to api.
The routes are a shared module between s3 projects which may diverge in their support based on different parameters (request method, etc).

This also means we won't have to pass another param to routes (were passing unsupportedQueries)


electrachong:
@ironman-machine 
SCALITY_ARSENAL_BRANCH=rf/unsupportedCheck

electrachong:
end-to-end retried here: http://ci.ironmann.io/gh/scality/Integration/13690

constants.js
```diff
@@ -87,6 +87,17 @@ const constants = {
         'restore': true,
         'torrent': true,
     },
+
+    // Headers supported by AWS that we do not currently support.
+    encryptionHeaders: [
```
LaurenSpiegel: I would generalize to unsupportedHeaders.  I would do an object for both as well.  Or as discussed an array if we loop through this and then check the reqHeaders.

lib/api/api.js
```diff
@@ -50,6 +51,18 @@ auth.setHandler(vault);
...
+
+        const validationRes =
+            validateQueryAndHeaders(request.method, request.query,
+                request.headers, log);
+        if (validationRes.error) {
+            log.debug('request query / header validation failed', {
+                error: validationRes.error,
+                method: 'api.callApiMethod',
+            });
+            return callback(validationRes.error);
```
nicolas2bert: We might need: `proccess.nextTick` here but also for the errors after.
So you might want to add them in another commit.

constants.js
electrachong: updated

lib/api/api.js
electrachong: updated

electrachong:
after changes
@ironman-machine
SCALITY_ARSENAL_BRANCH=rf/unsupportedCheck

electrachong:
end-to-end passed

PR #796, opened 2017-06-29
PORT: port/ft/S3C-291-delete-bucket-replication to master

bennettbuchanan:


bennettbuchanan:
Moving this commit into a single PR with changes pertaining to replication (instead of porting from ft/S3-REP.1).

PR #798, opened 2017-06-29
FIX: Use correct method for replication test

bennettbuchanan:


bennettbuchanan:
Moving this commit into a single PR with changes pertaining to replication.

PR #800, opened 2017-06-29
Add abortMPU for AWS backend

dora-korpar:


lib/data/external/AwsClient.js
```diff
@@ -208,6 +208,22 @@ class AwsClient {
...
+
+    abortMPU(key, uploadId, location, bucket, log, callback) {
+        const awsBucket = this._awsBucketName;
+        const awsKey = this._createAwsKey(bucket, key, this._bucketMatch);
+        const abortParams = {
+            Bucket: awsBucket, Key: awsKey, UploadId: uploadId,
+        };
+        return this._client.abortMultipartUpload(abortParams, err => {
+            if (err) {
+                log.trace('err from data backend on abortMPU',
```
LaurenSpiegel: this should be log.error

lib/data/external/AwsClient.js
```diff
@@ -208,6 +208,22 @@ class AwsClient {
...
+                { error: err, dataStoreNae: this._dataStoreName });
```
LaurenSpiegel: can you please make sure the actual error is showing up in the logs?  we might have to do an err.message for the error to actually appear.

lib/data/multipleBackendGateway.js
```diff
@@ -146,6 +146,20 @@ const multipleBackendGateway = {
         }
         return cb();
     },
+
+    abortMPU: (key, uploadId, location, bucket, log, cb) => {
+        const client = clients[location];
+        if (client.clientType === 'aws_s3') {
```
LaurenSpiegel: can we do if client.abortMPU so we are checking whether the method is implemented rather than whether it is particularly s3.  same on the completeMPU call.

lib/data/external/AwsClient.js
```diff
@@ -208,6 +208,22 @@ class AwsClient {
...
+                return callback(errors.InternalError);
```
LaurenSpiegel: let's customize the error description with something like:
'There was an error aborting the MPU on AWS S3. You should abort directly on AWS S3 using the same uploadId.'

Otherwise they will have an orphan on AWS.

lib/api/multipartDelete.js
```diff
@@ -69,23 +72,46 @@ function multipartDelete(authInfo, request, log, callback) {
                     return next(err, mpuBucket, mpuOverviewArray, destBucket);
                 });
         },
+        function getMPUMetadata(mpuBucket, mpuOverviewArray, destBucket, next) {
```
LaurenSpiegel: why was this moved up?

tests/multipleBackend/multipartUpload.js
```diff
@@ -235,6 +243,32 @@ describe('Multipart Upload API with AWS Backend', function mpuTestSuite() {
         });
     });
 
+    it('should abort a multipart upload on real AWS', done => {
+        mpuSetup(uploadId => {
```
LaurenSpiegel: let's do a test where there are actual parts put on aws

tests/multipleBackend/multipartUpload.js
```diff
@@ -235,6 +243,32 @@ describe('Multipart Upload API with AWS Backend', function mpuTestSuite() {
...
+                assert.equal(err, null, `Error aborting MPU: ${err}`);
+                s3.listParts({ Bucket: awsBucket, Key: objectKey,
+                UploadId: uploadId }, err => {
+                    assert.strictEqual(err.code, 'NoSuchUpload');
+                    done();
+                });
+            });
+        });
+    });
+
```
LaurenSpiegel: what happens if you initiate an mpu through scality s3 and then abort it directly on real aws?  how do we respond?

tests/multipleBackend/multipartUpload.js
dora-korpar: mpuSetup initiates the mpu and puts two parts :)

lib/data/external/AwsClient.js
dora-korpar: no orphans! 👍

lib/data/multipleBackendGateway.js
dora-korpar: I'll do a followup PR to change that logic in the other three MPU functions as well

tests/multipleBackend/multipartUpload.js
dora-korpar: It responds with no error, same as with a non-existant uploadId

LaurenSpiegel: except if it is in a region where an error is returned by aws, they would return NoSuchUpload and we would translate that to InternalError which is not actually bad behavior since they caused a diversion in state on scality s3 and aws s3 by aborting directly on aws.

dora-korpar:
@ironman-machine go

lib/api/multipartDelete.js
```diff
@@ -122,14 +142,26 @@ function multipartDelete(authInfo, request, log, callback) {
                         bucket: bucketName,
                         byteLength: partSizeSum,
                     });
+                    return next(null, destBucket, location);
+                });
+        },
+        function ifMultipleBackend(destBucket, location, next) {
+            if (config.backends.data === 'multiple') {
+                return multipleBackendGateway.abortMPU(objectKey,
+                uploadId, location, destBucket, log, err => {
```
jonathan-gramain: Could `location` be undefined here, say, if the object was put without a location constraint originally? If the answer is yes, it could be wise to check this and skip as well if no location exists for the object.

lib/api/multipartDelete.js
```diff
@@ -69,23 +72,46 @@ function multipartDelete(authInfo, request, log, callback) {
...
+            let splitter = constants.splitter;
+            // BACKWARD: Remove to remove the old splitter
+            if (mpuBucket.getMdBucketModelVersion() < 2) {
+                splitter = constants.oldSplitter;
+            }
+            // Reconstruct mpuOverviewKey
+            const mpuOverviewKey =
+                `overview${splitter}${objectKey}${splitter}${uploadId}`;
+
+            return metadata.getObjectMD(mpuBucket.getName(), mpuOverviewKey,
```
dora-korpar: Good idea!

lib/api/multipartDelete.js
dora-korpar: No, the location should always be defined here, because the check is for the controlling location constraint, which defaults to either the bucket location or request endpoint location if the object is put without one :)

lib/data/external/AwsClient.js
```diff
@@ -208,6 +208,23 @@ class AwsClient {
...
+        const awsBucket = this._awsBucketName;
+        const awsKey = this._createAwsKey(bucket, key, this._bucketMatch);
+        const abortParams = {
+            Bucket: awsBucket, Key: awsKey, UploadId: uploadId,
+        };
+        return this._client.abortMultipartUpload(abortParams, err => {
+            if (err) {
+                log.error('There was an error aborting the MPU on AWS S3. You' +
+                    ' should abort directly on AWS S3 using the same uploadId.',
+                { error: err, dataStoreName: this._dataStoreName });
```
LaurenSpiegel: does this error actually show up in the logs or do you have to do err.message and err. stack?

Issue #789, opened 2017-06-27
Multipart upload: Failure on non sequential part numbers


mikeys:
### Description

On multipart upload: the upload fails when using non sequential part numbers.

Based on [AWS S3 specification](http://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html):
_"Amazon S3 creates an object by concatenating the parts in ascending order based on the part number"_

That means that potentially you can have part numbers which are non sequential (1, 2, 3, 7) and as long as the Complete Multipart Upload request lists those parts, everything should still work correctly (the concatenation of the parts is made based on the part number ascending order).

Specifically (in `lib/api/completeMultipartUpload.js`):
```node
for (let i = 0; i < partLength; i++) {
    const part = jsonList.Part[i];
    const partNumber = Number.parseInt(part.PartNumber[0], 10);
    if (partNumber !== i + 1) {     
        // If not in order return InvalidPartOrder      
        if (partNumber <= partLength) {     
            return next(errors.InvalidPartOrder, destBucket);       
        }       
        // If missing part return InvalidPart       
        return next(errors.InvalidPart, destBucket);        
    }
```
The above seems incorrect since it requires the `partNumber` to go sequentially starting from 1.

### Steps to reproduce the issue

1. Simply initiate a multipart upload in one of your favorite AWS S3 SDKs
2. When using `upload.add_part`, assign larger but non sequential part numbers to sequential file parts (1: file_part1, 3: file_part2, 4: file_part3).
3. `InvalidPart` error is raised.

*Expected result*: 
Upload should be successful - concatenation of parts based on ascending part number order.

### Additional information: (Node.js version, Docker version, etc)
Node.js v6.9.5 without Docker.

### Possible fix
```node
const _partNumberFromPart = (part) => Number.parseInt(part.PartNumber[0], 10);
const _partNumberComparer = (a, b) => _partNumberFromPart(a) - _partNumberFromPart(b);

function parsePartsList(destBucket, objMD, mpuBucket, next) {
    if (request.post) {
        return parseXml(request.post, (err, jsonList) => {
            if (err) {
                return next(err, destBucket);
            }

             const parts = jsonList.Part;

            // "Amazon S3 creates an object by concatenating the parts in ascending
            // order based on the part number"
            jsonList.Part = parts.sort(_partNumberComparer);

            return next(err, destBucket, objMD, mpuBucket, jsonList);
        });
    }
    return next(errors.MalformedXML, destBucket);
}

[...]

PR #802, opened 2017-06-30
FIX: completeMPU

nicolas2bert:
FIXES(#789)

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -22,6 +22,8 @@ const logger = require('../utilities/logger');
 
 const versionIdUtils = versioning.VersionID;
 
+let splitter = constants.splitter;
+
 /*
    Format of xml request:
    <CompleteMultipartUpload>
@@ -159,7 +161,6 @@ function completeMultipartUpload(authInfo, request, log, callback) {
             return next(errors.MalformedXML, destBucket);
         },
         function getMPUMetadata(destBucket, objMD, mpuBucket, jsonList, next) {
-            let splitter = constants.splitter;
             // BACKWARD: Remove to remove the old splitter
             if (mpuBucket.getMdBucketModelVersion() < 2) {
                 splitter = constants.oldSplitter;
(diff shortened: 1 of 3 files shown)


lib/api/completeMultipartUpload.js
```diff
@@ -258,107 +259,121 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                // return an error
+                // in ascending order return an error
                 const partNumber = Number.parseInt(part.PartNumber[0], 10);
-                if (partNumber !== i + 1) {
-                    // If not in order return InvalidPartOrder
-                    if (partNumber <= partLength) {
+                if (i > 0) {
+                    const previousPartNumber =
+                      Number.parseInt(jsonList.Part[i - 1].PartNumber[0], 10);
+                    if (partNumber <= previousPartNumber) {
```
LaurenSpiegel: If you send a complete mpu call with an inventory that states a part number more than once, what error does aws send?

mikeys: @nicolas2bert I'm not 100% sure that there's a restriction on the parts order in the MPU Complete request.

E.g. 
```json
{
  "Part": [
    { "PartNumber": 8, "ETag": "..." },
    { "PartNumber": 5, "ETag": "..." },
    { "PartNumber": 13, "ETag": "..." }
  ]
}
```

May be legit. All that I could grasp from the AWS Doc is that the items are concatenated based on the part numbers.

nicolas2bert: Hi @mikeys, 

There is an actual meaning to the array order.

Tested against AWS:

```
{
  "Part": [
    { "PartNumber": 1, "ETag": "..." },
    { "PartNumber": 2, "ETag": "..." },
    { "PartNumber": 3, "ETag": "..." }
  ]
}
```
returns no error.
However, 
```
{
  "Part": [
    { "PartNumber": 1, "ETag": "..." },
    { "PartNumber": 3, "ETag": "..." },
    { "PartNumber": 2, "ETag": "..." }
  ]
}
```
returns 400 - InvalidPartOrder error

mikeys: @nicolas2bert You've checked it against the actual AWS S3 Service? cool. Thanks for getting into this.

nicolas2bert: yes I did @mikeys, sure 👍

lib/api/completeMultipartUpload.js
```diff
@@ -258,107 +266,127 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                // the ETag for the part already sent, return an error
-                if (partETag !== storedPartsAsObjects[i].ETag) {
-                    return next(errors.InvalidPart, destBucket);
-                }
-                // If any part other than the last part is less than 5MB,
-                // return an error
-                const partSize = Number.parseInt(storedPartsAsObjects[i]
-                    .size, 10);
+                let isPartUploaded = false;
+                while (storedPartsAsObjects.length > 0 && !isPartUploaded) {
```
electrachong: if you prefer, you can do:
```
let isPartUploaded = false
for ( ; storedPartsAsObjects.length > 0 && !isPartUploaded; storedPartsAsObjects = storedPartsAsObjects.splice(1));
```
Go with what you think is most readable!

lib/api/completeMultipartUpload.js
```diff
@@ -221,18 +222,25 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
             }
-            const storedPartsAsObjects = storedParts.map(item => ({
-                // In order to delete the part listing in the shadow
-                // bucket, need the full key
-                key: item.key,
-                ETag: `"${item.value.ETag}"`,
-                size: item.value.Size,
+            let storedPartsAsObjects = [];
+            const keysToDelete = [];
+            for (let i = 0; i < storedParts.length; i++) {
```
nicolas2bert: Yes forEach is probably more appropriate.
Interesting article I found: http://thejsguy.com/2016/07/30/javascript-for-loop-vs-array-foreach.html

lib/api/completeMultipartUpload.js
nicolas2bert: I think it is readable enough. If you don't think so, I can change it.

lib/api/completeMultipartUpload.js
```diff
@@ -258,107 +266,127 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                                cipheredDataKey: location.sseCipheredDataKey,
+                            };
+                            dataLocations.push(pieceRetrievalInfo);
+                            calculatedSize += pieceSize;
+                        }
+
+                        const partETagWithoutQuotes =
+                            storedPart.ETag.slice(1, -1);
+                        concatETags += partETagWithoutQuotes;
+                        // partUploadedCompleted.push(storedPart);
```
nicolas2bert: nope, thanks

nicolas2bert:
@LaurenSpiegel @electrachong I adressed your comments here :)

PR #804, opened 2017-07-05
DOCKER: SSL for host and path style

nicolas2bert:
Using Subject Alternative Names (SAN) to specify both host and path style host names.

### DIFF
--- a/docs/DOCKER.rst
+++ b/docs/DOCKER.rst
@@ -106,13 +106,11 @@ will give you the most detailed.
 SSL
 ^^^
 
-This variable specifies the Common Name ``<DOMAIN_NAME>`` used to create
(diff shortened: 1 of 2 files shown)


docs/DOCKER.rst
```diff
@@ -58,16 +58,16 @@ You will also have to mount your AWS credentials file:
     -v $(pwd)/locationConfig.json:/usr/src/app/locationConfig.json
     -v ~/.aws/credentials:/root/.aws/credentials -e S3DATA=multiple scality/s3server
 
-HOST\_NAME
+ENDPOINT or HOST_NAME
```
LaurenSpiegel: just endpoint

LaurenSpiegel:
@nicolas2bert, can we close this?

PR #805, opened 2017-07-05
fix: missing mpu bucket prefix

electrachong:
The constant was accidentally deleted when extracting routes to Arsenal, causing the mpu bucket prefix to default to "undefined". Thanks to @bennettbuchanan for noticing the bug.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -32,6 +32,12 @@ const constants = {
     oldSplitter: 'splitterfornow',
     usersBucket: 'users..bucket',
     oldUsersBucket: 'namespaceusersbucket',
+    // MPU Bucket Prefix is used to create the name of the shadow
+    // bucket used for multipart uploads.  There is one shadow mpu
+    // bucket per bucket and its name is the mpuBucketPrefix followed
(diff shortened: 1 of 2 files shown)


electrachong:
@ironman-machine

rahulreddy:
I think something critical like this a test is warranted. Do you think you can add a unit test asserting that?

electrachong:
@rahulreddy it could be as simple as a unit test asserting the constant is defined, is that fine? All our existing unit tests pass because they get the mpu bucket using constants.mpuBucketPrefix as well.

rahulreddy:
Yes that and you can modify one or add a new unit test where you can check the metadata backend has the bucket created with constants. mpuBucketPrefix (not sure if we are already checking that)

electrachong:
@rahulreddy updated. We do have an existing unit test that checks if the mpu bucket was created in in-mem metadata using the constant.

electrachong:
@ironman-machine

electrachong:
end-to-end passed

PR #807, opened 2017-07-06
Rf/extract conditional header validation

electrachong:
Dependent on https://github.com/scality/Arsenal/pull/295


electrachong:
end to end here: http://ci.ironmann.io/gh/scality/Integration/13996

electrachong:
end-to-end passed

PR #808, opened 2017-07-06
Fix credentials profile assignment

dora-korpar:
If using a credential profile with AWS instead of directly using accessKey/secretKey, the credentials have to be assigned to the global AWS config, not passed as part of the S3 config params


dora-korpar:
@ironman-machine go

PR #809, opened 2017-07-06
Rf/retrieve data params

electrachong:
Depends on https://github.com/scality/Arsenal/pull/294

### DIFF
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -96,7 +96,7 @@ const data = {
             });
     },
 
-    get: (objectGetInfo, log, cb) => {
+    get: (objectGetInfo, log, params, cb) => {
         // If objectGetInfo.key exists the md-model-version is 2 or greater.
         // Otherwise, the objectGetInfo is just the key string.
         const objGetInfo = (implName === 'sproxyd') ?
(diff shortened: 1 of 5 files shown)


lib/data/wrapper.js
```diff
@@ -96,7 +96,7 @@ const data = {
             });
     },
 
-    get: (objectGetInfo, log, cb) => {
+    get: (objectGetInfo, log, params, cb) => {
```
LaurenSpiegel: what are the possible params? where are they used?

electrachong: it isn't used in s3 at all, but this function has to have the same prototype as a function in Mystique which needs to pass extra params, so that responseStreamData in Arsenal can call either as the 'dataRetrievalFn' without differentiating between the originating project.

jonathan-gramain: I haven't digged into the topic, but on top of head, couldn't those params be part of `objectGetInfo`? Mystique wrapper would extract the relevant params from it and pass them as separate params info then to the Mystique front-end. Unless they can conflict with existing params from `objectGetInfo`, merging them could make the API semantic slightly simpler and avoid the change.

electrachong: Oh. A much better idea, and I can get rid of this pr and the Arsenal branch entirely :wink:

Thanks!

PR #810, opened 2017-07-06
FT: Support completeMPU, objectCopy replication

bennettbuchanan:
Update object MD replicationInfo field when using completeMPU and objectCopy APIs and replication applies to the object.


bennettbuchanan:
@ironman-machine try

lib/api/completeMultipartUpload.js
```diff
@@ -433,6 +434,11 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                     metaStoreParams.versioning = options.versioning;
                     metaStoreParams.isNull = options.isNull;
                     metaStoreParams.nullVersionId = options.nullVersionId;
+                    const replicationInfo = getReplicationInfo(objectKey,
```
jonathan-gramain: Since this new processing is not related to versioning, could we rename the `versioningPreprocessing` function to some more generic name?

tests/unit/api/objectReplicationMD.js
```diff
@@ -114,6 +122,92 @@ function createBucketWithReplication(hasStorageClass) {
     });
 }
 
+// Initiate an MPU, put a part with the given body, and complete the MPU.
+function putMPU(key, body, cb) {
+    return async.waterfall([
```
rahulreddy: Unit tests have to use dummy data, otherwise these might as well be a functional tests. You can setup the mpu bucket and part data in `metadata.keyMaps` and when you do completeMultipartUpload it would "just work".

lib/api/completeMultipartUpload.js
bennettbuchanan: Thank you. I think it would be appropriate to instead move this to the previous waterfall function.

tests/unit/api/objectReplicationMD.js
bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine try

tests/unit/api/objectReplicationMD.js
```diff
@@ -27,6 +24,7 @@ const authInfo = makeAuthInfo(canonicalID);
 const ownerID = authInfo.getCanonicalID();
 const namespace = 'default';
 const bucketName = 'source-bucket';
+const mpuShadowBucket = `${constants.mpuBucketPrefix + bucketName}`;
```
rahulreddy: You mean 
```js
const mpuShadowBucket = `${constants.mpuBucketPrefix}${bucketName}`;
```

bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed on rebuild: http://ci.ironmann.io/gh/scality/Integration/13960.

lib/api/completeMultipartUpload.js
```diff
@@ -406,6 +407,11 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 multipart: true,
                 log,
             };
+            const replicationInfo = getReplicationInfo(objectKey, destBucket,
```
rahulreddy: since `getReplicationInfo` returns `undefined` if there is no replication config, you can avoid the conditional and set
```js
const metaStoreParams = {
     authInfo,
     objectKey,
     replicationInfo: getReplicationInfo(objectKey, destBucket, false, calculatedSize),
     ...
}
```

lib/api/completeMultipartUpload.js
```diff
@@ -5,6 +5,7 @@ const { errors, versioning, s3middleware } = require('arsenal');
 
 const escapeForXml = s3middleware.escapeForXml;
 const { pushMetric } = require('../utapi/utilities');
+const getReplicationInfo = require('./apiUtils/object/replication');
```
rahulreddy: I realized that the filename does not reflect the module name. It should be
```js
const getReplicationInfo = require('./apiUtils/object/getReplicationInfo');
```

lib/api/completeMultipartUpload.js
bennettbuchanan: I will update this. I had considered that, but thought it would be better to not assign a property as `undefined`.

lib/api/completeMultipartUpload.js
bennettbuchanan: Thank you for spotting this. Updated.

bennettbuchanan:
@ironman-machine try

lib/api/completeMultipartUpload.js
rahulreddy: `undefined` doesn't show up when you stringify the object, so that's good.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed.

PR #812, opened 2017-07-07
REFACTOR: do not require auth file if env variables are set

LaureVergeron:
Small suggestion to avoid requiring the
conf/authdata.json file if custom credentials have
been set as container environment variables.


LaureVergeron:
Rebased and checks passed. Thanks @LaurenSpiegel

PR #813, opened 2017-07-07
Fix AWS creds in testing

dora-korpar:
Allows multiple backend tests to be run with credential profile specified in locationConfig, instead of defaulting to 'default'


dora-korpar:
@ironman-machine go

PR #814, opened 2017-07-08
FT: Add configurable replication storageClass to objectMD

bennettbuchanan:
Allow including a custom array of replicationEndpoints in the `config.json`, any of which would be considered valid when putting the replicationConfiguration.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -221,6 +221,37 @@ class Config {
             this.replicationGroupId = 'RG001';
         }
 
+        this.replicationEndpoints = [];
+        if (config.replicationEndpoints) {
+            const { replicationEndpoints } = config;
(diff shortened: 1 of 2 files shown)


lib/Config.js
```diff
@@ -221,6 +221,17 @@ class Config {
             this.replicationGroupId = 'RG001';
         }
 
+        this.replicationEndpoints = [];
```
jonathan-gramain: One day not too far away we should consider using `joi` to validate S3 config structure and types, like is done in backbeat (thanks @rahulreddy for the idea), this would make validation much more concise and less error prone. But it's ok for now of course.

bennettbuchanan:
@ironman-machine try

PR #815, opened 2017-07-10
backport to rel/7.0: missing mpu bucket prefix [S3C-632]

electrachong:
(cherry picked from commit 2fb0b6855a541dfa3242ea2515ec5e1d7fe7cddf)

I think we should backport this fix to rel/7.0 since the release was affected by this bug as well.


electrachong:
@ironman-machine

electrachong:
end-to-end passed (I think regular ci tests are failing because they are not run with rel/7.0 arsenal branch)

rahulreddy:
Wasn't the bug introduced when the routes were extracted in Arsenal which was after the 7.0 release? If yes, why do we need this to be backported?

electrachong:
I think it was merged before the 7.0 release / branch was created:
https://github.com/scality/S3/commits/rel/7.0

electrachong:
@ironman-machine after locking dependency pakages

electrachong:
end to end passed, altho regular ci is still failing (I guess this time because of cache..)

PR #816, opened 2017-07-11
Add tests for AWS object copy

dora-korpar:
Adds tests for objectCopy API with AWS backend. ~~Nothing had to be added to the API itself, but~~ a minor change was included because it directly impacted the testing. This is adding an error return on a GET call to AWS.
New features include 0-byte object support and copying objects across backends while retaining all other metadata.

### DIFF
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -4,6 +4,7 @@ const { errors, versioning, s3middleware } = require('arsenal');
 const getMetaHeaders = s3middleware.userMetadata.getMetaHeaders;
 const validateHeaders = s3middleware.validateConditionalHeaders;
 
+const constants = require('../../constants');
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const locationConstraintCheck
     = require('./apiUtils/object/locationConstraintCheck');
@@ -19,8 +20,10 @@ const removeAWSChunked = require('./apiUtils/object/removeAWSChunked');
 const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
 const validateWebsiteHeader = require('./apiUtils/object/websiteServing')
     .validateWebsiteHeader;
+const { config } = require('../Config');
 
 const versionIdUtils = versioning.VersionID;
+const locationHeader = constants.objectLocationConstraintHeader;
 
 /**
  * Preps metadata to be saved (based on copy or replace request header)
@@ -84,6 +87,14 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         });
         return { error: userMetadata };
     }
+    // If metadataDirective is 'COPY' and location constraint header is
+    // specified, new location constraint should override source object's
+    if (whichMetadata === 'COPY' && headers[locationHeader]) {
+        userMetadata[locationHeader] = headers[locationHeader];
+    }
+    // If location constraint header is not included, locations match
+    const locationMatch = headers[locationHeader] ?
+        sourceObjMD[locationHeader] === headers[locationHeader] : true;
 
     // If tagging directive is REPLACE but you don't specify any
     // tags in the request, the destination object will
@@ -123,6 +134,7 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         taggingCopy,
         replicationInfo: getReplicationInfo(objectKey, destBucketMD, false,
             sourceObjMD['content-length']),
+        locationMatch,
     };
 
     // In case whichMetadata === 'REPLACE' but contentType is undefined in copy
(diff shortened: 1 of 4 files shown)


lib/data/external/AwsClient.js
```diff
@@ -32,7 +32,7 @@ class AwsClient {
             ContentLength: size,
         };
         if (keyContext.cipherBundle) {
-            uploadParams.ServerSideEncryption = 'aws:kms';
+            uploadParams.ServerSideEncryption = 'AES256';
```
LaurenSpiegel: I agree with this change but why is it in this pr?

lib/data/external/AwsClient.js
```diff
@@ -62,6 +62,7 @@ class AwsClient {
         const stream = request.createReadStream().on('error', err => {
             log.error('error streaming data from AWS', { error: err,
                 dataStoreName: this._dataStoreName });
+            return callback(err);
```
LaurenSpiegel: why is this added? didn't we discuss that this callback is confusing since it is not actually called?  once the listener is put on the stream, the headers have already been sent back to the requesting client and we break the connection if there is an error.

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy.js
```diff
@@ -0,0 +1,178 @@
...
+            .catch(err => {
+                process.stdout.write(`Error in afterEach: ${err}\n`);
+                throw err;
+            });
+        });
+
+        it('should copy an object from mem to AWS', done => {
+            putSourceObj('mem', () => {
+                const copyParams = Object.assign({
+                    MetadataDirective: 'REPLACE',
```
LaurenSpiegel: can we test that the metadata has been properly copied?

tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -51,7 +51,7 @@ function awsGetCheck(objectKey, s3MD5, awsMD5, cb) {
             if (process.env.ENABLE_KMS_ENCRYPTION === 'true') {
                 // doesn't check ETag because it's different
                 // with every PUT with encryption
-                assert.strictEqual(res.ServerSideEncryption, 'aws:kms');
+                assert.strictEqual(res.ServerSideEncryption, 'AES256');
```
LaurenSpiegel: separate pr?

lib/data/external/AwsClient.js
dora-korpar: In objectCopy, data.get is called directly, so the get error isn't caught and the API continues with the data.put, which makes the server hangs indefinitely

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine go2

lib/data/external/AwsClient.js
LaurenSpiegel: ah, makes sense

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy.js
LaurenSpiegel: Thanks. As discussed, if someone is copying from location constraint x to location constraint y and the source object metadata has a location header (scal-location-constraint) and they have selected the "COPY" option for metadata, we should copy all of the metadata except for the location header (scal-location-constraint) which we should have as the actual destination location.

dora-korpar:
@ironman-machine go3

lib/api/objectCopy.js
```diff
@@ -84,6 +84,12 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         });
         return { error: userMetadata };
     }
+    // If metadataDirective is 'COPY' and location constraint header is
+    // specified, new location constraint should override source object's
+    const locationHeader = 'x-amz-meta-scal-location-constraint';
```
LaurenSpiegel: this is in the constants file already so can pull it from there

lib/api/objectCopy.js
```diff
@@ -280,7 +286,9 @@ function objectCopy(authInfo, request, sourceBucket,
             // skip if source and dest the same or 0 byte object
             // still send along serverSideEncryption info so algo
             // and masterKeyId stored properly in metadata
-            if (sourceIsDestination || dataLocator.length === 0) {
+            if ((sourceIsDestination && !storeMetadataParams.metaHeaders
+            ['x-amz-meta-scal-location-constraint'])
```
LaurenSpiegel: use the constant here

LaurenSpiegel: also, isn't just checking whether this header is there not enough? shouldn't the check be that the specified location (if specified) is different from the data location of the source?  They could specify the location header but specify it to the same exact place.

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy.js
```diff
@@ -0,0 +1,206 @@
...
+const { createEncryptedBucketPromise } =
+    require('../../lib/utility/createEncryptedBucket');
+
+const awsLocation = 'aws-test';
+const bucket = 'buckettestmultiplebackendobjectcopy';
+const key = `somekey-${Date.now()}`;
+const copyKey = `copyKey-${Date.now()}`;
+const body = Buffer.from('I am a body', 'utf8');
+const correctMD5 = 'be747eb4b75517bf6b3cf7c5fbb62f3a';
+const locMetaHeader = 'scal-location-constraint';
```
LaurenSpiegel: use the constant with the x-amz-meta sliced off

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine gogo

PR #817, opened 2017-07-11
ListParts for AWS backend

dora-korpar:
Implements the listParts API for AWS backend

### DIFF
--- a/lib/api/listParts.js
+++ b/lib/api/listParts.js
@@ -10,6 +10,8 @@ const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
 const escapeForXml = s3middleware.escapeForXml;
 const { pushMetric } = require('../utapi/utilities');
 
+const { config } = require('../../lib/Config');
+const multipleBackendGateway = require('../data/multipleBackendGateway');
 
   /*
   Format of xml response:
@@ -129,14 +131,39 @@ function listParts(authInfo, request, log, callback) {
         function waterfall2(destBucket, next) {
             metadataValMPUparams.log = log;
             services.metadataValidateMultipart(metadataValMPUparams,
-                (err, mpuBucket, mpuOverview) => {
+                (err, mpuBucket, mpuOverviewObj) => {
                     if (err) {
                         return next(err, destBucket, null);
                     }
-                    return next(null, destBucket, mpuBucket, mpuOverview);
+                    return next(null, destBucket, mpuBucket, mpuOverviewObj);
                 });
         },
-        function waterfall3(destBucket, mpuBucket, mpuOverviewArray, next) {
+        function waterfall3(destBucket, mpuBucket, mpuOverviewObj, next) {
+            if (config.backends.data === 'multiple') {
+                const location = mpuOverviewObj.controllingLocationConstraint;
+                return multipleBackendGateway.listParts(objectKey, uploadId,
+                location, destBucket, partNumberMarker, maxParts, log,
+                (err, backendPartList) => {
+                    if (err) {
+                        return next(err, destBucket);
+                    } else if (backendPartList) {
+                        return next(null, destBucket, mpuBucket,
+                            mpuOverviewObj, backendPartList);
+                    }
+                    return next(null, destBucket, mpuBucket, mpuOverviewObj,
+                        null);
+                });
+            }
+            return next(null, destBucket, mpuBucket, mpuOverviewObj, null);
+        },
+        function waterfall4(destBucket, mpuBucket, mpuOverviewObj,
+        backendPartList, next) {
+            // if parts were returned from cloud backend, they were not
+            // stored in Scality S3 metadata, so this step can be skipped
+            if (backendPartList) {
+                return next(null, destBucket, mpuBucket, backendPartList,
+                    mpuOverviewObj);
+            }
             // BACKWARD: Remove to remove the old splitter
             if (mpuBucket.getMdBucketModelVersion() < 2) {
                 splitter = constants.oldSplitter;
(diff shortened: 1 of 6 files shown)


dora-korpar:
@ironman-machine go

lib/api/listParts.js
```diff
@@ -136,11 +139,50 @@ function listParts(authInfo, request, log, callback) {
...
                 splitter = constants.oldSplitter;
             }
+
+            if (config.backends.data === 'multiple') {
+                // Reconstruct mpuOverviewKey to serve
+                // as key to pull metadata originally stored when mpu initiated
+                const mpuOverviewKey =
+                    `overview${splitter}${objectKey}${splitter}${uploadId}`;
+
+                return metadata.getObjectMD(mpuBucket.getName(), mpuOverviewKey,
```
LaurenSpiegel: if you refactor services.metadataValidateMultipart to be a get of the mpuOverviewKey instead of a listing you can avoid this additional call.  It was a listing for reasons that no longer apply (the key used to contain more information).  That function would be much clearer if it was just a get and then you would have everything you need without an extra call to metadata.

lib/api/listParts.js
```diff
@@ -167,9 +210,16 @@ function listParts(authInfo, request, log, callback) {
                 // - {uploadId}
                 // - {splitter}
                 // - {partNumber}
-                const index = item.key.lastIndexOf(splitter);
+                let partNumber;
+                if (item.key) {
+                    const index = item.key.lastIndexOf(splitter);
+                    partNumber =
+                        parseInt(item.key.substring(index + splitterLen), 10);
+                } else {
```
LaurenSpiegel: please add note that otherwise the listing came from real AWS so has partNumber as attribute already

lib/data/external/AwsClient.js
```diff
@@ -158,6 +158,39 @@ class AwsClient {
...
+    listParts(key, uploadId, bucket, partNumberMarker, maxParts, log,
+    callback) {
+        const awsBucket = this._awsBucketName;
+        const awsKey = this._createAwsKey(bucket, key, this._bucketMatch);
+        const params = { Bucket: awsBucket, Key: awsKey, UploadId: uploadId,
+            PartNumberMarker: partNumberMarker, MaxParts: maxParts };
+        return this._client.listParts(params, (err, partList) => {
+            if (err) {
+                log.error('err from data backend on listPart',
+                { error: err.message, dataStoreName: this._dataStoreName });
```
LaurenSpiegel: also log the err.stack

lib/data/external/AwsClient.js
```diff
@@ -158,6 +158,39 @@ class AwsClient {
...
+                return callback(errors.InternalError);
+            }
+            // build storedParts object to mimic Scality S3 backend returns
+            const storedParts = {};
+            storedParts.IsTruncated = partList.IsTruncated;
+            storedParts.Contents = [];
+            storedParts.Contents = partList.Parts.map(item => {
```
LaurenSpiegel: It would be better if we didn't have to loop through these items twice but I suppose this is less messy code-wise.  We could have the create xml function have an if clause where if it's from aws, the xml creation is different so there is one pass through. What do you think?

tests/multipleBackend/multipartUpload.js
```diff
@@ -137,6 +144,57 @@ function assertMpuCompleteResults(compResult) {
...
 
+function assertListResults(listResult, testName, uploadId) {
+    parseString(listResult, (err, json) => {
+        assert.equal(err, null, `Error parsing list part results: ${err}`);
+        assert.strictEqual(json.ListPartsResult.Key[0], objectKey);
+        assert.strictEqual(json.ListPartsResult.UploadId[0], uploadId);
+        assert.strictEqual(json.ListPartsResult.Initiator[0].ID[0],
+            authInfo.getCanonicalID());
+
+        if (testName === 'partNumMarker') {
```
LaurenSpiegel: a comment about what we are testing here would be helpful

tests/multipleBackend/multipartUpload.js
```diff
@@ -137,6 +144,57 @@ function assertMpuCompleteResults(compResult) {
...
+        } else {
+            assert.strictEqual(json.ListPartsResult.PartNumberMarker,
+                undefined);
+            assert.strictEqual(json.ListPartsResult.Part[0].PartNumber[0], '1');
+            assert.strictEqual(json.ListPartsResult.Part[0].ETag[0],
+                `"${awsETagBigObj}"`);
+            assert.strictEqual(json.ListPartsResult.Part[0].Size[0],
+                '10485760');
+
+            if (testName === 'maxParts') {
```
LaurenSpiegel: here as well

tests/multipleBackend/multipartUpload.js
```diff
@@ -247,11 +313,58 @@ describe('Multipart Upload API with AWS Backend', function mpuTestSuite() {
...
+                query: { uploadId } }, listRequest);
+            listParts(authInfo, listParams, log, (err, result) => {
+                assert.equal(err, null, `Error listing parts on AWS: ${err}`);
+                assertListResults(result, null, uploadId);
+                abortMPU(uploadId, done);
+            });
+        });
+    });
+
+    it('should list maxParts number of parts', done => {
```
LaurenSpiegel: should only return number of parts equal to specified maxParts

lib/data/external/AwsClient.js
dora-korpar: Discussed, and we'll leave as is for now with later opportunity for optimization.

lib/services.js
```diff
@@ -410,57 +410,35 @@ const services = {
...
-                    const index2 = key.lastIndexOf(splitter);
+                    const initiatorID = storedMetadata.initiator.ID;
+                    const ownerID = storedMetadata['owner-id'];
                     const mpuOverview = [
-                        key.substring(0, index),
-                        key.substring(index + splitterLen, index2),
-                        key.substring(index2 + splitterLen),
-                        eventualStorageBucket,
+                        // 'overview' left as placeholder in array for
+                        // backwards compatibility
```
LaurenSpiegel: If we can change this into an object from an array without too much trouble (too many code changes in other api's), we should do so.

lib/services.js
```diff
@@ -410,57 +410,35 @@ const services = {
...
-                        initiatorDisplayName,
+                        storedMetadata.initiator.DisplayName,
                         ownerID,
-                        ownerDisplayName,
-                        storageClass,
-                        initiated,
+                        storedMetadata['owner-display-name'],
+                        storedMetadata['x-amz-storage-class'],
+                        storedMetadata.initiated,
+                        storedMetadata.controllingLocationConstraint,
```
LaurenSpiegel: don't we have getters for the object model attributes now?

dora-korpar:
@ironman-machine go

PR #818, opened 2017-07-11
Remove changed global vars

dora-korpar:
I had to learn the hard way not to cheat with changing global vars. This PR moves any changing variables in the multipleBackend MPU test suite (the only offender) to the local scope. Should also help with some of the AWS socket hangups


dora-korpar:
@ironman-machine go

PR #819, opened 2017-07-12
DOCKER: adding ENDPOINT and secret credentials

nicolas2bert:
This PR adds:
- ENDPOINT environment variable ( ENDPOINT === HOST_NAME)
- secret credentials for Zenko

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,6 +6,10 @@ set -e
 # modifying config.json
 JQ_FILTERS_CONFIG="."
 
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -6,6 +6,10 @@ set -e
 # modifying config.json
 JQ_FILTERS_CONFIG="."
 
+if [[ "$ENDPOINT" ]]; then
+    HOST_NAME= "$ENDPOINT"
```
RachedBenMustapha: I think we should remove the extra space

PR #822, opened 2017-07-14
[backbeat] bf: sanity check on header presence

jonathan-gramain:
ensure x-scal-canonical-id and content-md5 headers are provided in the `PUT .../data` backbeat route.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -58,8 +58,16 @@ PUT /_/backbeat/<bucket name>/<object key>/data
 */
 
 function putData(request, response, bucketInfo, objMd, log, callback) {
-    const canonicalID = request.headers['x-scal-canonicalId'];
-    const contentMd5 = request.headers['content-md5'];
+    const canonicalID = request.headers['x-scal-canonical-id'];
+    if (canonicalID === undefined) {
+        log.error('bad request: missing x-scal-canonical-id header');
+        return callback(errors.BadRequest);
+    }
+    const contentMD5 = request.headers['content-md5'];
+    if (contentMD5 === undefined) {
+        log.error('bad request: missing content-md5 header');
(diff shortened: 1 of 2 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -58,8 +58,16 @@ PUT /_/backbeat/<bucket name>/<object key>/data
 */
 
 function putData(request, response, bucketInfo, objMd, log, callback) {
-    const canonicalID = request.headers['x-scal-canonicalId'];
-    const contentMd5 = request.headers['content-md5'];
+    const canonicalID = request.headers['x-scal-canonical-id'];
+    if (canonicalID === undefined) {
+        log.error('bad request: missing x-scal-canonical-id header');
```
LaurenSpiegel: isn't this more like a user error which should be logged at debug level?

lib/routes/routeBackbeat.js
```diff
@@ -58,8 +58,16 @@ PUT /_/backbeat/<bucket name>/<object key>/data
...
+        return callback(errors.BadRequest);
+    }
+    const contentMD5 = request.headers['content-md5'];
+    if (contentMD5 === undefined) {
+        log.error('bad request: missing content-md5 header');
```
LaurenSpiegel: same

lib/routes/routeBackbeat.js
ploki: maybe it's a backbeat specific route and such cannot be considered coming from a user?

rahulreddy: Yes, canonicalID is critical for sending request to sproxyd which BB should be sending always, so makes sense to log it as error.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
@ironman-machine try

PR #826, opened 2017-07-18
FIX: change AWS encryption type

dora-korpar:
Changes encryption type used on AWS backend to 'AES256' to match Scality S3 encryption type.


tests/functional/aws-node-sdk/test/multipleBackend/put.js
```diff
@@ -51,7 +51,7 @@ function awsGetCheck(objectKey, s3MD5, awsMD5, cb) {
             if (process.env.ENABLE_KMS_ENCRYPTION === 'true') {
                 // doesn't check ETag because it's different
```
ploki: Just curious, you mean the returned ETag doesn't match the MD5 sum of the actual object?

dora-korpar: Not if encryption is enabled

ploki: thanks

PR #829, opened 2017-07-20
FT: AWS TAGGING

dora-korpar:
Implement object tagging for AWS backend.

Depends on Arsenal#310

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -83,6 +83,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         namespace: request.namespace,
         objectKey,
         metaHeaders,
+        tagging: request.headers['x-amz-tagging'],
     };
     // If the request was made with a pre-signed url, the x-amz-acl 'header'
     // might be in the query string rather than the actual headers so include
@@ -122,6 +123,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
     const backendInfo = backendInfoObj.backendInfo;
     const location = backendInfo.getControllingLocationConstraint();
     const locationType = config.locationConstraints[location].type;
+    metadataStoreParams.dataStoreName = location;
     /* eslint-disable camelcase */
     const dontSkipBackend = externalBackends;
     /* eslint-enable camelcase */
--- a/lib/data/multipleBackendGateway.js
+++ b/lib/data/multipleBackendGateway.js
@@ -199,6 +199,21 @@ const multipleBackendGateway = {
         }
         return cb();
     },
+
+    objectTagging: (method, key, bucket, objectMD, log, cb) => {
+        const location = objectMD.dataStoreName;
+        const client = clients[location];
+        if (client[`object${method}Tagging`]) {
+            return client[`object${method}Tagging`](key, bucket, objectMD, log,
+            err => {
+                if (err) {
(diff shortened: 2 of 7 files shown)


lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -83,6 +83,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         namespace: request.namespace,
         objectKey,
         metaHeaders,
+        tagging: request.headers['x-amz-tagging'],
```
LaurenSpiegel: are these validated before getting here?

lib/data/multipleBackendGateway.js
```diff
@@ -199,6 +199,21 @@ const multipleBackendGateway = {
         }
         return cb();
     },
+
+    objectTagging: (method, key, bucket, objectMD, log, cb) => {
+        const location = objectMD.location[0].dataStoreName;
```
LaurenSpiegel: this won't work for 0 byte objects.  we need to know from the metadata of the object, what the data backend is.

tests/functional/aws-node-sdk/test/multipleBackend/objectTagging.js
```diff
@@ -0,0 +1,221 @@
...
+            });
+        });
+
+        describe('putObject with tags and putObjectTagging', () => {
+            testBackends.forEach(backend => {
+                it(`should put an object with tags to ${backend} backend`,
+                done => {
+                    const params = Object.assign({ Tagging: tagString, Metadata:
+                        { 'scal-location-constraint': backend } }, putParams);
+                    s3.putObject(params, err => {
```
LaurenSpiegel: I assume this is not working for the aws backend since we are not pulling the dataStore properly on 0 byte objects? We should test with both 0 byte objects and objects with some data to make sure we are testing both code paths.

lib/api/apiUtils/object/createAndStoreObject.js
dora-korpar: No, it doesn't look like they're validated anywhere. @nicolas2bert ?

nicolas2bert: `request.headers['x-amz-tagging']` is validate later when saving MD.

dora-korpar:
@ironman-machine go

tests/functional/aws-node-sdk/test/multipleBackend/objectTagging.js
```diff
@@ -0,0 +1,264 @@
...
+                const params = Object.assign({ Metadata:
+                    { 'scal-location-constraint': awsLocation } }, putParams);
+                s3.putObject(params, err => {
+                    assert.equal(err, null);
+                    awsS3.deleteObject(awsDelParams, err => {
+                        assert.equal(err, null);
+                        const putTagParams = { Bucket: bucket, Key: key,
+                            Tagging: putTags };
+                        s3.putObjectTagging(putTagParams, err => {
+                            assert.strictEqual(err.code, 'InternalError');
```
jonathan-gramain: `InternalError` sounds a bit too generic and scary for this case, and it does not give back any useful info to the user, maybe simply ObjNotFound would be more appropriate? Or maybe we need to create new specific errors when there are inconsistencies caused by the user accessing AWS directly. I understand that we're talking about the object being removed from AWS but not from scality S3 backend.

jonathan-gramain:
oh, just noticed it was merged half an hour ago ^^

PR #842, opened 2017-07-28
FT: add AWS/Azure customize error description

nicolas2bert:
This PR includes for all of the Azure and AWS errors a "customizeDescription" so the user will instantly see the AWS/Azure error without having to look in the logs.

### DIFF
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -3,6 +3,7 @@ const AWS = require('aws-sdk');
 const MD5Sum = s3middleware.MD5Sum;
 const createLogger = require('../multipleBackendLogger');
 const { prepareStream } = require('../../api/apiUtils/object/prepareStream');
+const logHelper = require('./utils').logHelper;
 
 class AwsClient {
     constructor(config) {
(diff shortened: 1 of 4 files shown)


lib/data/external/AwsClient.js
```diff
@@ -43,7 +43,10 @@ class AwsClient {
                     const log = createLogger(reqUids);
                     log.error('err from data backend',
                     { error: err, dataStoreName: this._dataStoreName });
```
jonathan-gramain: While we are at error management, we may want to log `err.message` instead, here and below, as @rahulreddy told me this is what elastic search expects, and also error messages are often generated so they will not appear when logging the error object.

jonathan-gramain: And it may be a good time to create a logging helper since there are a lot of similar such calls in this file.

nicolas2bert:
@ironman-machine ok

nicolas2bert:
@jonathan-gramain I addressed your comments :)

nicolas2bert:
@ironman-machine ok

jonathan-gramain:
@nicolas2bert Thank you!

lib/data/external/utils.js
```diff
@@ -0,0 +1,8 @@
+const utils = {
+    logHelper(log, level, description, error, dataStoreName) {
```
jonathan-gramain: Being nit-picky, I was rather thinking of a local helper that would embed the `error from AWS` message and specifics, so that it has the same format everywhere. I don't think this one has much value especially since you can't log specific objects like the normal log function. But if you don't have more time to spend on this it's alright, the essential is that the error is logged ;)

nicolas2bert:
@ironman-machine ok

nicolas2bert:
@ironman-machin try

nicolas2bert:
@ironman-machine try

PR #844, opened 2017-07-31
COMPAT: Set part size as content-length

bennettbuchanan:
For a `getObject` request with a `PartNumber` param specified, we want to return the part's size as content length, and not the entire object's size.


bennettbuchanan:
@ironman-machine try

PR #847, opened 2017-08-01
FIX: Get obj part when originally copied from MPU

bennettbuchanan:
When getting an object's part that was originally copied from an MPU, we need to dynamically search the `objMD.location` array to retrieve all locations that comprise the object part. 

We then use those locations to calculate the byte range which defines the part, return those bytes as the body and set the body size as the content-length.

✅  Tested with `AWS_ON_AIR=true`.

### DIFF
--- a/lib/api/objectGet.js
+++ b/lib/api/objectGet.js
@@ -154,15 +154,35 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
                 }
             }
             if (partNumber) {
-                const partObj = objMD.location[partNumber - 1];
-                if (partObj === undefined) {
+                const locations = [];
+                for (let i = 0; i < objMD.location.length; i++) {
+                    const { dataStoreETag } = objMD.location[i];
+                    // Location objects prior to GA7.1 do not include the
+                    // dataStoreETag field so we cannot find the part range
+                    if (dataStoreETag === undefined) {
+                        return callback(errors.NotImplemented
+                            .customizeDescription('PartNumber parameter for ' +
+                            'this object is not supported'));
+                    }
+                    const locationPartNumber =
+                        Number(dataStoreETag.slice(':')[0], 10);
+                    // Get all parts that belong to the requested part number
+                    if (partNumber === locationPartNumber) {
+                        locations.push(objMD.location[i]);
+                    } else if (locationPartNumber > partNumber) {
+                        break;
+                    }
+                }
+                if (locations.length === 0) {
                     return callback(errors.InvalidPartNumber, null,
                         corsHeaders);
                 }
-                const { start, size } = partObj;
-                responseMetaHeaders['Content-Length'] = size;
-                const partRange = [start, start + size - 1];
-                dataLocator = setPartRanges(dataLocator, partRange);
+                const { start } = locations[0];
+                const endLocation = locations[locations.length - 1];
+                const end = endLocation.start + endLocation.size;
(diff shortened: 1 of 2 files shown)


lib/api/objectGet.js
```diff
@@ -154,14 +154,18 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
                 }
             }
             if (partNumber) {
-                const partObj = objMD.location[partNumber - 1];
-                if (partObj === undefined) {
+                // Get all parts that belong to the requested part number
+                const locationParts = objMD.location.filter(location =>
+                    location.dataStoreETag.startsWith(`${partNumber}`));
```
ploki: you shoud consider adding a trailing colon in the string to match. e.g.
```location.dataStoreETag.startsWith(`${partNumber}:`));```

In addition, this must be compatible with object with no ETag in the location array

Comment (bennettbuchanan) [inline lib/api/objectGet.js]: Nice catch. Thanks!

**Comment by bennettbuchanan** [comment]
@ironman-machine try

File: lib/api/objectGet.js [hunk 2]
```diff
@@ -154,15 +154,31 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
...
             if (partNumber) {
-                const partObj = objMD.location[partNumber - 1];
-                if (partObj === undefined) {
+                const locations = [];
+                for (let i = 0; i < objMD.location.length; i++) {
+                    const location = objMD.location[i];
+                    // Location objects prior to GA7.1 do not include the
+                    // dataStoreETag field so we cannot find the part range
+                    if (location.dataStoreETag === undefined) {
+                        return callback(errors.InvalidObjectState
```
LaurenSpiegel: I would do errors.NotImplemented

lib/api/objectGet.js
```diff
@@ -154,15 +154,31 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
...
+                            .customizeDescription('PartNumber param for ' +
+                            'getObject request is not supported'));
+                    }
+                    // Get all parts that belong to the requested part number
+                    if (location.dataStoreETag.startsWith(`${partNumber}:`)) {
+                        locations.push(location);
+                    }
```
LaurenSpiegel: You can break early once you have all your parts since they are in order.

lib/api/objectGet.js
bennettbuchanan: Updated.

lib/api/objectGet.js
bennettbuchanan: Added a break if the location part number is > the requested part number.

lib/api/objectGet.js
```diff
@@ -154,15 +154,31 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
...
-                if (partObj === undefined) {
+                const locations = [];
+                for (let i = 0; i < objMD.location.length; i++) {
+                    const location = objMD.location[i];
+                    // Location objects prior to GA7.1 do not include the
+                    // dataStoreETag field so we cannot find the part range
+                    if (location.dataStoreETag === undefined) {
+                        return callback(errors.InvalidObjectState
+                            .customizeDescription('PartNumber param for ' +
+                            'getObject request is not supported'));
```
bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine try

PR #849, opened 2017-08-03
Ft/paralleltesting

nicolas2bert:


tests.js
```diff
@@ -0,0 +1,186 @@
...
+                    return cb(`Failed killing server: ${cmd}`);
+                });
+            }
+            return cb();
+        }
+        return cb(`Failed: ${cmd}`);
+    });
+}
+
+if (process.env.CIRCLE_NODE_INDEX === '0') {
```
dora-korpar: is the CIRCLE_NODE_INDEX set in CircleCi?

nicolas2bert: yes :)

tests.bash
```diff
@@ -0,0 +1,162 @@
...
+
+  S3BACKEND=mem npm start > $CIRCLE_ARTIFACTS/server_mem_s3cmd.txt & bash wait_for_local_port.bash 8000 40 && npm run ft_s3cmd
+
+  killandsleep 8000
+
+  S3BACKEND=mem npm start > $CIRCLE_ARTIFACTS/server_mem_s3curl.txt & bash wait_for_local_port.bash 8000 40 && npm run ft_s3curl
+
+  killandsleep 8000
+
+  S3BACKEND=mem npm start > $CIRCLE_ARTIFACTS/server_mem_s3curl.txt & bash wait_for_local_port.bash 8000 40 && npm run ft_s3curl
```
JianqinWang: I think the same test has been run twice?

tests.bash
```diff
@@ -0,0 +1,162 @@
...
+fi
+
+if [ $CIRCLE_NODE_INDEX -eq 3 ]
+then
+
+  S3BACKEND=file S3VAULT=mem MPU_TESTING=yes npm start > $CIRCLE_ARTIFACTS/server_file_awssdk.txt & bash wait_for_local_port.bash 8000 40 && npm run ft_awssdk
+
+  killandsleep 8000
+
+  S3BACKEND=file S3VAULT=mem MPU_TESTING=yes npm start > $CIRCLE_ARTIFACTS/server_file_awssdk.txt & bash wait_for_local_port.bash 8000 40 && npm run ft_awssdk
```
JianqinWang: same test ran twice?

tests.bash
```diff
@@ -0,0 +1,162 @@
...
+
+  killandsleep 8000
+
+  S3BACKEND=file S3VAULT=mem npm start > $CIRCLE_ARTIFACTS/server_file_rawnode.txt & bash wait_for_local_port.bash 8000 40 && npm run ft_node
+
+  killandsleep 8000
+
+  # Run S3 with file Backend + KMS Encryption ; run ft_tests
+
+  S3BACKEND=file S3VAULT=mem MPU_TESTING=yes npm start > $CIRCLE_ARTIFACTS/server_file_kms_awssdk.txt & bash wait_for_local_port.bash 8000 40
```
JianqinWang: Possibly missing: '&& ENABLE_KMS_ENCRYPTION=true npm run ft_awssdk'

tests.bash
```diff
@@ -0,0 +1,162 @@
...
+set -e #exit at the first error
+
+MYPWD=$(pwd)
+
+killandsleep () {
+  kill -9 $(lsof -t -i:$1) || true
+  sleep 10
+}
+
+if [ $CIRCLE_NODE_INDEX -eq 0 ]
```
jonathan-gramain: Is it specified somewhere that we have 4 parallel runners? If not, may be good to enforce it so that it does not change under our feet and breaks tests (if there's a way)

rahulreddy: It's manual, its how we do it in Integration as well.

PR #850, opened 2017-08-03
ft: propogate redis password from config

rahulreddy:
This commit takes the optional Redis password defined in the config and
propagates the config to localCache and Utapi.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -148,9 +148,15 @@ class Config {
                 'config: invalid host for localCache. host must be a string');
             assert(typeof config.localCache.port === 'number',
                 'config: invalid port for localCache. port must be a number');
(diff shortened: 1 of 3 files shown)


rahulreddy:
@ironman-machine 
SCALITY_UTAPI_BRANCH=ft/S3C-697-redis-passsword
try

lib/Config.js
```diff
@@ -201,6 +214,13 @@ class Config {
                     'bad config: redis.port must be a number');
                 this.utapi.redis.host = config.utapi.redis.host;
                 this.utapi.redis.port = config.utapi.redis.port;
+                if (config.utapi.redis.password !== undefined) {
```
bennettbuchanan: Could these be reduced to one condition after the `if...else` block?

PR #856, opened 2017-08-07
Port/6.4 to master

rahulreddy:


rahulreddy:
@ironman-machine try

lib/RedisClient.js
```diff
@@ -7,11 +7,13 @@ module.exports = class RedisClient {
...
     * @param {string} host - Redis host
     * @param {number} port - Redis port
+    * @param {string} password - Redis password
     */
-    constructor(host, port) {
+    constructor(host, port, password) {
         this._client = new Redis({
             host,
             port,
+            password,
```
philipyoo: Does `password` here need to be wrapped in an object?
https://github.com/luin/ioredis/blob/master/API.md#new_Redis_new

rahulreddy: The `[param]` format is used to indicate that the param is optional.

PR #859, opened 2017-08-09
ft: update Backbeat routes format

rahulreddy:
This commit updates Backbeat routes format to gain flexibility on
the namespace.


rahulreddy:
@ironman-machine try

PR #861, opened 2017-08-10
FIX: Allow configurable storageClass

bennettbuchanan:


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed on retry: http://ci.ironmann.io/gh/scality/Integration/14346

PR #864, opened 2017-08-14
Fix batch delete param

dora-korpar:
batchDelete expects the first param to be an array, but was being passed an object, which caused misleading orphan keys to not be deleted if the completedHash did not match.

Addresses JIRA S3C-709


lib/api/apiUtils/object/storeObject.js
```diff
@@ -26,7 +26,9 @@ function checkHashMatchMD5(stream, hashedStream, dataRetrievalInfo, log, cb) {
             contentMD5,
         });
         log.trace('contentMD5 does not match, deleting data');
```
jonathan-gramain: Could you take this change opportunity to turn the above log to an error log, and log the object info with it? I think it's important to have a trace in logs when we delete data because of edge conditions that are not user-triggered. Also the other log could be removed since they kind of duplicate each other.

dora-korpar: Sure thing :)

dora-korpar: Wait, sorry, just for clarification - you think there should be only one log `log.error`? Or a `log.error` and `log.trace`?

jonathan-gramain: I meant just one log error, because those two log lines seem to duplicate info (e.g. you may add the 'deleting data' bit to the first log line and make it an error). Thanks!

jonathan-gramain: Maybe I misunderstood what is the md5 here, in light of Lauren's comment, if the MD5 is the one coming from the user, we should not log error then since we do not want a malicious user to flood the server with error logs. But it can still be a good idea to merge the log lines into a single one even if they are debug logs.

jonathan-gramain: In this case, it turns out to be a bigger issue than I though since a user could create orphan data on the server just by sending malformed requests. But again I'm not sure what is this actual code checking.

tests/unit/api/objectPut.js
```diff
@@ -278,6 +278,34 @@ describe('objectPut API', () => {
...
+
+        bucketPut(authInfo, testPutBucketRequest, log, () => {
+            objectPut(authInfo, testPutObjectRequest, undefined, log,
+            err => {
+                assert.deepStrictEqual(err, errors.BadDigest);
+                // orphan objects don't get deleted
+                // until the next tick
+                // in memory
+                process.nextTick(() => {
+                    // Data store starts at index 1
```
LaurenSpiegel: can you please run the cleanup() function from helpers.js before the test to make sure you are checking the correct indexes?  the test before this one seems to be checking ds[2] so if you are putting more data, it might be in ds[3]?  basically, you should run cleanup() before the test and make sure all of the ds is empty.

LaurenSpiegel:
I think we need to target 6.4 for this and then forward port to 7 and then to master.

tests/unit/api/objectPut.js
dora-korpar: It's run in the beforeEach :)

LaurenSpiegel: nifty 😄

dora-korpar:
Closing to target rel/6.4

PR #866, opened 2017-08-15
FIX: Check whether config.json changes

LaurenSpiegel:
Before overwriting config.json, confirm changes
have been made with env variables.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -94,8 +94,10 @@ if [[ "$RECORDLOG_ENABLED" ]]; then
     JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .recordLog.enabled=true"
 fi
 
-jq "$JQ_FILTERS_CONFIG" config.json > config.json.tmp
-mv config.json.tmp config.json
+if [[ $JQ_FILTERS_CONFIG != "." ]]; then
+    jq "$JQ_FILTERS_CONFIG" config.json > config.json.tmp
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -94,8 +94,10 @@ if [[ "$RECORDLOG_ENABLED" ]]; then
     JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .recordLog.enabled=true"
 fi
 
+if [[ JQ_FILTERS_CONFIG != "." ]]; then
```
jonathan-gramain: shouldn't it be `$JQ_FILTERS_CONFIG`?

docker-entrypoint.sh
```diff
@@ -94,8 +94,10 @@ if [[ "$RECORDLOG_ENABLED" ]]; then
...
 jq "$JQ_FILTERS_CONFIG" config.json > config.json.tmp
```
jonathan-gramain: minor: indent this block

electrachong:
Are they supposed to be unable to mount the config.json file if they don't specify any env variables?

An example from our [docs](http://s3-server.readthedocs.io/en/latest/DOCKER/#specifying-your-own-host-name):
`docker run -v $(pwd)/config.json:/usr/src/app/config.json -p 8000:8000 -d scality/s3server`

LaurenSpiegel:
thanks, @jonathan-gramain, updated.

PR #867, opened 2017-08-15
Doc/update docker section

LaureVergeron:
See http://s3-server.readthedocs.io/en/doc-update_docker_section/DOCKER/ for layout. Tried to put all tunables at the top, then a few examples of setup as the 42 students were confused by the current layout.

### DIFF
--- a/docs/DOCKER.rst
+++ b/docs/DOCKER.rst
@@ -1,62 +1,66 @@
 Docker
 ======
 
--  `For continuous integration with
-   Docker <#for-continuous-integration-with-docker>`__
 -  `Environment Variables <#environment-variables>`__
--  `In production with Docker <#in-production-with-docker>`__
--  `Using Docker Volume in
-   production <#using-docker-volume-in-production>`__
--  `Adding modifying or deleting accounts or users
-   credentials <#adding-modifying-or-deleting-accounts-or-users-credentials>`__
--  `Specifying your own host name <#specifying-your-own-host-name>`__
--  `Running as an unprivileged
-   user <#running-as-an-unprivileged-user>`__
-
-For continuous integration with Docker
---------------------------------------
-
-When you start the Docker Scality Zenko CloudServer image, you can adjust the
-configuration of the Scality Zenko CloudServer instance by passing one or more
-environment variables on the docker run command line.
+-  `Tunables and setup tips <#tunables-and-setup-tips>`__
+-  `Examples for continuous integration with
+   Docker <#continuous-integration-with-docker-hosted CloudServer>`__
+-  `Examples for going in production with Docker <#in-production-with-docker-hosted CloudServer>`__
 
 Environment Variables
-~~~~~~~~~~~~~~~~~~~~~
+---------------------
+
+S3DATA
+~~~~~~
 
 S3DATA=multiple
 ^^^^^^^^^^^^^^^
+Allows you to run Scality Zenko CloudServer with multiple data backends, defined
+as regions.
+When using multiple data backends, a custom ``locationConfig.json`` file is
+mandatory. It will allow you to set custom regions. You will then need to
+provide associated rest_endpoints for each custom region in your
+``config.json`` file.
+`Learn more about multiple backends configuration <../GETTING_STARTED/#location-configuration>`__
 
-This runs Scality Zenko CloudServer with multiple data backends. `More
-info <../GETTING_STARTED/#location-configuration>`__
+If you are using Scality RING endpoints, please refer to your customer
+documentation.
 
-.. code:: shell
-
-    docker run -d --name s3server -p 8000:8000 -e S3DATA=multiple scality/s3server
-
-For running an S3 AWS backend, you will have to add a new section
-(with the ``aws_s3`` location type) to your ``locationConfig.json`` file:
+Running it with an AWS S3 hosted backend
+""""""""""""""""""""""""""""""""""""""""
+To run CloudServer with an S3 AWS backend, you will have to add a new section
+to your ``locationConfig.json`` file with the ``aws_s3`` location type:
 
 .. code:: json
-
+(...)
     "aws-test": {
         "type": "aws_s3",
         "details": {
             "awsEndpoint": "s3.amazonaws.com",
             "bucketName": "yourawss3bucket",
             "bucketMatch": true,
-            "credentialsProfile": "default"
+            "credentialsProfile": "aws_hosted_profile"
         }
     }
+(...)
 
-You will also have to mount your AWS credentials file:
+You will also have to edit your AWS credentials file to be able to use your
+command line tool of choice. This file should mention credentials for all the
+backends you're using. You can use several profiles when using multiple
+profiles.
+
+.. code:: json
+[default]
+aws_access_key_id=accessKey1
+aws_secret_access_key=verySecretKey1
+[aws_hosted_profile]
+aws_access_key_id={{YOUR_ACCESS_KEY}}
+aws_secret_access_key={{YOUR_SECRET_KEY}}
+
+Just as you need to mount your locationConfig.json, you will need to mount your
+AWS credentials file at run time:
 ``-v ~/.aws/credentials:/root/.aws/credentials`` on Linux, OS X, or Unix or
 ``-v C:\Users\USERNAME\.aws\credential:/root/.aws/credentials`` on Windows
-
-.. code:: shell
-
-    docker run --name s3server -p 8000:8000
-    -v $(pwd)/locationConfig.json:/usr/src/app/locationConfig.json
-    -v ~/.aws/credentials:/root/.aws/credentials -e S3DATA=multiple scality/s3server
 
 NOTE: One account can't copy to another account with a source and
 destination on real AWS unless the account associated with the
@@ -64,8 +68,20 @@ access Key/secret Key pairs used for the destination bucket has rights
 to get in the source bucket. ACL's would have to be updated
 on AWS directly to enable this.
 
-ENDPOINT
+S3BACKEND
+~~~~~~
+
+S3BACKEND=file
+^^^^^^^^^^^
+When storing file data, for it to be persistent you must mount docker volumes
+for both data and metadata. See `this section <#using-docker-volumes-in-production>`__
+
+S3BACKEND=mem
 ^^^^^^^^^^
+This is ideal for testing - no data will remain after container is shutdown.
+
+ENDPOINT
+~~~~~~~~
 
 This variable specifies your endpoint. If you have a domain such as
 new.host.com, by specifying that here, you and your users can direct s3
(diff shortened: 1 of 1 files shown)


docs/DOCKER.rst
```diff
@@ -1,43 +1,35 @@
...
-^^^^^^^^^^^^^^^
+---------------------
 
-This runs Scality Zenko CloudServer with multiple data backends. `More
-info <../GETTING_STARTED/#location-configuration>`__
+S3DATA
+~~~~~~
 
-.. code:: shell
+S3DATA=multiple and S3DATA=scality
```
LaurenSpiegel: Please do not document S3DATA=scality.  People should be using S3DATA=multiple

docs/DOCKER.rst
```diff
@@ -1,43 +1,35 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Either of these runs Scality Zenko CloudServer with multiple data backends. The
+scality value refers to installations with a RING as one of the backends.
```
LaurenSpiegel: please remove this

docs/DOCKER.rst
```diff
@@ -1,43 +1,35 @@
...
+When using multiple data backends, a custom locationConfig.json file is
+mandatory. It will provide custom regions and associated rest_endpoints
```
LaurenSpiegel: the locationConfig.json does not set forth the rest_endpoints.  that is in the config.json.

docs/DOCKER.rst
```diff
@@ -1,43 +1,35 @@
...
+configuration `More info <../GETTING_STARTED/#location-configuration>`__
 
-    docker run -d --name s3server -p 8000:8000 -e S3DATA=multiple scality/s3server
+If you are using RING endpoints, please refer to your customer documentation.
```
LaurenSpiegel: I don't think we should have this here.

docs/DOCKER.rst
```diff
@@ -47,19 +39,28 @@ For running an S3 AWS backend, you will have to add a new section
             "credentialsProfile": "default"
         }
     }
+(...)
 
-You will also have to mount your AWS credentials file:
+You will also have to edit your AWS credentials file to be able to use your
+command line tool of choice. This file should only state credentials for S3 AWS
```
LaurenSpiegel: this does not have to do with command line tools.  We use the aws credentials file to pull the credentials for the aws sdk client within cloud server.

docs/DOCKER.rst
```diff
@@ -47,19 +39,28 @@ For running an S3 AWS backend, you will have to add a new section
...
+hosted data, not for your Zenko CloudServer hosted data.
```
LaurenSpiegel: I don't understand this sentence.  You can have numerous profiles in your aws credentials file. they need to mount the file that has any profile mentioned in their aws location.  Above, where it says "credentialsProfile":"default", this could say any other profile they choose.

docs/DOCKER.rst
```diff
@@ -47,19 +39,28 @@ For running an S3 AWS backend, you will have to add a new section
...
+As for your locationConfig.json, you will need to mount your AWS credentials
```
LaurenSpiegel: Just as you have to mount your locationConfig.json, you will also need to mount your aws credentials file.

docs/DOCKER.rst
```diff
@@ -47,19 +39,28 @@ For running an S3 AWS backend, you will have to add a new section
...
+file at run time:
 ``-v ~/.aws/credentials:/root/.aws/credentials`` on Linux, OS X, or Unix or
 ``-v C:\Users\USERNAME\.aws\credential:/root/.aws/credentials`` on Windows
 
-.. code:: shell
+S3DATA=file
```
LaurenSpiegel: I think we should simplify this and only mention S3DATA=multiple. If they want to do a quickstart with S3BACKEND=file or S3BACKEND=mem then it's all file or mem, respectively.  If they want a file or mem data location, they can do that through the locationConfig.json.

docs/DOCKER.rst
```diff
@@ -47,19 +39,28 @@ For running an S3 AWS backend, you will have to add a new section
...
+When storing file data, for it to be persistent you must mount docker volumes
+for both data and metadata. See `this section <#using-docker-volumes-in-production>`__
 
-    docker run --name s3server -p 8000:8000
-    -v $(pwd)/locationConfig.json:/usr/src/app/locationConfig.json
-    -v ~/.aws/credentials:/root/.aws/credentials -e S3DATA=multiple scality/s3server
+S3DATA=mem
+^^^^^^^^^^
+This is ideal for testing - no data will remain after container is shutdown. Setting
+S3BACKEND=mem will achieve the same result.
```
LaurenSpiegel: Please remove S3DATA=mem reference.

docs/DOCKER.rst
```diff
@@ -280,3 +280,50 @@ For instance, you can modify these lines in the dockerfile:
...
+configuration of the Scality Zenko CloudServer instance by passing one or more
+environment variables on the docker run command line.
+
+Sample ways to run it for CI are:
+
+- With custom locations, including one hosted on AWS, and custom credentials mounted:
+
+.. code:: shell
+
+    docker run --name CloudServer -p 8000:8000
```
LaurenSpiegel: I think it's confusing to use Cloudserver since our docker image is s3server.

docs/DOCKER.rst
LaureVergeron: Do you mean we should only mention S3BACKEND=mem? Or not mention the mem mode at all? From what I see at events, people use in-memory storage a lot for testing purposes.

docs/DOCKER.rst
LaureVergeron: I thought we were not allowed to use s3server for new content anymore? Plus, eventually, the image should be renamed cloudserver, and the whole documentation talks about cloudserver, so I guess it should be OK?

docs/DOCKER.rst
LaureVergeron: Fixed

docs/DOCKER.rst
LaureVergeron: You're right, I wrote this in the hackathon context for a specific team, but that's innacurate in the greater world. Providing an alternative.

docs/DOCKER.rst
LaureVergeron: Yes, that's when we mount it. I meant that they need to input their credentials in there like they would for using S3. I've tried a new wording, let me know what you think.

docs/DOCKER.rst
LaureVergeron: We do have a few customers who check the open source documentation. Are you strongly against it, or can we keep it?

docs/DOCKER.rst
LaureVergeron: Fixed, thanks!

docs/DOCKER.rst
LaureVergeron: Should S3DATA=scality only be used in production? Or just never used?

docs/DOCKER.rst
LaureVergeron: As I mentioned below, a lot of people are using S3BACKEND=mem, so I believe it's worth mentioning it here.
Regarding quickstart, this section is not under the "Getting Started" section, but is its own section, so I think it should provide a bit more detail and config options about using CloudServer in Docker. I agree that the "Getting Started" section should be simplified, though.

LaurenSpiegel: We should not mention S3DATA=mem.  As I said, I'm fine with S3BACKEND=mem.

docs/DOCKER.rst
LaurenSpiegel: At least Scality RING then.

docs/DOCKER.rst
LaureVergeron: Changed

docs/DOCKER.rst
LaureVergeron: Changed, thanks

docs/DOCKER.rst
LaureVergeron: Changed, thanks

docs/DOCKER.rst
```diff
@@ -280,3 +295,50 @@ For instance, you can modify these lines in the dockerfile:
...
+    -v $(pwd)/locationConfig.json:/usr/src/app/locationConfig.json
+    -v ~/.aws/credentials:/root/.aws/credentials 
+    -e SCALITY_ACCESS_KEY_ID=accessKey1
+    -e SCALITY_SECRET_ACCESS_KEY=verySecretKey1
+    -e S3DATA=multiple -e S3BACKEND=mem scality/s3server
+
+In production with Docker hosted CloudServer
+--------------------------------------------
+
+In production, we expect that data will be remanent, that you will use the multiple
```
LaurenSpiegel: persistent?

LaureVergeron: Fixed, thanks

docs/DOCKER.rst
```diff
@@ -280,3 +295,51 @@ For instance, you can modify these lines in the dockerfile:
...
+
+- With custom locations, including one hosted on AWS, and custom credentials mounted:
+
+.. code:: shell
+
+    docker run --name CloudServer -p 8000:8000
+    -v $(pwd)/locationConfig.json:/usr/src/app/locationConfig.json
+    -v $(pwd)/authdata.json:/usr/src/app/conf/authdata.json
+    -v ~/.aws/credentials:/root/.aws/credentials
+    -e S3DATA=multiple -e S3BACKEND=mem scality/s3server
```
LaurenSpiegel: shouldn't you be mounting the data and metadata directories as well?

docs/DOCKER.rst
```diff
@@ -280,3 +295,51 @@ For instance, you can modify these lines in the dockerfile:
...
+environment variables (see `this section <#scality-access-key-id-and-scality-secret-access-key>`__):
+
+.. code:: shell
+
+    docker run --name CloudServer -p 8000:8000
+    -v $(pwd)/locationConfig.json:/usr/src/app/locationConfig.json
+    -v ~/.aws/credentials:/root/.aws/credentials 
+    -e SCALITY_ACCESS_KEY_ID=accessKey1
+    -e SCALITY_SECRET_ACCESS_KEY=verySecretKey1
+    -e S3DATA=multiple -e S3BACKEND=mem scality/s3server
```
LaurenSpiegel: same

docs/DOCKER.rst
LaureVergeron: Well, this section is for CI and refers to a usecase with S3BACKEND=mem, so I wouldn't mount data / md volumes here... (however, S3DATA is set to multiple to enable testing against multiple backends). Does it make sense ?

LaurenSpiegel: For CI I guess it does not make sense but for the example below (custom locations) it seems we should.

LaureVergeron: I've rephrased to clarify as they're both CI examples, but I've mounted volumes in one of them with a file usecase.

LaureVergeron:
Thanks for the reviews :)
It's now rebased, I'll let you merge?

electrachong:
@ironman-machine r+

PR #868, opened 2017-08-16
Fix batch delete param

dora-korpar:
batchDelete expects the first param to be an array, but was being passed an object, which caused misleading orphan keys to not be deleted if the completedHash did not match.

Addresses JIRA S3C-709


dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine go

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

PR #870, opened 2017-08-17
bf: decode URI in backbeat routes

jonathan-gramain:
URIs received were not decoded at all in backbeat /data and /metadata
routes, so the object key was keeping the encoded characters
as-is. Then the key was applied a second URL encoding pass in the
target URL, leading to the object being replicated with a different
object key.

Fix by applying the same decoding than in S3 server to the original
URL received, with '+' replaced by ' ' prior to decoding
percent-encoded characters.

### DIFF
--- a/tests/functional/raw-node/test/routes/routeBackbeat.js
+++ b/tests/functional/raw-node/test/routes/routeBackbeat.js
@@ -83,6 +83,7 @@ describeSkipIfAWS('backbeat routes:', () => {
     describe('backbeat PUT routes:', () => {
         const testArn = 'aws::iam:123456789012:user/bart';
         const testKey = 'testkey';
+        const testKeyUTF8 = '䆩鈁櫨㟔罳';
         const testData = 'testkey data';
         const testDataMd5 = crypto.createHash('md5')
                   .update(testData, 'utf-8')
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
@ironman-machine try

tests/functional/raw-node/test/routes/routeBackbeat.js
```diff
@@ -119,36 +120,50 @@ describeSkipIfAWS('backbeat routes:', () => {
...
+               key: testKeyUTF8, encodedKey: encodeURI(testKeyUTF8) },
+             { caption: 'with percents and spaces encoded as \'+\' in key',
+               key: '50% full or 50% empty',
+               encodedKey: '50%25+full+or+50%25+empty' },
+            ].forEach(testCase => {
+                it(testCase.caption, done => {
+                    async.waterfall([next => {
+                        makeBackbeatRequest(
+                            { method: 'PUT', bucket: TEST_BUCKET,
+                              objectKey: testCase.encodedKey,
```
philipyoo: Is this variable going to be used in the future? I don't see it being used in `makeRequest` called in `makeBackbeatRequest`

jonathan-gramain: If you're talking about `objectKey`, it's used to make the url path in the body of `makeBackbeatRequest`.

philipyoo: ah missed that, thanks :)

PR #874, opened 2017-08-17
FT: Send key and dataStoreName for putData response

bennettbuchanan:
Depends on https://github.com/scality/backbeat/pull/65.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -103,8 +103,6 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
             const dataRetrievalInfo = [{
                 key,
                 dataStoreName,
-                size: payloadLen,
-                start: 0,
             }];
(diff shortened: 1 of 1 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -105,6 +110,7 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
                 dataStoreName,
                 size: payloadLen,
                 start: 0,
+                dataStoreETag: `${partNumber}:${md5}`,
```
rahulreddy: I don't see the value of sending partnumber and md5 from backbeat only to be sent back to backbeat. We can just construct this in BB instead.

bennettbuchanan:
@ironman-machine 
SCALITY_BACKBEAT_BRANCH=ft/add-partNumber-param-to-putData
try

bennettbuchanan:
@ironman-machine try

PR #875, opened 2017-08-17
Forward/rel/6.4

dora-korpar:


dora-korpar:
@ironman-machine try

rahulreddy:
@ironman-machine try

lib/Config.js
```diff
@@ -377,9 +377,15 @@ class Config {
                 'config: invalid host for localCache. host must be a string');
             assert(typeof config.localCache.port === 'number',
                 'config: invalid port for localCache. port must be a number');
+            if (config.localCache.password !== undefined) {
```
ThibaultRiviere: Does `config.localCache.password` can be `undefined` and in this case it's not a problem to set it after ?

package.json
```diff
@@ -58,7 +58,7 @@
     "ft_s3cmd": "cd tests/functional/s3cmd && mocha -t 40000 *.js",
     "ft_s3curl": "cd tests/functional/s3curl && mocha -t 40000 *.js",
     "ft_test": "npm-run-all -s ft_awssdk ft_s3cmd ft_s3curl ft_node",
-    "install_ft_deps": "npm install aws-sdk@2.28.0 bluebird@3.3.1 mocha@2.3.4 mocha-junit-reporter@1.11.1 tv4@1.2.7",
+    "install_ft_deps": "npm install aws-sdk@2.28.0 bluebird@3.3.1 mocha@2.3.4 mocha-junit-reporter@1.11.1 tv4@1.2.7 zombie@5.0.5",
```
LaurenSpiegel: I don't think we need zombie in rel/7.0 or master. It was a dependency for the bucket website tests that we ended up removing through a refactor by @electrachong I believe.

lib/Config.js
rahulreddy: @ThibaultRiviere Redis password is optional, so the config check will apply only when it's defined.

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

PR #876, opened 2017-08-18
Add dataStoreName to initMPU md

bennettbuchanan:


lib/data/multipleBackendGateway.js
```diff
@@ -215,9 +215,9 @@ const multipleBackendGateway = {
     },
 
     objectTagging: (method, key, bucket, objectMD, log, cb) => {
-        const location = objectMD.dataStoreName;
+        const location = objectMD.dataStoreName || undefined;
```
bennettbuchanan: Could this instead be `objectMD.controllingLocationConstraint` instead of adding another field `dataStoreName`?

In any case, seems `|| undefined` isn't necessary.

dora-korpar: the `|| undefined` is for backwards compatibility. I'll add a comment. And we use the `dataStoreName` attribute from the putObj and deleteObj APIs. A better fix would be to get rid of the `objectMD.controllingLocationConstraint` param, but it's used for other MPU things, and for the sake of expediency of the fix, I opted for having both.

lib/data/multipleBackendGateway.js
```diff
@@ -215,9 +215,9 @@ const multipleBackendGateway = {
...
         const client = clients[location];
-        if (client[`object${method}Tagging`]) {
+        if (location && client[`object${method}Tagging`]) {
```
rahulreddy: Should this be
```js
if( client && client[`object${method}Tagging`]) {
```
?

lib/data/multipleBackendGateway.js
bennettbuchanan: I was thinking that if `objectMd` does not have the field `dataStoreName`, it will be `undefined` anyway.

rahulreddy:
@ironman-machine try

PR #878, opened 2017-08-23
FIX: return NotImplemented put version to external data

nicolas2bert:
This PR returns error `NotImplemented` when putting a version to external data backend (ie AWS/AZURE)

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -16,6 +16,9 @@ const validateWebsiteHeader = require('./websiteServing')
     .validateWebsiteHeader;
 const externalBackends = constants.externalBackends;
 
+const externalVersioningErrorMessage = 'We do not currently support putting ' +
+'a versioned object to a location-constraint of type AWS or Azure.';
+
 
 function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
     metadataStoreParams, dataToDelete, deleteLog, requestMethod, callback) {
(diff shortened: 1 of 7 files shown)


lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -16,6 +16,9 @@ const validateWebsiteHeader = require('./websiteServing')
     .validateWebsiteHeader;
 const externalBackends = constants.externalBackends;
 
+const externalVersioningErrorMessage = 'For now, not allowed to put version ' +
+'on either AWS or Azure';
```
electrachong: maybe `For now, putting a version is not allowed on AWS or Azure backends` ?

nicolas2bert:
@ironman-machine try

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -120,10 +123,25 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             callback(backendInfoObj.err);
         });
     }
+
+    const vcfg = bucketMD.getVersioningConfiguration();
+    const isVersionedObj = vcfg && vcfg.Status === 'Enabled';
+    metadataStoreParams.isVersionedObj = isVersionedObj;
```
electrachong: I'm not 100% convinced we need to attach whether versioning is enabled on the bucket to the `metadataStoreParams` just to pass it back to the api so we don't have to call `bucketMd.getVersioningConfiguration` a second time. The `metadataStoreParams` is typically for storing information in the object's metadata.

If anything, we can store the result of `bucketMD.getVersioningConfiguration()` early in the objectPut api and then pass it directly in the arguments for `createAndStoreObject`, the only other place we are using this variable. You may have to modify the arguments for all the other places createAndStoreObject is used (if it's used anywhere else). But since bucketMD is passed anyway, I don't think it makes much difference if you just call `bucketMd.getVersioningConfiguration()` again. It's up to you!

Do you also have to update copyObject and completeMpu to check if the bucket has versioning enabled and return NotImplemented if the backend is AWS/Azure?

lib/api/apiUtils/object/createAndStoreObject.js
nicolas2bert: Thanks

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -16,6 +16,9 @@ const validateWebsiteHeader = require('./websiteServing')
     .validateWebsiteHeader;
 const externalBackends = constants.externalBackends;
 
+const externalVersioningErrorMessage = 'For now, putting a version is ' +
+'not allowed on AWS or Azure backends';
```
LaurenSpiegel: I would change this to: We do not currently support putting a versioned object to a location-constraint of type AWS or Azure.

nicolas2bert:
@ironman-machine try

PR #879, opened 2017-08-23
Include bucket, object and version counts in reports

RachedBenMustapha:
This change adds object, bucket and version counts to the report handler. Only implemented for bucketfile, which counts by inspecting all keys in the root DB.

### DIFF
--- a/lib/metadata/bucketfile/backend.js
+++ b/lib/metadata/bucketfile/backend.js
@@ -8,6 +8,7 @@ const { config } = require('../../Config');
 
 const errors = arsenal.errors;
 const MetadataFileClient = arsenal.storage.metadata.MetadataFileClient;
+const versionSep = arsenal.versioning.VersioningConstants.VersionId.Separator;
 
 const METASTORE = '__metastore';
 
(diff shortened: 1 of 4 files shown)


lib/metadata/bucketfile/backend.js
```diff
@@ -312,6 +313,59 @@ class BucketFileInterface {
...
+        let cbDone = false;
+
+        this.mdDB.rawListKeys(requestParams, (err, stream) => {
+            if (err) {
+                return cb(err);
+            }
+            stream
+                .on('data', e => {
+                    if (e.indexOf(METASTORE) === -1) {
+                        if (e.indexOf(constants.usersBucket) !== -1) {
```
bennettbuchanan: Minor: Using `includes` may be more readable.

lib/utilities/reportHandler.js
```diff
@@ -22,6 +22,28 @@ function isAuthorized(clientIP, req) {
...
+        (Date.now() - lastItemScanTime) > itemScanRefreshDelay) {
+        metadata.countItems(log, (err, res) => {
+            if (err) {
+                return cb(err);
+            }
+            lastItemScanTime = Date.now();
+            lastItemScanResult = res;
+            return cb(null, res);
+        });
+    } else {
```
bennettbuchanan: This else block is not needed since returning anyway.

bennettbuchanan: Minor: could return the method above instead of using an else block and returning undefined.

lib/metadata/wrapper.js
```diff
@@ -227,6 +227,18 @@ const metadata = {
         }
         return client.getDiskUsage(cb);
     },
+
+    countItems: (log, cb) => {
+        if (!client.countItems) {
+            log.debug('returning zero item counts as fallback', { implName });
```
electrachong: you'd rather return a response with 0 items instead of returning not implemented?

lib/utilities/reportHandler.js
```diff
@@ -22,6 +22,26 @@ function isAuthorized(clientIP, req) {
         req.headers['x-scal-report-token'] === config.reportToken;
 }
 
+const itemScanRefreshDelay = 1000 * 30 * 60; // 30 minutes
+let lastItemScanTime = undefined;
```
electrachong: this is a naive question, but does value reset each time the file is imported?

RachedBenMustapha: I may be wrong but I don't believe so, `require`'d modules are cached so module-init code does not run several times.

lib/metadata/wrapper.js
RachedBenMustapha: Yes, since it's management code, I'd rather have the error handling be for actual errors are are retryable, not for unsupported configurations that would never work anyway.

electrachong:
@ironman-machine r+

lib/utilities/reportHandler.js
```diff
@@ -22,6 +22,26 @@ function isAuthorized(clientIP, req) {
...
+        return metadata.countItems(log, (err, res) => {
+            if (err) {
+                return cb(err);
+            }
+            lastItemScanTime = Date.now();
+            lastItemScanResult = res;
+            return cb(null, res);
+        });
+    }
+    return cb(null, lastItemScanResult);
```
jonathan-gramain: I would suggest wrapping this in `process.nextTick` to avoid any chance of [releasing zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony)

electrachong:
r+

electrachong:
@ironman-machine r+

electrachong:
@ironman-machine r+

PR #880, opened 2017-08-23
FX: external backend ENCRYPTION

nicolas2bert:
In s3server we have bucket level encryption. 
AWS S3 has object level encryption. 
When we did the AWS backend on s3server, we decided to rely on AWS S3's object level encryption (instead of our own) so that encrypted objects would be readable if accessed directly from AWS S3. Now that we are bringing Azure into the picture, this gets complicated since Azure encryption is at the account level and s3server will not be messing with the account level settings.  

To make things consistent and to best carry-out the intent of users:

1) If a user creates an encrypted bucket, we always do the encryption on a put to this bucket (even if going to the azure or aws backends).

2) When setting up a locationConfig entry for the aws backend, we will allow them to set an option of "serverSideEncryption: true".  If they set this option, any time we put to this location constraint, we will send the appropriate headers to AWS so that they encrypt.

NOTE: I could not find any documentation about the CloudServer Encryption

### DIFF
--- a/docs/ARCHITECTURE.rst
+++ b/docs/ARCHITECTURE.rst
@@ -937,3 +937,53 @@ Algorithms
 The algorithms for each listing type can be found in the open-source
 `scality/Arsenal <https://github.com/scality/Arsenal>`__ repository, in
 `lib/algos/list <https://github.com/scality/Arsenal/tree/master/lib/algos/list>`__.
+
+Encryption
+===========
+
+With CloudServer, there are two possible methods of at-rest encryption.
+(1) We offer bucket level encryption where Scality CloudServer itself handles at-rest
+encryption for any object that is in an 'encrypted' bucket, regardless of what
+the location-constraint for the data is and
+(2) If the location-constraint specified for the data is of type AWS,
+you can choose to use AWS server side encryption.
+
+Note: bucket level encryption is not available on the standard AWS
+S3 protocol, so normal AWS S3 clients will not provide the option to send a
+header when creating a bucket. We have created a simple tool to enable you
+to easily create an encrypted bucket.
+
+Example:
+--------
+
+Creating encrypted bucket using our encrypted bucket tool in the bin directory
+
+.. code:: shell
+
+    ./create_encrypted_bucket.js -a accessKey1 -k verySecretKey1 -b bucketname -h localhost -p 8000
+
+
+AWS backend
+------------
+
+With real AWS S3 as a location-constraint, you have to configure the
+location-constraint as follows
+
+.. code:: json
+
+    "aws-test": {
+        "type": "aws_s3",
+        "legacyAwsBehavior": true,
+        "details": {
+            "serverSideEncryption": true,
+            ...
+        }
+    },
+
+Then, every time an object is put to that data location, we pass the following
+header to AWS: ``x-amz-server-side-encryption: AES256``
+
+Note: due to these options, it is possible to configure encryption by both
+CloudServer and AWS S3 (if you put an object to a CloudServer bucket which has
+the encryption flag AND the location-constraint for the data is AWS S3 with
+serverSideEncryption set to true).
--- a/tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
+++ b/tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
@@ -10,6 +10,7 @@ const { createEncryptedBucketPromise } =
 const { versioningEnabled } = require('../../../lib/utility/versioning-util');
 
 const awsLocation = 'aws-test';
+const awsLocationEncryption = 'aws-test-encryption';
 const bucket = 'buckettestmultiplebackendput';
 const key = `somekey-${Date.now()}`;
 const emptyKey = `emptykey-${Date.now()}`;
@@ -30,10 +31,9 @@ const describeSkipIfNotMultiple = (config.backends.data !== 'multiple'
 
 const awsTimeout = 20000;
 const retryTimeout = 10000;
+let awsBucket;
 
-function awsGetCheck(objectKey, s3MD5, awsMD5, cb) {
-    const awsBucket = config.locationConstraints[awsLocation].
-        details.bucketName;
+function awsGetCheck(objectKey, s3MD5, awsMD5, location, cb) {
     s3.getObject({ Bucket: bucket, Key: objectKey },
     function s3GetCallback(err, res) {
         if (err && err.code === 'NetworkingError') {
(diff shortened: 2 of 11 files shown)


docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,50 @@ Algorithms
...
+
+Encryption
+===========
+
+About Server Side Encryption CloudServer has its own bucket level encryption.
+
+Differentiations
+------------------
+
+With the real AWS S3, to enable encryption, you have to specify on the object
```
LaurenSpiegel: With real AWS S3 as a location-constraint, you have to configure the location-constraint as follows....

Then, every time an object is put to that data location, we pass the following header to AWS: `x-amz-server-side-encryption: AES256`

docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,50 @@ Algorithms
...
+level a ``x-amz-server-side-encryption: AES256`` header when you upload an object
+
```
LaurenSpiegel: Please note that your description above is not correct.  The user does not send the header -- we do.

docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,50 @@ Algorithms
...
+This enables encryption for all the objects stored in your Scality CloudServer **bucket**.
+
+Example:
+--------
+
+Creating encrypted bucket named: encryptedbucket
+
+.. code:: shell
+
+    s3cmd mb s3://encryptedbucket --add-header x-amz-scal-server-side-encryption:AES256
```
LaurenSpiegel: This won't work.  There is no ability to add a header on a bucket put in the regular clients since this is a divergence from the regular S3 spec.  instead, point them to our create encrypted bucket tool in the bin directory.

docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,50 @@ Algorithms
...
+        "legacyAwsBehavior": true,
+        "serverSideEncryption": true,
+        "details": {
+            ...
+        }
+    },
+
+If you set this option, any time you upload an object with this location constraint,
+your CloudServer will send ``x-amz-scal-server-side-encryption: AES256`` header to AWS so the
+object will directly be encrypted by AWS (AES256 encryption only).
```
LaurenSpiegel: Please add a note:

Note that due to these options, it is possible to configure encryption by both CloudServer and AWS S3 (if you put an object to a CloudServer bucket which has the encryption flag AND the location-constraint for the data is AWS S3 with serverSideEncryption set to true).

tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
```diff
@@ -44,12 +45,12 @@ function awsGetCheck(objectKey, s3MD5, awsMD5, cb) {
...
         assert.strictEqual(res.Metadata['scal-location-constraint'],
-            awsLocation);
+            location);
         return awsS3.getObject({ Bucket: awsBucket, Key: objectKey },
         (err, res) => {
             assert.strictEqual(err, null, 'Expected success, got error ' +
             `on direct AWS call: ${err}`);
-            if (process.env.ENABLE_KMS_ENCRYPTION === 'true') {
+            if (location === awsLocationEncryption) {
                 // doesn't check ETag because it's different
```
LaurenSpiegel: please add one more test where you put to the regular aws test location in an encrypted bucket and you check to make sure the data on aws is actually encrypted (so get it directly from aws and check that it is not the same as the data that was first sent to cloudServer).

docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,48 @@ Algorithms
...
+With CloudServer, there are two possible methods of at-rest encryption.
+(1) We offer bucket level encryption where Scality CloudServer itself handles at-rest
+encryption for any object that is in an 'encrypted' bucket regardless of what
+the location-constraint for the data is and
+(2) If the location-constraint specified for the data is of type AWS,
+you can select to use AWS server side encryption.
+
+Example:
+--------
+
```
LaurenSpiegel: Note that bucket level encryption is not available on the standard AWS S3 protocol. So normal AWS S3 clients will not provide the option to send a header when creating a bucket. We have created a simple tool to enable you to easily create an encrypted bucket.

docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,53 @@ Algorithms
...
+With CloudServer, there are two possible methods of at-rest encryption.
+(1) We offer bucket level encryption where Scality CloudServer itself handles at-rest
+encryption for any object that is in an 'encrypted' bucket regardless of what
+the location-constraint for the data is and
+(2) If the location-constraint specified for the data is of type AWS,
+you can select to use AWS server side encryption.
+
+Note: bucket level encryption is not available on the standard AWS
+S3 protocol. So normal AWS S3 clients will not provide the option to send a
```
electrachong: *`protocol, so normal`

docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,53 @@ Algorithms
...
+
+Encryption
+===========
+
+With CloudServer, there are two possible methods of at-rest encryption.
+(1) We offer bucket level encryption where Scality CloudServer itself handles at-rest
+encryption for any object that is in an 'encrypted' bucket regardless of what
+the location-constraint for the data is and
+(2) If the location-constraint specified for the data is of type AWS,
+you can select to use AWS server side encryption.
```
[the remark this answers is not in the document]
electrachong: or *specify to use?

LaurenSpiegel:
@electrachong, I think leaving the encryption in the main document is fine for now. If people want more detail on the encryption we can expand it into its own doc later.

nicolas2bert:
@ironman-machine try

docs/ARCHITECTURE.rst
```diff
@@ -937,3 +937,53 @@ Algorithms
...
+
+With real AWS S3 as a location-constraint, you have to configure the
+location-constraint as follows
+
+.. code:: json
+
+    "aws-test": {
+        "type": "aws_s3",
+        "legacyAwsBehavior": true,
+        "serverSideEncryption": true,
```
LaurenSpiegel: can we nest this option under details?

LaurenSpiegel:
@ironman-machine r+

tests/multipleBackend/backendHealthcheckResponse.js
```diff
@@ -16,14 +14,4 @@ describe('Healthcheck response', () => {
             done();
         });
     });
-    it('should return result for every location constraint in ' +
```
LaurenSpiegel: we shouldn't just delete this test.  we should test to make sure at least one of every azure and aws backend is getting called.

tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
```diff
@@ -198,7 +203,41 @@ describe('MultipleBackend put object', function testSuite() {
...
                         `got error ${err}`);
                     setTimeout(() => {
-                        awsGetCheck(key, correctMD5, correctMD5, () => done());
+                        awsGetCheck(key, correctMD5, correctMD5, awsLocation,
+                          () => done());
+                    }, awsTimeout);
+                });
+            });
+
+            itSkipIfNoEncryption('should encrypt body putting an object to AWS',
```
LaurenSpiegel: make the assertions in the test conditional based on whether the bucket is encrypted and let's clarify the description of the test.

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine try

PR #881, opened 2017-08-23
Ft/backbeat md updates

jonathan-gramain:
Dependency of https://github.com/scality/backbeat/pull/75

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -120,6 +120,18 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             // FIXME: add error type MalformedJSON
             return callback(errors.MalformedPOSTRequest);
         }
+        const { headers, bucketName, objectKey } = request;
+        // check if it's metadata only operation
+        if (headers['x-scal-replication-content'] === 'METADATA') {
+            if (!objMd) {
+                // if the target does not exist, return an error to
+                // backbeat, who will have to retry the operation as a
+                // complete replication
+                return callback(errors.ObjNotFound);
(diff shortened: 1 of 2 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -120,6 +120,12 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             // FIXME: add error type MalformedJSON
             return callback(errors.MalformedPOSTRequest);
         }
+        const { headers, bucketName, objectKey } = request;
+        // check if it's metadata only operation
+        if (headers['x-scal-replication-content'] === 'METADATA') {
+            // copy original data locations
+            omVal.location = Array.from(objMd.location);
```
bennettbuchanan: Do we need a check if objMd is `undefined`? Looking here:
https://github.com/scality/S3/blob/93d20912a2e58fd389b7ec9765043fa2adb2779a/lib/metadata/metadataUtils.js#L196

rahulreddy: Yes, we have to. Good catch.

jonathan-gramain: This seems to be occurring in case the object does not exist on the target. I think in such case we should just reject the whole request.

jonathan-gramain: Only in metadata-only mode of course

electrachong:
r+

electrachong:
@ironman-machine r+

electrachong:
@ironman-machine r-

jonathan-gramain:
@ironman-machine try

PR #883, opened 2017-08-24
bf: refuse backbeat PUT if bucket is not versioned

jonathan-gramain:
Both data and metadata routes now return InvalidBucketState (as HTTP 409)
if bucket versioning is not enabled, to prevent various issues linked with
putting versions on a non-versioned bucket.


jonathan-gramain:
@ironman-machine try

rahulreddy:
End to end passed here http://ci.ironmann.io/gh/scality/Integration/14615

PR #884, opened 2017-08-26
Ft/azure putpart and abort mpu

dora-korpar:
putPart and abortMPU APIs for Azure data backend

### DIFF
--- a/lib/api/multipartDelete.js
+++ b/lib/api/multipartDelete.js
@@ -77,26 +77,34 @@ function multipartDelete(authInfo, request, log, callback) {
                 const location = mpuOverviewObj.
                     controllingLocationConstraint;
                 return multipleBackendGateway.abortMPU(objectKey, uploadId,
-                location, destBucket, log, err => {
+                location, destBucket, log, (err, skipDataDelete) => {
                     if (err) {
                         return next(err, destBucket);
                     }
-                    return next(null, mpuBucket, destBucket);
+                    return next(null, mpuBucket, destBucket,
+                    skipDataDelete);
                 });
             }
-            return next(null, mpuBucket, destBucket);
+            return next(null, mpuBucket, destBucket, false);
         },
-        function getPartLocations(mpuBucket, destBucket, next) {
+        function getPartLocations(mpuBucket, destBucket, skipDataDelete,
+        next) {
             services.getMPUparts(mpuBucket.getName(), uploadId, log,
                 (err, result) => {
                     if (err) {
                         return next(err, destBucket);
                     }
                     const storedParts = result.Contents;
-                    return next(null, mpuBucket, storedParts, destBucket);
+                    return next(null, mpuBucket, storedParts, destBucket,
+                    skipDataDelete);
                 });
         },
-        function deleteData(mpuBucket, storedParts, destBucket, next) {
+        function deleteData(mpuBucket, storedParts, destBucket,
+        skipDataDelete, next) {
+            // for Azure we do not need to delete data
+            if (skipDataDelete) {
+                return next(null, mpuBucket, storedParts, destBucket);
+            }
             // The locations were sent to metadata as an array
             // under partLocations.  Pull the partLocations.
             let locations = storedParts.map(item => item.value.partLocations);
--- a/lib/api/objectPutPart.js
+++ b/lib/api/objectPutPart.js
@@ -196,22 +196,27 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                             'objectPutPart::multipleBackendGateway.uploadPart',
                         });
                         return next(err, destinationBucket);
-                    } else if (partInfo) {
+                    } else if (partInfo &&
+                        partInfo.dataStoreType === 'aws_s3') {
                         // if data backend handles MPU, skip to end of waterfall
                         return next(skipError, destinationBucket,
                         partInfo.dataStoreETag);
+                    } else if (partInfo && partInfo.dataStoreType === 'azure') {
+                        return next(null, destinationBucket,
+                            objectLocationConstraint, cipherBundle, splitter,
+                            partInfo);
                     }
                     // if data backend doesn't handle MPU, continue waterfall
                     return next(null, destinationBucket,
-                    objectLocationConstraint, cipherBundle, splitter);
+                    objectLocationConstraint, cipherBundle, splitter, null);
                 });
             }
             return next(null, destinationBucket, objectLocationConstraint,
-            cipherBundle, splitter);
+            cipherBundle, splitter, null);
         },
         // Get any pre-existing part.
         (destinationBucket, objectLocationConstraint, cipherBundle,
-            splitter, next) => {
+            splitter, partInfo, next) => {
             const paddedPartNumber = _getPaddedPartNumber(partNumber);
             const partKey = _getPartKey(uploadId, splitter, paddedPartNumber);
             return metadata.getObjectMD(mpuBucketName, partKey, {}, log,
@@ -238,12 +243,20 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                     }
                     return next(null, destinationBucket,
                         objectLocationConstraint, cipherBundle,
-                        partKey, prevObjectSize, oldLocations);
+                        partKey, prevObjectSize, oldLocations, partInfo);
                 });
         },
         // Store in data backend.
         (destinationBucket, objectLocationConstraint, cipherBundle,
-            partKey, prevObjectSize, oldLocations, next) => {
+        partKey, prevObjectSize, oldLocations, partInfo, next) => {
+            // NOTE: set oldLocations to null so we do not batchDelete for now
+            if (partInfo && partInfo.dataStoreType === 'azure') {
+                // skip to storing metadata
+                return next(null, destinationBucket, partInfo,
+                  partInfo.dataStoreETag,
+                  cipherBundle, partKey, prevObjectSize, null,
+                  objectLocationConstraint);
+            }
             const objectContext = {
                 bucketName,
                 owner: canonicalID,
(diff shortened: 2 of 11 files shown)


dora-korpar:
@ironman-machine try

dora-korpar:
E2E passed here: http://ci.ironmann.io/gh/scality/Integration/14724

lib/api/objectPutPart.js
```diff
@@ -238,12 +243,20 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
                         objectLocationConstraint, cipherBundle,
-                        partKey, prevObjectSize, oldLocations);
+                        partKey, prevObjectSize, oldLocations, partInfo);
                 });
         },
         // Store in data backend.
         (destinationBucket, objectLocationConstraint, cipherBundle,
-            partKey, prevObjectSize, oldLocations, next) => {
+        partKey, prevObjectSize, oldLocations, partInfo, next) => {
+            // NOTE: set oldLocations to null so we do not batchDelete for now
```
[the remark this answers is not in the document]
LaurenSpiegel: As discussed, it sounds like if you put part 1 again and it has 5 subparts, it will overwrite the prior say 5 subparts.  We should test that -- make sure if you put a block with the same name again, it does not error and that the committed final blob has the new data. If there were 6 subparts before, the last one will be cleaned up when you commit a block list without it.

lib/data/external/AzureClient.js
```diff
@@ -125,6 +236,74 @@ class AzureClient {
         });
     }
 
+    uploadPart(request, streamingV4Params, partStream, size, key, uploadId,
```
LaurenSpiegel: As discussed, we have to determine what happens if you have an existing blob called foo and then you put a block on it. If it erases the original block, we have a problem since our metadata will still think it is there.

LaurenSpiegel: There is a similar issue on delete.  If someone deletes foo and our metadata thinks it is there because there was a prior foo and it actually wipes out our ongoing block list we have a problem.  There are two types of delete with the azure sdk -- delete and deleteifExists. deleteIfExists should not delete your blocks so should be used on a delete call.

lib/data/external/AzureClient.js
```diff
@@ -53,6 +60,110 @@ class AzureClient {
         return this._translateMetaHeaders(metaHeaders);
     }
 
+    _putSinglePart(request, params, log, cb) {
```
LaurenSpiegel: is this code different from what is in mystique? can these functions be added to arsenal?

lib/data/external/AzureClient.js
```diff
@@ -125,6 +236,74 @@ class AzureClient {
...
+        const params = { bucketName: this._azureContainerName,
+            partNumber, size, objectKey: azureKey, contentMD5, uploadId };
+        const dataRetrievalInfo = {
+            key: partNumber,
+            dataStoreName: this._dataStoreName,
+            dataStoreType: 'azure',
+        };
+
+        if (size === 0) {
+            // Azure does not allow putting empty blocks; instead we
```
LaurenSpiegel: I think this comment is wrong.

lib/data/external/AzureClient.js
```diff
@@ -125,6 +236,74 @@ class AzureClient {
...
+            dataRetrievalInfo.dataStoreETag = dataStoreETag;
+            dataRetrievalInfo.numberSubParts = numberSubParts;
+            return callback(null, dataRetrievalInfo);
+        });
+    }
+
+    abortMPU(key, uploadId, bucket, log, callback) {
+        const azureBucket = this._azureContainerName;
+        const azureKey = this._createAzureKey(bucket, key, this._bucketMatch);
+        return this._client.deleteBlob(azureBucket, azureKey, err => {
```
LaurenSpiegel: what happens if there was already a fully fledged key under this name and blocks were added and then there is an abort? assuming you can add blocks to an existing key, we don't want to wipe out the original key on the abort call.

tests/functional/aws-node-sdk/test/multipleBackend/mpuAbort/azureAbortMPU.js
```diff
@@ -0,0 +1,123 @@
...
+
+const describeSkipIfNotMultiple = (config.backends.data !== 'multiple'
+    || process.env.S3_END_TO_END) ? describe.skip : describe;
+
+let bucketUtil;
+let s3;
+
+describeSkipIfNotMultiple('Abort MPU on Azure data backend', function
+describeF() {
+    this.timeout(250000);
```
LaurenSpiegel: ouch. what do you think is taking this long?

tests/functional/aws-node-sdk/test/multipleBackend/mpuAbort/azureAbortMPU.js
```diff
@@ -0,0 +1,123 @@
...
+                        }, err => next(err)),
+                        next => azureClient.getBlobProperties(
+                        azureContainerName,
+                        this.test.key, err => {
+                            assert.strictEqual(err.statusCode, 404);
+                            assert.strictEqual(err.code, 'NotFound');
+                            return next();
+                        }),
+                    ], done);
+                });
```
LaurenSpiegel: we need to test what happens when there was a prior key with the same name.

tests/functional/aws-node-sdk/test/multipleBackend/mpuParts/azurePutPart.js
```diff
@@ -0,0 +1,218 @@
...
+            assert.strictEqual(l.Name, getBlockId(uploadId,
+                expectedParts[index].partnbr, expectedParts[index].subpartnbr));
+        });
+        cb();
+    });
+}
+
+describeSkipIfNotMultiple('MultipleBackend put part to AZURE', function
+describeF() {
+    this.timeout(250000);
```
LaurenSpiegel: same issue

tests/functional/aws-node-sdk/test/multipleBackend/mpuParts/azurePutPart.js
```diff
@@ -0,0 +1,218 @@
...
+                        Body: body2,
+                    }, (err, res) => {
+                        const eTagExpected = expectedETag(body2);
+                        assert.strictEqual(res.ETag, eTagExpected);
+                        return next(err);
+                    }),
+                    next => checkSubPart(this.test.key, this.test.uploadId,
+                    parts2, next),
+                ], done);
+            });
```
LaurenSpiegel: need to test what happens when there was already a key with the same name as the mpu key.

dora-korpar:
@ironman-machine try

lib/data/external/AzureClient.js
```diff
@@ -125,6 +236,53 @@ class AzureClient {
         });
     }
 
+    uploadPart(request, streamingV4Params, partStream, size, key, uploadId,
+    partNumber, bucket, log, callback) {
+        const azureKey = this._createAzureKey(bucket, key, this._bucketMatch);
+        const contentMD5 = request.headers['content-md5'];
```
nicolas2bert: Can you please do a check: `if (request.headers['content-md5'])` ?

dora-korpar:
@ironman-machine try

lib/api/multipartDelete.js
```diff
@@ -77,26 +77,34 @@ function multipartDelete(authInfo, request, log, callback) {
...
-                    return next(null, mpuBucket, storedParts, destBucket);
+                    return next(null, mpuBucket, storedParts, destBucket,
+                    cleanupMD);
                 });
         },
-        function deleteData(mpuBucket, storedParts, destBucket, next) {
+        function deleteData(mpuBucket, storedParts, destBucket,
+        cleanupMD, next) {
+            // for Azure we do not need to delete data
+            if (cleanupMD) {
```
LaurenSpiegel: let's call this skipDataDelete

lib/api/objectPutPart.js
```diff
@@ -280,14 +293,14 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
-                // to an array
-                'md-model-version': 3,
+                // Version 4 adds number-subparts for Azure data backend
+                'md-model-version': 4,
                 partLocations,
                 'key': partKey,
                 'last-modified': new Date().toJSON(),
                 'content-md5': hexDigest,
                 'content-length': size,
+                'number-subparts': dataGetInfo.numberSubParts,
```
LaurenSpiegel: @electrachong, do we need this or can we just rely on the total size?

lib/data/external/AzureClient.js
LaurenSpiegel: We are not using deleteIfExists above (around line 100). I think we need to or else we might delete an ongoing mpu.  Also, we need a test where there was an object foo.  then we start an mpu with the name foo.  then there is a delete call to foo.  the original foo object should be deleted but the in progress mpu should not be "harmed" (confirm by completing the mpu and making sure you can get the data.

lib/data/multipleBackendGateway.js
```diff
@@ -201,6 +201,10 @@ const multipleBackendGateway = {
 
     abortMPU: (key, uploadId, location, bucket, log, cb) => {
         const client = clients[location];
+        if (client.clientType === 'azure') {
+            const cleanupMD = true;
```
LaurenSpiegel: skipDataCleanup (or whatever we decided above)

tests/functional/aws-node-sdk/test/multipleBackend/mpuParts/azurePutPart.js
LaurenSpiegel: still need to test this with pre-existing key to make sure we are not overwriting.  you can use some of the setup stuff you have for the abort test but need this for put part as well.

dora-korpar:
@LaurenSpiegel Not sure why I can't reply directly to your comments, but for the first one involving adding a test, that one will be included with the completeMPU API and I've added the second one

dora-korpar:
@ironman-machine try

lib/api/objectPutPart.js
electrachong: you can rely on total size but Arsenal mpu utils for putting subparts needs to be updated to put the correct size for the subparts (I have an in-progress branch for it)

dora-korpar:
@ironman-machine try

LaurenSpiegel:
@ironman-machine r+

PR #889, opened 2017-09-03
CHORE: Update deps for release branch

LaurenSpiegel:


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "S3 connector",
   "main": "index.js",
   "engines": {
-    "node": "6.9.5"
+    "node": ">=6.9.5"
```
ThibaultRiviere: Is it normal?

LaurenSpiegel: It's not required for the release branch update but seemed like an efficient way to fix this. Because metadata is using yarn, to update our node version at all we need to have this in any project that is a dependency of metadata so we should be consistent across the projects.

LaurenSpiegel:
@ironman-machine r+

PR #893, opened 2017-09-08
provide pushMetric with objects

ploki:
Provide pushMetric with objects on the methods with storage side effects

The purpose of this PR is to permit other services accessing the same data set to keep their states consistent

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -480,9 +480,11 @@ function multiObjectDelete(authInfo, request, log, callback) {
         }
         const xml = _formatXML(quietSetting, errorResults,
             successfullyDeleted);
+        const deletedKeys = successfullyDeleted.map(item => item.key);
         pushMetric('multiObjectDelete', log, {
             authInfo,
             bucket: bucketName,
+            keys: deletedKeys,
             byteLength: totalContentLengthDeleted,
             numberOfObjects: numOfObjectsRemoved,
         });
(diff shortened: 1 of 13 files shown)


ploki:
@ironman-machine SCALITY_UTAPI_BRANCH=ft/publish-side-effects

ploki:
@ironman-machine try

ploki:
@ironman-machine try

lib/api/multiObjectDelete.js
```diff
@@ -483,6 +483,7 @@ function multiObjectDelete(authInfo, request, log, callback) {
         pushMetric('multiObjectDelete', log, {
             authInfo,
             bucket: bucketName,
+            keys: successfullyDeleted,
```
electrachong: it's ok if the array item is an object containing `{ versionId, key }` instead of being an array of strings?

ploki: No it isn't. Is the ```successfullyDeleted``` array always composed of objects of this kind?

electrachong: yes, at least in any release with versioning implemented (>= 7.0)

ploki:
@ironman-machine try

ploki:
@ironman-machine try

ploki:
again @ironman-machine try

lib/api/multiObjectDelete.js
```diff
@@ -480,9 +480,12 @@ function multiObjectDelete(authInfo, request, log, callback) {
         }
         const xml = _formatXML(quietSetting, errorResults,
             successfullyDeleted);
+        const deletedKeys = [];
+        successfullyDeleted.map(item => deletedKeys.push(item.key));
```
electrachong: you can do this, no need to create a separate array outside:
`const deletedKeys = successfullyDeleted.map(item => item.key);`

electrachong:
@ironman-machine r+

electrachong:
@ironman-machine 
SCALITY_UTAPI_BRANCH=
r+

electrachong:
@ironman-machine r+

PR #896, opened 2017-09-11
ft: add endpoint for backbeat w/ docker swarm

jonathan-gramain:
Add a static endpoint named cloudserver-front, which backbeat uses to
contact the local S3 when deployed with docker swarm.


electrachong:
@ironman-machine r+

PR #897, opened 2017-09-12
FIX: Object Copy to AWS

nicolas2bert:
object copy to AWS S3

### DIFF
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -21,6 +21,7 @@ const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
 const validateWebsiteHeader = require('./apiUtils/object/websiteServing')
     .validateWebsiteHeader;
 const { config } = require('../Config');
+const multipleBackendGateway = require('../data/multipleBackendGateway');
 
 const versionIdUtils = versioning.VersionID;
 const locationHeader = constants.objectLocationConstraintHeader;
@@ -37,7 +38,10 @@ const externalVersioningErrorMessage = 'We do not currently support putting ' +
  * @param {string} objectKey - destination key name
  * @param {object} destBucketMD - bucket metadata of bucket being copied to
  * @param {object} log - logger object
- * @return {object} storeMetadataParams or an error
+ * @return {object}
+ * - (storeMetadataParams
+ * - sourceLocation {string} - location type of the source
+ * - OR error
  */
 function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
     objectKey, destBucketMD, log) {
@@ -94,9 +98,10 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
     if (whichMetadata === 'COPY' && headers[locationHeader]) {
         userMetadata[locationHeader] = headers[locationHeader];
     }
+    const sourceLocation = sourceObjMD[locationHeader];
     // If location constraint header is not included, locations match
     const locationMatch = headers[locationHeader] ?
-        sourceObjMD[locationHeader] === headers[locationHeader] : true;
+        sourceLocation === headers[locationHeader] : true;
 
     // If tagging directive is REPLACE but you don't specify any
     // tags in the request, the destination object will
@@ -144,8 +149,7 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
     if (!storeMetadataParams.contentType) {
         storeMetadataParams.contentType = sourceObjMD['content-type'];
     }
-
-    return storeMetadataParams;
+    return { storeMetadataParams, sourceLocation };
 }
 
 /**
@@ -256,12 +260,17 @@ function objectCopy(authInfo, request, sourceBucket,
                     if (headerValResult.error) {
                         return next(errors.PreconditionFailed, destBucketMD);
                     }
-                    const storeMetadataParams =
+                    const { storeMetadataParams, error: metadataError,
+                    sourceLocation } =
                         _prepMetadata(sourceObjMD, request.headers,
                             sourceIsDestination, authInfo, destObjectKey,
                             destBucketMD, log);
-                    if (storeMetadataParams.error) {
-                        return next(storeMetadataParams.error, destBucketMD);
+                    if (metadataError) {
+                        return next(metadataError, destBucketMD);
+                    }
+                    if (storeMetadataParams.metaHeaders) {
+                        dataStoreContext.metaHeaders =
+                          storeMetadataParams.metaHeaders;
                     }
                     let dataLocator;
                     // If 0 byte object just set dataLocator to empty array
@@ -284,11 +293,11 @@ function objectCopy(authInfo, request, sourceBucket,
                         }
                     }
                     return next(null, storeMetadataParams, dataLocator,
-                        destBucketMD, destObjMD);
+                        destBucketMD, destObjMD, sourceLocation);
                 });
         },
         function goGetData(storeMetadataParams, dataLocator, destBucketMD,
-            destObjMD, next) {
+            destObjMD, sourceLocation, next) {
             const serverSideEncryption = destBucketMD.getServerSideEncryption();
 
             const backendInfoObj = locationConstraintCheck(request,
@@ -349,6 +358,38 @@ function objectCopy(authInfo, request, sourceBucket,
                 }
                 return next(null, storeMetadataParams, dataLocator, destObjMD,
                     serverSideEncryption, destBucketMD);
+            }
+
+            // NOTE: using copyObject only if copying object from one external
+            // backend to the same external backend
+            const locationTypeMatch = request.headers[locationHeader] ?
+            config.locationConstraints[sourceLocation].type ===
+            config.locationConstraints[request.headers[locationHeader]].type
+            : true;
+            if (locationTypeMatch &&
+            constants.externalBackends[locationType]
+            && dataLocator.length === 1) {
+                const location = storeMetadataParams
+                .metaHeaders[locationHeader];
+                const objectGetInfo = dataLocator[0];
+                return multipleBackendGateway.copyObject(request, location,
+                sourceBucket, sourceObject, sourceLocation, log,
+                (error, objectRetrievalInfo) => {
+                    if (error) {
+                        return next(error, destBucketMD);
+                    }
+                    const putResult = {
+                        key: objectRetrievalInfo.key,
+                        dataStoreName: objectRetrievalInfo.
+                            dataStoreName,
+                        size: storeMetadataParams.size,
+                        dataStoreETag: objectGetInfo.dataStoreETag,
+                        start: objectGetInfo.start,
+                    };
+                    const putResultArr = [putResult];
+                    return next(null, storeMetadataParams, putResultArr,
+                        destObjMD, serverSideEncryption, destBucketMD);
+                });
             }
 
             // dataLocator is an array.  need to get and put all parts
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -1,9 +1,11 @@
 const { errors, s3middleware } = require('arsenal');
 const AWS = require('aws-sdk');
 const MD5Sum = s3middleware.MD5Sum;
+const getMetaHeaders = s3middleware.userMetadata.getMetaHeaders;
 const createLogger = require('../multipleBackendLogger');
 const { prepareStream } = require('../../api/apiUtils/object/prepareStream');
-const logHelper = require('./utils').logHelper;
+const { logHelper, trimXMetaPrefix } = require('./utils');
+const { config } = require('../../Config');
 
 class AwsClient {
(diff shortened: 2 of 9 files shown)


lib/api/objectCopy.js
```diff
@@ -351,6 +352,29 @@ function objectCopy(authInfo, request, sourceBucket,
                     serverSideEncryption, destBucketMD);
             }
 
+            if (storeMetadataParams.locationMatch && locationType &&
+            constants.externalBackends[locationType]
+            && dataLocator.length === 1) {
```
LaurenSpiegel: would the length ever be greater than 1? are you just checking to make sure there is any data at all?

lib/api/objectCopy.js
```diff
@@ -351,6 +352,29 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                const objectGetInfo = dataLocator[0];
+                const location = storeMetadataParams
+                .metaHeaders[locationHeader];
+                multipleBackendGateway.copyObject(request, location,
```
LaurenSpiegel: does sending the request work considering the bucketMatch option?

lib/api/objectCopy.js
nicolas2bert: Length should always be 1 for external dataLocator.

lib/data/external/AwsClient.js
```diff
@@ -355,6 +357,35 @@ class AwsClient {
             return callback();
         });
     }
+    copyObject(request, locationType, objectGetInfo, log,
+    callback) {
+        const sourceObjectKey =
+        typeof(objectGetInfo) === 'string' ? objectGetInfo :
```
electrachong: what the situations in which is is just a string and when it is an object?

lib/data/multipleBackendGateway.js
```diff
@@ -221,6 +221,22 @@ const multipleBackendGateway = {
...
+            log, (err, key) => {
+                const dataRetrievalInfo = {
+                    key,
+                    dataStoreName: locationType,
+                    dataStoreType: client.clientType,
+                };
+                cb(err, dataRetrievalInfo);
+            });
+        }
+        return cb();
```
electrachong: So if the client is not AWS and a copy object request is made, do we handle that?
`process.nextTick` ?

lib/data/external/AwsClient.js
nicolas2bert: I assume that in previous version objectGetInfo was a string representing the object's key. So we added this check for backwards compatibility. Ps: I added a comment :)

lib/api/objectCopy.js
```diff
@@ -351,6 +356,29 @@ function objectCopy(authInfo, request, sourceBucket,
                     serverSideEncryption, destBucketMD);
             }
 
+            if (storeMetadataParams.locationMatch && locationType &&
```
dora-korpar: I don't think you need to check if locationType exists because if it doesn't the next condition just won't be met, right?

lib/data/external/AwsClient.js
```diff
@@ -355,6 +357,35 @@ class AwsClient {
             return callback();
         });
     }
+    copyObject(request, locationType, objectGetInfo, log,
```
dora-korpar: this is trivial, but I think the name of the second param should be 'location' instead of 'locationType' because it will be the name of the actual location constraint instead of like aws or azure

lib/data/multipleBackendGateway.js
nicolas2bert: I added a note too.
We use `copyObject` only if copying object from one external backend to the same external backend

lib/api/objectCopy.js
```diff
@@ -351,6 +356,31 @@ function objectCopy(authInfo, request, sourceBucket,
                     serverSideEncryption, destBucketMD);
             }
 
+            // using copyObject only if copying object from one external
+            // backend to the same external backend
+            if (storeMetadataParams.locationMatch && locationType &&
```
dora-korpar: I don't think you need to check that locationType exists here because if it doesn't the next condition will just be false

lib/data/multipleBackendGateway.js
electrachong: So could someone make a request to copy Azure object to another location in Azure?
Just wondering if we need to return NotImplemented or not -- assuming we don't have a `copyObject` for the Azure backend client yet

lib/api/objectCopy.js
```diff
@@ -351,6 +356,31 @@ function objectCopy(authInfo, request, sourceBucket,
                     serverSideEncryption, destBucketMD);
             }
 
+            // using copyObject only if copying object from one external
+            // backend to the same external backend
```
electrachong: Do we check and handle if someone requests to copy object from one ext. backend to a different one? E.g. return an error?

nicolas2bert: if copy object from one ext. backend to a different one, we do a GET and then a PUT 
PS: condition here: https://github.com/scality/S3/pull/897/files#diff-5ce9bf7b97556522975c2fae1a6944efR361

lib/api/objectCopy.js
nicolas2bert: ok thanks :)

lib/data/external/AwsClient.js
nicolas2bert: you are right, thanks!

lib/data/multipleBackendGateway.js
nicolas2bert: not yet, so I added a NotImplemented error, thanks :)

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy.js
```diff
@@ -86,6 +85,10 @@ destBucket, destLoc, awsKey, mdDirective, isEmptyObj, callback) {
             assert.deepStrictEqual(sourceRes.Metadata['test-header'],
                 destRes.Metadata['test-header']);
         }
+        if (destLoc === awsLocation) {
+            // if copying to AWS  => retain metadata even though directive: COPY
```
electrachong: do you mean `even though directive: REPLACE`?
do you mean we retain all the metadata even if directive is `REPLACE` or just the location metaheader?

lib/data/external/AwsClient.js
```diff
@@ -355,6 +357,35 @@ class AwsClient {
             return callback();
         });
     }
+    copyObject(request, objectGetInfo, log, callback) {
+        // for backwards compatibility
+        const sourceObjectKey =
+        typeof(objectGetInfo) === 'string' ? objectGetInfo :
+          objectGetInfo.key;
```
LaurenSpiegel: no need for this because if it's aws to aws or azure to azure it would not be the old format

lib/data/external/utils.js
```diff
@@ -3,6 +3,14 @@ const utils = {
         log[level](description, { error: error.message, stack: error.stack,
           dataStoreName });
     },
+    removeMeta(obj) {
```
LaurenSpiegel: let's call it trimXMetaPrefix

nicolas2bert:
@ironman-machine try

lib/data/external/AwsClient.js
```diff
@@ -356,6 +358,32 @@ class AwsClient {
...
+            Bucket: this._awsBucketName,
+            Key: awsKey,
+            CopySource: `${this._awsBucketName}/${sourceObjectKey}`,
+            Metadata: metaHeaders,
+            MetadataDirective: metadataDirective,
+        }, err => {
+            if (err) {
+                logHelper(log, 'error', 'error from data backend on ' +
+                'copyObject', err, this._dataStoreName);
+                return callback(errors.InternalError
```
LaurenSpiegel: As discussed, if it's access denied we should probably pass that through so requester knows that no access on either source or destination bucket. Also, we need a test where the access key pair for the source bucket and destination bucket are different and there is an access denied.

lib/api/objectCopy.js
```diff
@@ -37,7 +38,13 @@ const externalVersioningErrorMessage = 'We do not currently support putting ' +
  * @param {string} objectKey - destination key name
  * @param {object} destBucketMD - bucket metadata of bucket being copied to
  * @param {object} log - logger object
- * @return {object} storeMetadataParams or an error
+ * @return {object}
+ * - (storeMetadataParams
+ * - locationTypeMatch - bool - true if location type of the source matches
+ *   location type of the destination, false if not.
+ * - sourceExternalBucketName - string - if external location name of the
+ *   external source bucket or containder)
```
LaurenSpiegel: container

lib/api/objectCopy.js
```diff
@@ -98,6 +105,25 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
...
 
+    const locationTypeMatch = headers[locationHeader] ?
+    config.locationConstraints[sourceObjMD[locationHeader]].type ===
+    config.locationConstraints[headers[locationHeader]].type : true;
+
+    let sourceExternalBucketName;
+    if (sourceObjMD[locationHeader] &&
+      config.locationConstraints[sourceObjMD[locationHeader]].type
+      === 'aws_s3') {
+        sourceExternalBucketName =
```
LaurenSpiegel: we should add a method to the Config class that is getExternalBucketName() so that you don't have to have all of this logic here and it will handle the different layouts of the azure and aws configs.

lib/api/objectCopy.js
```diff
@@ -351,6 +382,32 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                log, (error, objectRetrievalInfo) => {
+                    if (error) {
+                        return next(error, destBucketMD);
+                    }
+                    const putResult = {
+                        key: objectRetrievalInfo.key,
+                        dataStoreName: objectRetrievalInfo.
+                            dataStoreName,
+                        size: storeMetadataParams.size,
+                    };
```
LaurenSpiegel: don't you need start and etag like the other put results below?

lib/data/external/AwsClient.js
```diff
@@ -356,6 +358,41 @@ class AwsClient {
             return callback();
         });
     }
+    copyObject(request, sourceAwsBucketName, sourceObjectKey, log, callback) {
+        // for backwards compatibility
```
LaurenSpiegel: delete comment

lib/data/external/AwsClient.js
```diff
@@ -356,6 +358,41 @@ class AwsClient {
...
+            Bucket: this._awsBucketName,
+            Key: awsKey,
+            CopySource: `${sourceAwsBucketName}/${sourceObjectKey}`,
+            Metadata: metaHeaders,
+            MetadataDirective: metadataDirective,
+        }, err => {
+            if (err) {
+                if (err.code === 'AccessDenied') {
+                    logHelper(log, 'error', 'Unable to access ' +
+                    `${sourceAwsBucketName} AWS bucket`, err,
```
LaurenSpiegel: how do you know the access denied is due to the source rather than the destination?

lib/data/external/utils.js
```diff
@@ -3,6 +3,14 @@ const utils = {
         log[level](description, { error: error.message, stack: error.stack,
           dataStoreName });
     },
+    trimXMetaPrefix(obj) {
+        const newObj = {};
+        Object.keys(obj).forEach(key => {
+            const newKey = key.substring(11);
```
LaurenSpiegel: add comment, that you are taking of the 'x-amz-meta-'

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy.js
```diff
@@ -261,7 +278,97 @@ describeSkipIfNotMultiple('MultipleBackend object copy', function testSuite() {
...
+                    });
+                },
+                (key, next) =>
+                assertGetObjects(key, bucket, awsLocation, copyKey,
+                  bucket, awsLocation2, copyKey, 'REPLACE', false,
+                  awsS3Two, awsLocation2, next),
+            ], done);
+        });
+
+        it('should return error AccessDenied copying an object on AWS to a ' +
```
LaurenSpiegel: As discussed, please add a note in the docs that one account can't copy to another account with a source and destination on real AWS unless the account associated with the access Key/secret Key pairs used for the source bucket has rights to put in the destination bucket. ACL's would have to be updated on AWS directly to enable this.

lib/api/objectCopy.js
nicolas2bert: I put this check directly to external backend. It's simpler and clearer.

lib/data/external/AwsClient.js
nicolas2bert: because the destination's AWS account copies the source.

nicolas2bert:
@ironman-machine ok

nicolas2bert:
@ironman-machine try

electrachong:
@ironman-machine r+

nicolas2bert:
@ironman-machine r+

nicolas2bert:
@ironman-machine r+

PR #900, opened 2017-09-15
ft: azure list parts

dora-korpar:
Adds listParts API for Azure data backend. No calls actually need to be made to Azure since the part metadata is stored in Scality S3, so only tests included.


dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine r+

PR #905, opened 2017-09-20
Ft/azure completempu

dora-korpar:
Adds complete MPU API for Azure data backend.

Also fixes implementation of delete for Azure. Now, if a user wants to delete an object from Azure, but there is an ongoing MPU with the same key name, an MPUinProgress error will be returned. This is because on a delete, Azure will delete all objects and parts associated with a key name.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -105,9 +105,12 @@ const constants = {
 
     // user metadata header to set object locationConstraint
     objectLocationConstraintHeader: 'x-amz-meta-scal-location-constraint',
-    // eslint-disable-next-line camelcase
+    /* eslint-disable camelcase */
     externalBackends: { aws_s3: true, azure: true },
     versioningNotImplBackends: { azure: true },
+    mpuMDStoredExternallyBackend: { aws_s3: true },
+    /* eslint-enable camelcase */
+    mpuMDStoredOnS3Backend: { azure: true },
 };
 
 module.exports = constants;
--- a/lib/api/apiUtils/object/processMpuParts.js
+++ b/lib/api/apiUtils/object/processMpuParts.js
@@ -1,0 +1,241 @@
+const { errors } = require('arsenal');
+const crypto = require('crypto');
+const constants = require('../../../../constants');
+
+/**
+ * createAggregateETag - creates ETag from concatenated MPU part ETags to
+ * mimic AWS
+ * @param {string} concatETags - string of concatenated MPU part ETags
+ * @param {array} partList - list of parts to complete MPU with
+ * @return {string} aggregateETag - final complete MPU obj ETag
+ */
+function createAggregateETag(concatETags, partList) {
+    // AWS documentation is unclear on what the MD5 is that it returns
+    // in the response for a complete multipart upload request.
+    // The docs state that they might or might not
+    // return the MD5 of the complete object. It appears
+    // they are returning the MD5 of the parts' MD5s so that is
+    // what we have done here. We:
+    // 1) concatenate the hex version of the
+    // individual ETags
+    // 2) convert the concatenated hex to binary
+    // 3) take the md5 of the binary
+    // 4) create the hex digest of the md5
+    // 5) add '-' plus the number of parts at the end
+
+    // Convert the concatenated hex ETags to binary
+    const bufferedHex = Buffer.from(concatETags, 'hex');
+    // Convert the buffer to a binary string
+    const binaryString = bufferedHex.toString('binary');
+    // Get the md5 of the binary string
+    const md5Hash = crypto.createHash('md5');
+    md5Hash.update(binaryString, 'binary');
+    // Get the hex digest of the md5
+    let aggregateETag = md5Hash.digest('hex');
+    // Add the number of parts at the end
+    aggregateETag = `${aggregateETag}-${partList.length}`;
+
+    return aggregateETag;
+}
+
+/**
+ * generateMpuPartStorageInfo - generates info needed for storage of
+ * completed MPU object
+ * @param {array} filteredPartList - list of parts filtered from metadata
+ * @return {object} partsInfo - contains three keys: aggregateETag,
+ * dataLocations, and calculatedSize
+ */
+function generateMpuPartStorageInfo(filteredPartList) {
+    // Assemble array of part locations, aggregate size
+    // and build string to create aggregate ETag
+    let calculatedSize = 0;
+    const dataLocations = [];
+    let concatETags = '';
+    const partsInfo = {};
+
+    filteredPartList.forEach((storedPart, index) => {
+        const partETagWithoutQuotes =
+            storedPart.ETag.slice(1, -1);
+        const dataStoreETag = `${index + 1}:${partETagWithoutQuotes}`;
+        concatETags += partETagWithoutQuotes;
+
+        // If part was put by a regular put part rather than a
+        // copy it is always one location.  With a put part
+        // copy, could be multiple locations so loop over array
+        // of locations.
+        for (let j = 0; j < storedPart.locations.length; j++) {
+            // If the piece has parts (was a put part object
+            // copy) each piece will have a size attribute.
+            // Otherwise, the piece was put by a regular put
+            // part and the size the of the piece is the full
+            // part size.
+            const location = storedPart.locations[j];
+            // If there is no location, move on
+            if (!location || typeof location !== 'object') {
+                continue;
+            }
+            let pieceSize = Number.parseInt(storedPart.size, 10);
+            if (location.size) {
+                pieceSize = Number.parseInt(location.size, 10);
+            }
+            const pieceRetrievalInfo = {
+                key: location.key,
+                size: pieceSize,
+                start: calculatedSize,
+                dataStoreName: location.dataStoreName,
+                dataStoreETag,
+                cryptoScheme: location.sseCryptoScheme,
+                cipheredDataKey: location.sseCipheredDataKey,
+            };
+            dataLocations.push(pieceRetrievalInfo);
+            // eslint-disable-next-line no-param-reassign
+            calculatedSize += pieceSize;
+        }
+    });
+
+    partsInfo.aggregateETag =
+        createAggregateETag(concatETags, filteredPartList);
+    partsInfo.dataLocations = dataLocations;
+    partsInfo.calculatedSize = calculatedSize;
+    return partsInfo;
+}
+
+/**
+ * validateAndFilterMpuParts - validates part list sent by user and filters
+ * parts stored in metadata against user part list
+ * @param {array} storedParts - array of parts stored in metadata
+ * @param {array} jsonList - array of parts sent by user for completion
+ * @param {string} mpuOverviewKey - metadata mpu key
+ * @param {string} splitter - mpu key divider
+ * @param {object} log - Werelogs instance
+ * @return {object} filtersPartsObj - contains 3 keys: partList, keysToDelete,
+ * and extraPartLocations
+ */
+function validateAndFilterMpuParts(storedParts, jsonList, mpuOverviewKey,
+splitter, log) {
+    let storedPartsCopy = [];
+    const filteredPartsObj = {};
+    filteredPartsObj.partList = [];
+
+    const keysToDelete = [];
+    storedParts.forEach(item => {
+        keysToDelete.push(item.key);
+        storedPartsCopy.push({
+            // In order to delete the part listing in the shadow
+            // bucket, need the full key
+            key: item.key,
+            ETag: `"${item.value.ETag}"`,
+            size: item.value.Size,
+            locations: Array.isArray(item.value.partLocations) ?
+                item.value.partLocations : [item.value.partLocations],
+        });
+    });
+    keysToDelete.push(mpuOverviewKey);
+
+    // Check list sent to make sure valid
+    const partLength = jsonList.Part.length;
+    // A user can put more parts than they end up including
+    // in the completed MPU but there cannot be more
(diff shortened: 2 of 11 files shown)


lib/api/completeMultipartUpload.js
```diff
@@ -153,235 +154,79 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                extraParts.forEach(part => {
-                    const locations = part.locations;
-                    locations.forEach(location => {
-                        if (!location || typeof location !== 'object') {
-                            return;
-                        }
-                        extraPartLocations.push(location);
-                    });
-                });
+            if (completeObjData && completeObjData.calculateEtag) {
```
LaurenSpiegel: why? what does this mean? instead of using this condition and the condition above, let's use one of your nifty helper functions to be more expressive about we are trying to determine here.

lib/api/completeMultipartUpload.js
```diff
@@ -467,19 +311,23 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                     // in cases where completing mpu overwrites a previous
                     // null version when versioning is suspended or versioning
                     // is not enabled, need to delete pre-existing data
+                    // unless the preexisting object and the completed mpu
+                    // are on Azure data backend
```
LaurenSpiegel: because we don't allow object puts/initiate mpu to azure when a bucket has versioning enabled? if the possibility is blocked, is this needed? if it's not blocked, shouldn't it be until we are able to have azure versioning on versioning?

lib/data/external/utils.js
```diff
@@ -15,6 +20,67 @@ const utils = {
     removeQuotes(word) {
         return word.slice(1, -1);
     },
+    // Before putting or deleting object on Azure, check if MPU exists with
```
LaurenSpiegel: are we covered in initiate mpu and object copy as well? i think we are by virtue of blocking the data putting but just want to confirm.

lib/api/completeMultipartUpload.js
```diff
@@ -153,235 +154,79 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                jsonList, storedMetadata, null, mpuOverviewKey);
         },
         function processParts(destBucket, objMD, mpuBucket, storedParts,
             jsonList, storedMetadata, completeObjData, mpuOverviewKey, next) {
-            let calculatedSize = 0;
-            if (completeObjData) {
+            // means MPU completion was done on third-party backend,
+            // and Etag was returned, so part info is not in our metadata
+            //  and related steps should be skipped
+            if (completeObjData && !completeObjData.calculateEtag) {
```
nicolas2bert: Your comment `Etag was returned` and then you check `!completeObjData.calculateEtag`. AWS should not be completeObjData.calculateEtag equals true?

lib/data/external/utils.js
```diff
@@ -15,6 +20,67 @@ const utils = {
...
+            // partLocations will always be an array of length 1
+            const partInfo = part.value.partLocations[0];
+            if (partNumbersToCommit.indexOf(partInfo.key) > -1) {
+                partInfo.partNumber = partInfo.key;
+                partsToCommit.push(partInfo);
+            }
+        });
+        return partsToCommit;
+    },
+    checkAzureBackendMatch(dataToDelete, completeObjData) {
```
LaurenSpiegel: please add space and jsdoc to clarify the purpose of this and the params

lib/data/external/utils.js
```diff
@@ -15,6 +20,67 @@ const utils = {
...
+        const partNumbersToCommit = [];
+        userPartList.forEach(part => {
+            partNumbersToCommit.push(parseInt(part.PartNumber[0], 10));
+        });
+
+        const partsToCommit = [];
+        storedParts.forEach(part => {
+            // partLocations will always be an array of length 1
+            const partInfo = part.value.partLocations[0];
+            if (partNumbersToCommit.indexOf(partInfo.key) > -1) {
```
LaurenSpiegel: isn't the partNumbersToCommit array an array of just numbers?

lib/services.js
```diff
@@ -241,22 +242,39 @@ const services = {
...
+                        return data.delete(objectMD.location, deleteLog);
+                    }
+                    return data.batchDelete(objectMD.location, null, null,
+                    deleteLog);
+                });
+        }
+
+        const objGetInfo = objectMD.location;
+        if (objGetInfo && objGetInfo[0].dataStoreType === 'azure') {
+            return externalBackendUtils.protectAzureBlocks(bucketName,
```
LaurenSpiegel: is there a way to keep the protection code in the azure client so we don't have to spread the logic around?

lib/api/completeMultipartUpload.js
```diff
@@ -153,235 +154,79 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                        extraPartLocations.push(location);
-                    });
-                });
+            if (completeObjData && completeObjData.calculateEtag) {
+                const dataLocations = [
+                    {
+                        key: completeObjData.key,
+                        size: calculatedSize,
+                        start: 0,
+                        dataStoreName: storedMetadata.dataStoreName,
```
nicolas2bert: Where do you get `storedMetadata.dataStoreName` from?

lib/services.js
dora-korpar: the function is in a util file in the external folder, and I put it there cos we don't really call any of the external clients without going through the gateway, so I thought that was a good compromise. I can move it if you'd still prefer

lib/data/external/utils.js
dora-korpar: yes, and the data that's stored as 'key' in our metadata is actually also the partNumber

tests/functional/aws-node-sdk/test/multipleBackend/mpuComplete/azureCompleteMPU.js
```diff
@@ -0,0 +1,228 @@
...
+
+const azureMpuUtils = s3middleware.azureHelper.mpuUtils;
+const describeSkipIfNotMultiple = (config.backends.data !== 'multiple'
+    || process.env.S3_END_TO_END) ? describe.skip : describe;
+
+const awsLocation = 'aws-test';
+const awsBucket = 'multitester555';
+const awsConfig = getRealAwsConfig(awsLocation);
+const awsS3 = new AWS.S3(awsConfig);
+const azureLocation = 'azuretest';
```
nicolas2bert: Can you add test with bucketMatch equals false, please?

lib/api/completeMultipartUpload.js
dora-korpar: no, this dataToDelete is only done when versioning is *not* enabled. since we don't have versioning on Azure, this call to batchDelete would delete the completed MPU object on Azure. We don't encounter this on AWS because versioning is always enabled.

lib/data/external/AzureClient.js
```diff
@@ -162,14 +171,44 @@ class AzureClient {
...
+            if (err) {
+                logHelper(log, 'error', 'err completing MPU on Azure datastore',
+                    err, this._dataStoreName);
+                return callback(errors.InternalError
+                    .customizeDescription('Error returned from ' +
+                    `Azure: ${err.message}`));
+            }
+            const completeObjData = {
+                key: azureKey,
+                calculateEtag: true,
```
nicolas2bert: `calculateEtag: false,` ?

lib/api/completeMultipartUpload.js
dora-korpar: It's stored in the metadata

lib/data/external/utils.js
LaurenSpiegel: Instead of doing a filter for just azure here, let's do a part validation function early on in waterfall. Can skip for aws based on not having metadata parts.

lib/api/apiUtils/object/processMpuParts.js
```diff
@@ -0,0 +1,206 @@
...
+        // in ascending order return an error
+        const partNumber = Number.parseInt(part.PartNumber[0], 10);
+        if (i > 0) {
+            const previousPartNumber =
+                Number.parseInt(jsonList.Part[i - 1].PartNumber[0], 10);
+            if (partNumber <= previousPartNumber) {
+                partsInfo.error = errors.InvalidPartOrder;
+                return partsInfo;
+            }
+        }
```
LaurenSpiegel: let's move this validation to your general part validation function.

constants.js
```diff
@@ -105,8 +105,11 @@ const constants = {
 
     // user metadata header to set object locationConstraint
     objectLocationConstraintHeader: 'x-amz-meta-scal-location-constraint',
-    // eslint-disable-next-line camelcase
+    /* eslint-disable camelcase */
     externalBackends: { aws_s3: true, azure: true },
+    mpuMDStoredExternallyBackend: { aws_s3: true },
+    /* eslint-enable camelcase */
+    mpuMDStoredOnS3Backend: { azure: true },
```
LaurenSpiegel: isn't this true for everything other than aws?

lib/services.js
LaurenSpiegel: I meant, when an action is called in the azure client itself, can we put the protection check there?  that way, anyone extending will not have to worry about calling protectAzureBlocks outside of the azure client.

lib/api/apiUtils/object/processMpuParts.js
```diff
@@ -2,7 +2,20 @@ const { errors } = require('arsenal');
 const crypto = require('crypto');
 const constants = require('../../../../constants');
 
-function createAggregateETag(concatETags, jsonList) {
+function createAggregateETag(concatETags, partList) {
```
LaurenSpiegel: please add jsdoc

lib/api/apiUtils/object/processMpuParts.js
```diff
@@ -13,173 +26,172 @@ function createAggregateETag(concatETags, jsonList) {
...
     // Add the number of parts at the end
-    aggregateETag = `${aggregateETag}-${jsonList.Part.length}`;
+    aggregateETag = `${aggregateETag}-${partList.length}`;
 
     return aggregateETag;
 }
 
-function processMpuParts(destBucket, storedParts, jsonList, mpuOverviewKey,
-splitter, log) {
+function generateMpuPartStorageInfo(filteredPartList) {
```
LaurenSpiegel: jsdoc here as well

lib/data/external/AzureClient.js
```diff
@@ -56,11 +58,47 @@ class AzureClient {
...
+            maxKeys: 1,
+        };
+
+        return metadata.listMultipartUploads(mpuBucketName, listingParams,
+        log, (err, mpuList) => {
+            if (err && !err.NoSuchBucket) {
+                log.error('Error listing MPUs for Azure delete',
+                    { error: err, dataStoreName });
+                return cb(errors.InternalError
+                    .customizeDescription('Error deleting object from ' +
```
LaurenSpiegel: can be used for delete and put so error message should reflect that

lib/data/multipleBackendGateway.js
```diff
@@ -273,6 +273,14 @@ const multipleBackendGateway = {
           'Can not copy object from ' +
           `${client.clientType} to ${client.clientType}`));
     },
+    protectAzureBlocks: (bucketName, objectKey, location, log, cb) => {
```
LaurenSpiegel: let's keep this in the azureclient itself and call it at the beginning of put and delete in the client itself.

LaurenSpiegel: as discussed, this is a sad state of affairs but we need it outside to avoid orphans.

lib/api/completeMultipartUpload.js
```diff
@@ -153,235 +156,97 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+            // if mpu was completed on backend that stored mpu MD externally,
+            // skip MD processing steps
+            if (completeObjData && skipMpuPartProcessing(completeObjData)) {
                 const dataLocations = [
                     {
                         key: completeObjData.key,
                         size: completeObjData.contentLength,
-                        start: calculatedSize,
+                        start: 0,
                         dataStoreName: storedMetadata.dataStoreName,
```
nicolas2bert: you could have add `dataStoreType` key  to be consistent with the locations schema.

tests/functional/aws-node-sdk/test/multipleBackend/mpuComplete/azureCompleteMPU.js
```diff
@@ -0,0 +1,250 @@
...
+            (uploadId, partArray) => {
+                const params = {
+                    Bucket: azureContainerName,
+                    Key: this.test.key,
+                    UploadId: uploadId,
+                    MultipartUpload: { Parts: partArray },
+                };
+                s3.completeMultipartUpload(params, err => {
+                    assert.equal(err, null, `Err completing MPU: ${err}`);
+                    setTimeout(() => getCheck(this.test.key, done),
```
nicolas2bert: Is the test failing here? 
On the Azure side, your blob should not be inside `azureContainerName` container with blob name equals to `azureContainerName/this.test.key`?

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine r+

dora-korpar:
@ironman-machine try

PR #906, opened 2017-09-20
FIX: AWS put copy part

nicolas2bert:


lib/data/external/AwsClient.js
```diff
@@ -402,6 +402,52 @@ class AwsClient {
...
     }
+    uploadPartCopy(request, awsSourceKey, sourceLocationConstraintName,
+      log, callback) {
+        const destBucketName = request.bucketName;
+        const destObjectKey = request.objectKey;
+        const destAwsKey = this._createAwsKey(destBucketName, destObjectKey,
+        this._bucketMatch);
+
+        const sourceAwsBucketName =
+        config.locationConstraints[sourceLocationConstraintName]
```
LaurenSpiegel: we should have a method in Config.js that handles this.  we shouldn't be digging this deep into the object outside of the config.

lib/data/external/AwsClient.js
```diff
@@ -402,6 +402,52 @@ class AwsClient {
...
+            Bucket: this._awsBucketName,
+            CopySource: `${sourceAwsBucketName}/${awsSourceKey}`,
+            CopySourceRange: copySourceRange,
+            Key: destAwsKey,
+            PartNumber: partNumber,
+            UploadId: uploadId,
+        };
+        return this._client.uploadPartCopy(params, (err, res) => {
+            if (err) {
+                console.log('err!!!', err);
```
LaurenSpiegel: clean up

tests/multipleBackend/objectPutCopyPart.js
```diff
@@ -234,6 +235,49 @@ function testSuite() {
...
+                s3.abortMultipartUpload(awsReq, err => {
+                    assert.equal(err, null, `Error aborting MPU: ${err}. ` +
+                    `You must abort MPU with upload ID ${uploadId} manually.`);
+                    done();
+                });
+            });
+        });
+    });
+
+    it.only('should copy part an object on AWS to a mpu with a different ' +
```
LaurenSpiegel: clean up

lib/data/external/AwsClient.js
nicolas2bert: Totally agree!

lib/Config.js
```diff
@@ -729,6 +729,13 @@ class Config extends EventEmitter {
         this.authData.accounts = accounts;
         this.emit('authdata-update');
     }
+
+    getAwsBucketName(locationConstraint) {
+        return this.locationConstraints[locationConstraint].details.bucketName;
+    }
+    isBucketMatch(locationConstraint) {
+        return this.locationConstraints[locationConstraint].details.bucketMatch;
```
LaurenSpiegel: this function and the one above could throw if the locationConstraint is unknown or doesn't have a details section. are we ok with that? should we have a check first and return undefined for the bucketname and the bucketMatch if not known?

nicolas2bert: `locationConstraint` comes from the locationConstraint metadata of the source object.
Before storing the object source locationConstraint, we check the object location constraint so we don't save an unknown location.
here: lib/api/apiUtils/object/locationConstraintCheck.js
Also, in our `Config.json` we make sure that `locationConstraints[l].details` is an object
So I don't think we need to check that the location is known and if `bucketName` or `BucketMatch` are not specified in our `locationConfig.json` it will be undefined (no error will be thrown) .

LaurenSpiegel: well, if we have corrupted metadata we could throw but that would be a good wake up call to a problem.

tests/locationConfigTests.json
```diff
@@ -66,6 +66,16 @@
             "credentialsProfile": "default"
         }
     },
+    "aws-test-mismatch": {
```
dora-korpar: you already have two regions called 'aws-test-mismatch'

dora-korpar: There are already two regions with the same name in the config, plz remove this and one of them :)

nicolas2bert: Thanks, I removed it.

nicolas2bert:
@ironman-machine r+

PR #907, opened 2017-09-20
FIX: COPY-AWS: using datastore key

nicolas2bert:
Do not need to re-create the source key name if we get it directly from metadata.location info :)

### DIFF
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -359,26 +359,23 @@ class AwsClient {
             return callback();
         });
     }
-    copyObject(request, sourceBucketName, sourceObjectKey, sourceLocation,
-    log, callback) {
+    copyObject(request, sourceKey, sourceLocationConstraintName, log,
+    callback) {
         const destBucketName = request.bucketName;
         const destObjectKey = request.objectKey;
         const destAwsKey = this._createAwsKey(destBucketName, destObjectKey,
           this._bucketMatch);
 
-        const sourceBucketMatch =
-        config.locationConstraints[sourceLocation].details.bucketMatch;
         const sourceAwsBucketName =
-        config.locationConstraints[sourceLocation].details.bucketName;
-        const sourceAwsKey = this._createAwsKey(sourceBucketName,
-          sourceObjectKey, sourceBucketMatch);
+        config.locationConstraints[sourceLocationConstraintName]
+        .details.bucketName;
 
         const metadataDirective = request.headers['x-amz-metadata-directive'];
         const metaHeaders = trimXMetaPrefix(getMetaHeaders(request.headers));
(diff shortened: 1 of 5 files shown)


nicolas2bert:
@ironman-machine try

lib/data/external/AwsClient.js
```diff
@@ -359,26 +359,21 @@ class AwsClient {
             return callback();
         });
     }
-    copyObject(request, sourceBucketName, sourceObjectKey, sourceLocation,
-    log, callback) {
+    copyObject(request, sourceKey, sourceLocation, log, callback) {
```
LaurenSpiegel: what's the difference between sourceKey and sourceLocation?

lib/data/external/AwsClient.js
```diff
@@ -359,26 +359,21 @@ class AwsClient {
...
         const destBucketName = request.bucketName;
         const destObjectKey = request.objectKey;
         const destAwsKey = this._createAwsKey(destBucketName, destObjectKey,
           this._bucketMatch);
 
-        const sourceBucketMatch =
-        config.locationConstraints[sourceLocation].details.bucketMatch;
```
LaurenSpiegel: do you not need this because the key already has the info?

lib/data/multipleBackendGateway.js
```diff
@@ -228,12 +228,11 @@ const multipleBackendGateway = {
...
     // backend to the same external backend
-    copyObject: (request, location, sourceBucket, sourceObject,
-    sourceLocation, log, cb) => {
+    copyObject: (request, location, sourceKey, sourceLocation, log, cb) => {
         const client = clients[location];
         if (client.copyObject) {
-            return client.copyObject(request, sourceBucket, sourceObject,
-            sourceLocation, log, (err, key) => {
+            return client.copyObject(request, sourceKey, sourceLocation, log,
+            (err, key) => {
```
LaurenSpiegel: add a test please

lib/data/external/AwsClient.js
nicolas2bert: yes :)

lib/data/external/AwsClient.js
nicolas2bert: `sourceKey` is the AWS source object name, Maybe I should call it `awsSourceKey`
`sourceLocation` is the location of the source object

LaurenSpiegel: I still am not clear on the difference.

nicolas2bert: In this case: `awsSourceKey` === `bucketname/objectname` and `sourceLocation` === `aws-test`

nicolas2bert: locationConstraintName

PR #909, opened 2017-09-22
FX: MPU  with bucketMatch set to false

nicolas2bert:


tests/locationConfigTests.json
```diff
@@ -56,6 +56,16 @@
             "credentialsProfile": "default"
         }
     },
+    "aws-test-mismatch": {
+        "type": "aws_s3",
+        "legacyAwsBehavior": true,
+        "details": {
+            "awsEndpoint": "s3.amazonaws.com",
+            "bucketName": "multitester555",
```
LaurenSpiegel: please make sure versioning is enabled on this so that healthchecks pass

tests/multipleBackend/objectPutPart.js
```diff
@@ -109,15 +110,12 @@ errorDescription) {
...
-                'x-amz-meta-scal-location-constraint': `${partLoc}`,
-            };
-        }
         const partReq = new DummyRequest(partReqParams, body1);
         return objectPutPart(authInfo, partReq, undefined, log, err => {
             assert.strictEqual(err, null);
-            if (bucketLoc !== awsLocation && mpuLoc !== awsLocation) {
+            if (bucketLoc !== awsLocation && mpuLoc !== awsLocation &&
+            bucketLoc !== awsLocationMismatch &&
+            mpuLoc !== awsLocationMismatch) {
```
LaurenSpiegel: what happens if it is an awsLocation/awsLocationMisMatch?

lib/api/initiateMultipartUpload.js
```diff
@@ -201,7 +201,7 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
         let uploadId;
         if (config.backends.data === 'multiple') {
             return multipleBackendGateway.createMPU(objectKey, metaHeaders,
-            destinationBucket, websiteRedirectHeader, locConstraint, log,
+            bucketName, websiteRedirectHeader, locConstraint, log,
```
LaurenSpiegel: why the changes from destBucket to bucketName throughout?  they were called destination bucket since the objects live in some middle-world as mpu's until they are completed.

dora-korpar:
I started adding some bucketMatch tests earlier, you should add the one for delete: https://github.com/scality/S3/blob/ft/add-mismatch-tests/tests/functional/aws-node-sdk/test/multipleBackend/delete/delete.js

tests/locationConfigTests.json
nicolas2bert: since it's the same account as aws-test one, it should be fine :)

tests/multipleBackend/objectPutPart.js
nicolas2bert: We do not save the metadata of the object part.

lib/api/initiateMultipartUpload.js
nicolas2bert: `destBucket` is an object representing the metadata of the bucket.
`bucketName` is the actual name of the bucket.

nicolas2bert:
@dora-korpar, this PR is related to MPU. You are right, we will need more tests for put/get/delete.
I will also have to add tests for objectCopy/ copypart.

nicolas2bert:
@ironman-machine r+

PR #913, opened 2017-09-27
ft: enable versioning put/get for aws backend

electrachong:
Reviews appreciated!

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -14,10 +14,10 @@ const getReplicationInfo = require('./getReplicationInfo');
 const { config } = require('../../../Config');
 const validateWebsiteHeader = require('./websiteServing')
     .validateWebsiteHeader;
-const externalBackends = constants.externalBackends;
+const { externalBackends, versioningNotImplBackends } = constants;
 
 const externalVersioningErrorMessage = 'We do not currently support putting ' +
-'a versioned object to a location-constraint of type AWS or Azure.';
+'a versioned object to a location-constraint of type Azure.';
 
 
 function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
@@ -129,9 +129,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
     const locationType = config.getLocationConstraintType(location);
     metadataStoreParams.dataStoreName = location;
 
-    // NOTE: remove the following when we will support putting a
-    // versioned object to a location-constraint of type AWS or Azure.
-    if (locationType && externalBackends[locationType]) {
+    if (versioningNotImplBackends[locationType]) {
         const vcfg = bucketMD.getVersioningConfiguration();
         const isVersionedObj = vcfg && vcfg.Status === 'Enabled';
 
@@ -167,13 +165,14 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             // So that data retrieval information for MPU's and
             // regular puts are stored in the same data structure,
             // place the retrieval info here into a single element array
-            const { key, dataStoreName, dataStoreType, dataStoreETag } =
-                dataGetInfo;
+            const { key, dataStoreName, dataStoreType, dataStoreETag,
+                dataStoreVersionId } = dataGetInfo;
             const prefixedDataStoreETag = dataStoreETag
                       ? `1:${dataStoreETag}`
                       : `1:${calculatedHash}`;
             const dataGetInfoArr = [{ key, size, start: 0, dataStoreName,
-                dataStoreType, dataStoreETag: prefixedDataStoreETag }];
+                dataStoreType, dataStoreETag: prefixedDataStoreETag,
+                dataStoreVersionId }];
             if (cipherBundle) {
                 dataGetInfoArr[0].cryptoScheme = cipherBundle.cryptoScheme;
                 dataGetInfoArr[0].cipheredDataKey =
--- a/lib/api/objectGet.js
+++ b/lib/api/objectGet.js
@@ -1,7 +1,7 @@
 const { errors, s3middleware } = require('arsenal');
-const validateHeaders = s3middleware.validateConditionalHeaders;
-
 const { parseRange } = require('arsenal/lib/network/http/utils');
+
+const multipleBackendGateway = require('../data/multipleBackendGateway');
 
 const { decodeVersionId } = require('./apiUtils/object/versioning');
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
@@ -10,6 +10,8 @@ const { pushMetric } = require('../utapi/utilities');
 const { getVersionIdResHeader } = require('./apiUtils/object/versioning');
 const setPartRanges = require('./apiUtils/object/setPartRanges');
 const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
+
+const validateHeaders = s3middleware.validateConditionalHeaders;
 
 /**
  * GET Object - Get an object
@@ -113,8 +115,7 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
                 objMD.location : [{ key: objMD.location }];
             // if the data backend is azure, there will only ever be at
             // most one item in the dataLocator array
-            if (dataLocator && dataLocator[0] &&
-              dataLocator[0].dataStoreType === 'azure') {
+            if (dataLocator[0] && dataLocator[0].dataStoreType === 'azure') {
                 dataLocator[0].azureStreamingOptions = streamingParams;
             }
 
@@ -187,12 +188,22 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
                 dataLocator = setPartRanges(dataLocator, byteRange);
             }
         }
-        pushMetric('getObject', log, {
-            authInfo,
-            bucket: bucketName,
-            newByteLength: responseMetaHeaders['Content-Length'],
+        return multipleBackendGateway.head(dataLocator, log.getSerializedUids(),
+        err => {
+            if (err) {
+                log.error('error from external backend checking for ' +
+                'object existence', { error: err });
+                return callback(err);
+            }
+            pushMetric('getObject', log, {
+                authInfo,
+                bucket: bucketName,
+                newByteLength:
+                responseMetaHeaders['Content-Length'],
+            });
+            return callback(null, dataLocator, responseMetaHeaders,
+                byteRange);
         });
-        return callback(null, dataLocator, responseMetaHeaders, byteRange);
     });
 }
 
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -24,11 +24,11 @@ class AwsClient {
         }
         return `${requestBucketName}/${requestObjectKey}`;
     }
-
     put(stream, size, keyContext, reqUids, callback) {
         const awsKey = this._createAwsKey(keyContext.bucketName,
            keyContext.objectKey, this._bucketMatch);
         const metaHeaders = trimXMetaPrefix(keyContext.metaHeaders);
+        const log = createLogger(reqUids);
         const uploadParams = {
             Bucket: this._awsBucketName,
             Key: awsKey,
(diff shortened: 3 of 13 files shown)


tests/functional/aws-node-sdk/test/multipleBackend/get/getAwsVersioning.js
```diff
@@ -0,0 +1,290 @@
...
+const bucket = 'buckettestmultiplebackendgetawsversioning';
+
+function getAndAssertResult(s3, params, cb) {
+    const { bucket, key, body, versionId, expectVersionId } = params;
+    s3.getObject({ Bucket: bucket, Key: key, VersionId: versionId },
+        (err, data) => {
+            assert.strictEqual(err, null, 'Expected success ' +
+                `getting object, got error ${err}`);
+            if (body) {
+                // NOTE: if we have to be concerned about kms encrpytion,
```
electrachong: disregard

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -130,8 +130,8 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
     metadataStoreParams.dataStoreName = location;
 
     // NOTE: remove the following when we will support putting a
-    // versioned object to a location-constraint of type AWS or Azure.
-    if (locationType && externalBackends[locationType]) {
+    // versioned object to a location-constraint of type Azure.
+    if (locationType === 'azure') {
```
ploki: I know it was already wrong before this PR, but this code must ask the data backend directly to know if it supports versioning. this should be considered a different issue

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -167,13 +167,14 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             // So that data retrieval information for MPU's and
             // regular puts are stored in the same data structure,
             // place the retrieval info here into a single element array
-            const { key, dataStoreName, dataStoreType, dataStoreETag } =
-                dataGetInfo;
+            const { key, dataStoreName, dataStoreType, dataStoreETag,
+                awsVersionId } = dataGetInfo;
```
ploki: It must be ```dataStoreVersionId```. ```awsVersionId``` is too specific

tests/functional/aws-node-sdk/test/multipleBackend/utils.js
```diff
@@ -99,4 +104,48 @@ utils.expectedETag = (body, getStringified = true) => {
...
+
+utils.mapToAwsPuts = (s3, bucket, key, dataArray, cb) => {
+    async.mapSeries(dataArray, (data, next) => {
+        s3.putObject({ Bucket: bucket, Key: key, Body: data,
+        Metadata: { 'scal-location-constraint': awsLocation } },
+            next);
+    }, (err, results) => {
+        assert.strictEqual(err, null, 'Expected success ' +
+            `putting object, got error ${err}`);
+        // NOTE: do we want to return just version ids or object with version
```
LaurenSpiegel: ?

lib/api/objectGet.js
```diff
@@ -117,6 +119,14 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
               dataLocator[0].dataStoreType === 'azure') {
                 dataLocator[0].azureStreamingOptions = streamingParams;
             }
+            if (dataLocator && dataLocator[0] &&
```
electrachong: @dora-korpar @nicolas2bert is it possible object stored in aws backend will have multiple items in the dataLocator array? :S

nicolas2bert: @electrachong no, only one: `dataLocator[0]`

electrachong: 👍 thanks!!

lib/api/apiUtils/object/createAndStoreObject.js
electrachong: 👍 updated

lib/data/external/AwsClient.js
```diff
@@ -53,13 +53,14 @@ class AwsClient {
                       `AWS: ${err.message}`)
                     );
                 }
-                return callback(null, awsKey);
+                const params = { awsVersionId: data.VersionId };
```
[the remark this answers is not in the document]
electrachong: updated to check

lib/data/external/AwsClient.js
```diff
@@ -69,17 +70,21 @@ class AwsClient {
...
+               const params = { awsVersionId: data.VersionId };
+               return callback(null, awsKey, params);
            });
     }
     get(objectGetInfo, range, reqUids, callback) {
         const log = createLogger(reqUids);
         // for backwards compatibility
         const key = typeof objectGetInfo === 'string' ? objectGetInfo :
           objectGetInfo.key;
+        const versionId = typeof objectGetInfo === 'string' ? undefined :
```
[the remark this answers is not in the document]
electrachong: updated

tests/functional/aws-node-sdk/test/multipleBackend/get/getAwsVersioning.js
```diff
@@ -0,0 +1,290 @@
...
+                // may need to assert ETag instead of body; but I assume
+                // the ETag is stored by our calculation of data, not from
+                // AWS servers
+                assert.strictEqual(data.Body.toString(), body);
```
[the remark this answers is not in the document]
electrachong: updated

tests/functional/aws-node-sdk/test/multipleBackend/get/getAwsVersioning.js
```diff
@@ -0,0 +1,290 @@
...
+            }
+            assert.strictEqual(data.VersionId, expectVersionId || undefined);
```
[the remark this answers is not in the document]
electrachong: for whatever reason I set the value of `expectVersionId` to false when we weren't expecting any value.
I updated the code to be more explicit

tests/functional/aws-node-sdk/test/multipleBackend/utils.js
electrachong: outdated comment, I removed it

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

lib/api/apiUtils/object/createAndStoreObject.js
LaurenSpiegel: instead of saying azure, let's have some sort of constant object for backends with unsupported versioning with azure set to true.

lib/api/objectGet.js
LaurenSpiegel: why are we doing this for just aws and why in the objectGet api versus in the aws client itself? I thought the point was to prevent us from starting to stream data and then dropping the connection once we realize there is no data. It seems it would be cleaner to put this in the get calls of the azure and aws clients.

lib/api/objectGet.js
```diff
@@ -187,12 +197,29 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
                 dataLocator = setPartRanges(dataLocator, byteRange);
             }
         }
+        if (awsGetPreCheck) {
```
LaurenSpiegel: let's move this out of here...

lib/data/external/AwsClient.js
```diff
@@ -7,6 +7,33 @@ const { prepareStream } = require('../../api/apiUtils/object/prepareStream');
 const { logHelper, removeQuotes, trimXMetaPrefix } = require('./utils');
 const { config } = require('../../Config');
 
+function _awsGetPreCheck(objectGetInfo, reqUids, callback) {
+    const log = createLogger(reqUids);
+    const key = typeof objectGetInfo === 'string' ? objectGetInfo :
```
LaurenSpiegel: not needed for aws backend keys.  the old format is very old. it pre-dates our aws backend.

lib/data/external/AwsClient.js
```diff
@@ -7,6 +7,33 @@ const { prepareStream } = require('../../api/apiUtils/object/prepareStream');
...
+        objectGetInfo.key;
+    const versionId = typeof objectGetInfo === 'string' ? undefined :
```
LaurenSpiegel: same

lib/data/external/AwsClient.js
```diff
@@ -7,6 +7,33 @@ const { prepareStream } = require('../../api/apiUtils/object/prepareStream');
...
+        Key: key,
+        VersionId: versionId,
+    }, err => {
+        if (err) {
+            logHelper(log, 'error', 'error heading object ' +
+            'from datastore', err, this._dataStoreName);
+            if (err.code === 'NotFound') {
+                const error = errors.InternalError.customizeDescription(
+                    'Unexpected error from AWS: "NoSuchVersion". Backend ' +
+                    'data may have diverged from our metadata records.'
```
LaurenSpiegel: why NoSuchVersion instead of NotFound? Also, I would do something like 'Data on AWS may have been altered outside of CloudServer.'

lib/data/external/AwsClient.js
```diff
@@ -15,6 +42,8 @@ class AwsClient {
         this._dataStoreName = config.dataStoreName;
         this._serverSideEncryption = config.serverSideEncryption;
         this._client = new AWS.S3(this._s3Params);
+        const awsGetPreCheck = _awsGetPreCheck.bind(this);
```
LaurenSpiegel: why not define as a method on the class then?

lib/data/external/AwsClient.js
```diff
@@ -43,23 +73,33 @@ class AwsClient {
...
                     return callback(errors.InternalError
                       .customizeDescription('Error returned from ' +
                       `AWS: ${err.message}`)
                     );
                 }
-                return callback(null, awsKey);
+                if (!data.VersionId) {
+                    const error = errors.InternalError.customizeDescription(
+                        'Invalid state. Please ensure versioning is enabled ' +
+                        'in AWS for the location constraint\'s bucket and ' +
```
LaurenSpiegel: I would just say for the 'location constraint' since we do allow location at the object level

lib/data/external/AwsClient.js
```diff
@@ -69,17 +109,27 @@ class AwsClient {
                      `AWS: ${err.message}`)
                    );
                }
-               return callback(null, awsKey);
+               if (!data.VersionId) {
+                   const error = errors.InternalError.customizeDescription(
+                       'Invalid state. Please ensure versioning is enabled ' +
+                       'in AWS for the location constraint\'s bucket and ' +
```
LaurenSpiegel: same and can you make a function to avoid the duplication?

lib/data/external/AwsClient.js
```diff
@@ -43,23 +73,33 @@ class AwsClient {
...
+                    const error = errors.InternalError.customizeDescription(
+                        'Invalid state. Please ensure versioning is enabled ' +
+                        'in AWS for the location constraint\'s bucket and ' +
+                        'try again.'
+                    );
+                    logHelper(log, 'error', 'missing version id for data ' +
+                    'backend object', error, this._dataStoreName);
+                    return callback(error);
+                }
+                const params = { dataStoreVersionId: data.VersionId };
```
LaurenSpiegel: do you foresee us having more params? if not, can we just send the versionId back?

lib/data/multipleBackendGateway.js
```diff
@@ -56,8 +56,8 @@ const multipleBackendGateway = {
                     dataStoreName: controllingLocationConstraint,
                     dataStoreType: client.clientType,
                 };
-                if (eTag) {
-                    dataRetrievalInfo.eTag = eTag;
+                if (backendParams) {
```
LaurenSpiegel: if we just send back the versionId it would be clearer what is going on here and wouldn't have to do an object assign.

tests/functional/aws-node-sdk/test/multipleBackend/get/get.js
```diff
@@ -279,20 +272,6 @@ describe('Multiple backend get object', function testSuite() {
                         done();
                     });
             });
-            it('should return an error on get done to object deleted from AWS',
```
LaurenSpiegel: shouldn't we still get an error but returned from your head check?

tests/functional/aws-node-sdk/test/multipleBackend/get/getAwsVersioning.js
```diff
@@ -0,0 +1,395 @@
...
+} = require('../utils');
+
+const awsLocation = 'aws-test';
+const someBody = 'testbody';
+const bucket = 'buckettestmultiplebackendgetawsversioning';
+
+let awsS3;
+
+function getAndAssertResult(s3, params, cb) {
+    const { bucket, key, body, versionId, expectVersionId } = params;
```
LaurenSpiegel: expectedVersionId

lib/api/objectGet.js
electrachong: as discussed, I updated to check if the `dataStoreType` is an external backend instead and use a multipleBackendsGateway.head method

lib/data/external/AwsClient.js
electrachong: updated

lib/data/external/AwsClient.js
electrachong: updated to create a `head` class method instead

tests/functional/aws-node-sdk/test/multipleBackend/get/get.js
electrachong: I just happened to move this test here: https://github.com/scality/S3/pull/913/files#diff-d1e4ee25cab34d7de04a5e66ff66e53bR346

electrachong:
@ironman-machine try

lib/api/objectGet.js
```diff
@@ -187,12 +193,31 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
                 dataLocator = setPartRanges(dataLocator, byteRange);
             }
         }
+        if (locationType && externalBackends[locationType]) {
```
LaurenSpiegel: instead of having the if clause here, just call the backend head and if the backend has head, it will return the needed response, otherwise it should be a noop and just return

lib/data/multipleBackendGateway.js
```diff
@@ -55,14 +55,17 @@ const multipleBackendGateway = {
...
                 };
-                if (eTag) {
-                    dataRetrievalInfo.eTag = eTag;
-                }
                 return callback(null, dataRetrievalInfo);
             });
     },
 
+    head: (objectGetInfo, reqUids, callback) => {
+        const client = clients[objectGetInfo.dataStoreName];
```
LaurenSpiegel: check if client has head call and if not, just cb

lib/data/multipleBackendGateway.js
```diff
@@ -43,7 +43,7 @@ const multipleBackendGateway = {
         }
 
         return client.put(writeStream, size, keyContext,
-            reqUids, (err, key, eTag) => {
+            reqUids, (err, key, dataStoreVersionId) => {
```
LaurenSpiegel: etag never needed?

lib/api/objectGet.js
electrachong: updated

lib/data/multipleBackendGateway.js
electrachong: updated

lib/data/multipleBackendGateway.js
electrachong: seems like it was left-over artifact when we were returning the eTag from aws, which we no longer do -- I checked with @dora-korpar to make sure

electrachong:
@ironman-machine try

electrachong:
retried twice, failed once due to a replication test; passed here:
http://ci.ironmann.io/gh/scality/Integration/15150

PR #915, opened 2017-09-27
TEST: Azure with BucketMatch equals false

nicolas2bert:
Testing Azure location Constraint with bucketMatch sets to false.


nicolas2bert:
@ironman-machine r+

PR #916, opened 2017-09-28
FT: Make REST endpoint mapping dynamic

RachedBenMustapha:
This allows changing REST endpoint to location constraint mappings on the fly, without restarting processes, by calling `Config.setRestEndpoints()`.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -40,6 +40,17 @@ function sproxydAssert(configSproxyd) {
     return sproxydFields;
 }
 
+function restEndpointsAssert(restEndpoints, locationConstraints) {
+    assert(typeof restEndpoints === 'object',
+        'bad config: restEndpoints must be an object of endpoints');
+    assert(Object.keys(restEndpoints).every(
+        r => typeof restEndpoints[r] === 'string'),
+        'bad config: each endpoint must be a string');
+    assert(Object.keys(restEndpoints).every(
+        r => typeof locationConstraints[restEndpoints[r]] === 'object'),
+        'bad config: rest endpoint target not in locationConstraints');
+}
+
 function locationConstraintAssert(locationConstraints) {
     const supportedBackends =
       ['mem', 'file', 'scality'].concat(Object.keys(externalBackends));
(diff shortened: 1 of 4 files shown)


lib/Config.js
```diff
@@ -737,6 +737,11 @@ class Config extends EventEmitter {
     getLocationConstraintType(locationConstraint) {
         return this.locationConstraints[locationConstraint].type;
     }
+
+    setRestEndpoints(restEndpoints) {
+        this.restEndpoints = restEndpoints;
```
LaurenSpiegel: Please validate that it is the correct format first and check that each endpoint value is a known location.

tests/unit/api/bucketPut.js
```diff
@@ -269,4 +269,29 @@ describe('bucketPut API', () => {
         });
         done();
     });
+
+    it('should pick up updated rest endpoint config', done => {
+        const bucketName = 'new-loc-bucket-name';
+        const newRestEndpoint = 'newly.defined.rest.endpoint';
+        const newLocation = 'scality-us-west-1';
```
LaurenSpiegel: the new location would have to be added to the locationConfig and parsed by the location constraint parser to create a new in memory client.

RachedBenMustapha: This new location (`'scality-us-west-1'`) is already present in the testing location config, the `bucketPut` call fails otherwise.

lib/Config.js
RachedBenMustapha: Updated, added the "known location part" that did not exist for the initial config load :)

LaurenSpiegel: I was hoping you would :)

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine clean

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine r+

PR #917, opened 2017-09-28
FT: AZURE COPY OBJECT

nicolas2bert:


lib/api/objectCopy.js
```diff
@@ -400,10 +405,12 @@ function objectCopy(authInfo, request, sourceBucket,
             return async.mapLimit(dataLocator, 1,
                 // eslint-disable-next-line prefer-arrow-callback
                 function copyPart(part, cb) {
-                    return data.get(part, null, log, (err, stream) => {
+                    const passThrough = new PassThrough();
+                    return data.get(part, passThrough, log, (err, stream) => {
                         if (err) {
                             return cb(err);
                         }
+                        const objStream = stream || passThrough;
```
LaurenSpiegel: I would have azure pass back a stream in its callback so you don't need this special case.

lib/data/external/AzureClient.js
```diff
@@ -206,6 +220,42 @@ class AzureClient {
...
+                      `${sourceContainerName} Azure Container`)
+                    );
+                }
+                logHelper(log, 'error', 'error from data backend on ' +
+                'copyObject', err, this._dataStoreName);
+                return callback(errors.InternalError
+                  .customizeDescription('Error returned from ' +
+                  `AWS: ${err.message}`)
+                );
+            }
```
LaurenSpiegel: we need to check the response from azure to see if the copy was successful or "pending".  If it is pending we should return an error such as InvalidObjectState with a customized description that the azure status is pending.  They don't guarantee that the call is done.  See -- https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob

lib/data/external/AzureClient.js
```diff
@@ -60,7 +61,20 @@ class AzureClient {
         const options = { metadata:
             this._translateMetaHeaders(keyContext.metaHeaders,
                 keyContext.tagging) };
-        this._client.createBlockBlobFromStream(this._azureContainerName,
+        if (size === 0) {
```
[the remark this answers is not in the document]
nicolas2bert: if copying an empty object to an Azure location

lib/data/external/AzureClient.js
nicolas2bert: ok thanks

nicolas2bert: @LaurenSpiegel I aborted it when status was pending.

LaurenSpiegel: As discussed, we will only do Azure server side copy if it is between the same account (which we know based on the endpoint). Otherwise, we will do our regular copy mechanism of get and put.

lib/api/objectCopy.js
```diff
@@ -318,14 +316,14 @@ function objectCopy(authInfo, request, sourceBucket,
...
+            let sourceLocationConstraintType;
             if (storeMetadataParams.metaHeaders[locationHeader]) {
-                locationType = config.getLocationConstraintType(
+                sourceLocationConstraintType = config.getLocationConstraintType(
                   storeMetadataParams.metaHeaders[locationHeader]);
             }
             // NOTE: remove the following when we will support putting a
             // versioned object to a location-constraint of type AWS or Azure.
-            if (constants.externalBackends[locationType]) {
+            if (constants.externalBackends[sourceLocationConstraintType]) {
```
LaurenSpiegel: @electrachong, note this will have to be updated.  We should move it to the wrapper when you change it to be just azure.

lib/data/wrapper.js
```diff
@@ -285,6 +288,173 @@ const data = {
         }
         return client.getDiskUsage(log.getSerializedUids(), cb);
     },
+
+    /**
+     * _retryDelete - Attempt to delete key again if it failed previously
```
LaurenSpiegel: ?

lib/data/wrapper.js
```diff
@@ -285,6 +288,173 @@ const data = {
...
+     * @param {object} serverSideEncryption - serverSideEncryption
+     * @param {object} log - Werelogs request logger
+     * @param {function} cb - callback
+     * @returns {function} cb - callback
+     */
+    copyObject: (request, sourceLocationConstraintType,
+      sourceLocationConstraintName, storeMetadataParams, dataLocator,
+      dataStoreContext, destBackendInfo, serverSideEncryption, log, cb) => {
+        // NOTE: using copyObject only if copying object from one external
+        // backend to the same external backend
```
LaurenSpiegel: and for Azure if it is the same account since Azure copy outside of an account is async

lib/data/wrapper.js
```diff
@@ -285,6 +288,173 @@ const data = {
...
+        }
+
+        // dataLocator is an array.  need to get and put all parts
+        // For now, copy 1 part at a time. Could increase the second
+        // argument here to increase the number of parts
+        // copied at once.
+        return async.mapLimit(dataLocator, 1,
+            // eslint-disable-next-line prefer-arrow-callback
+            function copyPart(part, cb) {
+                if (part.dataStoreType === 'azure') {
```
LaurenSpiegel: I see why you kept knowledge of azure here. @ploki will want to remove it but this is still a good first step to get the azure logic over to the wrapper and out of the api.

nicolas2bert:
@ironman-machine r+

nicolas2bert:
@ironman-machine r+

PR #919, opened 2017-10-04
rf: update auth config example file

jonathan-gramain:
Use proper ARNs for accounts, remove account users which is now unsupported.

This change will be necessary for the example file to be parsed correctly when https://github.com/scality/Arsenal/pull/346 is merged, but can be merged before since it will still be parsed correctly by the existing code in arsenal.


LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine r+

PR #922, opened 2017-10-05
FT: Add storageType to replicationInfo

bennettbuchanan:
Depends on https://github.com/scality/Arsenal/pull/347.

* Add the `storageType` field to the `replicationInfo` field of an object's metadata.
* If the given `storageClass` of a replication configuration rule matches a location in the locationConfig.json, we set the `storageType` to the value of `type` (for example, 'aws_s3') that is included in the given location constraint.
* The definition of a valid `storageClass` is expanded to include values that exist in the locationConfig.json.

### DIFF
--- a/tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
+++ b/tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
@@ -157,6 +157,64 @@ describe('aws-node-sdk test putBucketReplication configuration rules', () => {
             checkError(config, 'InvalidArgument', done));
     });
 
+    it('should not accept configuration when \'Role\' is a comma-separated ' +
+        'list of more than two valid Amazon Resource Names',
+        done => {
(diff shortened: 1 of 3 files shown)


tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
```diff
@@ -157,6 +157,24 @@ describe('aws-node-sdk test putBucketReplication configuration rules', () => {
...
+                'arn:aws:iam::account-id:role/resource-2,' +
+                'arn:aws:iam::account-id:role/resource-3';
+            const config = Object.assign({}, replicationConfig, { Role });
+            checkError(config, 'InvalidArgument', done);
+        });
+
+    replicationUtils.validRoleARNs.forEach(ARN => {
+        const config = Object.assign({}, replicationConfig, { Role: ARN });
+
+        it('should accept configuration when \'Role\' is a single valid ' +
```
rahulreddy: We should allow single Role ARNs only when storageType is aws_s3.

PR #924, opened 2017-10-05
FIX: Check only defined replication rule IDs

bennettbuchanan:
To be forward ported through to master branch in S3 and Arsenal.


LaurenSpiegel:
@ironman-machine r+

PR #926, opened 2017-10-05
ft: enable versioning deletes for aws backend

electrachong:
The versioning adventure continues. :clinking_glasses: 

Sorry for the length of the test file. The helper functions and comments are portions of that.


electrachong:
@LaurenSpiegel I have a question that I'd like to discuss with you sometime about how we manage overwriting/deleting "null" version data when versioning is suspended / not configured on the s3 side (if we want to do that at all).

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

lib/data/external/AwsClient.js
```diff
@@ -64,6 +52,23 @@ class AwsClient {
             return callback(null, awsKey, dataStoreVersionId);
         };
 
+        const params = {
+            Bucket: this._awsBucketName,
+            Key: awsKey,
+        };
+        if (keyContext.isDeleteMarker) {
```
LaurenSpiegel: it's a put of a delete marker?

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+    // can only get real aws config if not running end-to-end
+    const awsConfig = getRealAwsConfig(awsLocation);
+    awsS3 = new AWS.S3(awsConfig);
+}
+
+// order of items by index:
+// 0 - whether to expect a version id
+// 1 - whether version id should match request version id (undef if n/a)
+// 2 - whether x-amz-delete-marker response header should be true
+const _deleteResultSchema = {
```
LaurenSpiegel: nice

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+                return bucketUtil.deleteOne(bucket);
+            })
+            .catch(err => {
+                process.stdout.write('Error emptying/deleting bucket: ' +
+                `${err}\n`);
+                throw err;
+            });
+        });
+
+        it('versioning not configured: should delete a specific version ' +
```
LaurenSpiegel: versioning not configured on our s3: should put delete marker on key in AWS backend successfully

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+        'in AWS backend successfully', done => {
+            const key = `somekey-${Date.now()}`;
+            async.waterfall([
+                next => putToAwsBackend(s3, bucket, key, someBody,
+                    err => next(err)),
+                next => _awsGetLatestVerId(key, someBody, next),
+                (awsVerId, next) => delAndAssertResult(s3, { bucket,
+                    key, versionId: 'null', resultType: deleteVersion },
+                    err => next(err, awsVerId)),
+                (awsVerId, next) => _awsGetAssertDeleted({ key,
```
LaurenSpiegel: why is it actually deleting it? we don't want to delete anything on real AWS unless they have versioning enabled on our s3 and they request to delete a specific version. otherwise, it should be putting a delete marker on aws.  As discussed, it's ok that deleting since specifying null.  We should add a test where we specify an actual version id other than null and confirm that the AWS version is left alone.

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+                    versionId: awsVerId, errorCode: 'NoSuchVersion' }, next),
+            ], done);
+        });
+
+        it('versioning suspended: should delete a specific version in AWS ' +
```
LaurenSpiegel: same.  this should only be for deleting the null version on aws.

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+                // should be able to get object originally put from s3
+                (s3vid, next) => getAndAssertResult(s3, { bucket, key,
+                    body: someBody, expectedVersionId: s3vid }, next),
+                // latest version in aws should now be object originally put
+                next => _awsGetLatestVerId(key, someBody, next),
+            ], done);
+        });
+
+        it('multiple delete markers: should get NoSuchObject if only ' +
+        'only of the delete markers is deleted', done => {
```
LaurenSpiegel: only one of?

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+                next => delAndAssertResult(s3, { bucket, key, resultType:
+                    newDeleteMarker }, err => next(err)),
+                next => _getAssertDeleted(s3, { key, errorCode: 'NoSuchKey' },
+                    next),
+                next => _awsGetAssertDeleted({ key, errorCode: 'NoSuchKey' },
+                    next),
+            ], done);
+        });
+
+        it('versioning suspended: creating a delete marker should overwrite ' +
```
LaurenSpiegel: if versioning is suspended on our end, don't you mean delete on our s3 should put a delete marker on aws?

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+                // in aws we should still get another delete marker instead
+                // of the null version
+                next => _awsGetAssertDeleted({ key, errorCode: 'NoSuchKey' },
+                    next),
+            ], done);
+        });
+
+        it('versioning suspended: creating a delete marker should overwrite ' +
+        'existing null version that is not the latest version in s3 md ' +
+        '(but not in AWS due to edge case)',
```
LaurenSpiegel: by creating a delete marker, do you also mean just deleting the key on our s3?

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -0,0 +1,548 @@
...
+                    err => next(err)),
+                next => _getAssertDeleted(s3, { key, errorCode: 'NoSuchKey' },
+                    next),
+            ], done);
+        });
+    });
+});
+
+describeSkipIfNotMultiple('AWS backend delete object w. versioning: ' +
+    'using bucket location constraint', function testSuite() {
```
LaurenSpiegel: how does this differ from the describe above?

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -70,7 +70,9 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         return callback(err);
     }
 
-    const metaHeaders = isDeleteMarker ? [] :
+    const metaHeaders = objMD && isDeleteMarker ?
+        { 'x-amz-meta-scal-location-constraint':
+        objMD[constants.objectLocationConstraintHeader] } :
```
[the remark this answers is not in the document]
electrachong: This is mostly if someone put an object using the `x-amz-meta-scal-location-constraint` header instead of specifying the location constraint at the bucket level. If they create a delete marker by deleting that object, the location constraint is pulled from the existing object's metadata so the delete marker is created in the existing object's datastore ([line 123](https://github.com/scality/S3/pull/926/files/09b006811a5b2e03eb5549fdfd63f4d58228470f#diff-e45defa8c8c53634533cf165cc9232ceR123)).

I figured this is how it should work since the `x-amz-meta-scal-location-constraint` cannot be specified during a delete, but if we're going to abandon object level location constraints, this will be removed later on anyway.

lib/api/apiUtils/object/versioning.js
```diff
@@ -76,11 +76,18 @@ function _storeNullVersionMD(bucketName, objKey, objMD, options, log, cb) {
 * @param {string} objKey - name of object key
 * @param {object} options - metadata options for getting object MD
 * @param {string} options.versionId - version to get from metadata
+* @param {object} mst - info about the master version
```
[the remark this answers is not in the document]
electrachong: It's some versioning information extracted from the master version's object metadata, such as `versionId`, `isNull`, `isDeleteMarker` (see [here](https://github.com/scality/S3/pull/926/files/09b006811a5b2e03eb5549fdfd63f4d58228470f#diff-f595d5a19c0275ceb841fe9faa05e13eR193)). The only property we use in this function is `versionId` so that's the only one I included in the jsdoc.

lib/api/apiUtils/object/versioning.js
```diff
@@ -95,15 +102,21 @@ function _getDeleteLocations(bucketName, objKey, options, log, cb) {
             }
             const dataToDelete = Array.isArray(versionMD.location) ?
                 versionMD.location : [versionMD.location];
+            if (dataToDelete[0]) {
+                dataToDelete[0].deletingNonMasterNullVersion = true;
```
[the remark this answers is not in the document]
electrachong: This function is only called by `_deleteNullVersionMD` which is used for cleaning up null versions when creating new null versions.

lib/api/apiUtils/object/versioning.js
```diff
@@ -95,15 +102,21 @@ function _getDeleteLocations(bucketName, objKey, options, log, cb) {
...
 }
 
-function _deleteNullVersionMD(bucketName, objKey, options, log, cb) {
-    // before deleting null version md, retrieve location of data to delete
-    return _getDeleteLocations(bucketName, objKey, options, log,
+function _deleteNullVersionMD(bucketName, objKey, options, mst, log, cb) {
+    return _getDeleteLocations(bucketName, objKey, options, mst, log,
         (err, nullDataToDelete) => {
             if (err) {
+                process.stdout.write('=============================\n');
```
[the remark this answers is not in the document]
electrachong: This debugging info is in a dropme commit; the last time I ran a working end to end some tests related to null versions failed w/ `InternalError` but I wasn't able to run any working end to ends to print this info afterwards. You can view the code changes without the dropme here: 953dd54

lib/api/apiUtils/object/versioning.js
```diff
@@ -132,9 +149,8 @@ function processVersioningState(mst, vstat, cb) {
         if (vstat === 'Suspended') {
             options.versionId = '';
             options.isNull = true;
-            options.dataToDelete = mst.objLocation;
             // if null version exists, clean it up prior to put
-            if (mst.isNull) {
+            if (mst.exists) {
```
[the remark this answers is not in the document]
electrachong: I think I can change this back. It was related to something I was trying to do before, but it wasn't necessary in the end.

lib/data/external/AwsClient.js
electrachong: Yes 👍

lib/api/apiUtils/object/createAndStoreObject.js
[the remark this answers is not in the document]
LaurenSpiegel: Let's get rid of the ?:?: 

Also, instead of messing with metaheaders, let's add to the request object, the x-amz-scal-location-constraint that might be stored on an object so that if it is a delete marker we can save that info with the delete marker.

lib/api/apiUtils/object/versioning.js
```diff
@@ -76,11 +76,18 @@ function _storeNullVersionMD(bucketName, objKey, objMD, options, log, cb) {
...
 * @param {object} options - metadata options for getting object MD
 * @param {string} options.versionId - version to get from metadata
+* @param {object} mst - info about the master version
+* @param {string} mst.versionId - the master version's version id
 * @param {RequestLogger} log - logger instanceof
 * @param {function} cb - callback
 * @return {undefined} - and call callback with (err, dataToDelete)
 */
-function _getDeleteLocations(bucketName, objKey, options, log, cb) {
+function _getDeleteLocations(bucketName, objKey, options, mst, log, cb) {
```
LaurenSpiegel: _getNullVersionsToDelete ?

lib/data/external/AwsClient.js
LaurenSpiegel: please just add a comment stating that we call put to create a delete marker but in calling AWS it's actually a delete to them

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
LaurenSpiegel: as discussed, you can create a delete marker while versioning is suspended and it will overwrite the latest null version.  please add a comment.

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
LaurenSpiegel: add to describe statement above that they are using object location constraint

electrachong:
are end-to-ends fixed yet?
@ironman-machine try

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -61,6 +61,10 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         canonicalID, cipherBundle, request, isDeleteMarker, streamingV4Params,
         log, callback) {
     const size = isDeleteMarker ? 0 : request.parsedContentLength;
+    // although the request method may actually be 'DELETE' if creating a
+    // delete marker, for our purposes we consider this to be a 'PUT'
+    // operation
+    const requestMethod = 'PUT';
```
dora-korpar: I feel like doing this might mess with this line in data.batchDelete: 
https://github.com/scality/S3/blob/master/lib/data/wrapper.js#L229
Or is the idea that, if using AWS backend, we won't have to go through that batchDelete to delete objects anymore?

lib/data/external/AwsClient.js
```diff
@@ -117,18 +124,25 @@ class AwsClient {
         return callback(null, stream);
     }
     delete(objectGetInfo, reqUids, callback) {
-        // for backwards compatibility
-        const key = typeof objectGetInfo === 'string' ? objectGetInfo :
-          objectGetInfo.key;
+        const { key, dataStoreVersionId, deleteVersion } = objectGetInfo;
```
dora-korpar: no need for backwards compat anymore?

lib/data/external/AwsClient.js
```diff
@@ -117,18 +124,25 @@ class AwsClient {
...
         return this._client.deleteObject(params, err => {
             if (err) {
-                const log = createLogger(reqUids);
                 logHelper(log, 'error', 'error deleting object from ' +
                 'datastore', err, this._dataStoreName);
+                if (err.code === 'NoSuchVersion') {
+                    // data may have been deleted directly from the AWS backend
+                    // don't want to retry the delete and errors are not
+                    // sent back to client anyway, so no need to return err
+                    return callback();
```
dora-korpar: i don't follow the logic on why not return an internal error in this case?

lib/api/apiUtils/object/createAndStoreObject.js
electrachong: I think we already opted to not go through with the batchDelete if using AWS backend for cleaning up old data for normal PUTs. The idea is to do the same when creating delete markers, even though the request.method in that case is "DELETE" instead of "PUT", to be consistent.

lib/data/external/AwsClient.js
electrachong: Nope because if I understand correctly, if we have the object stored in metadata as existing on the AWS backend, that object must have been put with the new format. (Couldn't have put the object with the old format on AWS backend since AWS backend didn't exist when we were still using the old format)

lib/data/external/AwsClient.js
electrachong: We can return the error, but it will just result in the delete being retried. The error doesn't get sent back to the user because at that point we have already deleted the metadata for the object:
https://github.com/scality/S3/blob/master/lib/data/wrapper.js#L203

And since we can expect this error to occur if someone deletes the object directly from the AWS backend, I didn't think it was necessary to retry the delete a few times since retries will not address that problem. But maybe we can just in case, if you think it's worthwhile.

dora-korpar: I'm not super familiar with the intricacies of versioning, but I'd say if that's the only case when the NoSuchVersion error is returned, then I like skipping the retries :)

electrachong:
@ironman-machine try

electrachong:
@ironman-machine r+

electrachong:
retried here: http://ci.ironmann.io/gh/scality/Integration/15531

electrachong:
end to end finally passed; merging

PR #927, opened 2017-10-05
decouple API from multiple backend on uploadPart

ploki:
refactoring evaluation

### DIFF
--- a/lib/api/objectPutPart.js
+++ b/lib/api/objectPutPart.js
@@ -12,8 +12,6 @@ const kms = require('../kms/wrapper');
 const metadata = require('../metadata/wrapper');
 const { pushMetric } = require('../utapi/utilities');
 const logger = require('../utilities/logger');
-const { config } = require('../Config');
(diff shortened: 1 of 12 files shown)


ploki:
dear reviewer, only the last commit is relevant for this PR

lib/api/objectPutPart.js
```diff
@@ -179,12 +177,11 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
         },
         (destinationBucket, objectLocationConstraint, cipherBundle,
-        splitter, next) => {
-            if (config.backends.data === 'multiple') {
-                return multipleBackendGateway.uploadPart(request,
-                streamingV4Params, null, size, objectLocationConstraint,
-                objectKey, uploadId, partNumber, bucketName, log,
-                (err, partInfo) => {
+         splitter, next) => {
+            const res = data.uploadPart(
```
jonathan-gramain: style: I don't think `res` needs to be returned since the processing is asynchronous, just calling `data.uploadPart(...);` should be good.

PR #934, opened 2017-10-10
Fix: Wrap Azure calls with try/catch

dora-korpar:
Fix to handle when Azure requests throw errors instead of returning them in the callback.

Depends on https://github.com/scality/Arsenal/pull/350

### DIFF
--- a/tests/functional/aws-node-sdk/test/multipleBackend/put/putAzure.js
+++ b/tests/functional/aws-node-sdk/test/multipleBackend/put/putAzure.js
@@ -144,6 +144,20 @@ describeF() {
                 });
             });
 
+            it('should return error InternalError putting an invalid key ' +
+            'name to Azure', done => {
+                const params = {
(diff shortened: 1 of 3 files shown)


dora-korpar:
Test added :)

tests/functional/aws-node-sdk/test/multipleBackend/put/putAzure.js
```diff
@@ -142,6 +142,20 @@ describeF() {
                 });
             });
 
+            it('should return error InternalError putting an invalid key ' +
+            'name to Azure', done => {
+                const params = {
+                    Bucket: azureContainerName,
+                    Key: '.',
```
LaurenSpiegel: this causes the azure client to throw?

LaurenSpiegel: yes

PR #935, opened 2017-10-10
FT: Add replication to AWS backbeat route

bennettbuchanan:
Adds route for replication of single-part objects to AWS.

Forthcoming replication operation support:
* MPU objects
* Metadata-only operations (zero-byte objects, putting object ACLs, etc.)
* Object deletion

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -11,6 +11,8 @@ const { dataStore } = require('../api/apiUtils/object/storeObject');
 const prepareRequestContexts = require(
 '../api/apiUtils/authorization/prepareRequestContexts');
 const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
+const { BackendInfo } = require('../api/apiUtils/object/BackendInfo');
+const { locationConstraints } = require('../Config').config;
 
 auth.setHandler(vault);
 
@@ -57,21 +59,34 @@ function _getRequestPayload(req, cb) {
     .on('end', () => cb(null, Buffer.concat(payload, payloadLen).toString()));
 }
 
+function _isValidLocation(request) {
+    const { bucketName, headers } = request;
+    const storageType = headers['x-scal-storage-type'];
+    const storageLocation = headers['x-scal-storage-class'];
+    return locationConstraints[storageLocation] &&
+        locationConstraints[storageLocation].type === storageType &&
+        locationConstraints[storageLocation].details.bucketName === bucketName;
+}
+
 /*
 PUT /_/backbeat/metadata/<bucket name>/<object key>
 PUT /_/backbeat/data/<bucket name>/<object key>
+PUT /_/backbeat/multiplebackendputobject/<bucket name>/<object key>
 */
 
 function putData(request, response, bucketInfo, objMd, log, callback) {
+    let errMessage;
     const canonicalID = request.headers['x-scal-canonical-id'];
     if (canonicalID === undefined) {
-        log.error('bad request: missing x-scal-canonical-id header');
-        return callback(errors.BadRequest);
+        errMessage = 'bad request: missing x-scal-canonical-id header';
+        log.error(errMessage);
+        return callback(errors.BadRequest.customizeDescription(errMessage));
     }
     const contentMD5 = request.headers['content-md5'];
     if (contentMD5 === undefined) {
-        log.error('bad request: missing content-md5 header');
-        return callback(errors.BadRequest);
+        errMessage = 'bad request: missing content-md5 header';
+        log.error(errMessage);
+        return callback(errors.BadRequest.customizeDescription(errMessage));
     }
     const context = {
         bucketName: request.bucketName,
@@ -94,6 +109,10 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         context, CIPHER, request, payloadLen, {},
         backendInfo, log, (err, retrievalInfo, md5) => {
             if (err) {
+                log.error('error putting data', {
+                    error: err,
+                    method: 'putData',
+                });
                 return callback(err);
             }
             if (contentMD5 !== md5) {
@@ -144,6 +163,10 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         return metadata.putObjectMD(bucketName, objectKey, omVal, options, log,
             (err, md) => {
                 if (err) {
+                    log.error('error putting object metadata', {
+                        error: err,
+                        method: 'putMetadata',
+                    });
                     return callback(err);
                 }
                 return _respond(response, md, log, callback);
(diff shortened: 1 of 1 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -57,9 +59,29 @@ function _getRequestPayload(req, cb) {
...
+    const { locationConstraints } = config;
+    return storageLocation &&
+        locationConstraints[storageLocation] &&
+        locationConstraints[storageLocation].type === 'aws_s3';
+}
+
 /*
 PUT /_/backbeat/metadata/<bucket name>/<object key>
 PUT /_/backbeat/data/<bucket name>/<object key>
+PUT /_/backbeat/aws/<bucket name>/<object key>
```
jonathan-gramain: If I understand the design correctly (correct me if I'm wrong), this route shall be agnostic of the backend since it will be driven by location constraints, so it should not be named after AWS directly but have a more generic name? Could be `/_/backbeat/object/...`, for example, meant to be used as a single-call object interface.

bennettbuchanan: Updated to `/_/backbeat/multipleBackendData/<bucket name>/<object key>`.

lib/routes/routeBackbeat.js
```diff
@@ -151,10 +176,59 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
     });
 }
 
+function putDataMultipleBackends(request, response, bucketInfo, objMd, log,
+    callback) {
+    const storageLocation = request.headers['x-scal-storage-class'];
+    const canonicalID = request.headers['x-scal-canonical-id'];
+    if (canonicalID === undefined) {
```
electrachong: what if canonicalID (and the headers content-md5 & scal-version-id below) is an empty string?
is some validation for format performed later on?

lib/routes/routeBackbeat.js
```diff
@@ -151,10 +164,59 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+        owner: canonicalID,
+        namespace: NAMESPACE,
+        objectKey: request.objectKey,
+        metaHeaders,
+    };
+    const payloadLen = parseInt(request.headers['content-length'], 10);
+    const backendInfo = new BackendInfo(storageLocation);
+    return dataStore(context, CIPHER, request, payloadLen, {}, backendInfo, log,
+        (err, retrievalInfo, md5) => {
+            if (err) {
```
rahulreddy: Please log the error here.

lib/routes/routeBackbeat.js
```diff
@@ -198,18 +265,31 @@ function routeBackbeat(clientIP, request, response, log) {
...
-                                             request.resourceType });
+                    resourceType: request.resourceType
+                });
                 return next(errors.MethodNotAllowed);
             }
-            const versioningConfig = bucketInfo.getVersioningConfiguration();
-            if (!versioningConfig || versioningConfig.Status !== 'Enabled') {
-                log.debug('bucket versioning is not enabled',
-                    { method: request.method,
+            if (usingMultipleBackend) {
```
rahulreddy: This can be combined into one
```js
 if (usingMultipleBackend && !_isValidLocation(request)) {

}
```

lib/routes/routeBackbeat.js
```diff
@@ -198,18 +265,31 @@ function routeBackbeat(clientIP, request, response, log) {
...
+                        method: request.method,
                         bucketName: request.bucketName,
                         objectKey: request.objectKey,
-                        resourceType: request.resourceType });
-                return next(errors.InvalidBucketState);
+                        resourceType: request.resourceType,
+                    });
+                    return next(errors.InvalidRequest);
+                }
+            } else {
```
rahulreddy: You would not need `else` here if you returned early above.

lib/routes/routeBackbeat.js
```diff
@@ -185,6 +248,10 @@ function routeBackbeat(clientIP, request, response, log) {
             return next(err, userInfo);
         }, 's3', requestContexts),
         (userInfo, next) => {
+            if (usingMultipleBackend) {
+                // Bucket and object do not exist in metadata.
+                return next(null, null, null);
```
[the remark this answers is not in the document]
bennettbuchanan: The callback `next` would be `undefined` in the subsequent waterfall step without these `null` values.

lib/routes/routeBackbeat.js
bennettbuchanan: I updated to return early above, but thought it would be cleaner to reuse the return value just once:
```js
return backbeatRoutes[request.method][request.resourceType](
    request, response, bucketInfo, objMd, log, next);
```

lib/routes/routeBackbeat.js
bennettbuchanan: I want to return the route if using a multiple backend, and an error if the location is invalid.

lib/routes/routeBackbeat.js
```diff
@@ -151,10 +164,59 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
     });
 }
 
+function multipleBackendPutObject(request, response, bucketInfo, objMd, log,
+    callback) {
+    const storageLocation = request.headers['x-scal-storage-class'];
+    const canonicalID = request.headers['x-scal-canonical-id'];
+    if (canonicalID === undefined) {
+        log.error('bad request: missing x-scal-canonical-id header');
+        return callback(errors.BadRequest);
```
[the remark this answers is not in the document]
bennettbuchanan: Does this apply to all the `BadRequest` errors in this file?

lib/routes/routeBackbeat.js
bennettbuchanan: I've updated to log errors here, across the routes.

lib/routes/routeBackbeat.js
```diff
@@ -57,9 +59,20 @@ function _getRequestPayload(req, cb) {
     .on('end', () => cb(null, Buffer.concat(payload, payloadLen).toString()));
 }
 
+function _isValidLocation(request) {
+    const { locationConstraints } = config;
```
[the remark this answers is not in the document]

lib/routes/routeBackbeat.js
bennettbuchanan: I assume the values stored in object metadata are valid, so we just need to check that the header is sent.

lib/routes/routeBackbeat.js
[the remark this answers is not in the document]
rahulreddy: Ah, I forgot about the waterfall nuances.

lib/routes/routeBackbeat.js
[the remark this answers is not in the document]
rahulreddy: Yup.

lib/routes/routeBackbeat.js
rahulreddy: I see the login one, missed it during the review.

lib/routes/routeBackbeat.js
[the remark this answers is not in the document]

LaurenSpiegel:
@ironman-machine r+

PR #937, opened 2017-10-13
S3C-1003 FX: COPY External backends encryption

nicolas2bert:
We need to exclude certain scenarios from using AWS/Azure server-side copy (for copy object and copy part) due to encryption.
Server side copy should only be allowed:
1) if source object and destination object are both on aws or both on azure.
2) if azure to azure, must be the same storage acocunt
3) if the source bucket is not an encrypted bucket and the destination bucket is not an encrypted bucket (unless the copy is all within the same bucket).
Also, we should make sure we are sending the sse header to aws if the destination location has SSE set to true in the locationConfig.

### DIFF
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -6,6 +6,8 @@ const createLogger = require('../multipleBackendLogger');
 const { prepareStream } = require('../../api/apiUtils/object/prepareStream');
 const { logHelper, removeQuotes, trimXMetaPrefix } = require('./utils');
 const { config } = require('../../Config');
+const constants = require('../../../constants');
+const locationHeader = constants.objectLocationConstraintHeader;
 
 class AwsClient {
     constructor(config) {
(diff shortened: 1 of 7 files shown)


lib/data/external/AwsClient.js
```diff
@@ -403,21 +405,28 @@ class AwsClient {
...
+        const metaHeaders = trimXMetaPrefix(getMetaHeaders(headers));
+        const s3Params = {
             Bucket: this._awsBucketName,
             Key: destAwsKey,
             CopySource: `${sourceAwsBucketName}/${sourceKey}`,
             Metadata: metaHeaders,
             MetadataDirective: metadataDirective,
-        }, err => {
+        };
+        const isAWSServerSideEncrytion = headers[locationHeader] ?
```
LaurenSpiegel: what if the destination bucket is an AWS serviceSideEncryption location?

lib/data/wrapper.js
```diff
@@ -309,23 +363,35 @@ const data = {
...
         const locationTypeMatch = request.headers[locationHeader] ?
         config.getLocationConstraintType(sourceLocationConstraintName) ===
         config.getLocationConstraintType(request.headers[locationHeader])
         : true;
-        if (locationTypeMatch &&
+        const sourceBucketName = sourceBucketMD.getName();
+        const destBucketName = request.bucketName;
+        const isSameBucket = sourceBucketName === destBucketName;
+        const serverSideEncryption = destBucketMD.getServerSideEncryption();
+        const bucketNotEncrypted = serverSideEncryption
```
LaurenSpiegel: bucketsNotEncrypted

lib/data/wrapper.js
```diff
@@ -309,23 +363,35 @@ const data = {
...
+        === sourceBucketMD.getServerSideEncryption() === null;
+        if (locationTypeMatch && (isSameBucket || bucketNotEncrypted) &&
```
LaurenSpiegel: I would wrap all of this logic determining whether we can do server side copy in a utility function. That way it will be clearer and can be used for copy part as well. Plus, you can unit test it.

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -284,6 +286,30 @@ function testSuite() {
...
+                    `error: ${err}`);
+                    assert.strictEqual(result.CopyObjectResult.ETag,
+                        `"${correctMD5}"`);
+                    assertGetObjects(key, bucket, awsLocation, copyKey, bucket,
+                        awsLocationEncryption, copyKey, 'REPLACE', false, awsS3,
+                        awsLocation, done);
+                });
+            });
+        });
+
```
LaurenSpiegel: I would add unit tests for our logic of whether server side copy is ok.

lib/data/external/AwsClient.js
nicolas2bert: We only care about the object metadata location.
Each objects has its own location and we use this location to encrypt it or not.

PR #939, opened 2017-10-13
fix: dataStoreName stored on put

dora-korpar:


dora-korpar:
@ironman-machine try

dora-korpar:
Turns out @electrachong did a much more thorough fix of this same issue here: https://github.com/scality/S3/pull/929/files
which was reported a while ago, so we are using that solution instead

PR #941, opened 2017-10-17
chore: move StatsClient and RedisClient to arsenal

philipyoo:
Moving StatsClient and RedisClient to arsenal to reuse these classes
in other repos.

Module ioredis is moved to devDependencies because it is still being
used in the healthcheck tests.

This should be merged after https://github.com/scality/Arsenal/pull/353


electrachong:
@ironman-machine r+

PR #942, opened 2017-10-17
S3C-1018 FIX: copy with locationConstraint

nicolas2bert:
On object copy/copy part we should be copying to the data location of the destination bucket unless the user specifies REPLACE and uses the `x-amz-meta-scal-location-constraint` header.

### DIFF
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -28,12 +28,14 @@ const externalVersioningErrorMessage = 'We do not currently support putting ' +
 
 /**
  * Preps metadata to be saved (based on copy or replace request header)
+ * @param {object} request - request
  * @param {object} sourceObjMD - object md of source object
  * @param {object} headers - request headers
  * @param {boolean} sourceIsDestination - whether or not source is same as
  * destination
  * @param {AuthInfo} authInfo - authInfo from Vault
  * @param {string} objectKey - destination key name
+ * @param {object} sourceBucketMD - bucket metadata of source bucket
  * @param {object} destBucketMD - bucket metadata of bucket being copied to
  * @param {object} log - logger object
  * @return {object}
@@ -41,8 +43,8 @@ const externalVersioningErrorMessage = 'We do not currently support putting ' +
  * - sourceLocationConstraintName {string} - location type of the source
  * - OR error
  */
-function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
-    objectKey, destBucketMD, log) {
+function _prepMetadata(request, sourceObjMD, headers, sourceIsDestination,
+    authInfo, objectKey, sourceBucketMD, destBucketMD, log) {
     let whichMetadata = headers['x-amz-metadata-directive'];
     // Default is COPY
     whichMetadata = whichMetadata === undefined ? 'COPY' : whichMetadata;
@@ -91,15 +93,31 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         });
         return { error: userMetadata };
     }
-    // If metadataDirective is 'COPY' and location constraint header is
-    // specified, new location constraint should override source object's
-    if (whichMetadata === 'COPY' && headers[locationHeader]) {
-        userMetadata[locationHeader] = headers[locationHeader];
+    // If metadataDirective is:
+    // - 'COPY' and source object has a location constraint in its metadata
+    // we use the bucket destination location constraint
+    if (whichMetadata === 'COPY'
+        && userMetadata[locationHeader]
+        && destBucketMD.getLocationConstraint()) {
+        userMetadata[locationHeader] = destBucketMD.getLocationConstraint();
     }
-    const sourceLocationConstraintName = sourceObjMD[locationHeader];
+    const backendInfoObjSource = locationConstraintCheck(request,
+        sourceObjMD, sourceBucketMD, log);
+    if (backendInfoObjSource.err) {
+        return { error: backendInfoObjSource.err };
+    }
+    const sourceLocationConstraintName = backendInfoObjSource.controllingLC;
+
+    const backendInfoObjDest = locationConstraintCheck(request,
+        userMetadata, destBucketMD, log);
+    if (backendInfoObjDest.err) {
+        return { error: backendInfoObjSource.err };
+    }
+    const destLocationConstraintName = backendInfoObjDest.controllingLC;
+
     // If location constraint header is not included, locations match
-    const locationMatch = headers[locationHeader] ?
-        sourceLocationConstraintName === headers[locationHeader] : true;
+    const locationMatch =
+    sourceLocationConstraintName === destLocationConstraintName;
 
     // If tagging directive is REPLACE but you don't specify any
     // tags in the request, the destination object will
@@ -147,7 +165,8 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
     if (!storeMetadataParams.contentType) {
         storeMetadataParams.contentType = sourceObjMD['content-type'];
     }
-    return { storeMetadataParams, sourceLocationConstraintName };
+    return { storeMetadataParams, sourceLocationConstraintName,
+      backendInfoObjDest };
 }
 
 /**
@@ -259,10 +278,10 @@ function objectCopy(authInfo, request, sourceBucket,
                         return next(errors.PreconditionFailed, destBucketMD);
                     }
                     const { storeMetadataParams, error: metadataError,
-                    sourceLocationConstraintName } =
-                        _prepMetadata(sourceObjMD, request.headers,
+                    sourceLocationConstraintName, backendInfoObjDest } =
+                        _prepMetadata(request, sourceObjMD, request.headers,
                             sourceIsDestination, authInfo, destObjectKey,
-                            destBucketMD, log);
+                            sourceBucketMD, destBucketMD, log);
                     if (metadataError) {
                         return next(metadataError, destBucketMD);
                     }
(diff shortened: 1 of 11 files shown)


lib/api/objectCopy.js
```diff
@@ -91,10 +91,11 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         });
         return { error: userMetadata };
     }
-    // If metadataDirective is 'COPY' and location constraint header is
-    // specified, new location constraint should override source object's
-    if (whichMetadata === 'COPY' && headers[locationHeader]) {
-        userMetadata[locationHeader] = headers[locationHeader];
+    // If metadataDirective is 'COPY' or 'REPLACE' with no location constraint,
+    // we use the bucket destination location constraint
+    if ((whichMetadata === 'COPY' || (whichMetadata === 'REPLACE' &&
```
electrachong: As discussed it doesn't make sense to me to copy the bucket location constraint to the obj location header if there was no obj location header on the original object. The check can be:
`if (sourceObjMd[locationHeader] && (whichMetadata === 'COPY' || (whichMetadata === 'REPLACE' && !userMetadata[locationHeader])))`
if that makes sense: if there was an obj location constraint on the original object and the directive is copy, or the directive is replace and there is no new obj location constraint, then we store the bucket location constraint as the obj location constraint

lib/api/objectCopy.js
```diff
@@ -91,12 +93,20 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         });
         return { error: userMetadata };
     }
-    // If metadataDirective is 'COPY' and location constraint header is
-    // specified, new location constraint should override source object's
-    if (whichMetadata === 'COPY' && headers[locationHeader]) {
-        userMetadata[locationHeader] = headers[locationHeader];
+    // If metadataDirective is:
+    // - 'COPY' and source object has a locaction contrain in its metadata
```
dora-korpar: 'location constraint'

lib/api/objectCopy.js
```diff
@@ -91,12 +93,20 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
...
+    // we use the bucket destination location constraint
```
dora-korpar: also, just for clarification, this new functionality is:
with MD directive 'COPY', the new object will be put to the destination bucket location, even if the source object has a location constraint in its metadata?

lib/api/objectCopy.js
```diff
@@ -91,12 +93,20 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
...
-    const sourceLocationConstraintName = sourceObjMD[locationHeader];
+    const backendInfoObjSource = locationConstraintCheck(request,
+        sourceObjMD, sourceBucketMD, log);
+    if (backendInfoObjSource.err) {
+        return { error: backendInfoObjSource.err };
+    }
+    const sourceLocationConstraintName = backendInfoObjSource.controllingLC;
     // If location constraint header is not included, locations match
     const locationMatch = headers[locationHeader] ?
         sourceLocationConstraintName === headers[locationHeader] : true;
```
LaurenSpiegel: shouldn't you be comparing the sourceLocationConstraintName to the destinationLocation (which could be the userMetadata[locationHeader] but might not be if you haven't set it? In any case, the headers seem to be the wrong place to look and we should always be checking against the actual destination.

lib/data/wrapper.js
```diff
@@ -333,37 +333,41 @@ const data = {
...
+            const destLocationConstraintName =
+                destBackendInfo.getControllingLocationConstraint();
+            const locationTypeMatch =
+            config.getLocationConstraintType(sourceLocationConstraintName) ===
+            config.getLocationConstraintType(destLocationConstraintName);
+            if (locationTypeMatch &&
+            (sourceLocationConstraintType === 'aws_s3' ||
+            (sourceLocationConstraintType === 'azure' &&
+            config.isSameAzureAccount(sourceLocationConstraintName,
+            destLocationConstraintName)))) {
```
LaurenSpiegel: please make this conditional a utility function that returns true or false.  also, please add unit tests for the different scenarios.

lib/kms/utilities.js
```diff
@@ -53,6 +54,13 @@ function _createEncryptedBucket(host,
     if (verbose) {
         logger.info('request headers', { headers: request._headers });
     }
+    if (locationConstraint && locationConstraint !== 'undefined') {
```
LaurenSpiegel: why a string of undefined?

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/azureObjectCopy.js
```diff
@@ -218,30 +256,98 @@ describeSkipIfNotMultiple('MultipleBackend object copy', function testSuite() {
...
+        function itFn(done) {
+            putSourceObj(this.test.key, azureLocation, null, bucketAzure,
+            () => {
+                const copyParams = {
+                    Bucket: bucketAzure,
+                    Key: this.test.copyKey,
+                    CopySource: `/${bucketAzure}/${this.test.key}`,
                     MetadataDirective: 'COPY',
-                    Metadata: { 'scal-location-constraint': azureLocation },
+                    Metadata: { 'scal-location-constraint': memLocation },
```
LaurenSpiegel: why memLocation?  To just show sending this doesn't matter?

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -128,18 +139,93 @@ function testSuite() {
...
+                process.stdout.write(`Deleting bucket ${bucketAws}\n`);
+                return bucketUtil.deleteOne(bucketAws);
+            })
             .catch(err => {
                 process.stdout.write(`Error in afterEach: ${err}\n`);
                 throw err;
             });
         });
 
+        it('should copy an object from mem to AWS relying on bucket location',
```
LaurenSpiegel: relying on destination bucket location

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -128,18 +139,93 @@ function testSuite() {
...
+                    assert.strictEqual(result.CopyObjectResult.ETag,
+                        `"${correctMD5}"`);
+                    assertGetObjects(key, bucket, undefined, copyKey,
+                        bucketAws, undefined, copyKey, 'COPY', false, awsS3,
+                        awsLocation, done);
+                });
+            });
+        });
+
+        it('should copy an object from AWS to mem relying on bucket location',
```
LaurenSpiegel: relying on destination bucket location

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -186,60 +272,83 @@ function testSuite() {
...
-                        awsLocation, done);
+                    assertGetObjects(key, bucket, undefined, copyKey,
+                        bucketAws, undefined, copyKey, 'REPLACE', false,
+                        awsS3, awsLocation, done);
+                });
+            });
+        });
+
+        it('should return NotImplemented copying an object from mem to a ' +
+        'versioning enable AWS bucket', done => {
```
LaurenSpiegel: @electrachong heads up

lib/Config.js
```diff
@@ -766,6 +766,19 @@ class Config extends EventEmitter {
         this.getAzureEndpoint(locationConstraintSrc) ===
         this.getAzureEndpoint(locationConstraintDest) : true;
     }
+
+    externalBackendCopy(locationConstraintSrc, locationConstraintDest) {
```
LaurenSpiegel: I would put this function in the external backend utils file.

lib/api/objectCopy.js
```diff
@@ -91,15 +93,31 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
         });
         return { error: userMetadata };
     }
-    // If metadataDirective is 'COPY' and location constraint header is
-    // specified, new location constraint should override source object's
-    if (whichMetadata === 'COPY' && headers[locationHeader]) {
-        userMetadata[locationHeader] = headers[locationHeader];
+    // If metadataDirective is:
+    // - 'COPY' and source object has a locaction constrain in its metadata
```
electrachong: `location constraint`

lib/data/wrapper.js
```diff
@@ -326,44 +323,40 @@ const data = {
...
-                };
-                const putResultArr = [putResult];
-                return cb(null, putResultArr);
-            });
+        if (config.backends.data === 'multiple') {
+            const destLocationConstraintName =
+                destBackendInfo.getControllingLocationConstraint();
+            if (config.externalBackendCopy(sourceLocationConstraintName,
+            destLocationConstraintName)) {
+                const objectGetInfo = dataLocator[0];
```
electrachong: no need for backwards compat here?

lib/api/objectCopy.js
```diff
@@ -91,15 +93,31 @@ function _prepMetadata(sourceObjMD, headers, sourceIsDestination, authInfo,
...
+    // we use the bucket destination location constraint
+    if (whichMetadata === 'COPY'
+        && userMetadata[locationHeader]
+        && destBucketMD.getLocationConstraint()) {
+        userMetadata[locationHeader] = destBucketMD.getLocationConstraint();
```
electrachong: Based on this, are we ignoring object level location constraint which may have been on the copy object request, even if it takes priority over bucket location constraint?

And just to confirm what we do when REPLACE:
If copy object request has obj-lvl location constraint, we copy to dest location header?
If copy object request has bucket-lvl location constraint and no obj-lvl location constraint, and source object had a location header, we copy the bucket-lvl location constraint to the dest location header?

lib/api/objectCopy.js
nicolas2bert: Thanks :)

lib/api/objectCopy.js
nicolas2bert: Here we use the metadata header from the request if copy https://github.com/scality/S3/pull/942/files#diff-5ce9bf7b97556522975c2fae1a6944efR86

Yes, If copy object request has `x-amz-meta-scal-location-constraint`, we copy to dest location constraint

If you copy an object with `azure-test` location using `REPLACE` and no `x-amz-meta-scal-location-constraint` to a bucket with `mem-test` location constraint , the copied object location constraint will be empty so we will use the location of the destination bucket, ie `mem-test`.

Does it make sense?

electrachong: At line 86, `userMetadata` looks like the source object MD, not the request headers (if COPY). ?

Can you give me the line # for where you copy the request `x-amz-meta-scal-location-constraint` to the dest location constraint? I think it will be quicker for you to find than for me to look.

> If you copy an object with azure-test location using REPLACE and no x-amz-meta-scal-location-constraint to a bucket with mem-test location constraint , the copied object location constraint will be empty so we will use the location of the destination bucket, ie mem-test.
It makes sense but where do you copy the destination bucket location constraint? Since in the condition above it looks like it only applies for `COPY`?

nicolas2bert: Sorry I mean "Here we use the metadata header from the request if **REPLACE**" https://github.com/scality/S3/pull/942/files#diff-5ce9bf7b97556522975c2fae1a6944efR88

electrachong: OK, as discussed:
- when `REPLACE` and there was source object `x-amz-meta-scal-location-constraint` but no request `x-amz-meta-scal-location-constraint`, we leave dest `x-amz-meta-scal-location-constraint` empty and default to using the bucket location constraint
- when `COPY` we disregard the request headers for metadata, since in concept we are pulling the metadata from the original object.

Thanks for the clarification!

nicolas2bert:
@ironman-machine try

PR #943, opened 2017-10-17
FT: Add backbeat routes for replication of MPU to AWS

bennettbuchanan:


lib/routes/routeBackbeat.js
```diff
@@ -72,6 +73,9 @@ function _isValidLocation(request) {
 PUT /_/backbeat/metadata/<bucket name>/<object key>
 PUT /_/backbeat/data/<bucket name>/<object key>
 PUT /_/backbeat/multiplebackendputobject/<bucket name>/<object key>
+PUT /_/backbeat/multiplebackendinitiatempu/<bucket name>/<object key>
```
jonathan-gramain: I would advise to use POST rather than PUT for initiate MPU and complete MPU routes because those are command routes. Also related to this, `multiplebackendputobject` could be named `multiplebackendobject` since we disambiguate with the HTTP action (PUT in this case), but no need to change it since it's not part of this PR.

My personal preference to avoid having too many routes, or form a sort of hierarchy which can help to leverage some common processing code for all routes, would be to use AWS method of having a common route and the specific action in the query string, as e.g. `POST /_/backbeat/multiplebackend/bucket/key?initiatempu`. But at this point I may be a little bit nit-picky :laughing:

bennettbuchanan: I like this idea. I've updated this PR with those changes.

lib/routes/routeBackbeat.js
```diff
@@ -229,11 +233,108 @@ function multipleBackendPutObject(request, response, bucketInfo, objMd, log,
...
+function multipleBackendCompleteMPU(request, response, bucketInfo, objMd, log,
+    callback) {
+    const storageLocation = request.headers['x-scal-storage-class'];
+    const uploadId = request.headers['x-scal-upload-id'];
+    if (uploadId === undefined) {
+        log.error('bad request: missing upload-id header');
+        return callback(errors.BadRequest);
+    }
+    const data = [];
+    request.on('data', chunk => data.push(chunk.toString()));
```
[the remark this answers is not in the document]
bennettbuchanan: Thanks! Updated.

lib/routes/routeBackbeat.js
```diff
@@ -287,18 +429,9 @@ function routeBackbeat(clientIP, request, response, log) {
...
-                        bucketName: request.bucketName,
-                        objectKey: request.objectKey,
-                        resourceType: request.resourceType,
-                    });
-                    return next(errors.InvalidRequest);
-                }
-                return backbeatRoutes[request.method][request.resourceType](
-                    request, response, bucketInfo, objMd, log, next);
+            if (request.resourceType === 'multiplebackenddata') {
+                return backbeatRoutes[request.method][request.resourceType]
```
jonathan-gramain: Did you check the validity of `request.query.operation` earlier? If not it would be good to gracefully return an error if the operation does not exist.

lib/routes/routeBackbeat.js
```diff
@@ -59,19 +60,78 @@ function _getRequestPayload(req, cb) {
...
+        return errors.InvalidRequest.customizeDescription(errMessage);
+    }
+    return undefined;
 }
 
 /*
 PUT /_/backbeat/metadata/<bucket name>/<object key>
 PUT /_/backbeat/data/<bucket name>/<object key>
-PUT /_/backbeat/multiplebackendputobject/<bucket name>/<object key>
+PUT /_/backbeat/multiplebackenddata/<bucket name>/<object key>
```
jonathan-gramain: For those two PUT commands, why not having the simpler and more REST-friendly convention: if no PartNumber attribute is provided, it's a put of the whole object (since it's provided as a `x-scal-part-number` header it works by checking the header absence), otherwise it's a put on this specific part.

You can have a common handler for PUTs on `multiplebackenddata` and route simply to the proper function depending on the header presence.

lib/routes/routeBackbeat.js
bennettbuchanan: Good point. Included checks for this.

lib/routes/routeBackbeat.js
bennettbuchanan: As discussed, will leave as is for now. 😸

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine 
SCALITY_BACKBEAT_BRANCH=ft/S3C-971/update-client-and-objectMD
try

bennettbuchanan:
E2E passed on retry: http://ci.ironmann.io/gh/scality/Integration/15928.

bennettbuchanan:
@ironman-machine 
SCALITY_BACKBEAT_BRANCH=
try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
Passed all E2E tests, but retried by CircleCi: http://ci.ironmann.io/gh/scality/Integration/16060.

PR #945, opened 2017-10-18
FT: Add backbeat route for putting delete marker

bennettbuchanan:


lib/routes/routeBackbeat.js
```diff
@@ -229,11 +230,28 @@ function multipleBackendPutObject(request, response, bucketInfo, objMd, log,
...
+function multipleBackendDeleteObject(request, response, bucketInfo, objMd,
+    log, callback) {
+    const storageLocation = request.headers['x-scal-storage-class'];
+    const objectGetInfo = {
+        key: request.objectKey,
+        dataStoreName: storageLocation,
+    };
+    const reqUids = log.getSerializedUids();
+    return multipleBackendGateway.delete(objectGetInfo, reqUids, err => {
+        if (err) {
```
electrachong: do you want to add a log at this layer after all?

lib/routes/routeBackbeat.js
```diff
@@ -229,11 +231,40 @@ function multipleBackendPutObject(request, response, bucketInfo, objMd, log,
...
+        return _respond(response, dataRetrievalInfo, log, callback);
+    });
+}
+
 const backbeatRoutes = {
     PUT: {
         data: putData,
         metadata: putMetadata,
         multiplebackendputobject: multipleBackendPutObject,
+        multiplebackenddeleteobject: multipleBackendDeleteObject,
```
philipyoo: I think this needs to be removed?

bennettbuchanan: Thanks, removed.

lib/routes/routeBackbeat.js

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E tests passed (CircleCi retried): http://ci.ironmann.io/gh/scality/Integration/16085#build-parameters.

PR #946, opened 2017-10-18
FT: Add replication of object tags

bennettbuchanan:


bennettbuchanan:
@ironman-machine 
SCALITY_ARSENAL_BRANCH=ft/S3C-972/add-datastoreVersionId-to-replicationInfo
try

bennettbuchanan:
@ironman-machine try
SCALITY_ARSENAL_BRANCH=
try

bennettbuchanan:
@ironman-machine try
SCALITY_ARSENAL_BRANCH=
try

lib/routes/routeBackbeat.js
```diff
@@ -373,20 +380,78 @@ function completeMultipartUpload(request, response, log, callback) {
...
+    const err = _checkMultipleBackendRequest(request, log);
+    if (err) {
+        return callback(err);
+    }
+    const storageLocation = request.headers['x-scal-storage-class'];
+    const dataStoreVersionId = request.headers['x-scal-data-store-version-id'];
+    const data = [];
+    request.on('data', chunk => data.push(chunk));
+    request.on('end', () =>
+        parseTagXml(Buffer.concat(data), log, (err, tags) => {
```
rahulreddy: > If totalLength is not provided, it is calculated from the Buffer instances in list. This however causes an additional loop to be executed in order to calculate the totalLength, so it is faster to provide the length explicitly if it is already known.

You should aggregate each chunk's length and then provide it to the Buffer.concat method.

lib/api/objectPutTagging.js
```diff
@@ -78,10 +78,11 @@ function objectPutTagging(authInfo, request, log, callback) {
             const params = objectMD.versionId ? { versionId:
               objectMD.versionId } : {};
             const replicationInfo =
-                getReplicationInfo(objectKey, bucket, true);
+                getReplicationInfo(objectKey, bucket, true, 0, 'PUT_TAGGING');
```
rahulreddy: Please create a constant after the requires, this way we don't have to hunt for these constants to maintain them.
```js
const REPLICATION_ACTION = 'PUT_TAGGING';
```

lib/api/objectDeleteTagging.js
```diff
@@ -71,10 +71,12 @@ function objectDeleteTagging(authInfo, request, log, callback) {
             objectMD.tags = {};
             const params = objectMD.versionId ? { versionId:
               objectMD.versionId } : {};
-            const replicationInfo = getReplicationInfo(objectKey, bucket, true);
+            const replicationInfo = getReplicationInfo(objectKey, bucket, true,
+                0, 'DELETE_TAGGING');
```
rahulreddy: Please create a constant after the requires, this way we don't have to hunt for these constants to maintain them.
```js
const REPLICATION_ACTION = 'DELETE_TAGGING';
```

bennettbuchanan: Updated.

lib/api/objectPutTagging.js
bennettbuchanan: Updated.

lib/routes/routeBackbeat.js

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

PR #952, opened 2017-10-20
FIX: putPart - back to metadata version 3

nicolas2bert:


lib/api/objectPutPart.js
```diff
@@ -293,17 +293,13 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
+                'md-model-version': 3,
                 partLocations,
                 'key': partKey,
                 'last-modified': new Date().toJSON(),
                 'content-md5': hexDigest,
                 'content-length': size,
-                // NOTE: We can go back to Version 3 since we get the number of
-                // subparts from locations array
-                // NOTE: remove 'number-subparts'
-                'number-subparts': dataGetInfo.numberSubParts,
```
LaurenSpiegel: please confirm this is not used anywhere in the code (searching s3 and arsenal).

nicolas2bert: I checked, we are not

LaurenSpiegel:
@ironman-machine r+

PR #954, opened 2017-10-20
chore: clean tests to use multiple backend utils

electrachong:
I copied functions and some constants to the utils for previous pull requests, but left the clean-up to afterwards so the pull request wouldn't be too long. This is the clean-up PR.


electrachong:
@ironman-machine try

electrachong:
End to end passed although auto-retried due to a bug. Merging manually

PR #956, opened 2017-10-20
ft: enable versioning to aws backend for mpus

electrachong:


lib/api/completeMultipartUpload.js
```diff
@@ -18,8 +18,7 @@ const { versioningPreprocessing, checkQueryVersionId }
 const metadata = require('../metadata/wrapper');
 const services = require('../services');
 const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
-const { checkAzureBackendMatch, skipMpuPartProcessing } =
```
electrachong: maybe need to remove checkAzureBackendMatch if it's not used anywhere else

lib/data/wrapper.js
```diff
@@ -59,6 +59,26 @@ if (config.backends.data === 'mem') {
...
+    const skipMethods = { PUT: true, POST: true };
+    if (!Array.isArray(locations) || !locations[0] ||
+        !locations[0].dataStoreType) {
+        return false;
+    }
+    const isSkipBackend = skipBackend[locations[0].dataStoreType];
+    const isMatchingBackends =
+        locations[0].dataStoreName === newObjDataStoreName;
+    const isSkipMethod = skipMethods[requestMethod];
+    if (!isSkipBackend || !isMatchingBackends || !isSkipMethod) {
```
bennettbuchanan: Minor: the result can just be returned instead of using an if condition.

lib/data/external/AwsClient.js
```diff
@@ -327,6 +329,12 @@ class AwsClient {
                   `AWS: ${err.message}`)
                 );
             }
+            if (!completeMpuRes.VersionId) {
```
bennettbuchanan: Does versioning have to be enabled on the bucket to use multiple backend routes? If so, I guess I would not expect the complete MPU operation to succeed if also receiving this error. But maybe we want to be flexible here.

electrachong: Versioning has to be enabled on AWS to use multiple backends with AWS. That's why we return an error if no version id is returned (means versioning was suspended at some point)

lib/data/wrapper.js
electrachong: Thanks!

lib/data/external/AwsClient.js
LaurenSpiegel: are we doing the same on puts?

lib/data/wrapper.js
```diff
@@ -59,6 +59,23 @@ if (config.backends.data === 'mem') {
  */
 const MAX_RETRY = 2;
 
+// This check is done because on a put, complete mpu or copy request to AWS,
+// if the object already exists on AWS, the existing object should not be
```
LaurenSpiegel: what about azure?

lib/data/wrapper.js
```diff
@@ -59,6 +59,23 @@ if (config.backends.data === 'mem') {
...
+// deleted, which is the functionality for all other backends
+function _shouldSkipDelete(locations, requestMethod, newObjDataStoreName) {
+    const skipBackend = externalBackends;
```
LaurenSpiegel: what does this do?

lib/data/wrapper.js
```diff
@@ -59,6 +59,23 @@ if (config.backends.data === 'mem') {
...
+    const skipMethods = { PUT: true, POST: true };
+    if (!Array.isArray(locations) || !locations[0] ||
+        !locations[0].dataStoreType) {
+        return false;
```
LaurenSpiegel: why do you have to still delete if there is no data?

lib/data/external/AwsClient.js
electrachong: @LaurenSpiegel yes, we are

lib/data/wrapper.js
electrachong: Added Azure to the comment.

lib/data/wrapper.js
electrachong: Nothing much, it's leftover from the original check. I removed it

lib/data/wrapper.js
electrachong: The batch delete uses an async.eachLimit method so it won't do anything for an empty array even if we return `false`.
I return early if it's an empty array here so we won't get an error below accessing the `dataStoreType` and `dataStoreName` properties of the 0th index.

electrachong:
@ironman-machine try

electrachong:
end to end passed here: http://ci.ironmann.io/gh/scality/Integration/15988#build-parameters

PR #957, opened 2017-10-23
ft: enable aws backend object copy w/ versioning

electrachong:


lib/api/objectCopy.js
```diff
@@ -409,9 +407,14 @@ function objectCopy(authInfo, request, sourceBucket,
                     // object with same name, so long as the source is not
                     // the same as the destination
                     if (!sourceIsDestination && dataToDelete) {
-                        data.batchDelete(dataToDelete, request.method, null,
-                                logger.newRequestLoggerFromSerializedUids(
-                                    log.getSerializedUids()));
+                        const newDataStoreName =
+                            Array.isArray(destDataGetInfoArr) &&
```
LaurenSpiegel: don't we have a cleaner way to access the destination data store name?

electrachong: it looks like we make sure destDataGetInfoArray is an array (https://github.com/scality/S3/pull/957/commits/0cc254142d1b3fe833211dd379eef6d6bd31d819#diff-5ce9bf7b97556522975c2fae1a6944efL283) so I guess we don't need all the checks.

LaurenSpiegel: i meant, don't we have separate metadata of the new data store so we don't have to dig into the location array?

lib/data/external/AwsClient.js
```diff
@@ -327,6 +329,12 @@ class AwsClient {
                   `AWS: ${err.message}`)
                 );
             }
+            if (!completeMpuRes.VersionId) {
```
bennettbuchanan: This is from the previous commit?

electrachong: Yeah, if we merge the other pr first and rebase I think it should disappear from the diff
I left the change there because I need some of the other changes to this file for this PR (the `missingVerIdInternalError` const)

lib/api/objectCopy.js
electrachong: we store the metadata just before this batch delete step, but we pass the same `destDataGetInfoArr`as the object location to store and `services.metadataStoreObject` doesn't return the object MD, so we'd have to make a separate call to get the new object MD to get the location directly from the metadata (unless we want to refactor)

electrachong: I do see there is a separate `dataStoreName` property for object MD but I don't think we set it for the new metadata in objectCopy...

LaurenSpiegel: we should be. not even in @nicolas2bert's new pr?

electrachong: I think Nicolas introduces a change in his PR where he gets the dest location constraint which we might be able to use to set the `dataStoreName` prop appropriately. So far we're only copying an object to a destination location based on the request/source objMd location header (depending on whether we use REPLACE or COPY), I think.

LaurenSpiegel: by so far you mean before Nicolas's pr? The three of us should discuss tomorrow.

electrachong:
@ironman-machine try

electrachong:
end to end passed

lib/api/objectCopy.js
```diff
@@ -311,20 +313,23 @@ function objectCopy(authInfo, request, sourceBucket,
...
-            const destLocationConstraintName = backendInfoObjDest.controllingLC;
-
-            // skip if source and dest and location constraint the same
+            // skip if source and dest and location constraint the same and
+            // versioning is not enabled
             // still send along serverSideEncryption info so algo
             // and masterKeyId stored properly in metadata
-            if (sourceIsDestination && storeMetadataParams.locationMatch) {
+            if (sourceIsDestination && storeMetadataParams.locationMatch
+                && !isVersionedObj) {
```
LaurenSpiegel: @nicolas note this.

PR #962, opened 2017-10-24
S3C-1003 FX: COPY External backends encryption

nicolas2bert:
We need to exclude certain scenarios from using AWS/Azure server-side copy (for copy object and copy part) due to encryption.
Server side copy should only be allowed:
1) if source object and destination object are both on aws or both on azure.
2) if azure to azure, must be the same storage account
3) if the source bucket is not an encrypted bucket and the destination bucket is not an encrypted bucket (unless the copy is all within the same bucket).
Also, we should make sure we are sending the sse header to aws if the destination location has SSE set to true in the locationConfig.

### DIFF
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -245,6 +245,11 @@ const data = {
         if (_shouldSkipDelete(locations, requestMethod, newObjDataStoreName)) {
             return;
         }
+        log.trace('initiating batch delete', {
+            keys: locations,
+            implName,
+            method: 'batchDelete',
+        });
         async.eachLimit(locations, 5, (loc, next) => {
             data.delete(loc, log, next);
         },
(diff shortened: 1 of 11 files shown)


lib/data/wrapper.js
```diff
@@ -305,6 +305,62 @@ const data = {
...
+    destBackendInfo, log, cb) => {
+        if (serverSideEncryption) {
+            return kms.createCipherBundle(
+            serverSideEncryption,
+            log, (err, cipherBundle) => {
+                if (err) {
+                    log.debug('error getting cipherBundle');
+                    return cb(errors.InternalError);
+                }
+                return data.put(cipherBundle, stream,
```
electrachong: minor: this and the `data.put` call below look the same except you are not passing a cipherBundle, so you might be able to refactor them into a common function / callback, or use async to reduce the lines of code. Up to you if it's worth it though

lib/data/wrapper.js
```diff
@@ -323,19 +379,22 @@ const data = {
...
      * @param {function} cb - callback
      * @returns {function} cb - callback
      */
     copyObject: (request,
       sourceLocationConstraintName, storeMetadataParams, dataLocator,
-      dataStoreContext, destBackendInfo, serverSideEncryption, log, cb) => {
+      dataStoreContext, destBackendInfo, sourceBucketMD, destBucketMD, log,
+      cb) => {
+        const serverSideEncryption = destBucketMD.getServerSideEncryption();
         if (config.backends.data === 'multiple') {
```
electrachong: you could reduce the amount of nesting if you get the destLocationConstraintName before and do:
```
if (config.backends.data === 'multiple' && utils.externalBackendCopy(sourceLocationConstraintName,
destLocationConstraintName, sourceBucketMd, destBucketMD)) {
...
}
```
or pass `config.backends.data` to `utils.externalBackendCopy` to be included in the check

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -13,6 +13,7 @@ const { describeSkipIfNotMultiple, awsS3, memLocation, awsLocation,
     require('../utils');
 const bucket = 'buckettestmultiplebackendobjectcopy';
 const bucketAws = 'bucketawstestmultiplebackendobjectcopy';
+const bucketEncrypted = 'bucketenryptedtestmultiplebackendobjectcopy';
```
LaurenSpiegel: can we call this awsserversideencryptionbucket?  It's confusing to call it bucketEncrypted since that implies our bucket encryption.

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -292,6 +304,52 @@ function testSuite() {
             });
         });
 
+        it('should copy an object on AWS with encryption', done => {
```
LaurenSpiegel: please be more specific about what type of encryption -- aws server side encryption

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -292,6 +304,52 @@ function testSuite() {
...
+                    assert.strictEqual(result.CopyObjectResult.ETag,
+                        `"${correctMD5}"`);
+                    assertGetObjects(key, bucket, awsLocation, copyKey, bucket,
+                        awsLocationEncryption, copyKey, 'REPLACE', false, awsS3,
+                        awsLocation, done);
+                });
+            });
+        });
+
+        it('should copy an object on AWS with bucket encryption', done => {
```
LaurenSpiegel: same

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -105,7 +105,7 @@ callback) {
     });
 }
 
-describeSkipIfNotMultiple('MultipleBackend object copy: AWS',
+describeSkipIfNotMultiple.only('MultipleBackend object copy: AWS',
```
LaurenSpiegel: no!!!

nicolas2bert:
@ironman-machine try

nicolas2bert:
E2E retried here: http://ci.ironmann.io/gh/scality/Integration/16423#build-parameters

PR #963, opened 2017-10-24
Fix: backwards compat on copy part

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine try

lib/api/objectPutCopyPart.js
```diff
@@ -115,8 +115,15 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                             errors.NoSuchKey;
                         return next(err, destBucketMD);
                     }
-                    const sourceLocationConstraintName =
-                      sourceObjMD.dataStoreName;
+                    let sourceLocationConstraintName =
+                        sourceObjMD.dataStoreName;
+                    // for backwards compatibility before storing dataStoreName
+                    if (!sourceLocationConstraintName &&
```
jonathan-gramain: In a future refactoring, I would prefer to see this compat handled by `ObjectMD`/`ObjectMDLocation` classes instead of here so that we don't scatter md compatibility code is various places, but fine for now since it implies using ObjectMD in place of raw metadata. Maybe add a TODO?

LaurenSpiegel:
@ironman-machine try

PR #966, opened 2017-10-26
FWD: rel/7.1 to master

bennettbuchanan:


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
E2E passed, but CircleCi retried: http://ci.ironmann.io/gh/scality/Integration/16270.

PR #967, opened 2017-10-26
Ft/external health checks

nicolas2bert:
- Instead of returning a 500 on a deep health check due to an error from AWS/Azure, we will return a 200 but in the body of the response, return the error. 

- Each s3 server worker will not call a deep health check to Azure/AWS if it has been less than 1 minute since its last health check. It will just return whatever the last response it got from AWS/Azure until it calls AWS/Azure again.

- For regular operations from s3 server to AWS/Azure, s3 server will return a 503 error if there is an error from AWS/Azure so that the load balancer does not blacklist that s3 server because of an AWS/Azure error.


tests/multipleBackend/backendHealthcheckResponse.js
```diff
@@ -10,15 +10,15 @@ const locConstraints = Object.keys(config.locationConstraints);
...
             assert.strictEqual(err, null,
                 `Expected success but got error ${err}`);
             done();
         });
     });
     it('should return result for every location constraint in ' +
     'locationConfig and at least one of every external locations', done => {
-        clientCheck(log, (err, results) => {
+        clientCheck(true, log, (err, results) => {
             locConstraints.forEach(constraint => {
```
LaurenSpiegel: will you please add a test where you set clientCheck to false? will you please also unit test the utility function for the healthchecks to make sure we will get a refreshed response after a minute?  You'll have to mock the checkThisOne.healthcheck function so you are just returning different responses.

nicolas2bert:
@ironman-machine try

nicolas2bert:
@ironman-machine try

PR #968, opened 2017-10-27
rf: azure location config, use account / key

electrachong:


lib/Config.js
```diff
@@ -51,6 +53,42 @@ function restEndpointsAssert(restEndpoints, locationConstraints) {
...
+    });
+    [storageAccessKeyFromEnv, azureStorageAccessKey].forEach(value => {
+        if (value !== undefined) {
+            assert(azureAccessKeyRegex.test(value),
+                `bad location constraint: "${location}" ` +
+                'azureStorageAccessKey is not a valid base64 string');
+        }
+    });
+    assert(azureContainerName, 'bad location constraint: ' +
+        `"${location}" azureContainerName must be defined`);
```
bennettbuchanan: This would assert false if `locationObj.details.azureContainerName` was defined as something falsy. I don't know why someone would set their config that way, but I guess it's possible.

tests/unit/testConfigs/locConstraintAssert.js
```diff
@@ -103,4 +103,125 @@ describe('locationConstraintAssert', () => {
...
+            bucketMatch: false,
+            azureContainerName: 'test',
+        });
+        assert.doesNotThrow(() => {
+            locationConstraintAssert({ 'azurefaketest': locationConstraint,
+            'us-east-1': usEast1 });
+        },
+        '/should not throw for a valid azure location constraint/');
+    });
+    it('should throw error for if type is azure and azureContainerName is ' +
```
bennettbuchanan: ~for if~ if

lib/Config.js
electrachong: Made it more explicit to check if it is actually undefined and added some checks to make sure it is a string. The isValidBucketName function will check for length (make sure it's not an empty string)

tests/unit/testConfigs/locConstraintAssert.js
electrachong: thanks, updated those

lib/Config.js
```diff
@@ -751,20 +799,34 @@ class Config extends EventEmitter {
...
     }
 
     isSameAzureAccount(locationConstraintSrc, locationConstraintDest) {
-        return locationConstraintDest ?
-        this.getAzureEndpoint(locationConstraintSrc) ===
-        this.getAzureEndpoint(locationConstraintDest) : true;
+        if (!locationConstraintDest) {
+            return true;
+        }
+        const accountNames = [locationConstraintSrc, locationConstraintDest]
```
jonathan-gramain: style: for a simple comparison, I think calling twice `getAzureStorageAccountName` and comparing the results would be more straigthtforward than relying on a map.

tests/functional/aws-node-sdk/test/multipleBackend/utils.js
```diff
@@ -61,44 +61,34 @@ const utils = {
...
+        azureStorageAccountName: 'AZURE_STORAGE_ACCOUNT_NAME',
+        azureStorageAccessKey: 'AZURE_STORAGE_ACCESS_KEY',
+    };
 
-    if (isTestingAzure) {
-        if (process.env[`${azureLocation}_AZURE_BLOB_SAS`]) {
-            azureBlobSAS = process.env[`${azureLocation}_AZURE_BLOB_SAS`];
-            isTestingAzure = true;
-        } else if (config.locationConstraints[azureLocation] &&
+    const isTestingAzure = Object.keys(envMap).every(key => {
```
jonathan-gramain: Maybe you can just rely on `azureStorageAccountName` presence, since an azure test would not make sense without the access key set as well.

lib/data/external/AzureClient.js
```diff
@@ -11,11 +11,12 @@ const azureMpuUtils = s3middleware.azureHelper.mpuUtils;
...
-        this._azureBlobEndpoint = config.azureBlobEndpoint;
-        this._azureBlobSAS = config.azureBlobSAS;
+        this._azureStorageCredentials = config.azureStorageCredentials;
         this._azureContainerName = config.azureContainerName;
-        this._client = azure.createBlobServiceWithSas(
-          this._azureBlobEndpoint, this._azureBlobSAS);
+        this._client = azure.createBlobService(
+            this._azureStorageCredentials.storageAccountName,
+            this._azureStorageCredentials.storageAccessKey);
+        this._azureStorageEndpoint = this._client.host.primaryHost;
```
LaurenSpiegel: are we relying on a documented property of the azure node sdk here?  do they have a method for the endpoint?  is this the best way to do this?

lib/data/locationConstraintParser.js
```diff
@@ -74,11 +75,12 @@ function parseLC() {
             clients[location].clientType = 'aws_s3';
         }
         if (locationObj.type === 'azure') {
-            const azureBlobEndpoint = config.getAzureEndpoint(location);
-            const azureBlobSAS = config.getAzureBlobSas(location);
+            assert(locationObj.details, 'bad location constraint: details ' +
```
LaurenSpiegel: why is this assertion here rather than in config.js?

lib/data/external/AzureClient.js
electrachong: Doesn't seem documented, they also don't have a method (they just access the StorageServiceClient.host.primaryHost property to get the value when sending a request). Maybe it's safest to require the blob endpoint explicitly again?

constants.js
```diff
@@ -111,6 +111,10 @@ const constants = {
     mpuMDStoredExternallyBackend: { aws_s3: true },
     /* eslint-enable camelcase */
     mpuMDStoredOnS3Backend: { azure: true },
+    azureAccountNameRegex: /^[a-z0-9]{3,24}$/,
+    azureAccessKeyRegex:
+        new RegExp('^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==' +
```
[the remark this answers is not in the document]
electrachong: I think we can just use the general base64Regex for the azureAccessKey check then.
Thanks for the tip

lib/Config.js
electrachong: That's true, I updated it

tests/functional/aws-node-sdk/test/multipleBackend/utils.js
electrachong: We also create the blobservice client below, and we need to get the azureStorageAccessKey (stored in params in the every step) for that.

lib/data/external/AzureClient.js
LaurenSpiegel: do you have to tell the client the endpoint?  since it has this primary concept loaded, why do you need to tell it what it knows?

electrachong: we don't need it for the blob storage client, there is so far one case where we use it:
for the copy source uri https://github.com/scality/S3/pull/968/files#diff-96a8cd65c80ea7fb9b908a81388494d4R366
We always use a put object if copying from one account to another (instead of copy object) so the storage endpoint for the copy source uri should be the same as the destination endpoint

electrachong:
@ironman-machine try

electrachong:
end to end retried here: http://ci.ironmann.io/gh/scality/Integration/16559

electrachong:
@ironman-machine try

PR #971, opened 2017-10-29
MAINT: Too much log ci

alexandre-merle:
This specific log fill 10K of the log file into the CI


LaurenSpiegel:
thanks @alexandre-merle. Fixing in another pr.

PR #972, opened 2017-10-29
FIX: Locations max issue

alexandre-merle:
For parts with many locations, setPartRanges was going one location more
than needed, resulting in sending more data than the content-length.

### DIFF
--- a/lib/api/apiUtils/object/setPartRanges.js
+++ b/lib/api/apiUtils/object/setPartRanges.js
@@ -37,7 +37,7 @@ function setPartRanges(dataLocations, outerRange) {
     const max = end - begin + 1;
     let total = 0;
     for (let i = 0; i < dataLocations.length; i++) {
-        if (total >= max) {
+        if (total >= (max - 1)) {
             break;
         }
         const partStart = parseInt(dataLocations[i].start, 10);


alexandre-merle:
@ironman-machine try

alexandre-merle:
failed, but fixing the first issue

alexandre-merle:
@ironman-machine
SCALITY_BACKBEAT_BRANCH=test/aws-sdk-version
try

alexandre-merle:
@ironman-machine try

alexandre-merle:
@ironman-machine 
SCALITY_INTEGRATION_BRANCH=test/backbeat-issue
try

alexandre-merle:
@ironman-machine try

alexandre-merle:
@ironman-machine try

alexandre-merle:
@ironman-machine r+

alexandre-merle:
oups, miss command

alexandre-merle:
@ironman-machine try

alexandre-merle:
@ironman-machine 
SCALITY_INTEGRATION_BRANCH=test/backbeat-wait-complete
try

LaurenSpiegel:
How did this come up as an issue?

lib/api/apiUtils/object/setPartRanges.js
```diff
@@ -37,7 +37,7 @@ function setPartRanges(dataLocations, outerRange) {
     const max = end - begin + 1;
     let total = 0;
     for (let i = 0; i < dataLocations.length; i++) {
-        if (total >= max) {
+        if (total >= (max - 1)) {
```
LaurenSpiegel: On line 36, it says we are adding 1 because range is inclusive. Why this change?

alexandre-merle:
@LaurenSpiegel i have done a lot of tests this week end, and this is an issue i see in the container 5 of backbeat, this PR fix the issue.
For the context, backbeat was requesting a location start: 11, size: 11 and get back 12 character from S3, which leads to instable connections, 'data backend' econnreset, and other stuff

alexandre-merle:
@LaurenSpiegel the related integration tests you can look into is http://ci.ironmann.io/gh/scality/Integration/16518#artifacts, you have a network trace inside it showing the bug

lib/api/apiUtils/object/setPartRanges.js
jonathan-gramain: This seems wrong, the problem is that we don't specify clearly the convention of whether `end` is inclusive or exclusive, ending up having different callers using different conventions. See e.g. https://github.com/scality/S3/blob/master/lib/api/objectGet.js#L188, where the part end boundary is exclusive but `byteRange` is inclusive. I haven't gone through other callers, my gut feeling is that it's meant to be inclusive (hence the comment about `+1` in `setPartRanges`).

I suggest we should go through those callers and fix if they don't follow the convention (hopefully only the part get is in this case but we never know), and document whether it's inclusive or exclusive in JSDoc of `setPartRanges` to avoid future mistakes.

jonathan-gramain: Btw, thanks @alexandre-merle for nailing down the issue :+1:

jonathan-gramain:
Please review https://github.com/scality/S3/pull/974, if that's ok and passes the tests it should replace this PR.

PR #973, opened 2017-10-30
bf: fix possible timeout in unit tests

jonathan-gramain:
Prevent S3 unit tests from trying to access a local dmd which is not
running, by adding an option to BucketFileInterface constructor not to
attempt to open the database.

This is only for unit tests that use mocking with BucketFileInterface
and should be limited to this use.


nicolas2bert:
@ironman-machine try

PR #980, opened 2017-11-01
FIX: replicationEndpoints config with multiplebackend

bennettbuchanan:
When defining replication endpoints using an external backend during deployment, we want to support an optional list of servers when the `type` is a key in the `externalBackends` constant. As specified here: https://github.com/scality/Federation/blob/master/env/client-template/group_vars/all#L400.


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

PR #981, opened 2017-11-02
fix: use credentials for getRealAwsConfig if fit

electrachong:
We relied on having a `credentialsProfile` for multiple backend tests but ended up using the `credentials` config option in the Integration set-up. We may as well support both options to get the right AWS credentials for our aws multiple backend tests.


electrachong:
@ironman-machine try

PR #985, opened 2017-11-03
Forward: rel/6.3 to rel/6.4

alexanderchan-scality:
Forward from rel/6.3 to rel/6.4

FIX: JSON parse on an undefined


LaurenSpiegel:
@ironman-machine r+

PR #986, opened 2017-11-03
Forward rel/6.4 to rel/7.0

alexanderchan-scality:
Forward port from rel/6.4 to rel/7.0
Removes rel/6.4 getLocationTemp.js


tests/functional/aws-node-sdk/test/bucket/getLocationTemp.js
```diff
@@ -0,0 +1,55 @@
+import assert from 'assert';
```
LaurenSpiegel: the commit that introduced this test should not be forward ported.  this test was just for 6.4.

alexanderchan-scality:
@ironman-machine try

tests/unit/metadata/bucketclient/backend.js
```diff
@@ -0,0 +1,89 @@
+const assert = require('assert');
+const http = require('http');
+const BucketClientInterface =
+    require('../../../../lib/metadata/bucketclient/backend').default;
```
ThibaultRiviere: default?

alexanderchan-scality: Nice catch. Didn't see that when trying to forward 6.4 to 7.0

alexanderchan-scality:
Effectively, yes. The result is from merging 6.4 to 7.0, and needing to drop some commits as they are no longer required.

ThibaultRiviere:
Without using merge you will need to to it each time you want to forward something from `6.4` or using only cherry-pick  now.

LaurenSpiegel:
@ironman-machine try

Issue #367, opened 2016-11-08
COMPAT: AWS rejects overwriting Owner ID for objectPutACL


electrachong:
Should return `AccessDenied` if attempting to put object acl that changes the owner ID of the resource.

On **ceph/s3-tests**, this test fails:
* `FAIL: s3tests.functional.test_s3.test_object_giveaway`

PR #987, opened 2017-11-06
ft:Remove skip condition if E2E test

LaurenSpiegel:
To be merged after:

Integration #585 (run multiple backend tests, fix endpoint name) 
Arsenal #367 (fix stream destroy for azure put part)
S3 #982 (fix string being sent to utapi)
S3 #986 (forward port json parse bug fix and then forward port to 7.1 and master)


tests/locationConfigTests.json
```diff
@@ -5,7 +5,12 @@
         "legacyAwsBehavior": true,
         "details": {}
     },
-    "file-test": {
+    "scality-internal-file": {
+        "type": "file",
+        "legacyAwsBehavior": false,
+        "details": {}
+    },
+    "file": {
```
electrachong: I remember the collision between this ('file' as location name) and implName (sometimes 'file') caused some undetected bugs and we tried to rename the location names so there would be no such collisions, is there a reason to go back?

In any case since these prs are crucial for the rcs/ga I think we can let it slide but make a note to change these back to different names afterwards if there isn't a reason we are sticking to just 'file' or 'mem' for the location names.

electrachong: What is the difference between "scality-internal-file" and "file"? And "scality-internal-mem" and "mem"? It looks like we are usually using just the "scality-internal" ones for tests, are we using the "file" and "mem" locations for anything?

LaurenSpiegel: @electrachong, you are right. My fault. @nicolas2bert, will you please check if we can get rid of those added locations?

PR #989, opened 2017-11-06
S3C-1039 FX: location legacy

nicolas2bert:
TODO: add tests


lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -198,6 +218,9 @@ class BackendInfo {
         if (bucketLC) {
             return bucketLC;
         }
+        if (this.legacyLocationConstraint) {
```
LaurenSpiegel: we need to document this.  Add to the jira ticket and tag Jeffrey.

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -198,6 +218,9 @@ class BackendInfo {
...
+            return this.legacyLocationConstraint;
+        }
```
LaurenSpiegel: please also write a couple tests.

lib/api/apiUtils/object/BackendInfo.js
LaurenSpiegel: We have to be careful that for new buckets we aren't defaulting to creating them in legacy. The default for new buckets should be the endpoint and if the endpoint is not known, us-east-1.

nicolas2bert: We don't use this function when creating a bucket. 
We use this one https://github.com/scality/S3/blob/master/lib/api/bucketPut.js#L29. So it should be fine :)

tests/multipleBackend/objectPut.js
```diff
@@ -132,8 +132,9 @@ describeSkipIfE2E('objectPutAPI with multiple backends', function testSuite() {
         });
     });
 
-    it('should put an object to file based on request endpoint', done => {
-        put(null, null, 'localhost', () => {
+    it('should put an object to us-east-1 based on bucket location if ' +
```
LaurenSpiegel: why change? (should at least say us-east-1 which is file)

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -33,26 +34,24 @@ describe('BackendInfo class', () => {
             assert((res.description).indexOf('Bucket ' +
             'Location Error') > -1);
         });
-        it('should return object with applicable error if requestEndpoint ' +
+        it('should return isValid if requestEndpoint ' +
         'is invalid, no objectLocationConstraint or bucketLocationConstraint' +
-        'was provided and data backend is set to "scality"', () => {
+        'was provided, data backend is set to "scality" and legacy location ' +
```
LaurenSpiegel: shouldn't it be data backend set to multiple (rather than scality)?

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -33,26 +34,24 @@ describe('BackendInfo class', () => {
...
+        'contraint', () => {
             config.backends.data = 'scality';
             const res = BackendInfo.controllingBackendParam(
                 undefined, undefined, 'notValid', log);
-            assert.equal(res.isValid, false);
-            assert((res.description).indexOf('Endpoint ' +
-            'Location Error') > -1);
+            assert.equal(res.isValid, true);
         });
-        it('should return object with applicable error if requestEndpoint ' +
```
LaurenSpiegel: is there a way to add the error test now that we have legacy in our config?

tests/unit/multipleBackend/BackendInfo.js
nicolas2bert: I have a test with multiple right after this one.
`config.backends.data = 'scality'` is the same as `config.backends.data = 'multiple'`
https://github.com/scality/S3/blob/master/lib/Config.js#L762

tests/multipleBackend/objectPut.js
nicolas2bert: This test was not testing `putting an object to file based on request endpoint` because when creating a bucket without any location => bucket location === us-east-1
So object's data will be stored in us-east-1 location if object location constraint is empty.
Note: I m gonna replace undefined to null, it actually does not matter.

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -11,12 +13,15 @@ class BackendInfo {
     * @param {string | undefined } bucketLocationConstraint - location
     * constraint for bucket based on bucket metadata
     * @param {string} requestEndpoint - endpoint to which request was made
+    * @param {string | undefined } legacyLocationConstraint - legacy location
+    * constraint
     */
     constructor(objectLocationConstraint, bucketLocationConstraint,
-        requestEndpoint) {
+        requestEndpoint, legacyLocationConstraint) {
```
electrachong: I feel like you don't technically need to pass it as an argument since you can just call BackendInfo.getLegacyLocationConstraint() again

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -140,6 +145,12 @@ class BackendInfo {
             `Your bucket location "${escapeForXml(bucketLocationConstraint)}"` +
             ' is not in your location config - Please update.' };
         }
+        const legacyLocationConstraint =
+        BackendInfo.getLegacyLocationConstraint();
+        if (legacyLocationConstraint) {
+            log.trace('legacy location is valid');
+            return { isValid: true, legacyLocationConstraint };
```
electrachong: do you need the `legacyLocationConstraint` property returned for anything else than feeding it to the constructor to instantiate the class?

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -33,26 +34,24 @@ describe('BackendInfo class', () => {
             assert((res.description).indexOf('Bucket ' +
             'Location Error') > -1);
         });
-        it('should return object with applicable error if requestEndpoint ' +
+        it('should return isValid if requestEndpoint ' +
         'is invalid, no objectLocationConstraint or bucketLocationConstraint' +
-        'was provided and data backend is set to "scality"', () => {
+        'was provided, data backend is set to "scality" and legacy location ' +
+        'contraint', () => {
```
electrachong: *constraint

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -33,26 +34,24 @@ describe('BackendInfo class', () => {
...
-            'Location Error') > -1);
+            assert.equal(res.isValid, true);
         });
-        it('should return object with applicable error if requestEndpoint ' +
+        it('should return isValid if requestEndpoint ' +
         'is invalid, no objectLocationConstraint or ' +
         'bucketLocationConstraint was provided and ' +
-        'data backend is set to "multiple"', () => {
+        'data backend is set to "multiple" and legacy location ' +
+        'contraint', () => {
```
electrachong: *constraint

tests/unit/multipleBackend/BackendInfo.js
```diff
@@ -33,26 +34,24 @@ describe('BackendInfo class', () => {
...
             config.backends.data = 'multiple';
             const res = BackendInfo.controllingBackendParam(
                 undefined, undefined, 'notValid', log);
-            assert.equal(res.isValid, false);
-            assert((res.description).indexOf('Endpoint ' +
-            'Location Error') > -1);
+            assert.equal(res.isValid, true);
```
electrachong: why not `assert.strictEqual`

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -183,9 +202,12 @@ class BackendInfo {
     * Order of priority is:
     * (1) objectLocationConstraint,
     * (2) bucketLocationConstraint,
-    * (3) default locationConstraint for requestEndpoint  if requestEndpoint
+    * (3) legacyLocationContraint,
```
electrachong: *`legacyLocationConstraint`

lib/api/apiUtils/object/BackendInfo.js
nicolas2bert: I guess it's mostly for performance, I don't want to call it in `controllingBackendParam()` and  every time I call `getControllingLocationConstraint()`

lib/api/apiUtils/object/BackendInfo.js
```diff
@@ -183,9 +202,12 @@ class BackendInfo {
...
+    * (4) default locationConstraint for requestEndpoint  if requestEndpoint
     *     is listed in restEndpoints in config.json
-    * (4) default data backend
+    * (5) default data backend
+    * @param {string | undefined} legacyLocationContraint -
```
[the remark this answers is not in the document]
nicolas2bert: I forgot to remove it. thanks.

lib/api/apiUtils/object/BackendInfo.js
nicolas2bert: not for now.

tests/unit/multipleBackend/BackendInfo.js
nicolas2bert: Thanks

nicolas2bert:
@ironman-machine try

nicolas2bert:
E2E flakiness, Retried - http://ci.ironmann.io/gh/scality/Integration/16967

PR #990, opened 2017-11-07
 rf: create azure container if flightCheckOnStartUp

electrachong:
Add flight check during start-up option to client check. If a flight check, behavior is:
 *  - return an error to halt start-up if there is any error in the client check
 *  - check all external backend locations (instead of just one random location per external backend)
 *  - create a container for an Azure location if one is missing

### DIFF
--- a/lib/utilities/healthcheckHandler.js
+++ b/lib/utilities/healthcheckHandler.js
@@ -30,13 +30,16 @@ function writeResponse(res, error, log, results, cb) {
 /**
  * Calls each client's healthcheck and translates responses to
  * InternalError message if appropriate
- * @param {boolean} errorOnExternalBackend - whether to return a 500
- * due to a healthcheck fail on an external backend
+ * @param {boolean} flightCheckOnStartUp - whether client check is a flight
+ * check on startup. If performing a flight check, behavior is:
+ *  - return an error to halt start-up if there is any error in the client check
+ *  - check all external backend locations
+ *  - create a container for Azure location if one is missing
  * @param {object} log - werelogs logger instance
  * @param {function} cb - callback
  * @return {undefined}
  */
-function clientCheck(errorOnExternalBackend, log, cb) {
+function clientCheck(flightCheckOnStartUp, log, cb) {
     const clients = [
         data,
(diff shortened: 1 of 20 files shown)


tests/multipleBackend/backendHealthcheckResponse.js
```diff
@@ -1,39 +1,33 @@
...
-describe('Healthcheck response', () => {
-    it('should return no error with errorOnExternalBackend set to false',
-    done => {
-        clientCheck(true, log, err => {
-            assert.strictEqual(err, null,
-                `Expected success but got error ${err}`);
-            done();
-        });
-    });
+describe.only('Healthcheck response', () => {
```
electrachong: need to remove

lib/utilities/healthcheckHandler.js
```diff
@@ -30,13 +30,16 @@ function writeResponse(res, error, log, results, cb) {
 /**
  * Calls each client's healthcheck and translates responses to
  * InternalError message if appropriate
- * @param {boolean} errorOnExternalBackend - whether to return a 500
- * due to a healthcheck fail on an external backend
+ * @param {boolean} flightCheckOnStartUp - whether client check is a flight
+ * check on startup. If a flight check, behavior is:
```
JianqinWang: "If performing a flight check"?

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

dora-korpar:
How come we're not creating an AWS bucket if it doesn't exist in the flight check?

electrachong:
@dora-korpar The reasoning is we don't want a normal health check to have side effects (e.g. when nginx is querying the server every 2 minutes). We think of the flight check on start up as different than a normal healthcheck, since we are checking to make sure everything specified for clients/locations is operational and will stop the server if it is not.

electrachong:
@ironman-machine r+

PR #991, opened 2017-11-07
Forward rel/7.0 to rel/7.1

alexanderchan-scality:
Forward from rel/7.0 to rel/7.1

### DIFF
--- a/tests/functional/aws-node-sdk/test/multipleBackend/objectTagging.js
+++ b/tests/functional/aws-node-sdk/test/multipleBackend/objectTagging.js
@@ -154,6 +154,9 @@ function mpuWaterfall(params, cb) {
 
 describeSkipIfNotMultiple('Object tagging with multiple backends',
 function testSuite() {
+    if (!process.env.S3_END_TO_END) {
+        this.retries(2);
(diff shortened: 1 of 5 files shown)


nicolas2bert:
@ironman-machine r+

tests/functional/aws-node-sdk/test/multipleBackend/objectTagging.js
```diff
@@ -154,6 +154,9 @@ function mpuWaterfall(params, cb) {
 
 describeSkipIfNotMultiple('Object tagging with multiple backends',
 function testSuite() {
+    if (!process.env.S3_END_TO_END) {
```
LaurenSpiegel: why do we need this condition?

alexanderchan-scality: I had followed the form of what has been added to master, such that retries should not be allowed when the var is set. I see that it won't be used in rel/6.3, I will make the proper changes.

nicolas2bert:
@alexanderchan-scality, just wondering why has it never been merged?

PR #992, opened 2017-11-07
chore: add keep-alive to real awsclient config

electrachong:
May help save the time of re-establishing the connection between each request to AWS.


electrachong:
Merged in https://github.com/scality/S3/pull/987

PR #996, opened 2017-11-09
chore: omit timeout on first try - multi tests

electrachong:
On the last successful [end to end](http://ci.ironmann.io/gh/scality/Integration/16932), it took 2 hours to run multiple backend tests. Trying to see if omitting the timeout on the first attempt to get data directly from AWS can reduce the run-length at all.

### DIFF
--- a/tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
+++ b/tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
@@ -25,16 +25,29 @@ let s3;
 const awsTimeout = 30000;
 const retryTimeout = 10000;
 
+function _retryWithTimeouts(fn, numberRetry) {
+    const timeout = {
+        0: 0,
+        1: retryTimeout,
+        2: awsTimeout,
+    };
+    const maxRetries = 2;
+    try {
+        setTimeout(() => {
+            fn();
+        }, timeout[numberRetry]);
+    } catch (e) {
+        if (numberRetry !== maxRetries) {
+            _retryWithTimeouts(fn, numberRetry + 1);
+        } else {
+            throw e;
+        }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


electrachong:
@ironman-machine try

electrachong:
cancelled previous end-to-end, retry:
@ironman-machine try

electrachong:
@ironman-machine try

tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
```diff
@@ -25,16 +25,29 @@ let s3;
 const awsTimeout = 30000;
 const retryTimeout = 10000;
 
+function _retryWithTimeouts(fn, numberRetry) {
```
nicolas2bert: What is the second argument for? Won't it be `0` every time?

electrachong: on line 41 I call the same fn increasing the numberRetry argument

nicolas2bert: ouch, yes sorry.

nicolas2bert: I just thought you could have done like you did here: https://github.com/scality/S3/pull/996/files#diff-759441a4bc174becdbb400c56b574584R104. But it's the same.

electrachong: yes, I let the retryNumber be optional in the other functions because I was a bit lazy to edit all existing function calls to add the retryNumber argument.
In this case I had to replace all the setTimeout calls and remove the awsTimeout argument in the file so it wasn't too much extra work to expect the argument instead of letting it be optional.

PR #997, opened 2017-11-10
Forward rel/6.3 to rel/6.4

alexanderchan-scality:
Forward from rel/6.3 to rel/6.4
Adds a check for empty case in multiObjectDelete.


LaurenSpiegel:
@ironman-machine r+

PR #998, opened 2017-11-10
Forward rel/6.4 to rel/7.0

alexanderchan-scality:
Forward from rel/6.4 to rel/7.0
Adds a check for empty case in multiObjectDelete.


alexanderchan-scality:
@ironman-machine try

PR #999, opened 2017-11-11
FIX: Add more header support to replicate with AWS

bennettbuchanan:


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
Failed due to flaky 10,000 part MPU Scality backend replication test, which is known.

Rerunning now.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

lib/data/external/AwsClient.js
```diff
@@ -187,13 +198,30 @@ class AwsClient {
...
+    createMPU(key, metaHeaders, bucketName, websiteRedirectHeader, contentType,
+        cacheControl, contentDisposition, contentEncoding, log, callback) {
+        const metaHeadersTrimmed = {};
+        Object.keys(metaHeaders).forEach(header => {
+            if (header.startsWith('x-amz-meta-')) {
+                const headerKey = header.substring(11);
+                metaHeadersTrimmed[headerKey] = metaHeaders[header];
+            }
+        });
+        if (metaHeadersTrimmed !== {}) {
```
rahulreddy: This needs to be changed to `Object.keys.length > 0` as this condition always evaluates to true.
Note: What's the downside of sending empty object? I don't think it will throw an error.

bennettbuchanan: You're right, I will just assign the object here to remove this conditional. If it's empty, that will work.

bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
All E2E tests passed, although the CI timed out here: http://ci.ironmann.io/gh/scality/Integration/17012

PR #1001, opened 2017-11-13
FT: split external backend tests

nicolas2bert:
Depends on https://github.com/scality/Integration/pull/594


rahulreddy:
Closing. We will come up with a better approach after reviewing current test suite.

PR #1002, opened 2017-11-14
fix: error on versioning put to azure loc bucket

dora-korpar:
For buckets put to backend locations that do not support versioning (Azure atm), we should return a NotImplemented error if the user tries to put versioning on the bucket.

### DIFF
--- a/lib/api/bucketPutVersioning.js
+++ b/lib/api/bucketPutVersioning.js
@@ -6,6 +6,12 @@ const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const metadata = require('../metadata/wrapper');
 const { metadataValidateBucket } = require('../metadata/metadataUtils');
 const { pushMetric } = require('../utapi/utilities');
+const versioningNotImplBackends =
+    require('../../constants').versioningNotImplBackends;
+const { config } = require('../Config');
+
+const externalVersioningErrorMessage = 'We do not currently support putting ' +
+'a versioned object to a location-constraint of type Azure.';
 
 /**
  * Format of xml request:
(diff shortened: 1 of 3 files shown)


dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

lib/api/bucketPutVersioning.js
```diff
@@ -50,6 +56,16 @@ function _parseXML(request, log, cb) {
     });
 }
 
+function _checkBackendVersioningImplemented(bucket) {
+    const bucketLocation = bucket.getLocationConstraint();
+    const bucketLocationType = config.locationConstraints[bucketLocation].type;
```
nicolas2bert: You can use: `getLocationConstraintType()`
https://github.com/scality/S3/blob/master/lib/Config.js#L797

lib/api/bucketPutVersioning.js
```diff
@@ -50,6 +56,16 @@ function _parseXML(request, log, cb) {
     });
 }
 
+function _checkBackendVersioningImplemented(bucket) {
```
nicolas2bert: You can add this function in Config.js

dora-korpar: This isn't called on the bucket included in the locationConfig, though. It's called on the local S3 bucket

nicolas2bert:
@ironman-machine r+

PR #1003, opened 2017-11-15
fix: try-catch should catch getAws thrown errors

electrachong:
We tried to reduce the length of end to end runs before by skipping timeouts on first attempts to get data from AWS, but some thrown errors were not getting caught because try-catch scopes do not extend to code inside an asynchronous function's callback.

Refactored so we only use try-catch when we are nesting the statements which may throw errors we expect to be caught immediately within.


electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

electrachong:
@ironman-machine 
SCALITY_FEDERATION_BRANCH=skip/check-upgrade
try

PR #1004, opened 2017-11-15
FIX: S3C-1087 put and copy metadata for external backends

nicolas2bert:
- When copying to Azure we want to keep metadata up to date on the Azure side.

- Put and copying to Azure we want to trim `x-amz-meta`prefix from metadata header keys.

### DIFF
--- a/lib/data/external/utils.js
+++ b/lib/data/external/utils.js
@@ -112,6 +112,25 @@ const utils = {
         }
         return cb();
     },
+    translateAzureMetaHeaders(metaHeaders, tags) {
+        const translatedMetaHeaders = {};
+        if (tags) {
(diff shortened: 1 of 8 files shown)


lib/data/external/utils.js
```diff
@@ -112,6 +112,25 @@ const utils = {
         }
         return cb();
     },
+    translateMetaHeaders(metaHeaders, tags) {
```
electrachong: this only works for azure right? maybe translateAzureMetaHeaders?

nicolas2bert:
@ironman-machine r+

nicolas2bert:
@ironman-machine r+

nicolas2bert:
@ironman-machine r+

PR #774, opened 2017-06-16
bf: immutable optim option on S3 sproxyd client

jonathan-gramain:
This is a new option in scality/sproxydclient that explicitly enables
immutable optims, because they are now disabled by default it's now
necessary to provide the option.

Goes along with PR https://github.com/scality/sproxydclient/pull/119

PR #1006, opened 2017-11-16
forward port of #1005

jonathan-gramain:
forward port of #1005 (originally a backport of #774).

Tied to https://github.com/scality/sproxydclient/pull/129


harperscality:
Seems we don't want this in 6.4

PR #1010, opened 2017-11-17
FIX: Get datastoreVersionId from metadata

bennettbuchanan:
Depends on https://github.com/scality/backbeat/pull/135.

Object tagging operations that are made before the source object has completed replication do not have the `datastoreVersionId` in the Kafka entry. 

Because all operations for the multiple backend are serialized, when any tagging operation is being replicated we can guarantee that the source object's metadata will have been updated at this point. Hence we need to retrieve the `replicationInfo.datastoreVersionId` from the source S3 before making the request to an external location.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -1,9 +1,8 @@
 const url = require('url');
 const async = require('async');
 
-const { auth, errors, s3middleware } = require('arsenal');
+const { auth, errors } = require('arsenal');
 const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
-const { parseTagXml } = s3middleware.tagging;
 const vault = require('../auth/vault');
 const metadata = require('../metadata/wrapper');
 const locationConstraintCheck = require(
@@ -11,7 +10,8 @@ const locationConstraintCheck = require(
 const { dataStore } = require('../api/apiUtils/object/storeObject');
 const prepareRequestContexts = require(
 '../api/apiUtils/authorization/prepareRequestContexts');
-const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
+const { metadataValidateBucketAndObj,
+    metadataGetObject } = require('../metadata/metadataUtils');
 const { BackendInfo } = require('../api/apiUtils/object/BackendInfo');
 const { locationConstraints } = require('../Config').config;
 const multipleBackendGateway = require('../data/multipleBackendGateway');
@@ -89,11 +89,23 @@ function _checkMultipleBackendRequest(request, log) {
         log.error(errMessage);
         return errors.BadRequest.customizeDescription(errMessage);
     }
-    if ((operation === 'puttagging' || operation === 'deletetagging') &&
-        headers['x-scal-data-store-version-id'] === undefined) {
-        errMessage = 'bad request: missing x-scal-data-store-version-id header';
-        log.error(errMessage);
-        return errors.BadRequest.customizeDescription(errMessage);
+    if (operation === 'puttagging' || operation === 'deletetagging') {
+        if (headers['x-scal-data-store-version-id'] === undefined) {
+            errMessage =
+                'bad request: missing x-scal-data-store-version-id header';
+            log.error(errMessage);
+            return errors.BadRequest.customizeDescription(errMessage);
+        }
+        if (headers['x-scal-source-bucket'] === undefined) {
+            errMessage = 'bad request: missing x-scal-source-bucket header';
+            log.error(errMessage);
+            return errors.BadRequest.customizeDescription(errMessage);
+        }
+        if (headers['x-scal-source-version-id'] === undefined) {
+            errMessage = 'bad request: missing x-scal-source-version-id header';
+            log.error(errMessage);
+            return errors.BadRequest.customizeDescription(errMessage);
+        }
     }
     if (operation === 'putobject' &&
         headers['content-md5'] === undefined) {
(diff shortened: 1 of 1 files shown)


bennettbuchanan:
@ironman-machine 
SCALITY_BACKBEAT_BRANCH=fix/S3C-1100/update-headers-for-tagging-and-set-datastoreversionid
try

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3C-972/add-replication-to-aws-tests
try

lib/routes/routeBackbeat.js
```diff
@@ -439,15 +445,22 @@ function putObjectTagging(request, response, log, callback) {
...
-    request.on('data', chunk => {
-        totalLength += chunk.length;
-        data.push(chunk);
-    });
-    request.on('end', () =>
-        parseTagXml(Buffer.concat(data, totalLength), log, (err, tags) => {
+    const sourceVersionId = request.headers['x-scal-source-version-id'];
+    const sourceBucket = request.headers['x-scal-source-bucket'];
+    const sourceTags = request.headers['x-scal-tags'];
+    return metadataGetObject(sourceBucket, request.objectKey,
```
electrachong: did you consider only going to get the updated object metadata if `x-scal-data-store-version-id` is undefined from Backbeat? Or you decided it's better to be consistent and always fetch the updated metadata?

bennettbuchanan: I did consider that, but figured it would be simpler to always retrieve it from metadata here. That way, because the value may be `''` in the Kafka entry, we can remove the retrieval of `datastoreVersionId` in backbeat (and sending it as an additional 'x-scal-datastore-version-id' header) and avoid any conditional checks here.

lib/routes/routeBackbeat.js
```diff
@@ -90,8 +90,14 @@ function _checkMultipleBackendRequest(request, log) {
...
-        headers['x-scal-data-store-version-id'] === undefined) {
-        errMessage = 'bad request: missing x-scal-data-store-version-id header';
+        headers['x-scal-source-bucket'] === undefined) {
+        errMessage = 'bad request: missing x-scal-source-bucket header';
+        log.error(errMessage);
+        return errors.BadRequest.customizeDescription(errMessage);
+    }
+    if ((operation === 'puttagging' || operation === 'deletetagging') &&
+        headers['x-scal-source-version-id'] === undefined) {
+        errMessage = 'bad request: missing x-scal-version-id header';
```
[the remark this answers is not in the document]
bennettbuchanan: Thanks! Will update to `x-scal-source-version-id`.

My sense is that we can just validate that the headers were sent while assuming the values are correct from the Kafka entry.

lib/routes/routeBackbeat.js
```diff
@@ -90,8 +90,14 @@ function _checkMultipleBackendRequest(request, log) {
         return errors.BadRequest.customizeDescription(errMessage);
     }
     if ((operation === 'puttagging' || operation === 'deletetagging') &&
```
jonathan-gramain: The following structure would be slightly better, grouping all tests related to a subset of operations:

```
if (operation === 'puttagging' || operation === 'deletetagging') {
    if (A) {}
    if (B) {}
}
```

lib/routes/routeBackbeat.js
jonathan-gramain: Functionally it's equivalent, but there's a non-negligible cost to fetch the metadata on S3 because it may end up being a storage request (unless cached) where we already optimize metadata-only sync to avoid hitting the source S3. I would suggest not to do this request when the source version is readily available in kafka, it should be safe since source version shall be immutable for a particular object version.

Depending on workloads it may not be important or may not even go through this code, but in some kind of workloads it could make a significant difference (e.g. for mass-tagging of objects that already exist and have been synced, this will spare a metadata request for each object).

Code-wise it should be pretty simple with a wrapper of `metadataGetObject` branching to a `process.nextTick(cb, info)` when the version is available locally. It can be an extra test case as well but it's worth it IMO.

lib/routes/routeBackbeat.js

lib/routes/routeBackbeat.js
bennettbuchanan: Okay, I've updated this. I had thought that a request to retrieve Kafka entry values and metadata values would be about the same cost.

lib/routes/routeBackbeat.js
```diff
@@ -89,17 +89,23 @@ function _checkMultipleBackendRequest(request, log) {
...
-        log.error(errMessage);
-        return errors.BadRequest.customizeDescription(errMessage);
-    }
-    if ((operation === 'puttagging' || operation === 'deletetagging') &&
-        headers['x-scal-source-version-id'] === undefined) {
-        errMessage = 'bad request: missing x-scal-version-id header';
-        log.error(errMessage);
-        return errors.BadRequest.customizeDescription(errMessage);
+    if (operation === 'puttagging' || operation === 'deletetagging') {
+        if (headers['x-scal-data-store-version-id'] === undefined) {
```
electrachong: is it always `''` coming from backbeat and not `undefined` if we do not have the replicated version id stored in the kafka entry?

bennettbuchanan: It will be `''`.

bennettbuchanan: Replication info in the Kafka entry when the tagging is put after replication:

```
{\"status\":\"PENDING\",\"content\":[\"METADATA\",\"PUT_TAGGING\"],\"destination\":\"arn:aws:s3:::destination-bucket\",\"storageClass\":\"awsbackend\",\"role\":\"arn:aws:iam::887847504243:role/RoleForReplication-1510950202478\",\"storageType\":\"aws_s3\",\"dataStoreVersionId\":\"0hPAGSGpo7o03eQpKfvivoVHJHKlNkBJ\"}
```

and when tagging is put before replication:

```
{\"status\":\"PENDING\",\"content\":[\"METADATA\",\"PUT_TAGGING\"],\"destination\":\"arn:aws:s3:::destination-bucket\",\"storageClass\":\"awsbackend\",\"role\":\"arn:aws:iam::887847504243:role/RoleForReplication-1510951189163\",\"storageType\":\"aws_s3\",\"dataStoreVersionId\":\"\"}
```

lib/routes/routeBackbeat.js
```diff
@@ -128,6 +140,38 @@ function _checkMultipleBackendRequest(request, log) {
...
 }
 
+function handleTaggingOperation(request, response, method, dataStoreVersionId,
+    log, callback) {
+    const storageLocation = request.headers['x-scal-storage-class'];
+    const objectMD = {
+        dataStoreName: storageLocation,
+        location: [{ dataStoreVersionId }],
+    };
+    if (method === 'Put') {
```
electrachong: just a small comment, `method` may confuse a bit because the HTTP method for a put object tagging is actually `POST`. we could keep the operation semantics (`puttagging` or `deletetagging`) instead

bennettbuchanan: I agree and I've renamed it to `type`. I was following the naming scheme for the `objectTagging` method in the multiple backend gateway.

bennettbuchanan:
E2E passed with dependent branch: 'http://ci.ironmann.io/gh/scality/Integration/17401.

rahulreddy:
@ironman-machine clean r+

rahulreddy:
@ironman-machine 
SCALITY_BACKBEAT_BRANCH=fix/S3C-1100/update-headers-for-tagging-and-set-datastoreversionid
r- r+

rahulreddy:
@ironman-machine
SCALITY_BACKBEAT_BRANCH=fix/S3C-1100/update-headers-for-tagging-and-set-datastoreversionid
r- r+

PR #1019, opened 2017-11-22
Fix/s3 c 1120 aws proxy http payloads

rahulreddy:
**fix: use V2 auth for HTTP only AWS requests**

 To use an AWS location constraint in an HTTP only proxy environment in the context of Backbeat, V2 auth is used to avoid the body signing exception thrown by AWS SDK.


rahulreddy:
@ironman-machine try

PR #1023, opened 2017-11-29
FT: S3C-1065: sparse objects

ploki:
# FT: S3C-1065: sparse objects

## Description

handle null parts on sparse objects

### Motivation and context

Null parts are parts with null keys which are generated by the NFS server when growing a file with truncate

### Related issues

* scality/scality-nfsd#103
* scality/Arsenal#377
* scality/Integration#601

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -108,6 +108,10 @@ const constants = {
     legacyLocations: ['sproxyd', 'legacy'],
     /* eslint-disable camelcase */
     externalBackends: { aws_s3: true, azure: true },
+    // some of the available data backends  (if called directly rather
+    // than through the multiple backend gateway) need a key provided
+    // as a string as first parameter of the get/delete methods.
+    clientsRequireStringKey: { sproxyd: true, cdmi: true },
     // healthcheck default call from nginx is every 2 seconds
     // for external backends, don't call unless at least 1 minute
     // (60,000 milliseconds) since last call
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -8,6 +8,7 @@ const multipleBackendGateway = require('./multipleBackendGateway');
 const utils = require('./external/utils');
 const { config } = require('../Config');
 const MD5Sum = s3middleware.MD5Sum;
+const NullStream = s3middleware.NullStream;
 const assert = require('assert');
 const kms = require('../kms/wrapper');
 const externalBackends = require('../../constants').externalBackends;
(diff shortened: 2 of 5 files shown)


lib/data/wrapper.js
```diff
@@ -182,6 +183,10 @@ const data = {
         if (response) {
             objGetInfo.response = response;
         }
+        if (objGetInfo.key === null) {
```
LaurenSpiegel: do you have to set the key to null?  Would it be possible to set it to something more descriptive? Very legacy deployments would have objectGetInfo.key = undefined which could be confusing.

constants.js
```diff
@@ -108,6 +108,9 @@ const constants = {
     legacyLocations: ['sproxyd', 'legacy'],
     /* eslint-disable camelcase */
     externalBackends: { aws_s3: true, azure: true },
+    // some of the available data backends need a key provided as a string
```
LaurenSpiegel: "some of the available data backends (if called directly rather than through the multiple backend gateway) need..."

lib/data/wrapper.js
```diff
@@ -171,18 +172,27 @@ const data = {
     },
 
     get: (objectGetInfo, response, log, cb) => {
-        // If objectGetInfo.key exists the md-model-version is 2 or greater.
-        // Otherwise, the objectGetInfo is just the key string.
-        const objGetInfo = (implName === 'sproxyd' || implName === 'cdmi') ?
-            objectGetInfo.key : objectGetInfo;
+        const isMdModelVersion2 = typeof(objectGetInfo) === 'string';
+        const isRequiredStringKey = constants.clientsRequireStringKey[implName];
```
LaurenSpiegel: so beautiful i could cry 😄

lib/data/wrapper.js
```diff
@@ -171,18 +172,27 @@ const data = {
...
+        const key = isMdModelVersion2 ? objectGetInfo : objectGetInfo.key;
+        const clientGetInfo = isRequiredStringKey ? key : objectGetInfo;
         const range = objectGetInfo.range;
+
+        // If the key is explicitly set to null, the part to
+        // be read doesn't really exist and is only made of zeroes
+        // otherwise, the key is always defined
```
LaurenSpiegel: please add -- "This functionality is used by Scality-NFSD."

lib/data/wrapper.js
jonathan-gramain: even better -> `constants.clientsRequireStringKey[implName] || false` :)

lib/data/wrapper.js
```diff
@@ -219,16 +230,23 @@ const data = {
 
     delete: (objectGetInfo, log, cb) => {
         const callback = cb || log.end;
-        // If objectGetInfo.key exists the md-model-version is 2 or greater.
-        // Otherwise, the objectGetInfo is just the key string.
-        const objGetInfo = (implName === 'sproxyd' || implName === 'cdmi') ?
-            objectGetInfo.key : objectGetInfo;
+        const isMdModelVersion2 = typeof(objectGetInfo) === 'string';
```
jonathan-gramain: Eventually we would better use `ObjectMD` class to centralize compat stuff in a single location instead of having to deal with this at various locations and take the risk of screwing or forgetting to handle it sometimes. But of course this is not the scope here.

lib/data/wrapper.js
```diff
@@ -219,16 +230,23 @@ const data = {
...
-        });
-        _retryDelete(objGetInfo, log, 0, err => {
+            implName, key, method: 'delete' });
+        // If the key is explicitly set to null, the part to
+        // be deleted doesn't really exist.
+        // This functionality is used by Scality-NFSD.
+        // Otherwise, the key is always defined
+        assert(key === null || key !== undefined);
+        if (key === null) {
+            callback(null);
```
jonathan-gramain: Do you need to send back `null` here? Otherwise go for `callback()` instead.

lib/data/wrapper.js
```diff
@@ -252,7 +270,7 @@ const data = {
             method: 'batchDelete',
         });
         async.eachLimit(locations, 5, (loc, next) => {
-            data.delete(loc, log, next);
+            process.nextTick(() => data.delete(loc, log, next));
```
jonathan-gramain: Why adding `nextTick`?

ploki:
@ironman-machine after=[scality/Arsenal#377]

lib/data/wrapper.js
ploki: because there are a lot of ```assert.strictEqual(err, null);``` here and there

lib/data/wrapper.js
ploki: from the commit message:
```
Because of the data/wrapper batchDelete function becoming 'more' asynchronous
by calling data.delete from nextTick (that prevents a call stack overflow on
large array), some of the process.nextTick calls in the unit tests have been
replaced by setImmediate. (setImmediate is scheduled AFTER nextTick)
```

lib/data/wrapper.js
ploki: one hundred percent positive about this

lib/data/wrapper.js
ploki: already falsey if not true and consistent with the usage of other similarly defined constants

rahulreddy:
@ironman-machine 
r=aa34f86
after=https://github.com/scality/Arsenal/pull/377

rahulreddy:
@ironman-machine
r=aa34f86
after=scality/Arsenal#377

rahulreddy:
@ironman-machine 
r-
r=aa34f86

PR #1024, opened 2017-11-29
Bf/s3 c 1135 aws range

rahulreddy:
test: get object from AWS using range query

bf: fix range query format for GETs on AWS

### DIFF
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -125,7 +125,7 @@ class AwsClient {
             Bucket: this._awsBucketName,
             Key: key,
             VersionId: dataStoreVersionId,
-            Range: range,
+            Range: range ? `bytes=${range[0]}-${range[1]}` : null,
         }).on('success', response => {
             log.trace('AWS GET request response headers',
               { responseHeaders: response.httpResponse.headers });
--- a/tests/functional/aws-node-sdk/test/multipleBackend/get/get.js
+++ b/tests/functional/aws-node-sdk/test/multipleBackend/get/get.js
@@ -20,6 +20,7 @@ const bigObject = `bigObject-${Date.now()}`;
 const mismatchObject = `mismatch-${Date.now()}`;
 const body = Buffer.from('I am a body', 'utf8');
 const bigBody = Buffer.alloc(10485760);
+const bigBodyLen = bigBody.length;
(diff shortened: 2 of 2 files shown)


lib/data/external/AwsClient.js
```diff
@@ -125,7 +125,7 @@ class AwsClient {
             Bucket: this._awsBucketName,
             Key: key,
             VersionId: dataStoreVersionId,
-            Range: range,
+            Range: range ? `bytes ${range[0]}-${range[1]}` : null,
```
jonathan-gramain: The proper HTTP format for a full range request is `Range: bytes=A-B`, so I assume the AWS client expects the same.

lib/data/external/AwsClient.js
```diff
@@ -125,7 +125,7 @@ class AwsClient {
             Bucket: this._awsBucketName,
             Key: key,
             VersionId: dataStoreVersionId,
-            Range: range,
+            Range: range ? `bytes=${range[0]}-${range[1]}` : null,
```
bennettbuchanan: Not sure if it matters to AWS, but maybe use `undefined` instead of `null`.

lib/data/locationConstraintParser.js
```diff
@@ -43,8 +43,9 @@ function parseLC() {
                 new http.Agent({ keepAlive: true });
             const protocol = locationObj.details.https ? 'https' : 'http';
             const httpOptions = locationObj.details.proxy ?
-                { proxy: locationObj.details.proxy, agent: connectionAgent }
-                : { agent: connectionAgent };
+                { proxy: locationObj.details.proxy, agent: connectionAgent,
+                timeout: 0, retries: 0 }
```
LaurenSpiegel: why is this in here?

lib/data/external/AwsClient.js
rahulreddy: The caller was sending `null` before, so I used the same.

rahulreddy:
@LaurenSpiegel We will be following up with fixes for Azure in a different PR.

rahulreddy:
@ironman-machine try

lib/data/locationConstraintParser.js
rahulreddy: I will move it to another PR

rahulreddy:
@ironman-machine try

PR #1025, opened 2017-11-30
S3C-1115 FT: Adds GCP Base Files

alexanderchan-scality:
Adds the base files for building the GCP service using the AWS-SDK.

### DIFF
--- a/lib/data/external/GCP/GcpService.js
+++ b/lib/data/external/GCP/GcpService.js
@@ -1,0 +1,199 @@
+const AWS = require('aws-sdk');
+const { errors } = require('arsenal');
+const Service = AWS.Service;
+
+const GcpSigner = require('./GcpSigner');
+
+AWS.apiLoader.services.gcp = {};
+const GCP = Service.defineService('gcp', ['2017-11-01'], {
+    getSignerClass() {
+        return GcpSigner;
+    },
+
+    validateService() {
+        if (!this.config.region) {
+            this.config.region = 'us-east-1';
+        }
+    },
+
+    upload(params, options, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: upload not implemented'));
+    },
+
+    // Service API
+    listBuckets(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: listBuckets not implemented'));
+    },
+
+    // Bucket APIs
+    getBucketLocation(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: getBucketLocation not implemented'));
+    },
+
+    deleteBucket(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: deleteBucket not implemented'));
+    },
+
+    headBucket(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: headBucket not implemented'));
+    },
+
+    listObjects(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: listObjects not implemented'));
+    },
+
+    listObjectVersions(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: listObjecVersions not implemented'));
+    },
+
+    putBucket(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: putBucket not implemented'));
+    },
+
+    getBucketAcl(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: getBucketAcl not implemented'));
+    },
+
+    putBucketAcl(params, callback) {
+        return callback(errors.NotImplemented
+            .customizeDescription('GCP: putBucketAcl not implemented'));
+    },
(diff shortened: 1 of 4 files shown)


alexanderchan-scality:
@ironman-machine try

rahulreddy:
@alexanderchan-scality As per our discussion, let's make a GCP client which extends AWS Client class.

alexanderchan-scality:
@rahulreddy Will do. I must have mixed up the ideas. I will fix it.

alexanderchan-scality:
@ironman-machine try

alexanderchan-scality:
Yes. I intend implement the api methods first then the functional tests. Those still require the credentials to be setup, and I believe separating the PR as such will make it easier on review.

lib/data/external/gcp-2017-11-01.api.json
```diff
@@ -0,0 +1,174 @@
+{
```
dora-korpar: Could we put all these base files at are loose in the `external` folder into their own `GCP` folder? Like `external/GCP`. I think we should try to keep things as easy to navigate/manage as possible :)

alexanderchan-scality: I see. Thanks for the tip. I will move the files into their own folder like you mentioned.

lib/data/locationConstraintParser.js
```diff
@@ -64,14 +65,16 @@ function parseLC() {
                 s3Params.secretAccessKey =
                     locationObj.details.credentials.secretKey;
             }
-            clients[location] = new AwsClient({
+            const config = {
                 s3Params,
                 awsBucketName: locationObj.details.bucketName,
```
dora-korpar: I'm also generally a bit confused about how the config will look for a GCP location constraint. With this setup, the user will have to have an AWS account name and access key as well as GCP credentials, correct?

Also, with the two location constraint types sharing these config options, maybe we should rename awsBucketNmae to just bucketName?

dora-korpar: Discussed credentials, looks like only GCP creds are needed. Still think the `awsBucketName` key should be changed :)

alexanderchan-scality: Yup. For the config, we will just need the GCP interoperability credentials. The credentials are very similar to that of AWS's, an access id and a secret key, and with it, the location constraint for GCP will just be:
```
"gcpbackend": {
        "type": "gcp",
        "legacyAwsBehavior": true,
        "details": {
            "awsEndpoint": "storage.googleapis.com",
            "bucketName": "bucket name",
            "bucketMatch": true,
            "credentialsProfile": "profile name"
        }
    }
```  
Those credentials will be store as one of the profiles in AWS, so retrieving it will just be specifying the profile name.  
As for the variable names, yup... I had overlooked that

electrachong:
Agree that it might be better to separate adding the location constraint config and validation in another PR. On the other hand, it might be nice to have functional tests to reference in each of the subsequent PRs for implementing the APIs. Do you think it is too onerous or time-consuming at the moment to base a PR which adds the test location constraint on this one, and rebase all of the API prs on that PR?

rahulreddy:
Commenting for clarity, this should be the only PR without tests. All other subsequent PRs which implement the APIs should have functional tests that go with them.
As long as this PR does not expose new functionality to the end user and it doesn't break existing tests it should be ok. Location constraint changes should be introduced separately.

alexanderchan-scality:
I see. I will combine the API implementations and their corresponding test as a PR each. I had separated them because the location constraint hasn't been introduced and just for the sake getting feedback on them.

alexanderchan-scality:
@ironman-machine try

alexanderchan-scality:
@ironman-machine try

alexanderchan-scality:
@ironman-machine try

lib/data/external/GCP/GcpService.js
```diff
@@ -0,0 +1,199 @@
...
+Object.defineProperty(AWS.apiLoader.services.gcp, '2017-11-01', {
+    get: function get() {
+        const model = require('./gcp-2017-11-01.api.json');
+        return model;
+    },
+    enumerable: true,
+    configurable: true,
+});
+
+Object.assign(GCP.prototype, {
```
rahulreddy: Can you opt for class and extends instead of modifying the prototype directly?

alexanderchan-scality: The defineService returns an object, so I had to modify the prototype directly.

alexanderchan-scality:
@ironman-machine try

rahulreddy:
@ironman-machine try

alexanderchan-scality:
@ironman-machine try

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r+

alexanderchan-scality:
@ironman-machine 
SCALITY_INTEGRATION_BRANCH=test/10k-flakiness
try

rahulreddy:
Build nfsd seems to be be failing constantly across all the containers, unrelated to this PR.

rahulreddy:
@ironman-machine clean r+

PR #1034, opened 2017-12-02
FT: Add Config::setLocationConstraints

bennettbuchanan:
This PR is the first step in allowing the location constraints to be updated dynamically.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -812,6 +812,12 @@ class Config extends EventEmitter {
         this.emit('rest-endpoints-update');
     }
 
+    setLocationConstraints(locationConstraints) {
+        restEndpointsAssert(this.restEndpoints, locationConstraints);
+        this.locationConstraints = locationConstraints;
+        this.emit('location-constraints-update');
+    }
+
     getAzureEndpoint(locationConstraint) {
         let azureStorageEndpoint =
         process.env[`${locationConstraint}_AZURE_STORAGE_ENDPOINT`] ||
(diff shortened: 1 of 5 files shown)


tests/unit/multipleBackend/setLocationConstraints.js
```diff
@@ -0,0 +1,34 @@
...
+    it('should update location constraint config', () => {
+        assert.notDeepStrictEqual(config.locationConstraints, expectedLCs);
+        config.setLocationConstraints(expectedLCs);
+        assert.deepStrictEqual(config.locationConstraints, expectedLCs);
+    });
+
+    it('should update multiple backend clients', () => {
+        assert.strictEqual(parseLC(config)[newLCKey], undefined);
+        config.setLocationConstraints(expectedLCs);
+        assert(parseLC(config)[newLCKey]);
```
LaurenSpiegel: shouldn't you assert that it's actually an s3 client?

lib/Config.js
```diff
@@ -804,6 +804,12 @@ class Config extends EventEmitter {
         this.emit('rest-endpoints-update');
     }
 
+    setLocationConstraints(locationConstraints) {
```
LaurenSpiegel: so to add a location constraint, pensieve has to send the whole lc back to s3? where does pensieve store the full location constraint file?

bennettbuchanan: Correct. My understanding is that it will be stored in a bucket used by zenko, and the retrieval of resources from there will managed by the Pensieve API.

tests/unit/multipleBackend/setLocationConstraints.js
bennettbuchanan: Updated.

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine r+

bennettbuchanan:
E2E passed on retry here: http://ci.ironmann.io/gh/scality/Integration/17747#build-parameters.

PR #1038, opened 2017-12-05
FT: Add azure type for replication information

bennettbuchanan:
* Add azure type for replication information.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -126,7 +126,8 @@ function _checkMultipleBackendRequest(request, log) {
     const location = locationConstraints[headers['x-scal-storage-class']];
     const isValidLocation = location &&
         location.type === headers['x-scal-storage-type'] &&
-        location.details.bucketName === bucketName;
+        (location.details.bucketName === bucketName ||
+        location.details.azureContainerName === bucketName);
     if (!isValidLocation) {
         errMessage = 'invalid request: invalid location constraint in request';
         log.debug(errMessage, {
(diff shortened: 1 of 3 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -255,7 +256,11 @@ function putObject(request, response, log, callback) {
     const cacheControl = request.headers['x-scal-cache-control'];
     const contentDisposition = request.headers['x-scal-content-disposition'];
     const contentEncoding = request.headers['x-scal-content-encoding'];
-    const metaHeaders = {
+    const backendType = locationConstraints[storageLocation].type;
+    const metaHeaders = backendType === 'azure' ? {
```
jonathan-gramain: Why using a different convention for azure than for AWS? Using the same would simplify things at various levels.

FWIW, we also seem to use `x-amz-meta-` convention for azure backend in S3, as found by searching this prefix in lib/data/external/AzureClient.js, and there's a `_translateMetaHeaders()` helper to pass them on to Azure API.

bennettbuchanan: I was actually wondering that, but this `_translateMetaHeaders` helper in multiple backend Azure client does not trim the prefix `x-amz-meta-` like the multiple backend AWS client does. I am assuming that is intentional, but maybe we need to be doing so in the multiple backend Azure client.

bennettbuchanan: I've updated this PR to trim any `x-amz-meta-` prefix in Azure client here: https://github.com/scality/S3/pull/1038/files#diff-96a8cd65c80ea7fb9b908a81388494d4R66.

bennettbuchanan: As per our conversation, we will maintain the user metadata key without the `x_amz_meta-` prefix for the context of replication. For all other user metadata being put to azure, it will have this prefix. So we will keep this condition here in the backbeat route.

rahulreddy: Can you also add a `TODO:` comment capturing the reason, so that we can follow up with an update later?

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

PR #1053, opened 2017-12-13
[WIP] Fix versioning-related Utapi metrics [SC3-368]

electrachong:
To-do: 
 * objectPut, objectCopy, completeMpu: track correct old content length when overwriting null version that is not the latest version;
  * deleteObject: when creating a delete marker when versioning is suspended: 
     * - decrement storage utilized by the null version, if any.
     * - number of objects should remain unchanged (since we both delete a version, and put a delete marker).

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -23,6 +23,7 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
                 return callback(err);
             }
             if (dataToDelete) {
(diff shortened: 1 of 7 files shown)


electrachong:
end to end tried here: http://ci.ironmann.io/gh/scality/Integration/17865

electrachong:
@ironman-machine 
SCALITY_INTEGRATION_BRANCH=fix/utapi-versioning-metrics
try

electrachong:
@ironman-machine try

electrachong:
@ironman-machine try

PR #1054, opened 2017-12-13
FT: Update multi backend for replication to Azure

bennettbuchanan:


lib/routes/routeBackbeat.js
```diff
@@ -407,24 +430,46 @@ function completeMultipartUpload(request, response, log, callback) {
...
         }
+        const partList =
+            getPartList(parts, request.objectKey, uploadId, storageLocation);
+        // Azure client will set user metadata at this point.
+        const metaHeaders = {
+            'scal-replication-status': 'REPLICA',
+            'scal-version-id': sourceVersionId,
+        };
+        if (userMetadata) {
+            Object.assign(metaHeaders, JSON.parse(userMetadata));
```
jonathan-gramain: I think you should not trust `userMetadata` since it comes directly from the received headers, i.e. put the `JSON.parse` in a `try..catch` and return a request error on failure. You may add additional checks if need be but I don't know if there's any useful check.

lib/routes/routeBackbeat.js
```diff
@@ -89,7 +95,8 @@ function _checkMultipleBackendRequest(request, log) {
         log.error(errMessage);
         return errors.BadRequest.customizeDescription(errMessage);
     }
-    if ((operation === 'puttagging' || operation === 'deletetagging') &&
+    if (headers['x-scal-storage-type'] === 'aws' &&
```
jonathan-gramain: S3 has a constant `versioningNotImplBackends` (in constants.js), you may use the same here, as it's likely to forget updating such piece of code when adding a new backend compared to updating a global constant (you can copy the constant in backbeat constants to avoid doing an arsenal PR but ideally it should be uniquely defined in arsenal).

lib/routes/routeBackbeat.js
bennettbuchanan: Nice catch. Updated.

lib/routes/routeBackbeat.js
bennettbuchanan: Good idea. Updated.

lib/data/multipleBackendGateway.js
```diff
@@ -198,11 +198,14 @@ const multipleBackendGateway = {
     },
 
     completeMPU: (key, uploadId, location, jsonList, mdInfo, bucketName,
```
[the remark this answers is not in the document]
bennettbuchanan: As per our conversation, most of these methods for multiple backends could benefit from a similar refactor, so we can make the change across all methods in a separate refactor PR.

lib/routes/routeBackbeat.js
```diff
@@ -141,6 +146,26 @@ function _checkMultipleBackendRequest(request, log) {
     return undefined;
 }
 
+function getPartList(parts, objectKey, uploadId, storageLocation) {
+    const partList = {};
+    if (locationConstraints[storageLocation].type === 'azure') {
+        partList.uncommittedBlocks = [];
```
rahulreddy: You can use map to make this cleaner.
```js
partList.uncommittedBlocks = parts.map(part => {


```

lib/routes/routeBackbeat.js
```diff
@@ -339,7 +364,7 @@ function putObject(request, response, log, callback) {
                 return callback(errors.BadDigest);
             }
             const dataRetrievalInfo = {
-                versionId: retrievalInfo.dataStoreVersionId,
+                versionId: retrievalInfo.dataStoreVersionId || '',
```
rahulreddy: Can you add a todo here, mentioning to remove when Azure versioning is implemented.

lib/routes/routeBackbeat.js
bennettbuchanan: I want to use spread syntax here to push all items of the array `subPartIds`.

https://github.com/scality/S3/pull/1054/files/6547c168ac5060473d1da7904e86fece26c664e5#diff-5c3dc4783e2d0f782449eea237c8f691R161

lib/routes/routeBackbeat.js

rahulreddy:
@ironman-machine r+

bennettbuchanan:
Automatically retried successfully here: http://ci.ironmann.io/gh/scality/Integration/18383#build-parameters

rahulreddy:
@ironman-machine r-

PR #1060, opened 2017-12-27
bf: close/end readable/response streams on errors

rahulreddy:
This fixes the leakage of sockets in CLOSE_WAIT state by closing the streams and
destroying the sockets when the client has abruptly closed the connection.
Depends on https://github.com/scality/sproxydclient/pull/132

### DIFF
--- a/lib/utilities/retrieveData.js
+++ b/lib/utilities/retrieveData.js
@@ -1,31 +1,74 @@
-export default function retrieveData(locations, dataRetrievalFn,
-    response, logger, errorHandlerFn) {
-    if (locations.length === 0) {
-        return response.end();
-    }
-    if (errorHandlerFn === undefined) {
-        // eslint-disable-next-line
-        errorHandlerFn = () => { response.connection.destroy(); };
-    }
-    const current = locations.shift();
-    return dataRetrievalFn(current, logger,
-        (err, readable) => {
+const { eachSeries } = require('async');
+const responseErr = new Error();
+responseErr.code = 'ResponseError';
+responseErr.message = 'response closed by client request before all data sent';
+
+export default function retrieveData(locations, retrieveDataFn, response, log) {
+    let responseDestroyed = false;
+    const _destroyResponse = () => {
+        // destroys the socket if available
+        response.destroy();
+        responseDestroyed = true;
+    };
+    response.once('close', () => {
+        log.debug('received close event before response end');
+        _destroyResponse();
+    });
+
+    eachSeries(locations,
+        (current, next) => retrieveDataFn(current, log, (err, readable) => {
+            let cbCalled = false;
+            const _next = err => {
+                // Avoid multiple callbacks since it's possible that response's
+                // close event and the readable's end event are emitted at
+                //  the same time.
+                if (!cbCalled) {
+                    cbCalled = true;
+                    next(err);
+                }
+            };
+
             if (err) {
-                logger.error('failed to get object', {
+                log.error('failed to get object', {
                     error: err,
                     method: 'retrieveData',
                 });
-                return errorHandlerFn(err);
+                _destroyResponse();
+                return _next(err);
             }
-            readable.on('error', err => {
-                logger.error('error piping data from source');
-                errorHandlerFn(err);
+            if (responseDestroyed) {
+                log.debug('response destroyed before readable could stream');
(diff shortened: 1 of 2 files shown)


rahulreddy:
@ironman-machine 
SCALITY_SPROXYDCLIENT_BRANCH=bf/S3C-1026-memory-leak
try

lib/utilities/retrieveData.js
```diff
@@ -1,31 +1,73 @@
...
+const { eachSeries } = require('async');
+const responseErr = new Error();
+responseErr.code = 'ResponseError';
+responseErr.message = 'response errored out before readable finished';
+
+export default function retrieveData(locations, retrieveDataFn, response, log) {
+    let responseDestroyed = false;
+    const _destroyResponse = () => {
+        // destroys the socket if available
+        response.destroy();
```
philipyoo: Just checking, does this need to be `response.connection` or `response.socket`?

lib/utilities/retrieveData.js
```diff
@@ -1,31 +1,73 @@
...
-        // eslint-disable-next-line
-        errorHandlerFn = () => { response.connection.destroy(); };
-    }
-    const current = locations.shift();
-    return dataRetrievalFn(current, logger,
-        (err, readable) => {
+const { eachSeries } = require('async');
+const responseErr = new Error();
+responseErr.code = 'ResponseError';
+responseErr.message = 'response errored out before readable finished';
```
LaurenSpiegel: maybe 'response closed by client request before all data sent'

lib/utilities/retrieveData.js
```diff
@@ -1,31 +1,73 @@
...
+
+export default function retrieveData(locations, retrieveDataFn, response, log) {
+    let responseDestroyed = false;
+    const _destroyResponse = () => {
+        // destroys the socket if available
+        response.destroy();
+        responseDestroyed = true;
+    };
+    response.on('close', () => {
```
LaurenSpiegel: why 'on' here but 'once' in sproxydclient and below?

lib/utilities/retrieveData.js
```diff
@@ -1,31 +1,73 @@
...
+    eachSeries(locations,
+        (current, next) => retrieveDataFn(current, log, (err, readable) => {
+            let cbCalled = false;
+            const _next = err => {
+                // Avoid multiple callbacks since it's possible that response's
+                // close event and the readable's end event are emitted at
+                //  the same time.
+                if (!cbCalled) {
+                    next(err);
+                    cbCalled = true;
```
LaurenSpiegel: just to be safe, set this to true before calling next(err)?

lib/utilities/retrieveData.js
```diff
@@ -1,31 +1,73 @@
...
                 });
-                return errorHandlerFn(err);
+                return _next(err);
             }
-            readable.on('error', err => {
-                logger.error('error piping data from source');
-                errorHandlerFn(err);
+            if (responseDestroyed) {
+                log.debug('response destroyed before readable could stream');
+                readable.emit('close');
```
LaurenSpiegel: does this do anything?

lib/utilities/retrieveData.js
rahulreddy: Just being careful here to avoid EventEmitter leak since it's in a loop.

lib/utilities/retrieveData.js
rahulreddy: Sure.

lib/utilities/retrieveData.js
rahulreddy: This close event has a listener in sproxydclient here https://github.com/scality/sproxydclient/pull/132 which aborts the request to sproxyd and destroys the socket.

lib/utilities/retrieveData.js
rahulreddy: Both `response.connection` and `response.socket` are set to null by node.js once the socket is destroyed. This method offers a degree of safety where it checks if `response.socket` is defined before calling `response.socket.destroy()`

rahulreddy:
@ironman-machine clean after=https://github.com/scality/sproxydclient/pull/132

rahulreddy:
@ironman-machine clean r+

PR #1064, opened 2017-12-29
FT: Implement MongoDB backend

### LINKED ISSUE #824: MongoDB instead of levelDB for metedata storage
Hello

Is there any plans in you schedule to use Mongo DB as a metadata storage instead of levelDB? If not is there any guide on how to replace LevelDB with Mongo DB correctly?

vrancurel:
MongoDB driver for Metadata

FIX #824


lib/metadata/mongoclient/backend.js
```diff
@@ -0,0 +1,554 @@
...
+                    throw (errors.InternalError);
+                }
+            });
+    }
+    
+    getCollection(name) {
+	/* mongo has a problem with .. in collection names */
+	if (name === constants.usersBucket)
+	    name = "users__bucket";
+	return this.db.collection(name);
```
LaurenSpiegel: do we want to use the strict version with a callback so that we can provide an error if the collection does not exist or handle appropriately? Skip comment. As discussed, we are ok having the collection created on a put object.

lib/metadata/mongoclient/readStream.js
```diff
@@ -0,0 +1,90 @@
...
+	    var value = undefined;
+	    
+	    if (doc) {
+		var key = doc._id;
+		var value = JSON.stringify(doc.value);
+	    }
+	    
+	    if (key === undefined && value === undefined) {
+		this.push(null);
+	    } else if (this._options.keys !== false &&
```
LaurenSpiegel: what if the key is undefined?

lib/api/apiUtils/object/removeAWSChunked.js
```diff
@@ -6,7 +6,8 @@
  * @return {string} new value w. 'aws-chunked'/'aws-chunked,' substring removed
  */
 function removeAWSChunked(sourceHeader) {
-    if (sourceHeader === undefined) {
+    if (sourceHeader === undefined ||
+        sourceHeader === null) {
```
LaurenSpiegel: why did you have to change this? seems unrelated.

LaurenSpiegel:
@ironman-machine try

vrancurel:
@ironman-machine try

tests.bash
```diff
@@ -126,6 +126,14 @@ then
 
   killandsleep 8000
 
+  docker pull mongo:3.4.10
+  mkdir /tmp/mongodb
+  docker run -d -p 27018:27017 -v /tmp/mongodb:/data/db mongo:3.4.10
+  bash wait_for_local_port.bash 27018 40
+  S3BACKEND=mem MPU_TESTING=yes S3METADATA=mongodb npm start > $CIRCLE_ARTIFACTS/server_mongodb_awssdk.txt & bash wait_for_local_port.bash 8000 40 && npm run ft_awssdk
```
ttvi: It's just awssdk here, maybe you need to check this backend with the other test suits (s3cmd, s3curl, jaws, node) as well.

vrancurel: Changing ft_awssdk by ft_test

vrancurel:
Updated

vrancurel:
@ironman-machine try

PR #1070, opened 2018-01-06
FT: Update replication info for one-to-many

bennettbuchanan:
Depends on https://github.com/scality/backbeat/pull/162.
Depends on https://github.com/scality/Arsenal/pull/394.

### DIFF
--- a/lib/api/apiUtils/object/getReplicationInfo.js
+++ b/lib/api/apiUtils/object/getReplicationInfo.js
@@ -1,32 +1,47 @@
 const s3config = require('../../../Config').config;
 
-function _getReplicationInfo(rule, replicationConfig, content, operationType) {
-    const storageClasses = rule.storageClass ?
-        rule.storageClass.split(',') : [];
+function _getBackend(objectMD, site) {
+    const backends = objectMD ? objectMD.replicationInfo.backends : [];
+    const backend = backends.find(o => o.site === site);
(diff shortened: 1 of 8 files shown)


config.json
```diff
@@ -66,7 +67,7 @@
         "port": 9991
     },
     "recordLog": {
-        "enabled": false,
+        "enabled": true,
```
bennettbuchanan: Remove this.

lib/api/apiUtils/object/getReplicationInfo.js
```diff
@@ -1,33 +1,48 @@
 const s3config = require('../../../Config').config;
 
-function _getReplicationInfo(rule, replicationConfig, content, operationType) {
-    const storageClasses = rule.storageClass ?
-        rule.storageClass.split(',') : [];
+function _getBackend(objectMD, site) {
```
rahulreddy: I forgot the context for this one. Why would the backend not exist? And also, it's weird that a get call is updating the status.

bennettbuchanan: If it's the first time putting an object to be replicated it will push the new backend to the array. In the case where you are updating an object version (for example putting tags), we want to preserve the datastoreVersionId. 

I guess the comment is a bit misleading in the sense that it's not really "updating" the status, but rather preserving any existing object values in the `backends` array.

PR #1076, opened 2018-01-13
bf: use stream write method instead of piping

rahulreddy:


lib/utilities/retrieveData.js
```diff
@@ -60,7 +60,17 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
                 log.error('error piping data from source');
                 return _next(err);
             });
-            return readable.pipe(response, { end: false });
+            readable.on('data', chunk => {
+                const write = response.write(chunk);
+                if (!write) {
+                    readable.pause();
+                    response.once('drain', () => {
+                        response.write(chunk);
```
LaurenSpiegel: won't this write the same chunk again?  The docs say -- "false if the stream wishes for the calling code to wait for the 'drain' event to be emitted before continuing to write **additional** data"

jonathan-gramain: Instead of `response.write` you may want to call `readable.resume()` here

lib/utilities/retrieveData.js
```diff
@@ -60,7 +60,17 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
                 log.error('error piping data from source');
                 return _next(err);
             });
-            return readable.pipe(response, { end: false });
+            readable.on('data', chunk => {
```
jonathan-gramain: Can you give a short hint on why this solves the memory issue in the PR?

lib/utilities/retrieveData.js
jonathan-gramain: My bad, you call it just below, so removing `response.write` should be fine.

rahulreddy: I was under the assumption that I had to write the same chunk again...reading the docs looks like it does accept the chunk and we have to pause for additional data. Thanks for catching that!

lib/utilities/retrieveData.js
rahulreddy: It's response.write method which solves the issue. I don't have a technical argument as to why, because stream.pipe essentially does the same.

jonathan-gramain: ok, good to know

rahulreddy:
Ok, this isn't making much difference. Funny thing is if I comment out this stub of code, it works.
```js
if (!write) {
    readable.pause();
    response.once('drain', () => {
        readable.resume();
    });
}
```

LaurenSpiegel:
can you do anything with the paused readable once the client connection has dropped?

rahulreddy:
I am not sure since the problem lies in how the socket memory is being handled. I think the best approach forward is to do some network traces and observe the differences between the two.

PR #1079, opened 2018-01-16
FT: Protect reserved bucketname

LaurenSpiegel:
This used to be protected by virtue of having capitalized letters but now that it is permitted in arsenal, protect here.


rahulreddy:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine try

PR #1085, opened 2018-01-18
TEST: Add content length check on get

LaurenSpiegel:
Node sdk right now apparently does not check that body size is what was expected. I spent some time trying to figure out why we had an etag mismatch on a test with the aws backend --
 http://ci.ironmann.io/gh/scality/S3/9975

After looking through the logs, it's clear the connection with aws dropped before all of the data was retrieved but this was not apparent from the test.  This should help in the future.


LaurenSpiegel:
@ironman-machine try

LaurenSpiegel:
@ironman-machine r+

LaurenSpiegel:
@ironman-machine try

bennettbuchanan:
@ironman-machine r+

bennettbuchanan:
The "r+" command was successful but merging failed due to pending ci/circleci_enterprise tests. Those tests have now completed successfully. Merging manually.

PR #1086, opened 2018-01-24
Fix: Revert default log level to info

LaurenSpiegel:


bennettbuchanan:
@ironman-machine r+

bennettbuchanan:
@ironman-machine r+

PR #1087, opened 2018-01-25
S3C-1026 S3 consumes too much TCP memory

flavien-scality:
destroy() the readable end (sproxyd) when the client closes the connection,
instead of sending a 'close' event.
Cleanup other error cases on the way.

# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/utilities/retrieveData.js
+++ b/lib/utilities/retrieveData.js
@@ -33,7 +33,6 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
                     error: err,
                     method: 'retrieveData',
                 });
-                _destroyResponse();
                 return _next(err);
             }
             if (responseDestroyed) {
@@ -47,17 +46,19 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
                 if (!responseDestroyed) {
                     _destroyResponse();
                 }
-                readable.emit('close');
+                readable.destroy();
                 return _next(responseErr);
             });
             // readable stream successfully consumed
-            readable.on('end', () => {
+            readable.once('end', () => {
+                readable.unpipe(response);
                 log.debug('readable stream end reached');
                 return _next();
             });
             // errors on server side with readable stream
-            readable.on('error', err => {
+            readable.once('error', err => {
                 log.error('error piping data from source');
+                _destroyResponse();
                 return _next(err);
             });
             return readable.pipe(response, { end: false });
@@ -65,6 +66,7 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
             if (err) {
                 log.debug('abort response due to client error', {
                     error: err.code, errMsg: err.message });
+		return _destroyResponse();
             }
             // call end for all cases (error/success) per node.js docs
             // recommendation


lib/utilities/retrieveData.js
```diff
@@ -47,24 +46,27 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
...
+            readable.once('end', () => {
+                readable.unpipe(response);
                 log.debug('readable stream end reached');
                 return _next();
             });
             // errors on server side with readable stream
-            readable.on('error', err => {
+            readable.once('error', err => {
                 log.error('error piping data from source');
+                _destroyResponse();
```
LaurenSpiegel: why destroyResponse here and on line 69?  Won't it hit both?

lib/utilities/retrieveData.js
```diff
@@ -47,24 +46,27 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
                 if (!responseDestroyed) {
                     _destroyResponse();
                 }
-                readable.emit('close');
+                readable.destroy();
```
LaurenSpiegel: so this is the magic line freeing up the memory? It basically pushes null into the readable stream.  https://github.com/nodejs/node/blob/master/lib/_stream_readable.js#L189

lib/utilities/retrieveData.js
```diff
@@ -47,24 +46,27 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
...
                 return _next(responseErr);
             });
             // readable stream successfully consumed
-            readable.on('end', () => {
+            readable.once('end', () => {
+                readable.unpipe(response);
```
LaurenSpiegel: is this unpipe necessary?

lib/utilities/retrieveData.js
flavien-scality: Yes, it is : by destroying the readable, it causes the underlying socket to be closed, so the kernel frees the bytes that have been buffered in the receive queue of the node process, and signals the other end  that it will not read from it again, causing the send buffer associated to nginx to be freed as well, and nginx getting a notification that the socket has been closed.
As for the implementation of readable.destroy, I honestly don't know. We have to be careful though : you're pointing to the branch master of node, but we're using a v4.8.4 (which is so old that it's even not available in the tags for the node project ; oldest is 4.8.6).
It just seems more reasonable to call destroy() than to emit a close event... and it closes the socket, which is what we were missing so far.

lib/utilities/retrieveData.js
flavien-scality: @rahulreddy had added it before I got involved and we did not have a chance to talk about it since I only did the diff after his departure. It looks reasonable to me : we read the whole answer, so unpiping the response looks fair. Are you thinking about any side effect ?

lib/utilities/retrieveData.js
flavien-scality: Same here : it was added before I got involved.
I agree with you, the logic of "_destroyResponse()" could be improved (ie test the boolean inside the function so that even if it's called twice it does not hurt ; even though I'm pretty sure calling it twice is fine anyway).

lib/utilities/retrieveData.js
rahulreddy: I remember trying this trick where I push `null` into the readable stream, it didn't seem to have much effect.
I am emitting `close` here so that I can have an listener in sproxydclient to destroy to the client request in-progress. https://github.com/scality/sproxydclient/blob/rel/6.4/lib/sproxyd.js#L231
This is critical and should not be removed, otherwise the listener is never called. That listener (per docs) should drop the unconsumed data in the streams and then destroy the socket. The event name emitted can be anything, I chose `close` as I thought that it seemed appropriate. Also, having the listener is sproxydclient makes sense so that any other projects using the lib doesn't have to re-implement the fix, all they have to do is emit `close`.

lib/utilities/retrieveData.js
rahulreddy: This was rando thing we tried. If you look at the implementation, unpipe just removes the writable from the list and emits 'unpipe'.

lib/utilities/retrieveData.js
rahulreddy: This can be removed.

lib/utilities/retrieveData.js
```diff
@@ -47,24 +46,27 @@ export default function retrieveData(locations, retrieveDataFn, response, log) {
...
                 log.error('error piping data from source');
+                _destroyResponse();
                 return _next(err);
             });
             return readable.pipe(response, { end: false });
         }), err => {
             if (err) {
                 log.debug('abort response due to client error', {
                     error: err.code, errMsg: err.message });
+		return _destroyResponse();
```
rahulreddy: I originally did not destroyResponse for all requests because for the situation where readable emits an error we have to send a 500 to the client instead of destroying the socket altogether.

lib/utilities/retrieveData.js
flavien-scality: So may be the fix is to let it this way here emitting a `close` and change sproxydclient so that it does a `destroy` of the socket (it currently calls `abort`) ? Sounds better ?

lib/utilities/retrieveData.js
flavien-scality: ok, I'll remove it then

lib/utilities/retrieveData.js
flavien-scality: Oh. You're right. Got it. I'll remove that change.

lib/utilities/retrieveData.js
rahulreddy: `abort` actually destroys the socket as well. https://github.com/nodejs/node/blob/master/lib/_http_client.js#L265

lib/utilities/retrieveData.js
flavien-scality: replying to self : If one of the reads frails from sproxyd, we cannot send back a 500 to the client : we already sent the headers with a 200, and content-length of the whole thing, so the only way to abort is for s3 to close the connection to the client when this happens.

ThibaultRiviere:
@flavien-scality @rahulreddy what's the status here?

harperscality:
@flavien-scality closing this as it seems finished elsewhere (#1100)
If you don't need this, don't hesitate to delete your branch

PR #1088, opened 2018-01-26
Fix: Prevent memory issue

LaurenSpiegel:
Fixes S3C-1026
Uses a passthrough stream with a 0 byte high watermark to
prevent excess buffering in streams.


LaurenSpiegel:
@ironman-machine try

LaurenSpiegel:
@ironman-machine try

LaurenSpiegel:
@ironman-machine try

PR #1095, opened 2018-01-30
Ft/s3 c 1115/gcp apis object

alexanderchan-scality:
Reference docs:
PUT: https://cloud.google.com/storage/docs/xml-api/put-object-upload
GET: https://cloud.google.com/storage/docs/xml-api/get-object-download
HEAD: https://cloud.google.com/storage/docs/xml-api/head-object
DELETE: https://cloud.google.com/storage/docs/xml-api/delete-object

### DIFF
--- a/lib/data/external/GCP/GcpService.js
+++ b/lib/data/external/GCP/GcpService.js
@@ -1,11 +1,19 @@
 const googleAuth = require('google-auto-auth');
 const async = require('async');
 const AWS = require('aws-sdk');
+const stream = require('stream');
 const { errors } = require('arsenal');
 const Service = AWS.Service;
 
 const GcpSigner = require('./GcpSigner');
 
+/**
+ * genAuth - create a google authorizer for generating request tokens
+ * @param {object} authParams - params that contains the credentials for
+ * generating the authorizer
+ * @param {function} callback - callback function to call with the authorizer
+ * @return {undefined}
+ */
 function genAuth(authParams, callback) {
     async.tryEach([
         function authKeyFile(next) {
@@ -32,6 +40,12 @@ const GCP = Service.defineService('gcp', ['2017-11-01'], {
     _jsonAuth: null,
     _authParams: null,
 
+    /**
+     * getToken - generate a token for authorizing JSON API requests
+     * @param {function} callback - callback function to call with the
+     * generated token
+     * @return {undefined}
+     */
     getToken(callback) {
         if (this._jsonAuth) {
             return this._jsonAuth.getToken(callback);
@@ -69,12 +83,17 @@ const GCP = Service.defineService('gcp', ['2017-11-01'], {
         return this.listObjects(params, callback);
     },
 
-    // TO-DO: Implemented the following APIs
-    upload(params, options, callback) {
-        return callback(errors.NotImplemented
-            .customizeDescription('GCP: upload not implemented'));
+    // Object APIs
+    upload(params, callback) {
+        if (params.Body instanceof stream) {
+            return callback(errors.NotImplemented
+                .customizeDescription(
+                    'GCP: Upload with stream body not implemented'));
+        }
+        return this.putObject(params, callback);
     },
 
+    // TO-DO: Implement the following APIs
     // Service API
     listBuckets(params, callback) {
         return callback(errors.NotImplemented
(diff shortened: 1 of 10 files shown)


lib/data/external/GCP/GcpService.js
```diff
@@ -69,12 +83,17 @@ const GCP = Service.defineService('gcp', ['2017-11-01'], {
...
+    upload(params, callback) {
+        if (params.Body instanceof stream) {
+            return callback(errors.NotImplemented
+                .customizeDescription(
+                    'GCP: Upload with stream body not implemented'));
+        }
+        return this.putObject(params, callback);
     },
 
+    // TO-DO: Implemented the following APIs
```
bennettbuchanan: ~Implemented~ Implement

tests/functional/raw-node/test/GCP/object/delete.js
```diff
@@ -0,0 +1,116 @@
...
+const { getRealAwsConfig } =
+    require('../../../../aws-node-sdk/test/support/awsConfig');
+
+const credentialOne = 'gcpbackend';
+
+describe('GCP: DELETE Object', function testSuite() {
+    this.timeout(30000);
+    const config = getRealAwsConfig(credentialOne);
+    const gcpClient = new GCP(config);
+
```
bennettbuchanan: Seems like you could have a `beforeEach` hook here that creates the bucket rather than do it twice in the nested `describe` statements.

tests/functional/raw-node/test/GCP/object/get.js
```diff
@@ -0,0 +1,142 @@
...
+                    process.stdout.write('Created bucket\n');
+                }
+                makeGcpRequest({
+                    method: 'PUT',
+                    bucket: this.currentTest.bucketName,
+                    objectKey: this.currentTest.key,
+                    authCredentials: config.credentials,
+                }, (err, res) => {
+                    if (err) {
+                        process.stdout.write(`err in creating object ${err}\n`);
```
bennettbuchanan: Here I think it's best to return the callback with the error, otherwise we end up with the custom error when `res` is undefined.

tests/functional/raw-node/test/GCP/object/get.js
```diff
@@ -0,0 +1,142 @@
...
+        beforeEach(function beforeFn(done) {
+            this.currentTest.bucketName = `somebucket-${Date.now()}`;
+            this.currentTest.key = `somekey-${Date.now()}`;
+            gcpRequestRetry({
+                method: 'PUT',
+                bucket: this.currentTest.bucketName,
+                authCredentials: config.credentials,
+            }, 0, err => {
+                if (err) {
+                    process.stdout.write(`err in creating bucket ${err}\n`);
```
bennettbuchanan: We can just return the callback with the error here.

tests/functional/raw-node/test/GCP/object/get.js
```diff
@@ -0,0 +1,142 @@
...
+                } else {
+                    process.stdout.write('Created bucket\n');
```
bennettbuchanan: In this repo I see tests both with explicit steps written to stdout and tests without them. I am in favor of removing stdout messages like this since the information can be assumed when no error occurred. What do you think?

alexanderchan-scality: I agree. I had them for debugging; will clean up the stdout messages

lib/data/external/GCP/GcpService.js
```diff
@@ -1,11 +1,19 @@
...
 const { errors } = require('arsenal');
 const Service = AWS.Service;
 
 const GcpSigner = require('./GcpSigner');
 
+/**
+ * genAuth - create a google authorizer for generating request tokens
+ * @param {object} authParams - params that contains the credentials for
+ * generating the authorizer
+ * @param {function} callback - callback function to call with the an authorizer
```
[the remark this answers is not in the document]
alexanderchan-scality: Thanks! will fix.

lib/data/external/GCP/GcpService.js
alexanderchan-scality: Thanks! will fix.

tests/functional/raw-node/test/GCP/object/delete.js
alexanderchan-scality: Thanks! will restructure the hooks.

tests/functional/raw-node/test/GCP/object/put.js
```diff
@@ -0,0 +1,116 @@
...
+                Key: this.test.key,
+            }, (err, res) => {
+                assert.notStrictEqual(res.VersionId, this.test.uploadId);
+                return done();
+            });
+        });
+    });
+
+    describe('without existing object in bucket', () => {
+        beforeEach(function beforeFn(done) {
```
bennettbuchanan: Looks like no need for `beforeEach` hook here.

alexanderchan-scality: I have that beforeEach to set a variable used in an afterEach

bennettbuchanan: It's minor, but can you define `this.test.key` in the `it` block instead, before putting the object? Then we can remove the hook.

alexanderchan-scality: yes. that is possible.

alexanderchan-scality:
@ironman-machine try

bennettbuchanan:
@ironman-machine r+

bennettbuchanan:
Retried successfully here: http://ci.ironmann.io/gh/scality/Integration/19047

PR #1099, opened 2018-01-30
Ft/s3 c 1115/gcp copy

alexanderchan-scality:
Reference Docs:
Copy: https://cloud.google.com/storage/docs/json_api/v1/objects/copy
Update: https://cloud.google.com/storage/docs/json_api/v1/objects/update


lib/data/external/GCP/GcpApis/updateMetadata.js
```diff
@@ -0,0 +1,76 @@
...
+                        JSON.parse(body) : body;
+                } catch (error) { res = undefined; }
+                if (res && res.error && res.error.code >= 300) {
+                    return next(
+                        new JsonError(res.error.message, res.error.code));
+                }
+                return next(null, res);
+            });
+        },
+    ], (err, result) => callback(err, result));
```
bennettbuchanan: Can just pass `callback` here.

alexanderchan-scality:
@ironman-machine try

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,122 @@
...
+ * @param {string} params.Key - object key
+ * @param {string} param.CopySource - source object
+ * @param {function} callback - callback function to call with the copy object
+ * result
+ * @return {undefined}
+ */
+function _copyObject(params, callback) {
+    const { CopySource } = params;
+    const sourceArray = CopySource.split('/');
+    const sourceBucket = sourceArray[0] ? sourceArray[0] : sourceArray[1];
```
bennettbuchanan: Is `sourceArray[0]` ever undefined?

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,122 @@
...
+                });
+            }
+            const resObj = {
+                CopyObjectResult: {
+                    ETag: result.etag,
+                    LastModified: result.updated,
+                },
+                VersionId: result.generation,
+            };
+            return next('SkipMetadata', resObj);
```
bennettbuchanan: In this method, it might be clearer to just return `next(null, result)` here and `next(null, res)` above. Then in the end of the waterfall construct the object once. That way we don't have to construct the object twice or send 'SkipMetadata' as an error.

alexanderchan-scality:
@ironman-machine try

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,112 @@
...
+ * for GCP Backend
+ * @param {object} params - update metadata params
+ * @param {string} params.Bucket - bucket name
+ * @param {string} params.Key - object key
+ * @param {string} param.CopySource - source object
+ * @param {function} callback - callback function to call with the copy object
+ * result
+ * @return {undefined}
+ */
+function _copyObject(params, callback) {
```
bennettbuchanan: Is this meant to be `copyObject` instead of `_copyObject` since it is exported?

lib/data/external/GCP/GcpUtils.js
```diff
@@ -1 +1,30 @@
...
+    constructor(type, code, desc) {
+        super(type);
+        this.code = code;
+        this.description = desc;
+        this[type] = true;
+    }
+}
+
+function checkSourceString(sourceString) {
+    return Boolean(sourceString.match(/^\/{0,1}[^/]+(\/[^/]+)+$/g));
```
bennettbuchanan: I would have to check, but it seems like a source object with multiple slashes (ex. `a/b/c`) should not be valid. Do you know? 

In any case, instead of the regexp I think perhaps it would be simpler to validate using `.split('/')`, especially since you call that method in a subsequent step.

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,112 @@
...
+    const { CopySource } = params;
+    if (!checkSourceString(CopySource)) {
```
bennettbuchanan: If `CopySource` is undefined, this function throws a TypeError. Looks like we should check that `CopySource` is defined before checking its validity.

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,112 @@
...
+        if (CopySource) {
+            return callback(errors.InvalidArgument);
+        }
+        return callback(errors.MissingParameter);
+    }
+    const sourceArray = CopySource.split('/');
+    const sourceBucket = sourceArray[0] ? sourceArray[0] : sourceArray[1];
```
bennettbuchanan: Can `sourceArray[0]` ever be falsy'?

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,112 @@
...
+            // if metadata directive is REPLACE then perform a metadata update
+            // otherwise default to COPY
+            if (params.MetadataDirective === 'REPLACE') {
+                const updateParams = {
+                    Bucket: params.Bucket,
+                    Key: params.Key,
+                    Metadata: params.Metadata || {},
+                    VersionId: result.generation,
+                };
+                return this.updateMetadata(updateParams, (err, res) => {
```
bennettbuchanan: Can pass `next` directly here to `this.updateMetadata` to save some lines. 😸

lib/data/external/GCP/GcpApis/updateMetadata.js
```diff
@@ -0,0 +1,70 @@
...
+ * objectTagging, in which creation/deletion of medatadata is required for GCP,
+ * and copyObject.
+ * @param {object} params - update metadata params
+ * @param {string} params.Bucket - bucket name
+ * @param {string} params.Key - object key
+ * @param {string} params.VersionId - object version id
+ * @param {function} callback - callback function to call with the object result
+ * @return {undefined}
+ */
+function _updateMetadata(params, callback) {
```
bennettbuchanan: `_updateMetadata` => `updateMetadata`.

lib/data/external/GCP/GcpApis/copyObject.js
bennettbuchanan: Can `sourceArray[0]` every be falsy? I guess it could happen with source with a leading slash, but seems like that should be invalid anyway.

tests/functional/raw-node/test/GCP/object/copy.js
```diff
@@ -0,0 +1,180 @@
...
+                    objectKey: this.currentTest.copyKey,
+                    authCredentials: config.credentials,
+                }, err => {
+                    if (err) {
+                        process.stdout
+                            .write(`err in deleting copy object ${err}\n`);
+                    }
+                    return next(err);
+                }),
+            ], err => done(err));
```
bennettbuchanan: Can pass `done` directly here and below.

tests/functional/raw-node/test/GCP/object/copy.js
```diff
@@ -0,0 +1,180 @@
...
+    });
+
+    describe('without existing object in bucket', () => {
+        it('should return 404 and \'Not Found\'', done => {
+            const missingObject = `nonexistingkey-${Date.now()}`;
+            const someKey = `somekey-${Date.now()}`;
+            gcpClient.copyObject({
+                Bucket: bucketName,
+                Key: someKey,
+                CopySource: `/${bucketName}/${missingObject}`,
```
bennettbuchanan: I guess a lot of my questions hinge on what the form of the `CopySource` param should take. Seems like it should not need a leading slash. Can you confirm?

lib/data/external/GCP/GcpUtils.js
alexanderchan-scality: Multiple slashes should be a valid source string. Example, in `a/b/c`, `a` is the bucket and `b/c` is the object key.

lib/data/external/GCP/GcpApis/copyObject.js
alexanderchan-scality: Overlooked this one, will add the check.

lib/data/external/GCP/GcpApis/copyObject.js
alexanderchan-scality: Yes. Ex. `/a/b`, the split will generate array [ '', 'a', 'b' ]

alexanderchan-scality: That is what I had thought. I did a few tests against the S3 API, and source string with a starting slash are still valid.

lib/data/external/GCP/GcpUtils.js
bennettbuchanan: True. The matching should also factor in source object with trailing slashes then (ex. `a/b/c/`).

lib/data/external/GCP/GcpApis/copyObject.js
bennettbuchanan: Ah, of course! 😄

bennettbuchanan: Ok, did not realize leading slash is handled.

lib/data/external/GCP/GcpUtils.js
```diff
@@ -1 +1,31 @@
...
+        super(type);
+        this.code = code;
+        this.description = desc;
+        this[type] = true;
+    }
+}
+
+function checkSourceString(sourceString) {
+    if (!sourceString) return false;
+    return Boolean(sourceString.match(/^\/{0,1}[^/]+(\/[^/]+)+$/g));
```
jonathan-gramain: I would recommend an explicit boolean return value like `return sourceString.match(...) !== null`, more directly understandable IMO.

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,107 @@
...
+ * @param {string} params.Bucket - bucket name
+ * @param {string} params.Key - object key
+ * @param {string} param.CopySource - source object
+ * @param {function} callback - callback function to call with the copy object
+ * result
+ * @return {undefined}
+ */
+function copyObject(params, callback) {
+    const { CopySource } = params;
+    if (!checkSourceString(CopySource)) {
```
jonathan-gramain: You may check first if `CopySource` is `undefined`, then make `checkSourceString` also check if the type is a `string` in order to call `.match` on it. This way you would also catch errors if `CopySource` type is neither `string` nor `undefined`.

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,107 @@
...
+        if (typeof CopySource === 'string') {
+            return callback(errors.InvalidArgument);
+        }
+        return callback(errors.MissingParameter);
+    }
+    const sourceArray = CopySource.split('/');
+    const sourceBucket = sourceArray[0] ? sourceArray[0] : sourceArray[1];
```
jonathan-gramain: If you just get rid of the first element of `sourceArray` if empty, you can probably simplify the following code by not having to deal with `objectStart`.

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,107 @@
...
+    const objectStart = sourceArray[0] ? 1 : 2;
+    const sourceObject =
+        sourceArray.slice(objectStart).reduce(
+            (prev, curr) => {
```
jonathan-gramain: You can use a functional expression instead of a true function for this: `(prev, curr) => prev ? ...`

Also, \``${curr}`\` <=> `curr`

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,107 @@
...
+                const ret = prev ? `${prev}/${curr}` : `${curr}`;
+                return ret;
+            }, '');
+    return async.waterfall([
+        next => {
```
jonathan-gramain: You can simply write `next => this.getToken(next)`

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,107 @@
...
+            request({
+                method: 'POST',
+                baseUrl: this.config.jsonEndpoint,
+                proxy: this.config.proxy,
+                uri,
+                auth: { bearer: token } },
+            (err, resp, body) => {
+                if (err) {
+                    logHelper(logger, 'error',
+                        'copyObject: err in json method',
```
jonathan-gramain: `err` => `error` (applies to all similar error messages)

package.json
```diff
@@ -22,10 +22,11 @@
     "aws-sdk": "2.28.0",
     "arsenal": "scality/Arsenal",
     "async": "~2.5.0",
-    "google-auto-auth": "^0.9.1",
     "azure-storage": "^2.1.0",
     "bucketclient": "scality/bucketclient",
     "commander": "^2.9.0",
+    "google-auto-auth": "^0.9.1",
+    "mongodb": "^2.2.31",
```
jonathan-gramain: why are we adding a dependency to mongodb?

tests/functional/raw-node/test/GCP/object/copy.js
```diff
@@ -0,0 +1,180 @@
...
+const { GCP } = require('../../../../../../lib/data/external/GCP');
+const { makeGcpRequest } = require('../../../utils/makeRequest');
+const { gcpRequestRetry } = require('../../../utils/gcpUtils');
+const { getRealAwsConfig } =
+    require('../../../../aws-node-sdk/test/support/awsConfig');
+
+const credentialOne = 'gcpbackend';
+const bucketName = `somebucket-${Date.now()}`;
+
+describe.only('GCP: COPY Object', function testSuite() {
```
jonathan-gramain: remove `.only`

lib/data/external/GCP/GcpUtils.js
```diff
@@ -1 +1,58 @@
...
+    }
+}
+
+function jsonRespCheck(err, resp, body, method, callback) {
+    if (err) {
+        logHelper(logger, 'error',
+            `${method}: err in json method`,
+            errors.InternalError.customizeDescription('json method failed'));
+        return callback(errors.InternalError
+            .customizeDescription('err in JSON Request'));
```
jonathan-gramain: s/`err`/`error`/ (also at line 26 above)

lib/data/external/GCP/GcpApis/updateMetadata.js
```diff
@@ -0,0 +1,44 @@
+const async = require('async');
+const request = require('request');
+
+const { jsonRespCheck } = require('../GcpUtils');
+
+/**
+ * updateMetadata - update the metadata of an object. Only used in when
```
bennettbuchanan: used ~in~ when

lib/data/external/GCP/GcpUtils.js
```diff
@@ -1 +1,62 @@
...
+        this.description = desc;
+        this[type] = true;
+    }
+}
+
+function jsonRespCheck(err, resp, body, method, callback) {
+    if (err) {
+        logHelper(logger, 'error',
+            `${method}: error in json method`,
+            errors.InternalError.customizeDescription('json method failed'));
```
bennettbuchanan: Should we be passing the `dataStoreName` to this log helper function? https://github.com/scality/S3/blob/07245ac181a7e89deb1571ded058b33f5a024bf3/lib/data/external/utils.js#L25

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,83 @@
...
+ * @param {string} params.Bucket - bucket name
+ * @param {string} params.Key - object key
+ * @param {string} param.CopySource - source object
+ * @param {function} callback - callback function to call with the copy object
+ * result
+ * @return {undefined}
+ */
+function copyObject(params, callback) {
+    const { CopySource } = params;
+    if (!CopySource) {
```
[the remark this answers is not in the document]
alexanderchan-scality: Will change the returned error for empty string case

lib/data/external/GCP/GcpApis/copyObject.js
```diff
@@ -0,0 +1,83 @@
...
+        return callback(errors.MissingParameter);
+    }
+    if (typeof CopySource !== 'string') {
+        return callback(errors.InvalidArgument);
```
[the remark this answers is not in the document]
alexanderchan-scality: tried it against the aws cli, `InvalidArgument` is returned.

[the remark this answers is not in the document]
bennettbuchanan: I guess the CLI transforms the argument to a string before making the request. Using the SDK, it returns `InvalidParameterType`.

alexanderchan-scality:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
SCALITY_INTEGRATION_BRANCH=ultron/rel/7.4
SCALITY_S3_BRANCH=ft/S3C-1115/GCP-COPY
try

alexanderchan-scality:
@ironman-machine 
DEFAULT_BRANCH=rel/7.4
SCALITY_INTEGRATION_BRANCH=ultron/rel/7.4
SCALITY_S3_BRANCH=ft/S3C-1115/GCP-COPY
try

PR #1100, opened 2018-01-31
S3C-1026: tcp leak

flavien-scality:
Harden handling of events at different stages, so that we close sockets fds in
all situations, especially when the S3 client aborts the connection while we
are processing a GET request. This used to generate a FD leak, leading to
unread tcp buffers and memory leak (observed in netstat output).


LaurenSpiegel:
@ironman-machine try

PR #1102, opened 2018-02-01
Ft/s3 c 1179/gcp managed upload

alexanderchan-scality:
Review: cf5c359439810c32deff91179744f4b4a2b04f59
Fix: 0325c2338cafc88be54c5e1695d067dd1a5988b3

### DIFF
--- a/lib/data/external/GCP/GcpManagedUpload.js
+++ b/lib/data/external/GCP/GcpManagedUpload.js
@@ -1,0 +1,409 @@
+const async = require('async');
+const assert = require('assert');
+const stream = require('stream');
+const { errors } = require('arsenal');
+const { minimumAllowedPartSize, maximumAllowedPartCount } =
+    require('../../../../constants');
+const { createMpuList, logger } = require('./GcpUtils');
+const { logHelper } = require('../utils');
+
+
+function sliceFn(body, size) {
+    const array = [];
+    let partNumber = 1;
+    for (let ind = 0; ind < body.length; ind += size) {
+        array.push({
+            Body: body.slice(ind, ind + size),
+            PartNumber: partNumber++,
+        });
+    }
+    return array;
+}
+
+class GcpManagedUpload {
+    /**
+     * GcpMangedUpload - class to mimic the upload method in AWS-SDK
+     * To-Do: implement retry on failure like S3's upload
+     * @param {GcpService} service - client object
+     * @param {object} params - upload params
+     * @param {string} params.Bucket - bucket name
+     * @param {string} params.MPU - mpu bucket name
+     * @param {string} params.Overflow - overflow bucket name
+     * @param {string} params.Key - object key
+     * @param {object} options - config setting for GcpManagedUpload object
+     * @param {number} options.partSize - set object chunk size
+     * @param {number} options.queueSize - set the number of concurrent upload
+     * @return {object} - return an GcpManagedUpload object
+     */
+    constructor(service, params, options = {}) {
+        this.service = service;
+        this.params = params;
+        this.mainBucket =
+            this.params.Bucket || this.service.config.mainBucket;
+        this.mpuBucket =
+            this.params.MPU || this.service.config.mpuBucket;
+        this.overflowBucket =
+            this.params.Overflow || this.service.config.overflowBucket;
+
+        this.partSize = minimumAllowedPartSize;
+        this.queueSize = options.queueSize || 4;
+        this.validateBody();
+        this.setPartSize();
+        // multipart information
+        this.parts = {};
+        this.uploadedParts = 0;
+        this.activeParts = 0;
+        this.partBuffers = [];
+        this.partQueue = [];
+        this.partBufferLength = 0;
+        this.totalChunkedBytes = 0;
+        this.partNumber = 0;
+    }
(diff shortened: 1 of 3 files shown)


lib/data/external/GCP/GcpManagedUpload.js
```diff
@@ -0,0 +1,432 @@
+const async = require('async');
+const stream = require('stream');
+const { lstatSync } = require('fs');
+const { errors } = require('arsenal');
+const { minimumAllowedPartSize, maximumAllowedPartCount } =
+    require('../../../../constants');
+const { createMpuList } = require('./GcpUtils');
+
+function sliceFn(size) {
+    this.array = [];
```
bennettbuchanan: Why use `this` and not a constant variable?

lib/data/external/GCP/GcpManagedUpload.js
```diff
@@ -0,0 +1,432 @@
...
+
+    setPartSize() {
+        if (this.totalBytes) {
+            const newPartSize =
+                Math.ceil(this.totalBytes / maximumAllowedPartCount);
+            if (newPartSize > this.partSize) this.partSize = newPartSize;
+            this.totalParts = Math.ceil(this.totalBytes / this.partSize);
+        }
+        if (this.body instanceof Buffer && this.totalParts > 1) {
+            this.slicedParts = sliceFn.call(this.body, this.partSize);
```
bennettbuchanan: I'm not sure if we need to use the `call()` method because we are not using any values from the `this` object in the function.

lib/data/external/GCP/GcpManagedUpload.js
alexanderchan-scality: Will fix this.

lib/data/external/GCP/GcpManagedUpload.js
alexanderchan-scality: Will fix this.

lib/data/external/GCP/GcpManagedUpload.js
```diff
@@ -202,33 +178,35 @@ class GcpManagedUpload {
...
-                // stream type
-                this.body.on('error', err => this.cleanUp(err))
-                .on('readable', () => this.chunkStream())
-                .on('end', () => {
-                    this.isDoneChunking = true;
-                    this.chunkStream();
+        this.called = true;
+        this.callback = callback;
+        assert.strictEqual(typeof callback, 'function',
+            new Error('Missing callback function for GcpManagedUpload'));
```
bennettbuchanan: We shouldn't need to instantiate a new `Error` here because it will be thrown as an `AssertionError` anyway.

lib/data/external/GCP/GcpManagedUpload.js
```diff
@@ -258,49 +239,48 @@ class GcpManagedUpload {
...
+                this.service.uploadPart(params, (err, res) => {
+                    if (!err) {
+                        this.parts[uploadPart.PartNumber] = res.ETag;
+                        this.uploadedParts++;
+                    }
+                    return done(err);
+                });
+            }, next),
+        ], err => {
+            assert.ifError(err, new Error(
```
bennettbuchanan: Same here about `new Error`.

lib/data/external/GCP/GcpManagedUpload.js
```diff
@@ -65,80 +68,57 @@ class GcpManagedUpload {
      */
     validateBody() {
         this.body = this.params.Body;
-        if (!this.body) {
-            this.error = errors.InvalidArgument
-                .customizeDescription('Body parameter is required');
+        assert(this.body, errors.MissingRequestBodyError.customizeDescription(
```
bennettbuchanan: An `AssertionError` should suffice. However, I am wondering what impact that these assertions will have. Would this not effectively stop the server when we want to instead log an error and return an error? @jonathan-gramain 

Maybe I'm mistaken about that, but was wondering because you removed the test for this case.

alexanderchan-scality: Yes. the assertion does stop the server.

alexanderchan-scality: Have made the necessary changes to wrap the construction of the GcpManageUpload constructor in a try/catch. If an error is thrown, the callback will be called with the returned error.

alexanderchan-scality:
@ironman-machine
DEFAULT_BRANCH=rel/7.4
SCALITY_INTEGRATION_BRANCH=test/zenkobb
SCALITY_FEDERATION_BRANCH=ft/S3C-1114/GCP-BACKEND
SCALITY_S3_BRANCH=ft/S3C-1179/GCP-ManagedUpload
try

alexanderchan-scality:
@ironman-machine
DEFAULT_BRANCH=rel/7.4
SCALITY_INTEGRATION_BRANCH=test/zenkobb
SCALITY_FEDERATION_BRANCH=ft/S3C-1114/GCP-BACKEND
SCALITY_S3_BRANCH=ft/S3C-1179/GCP-ManagedUpload
try

alexanderchan-scality:
http://ci.ironmann.io/gh/scality/Integration/19533

PR #1104, opened 2018-02-05
S3C-1245 update dependencies

harperscality:


DavidPineauScality:
Merging manually without E2E, as this only affects package.json and tests are passing.

PR #1107, opened 2018-02-06
FX: GCP PUT API Test

alexanderchan-scality:
Fixes the use of an undefined variable that leads to creation of buckets
that aren't deleted on GCP.


bennettbuchanan:
@ironman-machine r+

PR #1114, opened 2018-02-08
Fx/gcp put test

alexanderchan-scality:
Changing base from master to feature branch ft/GCP
Refer to #1107


alexanderchan-scality:
@ironman-machine try

PR #1140, opened 2018-02-25
Fix: limit exported config items

RachedBenMustapha:
Only send `overlayVersion` for now to avoid sending unwanted items.


bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=master
r+

PR #1145, opened 2018-03-06
Ft/objdel add modified header check

dora-korpar:
Adds check for date modified headers in objectDelete API

### DIFF
--- a/lib/api/objectDelete.js
+++ b/lib/api/objectDelete.js
@@ -1,6 +1,6 @@
 const async = require('async');
-const { errors, versioning } = require('arsenal');
-
+const { errors, versioning, s3middleware } = require('arsenal');
+const checkDateModifiedHeaders = s3middleware.checkDateModifiedHeaders;
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const services = require('../services');
 const { pushMetric } = require('../utapi/utilities');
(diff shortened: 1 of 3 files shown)


lib/api/objectDelete.js
```diff
@@ -99,6 +99,21 @@ function objectDelete(authInfo, request, log, cb) {
...
+        function validateHeaders(bucketMD, objectMD, options, next) {
+            if (objectMD) {
+                const lastModified = objectMD['last-modified'];
+                const { modifiedSinceRes, unmodifiedSinceRes } =
+                    checkDateModifiedHeaders(request.headers, lastModified);
+                const err = modifiedSinceRes.error || unmodifiedSinceRes.error;
+                if (err) {
+                    return process.nextTick(() => next(err, bucketMD));
+                }
+                return process.nextTick(() =>
```
rahulreddy: This return is redundant as it will go the next one anyways.

philipyoo: curious, why `process.nextTick`?

dora-korpar: cos everything in the function is synchronous, so we gotta mimic async :)

philipyoo: cool beans ty for the explanation :)

PR #1147, opened 2018-03-06
FIX: Make changes for recent Arsenal updates

bennettbuchanan:


bennettbuchanan:
Closing this because fixes are to be added in https://github.com/scality/S3/pull/1129.

PR #1155, opened 2018-03-12
Ft/update test

dora-korpar:
With the Lifecycle model changes in https://github.com/scality/Arsenal/pull/440, S3 tests had to be updated


dora-korpar:
@ironman-machine try

bennettbuchanan:
@ironman-machine r+

PR #1162, opened 2018-03-15
FT: Add CRR statistics for Orbit

bennettbuchanan:
This is the first effort to use forthcoming Backbeat routes for retrieving metrics (to be added by https://github.com/scality/backbeat/pull/200). The next step is to extract metrics accumulation code and move it to Arsenal so that we can avoid making an HTTP request to Backbeat.

### DIFF
--- a/config.json
+++ b/config.json
@@ -34,6 +34,10 @@
         "site": "us-east-2",
         "type": "aws_s3"
     }],
+    "backbeat": {
+        "host": "localhost",
+        "port": 8900
+    },
     "cdmi": {
         "host": "localhost",
         "port": 81,
--- a/lib/utilities/reportHandler.js
+++ b/lib/utilities/reportHandler.js
@@ -78,50 +78,39 @@ function getCRRStats(log, cb) {
     log.debug('getting CRR stats', { method: 'getCRRStats' });
     // TODO: Reuse metrics code from Backbeat by moving it to Arsenal instead of
     // making an HTTP request to the Backbeat metrics route.
-    const params = { url: 'http://localhost:8900/_/metrics/crr/all' };
+    const { host, port } = config.backbeat;
+    const params = { url: `http://${host}:${port}/_/metrics/crr/all` };
     return request.get(params, (err, res) => {
         if (err) {
             log.error('failed to get CRR stats', {
                 method: 'getCRRStats',
                 error: err,
             });
-            return cb(err);
+            return cb(null, {});
         }
-        let body;
-        try {
-            body = JSON.parse(res.body);
-        } catch (parseErr) {
-            log.error('could not parse backbeat response', {
-                method: 'getCRRStats',
-                error: parseErr,
-            });
-            return cb(errors.InternalError
-                .customizeDescription('could not parse backbeat response'));
-        }
-        const { completions, backlog, throughput } = body;
+        const { completions, backlog, throughput } = res.body;
         if (!completions || !backlog || !throughput) {
             log.error('could not get metrics from backbeat', {
                 method: 'getCRRStats',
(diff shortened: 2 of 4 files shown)


lib/utilities/reportHandler.js
```diff
@@ -73,6 +74,56 @@ function getSystemStats() {
...
+    // TODO: Reuse metrics code from Backbeat by moving it to Arsenal instead of
+    // making an HTTP request to the Backbeat metrics route.
+    const params = { url: 'http://localhost:8900/_/metrics/crr/all' };
+    return request.get(params, (err, res) => {
+        if (err) {
+            log.error('failed to get CRR stats', {
+                method: 'getCRRStats',
+                error: err,
+            });
+            return cb(err);
```
RachedBenMustapha: Backbeat being down should not cause the whole report to fail so it would be wiser to return `(null, {})`.

lib/utilities/reportHandler.js
```diff
@@ -73,6 +74,56 @@ function getSystemStats() {
...
+        }
+        let body;
+        try {
+            body = JSON.parse(res.body);
```
RachedBenMustapha: `request(...).json()` would have the same effect

lib/utilities/reportHandler.js
```diff
@@ -73,6 +74,56 @@ function getSystemStats() {
...
+        if (!completions || !backlog || !throughput) {
+            log.error('could not get metrics from backbeat', {
+                method: 'getCRRStats',
+            });
+            return cb(errors.InternalError
+                .customizeDescription('could not get metrics from backbeat'));
+        }
+        const stats = {
+            completions: {
+                count: parseFloat(completions.results.count),
```
RachedBenMustapha: The swagger definition expects a `int64` instead of a float

lib/utilities/reportHandler.js
```diff
@@ -73,6 +74,56 @@ function getSystemStats() {
     };
 }
 
+function getCRRStats(log, cb) {
+    log.debug('getting CRR stats', { method: 'getCRRStats' });
+    // TODO: Reuse metrics code from Backbeat by moving it to Arsenal instead of
+    // making an HTTP request to the Backbeat metrics route.
+    const params = { url: 'http://localhost:8900/_/metrics/crr/all' };
```
RachedBenMustapha: even if it's temporary, having a configuration would be easier because we currently need to support 4 different stacks (swarm, kube, federation, sandboxes) and not all of them are able to reach backbeat at `localhost:8900`.

lib/utilities/reportHandler.js
RachedBenMustapha: I think the approximation is fine, the volume of data that zenko's meant to handle makes the rounding errors not matter much :)

config.json
```diff
@@ -34,6 +34,10 @@
         "site": "us-east-2",
         "type": "aws_s3"
     }],
+    "backbeat": {
```
RachedBenMustapha: To make it usable out of the box and ease testing, could we handle `CRR_METRICS_HOST` and `CRR_METRICS_PORT` environment variables to update this directly in `docker-endpoint.sh`?

RachedBenMustapha:
Almost there for the temporary version :)

config.json
bennettbuchanan: You got it! Updated so that you can start the image with a custom endpoint as such:

```
docker run -e CRR_METRICS_HOST=10.100.3.139 -e CRR_METRICS_PORT=1337 a7ae6fa66070
```

RachedBenMustapha:
Ready to squash :)

PR #1170, opened 2018-03-19
Ft/zenko endpoints

ssalaues:
Updates to the docker entry point script that allows csv style strings for multiple endpoint to be passed through. Still retains compatibility when only one endpoint is specified.

2nd commit fixes the mongodb hosts pass through since there was a name change from "hosts" to "replicaSetHosts"


vrancurel:
Merging manually because cannot be tested

PR #1172, opened 2018-03-20
S3C-1348 FT: Integrating 1-many locations replication into Orbit

nicolas2bert:


lib/management.js
```diff
@@ -146,6 +146,17 @@ function patchConfiguration(instanceId, newConf, log, cb) {
                     'constraints', { error });
                 return cb(error);
             }
+            try {
+                const locationsWithNoScality = locations
+                // NOTE: In Orbit, we don't need to have Scality location in our
+                // replication endpoind config, since we do not replicate to
+                // any Scality Instance yet.
+                .filter(key => locations[key].type !== 'file');
```
bennettbuchanan: Is it possible there would be a `mem` type, or some other type for which replication is not supported? 

I would suggest using a constant variable for supported replication backends, like the one to being introduced here https://github.com/scality/S3/pull/1154/files#diff-88903d9c789529ef76d3636ebafcdb40R118

nicolas2bert: `file` in orbit is the default location type we use, there is no `mem` but it might one day change so using `constants.replicationBackends` is, I think, a good idea. Thanks @bennettbuchanan

lib/management.js
```diff
@@ -163,6 +164,23 @@ function patchConfiguration(instanceId, newConf, log, cb) {
                     'constraints', { error });
                 return cb(error);
             }
+            try {
+                const locationsWithNoScality = Object.keys(locations)
```
jonathan-gramain: What about naming this `locationsWithReplicationBackend`?

nicolas2bert: Since I changed the way I filter it, your constant naming makes more sense. Thanks for the review @jonathan-gramain

PR #1174, opened 2018-03-20
Ft/zenko 20

anurag4DSB:
###  This pull request is for Zenko. A Prometheus server will pull metrics from S3 at the route: '/_/monitoring/metrics'. I created a new route handler for the same.

This change is required as all the components in Zenko are monitored by Prometheus. We need S3 too.

### DIFF
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -19,7 +19,7 @@ function writeResponse(res, error, log, results, cb) {
 
 
 function routeHandler(req, res, log, cb) {
-    if (req.method !== 'GET' && req.method !== 'POST') {
+    if (req.method !== 'GET') {
         return cb(errors.BadRequest, []);
     }
     client.collectDefaultMetrics();
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -26,7 +26,7 @@ function routeHandler(req, res, log, cb) {
     const promMetrics = client.register.metrics();
     const contentLen = Buffer.byteLength(promMetrics, 'utf8');
     res.setHeader('content-length', contentLen);
-    res.setHeader('content-type', 'application/json');
+    res.setHeader('content-type', client.register.contentType);
     res.end(promMetrics);
     return undefined;
 }
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -1,0 +1,65 @@
+const errors = require('arsenal');
+const client = require('prom-client');
+
+function writeResponse(res, error, log, results, cb) {
+    let statusCode = 200;
+    if (error) {
+        if (Number.isInteger(error.code)) {
+            statusCode = error.code;
+        } else {
+            statusCode = 500;
+        }
+    }
+    res.writeHead(statusCode, { 'Content-Type': 'application/json' });
+    res.write(JSON.stringify(results));
+    res.end(() => {
+        cb(error, results);
+    });
+}
+
+
+function routeHandler(req, res, log, cb) {
+    if (req.method !== 'GET' && req.method !== 'POST') {
+        return cb(errors.BadRequest, []);
+    }
+    client.collectDefaultMetrics();
+    const promMetrics = client.register.metrics();
+    const contentLen = Buffer.byteLength(promMetrics, 'utf8');
+    res.setHeader('content-length', contentLen);
+    res.setHeader('content-type', 'application/json');
+    res.end(promMetrics);
+    return undefined;
+}
+
+/**
+ * Checks if client IP address is allowed to make http request to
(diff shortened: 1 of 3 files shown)


lib/utilities/monitoringHandler.js
```diff
@@ -0,0 +1,65 @@
...
+        cb(error, results);
+    });
+}
+
+
+function routeHandler(req, res, log, cb) {
+    if (req.method !== 'GET') {
+        return cb(errors.BadRequest, []);
+    }
+    client.collectDefaultMetrics();
```
NicolasT: This should be only called once at application startup, not per-request?

lib/utilities/monitoringHandler.js
```diff
@@ -0,0 +1,65 @@
...
+    const promMetrics = client.register.metrics();
+    const contentLen = Buffer.byteLength(promMetrics, 'utf8');
+    res.setHeader('content-length', contentLen);
+    res.setHeader('content-type', 'application/json');
```
NicolasT: Wrong MIME-type for the response. There's a constant defined I believe in prom-client, see https://github.com/siimon/prom-client#utilites

lib/utilities/monitoringHandler.js
```diff
@@ -0,0 +1,65 @@
...
+    res.end(promMetrics);
+    return undefined;
```
NicolasT: Shouldn't be required.

lib/utilities/monitoringHandler.js
```diff
@@ -0,0 +1,65 @@
...
+
+
+function routeHandler(req, res, log, cb) {
+    if (req.method !== 'GET') {
+        return cb(errors.BadRequest, []);
+    }
+    client.collectDefaultMetrics();
+    const promMetrics = client.register.metrics();
+    const contentLen = Buffer.byteLength(promMetrics, 'utf8');
+    res.setHeader('content-length', contentLen);
```
NicolasT: Shouldn't be required, and maybe even set by the HTTP server module? Anyway, calculating the length can be somewhat expensive without a need.

lib/utilities/monitoringHandler.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+    res.writeHead(statusCode, { 'Content-Type': 'application/json' });
+    res.write(JSON.stringify(results));
+    res.end(() => {
+        cb(error, results);
+    });
+}
+
+
+function routeHandler(req, res, log, cb) {
```
NicolasT: In some control flows, `cb` is not called, which is slightly odd.

lib/utilities/monitoringHandler.js
anurag4DSB: This is to satisfy CI. Es-lint expects me to return

lib/utilities/monitoringHandler.js
anurag4DSB: This was suggested by the lead. It is required.

lib/utilities/monitoringHandler.js
anurag4DSB: So what do you suggest the timeout value to be. I thought we can control it better with the server controlling the timeout.

lib/utilities/monitoringHandler.js
anurag4DSB: As advised I am following the healthcheckHandler

lib/utilities/monitoringHandler.js
anurag4DSB: If I check the type, its string.
So it should be text/plain.
Is that okay?

lib/utilities/monitoringHandler.js
anurag4DSB: Also, existing intervals are automatically cleared when calling collectDefaultMetrics.

lib/utilities/monitoringHandler.js
NicolasT: > The content-type prometheus expects is also exported as a constant, both on the register and from the main file of this project, called contentType.

lib/utilities/monitoringHandler.js
NicolasT: I'd just set the interval to 5s or so, we won't be using a higher sampling rate, and not call it for every server request.

lib/utilities/monitoringHandler.js
```diff
@@ -0,0 +1,65 @@
...
+function writeResponse(res, error, log, results, cb) {
+    let statusCode = 200;
+    if (error) {
+        if (Number.isInteger(error.code)) {
+            statusCode = error.code;
+        } else {
+            statusCode = 500;
+        }
+    }
+    res.writeHead(statusCode, { 'Content-Type': 'text/plain' });
```
NicolasT: This is likely `application/json`, looking at what's below?

PR #1181, opened 2018-03-28
FX: granting '/replication' user permission to read/write buckets and objects

nicolas2bert:
This PR grants '/replication' user permission to read and write any buckets and objects.

### DIFF
--- a/lib/auth/in_memory/builder.js
+++ b/lib/auth/in_memory/builder.js
@@ -4,23 +4,23 @@ const serviceAccountPrefix =
 /** build simple authdata with only one account
  * @param {string} accessKey - account's accessKey
  * @param {string} secretKey - account's secretKey
+ * @param {string} canonicalId - account's canonical id
  * @param {string} [serviceName] - service name to use to generate can id
  * @return {object} authdata - authdata with account's accessKey and secretKey
  */
-function buildAuthDataAccount(accessKey, secretKey, serviceName) {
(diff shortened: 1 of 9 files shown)


lib/auth/in_memory/builder.js
```diff
@@ -4,23 +4,23 @@ const serviceAccountPrefix =
 /** build simple authdata with only one account
  * @param {string} accessKey - account's accessKey
  * @param {string} secretKey - account's secretKey
+ * @param {string} canonicalId - account's canonical id
  * @param {string} [serviceName] - service name to use to generate can id
  * @return {object} authdata - authdata with account's accessKey and secretKey
  */
-function buildAuthDataAccount(accessKey, secretKey, serviceName) {
+function buildAuthDataAccount(accessKey, secretKey, canonicalId, serviceName) {
```
RachedBenMustapha: Now that service account canonical ids contain the service type, we can remove the `serviceName` handling altogether.

nicolas2bert: I kept `serviceName` for legacy purpose. If the canonicalId is not specified `buildAuthDataAccount` create a canonical id as: `${serviceAccountPrefix}/${serviceName}`

nicolas2bert:
@dora-korpar I added some tests :)

RachedBenMustapha:
Merging into the feature branch

PR #1184, opened 2018-03-29
transient source

jonathan-gramain:
Linked to https://github.com/scality/backbeat/pull/228 and https://github.com/scality/Arsenal/pull/454

ft: set nonTransientLocation property in object metadata

Set the nonTransientLocation property to 'PENDING:preferred_site' in
object metadata (plain, mpu and copy), when the object location has
the 'isTransient' attribute set.

This will be used by CRR processors and status processor to know if
the object location has to be transitioned to the preferred CRR
target.

Return the Location attribute from backbeat route for backbeat to
write the proper non-transient location in the object metadata once
CRRed.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -164,6 +164,10 @@ function locationConstraintAssert(locationConstraints) {
             === 'boolean',
             'bad config: locationConstraints[region]' +
             '.legacyAwsBehavior is mandatory and must be a boolean');
+        assert(['undefined', 'boolean'].includes(
+            typeof locationConstraints[l].isTransient),
+               'bad config: locationConstraints[region]' +
+               '.isTransient must be a boolean');
         const details = locationConstraints[l].details;
         assert(typeof details === 'object',
             'bad config: locationConstraints[region].details is ' +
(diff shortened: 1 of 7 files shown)


lib/Config.js
```diff
@@ -389,6 +397,15 @@ class Config extends EventEmitter {
                     });
                 }
             });
+            const defaultEndpoint = replicationEndpoints.find(
+                endpoint => endpoint.default);
+            if (Object.keys(this.locationConstraints)
+                .map(k => this.locationConstraints[k])
+                .some(location => location.isTransient)) {
+                assert(defaultEndpoint.type, 'bad config: ' +
```
bennettbuchanan: I think it's worth noting if there is only one replication endpoint the default can be omitted. See: https://github.com/scality/S3/pull/1184/files#diff-9280903abc69f09eff9ab36070b6d751R355. I don't think it would make sense to have a transient location with only one replication endpoint, but just want to make sure we don't happen to run into a problem accessing `type` of `undefined`.

jonathan-gramain: ok will add a check for this, thanks!

jonathan-gramain:
POC only

PR #1188, opened 2018-03-30
Fix mongo tests

ssalaues:
Fixes the problem where the functional tests with mongo would not run correctly causing CircleCi to falsely pass. 

Uploaded a new ci-mongo docker image based off Federation and that self initializes to a single node replica set for compatibility.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -29,6 +29,8 @@ machine:
 
 dependencies:
   override:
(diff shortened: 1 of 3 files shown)


ssalaues:
Added exit $? at the end of the script so it always returns the last exit code instead of forcing an exit 0

ssalaues:
Rebased and fixed conflicts

ssalaues:
Rebased and changed the ci-mongo image to 3.4

PR #1190, opened 2018-03-30
FX: functional tests

nicolas2bert:
These changes should fix our tests.
This check was not skipping tests: https://github.com/scality/S3/blob/master/tests/functional/aws-node-sdk/test/multipleBackend/utils.js#L40

### DIFF
--- a/tests/functional/report/checkRoutes.js
+++ b/tests/functional/report/checkRoutes.js
@@ -69,7 +69,7 @@ describe('Report route', () => {
 
     it('should contain config', done => {
         queryReport(done, response => {
-            if (!response.config || !response.config.locationConstraints) {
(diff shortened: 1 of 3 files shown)


tests/functional/report/checkRoutes.js
```diff
@@ -69,7 +69,7 @@ describe('Report route', () => {
 
     it('should contain config', done => {
         queryReport(done, response => {
-            if (!response.config || !response.config.locationConstraints) {
+            if (!response.config || !response.config.overlayVersion) {
```
nicolas2bert: NOTE: The way we do the report config cleanup has changed: https://github.com/scality/S3/pull/1183/files#diff-2157ec89ef3a5f5eefc407fecbdba988R14

PR #1195, opened 2018-04-05
ft: provide pathStyle option to support S3C

rahulreddy:
This feature allows setting pathStyle requests option for location constraints so that
it can be relaxed for non-AWS S3 backends (for example S3 Connector)


rahulreddy:
@ironman-machine try

PR #1196, opened 2018-04-05
FIX: Add flaky test retry and additional logging

bennettbuchanan:
This test has been observed both passing and failing with the same code base (see node 6):

Passed: http://ci.ironmann.io/gh/scality/Integration/20988
Failed: http://ci.ironmann.io/gh/scality/Integration/20997

The proposed change is to let the test run three consecutive times before the suite fails and log additional output for further investigation, should the issue arise again. This is a step towards reducing its flakiness but not a final solution. A reliable fix should be made, or the test parameters (e.g., timeout length) could be re-evaluated.

### DIFF
--- a/tests/functional/aws-node-sdk/test/multipleBackend/get/getAzure.js
+++ b/tests/functional/aws-node-sdk/test/multipleBackend/get/getAzure.js
@@ -73,7 +73,8 @@ function testSuite() {
 
                 it(`should get an ${key.describe} object from Azure`, done => {
                     // Log the key name to help investigate potential flakiness.
-                    process.stdout.write(`key: ${testKey}`);
+                    process.stdout.write(`should get an ${key.describe} ` +
+                        `object from Azure test key name: ${testKey}`);
                     s3.getObject({ Bucket: azureContainerName, Key:
                       testKey },
                         (err, res) => {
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
Replying to my own comment, it may not be that easy to retry in the code if the data is streamed and the error happens after some data has already been sent back.

tests/functional/aws-node-sdk/test/multipleBackend/get/getAzure.js
```diff
@@ -53,7 +53,10 @@ function testSuite() {
             });
         });
         keys.forEach(key => {
-            describe(`${key.describe} size`, () => {
+            describe(`${key.describe} size`, function fn() {
```
rahulreddy: I just noticed this. Having describe in a forEach is a bad idea as it's not synchronous, meaning two tests run simultaneously and if they happen to use the same bucket name or key name, they are bound to have a collision. We have to clean this up later.

tests/functional/aws-node-sdk/test/multipleBackend/get/getAzure.js
```diff
@@ -53,7 +53,10 @@ function testSuite() {
...
+                // This test has been observed to be flaky, so allow the test to
+                // fail three times consecutively before the suite itself fails.
+                this.retries(2);
```
rahulreddy: The other option is to use `this.skip()` on tests known to be flaky.

PR #1197, opened 2018-04-06
FX: HTTP requests to AWS

nicolas2bert:
Error: `NetworkingError: Protocol "http:" not supported. Expected "https:"`

### DIFF
--- a/lib/management.js
+++ b/lib/management.js
@@ -128,6 +128,7 @@ function patchConfiguration(instanceId, newConf, log, cb) {
                                 Boolean(l.details.serverSideEncryption),
                             awsEndpoint: l.details.endpoint ||
                                 's3.amazonaws.com',
+                            https: true,
                         };
                     }
                     break;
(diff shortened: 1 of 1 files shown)


lib/management.js
```diff
@@ -128,6 +128,7 @@ function patchConfiguration(instanceId, newConf, log, cb) {
                                 Boolean(l.details.serverSideEncryption),
                             awsEndpoint: l.details.endpoint ||
                                 's3.amazonaws.com',
+                            https: true,
```
rahulreddy: This should have been the default if `https` option is not defined.

nicolas2bert: We use `setLocationConstraints()` to update the config in Orbit management.js https://github.com/scality/S3/blob/master/lib/management.js#L160 and so this function `locationConstraintAssert` never get called https://github.com/scality/S3/blob/master/lib/Config.js#L204. So I m guessing that for config management we have to specify it unless we change our checks here: https://github.com/scality/S3/blob/master/lib/data/locationConstraintParser.js#L63

rahulreddy:
Closing this in favor of https://github.com/scality/S3/pull/1200

PR #1204, opened 2018-04-12
ft/ZENKO-229 get using location

dora-korpar:
Adds functionality to get object from specific location by passing header 'x-amz-location-constraint'

### DIFF
--- a/lib/api/apiUtils/object/getReplicationBackendDataLocator.js
+++ b/lib/api/apiUtils/object/getReplicationBackendDataLocator.js
@@ -1,0 +1,45 @@
+const { errors } = require('arsenal');
+
+/**
+ * getReplicationBackendDataLocator - compares given location constraint to
+ * replication backends
+ * @param {object} locationObj - object containing location information
+ * @param {string} locationObj.location - name of location constraint
+ * @param {string} locationObj.key - keyname of object in location constraint
+ * @param {string} locationObj.locationType - type of location constraint
+ * @param {object} replicationInfo - information about object replication
+ * @param {array} replicationInfo.backends - array containing information about
+ * each replication location
+ * @param {string} replicationInfo.backends[].site - name of replication
+ * location
+ * @param {string} replicationInfo.backends[].status - status of replication
+ * @param {string} replicationInfo.backends[].dataStoreVersionId - version id
+ * of object at replication location
+ * @return {object} contains error if no replication backend matches or
+ * dataLocator object
+ */
+function getReplicationBackendDataLocator(locationObj, replicationInfo) {
+    const repBackendResult = {};
+    const locMatch = replicationInfo.backends.find(
+        backend => backend.site === locationObj.location);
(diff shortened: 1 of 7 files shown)


lib/api/apiUtils/object/replicationBackendCompare.js
```diff
@@ -0,0 +1,44 @@
...
+    const repBackendResult = {};
+    const locMatch = replicationInfo.backends.find(
+        backend => backend.site === locationObj.location);
+    if (!locMatch) {
+        repBackendResult.error = errors.InvalidLocationConstraint.
+            customizeDescription('Object is not replicated to location ' +
+            'passed in location header');
+        return repBackendResult;
+    }
+    if (locMatch.status === 'PENDING') {
```
jonathan-gramain: We may also want to handle `FAILED` status the same way, or alternatively, return the locator if status is `COMPLETED`, and fail otherwise and provide the replication status as an extra info in the returned message.

lib/api/apiUtils/object/replicationBackendCompare.js
```diff
@@ -0,0 +1,44 @@
...
+ * each replication location
+ * @param {string} replicationInfo.backends[].site - name of replication
+ * location
+ * @param {string} replicationInfo.backends[].status - status of replication
+ * @param {string} replicationInfo.backends[].dataStoreVersionId - version id
+ * of object at replication location
+ * @return {object} contains error if no replication backend matches or
+ * dataLocator object
+ */
+function replicationBackendCompare(locationObj, replicationInfo) {
```
jonathan-gramain: What about renaming to `getReplicationBackendDataLocator`, since this is the main purpose of the function (and especially since the function is exported as a utility function potentially useful for others - thinking of transient source :smile:).

dora-korpar:
@ironman-machine try

lib/api/apiUtils/object/getReplicationBackendDataLocator.js
```diff
@@ -0,0 +1,45 @@
...
+        backend => backend.site === locationObj.location);
+    if (!locMatch) {
+        repBackendResult.error = errors.InvalidLocationConstraint.
+            customizeDescription('Object is not replicated to location ' +
+            'passed in location header');
+        return repBackendResult;
+    }
+    if (['PENDING', 'FAILED'].includes(locMatch.status)) {
+        repBackendResult.error = errors.NoSuchKey.customizeDescription(
+            'Object replication to specified backend is not complete');
```
bennettbuchanan: I could imagine it being nice to include the status in the description, particularly if the status is failed. That way the user would know to retry the operation.

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

PR #1205, opened 2018-04-12
FIX: WebsiteRedirectLocation header response

giacomoguiulfo:
Compatibility fix with AWS S3. The WebsiteRedirectLocation will not be included in the get object or head responses if it was not set on the object.

AWS S3 output:
```json
{
    "AcceptRanges": "bytes",
    "ContentType": "binary/octet-stream",
    "LastModified": "Tue, 10 Apr 2018 05:42:07 GMT",
    "ContentLength": 0,
    "Metadata": {}
}
```

Cloudserver previous output:
```json
{
    "LastModified": "Tue, 10 Apr 2018 05:46:03 GMT",
    "ContentLength": 0,
    "WebsiteRedirectLocation": "",
    "Metadata": {}
}
```

### DIFF
--- a/lib/utilities/collectResponseHeaders.js
+++ b/lib/utilities/collectResponseHeaders.js
@@ -15,8 +15,7 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
   returnTagCount) {
     // Add user meta headers from objectMD
     const responseMetaHeaders = Object.assign({}, corsHeaders);
-    Object.keys(objectMD).filter(val => (val.substr(0, 11) === 'x-amz-meta-' ||
-        val === 'x-amz-website-redirect-location'))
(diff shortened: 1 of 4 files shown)


lib/utilities/collectResponseHeaders.js
```diff
@@ -15,8 +15,7 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
   returnTagCount) {
     // Add user meta headers from objectMD
     const responseMetaHeaders = Object.assign({}, corsHeaders);
-    Object.keys(objectMD).filter(val => (val.substr(0, 11) === 'x-amz-meta-' ||
-        val === 'x-amz-website-redirect-location'))
+    Object.keys(objectMD).filter(val => (val.substr(0, 11) === 'x-amz-meta-'))
```
philipyoo: Nothing wrong here, just another method you could use here is `String#startsWith` method and you won't need to specify the length

PR #1206, opened 2018-04-12
FIX: Add Accept-Ranges header to response

tmacro:
Compatibility fix for AWS S3.

AWS returns a `Accept-Ranges: bytes` header with a GET or HEAD request for an object.

 ```
{
    "AcceptRanges": "bytes",
    "ContentType": "binary/octet-stream",
    "LastModified": "Tue, 10 Apr 2018 05:42:07 GMT",
    "ContentLength": 0,
    "Metadata": {}
}
```
Cloudserver was missing this header, which can cause programs relying upon the `Range` header to be unable to detect support.

See #1134

### DIFF
--- a/lib/utilities/collectResponseHeaders.js
+++ b/lib/utilities/collectResponseHeaders.js
@@ -38,6 +38,9 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
         responseMetaHeaders['x-amz-server-side-encryption-aws-kms-key-id']
             = objectMD['x-amz-server-side-encryption-aws-kms-key-id'];
     }
+
+    responseMetaHeaders['Accept-Ranges'] = 'bytes';
+
     if (objectMD['cache-control']) {
         responseMetaHeaders['Cache-Control'] = objectMD['cache-control'];
     }
--- a/tests/functional/aws-node-sdk/test/object/get.js
+++ b/tests/functional/aws-node-sdk/test/object/get.js
@@ -224,7 +224,7 @@ describe('GET object', () => {
             });
 
         describe('Additional headers: [Cache-Control, Content-Disposition, ' +
-            'Content-Encoding, Expires]', () => {
+            'Content-Encoding, Expires, Accept-Ranges]', () => {
             describe('if specified in put object request', () => {
                 before(done => {
                     const params = {
@@ -255,6 +255,7 @@ describe('GET object', () => {
                           assert.strictEqual(res.ContentType, contentType);
                           assert.strictEqual(res.Expires,
                               new Date(expires).toGMTString());
+                          assert.strictEqual(res.AcceptRanges, 'bytes');
                           return done();
                       });
                 });
--- a/tests/unit/utils/collectResponseHeaders.js
+++ b/tests/unit/utils/collectResponseHeaders.js
@@ -19,4 +19,9 @@ describe('Middleware: Collect Response Headers', () => {
                 undefined);
         });
     });
+
+    it('should add the Accept-Ranges header', () => {
+        const headers = collectResponseHeaders({});
+        assert.strictEqual(headers['Accept-Ranges'], 'bytes');
+    });
 });

### CODE REVIEW

File: lib/utilities/collectResponseHeaders.js [hunk 1]
```diff
@@ -25,6 +25,8 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
     responseMetaHeaders['x-amz-version-id'] =
         getVersionIdResHeader(versioningCfg, objectMD);
 
+    responseMetaHeaders['Accept-Ranges'] = 'bytes';
```
bennettbuchanan: Awesome! Thinking out loud: perhaps this could be put near the other standard HTTP headers (e.g., 'Cache-Control', etc.), just below where the Amazon-specific headers (those beginning with 'x-amz-') are set.

lib/utilities/collectResponseHeaders.js
```diff
@@ -25,6 +25,7 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
     responseMetaHeaders['x-amz-version-id'] =
         getVersionIdResHeader(versioningCfg, objectMD);
 
+
```
JianqinWang: this may be too nitpicky, but it looks like there's an extra line here

lib/utilities/collectResponseHeaders.js
```diff
@@ -56,26 +60,26 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
...
-    Object.keys(objectMD.tags).length > 0) {
-        responseMetaHeaders['x-amz-tagging-count'] =
-          Object.keys(objectMD.tags).length;
-    }
-    if (objectMD.replicationInfo && objectMD.replicationInfo.status) {
-        responseMetaHeaders['x-amz-replication-status'] =
+        Object.keys(objectMD.tags).length > 0) {
+            responseMetaHeaders['x-amz-tagging-count'] =
+            Object.keys(objectMD.tags).length;
+        }
```
jonathan-gramain: It seems the indentation is off from this point

lib/utilities/collectResponseHeaders.js
```diff
@@ -56,26 +60,26 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
...
-        Array.isArray(objectMD.replicationInfo.backends)) {
-        objectMD.replicationInfo.backends.forEach(backend => {
-            const { status, site, dataStoreVersionId } = backend;
-            responseMetaHeaders[`x-amz-meta-${site}-replication-status`] =
+        }
+        if (objectMD.replicationInfo &&
+            // Use storageType to determine if user metadata is needed.
+            objectMD.replicationInfo.storageType &&
+            Array.isArray(objectMD.replicationInfo.backends)) {
+                objectMD.replicationInfo.backends.forEach(backend => {
```
jonathan-gramain: similar indentation issue

jonathan-gramain:
@rahulreddy As far as I know, the `Accept-Ranges` header is returned to inform of the server capabilities, it's not bound to a particular request, so it should be sent whether the object is 0-byte or not.

jonathan-gramain:
And if some of us wonder, it's HTTP-spec related, not AWS, so it's ok even if you don't see it documented in AWS API.

tmacro:
Here's a dump of the headers returned of a 0 byte file. `Accept-Ranges` is included

```
{  
    'accept-ranges': 'bytes',
    'content-length': '0',
    'content-type': 'binary/octet-stream',
    'date': 'Thu, 12 Apr 2018 '
            '23:34:19 GMT',
    'last-modified': 'Thu, 12 Apr '
                    '2018 23:32:09 '
                    'GMT',
    'server': 'AmazonS3',
}
```

tmacro:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=True
try

lib/utilities/collectResponseHeaders.js
```diff
@@ -38,12 +44,15 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
         responseMetaHeaders['x-amz-server-side-encryption-aws-kms-key-id']
             = objectMD['x-amz-server-side-encryption-aws-kms-key-id'];
     }
+
+    responseMetaHeaders['Accept-Ranges'] = 'bytes';
```
philipyoo, line 42: I see a duplicate here

rahulreddy:
@jonathan-gramain You would be surprised by how many times we were failing compatibility where AWS S3 was returning responses differently from the HTTP spec. It sounds common sense but not always. I usually encourage writing tests against corner cases so that it helps developers think of them.

PR #1207, opened 2018-04-13
Ft/more s3 locations support

alexanderchan-scality:
Adds more support options for S3 location constraints:
+ optional versioning for S3 backend
+ path style

### DIFF
--- a/lib/management/configuration.js
+++ b/lib/management/configuration.js
@@ -95,6 +95,9 @@ function patchConfiguration(instanceId, newConf, log, cb) {
             Object.keys(newConf.locations || {}).forEach(k => {
                 const l = newConf.locations[k];
                 const location = {};
+                let supportsVersioning = false;
+                let pathStyle = false;
+
                 switch (l.locationType) {
                 case 'location-mem-v1':
                     location.type = 'mem';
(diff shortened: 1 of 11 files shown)


lib/management.js
```diff
@@ -1,3 +1,4 @@
+const { URL } = require('url');
```
philipyoo: I don't think `url.URL` is supported in Node v6. Instead there is a property: `url.Url` but it doesn't seem to work as intended.

Based on this blog: https://www.contentful.com/blog/2016/11/02/node-v7-urls-deprecations-warnings-and-developer-experience/
Looks like support was added in v7?

alexanderchan-scality: https://nodejs.org/docs/latest-v6.x/api/url.html#url_the_whatwg_url_api
it is also added in 6.13

lib/management.js
```diff
@@ -111,11 +115,23 @@ function patchConfiguration(instanceId, newConf, log, cb) {
                         };
                     }
                     break;
-                case 'location-do-spaces-v1':
+                case 'location-scality-ring-s3-v1':
+                    pathStyle = true; // fallthrough
```
jonathan-gramain: I would recommend not relying on fallthrough capability of `switch`, instead using a map of properties per location name (`{ 'location-scality-bleh': { pathStyle: true, ...} }`) would be more readable and less error-prone.

alexanderchan-scality:
@ironman-machine try
RUN_BACKBEAT_CRR_TESTS=true

alexanderchan-scality:
@ironman-machine 
RUN_BACKBEAT_CRR_TESTS=true
try

lib/management/configuration.js
```diff
@@ -115,11 +119,23 @@ function patchConfiguration(instanceId, newConf, log, cb) {
                         };
                     }
                     break;
-                case 'location-do-spaces-v1':
+                case 'location-scality-ring-s3-v1':
+                    pathStyle = true; // fallthrough
```
jonathan-gramain: Reporting the style comment about using a mapping object rather than relying on fallthrough of a switch.

RachedBenMustapha: It's a good cleanup, we'll follow up on it

PR #1209, opened 2018-04-13
S3C-1412: start s3 unconditionally

scality-gdoumergue:
# Pull request

## Description

This PR makes S3 finish its initialization even if a sub-component fails.

### Motivation and context

Since 7.4.0, bucketd returns a 500 error to healthcheck requests as soon as a raft session doesn't have a leader. And this is a really helpful information we needed since a long time ago.

But it means that, even when 7 bucket raft sessions work properly, S3 won't start because its initial deep healthcheck prevents it to complete its startup.

This can lead to a total service interruption if the S3 connectors are rebooted/restarted when some, not all, buckets are unavailable.

No tests added.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -159,20 +159,20 @@ class S3Server {
     initiateStartup(log) {
         clientCheck(true, log, (err, results) => {
             if (err) {
-                log.info('initial health check failed, delaying startup', {
+                log.error('initial health check failed', {
                     error: err,
                     healthStatus: results,
                 });
-                setTimeout(() => this.initiateStartup(log), 2000);
             } else {
                 log.debug('initial health check succeeded');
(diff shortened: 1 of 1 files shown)


LaureVergeron:
It could be worthwhile to add a section either [here](https://github.com/scality/S3/blob/master/docs/DOCKER.rst), or [here](https://github.com/scality/S3/blob/master/docs/USING_PUBLIC_CLOUDS.rst#using-public-clouds-as-data-backends) mentioning that, if you can't access remote backends it might be helpful to look for a failing healthcheck.

rahulreddy:
We added a few comments on the Jira issue S3C-1412

DavidPineauScality:
Closing in favor of #1215, according to our discussion of S3C-1412

PR #1215, opened 2018-04-17
S3C-1412: Interpret healthCheck Status from Bucketd

DavidPineauScality:
## Description

As discussed in S3C-1412, it is necessary for S3 to interpret Bucketd health
status in order to provide more flexibility (relatively to the health check
mechanism) than failing due to a light partial unavailability of the platform.

This is done in the bucketclient backend's healthcheck method, in order to
comply with all the other backends.

Fixes S3C-1412

Note:
I still need to check whether tests exist for the modified piece of code (Doubtful given that it relies on an external service, which isn't supposed to be used in S3 unit/func testing), and update them if any is present.

### DIFF
--- a/lib/metadata/bucketclient/backend.js
+++ b/lib/metadata/bucketclient/backend.js
@@ -179,6 +179,9 @@ class BucketClientInterface {
                         body: reason,   // Provide analysis data
                     };
                     if (failure) {
+                        // Setting the `error` field is how the healthCheck
+                        // logic interprets it as an error. Don't forget it !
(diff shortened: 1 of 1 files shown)


DavidPineauScality:
@rahulreddy @RachedBenMustapha Pushed two fixes-commits once I got to testing it in a real platform, and playing with the available raft Sessions.

I'll leave it to you to squash them into the original commit when you've reviewed the code, and if you confirm that there is no major issue with this method.

rahulreddy:
@ironman-machine r+

PR #1225, opened 2018-04-20
rf: S3C-1399 Use CRR metrics from Arsenal

bennettbuchanan:


lib/utilities/reportHandler.js
```diff
@@ -76,19 +76,38 @@ function getSystemStats() {
...
+        internalStart: Date.now() - 900000, // 15 minutes ago.
+    }, log);
+    const redisKeys = {
+        ops: 'bb:crr:ops',
+        bytes: 'bb:crr:bytes',
+        opsDone: 'bb:crr:opsdone',
+        bytesDone: 'bb:crr:bytesdone',
+        failedCRR: 'bb:crr:failed',
+    };
+    const routes = backbeat.routes(redisKeys, sites);
```
jonathan-gramain: As a future cleanup, I'm thinking the `redisKeys` and `sites` filtering should not be processed by the routes selection function, but by backend methods.

bennettbuchanan: I didn't like this option much either, but was felt that avoiding it would lead to a significantly bigger refactor. I agree this could be another round of work.

lib/utilities/reportHandler.js
```diff
@@ -98,19 +116,19 @@ function getCRRStats(log, cb) {
         const stats = {
             completions: {
                 count: completions.results.count,
-                size: parseFloat(completions.results.size) * 1000,
+                size: Math.round(parseFloat(completions.results.size) * 1000),
```
bennettbuchanan: Added `Math.round` because multiplying by 1000 for certain numbers still results in a floating point (e.g. `parseFloat('8.19') * 1000`).

PR #1228, opened 2018-04-24
MongoDB as metadata and Scality as a data ba…

ssalaues:
Allows for the case where both MongoDB and Scality backend is enabled
and therefore would not require dataserver to be started.

The only way I could find to identify a scality backend with mongodb was via the auth. If there is a better way I will gladly update the PR.


rahulreddy:
Merged as the CI is in shambles.

PR #1229, opened 2018-04-24
bf: ZENKO-275 CRR to GCP MPU properties

bennettbuchanan:
Properties (e.g. ContentType) on the source MPU do not end up as values of the final object when doing CRR to GCP.


bennettbuchanan:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=true
try

bennettbuchanan:
E2E test passed, validating this change http://ci.ironmann.io/gh/scality/Integration/21638.

PR #1231, opened 2018-04-26
Ft/zenko 276 replication default read

alexanderchan-scality:
Review: https://github.com/scality/S3/pull/1231/commits/57a1bac41a707b22e7e71f46ae1e08c629a5df65

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -78,12 +78,26 @@ function sproxydAssert(configSproxyd) {
 function restEndpointsAssert(restEndpoints, locationConstraints) {
     assert(typeof restEndpoints === 'object',
         'bad config: restEndpoints must be an object of endpoints');
-    assert(Object.keys(restEndpoints).every(
-        r => typeof restEndpoints[r] === 'string'),
-        'bad config: each endpoint must be a string');
-    assert(Object.keys(restEndpoints).every(
-        r => typeof locationConstraints[restEndpoints[r]] === 'object'),
-        'bad config: rest endpoint target not in locationConstraints');
+    Object.keys(restEndpoints).forEach(r => {
+        const type = typeof restEndpoints[r];
+        assert(type === 'string' || type === 'object',
+            'bad config: each endpoint must be a string or an object');
+        if (type === 'string') {
+            assert(typeof locationConstraints[restEndpoints[r]] === 'object',
+            'bad config: rest endpoint target not in locationConstraints');
+        }
+        if (type === 'object') {
+            const { read, write } = restEndpoints[r];
+            assert(read && typeof read === 'string',
+                'bad config: invalid preferred read location');
+            assert(typeof locationConstraints[read] === 'object',
+            'bad config: preferred read target not in locationConstraints');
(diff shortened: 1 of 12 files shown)


lib/api/apiUtils/bucket/validateReplicationConfig.js
```diff
@@ -0,0 +1,25 @@
...
+ * @return {boolean} validity of replication configuration with rest endpoint
+ * configuration
+ */
+function validateConfiguration(config, bucket) {
+    const writeLocation = bucket.getLocationConstraint();
+    const readLocation = bucket.getReadLocationConstraint();
+    if (!config || !config.rules) {
+        return false;
+    }
+    const isValid = config.rules.some(rule => {
```
bennettbuchanan: Don't we want to check that every rule includes the storage class?

lib/api/bucketPut.js
```diff
@@ -18,6 +18,32 @@ config.on('location-constraints-update', () => {
...
+function _checkPreferredLocations(location, log) {
+    const retError = loc => {
+        const errMsg = 'value of the location you are attempting to set - ' +
+        `${loc} - is not listed in the locationConstraint config`;
+        log.trace(`locationConstraint is invalid - ${errMsg}`,
+          { locationConstraint: loc });
+        return { error: errors.InvalidLocationConstraint.
+          customizeDescription(errMsg) };
+    };
+    if (typeof location === 'string') {
```
bennettbuchanan: Seems like this could be reduced to a single condition.

alexanderchan-scality:
Added into https://github.com/scality/S3/pull/1227 as they are the same topic.

lib/api/apiUtils/bucket/validateReplicationConfig.js
alexanderchan-scality: My thought is that a valid configuration will contain the preferred read location as a storage class in one of the rules.

lib/Config.js
```diff
@@ -78,12 +78,26 @@ function sproxydAssert(configSproxyd) {
...
+        assert(type === 'string' || type === 'object',
+            'bad config: each endpoint must be a string or an object');
+        if (type === 'string') {
+            assert(typeof locationConstraints[restEndpoints[r]] === 'object',
+            'bad config: rest endpoint target not in locationConstraints');
+        }
+        if (type === 'object') {
+            const { read, write } = restEndpoints[r];
+            assert(read && typeof read === 'string',
+                'bad config: invalid preferred read location');
```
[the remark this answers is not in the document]
alexanderchan-scality: will add the explicit error message

lib/api/apiUtils/bucket/validateReplicationConfig.js
alexanderchan-scality: I can see that there might be an error if the replication rule has a prefix.

bennettbuchanan: Great! That's the kind of case I had in mind.

PR #1242, opened 2018-05-09
ft: ZENKO-319 adds sproxyd location type management

alexanderchan-scality:


lib/management/configuration.js
```diff
@@ -171,6 +171,20 @@ function patchConfiguration(instanceId, newConf, log, cb) {
                         };
                     }
                     break;
+                case 'location-scality-sproxyd-v1':
+                    location.type = 'scality';
+                    if (l.details && l.details.bootstrap && l.detaisl.path) {
```
jonathan-gramain: typo, should be `l.details.path`

lib/management/configuration.js
```diff
@@ -171,6 +171,21 @@ function patchConfiguration(instanceId, newConf, log, cb) {
...
                     }
                     break;
+                case 'location-scality-sproxyd-v1':
+                    location.type = 'scality';
+                    if (l.details && l.details.bootstrapList &&
+                        l.details.proxyPath) {
+                        locations.details = {
+                            connector: {
+                                sproxyd: {
+                                    chordCos: l.details.chordCos || null,
```
jonathan-gramain: We may want to add ARC support here as well, or not (@RachedBenMustapha ?)

rachedbenmustapha: @jonathan-gramain you're the sproxyd guru so correct me if I'm wrong :) We rely on already-deployed sproxyd and I believe ARC settings will be bound to the `proxyPath` we are clients of.

jonathan-gramain: Right, since we're using the `proxyPath` to an existing sproxyd that's what matters. But then is the chord COS used at all, aren't we passing the object bucket/key in the path directly?

rachedbenmustapha: @jonathan-gramain chord cos is passed to sproxydclient

rahulreddy:
@alexanderchan-scality Can you make this BertE compatible?

PR #1244, opened 2018-05-11
fix: use existing localCache redis configuration

RachedBenMustapha:
A new “redis” configuration item that can not be set easily in containers while one already exists called localCache which is already configurable


RachedBenMustapha:
Tests are currently failing because this fix uncovered bug ZENKO-346 which scality/Arsenal#490 fixes - will retry CI when scality/Arsenal#490 is merged

PR #1249, opened 2018-05-14
fx: running dataserver with mongodb metadata backend

nicolas2bert:


dataserver.js
```diff
@@ -8,7 +8,7 @@ if (config.backends.data === 'file' ||
     (config.backends.data === 'multiple' &&
      config.backends.metadata !== 'scality') &&
      (config.backends.auth !== 'scality' &&
-      config.backends.metadata !== 'mongodb')) {
```
nicolas2bert, line 11: @ssalaues is there any reason why we did check for `config.backends.metadata !== 'mongodb'`

ssalaues: The ```(config.backends.auth !== 'scality' && config.backends.metadata !== 'mongodb')``` condition allows for the case where both MongoDB metadata and Scality data backend is enabled and therefore would not require dataserver to be started. 

The issue that caused this addition:
S3 in federation with MongoDB as metadata will fail because there is no Dataserver parameters in the federation setup and shouldn't need Dataserver to be started anyway because of the "Scality" databackend aka RING.


Please be sure to test this on a federation deployment before merging as it may cause issues with federation installs.

ssalaues:
What issue does this fix?

dataserver.js
nicolas2bert: S3 in federation has config.backends.auth set to 'scality', no?
The issue, I see, is that when we run S3 locally with this conf:
`{ auth: 'mem',
  data: 'multiple',
  metadata: 'mongodb',
  kms: 'file' }` (`"start_mongo": "S3METADATA=mongodb npm-run-all --parallel start_dataserver start_s3server",`) even though we do `start_dataserver`, data server never gets started because of this check.

And we do need a data server to send report: https://github.com/scality/S3/blob/master/lib/utilities/reportHandler.js#L138

ssalaues: Ahh I see! Yes, S3 in federation does have config.backends.auth set to 'scality'. However the problem still remains that if we merge this, it very well may break the zenko federation deployment that has 'config.backends.metadata=mongodb'. Lets run a CI 'try' with ultron.

nicolas2bert:
@ironman-machine try

ssalaues:
@ironman-machine 
DEFAULT_BRANCH=z/1.0
try

ssalaues:
@ironman-machine 
try

PR #1250, opened 2018-05-15
bf: ZENKO-355 crr stats byte conversion

philipyoo:
Relies on https://github.com/scality/Arsenal/pull/491


rahulreddy:
CI is facing some issues which needs to be fixed outside of this PR. Tests are passing locally.

PR #1252, opened 2018-05-16
ft: make redis configurable through env variables

giacomoguiulfo:
This allows setting the Redis (sentinel) host and port through environmental variables.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -94,13 +94,22 @@ if [[ "$MONGODB_DATABASE" ]]; then
    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .mongodb.database=\"$MONGODB_DATABASE\""
 fi
 
-if [[ "$REDIS_HOST" ]]; then
-    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .localCache.host=\"$REDIS_HOST\""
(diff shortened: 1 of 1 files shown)


giacomoguiulfo:
@RachedBenMustapha  Correct me if I'm wrong but as far as I know, S3 connects to both redis-local (localCache) and redis-sentinel. Right now the docker-entrypoint.sh just lets you connect to only one by specifying REDIS_HOST.

RachedBenMustapha:
This change is not backward-compatible, we don't rename existing parameters to new uses but instead introduce new ones for new uses.

What part of s3 exactly needs access to sentinel?

giacomoguiulfo:
@RachedBenMustapha In Federation, when UTAPI is enabled it can use redis-sentinel (https://github.com/scality/Federation/blob/master/roles/run-s3/templates/config.json.j2#L52) or at least that is what it seems to be. I made a mistake, it should be ".utapi.redis" instead of ".redis". 

The reason I created this PR because of the addition of the redis-sentinel chart (https://github.com/scality/Zenko/pull/102). But it appears that UTAPI will be a separate chart so perhaps this is outside of the scope for now. Let me know what you think and this PR can be closed if you wish.

RachedBenMustapha:
I'd be for closing the PR since the need wasn't clearly defined and the proposed changes are not in line.

PR #1254, opened 2018-05-24
ft: ZENKO-404 lifecycle service account

jonathan-gramain:
Add support for service account 'service-lifecycle' in addition to
existing 'service-replication'. ACL checks are shunted in both cases,
we may implement more fine-grained access control later.


rahulreddy:
Merging this for now as we seem to be having timeout/service issues with Azure Storage.

PR #1255, opened 2018-05-24
ft: ZENKO-413 add GCP support for domain-styled endpoint requests

alexanderchan-scality:


alexanderchan-scality:
closed in favor of #1273

PR #1260, opened 2018-05-30
bf: speed up docker build

jonathan-gramain:
Maximize use of docker cache when only code files have changed

See same change on backbeat for reference (https://github.com/scality/backbeat/pull/239)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,17 +3,20 @@ MAINTAINER Giorgio Regni <gr@scality.com>
 
 WORKDIR /usr/src/app
 
-# Keep the .git directory in order to properly report version
-COPY . /usr/src/app
-
 RUN apt-get update \
-    && apt-get install -y jq python git build-essential --no-install-recommends \
+    && apt-get install -y jq --no-install-recommends
+
+COPY package.json /usr/src/app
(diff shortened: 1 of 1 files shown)


rahulreddy (dismissed):
Nice!

rachedbenmustapha (dismissed):

Dockerfile
```diff
@@ -3,18 +3,20 @@ MAINTAINER Giorgio Regni <gr@scality.com>
...
-COPY . /usr/src/app
-
 RUN apt-get update \
-    && apt-get install -y jq python git build-essential --no-install-recommends \
-    && npm install --production \
+    && apt-get install -y jq python git build-essential --no-install-recommends
+
+COPY package.json /usr/src/app
+RUN npm install --production \
     && apt-get autoremove --purge -y python git build-essential \
```
NicolasT: If you're removing `python`, `git` and others again, better don't install them in a prior step, because it makes the image (layers) unnecessarily bigger.

```
RUN install-runtime-dependencies
COPY package.json /path
RUN install-build-time-dependencies && npm install --production && uninstall-build-time-dependencies
COPY . /path
```

PR #1261, opened 2018-05-31
Ft/zenko 386 utapi service accounts

dora-korpar:
JIRA ticket: https://scality.atlassian.net/browse/ZENKO-386

Creates a route for CloudServer to request metrics from Utapi.
Use with https://github.com/scality/utapi/pull/175

### DIFF
--- a/lib/routes/routeUtapi.js
+++ b/lib/routes/routeUtapi.js
@@ -4,16 +4,11 @@ const { errors } = require('arsenal');
 const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
 const { config } = require('../Config');
 
-function _decodeURI(uri) {
-    // do the same decoding as in S3 server
-    return decodeURIComponent(uri.replace(/\+/g, ' '));
-}
-
 function _createUtapiRequestOptions(req) {
     const parsedUrl = url.parse(req.url, true);
-    const reqPath = _decodeURI(parsedUrl.pathname);
     // path will begin with /_/utapi
-    const utapiPath = `${reqPath.substring(8)}${parsedUrl.search}`;
+    const basePath = `${parsedUrl.pathname.substring(8)}`;
+    const utapiPath = parsedUrl.search ? `${basePath}${parsedUrl.search}` : `${basePath}`;
     const utapiPort = config.utapi.port || 8100;
     const utapiHost = config.utapi.host || 'localhost';
 
--- a/lib/routes/routeUtapi.js
+++ b/lib/routes/routeUtapi.js
@@ -1,0 +1,76 @@
+const url = require('url');
+const http = require('http');
+const { errors } = require('arsenal');
+const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
+const { config } = require('../Config');
+
+function _decodeURI(uri) {
+    // do the same decoding as in S3 server
+    return decodeURIComponent(uri.replace(/\+/g, ' '));
+}
+
+function _createUtapiRequestOptions(req) {
+    const parsedUrl = url.parse(req.url, true);
+    const reqPath = _decodeURI(parsedUrl.pathname);
+    // path will begin with /_/utapi
+    const utapiPath = `${reqPath.substring(8)}${parsedUrl.search}`;
+    const utapiPort = config.utapi.port || 8100;
+    const utapiHost = config.utapi.host || 'localhost';
+
+    return {
+        path: utapiPath,
+        port: utapiPort,
+        host: utapiHost,
+        method: 'POST',
+        headers: {
+            'content-type': 'application/json',
+            'cache-control': 'no-cache',
+        },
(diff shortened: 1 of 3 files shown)


lib/routes/routeUtapi.js
```diff
@@ -0,0 +1,76 @@
...
+    response.writeHead(200, httpHeaders);
+    response.end(body, 'utf8');
+}
+
+
+function routeUtapi(clientIP, request, response, log) {
+    log.debug('route request', { method: 'routeUtapi' });
+    let reqBody = '';
+    request.on('data', chunk => {
+        reqBody += chunk.toString();
```
bennettbuchanan, line 46: It's been my understanding that pushing to an array and then calling `.join` (as you've done below) has better performance than concatenating in this manner. Maybe that's inaccurate though—I actually couldn't find a definite source for that idea 😺.

philipyoo (dismissed):

lib/routes/routeUtapi.js
```diff
@@ -0,0 +1,76 @@
+const url = require('url');
+const http = require('http');
+const { errors } = require('arsenal');
+const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
+const { config } = require('../Config');
+
+function _decodeURI(uri) {
```
bennettbuchanan: Since this method is used in routeBackbeat.js, maybe we could go ahead and move to a common util.

bennettbuchanan (dismissed):
Awesome! LGTM, I've left some considerations. :shipit:

lib/routes/routeUtapi.js
dora-korpar: In this case, I tried the `.join` way of creating the `reqBody`, but when it was written to the new request, the format was messed up. This was the only way I was able to keep the JSON properly formatted on the passthrough to Utapi :)

lib/routes/routeUtapi.js
dora-korpar: I agree in general, but I left it in this case since it's just a 3-line function. Looked a little silly in a file all its own x)

philipyoo (dismissed):

JianqinWang (dismissed):

lib/routes/routeUtapi.js
JianqinWang: I'm writing more routes that will also be using dcodeURI - this can be a future minor rf

rahulreddy: IIRC this was done because s3cmd replaces spaces with `+`. I don't think we need it for Utapi/Backbeat/Metadata routes.

lib/routes/routeUtapi.js
```diff
@@ -0,0 +1,76 @@
...
+    // do the same decoding as in S3 server
+    return decodeURIComponent(uri.replace(/\+/g, ' '));
+}
+
+function _createUtapiRequestOptions(req) {
+    const parsedUrl = url.parse(req.url, true);
+    const reqPath = _decodeURI(parsedUrl.pathname);
+    // path will begin with /_/utapi
+    const utapiPath = `${reqPath.substring(8)}${parsedUrl.search}`;
```
rahulreddy: I recommend sanity checking the variables before accessing the properties and sanity checking the utapiPath as well.

lib/routes/routeUtapi.js
```diff
@@ -0,0 +1,76 @@
...
+    const utapiPort = config.utapi.port || 8100;
```
rahulreddy: Same thing here, please check if `config.utapi` is defined before accessing the port/host property.

lib/routes/routeUtapi.js
```diff
@@ -0,0 +1,76 @@
...
+                    return _respond(response, responseBody, log);
+                }
+                log.error('Utapi request failed', {
+                    statusCode: res.statusCode,
+                    body: responseBody });
+                return responseJSONBody(errors.InternalError, null,
+                    response, log);
+            });
+        });
+        utapiRequest.write(reqBody);
```
rahulreddy, line 66: Can't we just use pipe instead?
`utapiRequest.pipe(request)`?

lib/routes/routeUtapi.js
```diff
@@ -0,0 +1,76 @@
...
+    let reqBody = '';
+    request.on('data', chunk => {
+        reqBody += chunk.toString();
+    });
+    request.on('end', () => {
+        const options = _createUtapiRequestOptions(request);
+        const utapiRequest = http.request(options, res => {
+            const resBody = [];
+            res.setEncoding('utf8');
+            res.on('data', chunk => resBody.push(chunk));
```
rahulreddy, line 53: I recommend just using streams here instead of using a data listener.

rahulreddy (changes requested):

rahulreddy:
Closing. I will re-open once we start porting Utapi.

PR #1262, opened 2018-05-31
bf: S3C-1477 Add more verbose logging

tmacro:


lib/management/configuration.js
```diff
@@ -171,7 +171,7 @@ function patchConfiguration(newConf, log, cb) {
                 _config.setLocationConstraints(locations);
             } catch (error) {
                 log.info('could not apply configuration version location ' +
-                    'constraints', { error });
+                    'constraints', { msg: error.message, trace: error.stack });
```
jonathan-gramain: Two comments:

* We mostly use `error` attribute to log the error message instead of `msg`

* I don't recommend logging the stack trace except for `log.trace` kind of logs, in my experience they bloat the logs and make them less readable while usually not being this useful to track down the issues (one may disagree of course)

jonathan-gramain (approved):

lib/management/configuration.js
```diff
@@ -171,7 +171,7 @@ function patchConfiguration(newConf, log, cb) {
                 _config.setLocationConstraints(locations);
             } catch (error) {
                 log.info('could not apply configuration version location ' +
-                    'constraints', { error });
+                    'constraints', { error: error.message });
```
rachedbenmustapha, line 174: There may be something wrong with the exception thrown by `_config.setLocationConstraints()` so I would look at that instead.

We usually log actual error objects as the `error` attribute so the code is fine as it is. The `error` attribute of log lines actually has to be an object and not a string (we use the `errorMessage` attribute for error messages) otherwise any log indexer will reject the line because the implicit schema and the line don't match.

rachedbenmustapha (changes requested):

lib/management/configuration.js
tmacro: As this logging statement is inside a try-catch, you must explicitly build the logging object.

PR #1268, opened 2018-06-05
ft: add "no_proxy" environment variables

giacope:
This adds the no_proxy and NO_PROXY environmental variables, which should always have these values.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,9 @@ RUN apt-get update \
 
 VOLUME ["/usr/src/app/localData","/usr/src/app/localMetadata"]
 
+ENV NO_PROXY localhost,127.0.0.1
+ENV no_proxy localhost,127.0.0.1
+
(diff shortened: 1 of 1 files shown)


JianqinWang (approved):

dora-korpar (approved):

Dockerfile
```diff
@@ -17,6 +17,9 @@ RUN apt-get update \
 
 VOLUME ["/usr/src/app/localData","/usr/src/app/localMetadata"]
 
+ENV NO_PROXY localhost,127.0.0.1
+ENV no_proxy localhost,127.0.0.1
```
nicolas2bert, line 21: Just wondering, why do you need upper and lower case? :)

giacope: It's a convention. In fact, there is a third one ```no_PROXY```, but those 2 are the most common.

PR #1278, opened 2018-06-12
Feature/zenko 485 eve on s3

ssalaues:
Adds basic Eve functionality using both kube and docker workers that can be the template for future additions.

2nd commit modifies the Dockerfile slightly to improve CI and local build time

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Giorgio Regni <gr@scality.com>
 WORKDIR /usr/src/app
 
 # Keep the .git directory in order to properly report version
-COPY . /usr/src/app
+COPY ./package.json .
 
 RUN apt-get update \
     && apt-get install -y jq python git build-essential --no-install-recommends \
@@ -14,6 +14,8 @@ RUN apt-get update \
     && npm cache clear \
     && rm -rf ~/.node-gyp \
     && rm -rf /tmp/npm-*
+
+COPY ./ ./
 
 VOLUME ["/usr/src/app/localData","/usr/src/app/localMetadata"]
 
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -1,0 +1,102 @@
+---
+version: 0.2
+
+branches:
+  feature/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*:
+    stage: pre-merge
+  development/*:
+    stage: post-merge
+
+models:
+  - Git: &clone
+      name: Pull repo
+      repourl: '%(prop:git_reference)s'
+      shallow: True
+      retryFetch: True
+      haltOnFailure: True
+      # Machine Setup
+  - ShellCommandWthSecrets: &credentials
+      name: Setup Credentials
+      command: bash eve/workers/build/credentials.bash
+      haltOnFailure: True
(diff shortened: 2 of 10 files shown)


tmacro (dismissed):
LGTM

eve/main.yml
```diff
@@ -0,0 +1,106 @@
...
+
+  linting-coverage:
+    worker: 
+      type: docker
+      path: eve/workers/build
+      volumes: &default_volumes
+        - '/home/eve/workspace'
+    steps:
+      - Git: *clone
+      - ShellCommand: *npm-install
```
rahulreddy, line 52: Shall we use yarn? It's faster.

Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Giorgio Regni <gr@scality.com>
 WORKDIR /usr/src/app
 
 # Keep the .git directory in order to properly report version
-COPY . /usr/src/app
+COPY ./package.json .
```
alexanderchan-scality, line 7: Looks good. just curious as to why only the `package.json` is copied here rather than the whole dir

ssalaues:
Fixed commit messages

ssalaues:
Fixed post merge step

ssalaues:
example of the post-merge docker build step
https://eve.devsca.com/github/scality/s3/#/builders/3/builds/2566

eve/workers/build/Dockerfile
```diff
@@ -0,0 +1,52 @@
+FROM ubuntu:xenial
+
+#
+# Install packages needed by the buildchain
+#
+ENV LANG C.UTF-8
+COPY ./s3_packages.list ./buildbot_worker_packages.list /tmp/
+RUN apt-get update \
+    && cat /tmp/*packages.list | xargs apt-get install -y \
+    && wget https://nodejs.org/dist/v6.9.5/node-v6.9.5-linux-x64.tar.xz \
```
rahulreddy: Can we try using `n` to manage node.js distribution? This way we can automatically install `n 6.x` and get the latest minor-minor version.
https://github.com/tj/n

rahulreddy (dismissed):
Quick review. LGTM in general.

rahulreddy:
@bert-e wait

eve/main.yml
ssalaues: I'll try it out. Should work

eve/workers/build/Dockerfile
ssalaues: Yeah that's a good idea. I'll add that in.

alexanderchan-scality (dismissed):

Dockerfile
ssalaues: So this is where the magic of docker layers becomes useful. So say you are changing code and need to rebuild your docker image to test but you dependencies stay the same. If this was built previously, then on consecutive runs Docker will skip this step along with the following RUN step where all the system packages and npm packages are installed and use the cached layers from the previous builds.

It then will use the step at line 18 to copy the remaining code base into the image which wont be cached because of code changes.

jonathan-gramain (dismissed):

rahulreddy (approved):

rahulreddy:
Merging manually due to a know Azure issue in CI.

PR #1285, opened 2018-06-18
improvement: remove enterprise circle ci

rahulreddy:


philipyoo (approved):
looks like enterprise_ci doesn't like this change lol

ssalaues (approved):

PR #1287, opened 2018-06-19
ft: ZENKO-143 batchDelete backbeat route

jonathan-gramain:
- Implement 'POST /_/backbeat/batchdelete' backbeat route to get rid
  of an array of data locations. The route will be used by the garbage
  collector service.

  * This includes some reorganization of backbeat routes sanity checks

- Handle gracefully 404 errors from datastore backend:

  * no need to retry 404 errors as they are permanent

  * batch delete can also gracefully handle missing data locations and
    still delete other existing locations (may happen if a retry
    occurs, when replaying a kafka message among other cases).

- Support service-gc service account with account owner access rights.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -5,6 +5,7 @@ const { auth, errors, s3middleware } = require('arsenal');
 const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
 const { getSubPartIds } = s3middleware.azureHelper.mpuUtils;
 const vault = require('../auth/vault');
+const data = require('../data/wrapper');
 const metadata = require('../metadata/wrapper');
 const locationConstraintCheck = require(
     '../api/apiUtils/object/locationConstraintCheck');
@@ -34,7 +35,7 @@ function _decodeURI(uri) {
     return decodeURIComponent(uri.replace(/\+/g, ' '));
 }
 
-function normalizeBackbeatRequest(req) {
+function _normalizeBackbeatRequest(req) {
     /* eslint-disable no-param-reassign */
     const parsedUrl = url.parse(req.url, true);
     req.path = _decodeURI(parsedUrl.pathname);
@@ -44,6 +45,11 @@ function normalizeBackbeatRequest(req) {
     req.bucketName = pathArr[4];
     req.objectKey = pathArr.slice(5).join('/');
     /* eslint-enable no-param-reassign */
+}
+
+function _isObjectRequest(req) {
+    return ['data', 'metadata', 'multiplebackenddata']
+        .includes(req.resourceType);
 }
 
 function _respond(response, payload, log, callback) {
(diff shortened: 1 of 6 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -46,6 +47,11 @@ function normalizeBackbeatRequest(req) {
     /* eslint-enable no-param-reassign */
 }
 
+function _isObjectRequest(req) {
+    return ['data', 'metadata', 'multiplebackenddata']
```
bennettbuchanan, line 51: Does this need 'batchdelete' as well?

lib/routes/routeBackbeat.js
```diff
@@ -609,6 +622,50 @@ function deleteObjectTagging(request, response, log, callback) {
...
+                    locations,
+                    error: err,
+                });
+                return callback(err);
+            }
+            log.debug('batch delete successful', { locations });
+            return _respond(response, null, log, callback);
+        });
+    });
+    return undefined;
```
bennettbuchanan: Minor: I guess this is meant to satisfy the linter, in which case maybe just return `_getRequestPayload`.

tests/functional/raw-node/test/routes/routeBackbeat.js
```diff
@@ -452,4 +452,103 @@ describeSkipIfAWS('backbeat routes:', () => {
...
+                        jsonResponse: true,
+                    };
+                    makeRequest(options, done);
+                },
+                done => s3.getObject({
+                    Bucket: TEST_BUCKET,
+                    Key: 'batch-delete-test-key',
+                }, err => {
+                    // should error out as location shall no longer exists
+                    assert(err);
```
bennettbuchanan: Maybe worthwhile to assert the error code as well (i.e. 'NoSuchKey');

bennettbuchanan (dismissed):

lib/routes/routeBackbeat.js
```diff
@@ -637,21 +695,41 @@ const backbeatRoutes = {
...
+          (!request.resourceType ||
+           (_isObjectRequest(request) &&
+            (!request.bucketName || !request.objectKey)) ||
+           (!request.query.operation && useMultipleBackend));
+    const invalidRoute =
+          (backbeatRoutes[request.method] === undefined ||
+           backbeatRoutes[request.method][request.resourceType] === undefined ||
+           (backbeatRoutes[request.method][request.resourceType]
+            [request.query.operation] === undefined &&
+            useMultipleBackend));
```
philipyoo, line 713: I noticed this behavior in another PR, but in javascript, if the first statement is true, using the `&&` will evaluate and return the value of the second. So in this case, it will always return the value of `useMultipleBackend` if the operation query is undefined.
Not sure if this is expected behavior, just making sure

lib/routes/routeBackbeat.js
jonathan-gramain: No, as `batchdelete` is not an object request, it's a batch request and that's the point of this check as it does not expect a bucket/object name in the URL.

lib/routes/routeBackbeat.js
jonathan-gramain: Yes, here we only rely on the boolean value of the expressions being evaluated (`useMultipleBackend` is a boolean)

philipyoo: lol sorry I got confused here

lib/routes/routeBackbeat.js
jonathan-gramain: done

bennettbuchanan (dismissed):

philipyoo (dismissed):

philipyoo (dismissed):

tests/functional/raw-node/test/routes/routeBackbeat.js
```diff
@@ -452,4 +452,103 @@ describeSkipIfAWS('backbeat routes:', () => {
...
+                        jsonResponse: true,
+                    };
+                    makeRequest(options, done);
+                },
+                done => s3.getObject({
+                    Bucket: TEST_BUCKET,
+                    Key: 'batch-delete-test-key',
+                }, err => {
+                    // should error out as location shall no longer exist
+                    assert(err);
```
jonathan-gramain, line 511: @bennettbuchanan I reverted the check of `NoSuchKey` because what's missing is the data location, not the object itself, so it ends up erroring with a server error (503), and I don't think we need to be specific about this error.

bennettbuchanan (approved):

philipyoo (approved):

PR #1288, opened 2018-06-19
ft: add cloudserver support for advanced options

alexanderchan-scality:


lib/management/configuration.js
```diff
@@ -179,6 +179,8 @@ function patchConfiguration(newConf, log, cb) {
                         l.locationType });
                     return;
                 }
+                location.sizeLimitGB = l.sizeLimitGB || false;
```
rachedbenmustapha: I'd expect `0 GB` to mean 'no limit' instead of `false GB` :)

jonathan-gramain: I suggest leaving as `undefined` instead of `false` to avoid mixing data types (considered bad practice).

rachedbenmustapha (changes requested):

lib/management/configuration.js
alexanderchan-scality: will fix

alexanderchan-scality: will fix to use undefined for 0 value case

lib/management/configuration.js
```diff
@@ -179,6 +179,8 @@ function patchConfiguration(newConf, log, cb) {
                         l.locationType });
                     return;
                 }
+                location.sizeLimitGB = l.sizeLimitGB || undefined;
```
rachedbenmustapha: To ease debugging then, let's use `null` which conveys that it was manually assigned but chosen as missing-value, whereas `undefined` usually means that it was never manually defined.

rachedbenmustapha (changes requested):

rachedbenmustapha (approved):

jonathan-gramain (approved):

PR #1294, opened 2018-06-20
feature: ZENKO-315 Allow CRR without versioning

bennettbuchanan:


lib/routes/routeBackbeat.js
```diff
@@ -706,16 +706,6 @@ function routeBackbeat(clientIP, request, response, log) {
                 return backbeatRoutes[request.method][request.resourceType]
                     [request.query.operation](request, response, log, next);
             }
-            const versioningConfig = bucketInfo.getVersioningConfiguration();
```
bennettbuchanan: Allow the user to be responsible for setting versioning on the destination bucket, since at this point we don't know whether the source is an NFS bucket.

lib/metadata/ModelVersion.md
```diff
@@ -105,3 +105,15 @@ this._readLocationConstraint = readLocationConstraint || null;
...
+
+### Properties Added
+
+```javascript
+this._isNFS = isNFS || null;
+```
+
+### Usage
+
+Used to determine whether the bucket is on NFS
```
jonathan-gramain: `whether the bucket may be accessed through NFS` could be more accurate?

bennettbuchanan: 👍 Updated.

jonathan-gramain (dismissed):

jonathan-gramain (approved):

philipyoo (approved):

lib/api/objectDelete.js
```diff
@@ -121,6 +121,28 @@ function objectDelete(authInfo, request, log, cb) {
                 newDeleteMarker: false,
             };
             if (delOptions && delOptions.deleteData) {
+                if (bucketMD.isNFS() &&
```
rahulreddy: So this can be combined in the main if conditional without having another level?

rahulreddy (changes requested):

lib/api/objectDelete.js
bennettbuchanan: There is a sub-condition for NFS, otherwise it treats the operation like a regular delete (as was being done before).

bennettbuchanan: It would be easy to bump it up an indentation level for readability 😄.

rahulreddy (approved):

PR #1304, opened 2018-06-21
bf: ZENKO-576 fix 'supportsVersioning' flag in locations

jonathan-gramain:
Orbit management layer now relies on
constants.versioningNotImplBackends to set the 'supportsVersioning'
flag on locations.

### DIFF
--- a/lib/management/configuration.js
+++ b/lib/management/configuration.js
@@ -91,9 +91,11 @@ function patchConfiguration(newConf, log, cb) {
                 switch (l.locationType) {
                 case 'location-mem-v1':
                     location.type = 'mem';
+                    supportsVersioning = true;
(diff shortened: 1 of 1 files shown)


rachedbenmustapha (changes requested):
`supportsVersioning` is not using `versioningNotImplBackends` on purpose because some backends like digital ocean are of type `aws_s3` and do not support versioning - so I think that extra layer is still useful.

jonathan-gramain:
@RachedBenMustapha indeed, sorry for rushing a little bit on this!

rachedbenmustapha (approved):
much better, thanks :)

philipyoo (approved):

PR #1309, opened 2018-06-21
Feature/zenko 267 md ingestion routes

JianqinWang:
Dependent on https://github.com/scality/bucketclient/pull/139

### DIFF
--- a/lib/routes/routeMetadata.js
+++ b/lib/routes/routeMetadata.js
@@ -1,0 +1,169 @@
+const url = require('url');
+const async = require('async');
+
+const { auth, errors } = require('arsenal');
+const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
+const vault = require('../auth/vault');
+
+const BucketClientInterface = require('../metadata/bucketclient/backend');
+const prepareRequestContexts = require(
+'../api/apiUtils/authorization/prepareRequestContexts');
+const { decodeVersionId } = require('../api/apiUtils/object/versioning');
+
+const metadata = new BucketClientInterface();
+auth.setHandler(vault);
+
+function normalizeMetadataRequest(req) {
+    /* eslint-disable no-param-reassign */
+    const parsedUrl = url.parse(req.url, true);
+    req.path = parsedUrl.pathname;
+    req.query = parsedUrl.query;
+    const pathArr = req.path.split('/');
+    req.resourceType = pathArr[3];
+    req.bucketName = pathArr[4];
+    if (pathArr[5]) {
+        req.objectKey = pathArr.slice(5).join('/');
+    }
+    /* eslint-enable no-param-reassign */
+}
+
+function _respond(response, payload, log, callback) {
+    const body = typeof payload === 'object' ?
+        JSON.stringify(payload) : payload;
+    const httpHeaders = {
+        'x-amz-id-2': log.getSerializedUids(),
+        'x-amz-request-id': log.getSerializedUids(),
+        'content-type': 'application/json',
+        'content-length': Buffer.byteLength(body),
+    };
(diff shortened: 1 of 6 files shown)


lib/routes/routeMetadata.js
```diff
@@ -0,0 +1,175 @@
...
+        'content-length': Buffer.byteLength(body),
+    };
+    response.writeHead(200, httpHeaders);
+    response.end(body, 'utf8', callback);
+}
+
+/*
+- getting list of buckets for each raft session -
+GET /_/metadata/listbuckets/<raft session id>
+- getting list of objects for each bucket -
```
jonathan-gramain, line 46: minor: comments are inverted between lines 47 and 49

lib/routes/routeMetadata.js
```diff
@@ -0,0 +1,175 @@
...
+        listobjects: getObjectList,
+        getbucket: getBucketMetadata,
+        getobject: getObjectMetadata,
+    },
+};
+
+function routeMetadata(clientIP, request, response, log) {
+    log.debug('routing request', { method: 'routeBackbeat' });
+    normalizeMetadataRequest(request);
+    const useMultipleBackend = request.resourceType === 'multiplebackenddata';
```
jonathan-gramain: This is specific to backbeat routes, please remove

lib/routes/routeMetadata.js
```diff
@@ -0,0 +1,175 @@
...
+        (!request.query.operation && useMultipleBackend));
+    if (invalidRequest) {
+        log.debug('invalid request', {
+            method: request.method, bucketName: request.bucketName,
+            objectKey: request.objectKey, resourceType: request.resourceType,
+            query: request.query,
+        });
+        return responseJSONBody(errors.MethodNotAllowed, null, response, log);
+    }
+    const requestContexts = prepareRequestContexts('objectReplicate', request);
```
jonathan-gramain, line 118: Feels weird to use `objectReplicate` as a context here, which is also technical debt for backbeat (as backbeat does more than just replicating an object). But there may be implications with Vault too, I don't remember exactly, so it's possible that it is quicker to stick with this context.

If you have time to find out, can you add a comment?

jonathan-gramain (changes requested):

lib/routes/routeMetadata.js
```diff
@@ -0,0 +1,170 @@
...
+        'content-type': 'application/json',
+        'content-length': Buffer.byteLength(body),
+    };
+    response.writeHead(200, httpHeaders);
+    response.end(body, 'utf8', callback);
+}
+
+/*
+- getting list of buckets for each raft session -
+GET /_/metadata/listbuckets/<raft session id>
```
jonathan-gramain, line 45: Just thinking it could be nice to be able to use these routes in the future as an internal mongodb frontend for other purposes (e.g. listbuckets may be used by lifecycle extension to avoid reading from the log and maintaining a list of buckets on zookeeper). Not that this is the point of implementing anything here but thinking if we could also use the routes in a similar way with a mongodb backend later that would be neat!

jonathan-gramain (approved):

tmacro (approved):

lib/routes/routeMetadata.js
```diff
@@ -0,0 +1,170 @@
+const url = require('url');
+const async = require('async');
+
+const { auth, errors } = require('arsenal');
+const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
+const vault = require('../auth/vault');
+
+// const metadata = require('../metadata/wrapper');
```
bennettbuchanan: Extra comment 😃

bennettbuchanan (approved):
Nice! 🚢

PR #1313, opened 2018-06-25
Feature/zenko 142 location quota

dora-korpar:
Add a storage quota size per location! Make sure you never put more data than you want!
APIs affected:
ObjectPut
ObjectDelete
ObjectCopy
ObjectPutPart
ObjectPutCopyPart
AbortMPU

The storage limit check is done at the wrapper level, so is found in `data.wrapper` put and delete, and in `multipleBackendGateway` as needed.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -238,6 +238,11 @@ function locationConstraintAssert(locationConstraints) {
             // eslint-disable-next-line no-param-reassign
             locationConstraints[l].details.supportsVersioning = true;
         }
+        if (details.sizeLimitGB !== undefined) {
+            assert(typeof details.sizeLimitGB === 'number',
+                'bad config: locationConstraints[region].details.sizeLimitGB ' +
+                'must be a number (in gigabytes)');
+        }
 
         if (locationConstraints[l].type === 'azure') {
             azureLocationConstraintAssert(l, locationConstraints[l]);
@@ -754,6 +759,11 @@ class Config extends EventEmitter {
                     'bad config: utapi.expireMetricsTTL must be a number');
                 this.utapi.expireMetricsTTL = config.utapi.expireMetricsTTL;
             }
+        }
+        if (Object.keys(this.locationConstraints).some(
+        loc => this.locationConstraints[loc].details.sizeLimitGB)) {
+            assert(this.utapi, 'bad config: if storage size limit set on a ' +
+                'location constraint, Utapi must also be configured');
         }
 
         this.log = { logLevel: 'debug', dumpLevel: 'error' };
--- a/lib/api/apiUtils/object/locationStorageCheck.js
+++ b/lib/api/apiUtils/object/locationStorageCheck.js
@@ -1,0 +1,49 @@
+const { errors } = require('arsenal');
+
+const { config } = require('../../../Config');
+const { getLocationMetric, pushLocationMetric } =
+    require('../../../utapi/utilities');
+
+function _gbToBytes(gb) {
+    return gb * 1024 * 1024 * 1024;
+}
+
+/**
+ * locationStorageCheck - will ensure there is enough space left for object on
+ * PUT operations, or will update metric on DELETE
+ * NOTE: storage limit may not be exactly enforced in the case of concurrent
+ * requests when near limit
+ * @param {string} location - name of location to check quota
+ * @param {number} updateSize - new size to check against quota in bytes
+ * @param {object} log - werelogs logger
+ * @param {function} cb - callback function
+ * @return {undefined}
+ */
+function locationStorageCheck(location, updateSize, log, cb) {
+    const lc = config.locationConstraints;
+    const sizeLimitGB = lc[location] && lc[location].details ?
+        lc[location].details.sizeLimitGB : undefined;
+    if (updateSize === 0 || sizeLimitGB === undefined) {
+        return cb();
+    }
+    // no need to list location metric, since it should be decreased
+    if (updateSize < 0) {
+        return pushLocationMetric(location, updateSize, log, cb);
+    }
+    return getLocationMetric(location, log, (err, bytesStored) => {
+        if (err) {
+            log.error(`Error listing metrics from Utapi: ${err.message}`);
+            return cb(err);
+        }
+        const newStorageSize = bytesStored + updateSize;
+        const sizeLimitBytes = _gbToBytes(sizeLimitGB);
+        if (sizeLimitBytes < newStorageSize) {
+            return cb(errors.AccessDenied.customizeDescription(
+                `The assigned storage space limit for location ${location} ` +
+                'will be exceeded'));
+        }
+        return pushLocationMetric(location, updateSize, log, cb);
+    });
+}
+
+module.exports = locationStorageCheck;
--- a/lib/api/multipartDelete.js
+++ b/lib/api/multipartDelete.js
@@ -106,6 +106,13 @@ function multipartDelete(authInfo, request, log, callback) {
             if (skipDataDelete) {
                 return next(null, mpuBucket, storedParts, destBucket);
             }
+            for (let i = 0; i < storedParts.length; i++) {
+                if (storedParts[i].value.partLocations.length > 0) {
+                    // eslint-disable-next-line no-param-reassign
+                    storedParts[i].value.partLocations[0].size =
+                        storedParts[i].value.Size;
+                }
+            }
             // The locations were sent to metadata as an array
             // under partLocations.  Pull the partLocations.
             let locations = storedParts.map(item => item.value.partLocations);
--- a/lib/data/wrapper.js
+++ b/lib/data/wrapper.js
@@ -14,6 +14,8 @@ const kms = require('../kms/wrapper');
 const externalBackends = require('../../constants').externalBackends;
 const constants = require('../../constants');
 const { BackendInfo } = require('../api/apiUtils/object/BackendInfo');
+const locationStorageCheck =
+    require('../api/apiUtils/object/locationStorageCheck');
 const RelayMD5Sum = require('../utilities/RelayMD5Sum');
 const skipError = new Error('skip');
 
@@ -50,15 +52,6 @@ if (config.backends.data === 'mem') {
     implName = 'cdmi';
 }
 
-/**
- * _retryDelete - Attempt to delete key again if it failed previously
- * @param { string | object } objectGetInfo - either string location of object
- *      to delete or object containing info of object to delete
- * @param {object} log - Werelogs request logger
- * @param {number} count - keeps count of number of times function has been run
- * @param {function} cb - callback
- * @returns undefined and calls callback
- */
 const MAX_RETRY = 2;
 
 // This check is done because on a put, complete mpu or copy request to
@@ -77,6 +70,15 @@ function _shouldSkipDelete(locations, requestMethod, newObjDataStoreName) {
     return (isSkipBackend && isMatchingBackends && isSkipMethod);
 }
 
+/**
+ * _retryDelete - Attempt to delete key again if it failed previously
+ * @param { string | object } objectGetInfo - either string location of object
+ *      to delete or object containing info of object to delete
+ * @param {object} log - Werelogs request logger
+ * @param {number} count - keeps count of number of times function has been run
+ * @param {function} cb - callback
+ * @return {undefined} calls callback
+ */
 function _retryDelete(objectGetInfo, log, count, cb) {
     if (count > MAX_RETRY) {
         return cb(errors.InternalError);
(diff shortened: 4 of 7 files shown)


lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -33,7 +34,9 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
                 data.batchDelete(dataToDelete, requestMethod,
                 newDataStoreName, deleteLog);
             }
-            return callback(null, result);
+            const returnObj = Object.assign(
+                { dataStoreName: dataGetInfo[0].dataStoreName }, result);
```
jonathan-gramain: We probably need to check`Array.isArray(dataGetInfo)` first to ensure backward compat, like done at line 32.

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param {string} location - name of location to check quota
+ * @param {object} log - werelogs logger
+ * @param {function} cb - callback function
+ * @return {undefined}
+ */
+function locationStorageCheck(newObjSize, prevObjSize, isDeleteMarker, location,
+log, cb) {
+    const sizeLimitGB = config.locationConstraints[location].
+        details.sizeLimitGB;
+    const sizeDiff = newObjSize - prevObjSize;
```
jonathan-gramain: This may work when `prevObjSize === null` but I would rather not do arithmetic with `null` and do an explicit check for this.

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -0,0 +1,54 @@
...
+        }
+        const newStorageSize = bytesStored + sizeDiff;
+        const sizeLimitBytes = _gbToBytes(sizeLimitGB);
+        if (sizeLimitBytes < newStorageSize) {
+            return cb(errors.AccessDenied.customizeDescription(
+                `The assigned storage space limit for location ${location} ` +
+                'will be exceeded'));
+        }
+        if ((newStorageSize / sizeLimitBytes) > 0.8) {
+            log.warn(`${location} location storage space is above 80% ` +
```
jonathan-gramain: I would prefer not logging for every check done (especially in warning level) since it may be a "normal" situation (at least temporarily) for some users to reach 80% storage, leading to a lot of log lines that users will likely not want, or that they will never read unless they actually see errors happening. One may argue that we could log at 99% instead which is unlikely to be "normal", however it's hard to define a clear line here and usefulness to log at 99% is dubious (in terms of time for action to take before reaching an error).

I think we can (should IMO) get rid of this log for now, or at the very least, add a global variable that tells if we are above or below that boundary and only log at the time we cross the boundary. But even then, if we keep putting and deleting objects around 80% of usage, we will log as much and it may be even more confusing. We could also have separate "high tide" and "low tide" values where to log respectively when crossing beyond and under those limits to prevent this, but it's becoming quite custom and not necessarily very useful, hence I think it's better to just stick with the hard check when reaching the storage limit until we agree on a proper feedback method.

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -0,0 +1,54 @@
...
+    const sizeDiff = newObjSize - prevObjSize;
+
+    if (isDeleteMarker || sizeDiff === 0 || sizeLimitGB === undefined) {
+        return cb();
+    }
+    // no need to list location metric, since it should be decreased
+    if (sizeDiff < 0) {
+        return pushLocationMetric(location, sizeDiff, log, cb);
+    }
+    return listLocationMetric(location, log, (err, bytesStored) => {
```
jonathan-gramain: Correct me if I'm wrong, "list" carries a notion of multiple values returned, here it seems to me we just fetch the location's metric value so should be called `getLocationMetric`. Since that impacts utapi client we can skip this change for now and keep it as a TODO.

lib/api/objectPutCopyPart.js
```diff
@@ -229,6 +230,23 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                             sourceVerId, sourceLocationConstraintName);
                     });
         },
+        function checkStorageQuote(dataLocator, destBucketMD,
```
jonathan-gramain: typo -> `checkStorageQuota`, but see comment above about renaming to `checkStorageLimit`

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -0,0 +1,54 @@
...
+const { config } = require('../../../Config');
+const { listLocationMetric, pushLocationMetric } =
+    require('../../../utapi/utilities');
+
+function _gbToBytes(gb) {
+    return gb * 1024 * 1024 * 1024;
+}
+
+/**
+ *
```
jonathan-gramain: May be worth mentioning in the JSDoc that it might not enforce the limit exactly in case there are concurrent requests coming while reaching the storage limit (atomicity issue between `listLocationMetric` and `pushLocationMetric`)

lib/api/objectCopy.js
```diff
@@ -317,12 +318,35 @@ function objectCopy(authInfo, request, sourceBucket,
                         sourceLocationConstraintName, backendInfoDest);
                 });
         },
+        function checkStorageQuota(storeMetadataParams, dataLocator,
```
jonathan-gramain: I would stick with `checkStorageLimit` name to avoid mixing things up with the future quota feature (as per account).

lib/api/multipartDelete.js
```diff
@@ -148,6 +150,18 @@ function multipartDelete(authInfo, request, log, callback) {
                         keys: [objectKey],
                         byteLength: partSizeSum,
                     });
+                    if (location) {
+                        // partSizeSum passed as negative to decrement metric
+                        return pushLocationMetric(location, -partSizeSum, log,
```
jonathan-gramain: Not sure, but isn't this already done at line 118 (`data.delete()` which in turns calls the backend `_retryDelete` which decrements the metric as well)?

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -0,0 +1,54 @@
...
+        if (sizeLimitBytes < newStorageSize) {
+            return cb(errors.AccessDenied.customizeDescription(
+                `The assigned storage space limit for location ${location} ` +
+                'will be exceeded'));
+        }
+        if ((newStorageSize / sizeLimitBytes) > 0.8) {
+            log.warn(`${location} location storage space is above 80% ` +
+                'capacity');
+        }
+        return pushLocationMetric(location, sizeDiff, log, cb);
```
jonathan-gramain: I'm wondering if the fact that we have the `pushLocationMetric` inside the data backend (`_retryDelete`) will not have the effect of carrying the decrement twice (here and there)?

Overall, I haven't dug how or if this is the right thing to do, but it seems to me the logical answer is to increment the location metric in the data backend itself (`data.put`), so that it's in par with the decrement in `data.delete`, it's much easier to reason about and much less likely to forget to update some code paths (now or in the future). Again if there's a valid reason to do the increments outside the data wrapper layer, please disregard, just thinking out loud.

jonathan-gramain (changes requested):

lib/api/apiUtils/object/locationStorageCheck.js
jonathan-gramain: Please disregard, saw you changed it already :+1:

lib/api/apiUtils/object/createAndStoreObject.js
dora-korpar: Accidentally left this over from an earlier iteration of the update on delete!

lib/api/apiUtils/object/locationStorageCheck.js
dora-korpar: Was just trying to appease everyone based on the meeting we had a few weeks ago. I agree that it would flood the logs, so will remove

lib/api/apiUtils/object/locationStorageCheck.js
dora-korpar: Yeah, I originally put it in here since there is a path in the objectDelete API that goes through the `createAndStoreObject` file, but since that case is only for delete markers (which don't affect size count), it makes more sense to put the metric check in the wrapper

lib/api/multipartDelete.js
dora-korpar: Yes! Good catch 😅

lib/data/wrapper.js
```diff
@@ -256,6 +257,19 @@ const data = {
                 log.error('delete error from datastore',
                     { error: err, key: objectGetInfo.key, moreRetries: 'no' });
             }
+            if (!isMdModelVersion2) {
+                // pass size as negative so location metric is decremented
+                return pushLocationMetric(objectGetInfo.dataStoreName,
```
jonathan-gramain: You will want to pass the decrement when `err` is not null (for the `ObjNotFound` case)

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -0,0 +1,49 @@
...
+    const sizeLimitGB = lc[location] && lc[location].details ?
+        lc[location].details.sizeLimitGB : undefined;
+    if (updateSize === 0 || sizeLimitGB === undefined) {
+        return cb();
+    }
+    // no need to list location metric, since it should be decreased
+    if (updateSize < 0) {
+        return pushLocationMetric(location, updateSize, log, cb);
+    }
+    return getLocationMetric(location, log, (err, bytesStored) => {
```
bennettbuchanan, line 33: Does this have the effect that Redis must be running if using S3? These methods aren't checking whether Utapi is included in the config.

dora-korpar: Yes! Thanks, I added a check in Config.js

tests/unit/testConfigs/locConstraintAssert.js
```diff
@@ -286,4 +286,18 @@ describe('locationConstraintAssert', () => {
...
+        const usEast1 = new LocationConstraint();
+        const locationConstraint = new LocationConstraint('aws_s3', true,
+            { sizeLimitGB: true });
+        assert.throws(() => {
+            locationConstraintAssert({
+                'us-east-1': usEast1,
+                'awsstoragesizelimit': locationConstraint,
+            });
+        },
+        '/bad config: locationConstraints[region].details.sizeLimitGB ' +
```
jonathan-gramain, line 300: I think a string does not work here (see https://nodejs.org/api/assert.html#assert_assert_throws_block_error_message)

Do you want a regex instead, starting with `/bad config...`? In which case should work but cannot be split across lines.

jonathan-gramain (changes requested):
One comment on the unit test, otherwise looks good to me!

tests/unit/testConfigs/locConstraintAssert.js
dora-korpar: The test passes :) It follows the same format as the rest of the tests in the file

jonathan-gramain: Here's what I mean:

```
jonathan@jonathan-XPS-L421X:~/git/ironman/backbeat$ node
> const assert = require('assert')
undefined
> > assert.throws(() => { throw new Error('foo'); }, '/bar')
... undefined
... > assert.throws(() => { throw new Error('foo'); }, /bar/)
... Error: foo
...     at assert.throws (repl:1:29)
...     at tryBlock (assert.js:305:5)
...     at innerThrows (assert.js:322:18)
...     at Function.throws (assert.js:350:3)
...     at repl:1:8
...     at sigintHandlersWrap (vm.js:22:35)
...     at sigintHandlersWrap (vm.js:73:12)
...     at ContextifyScript.Script.runInThisContext (vm.js:21:12)
...     at REPLServer.defaultEval (repl.js:340:29)
...     at bound (domain.js:280:14)
```

jonathan-gramain: i.e. `assert.throws` does not enforce the second argument check if it's a string

jonathan-gramain: And the rest of the tests in the file may be broken as well ;)

dora-korpar: rip

jonathan-gramain: Actually I just realize that it was in the same file where I saw such errors and added a `//TODO` in a previous PR to evolve to use `joi` for conf parsing.

jonathan-gramain (approved):

lib/Config.js
```diff
@@ -755,6 +760,11 @@ class Config extends EventEmitter {
                 this.utapi.expireMetricsTTL = config.utapi.expireMetricsTTL;
             }
         }
+        if (Object.keys(this.locationConstraints).some(
```
bennettbuchanan, line 763: Maybe it's overkill, but it could be helpful to also assert the `redis` and `localCache` properties.
Edit: Actually, I guess that would be taken care of if `this.utapi` is set, so disregard this 😄.

dora-korpar: The section that asserts `config.utapi` asserts `redis` and `localcache`, which is why I check for `this.utapi` here. If `this.utapi` exists, `redis` and `localcache` will too.

bennettbuchanan (approved):

PR #1346, opened 2018-07-02
Improving secret handling in CI

tcarmet:
# Intro

Sorry for the wait, but here it is!

Up for an early review until we upgrade eve in production. Works on my dev environment (link). Need your review to see if everything looks good for you. Feel free to ask any change from your point of view.

I'll merge this PR once we've upgraded Eve (which should be on Wednesday morning if everything goes fine on our side).

# Description

Made the migration of using only Vault to provide all the secrets for S3. They can be used with the syntax `%(secret:my_secret)s`.

`ShellCommandWithSecrets` will be deprecated in the future. Keep that in mind when writing CI code.

Also using a new feature in Eve [EVE-962](https://scality.atlassian.net/browse/EVE-962) that allows to interpolate those variables inside a kubernetes pod.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -9,19 +9,39 @@ branches:
 
 models:
   - env: &global-env
-      b2backend_B2_ACCOUNT_ID: '%(secret:b2backend_b2_account_id)s'
-      b2backend_B2_STORAGE_ACCESS_KEY: >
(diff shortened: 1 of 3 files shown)


tcarmet:
closed this PR as I was targetting development/8.0... my bad

PR #1351, opened 2018-07-05
ft: ZENKO 637 add proxy cert env var

dora-korpar:
Add support for HTTPS_PROXY_CERTIFICATE environment variable

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -855,26 +855,29 @@ class Config extends EventEmitter {
             }
             this.outboundProxy.url = proxyUrl;
             this.outboundProxy.certs = {};
+            const envCert = process.env.HTTPS_PROXY_CERTIFICATE;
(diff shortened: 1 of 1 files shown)


NicolasT:
I'm not sure using an environment variable is the right way to pass along certificate stores... They can be big, and should likely be overlayed with `ca-certificates` on some deployments where an HTTP proxy is not the issue, but custom CAs need to be added to e.g. access a RING/S3 exposed using HTTPS using a self-signed CA.

jonathan-gramain (approved):

dora-korpar:
@NicolasT the environment var would be used just to pass the path to the certificate. Take a look at the JIRA issue for more details :)

NicolasT:
:man_facepalming: Thanks @dora-korpar

ssalaues (approved):

PR #1356, opened 2018-07-06
ft: Add storage quota config

ssalaues:
Allows for storage quota configuration to be enabled from docker entry point.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -116,6 +116,10 @@ if [[ "$RECORDLOG_ENABLED" ]]; then
     JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .recordLog.enabled=true"
 fi
 
+if [[ "$STORAGE_LIMIT_ENABLED" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .utapi.metrics[.utapi.metrics | length]=\"location\""
+fi
+
 if [[ "$CRR_METRICS_HOST" ]]; then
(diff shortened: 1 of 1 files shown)


dora-korpar (approved):

giacope (approved):

docker-entrypoint.sh
```diff
@@ -116,6 +116,10 @@ if [[ "$RECORDLOG_ENABLED" ]]; then
     JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .recordLog.enabled=true"
 fi
 
+if [[ "$STORAGE_QUOTA_ENABLED" ]]; then
```
bennettbuchanan: Why not set `utapi.metrics` with the metrics array value as `['location']`? That way we can extend it later for the other metric granularities.

bennettbuchanan: Seems like we may eventually want to introduce other metrics.

jonathan-gramain: Can you change the name to `STORAGE_LIMIT_ENABLED`? It is the official name of the feature and will avoid confusion with the upcoming quotas support (per user).

jonathan-gramain (changes requested):

bennettbuchanan (changes requested):

jonathan-gramain:
@bert-e wait

ssalaues:
Updated PR to accommodate the requested changes.

jonathan-gramain (approved):

bennettbuchanan (approved):

PR #1357, opened 2018-07-06
bugfix: ZENKO-315 Only create version for replica

bennettbuchanan:
If the object is from a source bucket without versioning (i.e. NFS), then we want to create a version for the replica object even though none was provided in the object metadata value. This then leaves the source object without versioning.


bennettbuchanan:
@bert-e wait

rahulreddy (approved):

bennettbuchanan:
Closing in favor of bugflix Jira tracking branch https://github.com/scality/S3/pull/1359.

PR #1359, opened 2018-07-06
bugfix: ZENKO-665 Only create version for replica

bennettbuchanan:
If the object is from a source bucket without versioning (i.e. NFS), then we want to create a version for the replica object even though none was provided in the object metadata value. This then leaves the source object without versioning.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -310,10 +310,17 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // version (updating master as well) but with specified
         // versionId
         const options = {
-            versioning: omVal.versionId !== undefined ||
-                omVal.replicationInfo.isNFS === true,
+            versioning: true,
             versionId: omVal.versionId,
         };
+        // If the object is from a source bucket without versioning (i.e. NFS),
+        // then we want to create a version for the replica object even though
+        // none was provided in the object metadata value.
+        if (omVal.replicationInfo.isNFS) {
+            const isReplica = omVal.replicationInfo.status === 'REPLICA';
+            options.versioning = isReplica;
+            omVal.replicationInfo.isNFS = !isReplica;
+        }
         log.trace('putting object version', {
             objectKey: request.objectKey, omVal, options });
         return metadata.putObjectMD(bucketName, objectKey, omVal, options, log,
--- a/bin/create_bucket_with_NFS_enabled.js
+++ b/bin/create_bucket_with_NFS_enabled.js
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env node
 'use strict'; // eslint-disable-line strict
 
 require('../lib/nfs/utilities.js').createBucketWithNFSEnabled();
--- a/lib/nfs/utilities.js
+++ b/lib/nfs/utilities.js
@@ -5,7 +5,7 @@ const https = require('https');
 const logger = require('../utilities/logger');
 
 function _createBucketWithNFSEnabled(host, port, bucketName, accessKey,
-    secretKey, verbose, ssl) {
+    secretKey, verbose, ssl, locationConstraint) {
     const options = {
         host,
         port,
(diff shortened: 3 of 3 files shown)


rahulreddy:
@bert-e wait

rahulreddy (approved):

ploki:
@bert-e wait

lib/routes/routeBackbeat.js
```diff
@@ -304,12 +304,15 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+        // Specify both 'versioning' and 'versionId' to create a "new" version
+        // (updating master as well) but with specified versionId. If the object
+        // is from a source bucket without versioning (i.e. NFS), then we want
+        // to create a version for the replica object even though none was
+        // provided in the object metadata value.
         const options = {
             versioning: omVal.versionId !== undefined ||
-                omVal.replicationInfo.isNFS === true,
+                (omVal.replicationInfo.isNFS === true &&
+                omVal.replicationInfo.status === 'REPLICA'),
```
ploki: Previously, the versioning field was unconditionally set to true.
If the purpose of the changeset is to prevent the agent from making a put with versioning on the source object when it has the attribute replicationInfo.isNFS set to true, and only that particular case. I think the following is more accurate:

```
versioning: omVal.replicationInfo.isNFS !== true || omVal.replicationInfo.status === 'REPLICA'
```

ploki (changes requested):

lib/routes/routeBackbeat.js
ploki: in addition to this. the omVal.replicationInfo.isNFS must be cleared if omVal.replicationInfo.status === 'REPLICA' is true

bennettbuchanan: The versioning field was set to `true` on the condition that a version ID was provided in the `omVal`. We want to make an exception to this if the object has `isNFS` set to `true` and it is a replica object (i.e., the case that metadata is being updated on the destination where the source is an NFS bucket).

ploki: > The versioning field was set to true on the condition that a version ID was provided in the omVal

You changed this in b8a2c8ede5ce610f10fb6d887c6e729ec4bd23c6
originally it was unconditionally set to true

the exception is: not put with versioning the source objectMD when isNFS is true on that object. which is what says the expression I put in my previous comment

bennettbuchanan: Ah, yes indeed this simplifies the condition. Updated!

lib/routes/routeBackbeat.js
```diff
@@ -304,12 +304,14 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+        const { isNFS, status } =  omVal.replicationInfo;
+        // Specify both 'versioning' and 'versionId' to create a "new" version
+        // (updating master as well) but with specified versionId. If the object
+        // is from a source bucket without versioning (i.e. NFS), then we want
+        // to create a version for the replica object even though none was
+        // provided in the object metadata value.
         const options = {
-            versioning: omVal.versionId !== undefined ||
-                omVal.replicationInfo.isNFS === true,
+            versioning: isNFS !== true || status === 'REPLICA',
```
ploki: Do you want to reset ```omVal.replicationInfo.isNFS``` when ```status === 'REPLICA'```? You probably don't want to propagate this property to the destination object.

bennettbuchanan: I think it should be okay as-is because we are not altering any of the other `replicationInfo` properties sent along from the source. That way the metadata of the objects remain identical.

bennettbuchanan: I can think of a case though where the destination bucket also has CRR configured on it and is thus a source for another bucket in which case maybe we do want to update this value.

ploki: I was thinking about that particular use case

jonathan-gramain: minor comment: we should avoid using comparison operators on booleans, unless there's a very specific reason to do it: use `!isNfs` instead.

jonathan-gramain (approved):

ploki (approved):

lib/routes/routeBackbeat.js
bennettbuchanan: Okay, just to track this: we do not replicate 'REPLICA' objects (as per AWS docs https://docs.aws.amazon.com/AmazonS3/latest/dev/crr-what-is-isnot-replicated.html), but will this leave as-is.

bennettbuchanan:
@bert-e wait

PR #1362, opened 2018-07-09
bf: ZENKO 646 allow mdsearch by replication status

dora-korpar:
As discussed with team leads, deep metadata search is going to require a rework of our search method.
This PR is a quick-fix to enable md search by 'replicationStatus', as that is the most pressing use-case atm.

### DIFF
--- a/lib/api/apiUtils/bucket/validateSearch.js
+++ b/lib/api/apiUtils/bucket/validateSearch.js
@@ -41,6 +41,7 @@ function _validateTree(whereClause, possibleAttributes) {
                 const field = node[operator][0];
                 if (!field.startsWith('tags.') &&
                     !possibleAttributes[field] &&
+                    !field.startsWith('replicationInfo.') &&
                     !field.startsWith('x-amz-meta-')) {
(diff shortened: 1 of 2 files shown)


lib/api/apiUtils/bucket/validateSearch.js
```diff
@@ -63,6 +64,11 @@ function _validateTree(whereClause, possibleAttributes) {
 function validateSearchParams(searchParams) {
     let ast;
     try {
+        // allow using 'replicationStatus' as search param to increase
+        // ease of use, pending metadata search rework
+        // eslint-disable-next-line no-param-reassign
+        searchParams = searchParams.replace(
```
jonathan-gramain, line 71: Unless it has been decided otherwise for good reason, let's try to avoid mixing camelCase and dash-separated as accepted search words for consistency (especially if search is case-insensitive, that I don't know), `replication-status` would be better in this respect.

jeremyds (approved):

jonathan-gramain (approved):

PR #1371, opened 2018-07-12
bugfix: gcp behind an https proxy

giacope:


tmacro (approved):

lib/management/configuration.js
```diff
@@ -158,8 +167,8 @@ function patchConfiguration(newConf, log, cb) {
                             bucketName: l.details.bucketName,
                             mpuBucketName: l.details.mpuBucketName,
                             bucketMatch: l.details.bucketMatch,
-                            gcpEndpoint: l.details.endpoint ||
-                                'storage.googleapis.com',
+                            gcpEndpoint: gcpEndpoint,
+                            https: https,
```
alexanderchan-scality: nothing major.
`gcpEndpoint: gcpEndpoint` can be written as just `gcpEndpoint`
and `https: https` => `https`

alexanderchan-scality (approved):

lib/management/configuration.js
giacope: That's cool.

rahulreddy:
@bert-e wait

lib/management/configuration.js
```diff
@@ -149,6 +149,15 @@ function patchConfiguration(newConf, log, cb) {
                 case 'location-gcp-v1':
                     location.type = 'gcp';
                     if (l.details.secretKey && l.details.secretKey.length > 0) {
+                        let https = true;
```
rahulreddy: We can just set `https: true` instead of this logic. It was implemented for `aws_s3` type because we wanted to supported HTTP only endpoint for S3 Connector while using https for AWS.

rahulreddy (changes requested):

giacope:
@rahulreddy made the requested change.

rahulreddy (approved):

rahulreddy:
Thanks @giacomoguiulfo . Approved!

PR #1372, opened 2018-07-13
bf: ZENKO 726 storage limit config level

dora-korpar:
Small misunderstanding between Orbit and Cloudserver config left the sizeLimitGB property being set at different config levels.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -168,6 +168,11 @@ function locationConstraintAssert(locationConstraints) {
             typeof locationConstraints[l].isTransient),
                'bad config: locationConstraints[region]' +
                '.isTransient must be a boolean');
+        assert(['undefined', 'number'].includes(
+            typeof locationConstraints[l].sizeLimitGB),
+                'bad config: locationConstraints[region].sizeLimitGB ' +
+                'must be a number (in gigabytes)');
+
         const details = locationConstraints[l].details;
         assert(typeof details === 'object',
             'bad config: locationConstraints[region].details is ' +
@@ -223,11 +228,6 @@ function locationConstraintAssert(locationConstraints) {
             // default to true
             // eslint-disable-next-line no-param-reassign
             locationConstraints[l].details.supportsVersioning = true;
-        }
-        if (details.sizeLimitGB !== undefined) {
-            assert(typeof details.sizeLimitGB === 'number',
-                'bad config: locationConstraints[region].details.sizeLimitGB ' +
-                'must be a number (in gigabytes)');
         }
 
         if (locationConstraints[l].type === 'azure') {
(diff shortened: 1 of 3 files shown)


tmacro (approved):

lib/Config.js
```diff
@@ -168,6 +168,11 @@ function locationConstraintAssert(locationConstraints) {
             typeof locationConstraints[l].isTransient),
                'bad config: locationConstraints[region]' +
                '.isTransient must be a boolean');
+        assert(['undefined', 'number'].includes(
```
rahulreddy, line 171: Looks like `management` utility set sizeLimitGB to `null` as well. https://github.com/scality/S3/blob/c7bc911b913e07d0a3e424d41bfdecdd198f11b0/lib/management/configuration.js#L188

jonathan-gramain (approved):

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -21,8 +21,7 @@ function _gbToBytes(gb) {
  */
 function locationStorageCheck(location, updateSize, log, cb) {
     const lc = config.locationConstraints;
-    const sizeLimitGB = lc[location] && lc[location].details ?
-        lc[location].details.sizeLimitGB : undefined;
+    const sizeLimitGB = lc[location] ? lc[location].sizeLimitGB : undefined;
```
rahulreddy, line 24: Same thing here, you may have to check for `null` as well

tests/unit/testConfigs/locConstraintAssert.js
```diff
@@ -290,14 +291,14 @@ describe('locationConstraintAssert', () => {
     it('should throw error if sizeLimitGB is not a number', () => {
         const usEast1 = new LocationConstraint();
         const locationConstraint = new LocationConstraint('aws_s3', true,
-            { sizeLimitGB: true });
+            null, true);
```
rahulreddy, line 294: This is a valid way Orbit sets the config.

rahulreddy:
BertE merged before I could finish my review.

tests/unit/testConfigs/locConstraintAssert.js
jonathan-gramain: I think the invalid part here is `true` for the storage limit argument which shall be an integer

lib/Config.js
dora-korpar: fix incoming

lib/api/apiUtils/object/locationStorageCheck.js
dora-korpar: I checked and having `lc[location].sizeLimitGB` set to `null` would result in `sizeLimitGB` being set to undefined cos `null` is falsey :)

PR #1374, opened 2018-07-16
Implementing buildbot logfiles configuration

tcarmet:
# Improving debug of containers logs

While trying to help @tmacro on a issue this morning I found the option of `logfiles` in buildbot documentation and thought that it could be a nice to have in S3 CI.

## documentation
You can checkout the full documentation about it [here](http://docs.buildbot.net/latest/manual/cfg-buildsteps.html#shellcommand). But here's what you need to know:

`logfiles`

Sometimes commands will log interesting data to a local file, rather than emitting everything to stdout or stderr. For example, Twisted’s trial command (which runs unit tests) only presents summary information to stdout, and puts the rest into a file named _trial_temp/test.log. It is often useful to watch these files as the command runs, rather than using /bin/cat to dump their contents afterwards.

## Personal thoughts
Checkout how this change affect the [UI](https://eve.devsca.com/github/scality/s3/#/builders/9/builds/1715)! You can see that you have your stdio field and now also have S3 and you will be able to follow up the logs while the command is running! Which could be useful.

IMHO this should be implemented implicitly eve side. But while the development is not done yet, we could have this working now.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -87,9 +87,15 @@ models:
       name: install modules
       command: npm install
       haltOnFailure: True
-  - ShellCommand: &s3-log
-      name: s3 logs
-      command: cat /artifacts/s3.log || exit 0
+  - Upload: &upload-artifacts
+      source: /artifacts
+      urls:
+        - "*"
+  - ShellCommand: &follow-s3-log
+      logfiles:
+        s3:
+          filename: /artifacts/s3.log
+          follow: true
 
 stages:
   pre-merge:
@@ -165,6 +171,7 @@ stages:
             set -ex
             bash wait_for_local_port.bash 8000 40
             npm run multiple_backend_test"
+          <<: *follow-s3-log
           env:
             <<: *multiple-backend-vars
             <<: *global-env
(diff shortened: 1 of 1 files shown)


tcarmet:
@bert-e wait

tcarmet:
Not ready for merge, as the work is not 100% done, I'm just getting your thoughts on this and with your approval on the idea I'll take the work forward

eve/main.yml
```diff
@@ -90,6 +90,11 @@ models:
   - ShellCommand: &s3-log
```
ssalaues: Would replace this correct?

eve/main.yml
```diff
@@ -204,6 +212,7 @@ stages:
             set -ex
             bash wait_for_local_port.bash 8000 40
             npm run ft_test
+          <<: *follow-s3-log
```
ssalaues: Does this mean we could see the logs during the build steps instead of waiting until the end?

ssalaues (commented):
Overall looks like a good idea!

giacope:
Great idea! 👍

tmacro:
This would be awesome!

eve/main.yml
tcarmet: Exactly. I think I'll replace this step by an upload. As we won't need it anymore

tcarmet:
@ssalaues to answer your question, yes the logs appears as long as the build/shellcommands are running. Just force a build on this branch to see how it looks like in live.

tmacro (approved):

rahulreddy (approved):
Very helpful, I missed this output. Thanks @tcarmet !

giacope (approved):

ssalaues (approved):

rahulreddy:
@bert-e reset

tcarmet:
My pleasure @rahulreddy ! 

Could be nice if we do the same for Zenko e2e tests, let me know if you guys need this and I'll do my best helping you out!

dora-korpar (approved):

PR #1375, opened 2018-07-16
bf(zenko-473): Handle a unspecified storage quota correctly

tmacro:
# Pull request template

### Motivation and context

in configuration.js location.sizeLimitGB is set as follows 

```
location.sizeLimitGB = l.sizeLimitGB || null;
```

where `l` is the location read from the overlay config

later when checking the quota when uploading an object the value is tested as follows

```
    const sizeLimitGB = lc[location] ? lc[location].sizeLimitGB : undefined;
    if (updateSize === 0 || sizeLimitGB === undefined) {
        return cb();
    }
```

This incorrectly lets a sizeLimitGB of null pass through for checking which causes all PUTs to fail with the error `The assigned storage space limit for location awsbackend will be exceeded` on the client.

The fix for this bug is adding an additional check for  `sizeLimitGB === null`

### DIFF
--- a/lib/api/apiUtils/object/locationStorageCheck.js
+++ b/lib/api/apiUtils/object/locationStorageCheck.js
@@ -22,7 +22,7 @@ function _gbToBytes(gb) {
 function locationStorageCheck(location, updateSize, log, cb) {
     const lc = config.locationConstraints;
     const sizeLimitGB = lc[location] ? lc[location].sizeLimitGB : undefined;
-    if (updateSize === 0 || sizeLimitGB === undefined) {
+    if (updateSize === 0 || sizeLimitGB === undefined || sizeLimitGB === null) {
         return cb();
     }
     // no need to list location metric, since it should be decreased


JianqinWang (approved):

alexanderchan-scality (approved):

giacope (approved):

lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -22,7 +22,7 @@ function _gbToBytes(gb) {
 function locationStorageCheck(location, updateSize, log, cb) {
     const lc = config.locationConstraints;
     const sizeLimitGB = lc[location] ? lc[location].sizeLimitGB : undefined;
-    if (updateSize === 0 || sizeLimitGB === undefined) {
```
dora-korpar, line 25: Are you sure this fixes the issue? Cos when I tested, if `lc[location].sizeLimitGB` was set to null, `sizeLimitGB` was set to undefined

tmacro: maybe different node versions?, I'm definitely able to reproduce on my end.

dora-korpar: Scratch that, I realize what I did wrong. Thx for fixing 👍

dora-korpar (approved):

PR #1402, opened 2018-07-20
bf: ZENKO 790 redis limit return

dora-korpar:
Redis suddenly decided to return key values as strings, so new put sizes were being concatenated instead of added to the existing location value


rahulreddy:
Are there no unit/functional tests for this method?

philipyoo (approved):

dora-korpar:
@rahulreddy Since the value in question came from redis, it would require integration tests to test, of which there are none currently 👎 Obvs, this is something we should do asap

tmacro (approved):

rahulreddy:
We have Redis running as part of S3 CI, we should be able to put something together to test this independently without Integration. I am approving this PR keeping in mind the time constraint but please follow up with a functional test for this.

rahulreddy (approved):

PR #1413, opened 2018-07-20
bf: ZENKO-745 unique location metrics redis keys

jonathan-gramain:
Use new 'objectId' field set by Orbit to each location for the redis
key, instead of the location name. This will avoid conflicts if
locations are re-created with the same name.

Take the opportunity to enforce an objectId to be set on each
location, even if the config is from a static file, to make its use
durable.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -160,6 +160,16 @@ function locationConstraintAssert(locationConstraints) {
         assert(supportedBackends.indexOf(locationConstraints[l].type) > -1,
             'bad config: locationConstraints[region].type must ' +
             `be one of ${supportedBackends}`);
+        assert(typeof locationConstraints[l].objectId === 'string',
+               'bad config: locationConstraints[region].objectId is ' +
+               'mandatory and must be a unique string across locations');
+        assert(Object.keys(locationConstraints)
+               .filter(loc => (locationConstraints[loc].objectId ===
+                               locationConstraints[l].objectId))
+               .length === 1,
+               'bad config: location constraint objectId ' +
+               `"${locationConstraints[l].objectId}" is not unique across ` +
+               'configured locations');
         assert(typeof locationConstraints[l].legacyAwsBehavior
             === 'boolean',
             'bad config: locationConstraints[region]' +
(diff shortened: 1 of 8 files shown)


dora-korpar:
`objectId` seems like a kind of generic and misleading keyname to me. Could we use something like `locationId` or `uniqueId`?
Also, would it not be simpler (from user end) for us to just generate a unique string for each location? For example, what if a user is doing the same thing you did in the locationConfig, and uses the location constraint name as the id, then we run into the same problem this issue is trying to fix.

jonathan-gramain:
@dora-korpar on `objectId`: I think the reason why this name is that Orbit intends to put such id to various kinds of objects, not just locations (@RachedBenMustapha ?). In this respect, mapping this field to another name like `locationId` would I think bring more confusion rather than less. If people think differently I'm open to changes though.

On the unique ID: it seems like we may use the same ID for more and more features, and it makes sense in general to have one. Orbit is exactly doing what you suggest already by generating a UUID for each location. If the user wants to run without orbit, its responsibility is to add unique IDs to his locations then, because we rely on the static config. Suggestions for improving this to make it simpler etc. are welcome, but right now I don't have better in mind.

dora-korpar:
Aha, didn't realize the id generation would be in orbit, but that makes sense. I'll look out for that pr then as well :)

jonathan-gramain:
@dora-korpar it's actually there with the private endpoint ;)

lib/Config.js
```diff
@@ -159,7 +159,17 @@ function locationConstraintAssert(locationConstraints) {
             'mandatory and must be a string');
         assert(supportedBackends.indexOf(locationConstraints[l].type) > -1,
             'bad config: locationConstraints[region].type must ' +
-            `be one of ${supportedBackends}`);
+               `be one of ${supportedBackends}`);
```
dora-korpar: accidental indentation?

jonathan-gramain: reverted indentation

dora-korpar (approved):

rachedbenmustapha:
@dora-korpar orbit does not care about locations, it's all about anonymous faceless objects :)

rachedbenmustapha (approved):

jonathan-gramain:
@bert-e reset

PR #1427, opened 2018-07-24
Feature/zenko 713 new management agent

jeremyds:
# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/management/agentClient.js
+++ b/lib/management/agentClient.js
@@ -4,6 +4,11 @@ const managementAgentMessageType = {
 };
 
 
+function isManagementAgentUsed() {
+    return process.env.MANAGEMENT_USE_AGENT === '1';
+}
+
 module.exports = {
     managementAgentMessageType,
+    isManagementAgentUsed,
 };
--- a/lib/management/agentClient.js
+++ b/lib/management/agentClient.js
@@ -1,0 +1,9 @@
+const managementAgentMessageType = {
+    /** Message that contains the loaded overlay */
+    NEW_OVERLAY: 1,
+};
+
+
+module.exports = {
+    managementAgentMessageType,
+};
--- a/lib/management/configuration.js
+++ b/lib/management/configuration.js
@@ -258,6 +258,8 @@ function saveConfigurationVersion(cachedOverlay, remoteOverlay, log, cb) {
                 if (error) {
                     log.error('could not save configuration',
                         { configurationVersion: remoteOverlay.version });
+                    cb(error);
+                    return;
                 }
                 metadata.putObjectMD(managementDatabaseName,
                     latestOverlayVersionKey, remoteOverlay.version, {}, log,
--- a/lib/management/configuration.js
+++ b/lib/management/configuration.js
@@ -256,12 +256,19 @@ function saveConfigurationVersion(cachedOverlay, remoteOverlay, log, cb) {
         metadata.putObjectMD(managementDatabaseName, objName, remoteOverlay,
             {}, log, error => {
                 if (error) {
-                    log.error('could not save configuration version',
+                    log.error('could not save configuration',
                         { configurationVersion: remoteOverlay.version });
                 }
                 metadata.putObjectMD(managementDatabaseName,
                     latestOverlayVersionKey, remoteOverlay.version, {}, log,
-                    error => cb(error, remoteOverlay));
+                    error => {
+                        if (error) {
+                            log.error('could not save configuration version', {
+                                configurationVersion: remoteOverlay.version,
+                            });
+                        }
+                        cb(error, remoteOverlay);
+                    });
             });
     } else {
         log.debug('no remote configuration to cache yet');
--- a/lib/management/index.js
+++ b/lib/management/index.js
@@ -12,6 +12,7 @@ const {
 const { initManagementCredentials } = require('./credentials');
 const { startWSManagementClient } = require('./push');
 const { startPollingManagementClient } = require('./poll');
+const { isManagementAgentUsed } = require('./agentClient');
 
 const initRemoteManagementRetryDelay = 10000;
 
@@ -78,6 +79,8 @@ function initManagement(log, callback) {
         return;
     }
     async.waterfall([
+        // eslint-disable-next-line arrow-body-style
+        cb => { return isManagementAgentUsed() ? metadata.setup(cb) : cb(); },
         cb => initManagementDatabase(log, cb),
         cb => metadata.getUUID(log, cb),
         (instanceId, cb) => initManagementCredentials(
--- a/lib/management/index.js
+++ b/lib/management/index.js
@@ -65,10 +65,12 @@ function startManagementListeners(instanceId, token) {
  *
  * @param  {werelogs~Logger} log Request-scoped logger to be able to trace
  *  initialization process
+ * @param  {function} callback Function to call once the overlay is loaded
+ *  (overlay)
  *
  * @returns {undefined}
  */
-function initManagement(log) {
+function initManagement(log, callback) {
     if ((process.env.REMOTE_MANAGEMENT_DISABLE &&
         process.env.REMOTE_MANAGEMENT_DISABLE !== '0')
         || process.env.S3BACKEND === 'mem') {
@@ -80,14 +82,13 @@ function initManagement(log) {
         cb => metadata.getUUID(log, cb),
         (instanceId, cb) => initManagementCredentials(
             managementEndpoint, instanceId, log, cb),
-        (instanceId, token, cb) => loadCachedOverlay(log, (err, conf) => {
-            if (err) {
-                return cb(err);
-            }
-            return patchConfiguration(conf, log,
-                err => cb(err, instanceId, token));
-        }),
-    ], (error, instanceId, token) => {
+        (instanceId, token, cb) =>
+            loadCachedOverlay(log, (err, overlay) => cb(err, instanceId,
+                                                        token, overlay)),
+        (instanceId, token, overlay, cb) =>
+            patchConfiguration(overlay, log,
+                               err => cb(err, instanceId, token, overlay)),
+    ], (error, instanceId, token, overlay) => {
         if (error) {
             log.error('could not initialize remote management, retrying later',
                 { error });
@@ -98,6 +99,9 @@ function initManagement(log) {
             log.info(`this deployment's Instance ID is ${instanceId}`);
             log.end('management init done');
             startManagementListeners(instanceId, token);
+            if (callback) {
+                callback(overlay);
+            }
         }
     });
 }
--- a/lib/management/push.js
+++ b/lib/management/push.js
@@ -141,10 +141,7 @@ function startWSManagementClient(pushEndpoint, instanceId, token) {
               || 'localhost';
             const port = process.env.SECURE_CHANNEL_DEFAULT_FORWARD_TO_PORT
               || _config.port;
-            socket = net.createConnection({
-                host: host,
-                port: port,
-            });
+            socket = net.createConnection(port, host);
 
             socket.on('data', data => {
                 ws.send(ChannelMessageV0.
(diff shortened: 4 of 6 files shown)


jeremyds:
for reviewers: only the last commit (ZENKO 713) is relevant, others are in [this PR.](https://github.com/scality/S3/pull/1378)

managementAgent.js
```diff
@@ -0,0 +1,67 @@
...
+    });
+
+    wss.on('error', error => {
+        logger.error('management agent websocket server error', { error });
+    });
+
+    addOverlayMessageListener((remoteOverlay) => {
+        wss.clients.forEach((client) => {
+            if (client.readyState !== client.OPEN) {
+                logger.warning('client socket not in ready state', { client });
```
jeremyds: logger.error

lib/management/configuration.js
```diff
@@ -13,6 +13,9 @@ const managementDatabaseName = 'PENSIEVE';
 const replicatorEndpoint = 'zenko-cloudserver-replicator';
 const { decryptSecret } = arsenal.pensieve.credentialUtils;
 
+
+let configurationListeners = [];
```
rachedbenmustapha: This can be const if it's pushed into

managementAgent.js
```diff
@@ -3,8 +3,16 @@ const Uuid = require('uuid');
 
 const logger = require('./lib/utilities/logger');
 const { initManagement } = require('./lib/management');
+const { addConfigurationListener } = require('./lib/management/configuration');
 const _config = require('./lib/Config').config;
+const WebSocket = require('ws');
+
+
+//TODO: signal handler to cleanly shutdown server
+//TODO: ping messages between server and client
```
rachedbenmustapha: I think it would be good to implement the ping part in this PR

rachedbenmustapha (changes requested):
In order to be more resilient to disconnections and timing issues, we should also send the latest overlay on the websocket as soon as a client is connected.

lib/management/configuration.js
```diff
@@ -300,10 +310,22 @@ function loadCachedOverlay(log, callback) {
         });
 }
 
+/**
```
jeremyds: should be removed, interactive rebase error I guess, now in push.js.

managementAgent.js
```diff
@@ -55,5 +134,19 @@ function managementAgentWS() {
                 }
             });
         });
-    });
+    }
+
+    checkBrokenConnections() {
+        this.wss.clients.forEach((client) => {
+            if (!client.isAlive) {
+                logger.error('close broken connection');
```
rachedbenmustapha: This should be info instead of error - errors are for abnormal conditions and a client restarting is normal.

lib/management/configuration.js
```diff
@@ -256,12 +259,19 @@ function saveConfigurationVersion(cachedOverlay, remoteOverlay, log, cb) {
         metadata.putObjectMD(managementDatabaseName, objName, remoteOverlay,
             {}, log, error => {
                 if (error) {
-                    log.error('could not save configuration version',
+                    log.error('could not save configuration',
```
rachedbenmustapha, line 259: It's not new with this change but since there's already a diff here: we should not continue to update the version pointer if there was a failure saving the overlay in the first place.

managementAgent.js
```diff
@@ -1,51 +1,130 @@
...
-    wss.on('listening', ws => {
-        logger.info('management agent websocket server listening', { port });
-    });
+        socket.isAlive = true;
+        socket.on('pong', hearthbeat.bind(socket));
 
-    wss.on('error', error => {
-        logger.error('management agent websocket server error', { error });
-    });
+        if (!socket.readyState === socket.OPEN) {
```
rachedbenmustapha: Cosmetic but it took me a few seconds to see that the condition is reversed, I suggest using `!==` instead

managementAgent.js
```diff
@@ -1,51 +1,130 @@
...
+        /* Define REPORT_TOKEN env variable needed by the management
+         * module. */
+        process.env.REPORT_TOKEN = process.env.REPORT_TOKEN
+          || _config.reportToken
+          || Uuid.v4();
+
+        metadata.setup(error => {
+            if (error) {
+                const msg = `failed to setup metadata: ${error}`;
+                throw new Error(msg);
```
rachedbenmustapha: We should not throw in async callbacks, instead call a callback with the failure so that the toplevel code can schedule a retry

managementAgent.js
```diff
@@ -1,51 +1,130 @@
...
+            }
+            initManagement(logger.newRequestLogger(), (overlay) => {
+                if (overlay) {
+                    this.loadedOverlay = overlay;
+                    this.startServer();
+                } else {
+                    const msg = `failed to init management: ${error}`;
+                    throw new Error('failed to init management');
```
rachedbenmustapha: Same here, we should not throw

managementAgent.js
```diff
@@ -1,51 +1,130 @@
...
+        logger.info('websocket server listening',
+                    { port: this.port });
+    }
+
+    onError(error) {
+        logger.error('websocket server error', { error });
+    }
+
+    onNewOverlay(remoteOverlay) {
+        //XXX patch configuration?
```
rachedbenmustapha: I don't think the server has anything to patch as the overlay only contains s3-protocol specific items. On thing that will come later is the `browserAccess` flag

managementAgent.js
```diff
@@ -1,51 +1,130 @@
...
-        logger.info('management agent websocket server listening', { port });
-    });
+        socket.isAlive = true;
+        socket.on('pong', hearthbeat.bind(socket));
 
-    wss.on('error', error => {
-        logger.error('management agent websocket server error', { error });
-    });
+        if (!socket.readyState === socket.OPEN) {
+            logger.error('client socket not in ready state', { socket });
```
rachedbenmustapha: This could include a lot of unpleasant data to look at that can be big or circular, so for clarity and for preserving performance we should select a few meaningful attributes (if available) such as `host`, `port`, `readyState` and so on

lib/management/agentClient.js
```diff
@@ -0,0 +1,16 @@
+const managementAgentMessageType = {
+    /** Message that contains a configuration overlay */
+    NEW_OVERLAY_VERSION: 1,
+    /** Message that contains the loaded overlay */
+    LOADED_OVERLAY: 2,
```
rachedbenmustapha: Why is there a different message type `LOADED_OVERLAY`? I think only one message type to push overlays would be clearer.

rachedbenmustapha (changes requested):

lib/management/agentClient.js
jeremyds: I thought it maybe useful to distinguish the loaded overlay from a new one. But as long as I don't have a precise use case in mind, let's have only one message type for the moment.

lib/management/configuration.js
jeremyds: not relevant anymore, code has been moved to push.js

lib/management/configuration.js
jeremyds: right, I am going to fix it in a separate commit.

managementAgent.js
jeremyds: done

managementAgent.js
jeremyds: done too

managementAgent.js
jeremyds: right, not easy to read

managementAgent.js
jeremyds: done

managementAgent.js
jeremyds: thanks for the answer, XXX to remove

managementAgent.js
jeremyds: done

managementAgent.js
```diff
@@ -1,59 +1,159 @@
...
+          || _config.reportToken
+          || Uuid.v4();
 
-setTimeout(() => {
-    managementAgentWS();
-}, 6);
+        metadata.setup(error => {
+            if (error) {
+                const msg = `failed to setup metadata: ${error}`;
+                return cb(new Error(msg));
```
rachedbenmustapha: Real `Error` objects usually don't print well in our json logs so just forwarding the error we get to the callback is safer. A log before that saying `log.error('failed to setup metadata', { error })` would be enough to track

lib/management/push.js
```diff
@@ -141,10 +146,7 @@ function startWSManagementClient(pushEndpoint, instanceId, token) {
               || 'localhost';
             const port = process.env.SECURE_CHANNEL_DEFAULT_FORWARD_TO_PORT
               || _config.port;
-            socket = net.createConnection({
-                host: host,
-                port: port,
-            });
+            socket = net.createConnection(host, port);
```
rachedbenmustapha: The order is wrong here, its port then host (nodejs nicety). Why the change?

rachedbenmustapha (changes requested):

lib/management/index.js
```diff
@@ -80,14 +82,14 @@ function initManagement(log) {
...
         (instanceId, cb) => initManagementCredentials(
             managementEndpoint, instanceId, log, cb),
-        (instanceId, token, cb) => loadCachedOverlay(log, (err, conf) => {
+        (instanceId, token, cb) => loadCachedOverlay(log, (err, overlay) => {
             if (err) {
                 return cb(err);
             }
-            return patchConfiguration(conf, log,
-                err => cb(err, instanceId, token));
+            return patchConfiguration(overlay, log,
```
nicolas2bert: It might be clearer to add another waterfall callback here for `patchConfiguration()`. 
`(overlay, instanceId, ... cb) => patchConfiguration(overlay...)`

managementAgent.js
```diff
@@ -1,59 +1,159 @@
...
+
+
+// TODO: auth?
+
+const CHECK_BROKEN_CONNECTIONS_FREQUENCY_MS = 15000;
+
+
+class ManagementAgentServer {
+    constructor() {
+        this.port = _config.managementAgent.port || 8010;
```
nicolas2bert: Not sure how `_config.managementAgent.port` can be `undefined` or `0` since you already asserted it in the `Config.js`. _Better safe than sorry?_

managementAgent.js
```diff
@@ -1,59 +1,159 @@
...
+        this.wss = null;
+        this.loadedOverlay = null;
+
```
nicolas2bert: You could also have done it this way:
```
this.stop = this.stop.bind(this);

process.on('SIGINT', this.stop);

process.on('SIGHUP', this.stop);

process.on('SIGQUIT', this.stop);

process.on('SIGTERM', this.stop);
```

nicolas2bert (approved):

managementAgent.js
jeremyds: yes, exactly

lib/management/index.js
jeremyds: right, done

lib/management/push.js
jeremyds: wow... linter complains about the object form {port: value, host: value}

managementAgent.js
jeremyds: done

managementAgent.js
jeremyds: done (I missed this comment)

lib/management/push.js
rachedbenmustapha: Maybe it wants `{ host, port }` since the vars have the same name as the keys

managementAgent.js
```diff
@@ -1,59 +1,169 @@
...
+                return;
+            }
+            client.isAlive = false;
+            client.ping();
+        });
+    }
 }
+
+const server = new ManagementAgentServer();
+server.start();
```
rachedbenmustapha, line 162: If `ManagementAgentServer.start()` fails and we don't provide a callback that exits, are we sure that the process will exit with an error status? When deployed in k8s for example if mongodb is not yet available we want to be sure that we retry initializing later. Arguably we could crash as well but it only works well when deployed inside an orchestrator.

rachedbenmustapha (changes requested):

rachedbenmustapha (approved):

PR #1431, opened 2018-07-25
improvement: ZENKO-819 use regional endpoints

rahulreddy:
This commit adds a retry for requests to AWS where the api call fails as the common endpoint
s3.amazonaws.com is not redirected to the regional endpoints anymore. If the original call
recives AuthorizationHeaderMalformed, a GET Bucket location call is made and the endpoint is
updated to use the correct region in it's hostname.


tmacro (approved):

jonathan-gramain (approved):
neat :shipit:

PR #1443, opened 2018-07-30
ft: ZENKO-869 increase GCP part limit

alexanderchan-scality:


alexanderchan-scality:
@bert-e reset

rahulreddy:
@bert-e wait

PR #1446, opened 2018-07-30
bf: ZENKO-751 update arsenal package

philipyoo:
object key max size is set to 915

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -223,8 +223,8 @@
       "dev": true
     },
     "arsenal": {
-      "version": "github:scality/Arsenal#fdbeed1c4eb5bff95662ad07470afcb05b077d38",
-      "from": "github:scality/Arsenal#fdbeed1",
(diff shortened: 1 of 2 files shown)


JianqinWang (approved):

anurag4DSB (approved):

jonathan-gramain (changes requested):
can you revert the unnecessary changes in package-lock.json?

philipyoo:
@jonathan-gramain So I can't say for certain, but seems like there are some known issues with package lock files and how they are generated. But the problems seem to be related to either version differences or OS differences (https://github.com/npm/npm/issues/17749)

I saw sha-512 -> sha-1 changes being made. To fix this, I had to remove my existing node module folder and npm install again.
I wasn't aware the lock file would be affected by the existing node modules dir, especially since my package.json changes were only for arsenal   o.O

philipyoo:
@bert-e reset

rahulreddy (approved):

philipyoo:
@bert-e reset

PR #1448, opened 2018-07-30
Bugfix/zenko 791 quota update on put fail

dora-korpar:
Location metric is updated before a data is actually put. If put fails, metric should be decremented appropriately.

### DIFF
--- a/lib/data/multipleBackendGateway.js
+++ b/lib/data/multipleBackendGateway.js
@@ -201,7 +201,22 @@ const multipleBackendGateway = {
                     return cb(err);
                 }
                 return client.uploadPart(request, streamingV4Params, stream,
-                    size, key, uploadId, partNumber, bucketName, log, cb);
+                size, key, uploadId, partNumber, bucketName, log,
+                (err, partInfo) => {
+                    if (err) {
+                        // if error putting part, counter should be decremented
+                        return locationStorageCheck(location, -size, log,
+                        error => {
+                            if (error) {
+                                log.error('Error decrementing location ' +
+                                    'metric following object PUT failure',
+                                    { error: error.message });
+                            }
+                            return cb(err);
+                        });
+                    }
+                    return cb(null, partInfo);
+                });
             });
         }
         return cb();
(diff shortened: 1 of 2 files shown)


lib/data/multipleBackendGateway.js
```diff
@@ -201,7 +201,22 @@ const multipleBackendGateway = {
...
+                        error => {
+                            if (error) {
+                                log.error('Error decrementing location ' +
+                                'metric after following object PUT failure: ' +
+                                `${error}`);
+                            }
+                            return cb(err);
+                        });
+                    }
+                    return cb(err, partInfo);
```
jonathan-gramain: You can replace `err` by `null` here.

lib/data/multipleBackendGateway.js
```diff
@@ -287,7 +302,18 @@ const multipleBackendGateway = {
                         dataStoreType: client.clientType,
                         dataStoreVersionId,
                     };
-                    cb(err, dataRetrievalInfo);
+                    if (err) {
+                        // if error copying obj, counter should be decremented
+                        locationStorageCheck(destLocationConstraintName,
```
jonathan-gramain: This looks fine per se, but to avoid surprises I would think it's better to map exactly what is done for the other rollback cases, i.e. do `return locationStorageCheck` here to make it synchronous with the response, and `return cb(null, dataRetrievalInfo)` on success.

lib/data/multipleBackendGateway.js
```diff
@@ -201,7 +201,22 @@ const multipleBackendGateway = {
...
                 return client.uploadPart(request, streamingV4Params, stream,
-                    size, key, uploadId, partNumber, bucketName, log, cb);
+                size, key, uploadId, partNumber, bucketName, log,
+                (err, partInfo) => {
+                    if (err) {
+                        // if error putting part, counter should be decremented
+                        return locationStorageCheck(location, -size, log,
+                        error => {
+                            if (error) {
+                                log.error('Error decrementing location ' +
```
jonathan-gramain, line 211: Let's stick with the general logging practice for errors: `log.error('some message', { error: err.message })`

jonathan-gramain (changes requested):

dora-korpar:
@bert-e reset

lib/data/multipleBackendGateway.js
```diff
@@ -287,7 +302,18 @@ const multipleBackendGateway = {
...
+                    if (err) {
+                        // if error copying obj, counter should be decremented
+                        return locationStorageCheck(destLocationConstraintName,
+                        -storeMetadataParams.size, log, error => {
+                            if (error) {
+                                log.error('Error decrementing location ' +
+                                    'metric after following object PUT failure',
+                                    { error: error.message });
+                            }
+                        });
```
jonathan-gramain: Then with the added `return` you now need to call `cb(err)` ;)

alexanderchan-scality (approved):

jonathan-gramain (approved):

rahulreddy (commented):
Please use `async` lib to avoid callback hell next time.

rahulreddy (approved):

dora-korpar:
@bert-e reset

PR #1462, opened 2018-08-02
Do not parse list entries from MD6.4.7

MathieuCassagne:


MathieuCassagne:
@bert-e reset

MathieuCassagne:
@bert-e help

ghost (approved):
LGTM, backwards compat is achieved by testing the entry.value's type before going further.

Good thing it was this simple to address within S3, but I'm not very confortable on having that here, as opposed to having that in the S3C-MD backend (that may be in arsenal?).

ghost:
As I'm unclear yet about what change/piece-of-code is at the source of this compatibliity issue, I cannot tell where the best spot to fix it would be.

lamphamsy (approved):

DavidPineauScality:
@bert-e bypass_leader_approval

tcarmet:
@bert-e bypass_leader_approval

PR #1465, opened 2018-08-06
bf: ZENKO-916 do not replicate ACL change

jonathan-gramain:
We should not reset the replication status to PENDING when changing
ACLs of an object. This because we choose not to replicate ACLs to
replication targets. Later on we may see this as a feature but not
now.

As per description in ZENKO-916, it fixes because putting ACLs is the
only known problematic case where it breaks with a transient source
location. E.g. object tags get replicated correctly because we only
replicate tags specifically in such case.

Note that https://scality.atlassian.net/browse/ZENKO-932 may look like a similar issue from a user standpoint but will be treated separately as the root cause is different.

### DIFF
--- a/lib/metadata/acl.js
+++ b/lib/metadata/acl.js
@@ -1,6 +1,5 @@
 const { errors } = require('arsenal');
 
-const getReplicationInfo = require('../api/apiUtils/object/getReplicationInfo');
 const aclUtils = require('../utilities/aclUtils');
 const constants = require('../../constants');
 const metadata = require('../metadata/wrapper');
@@ -17,12 +16,6 @@ const acl = {
         log.trace('updating object acl in metadata');
         // eslint-disable-next-line no-param-reassign
         objectMD.acl = addACLParams;
-        const replicationInfo = getReplicationInfo(objectKey, bucket, true);
-        if (replicationInfo) {
-            // eslint-disable-next-line no-param-reassign
-            objectMD.replicationInfo = Object.assign({},
-                objectMD.replicationInfo, replicationInfo);
-        }
         metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params, log,
             cb);
     },
--- a/tests/unit/api/objectReplicationMD.js
+++ b/tests/unit/api/objectReplicationMD.js
@@ -346,18 +346,27 @@ describe('Replication object MD without bucket replication config', () => {
                     return done();
                 }));
 
-        it("should update status to 'PENDING' and content to '['METADATA']' " +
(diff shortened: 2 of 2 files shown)


jonathan-gramain:
This change on E2E tests should ensure we do not change the replication status on ACL change: https://github.com/scality/Zenko/pull/252

JianqinWang (approved):

philipyoo (approved):

lib/metadata/acl.js
```diff
@@ -17,12 +16,6 @@ const acl = {
         log.trace('updating object acl in metadata');
         // eslint-disable-next-line no-param-reassign
         objectMD.acl = addACLParams;
-        const replicationInfo = getReplicationInfo(objectKey, bucket, true);
```
bennettbuchanan, line 20: I'm not sure how much work it would take to make this dynamically set based on the location constraint used. If it's a small change to check the external backend type, I'd say we could put this in an `if` block so that we don't have to revisit this change should we eventually want to support this CRR operation to another Zenko instance. Also, that would support backwards compatibility with S3C releases (though maybe it's irrelevant going forward). If, in your view, it's not worth the effort at this point then I'm good with that as well. 😄

rahulreddy (changes requested):
We would still want to replicate ACLs for S3C. Is there something we could do, like @bennettbuchanan suggested, don't replicate ACLs if the location is transient?

lib/metadata/acl.js
jonathan-gramain: Good point, I think for now I will leave as-is, as I'm not sure of the implications e.g. when mixing multiple backends where one at least would be a zenko (S3C) target. When eventually S3C will become 8.x, we will need to revisit the overall mechanism to make sure we support correctly replicating things to a S3C target amongst other clouds, and this will pop up as an action item.

As we have an integration test for ACL replication on S3C target I'm confident we will not forget to take care of this so should be fine for now especially considering the imminent release :)

jonathan-gramain:
@rahulreddy do you mean we need to replicate ACLs in zenko for S3C target? In such case, yes it could be necessary to rework. If you mean S3C as per the S3C way of replicating, see my reply, I think we can skip for now.

rahulreddy (approved):

rahulreddy:
After offline discussion with @jonathan-gramain we will come back and add support for ACLs public for Zenko and all ACLs for S3C.

jonathan-gramain:
Follow-up ticket: https://scality.atlassian.net/browse/ZENKO-948

PR #1467, opened 2018-08-07
Bugfix/zenko 922/add redis disconnect

alexanderchan-scality:


alexanderchan-scality:
@bert-e reset

philipyoo (approved):

alexanderchan-scality:
@bert-e reset

alexanderchan-scality:
@bert-e  reset

tests/functional/utilities/reportHandler.js
```diff
@@ -136,7 +139,32 @@ describe('reportHandler::getCRRStats', function testSuite() {
     this.timeout(20000);
     let redisClient;
 
-    before(done => {
+    function assertResults(res) {
```
JianqinWang: minor nitpick: imo helper functions should be at the beginning of file/outside of the test functions themselves

tests/functional/utilities/reportHandler.js
```diff
@@ -147,37 +175,59 @@ describe('reportHandler::getCRRStats', function testSuite() {
...
+                redisChecker.client('list', (err, res) => {
+                    assert.ifError(err, `Expect success, but got ${err}`);
+                    const clients = res.split('\n').filter(c => !!c);
+                    preCheckCount = clients.length;
+                    next();
+                });
+            },
+            checkResults: next => {
+                getCRRStats(logger, (err, res) => {
+                    assert.ifError(err,
```
JianqinWang: For the `assert.ifError`, I believe all other times it has been used in tests it only accepts 1 param (err). I also don't see documentation for the function accepting a 2nd param here: https://nodejs.org/api/assert.html#assert_assert_iferror_value

In any case, if `assert.ifError` finds an error, it will output the error info as a stack trace.

alexanderchan-scality:
@bert-e reset

eve/main.yml
```diff
@@ -65,13 +65,15 @@ models:
       S3BACKEND: "mem"
       MPU_TESTING: "yes"
       S3METADATA: mongodb
+      REDIS_TEST: "true"
```
rahulreddy: I am worried about introducing another environment variable. Can we do without?

lib/utilities/reportHandler.js
```diff
@@ -193,6 +193,12 @@ function getCRRStats(log, cb, _config) {
                 return done(null, retObj);
             }
         ),
+        retrieveRedisClients: done => {
+            if (process.env.REDIS_TEST === 'true') {
+                return backbeatMetrics.listClients(done);
```
rahulreddy: You want to return this only in the test mode because you are worried about revealing internals for security?

lib/utilities/reportHandler.js
```diff
@@ -201,16 +207,28 @@ function getCRRStats(log, cb, _config) {
...
+            const total = (res && res.total) || {};
+            const byLocation = (res && res.byLocation) || {};
+            const retObj = {
+                completions: total.completions,
+                failures: total.failures,
+                backlog: total.backlog,
+                throughput: total.throughput,
+                byLocation,
+            };
+            if (process.env.REDIS_TEST === 'true') {
```
rahulreddy: This can just be CI=true

rahulreddy (changes requested):

rahulreddy:
@bert-e wait

eve/main.yml
alexanderchan-scality: will fix and use `CI=true` instead

lib/utilities/reportHandler.js
alexanderchan-scality: It was more about not introducing a property that may cause the orbit api to reject the report due to unexpected schema

JianqinWang (approved):

jonathan-gramain (approved):

rahulreddy (approved):

rahulreddy:
@bert-e reset

PR #1475, opened 2018-08-08
ft: ZENKO-925 increase metrics expiry to 24hrs

philipyoo:


JianqinWang (approved):

tests/functional/utilities/reportHandler.js
```diff
@@ -88,7 +89,7 @@ describe('reportHandler::_crrRequest', function testSuite() {
             backbeatMetrics = new backbeat.Metrics({
                 redisConfig: config.redis,
                 validSites: sites,
-                internalStart: Date.now() - 900000,
+                internalStart: Date.now() - (EXPIRY * 1000),
```
JianqinWang: quick comment: why do you use a variable `EXPIRY` here when you simply used the number in the function file?

philipyoo:
@bert-e wait

tests/functional/utilities/reportHandler.js
philipyoo: ah, let me add an expiry constant to the function file. I think it will be nice to have it more explicit if changes in the future will need to be made here.

I think this could use a refactor to make more manageable in the future though, but for now, just want to increase expiry

alexanderchan-scality (approved):

tests/functional/utilities/reportHandler.js
JianqinWang: I had guessed that adding the variable here was so that it would be more manageable for future changes :) I agree with having the variable in both files

philipyoo:
@bert-e reset

dora-korpar (approved):

tests/functional/utilities/reportHandler.js
bennettbuchanan: I'd say for purposes of maintainability, it would be ideal to have the constant variable the same across files (e.g., instead of using both `EXPIRY` and `CRR_METRIC_EXPIRY`).

bennettbuchanan (approved):

philipyoo:
@bert-e wait

philipyoo:
@bennettbuchanan sgtm, pushing up a change soon!

philipyoo:
@bert-e status

philipyoo:
@bert-e reset

PR #1488, opened 2018-08-20
bugfix: ZENKO-1018 Disabled status for CRR config

bennettbuchanan:
Do not update the `replicationInfo` property for objects that match the replication config prefix but the rule's status is `Disabled`. This has the effect that the backbeat processor will not attempt CRR on objects that meet this condition.

### DIFF
--- a/lib/api/apiUtils/object/getReplicationInfo.js
+++ b/lib/api/apiUtils/object/getReplicationInfo.js
@@ -71,7 +71,8 @@ function getReplicationInfo(objKey, bucketMD, isMD, objSize, operationType,
     const config = bucketMD.getReplicationConfiguration();
     // If bucket does not have a replication configuration, do not replicate.
     if (config) {
-        const rule = config.rules.find(rule => objKey.startsWith(rule.prefix));
+        const rule = config.rules.find(rule =>
+            (objKey.startsWith(rule.prefix) && rule.enabled));
         if (rule) {
             return _getReplicationInfo(rule, config, content, operationType,
                 objectMD);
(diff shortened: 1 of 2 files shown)


rachedbenmustapha (approved):

JianqinWang (approved):

lib/api/apiUtils/object/getReplicationInfo.js
```diff
@@ -89,7 +89,8 @@ function getReplicationInfo(objKey, bucketMD, isMD, objSize, operationType,
     //   actions triggered by internal services, by design)
```
tmacro: We might want to update the comment to reflect this change. Aside from that LGTM

tmacro (approved):

lib/api/apiUtils/object/getReplicationInfo.js
bennettbuchanan: Good point! Updated.

jonathan-gramain (commented):
Does this fix apply to S3C? If so, can you retarget to 7.4?

bennettbuchanan:
@bert-e help

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

rahulreddy (approved):

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

PR #1523, opened 2018-09-04
Bugfix/zenko 1091 redirect payload

rahulreddy:
- improvement: revert ZENKO-819 regional endpoints
- bugfix: set region endpoint during location setup
This commit removes the current redirect code and adds a setup method
that makes a call to get the region of the destination bucket on AWS S3
and updates the endpoint to use the correct region specific endpoint.
This should avoid redirects from AWS which in turn messes up the payload
causing errors during PUTs. The setup call will be cleaned up to address
the technical debt of waiting for the callback.

### DIFF
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -1,5 +1,6 @@
-const { errors, s3middleware } = require('arsenal');
 const AWS = require('aws-sdk');
+const { errors, s3middleware } = require('arsenal');
+const werelogs = require('werelogs');
 const MD5Sum = s3middleware.MD5Sum;
 const getMetaHeaders = s3middleware.userMetadata.getMetaHeaders;
 const createLogger = require('../multipleBackendLogger');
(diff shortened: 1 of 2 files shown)


rahulreddy:
@bert-e reset

lib/data/locationConstraintParser.js
```diff
@@ -106,6 +106,9 @@ function parseLC() {
             }
             clients[location] = locationObj.type === 'gcp' ?
                 new GcpClient(clientConfig) : new AwsClient(clientConfig);
+            if (locationObj.type === 'aws_s3') {
```
alexanderchan-scality, line 109: should this `if` check also verify that the `aws_s3` location is an actual `aws` location before performing the setup operation?

jonathan-gramain (approved):
Alex may have a valid point (don't know if all AWS-compatible locations would support this).

lib/data/locationConstraintParser.js
rahulreddy: That's true. Let me see what we have.

rahulreddy: I checked the code and if it's non-AWS like DigitalOcean, Wasabi etc, at the best they will return NotImplemented and it will fail silently.

alexanderchan-scality: For `S3 Ring` locations, `cloudserver` does have the `getBucketLocation` implemented and will return a location; wouldn't the setup code overwrite the client with an incorrect endpoint?

lib/data/external/AwsClient.js
```diff
@@ -24,6 +24,37 @@ class AwsClient {
         this._serverSideEncryption = config.serverSideEncryption;
         this._supportsVersioning = config.supportsVersioning;
         this._client = new AWS.S3(this._s3Params);
+        this._logger = new werelogs.Logger('AwsClient');
+    }
+
+    setup(cb) {
+        this._client.getBucketLocation({ Bucket: this._awsBucketName },
```
vrancurel: can you add a comment saying that this call is itself redirected

lib/data/external/AwsClient.js
```diff
@@ -24,6 +24,37 @@ class AwsClient {
...
+        (err, res) => {
+            if (err && err.code !== 'AuthorizationHeaderMalformed') {
```
vrancurel: did you check the null ?

lib/data/locationConstraintParser.js
rahulreddy: Ah yeah you are right, since we do implement locations. I could probably check if endpoint ends with  'amazonaws.com' and fix it.

lib/data/external/AwsClient.js
rahulreddy: You mean if err is null? That condition should be taken care of.

lib/data/external/AwsClient.js
rahulreddy: Done.

rahulreddy:
@bert-e reset

alexanderchan-scality (approved):

PR #1534, opened 2018-09-07
bugfix: filter NoSuchKey errors when deleting on AWS/GCP

rahulreddy:
This accounts for 404s when trying to delete something that doesn't exist
or hasn't been replicated yet. It returns no error to the client as the delete
is a no-op and is inline with S3 API compatibility.
The test has been updated as well to match the behavior change.

### DIFF
--- a/lib/data/external/AwsClient.js
+++ b/lib/data/external/AwsClient.js
@@ -200,7 +200,7 @@ class AwsClient {
             if (err) {
                 logHelper(log, 'error', 'error deleting object from ' +
                 'datastore', err, this._dataStoreName, this.clientType);
-                if (err.code === 'NoSuchVersion') {
+                if (err.code === 'NoSuchVersion' || err.code === 'NoSuchKey') {
                     // data may have been deleted directly from the AWS backend
                     // don't want to retry the delete and errors are not
                     // sent back to client anyway, so no need to return err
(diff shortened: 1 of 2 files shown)


JianqinWang (approved):

tmacro (approved):

bennettbuchanan (approved):
Nice catch!

tests/functional/raw-node/test/GCP/object/delete.js
```diff
@@ -83,14 +83,13 @@ describe('GCP: DELETE Object', function testSuite() {
     });
 
     describe('without existing object in bucket', () => {
-        it('should return 404 and NoSuchKey', done => {
```
alexanderchan-scality: The `GCP raw-node` tests are for testing the `gcp client`'s implementation of making request to/handling responses from `Google Cloud`. It does not go through `Cloudserver`, so I think this should stay as is.

rahulreddy: Ah. I remember them now. Do we have tests where GCP goes through the AWSClient?

dora-korpar (approved):

tests/functional/raw-node/test/GCP/object/delete.js
alexanderchan-scality: There's  [multipleBackend/delete/deleteGcp.js](https://github.com/scality/cloudserver/blob/development/8.0/tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteGcp.js)...if that's what you're asking for.

jonathan-gramain (commented):
No change needed on GcpClient?

rahulreddy:
@jonathan-gramain GCPClient is not overriding this method.

rahulreddy:
@bert-e reset

jonathan-gramain (approved):

PR #1542, opened 2018-09-13
Feature/zenko 714 s3 management code2

jeremyds:
# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/management/agentClient.js
+++ b/lib/management/agentClient.js
@@ -1,14 +1,94 @@
+const WebSocket = require('ws');
+const arsenal = require('arsenal');
+
+const logger = require('../utilities/logger');
+const _config = require('../Config').config;
+const { patchConfiguration } = require('./configuration');
+const { reshapeExceptionError } = arsenal.errorUtils;
+
+
 const managementAgentMessageType = {
     /** Message that contains the loaded overlay */
     NEW_OVERLAY: 1,
 };
 
+const CONNECTION_RETRY_TIMEOUT_MS = 5000;
+
+
+function initManagementClient() {
+    const { host, port } = _config.managementAgent;
+
+    const ws = new WebSocket(`ws://${host}:${port}/watch`);
+
+    ws.on('open', () => {
+        logger.info('connected with management agent');
+    });
+
+    ws.on('close', (code, reason) => {
+        logger.info('disconnected from management agent', { reason });
+        setTimeout(initManagementClient, CONNECTION_RETRY_TIMEOUT_MS);
+    });
+
+    ws.on('error', error => {
+        logger.error('error on connection with management agent', { error });
+    });
+
+    ws.on('message', data => {
+        const method = 'initManagementclient::onMessage';
+        const log = logger.newRequestLogger();
+        let msg;
+
+        if (!data) {
+            log.error('message without data', { method });
+            return;
+        }
+        try {
+            msg = JSON.parse(data);
+        } catch (err) {
+            log.error('data is an invalid json', { method, err, data });
+            return;
+        }
+
+        if (msg.payload === undefined) {
+            log.error('message without payload', { method });
+            return;
+        }
+        if (typeof msg.messageType !== 'number') {
+            log.error('messageType is not an integer', {
+                type: typeof msg.messageType,
+                method,
+            });
+            return;
+        }
+
+        switch (msg.messageType) {
+        case managementAgentMessageType.NEW_OVERLAY:
+            patchConfiguration(msg.payload, log, err => {
+                if (err) {
+                    log.error('failed to patch overlay', {
+                        error: reshapeExceptionError(err),
+                        method,
+                    });
+                }
+            });
+            return;
+        default:
+            log.error('new overlay message with unmanaged message type', {
+                method,
+                type: msg.mmessageType,
+            });
+            return;
+        }
+    });
+}
 
 function isManagementAgentUsed() {
     return process.env.MANAGEMENT_USE_AGENT === '1';
 }
 
+
 module.exports = {
     managementAgentMessageType,
+    initManagementClient,
     isManagementAgentUsed,
 };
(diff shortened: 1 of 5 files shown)


lib/server.js
```diff
@@ -149,11 +153,15 @@ class S3Server {
 
         // TODO this should wait for metadata healthcheck to be ok
         // TODO only do this in cluster master
-        setTimeout(() => {
-            if (enableRemoteManagement) {
-                initManagement(logger.newRequestLogger());
-            }
-        }, 5000);
+        if (!isManagementAgentUsed()) {
```
rachedbenmustapha: The management agent client shouldn't be initialized either if we're disabling remote management altogether

lib/management/agentClient.js
```diff
@@ -1,14 +1,69 @@
...
+        if (typeof msg.messageType !== 'number') {
+            log.error('messageType is not an integer', {
+                type: typeof msg.messageType,
+            });
+            return;
+        }
+
+        switch (msg.messageType) {
+        case managementAgentMessageType.NEW_OVERLAY:
+            applyAndSaveOverlay(msg.payload, log);
```
rachedbenmustapha: In this particular case I don't think the agent client (cloudserver process) should be saving the overlay - it should just patch its configuration - and the overlay should be saved by the agent when it receives it, before broadcasting.

rachedbenmustapha (changes requested):

jeremyds:
@bert-e status

jeremyds:
@bert-e help

jeremyds:
@bert-e after_pull_request=1568

lib/server.js
jeremyds: done

lib/management/agentClient.js
```diff
@@ -1,14 +1,71 @@
...
+            log.error('messageType is not an integer', {
+                type: typeof msg.messageType,
+            });
+            return;
+        }
+
+        switch (msg.messageType) {
+        case managementAgentMessageType.NEW_OVERLAY:
+            patchConfiguration(msg.payload, log, err => {
+                if (err) { log.error('failed to patch overlay', { err }); }
```
rachedbenmustapha: field name in the logs should be `error` and probably needs to be reshaped (other users of `patchConfiguration` do reshape its returned error)

lib/management/agentClient.js
```diff
@@ -1,14 +1,71 @@
...
+            });
+            break;
```
rachedbenmustapha: Since the code has branched asynchronously here I suggest `return` instead of `break` to avoid surprises if anyone adds code at the end of the function later.

rachedbenmustapha (changes requested):

lib/management/agentClient.js
jeremyds: you re right, done

rachedbenmustapha (approved):

lib/management/agentClient.js
```diff
@@ -1,14 +1,78 @@
...
+    ws.on('error', error => {
+        logger.error('error on connection with management agent', { error });
+    });
+
+    ws.on('message', data => {
+        const log = logger.newRequestLogger();
+        const msg = JSON.parse(data);
+
+        if (msg.payload === undefined) {
+            log.error('message without payload');
```
nicolas2bert: You might want to add the `method` when the error occurs to ease the debugging: Exemple: `log.error('message without payload', { method: 'initManagementClient'});`

lib/management/agentClient.js
```diff
@@ -1,14 +1,78 @@
...
+            return;
+        }
+        if (typeof msg.messageType !== 'number') {
+            log.error('messageType is not an integer', {
+                type: typeof msg.messageType,
```
nicolas2bert, line 58: Same: `{ method: ... }`

lib/management/agentClient.js
```diff
@@ -1,14 +1,78 @@
...
+        setTimeout(initManagementClient, CONNECTION_RETRY_TIMEOUT_MS);
+    });
+
+    ws.on('error', error => {
+        logger.error('error on connection with management agent', { error });
+    });
+
+    ws.on('message', data => {
+        const log = logger.newRequestLogger();
+        const msg = JSON.parse(data);
```
nicolas2bert: Can `data` be `undefined` or an invalid JSON? If so, it will throw an error.

lib/management/agentClient.js
```diff
@@ -1,14 +1,78 @@
...
 const managementAgentMessageType = {
     /** Message that contains the loaded overlay */
     NEW_OVERLAY: 1,
 };
 
+const CONNECTION_RETRY_TIMEOUT_MS = 5000;
+
+
+function initManagementClient() {
+    const host = _config.managementAgent.host;
```
nicolas2bert: **Tip**: You might already know, ES6 allows you to do "object destructuring" : `const { host, port } = _config.managementAgent;`

lib/management/agentClient.js
```diff
@@ -1,14 +1,78 @@
...
+            patchConfiguration(msg.payload, log, err => {
+                if (err) {
+                    log.error('failed to patch overlay', {
+                        error: reshapeExceptionError(err),
+                    });
+                }
+            });
+            return;
+        default:
+            log.error('new overlay message version without payload');
```
nicolas2bert: new overlay message version without **message type**?

lib/management/agentClient.js
jeremyds: done

lib/management/agentClient.js
jeremyds: done

lib/management/agentClient.js
jeremyds: non specified on websocket documentation. It is based on MessageEvent which can be instanciate without data, set to null in this case. So you re rigjt, I should check data.

lib/management/agentClient.js
jeremyds: more elegant, done

lib/management/agentClient.js
jeremyds: stupid copy and paste :S

nicolas2bert (approved):

PR #1543, opened 2018-09-14
bf: ZENKO-1146 read from source until replicated to preferred read

jonathan-gramain:
When a preferred read location is defined for an object, and when
status of replication to this location is PENDING or FAILED, read
object from the source location instead of returning an error.

Note that reads are not guaranteed to succeed if the source location
is a transient source, because of a race condition with the garbage
collector. This will be addressed in a future change.

### DIFF
--- a/lib/api/objectGet.js
+++ b/lib/api/objectGet.js
@@ -156,12 +156,22 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
                     targetLocation, objMD.replicationInfo);
                 if (repBackendResult.error) {
                     log.error('Error with location constraint header', {
+                        bucketName, objectKey, versionId,
                         error: repBackendResult.error,
                         status: repBackendResult.status,
                     });
                     return callback(repBackendResult.error, null, corsHeaders);
                 }
-                dataLocator = repBackendResult.dataLocator;
+                const targetDataLocator = repBackendResult.dataLocator;
+                if (targetDataLocator) {
+                    dataLocator = targetDataLocator;
+                } else {
+                    log.debug('using source location as preferred read ' +
+                              'is unavailable', {
+                                  bucketName, objectKey, versionId,
+                                  reason: repBackendResult.reason,
+                              });
+                }
             }
             // if the data backend is azure, there will only ever be at
             // most one item in the dataLocator array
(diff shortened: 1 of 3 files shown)


lib/api/objectGet.js
```diff
@@ -156,12 +156,21 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
...
                         status: repBackendResult.status,
                     });
                     return callback(repBackendResult.error, null, corsHeaders);
                 }
-                dataLocator = repBackendResult.dataLocator;
+                const targetDataLocator = repBackendResult.dataLocator;
+                if (targetDataLocator) {
+                    dataLocator = targetDataLocator;
+                } else {
+                    log.info('Doing GET on source location', {
```
rahulreddy: This can be a `debug` level log and message could be `using source location as preferred read 
 is unavailable`, it would be more contextual.

tests/unit/multipleBackend/getReplicationBackendDataLocator.js
```diff
@@ -32,22 +32,25 @@ describe('Replication Backend Compare', () => {
...
-    it('should return error if backend status is PENDING', () => {
+    it('should return a status and reason if backend status is PENDING', () => {
         const repBackendResult =
             getReplicationBackendDataLocator(locCheckResult, repMatchPending);
-        assert(repBackendResult.error.NoSuchKey);
+        assert.strictEqual(repBackendResult.dataLocator, undefined);
         assert.strictEqual(repBackendResult.status, 'PENDING');
+        assert.notStrictEqual(repBackendResult.reason, undefined);
     });
     it('should return error if backend status is FAILED', () => {
```
rahulreddy: same as the previous one `return a status and reason `

rahulreddy (changes requested):
logic LGTM

lib/api/objectGet.js
jonathan-gramain: I had though it would be useful to have an info log because of the possible "race condition on read" known issue, it may help tracking those issues. But checking the request timestamps in cloudserver and backbeat should be enough to identify if a read error was caused by this race, so I can make this a debug log.

jonathan-gramain:
@bert-e reset

rahulreddy (approved):

jonathan-gramain:
@bert-e reset

ssalaues (approved):

JianqinWang (approved):

philipyoo:
lol too fast, but lgtm 👍

PR #1553, opened 2018-09-20
fs type location

vrancurel:
## Description
This PR add a new type of location called "fs".

### Motivation and context
Recently some Zenko users have expressed the need of being able to read
local FS objects from Zenko, as "symlinks" (see https://forum.zenko.io/t/using-existing-file-data/355/5).

Also we want to be able to support ingestion of metadata from existing filesystems (e.g. NFS).

### Why is this change required? What problem does it solve?

This change allows a location to put 0-byte objects with consistent size and MD5. The files
can be re-read in a pass-through data backend. 

The new "fs" type location supports a mountPath configuration that typically points on a
local folder or mount point that contains the file (e.g. a Kubernetes NFS PV).

Note that the data backend does not try to list or read the metadata from the mountPath,
because:
- The hierarchical filesystem structure is not compatible with the prefix and delimiter type of queries of a filesystem.
- File system entries are not lexicographically ordered.

As a result, the listing will be slow and impracticable for large filesystems, and not necessarily strictly compatible with S3.

But also:
- The MD5 of files is not available in file systems (but needed for doing S3 read requests). It would be too intense to compute at e.g. every HEAD query.

Instead it relies on a external tool, called periodically that checks for updates in the filesystem (see https://github.com/scality/rclone/pull/1).

The modified rclone tool make sure that the whole structure of the FS is properly mirrored in an S3 bucket. De facto being 100% compatible with all S3 LIST requests. Moreover the MD5 and size are properly set when creating the 0-byte file, therefore properly serving HEAD and GET S3 requests.

The mirroring is done in a way that existing entries are not populated twice. The tool can also track ctime changes and mirror attribute changes in the local filesystem.

### Caveats
With this design, changes are not mirrored in real-time. Anyways, being notified of real-time changes e.g. via inotify() is not possible on NFS. The alternative is to have access the the operation log of the filesystem which is not always the case.

Need to add tests.

### Code conforms with the [style guide]

(https://github.com/scality/Guidelines/blob/master/CONTRIBUTING.md#coding-style-guidelines)

### DIFF
--- a/lib/data/external/FsClient.js
+++ b/lib/data/external/FsClient.js
@@ -1,0 +1,156 @@
+const { errors, s3middleware } = require('arsenal');
+const werelogs = require('werelogs');
+const MD5Sum = s3middleware.MD5Sum;
+const getMetaHeaders = s3middleware.userMetadata.getMetaHeaders;
+const createLogger = require('../multipleBackendLogger');
+const { logHelper } = require('./utils');
+const { config } = require('../../Config');
+const fs = require('fs');
+
+class FsClient {
+    constructor(config) {
+        this.clientType = 'fs';
+        this.type = 'FS';
+        this._bucketName = config.bucketName;
+        this._bucketMatch = config.bucketMatch;
+        this._serverSideEncryption = config.serverSideEncryption;
+        this._dataStoreName = config.dataStoreName;
+        this._supportsVersioning = config.supportsVersioning;
+        this._mountPath = config.mountPath;
+        this._logger = new werelogs.Logger('FsClient');
+    }
+
+    setup(cb) {
+        return cb();
+    }
+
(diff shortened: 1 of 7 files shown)


lib/data/external/FsClient.js
```diff
@@ -0,0 +1,156 @@
...
+        return {
+            key: this._createFsKey(bucketName, objectKey, this._bucketMatch),
+            dataStoreName: this._dataStoreName,
+        };
+    }
+
+    put(stream, size, keyContext, reqUids, callback) {
+        const log = createLogger(reqUids);
+
+        if (size === 0) {
```
vrancurel, line 45: This could be replaced by a special header.

giacope:
Closing this since #1609 was merged.

PR #1559, opened 2018-09-21
Feature/s3 c 1499 apiv2

dora-korpar:
Implements v2 of the GET Bucket API (https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html).
Check out the design doc for the main differences between v1 & v2. 

Depends on https://github.com/scality/Arsenal/pull/574

### DIFF
--- a/docs/GET_BUCKET_V2.md
+++ b/docs/GET_BUCKET_V2.md
@@ -1,0 +1,69 @@
+# Get Bucket Version 2 Documentation
+
+## Description
+
+This feature implements version 2 of the GET Bucket (List Objects)
+operation, following AWS specifications
+(see https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html).
+
+## Requirements
+
+The user must have READ access to the bucket.
+
+## Design
+
+### Request
+
+The `delimiter`, `encoding-type`, `max-keys`, and `prefix` request parameters
+from GET Bucket v1 remain unchanged.
+In order to specify v2, the parameter `list-type` must be included and
+set to `2`.
+The `marker` v1 parameter's functionality has been split in two and replaced by
+`start-after` and `continuation-token` in v2. The `start-after` parameter is
+a specific object key after which the API will return key names. It is only
+valid in the first GET request. If both the `start-after` and
+`continuation-token` parameters are included in a request, the API will
+ignore the `start-after` parameter in favor of the `continuation-token`.
+If the GET Bucket v2 response is truncated, a `NextContinuationToken` will
+also be included. To list the next set of objects, the `NextContinuationToken`
+can be used as the `continuation-token` in the next request. The continuation
+token is an obfuscated string of 57 characters that CloudServer understands and
+interprets.
+By default, the v2 response does not include object owner information. To
+include owner information like the default v1 response, use the `fetch-owner`
+request parameter set to `true`.
+
+### Response
+
+The GET Bucket v1 and v2 responses are largely the same, with only a few changes.
+The `NextMarker` v1 parameter has been replaced by the
+`NextContinuationToken`. The `NextContinuationToken` is included with any
+truncated response, even if no delimiter is sent in the request. Its value is an
+obfuscated string that can be passed at the `continuation-token` in the next
+request, which will be interpreted by CloudServer.
+The `KeyCounter` parameter is returned in every response. Its value is the
+number of keys included in the response. It is always less than or equal to
+the `MaxKeys` value.
+If the `start-after` or `continuation-token` parameter is used in the
+request, it is also included in the response.
+By default, the v2 response does not include object owner information, unlike
+the v1 response. See the `Request` section for including it.
+
+### Continuation Token
+
+An example continuation token: 
+
+```
+NextContinuationToken: '1bunC4s+crlZNAAbKUGBLyajJUQKp22TOdUR6/01snxD2cZtjJD0ugA=='
+```
+
+In order to generate a comparable token, CloudServer uses base64 encoding to
+obfuscate the key name of the next object to be listed.
+Encoded continuation tokens are similarly decoded in order for listing to
+continue from the correct object.
+
+## Performing Get Bucket V2 Operation
+
+When performing the GET Bucket V2 operation, if the request is built manually,
+the parameter `list-type` must be included and set to `2`.
+Using the AWS cli client, the command becomes `list-objects-v2`.
--- a/lib/api/apiUtils/object/continueToken.js
+++ b/lib/api/apiUtils/object/continueToken.js
@@ -1,0 +1,28 @@
+/**
+ * generateToken - generates obfuscated continue token from object keyName
+ * @param {String} keyName - name of key to obfuscate
+ * @return {String} - obfuscated continue token
+ */
+function generateToken(keyName) {
+    if (keyName === '' || keyName === undefined) {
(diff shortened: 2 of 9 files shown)


docs/GET_BUCKET_V2.md
```diff
@@ -0,0 +1,69 @@
...
+### Continuation Token
+
+An example continuation token: 
+
+```
+NextContinuationToken: '1bunC4s+crlZNAAbKUGBLyajJUQKp22TOdUR6/01snxD2cZtjJD0ugA=='
+```
+
+In order to generate a comparable token, CloudServer uses base64 encoding to
+obfuscate the key name of the next object to be listed.
```
GiorgioRegni, line 61: base64 is not really obfuscation, more of a translation. Is it OK or do we need to really obfuscate/secure this token? Can it be used for an attack?

dora-korpar: AWS doesn't provide any documentation as to why the continuation token is obfuscated, but as far as I can tell, its only purpose is to avoid returning a key name that wasn't explicitly requested, so I think our method of 'translating' is fine :) @vrancurel

GiorgioRegni: For sure they don't explain anything about the token but it looks encrypted, @vrancurel you confirm that this is not an attack vector for the system?

lib/api/apiUtils/object/continueToken.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * generateToken - generates obfuscated continue token from object keyName
+ * @param {String} keyName - name of key to obfuscate
+ * @return {String} - obfuscated continue token
+ */
+function generateToken(keyName) {
+    if (keyName === '' || keyName === undefined) {
+        return undefined;
+    }
+    return Buffer.from(keyName).toString('base64');
```
JianqinWang, line 10: if we pass an empty string to `Buffer.from(keyName).toString('baset64')`, it will be able to return an empty string. On that note, if lines 8 and 20 in this file return `''` instead of `undefined`, line 182 (and line 192? if you had originally intended to add `|| ''`) in `lib/api/bucketGet.js` will not need the `||` conditional.

lib/api/bucketGet.js
```diff
@@ -154,13 +173,28 @@ function processMasterVersions(bucketName, listParams, list) {
...
+    if (listParams.v2) {
+        xmlParams.push(
+            { tag: 'StartAfter', value: listParams.startAfter || '' });
+        xmlParams.push(
+            { tag: 'FetchOwner', value: `${listParams.fetchOwner}` });
+        xmlParams.push({
+            tag: 'ContinuationToken',
+            value: generateToken(listParams.continuationToken) || '',
+        });
+        xmlParams.push({
```
JianqinWang, line 191: Is the response always guaranteed to have a `NextContinuationToken`? In the sample response on the AWS docs, it looks like non-truncated responses do not have this key-value pair.

docs/GET_BUCKET_V2.md
vrancurel: It is not clear why they obfuscate the token. They probably use a mapping table to have a fixed size token. We decided that for a v1 we wouldn't do that because it requires a DB etc.
I don't see why obfuscation is more secure than NextMarker, since it is possible to open hundred of cursors with the start-after parameter but they may have a reason.
In our case this is strictly equivalent to the listing v1 in term of security.

dora-korpar:
@bert-e reset

lib/api/bucketGet.js
dora-korpar: Yup, that's right. It won't be included in the response unless it's present because of line 209

dora-korpar:
@bert-e reset

docs/GET_BUCKET_V2.md
```diff
@@ -0,0 +1,69 @@
...
+The GET Bucket v1 and v2 responses are largely the same, with only a few changes.
+The `NextMarker` v1 parameter has been replaced by the
+`NextContinuationToken`. The `NextContinuationToken` is included with any
+truncated response, even if no delimiter is sent in the request. Its value is an
+obfuscated string that can be passed at the `continuation-token` in the next
+request, which will be interpreted by CloudServer.
+The `KeyCounter` parameter is returned in every response. Its value is the
+number of keys included in the response. It is always less than or equal to
+the `MaxKeys` value.
+If the `starts-after` or `continuation-token` parameter is used in the
```
bennettbuchanan: ~starts-after~ start-after

dora-korpar:
@bert-e reset

lib/api/bucketGet.js
```diff
@@ -241,9 +301,17 @@ function bucketGet(authInfo, request, log, callback) {
         listingType: 'DelimiterMaster',
         maxKeys: actualMaxKeys,
         delimiter: params.delimiter,
-        marker: params.marker,
         prefix: params.prefix,
     };
+    if (v2) {
```
bennettbuchanan, line 310: Based on the docs, perhaps the condition should be that `v2` equals `2` rather than any truthy value.

lib/metadata/in_memory/backend.js
```diff
@@ -233,6 +240,14 @@ const metastore = {
                 return cb(errors.InvalidArgument);
             }
 
+            if (continuationToken && typeof continuationToken !== 'string') {
+                return cb(errors.InvalidArgument);
```
bennettbuchanan, line 244: Could be nice to add a customized description about these argument checks. I guess this depends on AWS behavior though.

dora-korpar: Yeah, AWS just returns the standard 'InvalidArgument' error. And it keeps in line with the other checks here to return the standard

lib/api/bucketGet.js
dora-korpar: I'm using the `v2=true` value to make internal semantics easier to follow. The api still requires `list-type` to be set to `2`

bennettbuchanan: I do follow that, I mean that this `if` block condition could check `list-type` is set to `2` rather than any truthy value of `list-type`. That way, for example, a user couldn't set `list-type` to `1` and then get the v2 listing.

The next line which sets `listParams.v2 = true;` is all good for me 😄.

dora-korpar: ooh gotcha, yeah that's actually a good point

dora-korpar:
@bert-e wait

dora-korpar:
@bert-e reset

dora-korpar:
The pre-merge build is failing due to an unknown bug in a small subset of V2 tests wherein the listObjectsV2 S3 API call is misinterpreted as listObjectVersions, potentially due to an Eve and 7.4 mismatch. Extensive manual testing in both local and RING environments, along with pre-merge builds in branches based off development/8.0 and new end to end tests of the feature in Zenko all pass.

lib/api/bucketGet.js
dora-korpar: I updated this by including a check above that v2 is actually 2

bennettbuchanan: Nice, LGTM.

bennettbuchanan (approved):

JianqinWang (approved):

rahulreddy:
@bert-e wait

dora-korpar:
@bert-e reset

PR #1563, opened 2018-09-25
Feature/zenko 716 management process

jeremyds:
# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.


jeremyds:
@bert-e help

jeremyds:
@bert-e after_pull_request=1542

tests/functional/managementAgent/managementProcess.js
```diff
@@ -0,0 +1,368 @@
...
+             * request on the stats path. */
+            const url = URL.parse(process.env.CI_STAT_REPORT_URL);
+            const statServer = http.createServer((request, response) => {
+                if (request.url !== url.pathname) {
+                    response.writeHead(400, { 'Content-type': 'text/plan' });
+                    response.write('bad path');
+                    response.end();
+                    return;
+                }
+                response.writeHead(200, { 'Content-type': 'text/plan' });
```
jeremyds: plain

tests/functional/managementAgent/managementProcess.js
```diff
@@ -0,0 +1,368 @@
...
+                        firstMsgReceived = true;
+                        /* Send a new overlay to management process. */
+                        orbitSimulator.sendMessage(CONFIG_OVERLAY_MSG_TYPE,
+                                                   body);
+                        return;
+                    }
+
+                    /* Check we receive the second overlay. */
+                    assert.strictEqual(msg.payload.toString(), body);
+
```
jeremyds: ws.terminate

tests/functional/managementAgent/managementProcess.js
```diff
@@ -0,0 +1,368 @@
...
+               /* Ugly eventTarget internal fields hacking to avoid this web
+                * socket to answer to ping messages. It will make
+                * the management agent to close the connection after a timeout.
+                * Defining an onPing event does not help, this internal
+                * function is still called. */
+               ws._receiver._events.ping = function noop() {};
+           });
+       });
+
+    describe('with orbit simulator', () => {
```
jeremyds: not usefull anymore to have a describe specific to tests with orbit. It was done because I wanted to use before/after Each hooks but finally did not, so let's get an indentation level back.

jeremyds:
I deactivated file eve stage to test mine, I am facing ressource issue (CPU and then RAM) and before to handle this problem I want to make my stage and test suite run.
tl;drL DO NOT MERGE

lib/management/push.js
```diff
@@ -92,7 +89,9 @@ function startWSManagementClient(pushEndpoint, instanceId, token) {
     };
     const agent = createWSAgent(pushEndpoint, process.env, logger);
 
-    const url = `${pushEndpoint}/${instanceId}/ws`;
+    const url = process.env.CI_ORBIT_PUSH_ENDPOINT ||
```
rachedbenmustapha: Since the push endpoint is already configurable, this could follow the same pattern as for the API endpoint, instead of having a new CI-related env var.

lib/management/constants.js
```diff
@@ -0,0 +1,15 @@
+/* From RFC6455 which defines some reserved status code ranges, the range
+ * 4000-4999 are for private use. */
+const WS_STATUS_IDDLE = {
```
rachedbenmustapha: Typo: IDLE?

package.json
```diff
@@ -89,12 +89,13 @@
...
     "mem_backend": "S3BACKEND=mem node index.js",
     "perf": "mocha tests/performance/s3standard.js",
-    "start": "npm-run-all --parallel start_dmd start_s3server",
+    "start": "npm-run-all --parallel start_dmd start_management start_s3server",
     "start_mongo": "npm run cloudserver",
     "start_mdserver": "node mdserver.js",
     "start_dataserver": "node dataserver.js",
     "start_s3server": "node index.js",
     "start_dmd": "npm-run-all --parallel start_mdserver start_dataserver",
+    "start_management": "node managementAgent.js",
```
rachedbenmustapha, line 98: @ssalaues when this PR is merged (and a new docker image built) the cloudserver-manager pod will need to use this command line

rachedbenmustapha (changes requested):
One general comment that applies to the new env vars: they are all legitimate tunables so I don't think they need to have the `CI_` prefix. For consistency with `PUSH_ENDPOINT` that already exists I suggest the `PUSH_` prefix.

One question: is the agent able to work with `poll` mode as well? ie if it's configured to use `poll` instead of `push`, will it query a new overlay every minute and push the result to its connected cloudservers?

jeremyds:
@RachedBenMustapha:
> One question: is the agent able to work with poll mode as well? ie if it's configured to use poll instead of push, will it query a new overlay every minute and push the result to its connected cloudservers?

No. I did not handle the poll mode, I thought it was "unused" (the previous mode). I don't see a big work to make it works, I just need to call the agent listener callback.

rachedbenmustapha:
@jeremyds ok, if we switch off the agent-less management mode then the agent should pick up the duty of initiating poll mode (like startManagementListeners does).

It's not enabled by default and event discouraged, but it's still important to support as the kube deployment has configuration specific to this mode.

lib/management/constants.js
jeremyds: done

lib/management/push.js
jeremyds: Which endpoint are you speaking about? The one used here (line 93) is not configurable, it is hardocded to 'https://push.api.zenko.io/api/v1/instance' in lib/management/index.js
That being said I don't know if there is other use case than tests to change that, the PUSH endpoint is not supposed to change, is it?

rachedbenmustapha: There is a separate `PUSH_ENDPOINT` env var that is used and is configurable via the helm chart values.

It is different for dev, staging, production and laptop. It's even planned to have one per-customer for specific needs.

jeremyds:
@RachedBenMustapha: 

> For consistency with PUSH_ENDPOINT that already exists I suggest the PUSH_ prefix.

I don't get this one :S, which env variable are you speaking about? maybe ORBIT_PUSH_ENDPOINT or
 .MANAGEMENT_ENDPOINT?

rachedbenmustapha:
it's just called `PUSH_ENDPOINT` and is in `lib/management/index.js` (maybe it has disappeared in 8.1? if that's the case it needs to come back)

jeremyds:
> if we switch off the agent-less management mode then the agent should pick up the duty of initiating poll mode (like startManagementListeners does).

It's not enabled by default and event discouraged, but it's still important to support as the kube deployment has configuration specific to this mode.


The current code initailize the POLL management when the appropriate env variable is set. But in that case the new overlays received in poll mode don't trigger the agent listener callback and so, are not broadcasted to agent clients (however they are saved..... and patched). I did not handle at all the POLL mode

rachedbenmustapha:
Just verified about `PUSH_ENDPOINT`, this PR's commit "ft: ZENKO-716: management, use constant and env variables for tests" removed it, it needs to come back to keep compatibility and not upset the whole team who are using it 😆

lib/management/index.js
```diff
@@ -13,19 +13,16 @@ const { initManagementCredentials } = require('./credentials');
...
 
+const endpointPath = 'api/v1/instance';
 const managementEndpointRoot =
     process.env.MANAGEMENT_ENDPOINT ||
     'https://api.zenko.io';
-const managementEndpoint = `${managementEndpointRoot}/api/v1/instance`;
+const managementEndpoint = `${managementEndpointRoot}/${endpointPath}`;
 
-const pushEndpointRoot =
-    process.env.PUSH_ENDPOINT ||
```
jeremyds, line 26: @RachedBenMustapha here is the culprit

jeremyds: you we're faster to find it :(

jeremyds:
> Just verified about `PUSH_ENDPOINT`, this PR's commit "ft: ZENKO-716: management, use constant and env variables for tests" removed it, it needs to come back to keep compatibility and not upset the whole team who are using it

It is to be sure they update their working version 8-)

lib/management/push.js
```diff
@@ -86,13 +84,17 @@ function startWSManagementClient(pushEndpoint, instanceId, token) {
...
+    overlayMessageListener = newOverlayCallback;
+
     const socketsByChannelId = [];
     const headers = {
         'x-instance-authentication-token': token,
     };
     const agent = createWSAgent(pushEndpoint, process.env, logger);
 
-    const url = `${pushEndpoint}/${instanceId}/ws`;
+    const url = process.env.ORBIT_PUSH_ENDPOINT ||
```
rachedbenmustapha: The env var that already exists and is used by kube deployment is called `PUSH_ENDPOINT`

rachedbenmustapha (changes requested):

lib/management/push.js
jeremyds: Done

lib/management/index.js
```diff
@@ -13,19 +13,16 @@ const { initManagementCredentials } = require('./credentials');
...
 const initRemoteManagementRetryDelay = 10000;
 
+const endpointPath = 'api/v1/instance';
 const managementEndpointRoot =
     process.env.MANAGEMENT_ENDPOINT ||
     'https://api.zenko.io';
-const managementEndpoint = `${managementEndpointRoot}/api/v1/instance`;
+const managementEndpoint = `${managementEndpointRoot}/${endpointPath}`;
 
-const pushEndpointRoot =
```
rachedbenmustapha, line 25: I did not see this when reviewing just the last 2 commits: I meant to keep this here as `PUSH_ENDPOINT` instead of moving to a `ORBIT_PUSH_ENDPOINT` lookup in `startWSManagementClient` and hardcoding this one here.

So to be clear the code here should read:
```js
const pushEndpointRoot = process.env.PUSH_ENDPOINT || 'https://push.api.zenko.io';
const pushEndpoint = `${pushEndpointRoot}/api/v1/instance`;
```

lib/management/push.js
```diff
@@ -86,13 +84,17 @@ function startWSManagementClient(pushEndpoint, instanceId, token) {
...
+    overlayMessageListener = newOverlayCallback;
+
     const socketsByChannelId = [];
     const headers = {
         'x-instance-authentication-token': token,
     };
     const agent = createWSAgent(pushEndpoint, process.env, logger);
 
-    const url = `${pushEndpoint}/${instanceId}/ws`;
+    const url = process.env.PUSH_ENDPOINT ||
```
rachedbenmustapha: `${pushEndpoint}/${instanceId}/ws` should be used here as it has been correctly computed, and connecting the websocket client to `process.env.PUSH_ENDPOINT` directly will not work (it's missing paths).

rachedbenmustapha (dismissed):

lib/management/index.js
jeremyds: done

lib/management/push.js
jeremyds: done

lib/management/index.js
```diff
@@ -66,36 +57,39 @@ function startManagementListeners(instanceId, token) {
...
+                              envVarValue: initialID,
+                              metadataValue: instanceId,
+                          });
+                return callback(errors.InvalidParameterValue);
+            }
+            return cb(null, instanceId);
+        },
+        (instanceId, cb) => initManagementCredentials(instanceId, log, cb),
         (instanceId, token, cb) => {
             if (!isManagementAgentUsed()) {
```
jeremyds: should be removed, error while resolving conflict after rebasing on 8.1

nicolas2bert (approved):

PR #1574, opened 2018-10-05
cleanup: Remove duplicated code

vrancurel:
Bucketclient and bucketfile seem to be identical to Arsenal

There was a fix done in Cloudserver's bucketclient more recently but I think it is not necessary to report it to Arsenal. Please review carefully.


bennettbuchanan (approved):

JianqinWang:
@bert-e wait

lib/metadata/wrapper.js
```diff
@@ -2,7 +2,6 @@ const MetadataWrapper = require('arsenal').storage.metadata.MetadataWrapper;
 const { config } = require('../Config');
 const logger = require('../utilities/logger');
 const constants = require('../../constants');
-const bucketclient = require('bucketclient');
```
JianqinWang: From my pov, this is still needed :/ the MetadataWrapper is expecting this as a param

JianqinWang (changes requested):

vrancurel:
I have to modify Arsenal for including bucketclient. Was there a reason not do it @JianqinWang ?

lib/metadata/wrapper.js
vrancurel: question: why ?

JianqinWang:
@vrancurel I think it was because bucketclient has an Arsenal dependency, and @jonathan-gramain or @rahulreddy warned me that having that kind of circular dependency is not good

vrancurel:
There was a name clash between the local 'bucketclient' folder that I removed now and the 'bucketclient' defined in package.json. Please check that everything is correct now.

JianqinWang (approved):

rahulreddy:
@bert-e reset

PR #1581, opened 2018-10-18
ZENKO-833: include ingestion option for bucket creation

JianqinWang:
Dependent on: https://github.com/scality/Arsenal/pull/588

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -138,6 +138,8 @@ const constants = {
     base64Regex: new RegExp('^(?:[A-Za-z0-9+/]{4})*' +
         '(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$'),
     productName: 'APN/1.0 Scality/1.0 Scality CloudServer for Zenko',
+    // location constraint delimiter
+    zenkoSeparator: ':',
 };
 
 module.exports = constants;
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -13,10 +13,11 @@ const { isBackbeatUser } = require('../authorization/aclChecks');
 
 const usersBucket = constants.usersBucket;
 const oldUsersBucket = constants.oldUsersBucket;
+const zenkoSeparator = constants.zenkoSeparator;
 const userBucketOwner = 'admin';
 
 
-function addToUsersBucket(canonicalID, bucketName, log, cb) {
+function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
     // BACKWARD: Simplify once do not have to deal with old
     // usersbucket name and old splitter
 
@@ -28,7 +29,10 @@ function addToUsersBucket(canonicalID, bucketName, log, cb) {
         const splitter = usersBucketAttrs ?
             constants.splitter : constants.oldSplitter;
         let key = createKeyForUserBucket(canonicalID, splitter, bucketName);
-        const omVal = { creationDate: new Date().toJSON() };
+        const omVal = {
+            creationDate: new Date().toJSON(),
+            ingestion: bucketMD.getIngestion(),
+        };
         // If the new format usersbucket does not exist, try to put the
         // key in the old usersBucket using the old splitter.
         // Otherwise put the key in the new format usersBucket
@@ -92,7 +96,7 @@ function freshStartCreateBucket(bucket, canonicalID, log, callback) {
             return callback(err);
         }
         log.trace('created bucket in metadata');
-        return addToUsersBucket(canonicalID, bucketName, log, err => {
+        return addToUsersBucket(canonicalID, bucketName, bucket, log, err => {
             if (err) {
                 return callback(err);
             }
@@ -115,7 +119,7 @@ function freshStartCreateBucket(bucket, canonicalID, log, callback) {
  */
 function cleanUpBucket(bucketMD, canonicalID, log, callback) {
     const bucketName = bucketMD.getName();
-    return addToUsersBucket(canonicalID, bucketName, log, err => {
+    return addToUsersBucket(canonicalID, bucketName, bucketMD, log, err => {
         if (err) {
             return callback(err);
         }
@@ -172,9 +176,18 @@ function createBucket(authInfo, bucketName, headers,
         BucketInfo.currentModelVersion(), null, null, null,
         null, null, null, null, null, null, null, null,
         null, isNFSEnabled);
+    let locationConstraintVal = null;
 
-    if (locationConstraint !== undefined) {
-        bucket.setLocationConstraint(locationConstraint);
+    if (locationConstraint) {
+        const [locationConstraintStr, ingestion] =
+            locationConstraint.split(zenkoSeparator);
+        if (locationConstraintStr) {
+            locationConstraintVal = locationConstraintStr;
+            bucket.setLocationConstraint(locationConstraintStr);
+        }
+        if (ingestion === 'ingest') {
+            bucket.enableIngestion();
+        }
     }
     const parseAclParams = {
         headers,
@@ -245,7 +258,7 @@ function createBucket(authInfo, bucketName, headers,
         // error unless old AWS behavior (us-east-1)
         // Existing locationConstraint must have legacyAwsBehavior === true
         // New locationConstraint should have legacyAwsBehavior === true
-        if (isLegacyAWSBehavior(locationConstraint) &&
+        if (isLegacyAWSBehavior(locationConstraintVal) &&
             isLegacyAWSBehavior(existingBucketMD.getLocationConstraint())) {
             log.trace('returning 200 instead of 409 to mirror us-east-1');
             return cb(null, existingBucketMD);
--- a/lib/api/bucketPut.js
+++ b/lib/api/bucketPut.js
@@ -9,6 +9,7 @@ const { config } = require('../Config');
 const aclUtils = require('../utilities/aclUtils');
 const { pushMetric } = require('../utapi/utilities');
 const monitoring = require('../utilities/monitoringHandler');
+const { zenkoSeparator } = require('../../constants');
 
 let { restEndpoints, locationConstraints } = config;
 config.on('rest-endpoints-update', () => {
@@ -35,8 +36,10 @@ function checkLocationConstraint(request, locationConstraint, log) {
     // no locationConstraint provided.
     const { parsedHost } = request;
     let locationConstraintChecked;
+    let locationConstraintAddon;
     if (locationConstraint) {
-        locationConstraintChecked = locationConstraint;
+        locationConstraintAddon = locationConstraint.split(zenkoSeparator)[1];
+        locationConstraintChecked = locationConstraint.split(zenkoSeparator)[0];
     } else if (parsedHost && restEndpoints[parsedHost]) {
         locationConstraintChecked = restEndpoints[parsedHost];
     } else {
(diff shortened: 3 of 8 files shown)


lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -173,8 +173,13 @@ function createBucket(authInfo, bucketName, headers,
         null, null, null, null, null, null, null, null,
         null, isNFSEnabled);
 
-    if (locationConstraint !== undefined) {
-        bucket.setLocationConstraint(locationConstraint);
+    const locationConstraintParse =
```
jonathan-gramain: one maybe more elegant way could be `const { locationName, isIngestion } = locationConstraint.split(...)`, it's clearer to refer to named variables than indices in arrays.

But with respect to my earlier comment, we may as well do `const { locationName, locationProp } = ...`, then check if the prop is `ingest`, call `setIsIngestion`, that leaves room for other kind of properties (of course we could change this code later, but at least we make it clear we can just ignore properties that we don't know to keep backward compat).

Going even further, we could split everything that follows the delimiter as a list of properties, treat them as individual properties and apply the same processing to each, so e.g. `myawsloc:ingest,onsteroids` would still treat it as an ingest location even if we decide to add support for location on steroids later.

We don't have to go that far necessarily but I think we should at least avoid applying `setIsIngestion` to any property that follows without checking it upfront, to ease backward compatibility in the future.

constants.js
```diff
@@ -137,6 +137,8 @@ const constants = {
     azureAccountNameRegex: /^[a-z0-9]{3,24}$/,
     base64Regex: new RegExp('^(?:[A-Za-z0-9+/]{4})*' +
         '(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$'),
+    // location constraint delimiter for ingestionConfig
+    ingestionDelimiter: ':',
```
jonathan-gramain: I would not be surprised if we find later other properties to fit to locations, so a better name may be `locationPropertiesDelimiter`

JianqinWang: I think there's a delimiter that's used for replication (?) Currently trying to find that one and use that instead

rahulreddy: Let's make this generic and call it `zenkoSeparator`

rahulreddy (changes requested):
We need a test ensuring that we are able to create the bucket with that `ingest` option (send request, parse and setting of the property by BucketInfo).

constants.js
JianqinWang: nvm, I will be following @rahulreddy 's comment below :D

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -173,8 +174,13 @@ function createBucket(authInfo, bucketName, headers,
         null, null, null, null, null, null, null, null,
         null, isNFSEnabled);
 
-    if (locationConstraint !== undefined) {
-        bucket.setLocationConstraint(locationConstraint);
+    const { locationConstraintStr, isIngestion } =
+        locationConstraint.split(zenkoSeparator);
```
rachedbenmustapha: `locationConstraint` can still be undefined here i think?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -173,8 +174,13 @@ function createBucket(authInfo, bucketName, headers,
...
+    if (locationConstraintStr !== undefined) {
+        bucket.setLocationConstraint(locationConstraintStr);
+    }
+    if (isIngestion !== undefined) {
```
rachedbenmustapha: it would be safer to explicitly test that the suffix is `ingest`, otherwise a location constraint of `us-west-2:fortnite` or `my-location:do-not-ingest`  would still make it an ingest bucket

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -173,8 +174,13 @@ function createBucket(authInfo, bucketName, headers,
...
+        bucket.setIsIngestion(isIngestion);
```
rachedbenmustapha: and here, only if we matched the `ingest` suffix, set it to a hardcoded `true` to avoid unexpected values in the bucket metadata model

rachedbenmustapha (dismissed):

lib/api/apiUtils/bucket/bucketCreation.js
JianqinWang: the 'ingest' check is in the `BucketInfo` model in Arsenal, which expects `null`, `ingest`, or another different value to pause ingestion. Would appreciate any suggestions on what kind of value should be expected to set the ingestion status to disabled!

rachedbenmustapha: Ok I thought (because of the `is` prefix) that this attribute was just describing the nature of the bucket, not the current state (which I thought would be managed similar to the replication pause/resume in zookeeper if I remember correctly). Is there a particular reason to implement this in a different way?

JianqinWang: I just talked to @rahulreddy, this may stem from my vague naming - I will change it to just `ingestion`, since `isIngestion` implies a true/false value.

We're currently storing the ingestion value as `null`, `{ status: 'Enabled' }`, and `{ status: 'Disabled' }`. They aren't going to be used just yet since we aren't implementing suspend yet, but we're trying to make it so when we include it in the future it'll be here.

rachedbenmustapha: Is `Disabled` going to be the same as replication's `Suspended` or the same as our replication's pause/resume semantics? ie is the backlog processed on re-enable or just thrown away.

Depending on that, either `Suspended` or `Paused` make sense to really convey what's going on, and to be consistent with replication (which is a similar workflow, just opposite direction).

JianqinWang: I think we intend it to be similar to `Paused`, so that backlogs can be processed. tagging @rahulreddy again to confirm

If that's the case, instead of `Disabled`, I should have `status` as `Paused` instead?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -172,9 +173,20 @@ function createBucket(authInfo, bucketName, headers,
         BucketInfo.currentModelVersion(), null, null, null,
         null, null, null, null, null, null, null, null,
         null, isNFSEnabled);
+    let locationConstraintVal = null;
 
-    if (locationConstraint !== undefined) {
-        bucket.setLocationConstraint(locationConstraint);
+    if (locationConstraint) {
+        const locationConstraintStr =
+            locationConstraint.split(zenkoSeparator)[0];
```
nicolas2bert: Just wondering, do we check that the location constraint name does not contain a zenkoSeparator (`:`). e.g. what if `locationConstraint =  's3c:test1'`

JianqinWang: At least from Orbit, when adding a storage location users are not allowed to have ':' in the name

lib/api/apiUtils/bucket/bucketCreation.js
rachedbenmustapha: Yes, `Paused` seems appropriate

tests/functional/aws-node-sdk/test/bucket/put.js
```diff
@@ -224,5 +224,23 @@ describe('PUT Bucket - AWS.S3.createBucket', () => {
...
+                bucketUtil.s3.createBucketAsync(
+                    {
+                        Bucket: bucketName,
+                        CreateBucketConfiguration: {
+                            LocationConstraint: 'us-east-1:ingest',
+                        },
+                    }, (err, res) => {
+                    assert.ifError(err);
+                    assert.strictEqual(res.Location, `/${bucketName}`);
+                    return done();
```
rachedbenmustapha: An assertion that the ingest property is set and that the location constraint of the bucket is well formed would be good.

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -172,9 +173,20 @@ function createBucket(authInfo, bucketName, headers,
...
+        const ingestion =
+            locationConstraint.split(zenkoSeparator)[1];
```
philipyoo: minor: we can also do `const [locationConstraintStr, ingestion] = locationConstraint.split(zenkoSeparator)` and avoid an extra split 👍

lib/api/bucketPut.js
```diff
@@ -54,6 +57,12 @@ function checkLocationConstraint(request, locationConstraint, log) {
         return { error: errors.InvalidLocationConstraint.
             customizeDescription(errMsg) };
     }
+    if (locationConstraintAddon) {
+        // We are introducing a new format for location constraints.
+        // with `zenkoSeparator`, we are currently including the ingestion
+        // flag in the location constraint.
+        locationConstraintChecked = locationConstraint;
```
philipyoo, line 64: this block might not be needed. I only see `locationConstraintAddon` being set on L#41 and `locationConstraintChecked` is already set to whatever the location name is regardless if there is a split or not

lib/api/apiUtils/bucket/bucketCreation.js
JianqinWang: Based on a previous suggestion I had tried `const { locationConstraintStr, ingestion }` which errored out and caused me to use a day to debug. I did not look further into it, but I will try this instead. Thanks

lib/api/bucketPut.js
JianqinWang: I need this block because `locationConstraintChecked` needs to include any other values that was appended on the `locationConstraint` string, and not just the location name. The value is used later on in `createBucket`.

philipyoo: ah I see, sorry misread

ghost:
There were the following issues with this Pull Request

* Commit: b692a6f022990ca989ded50062952099767061bf
  - ✖ message may not be empty
  - ✖ type may not be empty
* Commit: 8f0e28bea97ded1f67fed3dfed50c75deea3982d
  - ✖ message may not be empty
  - ✖ type may not be empty
* Commit: bda8b275c09bab673a7351c6be15c86fb6377dea
  - ✖ message may not be empty
  - ✖ type may not be empty


You may need to [change the commit messages][ref] to comply with the repository contributing guidelines.

--------

🤖 This comment was generated by [**commitlint[bot]**][repo]. Please report issues [here][issues].

Happy coding!

[ref]: https://help.github.com/articles/changing-a-commit-message/
[repo]: https://github.com/z0al/commitlint-bot
[issues]: https://github.com/z0al/commitlint-bot/issues

ghost:
There were the following issues with this Pull Request

* Commit: b692a6f022990ca989ded50062952099767061bf
  - ✖ message may not be empty
  - ✖ type may not be empty
* Commit: 8f0e28bea97ded1f67fed3dfed50c75deea3982d
  - ✖ message may not be empty
  - ✖ type may not be empty
* Commit: bda8b275c09bab673a7351c6be15c86fb6377dea
  - ✖ message may not be empty
  - ✖ type may not be empty
* Commit: c4893357c28123d1d96e84a07039e4ebc2651faa
  - ✖ message may not be empty
  - ✖ type may not be empty


You may need to [change the commit messages][ref] to comply with the repository contributing guidelines.

--------

🤖 This comment was generated by [**commitlint[bot]**][repo]. Please report issues [here][issues].

Happy coding!

[ref]: https://help.github.com/articles/changing-a-commit-message/
[repo]: https://github.com/z0al/commitlint-bot
[issues]: https://github.com/z0al/commitlint-bot/issues

philipyoo (approved):

rahulreddy (approved):

JianqinWang:
@bert-e reset

rahulreddy (approved):

JianqinWang:
@bert-e approve

PR #1606, opened 2018-10-31
feat(ZENKO-1044): Add ceph backend tests

tmacro:
Add functional tests running against ceph as a backend

### DIFF
--- a/eve/workers/ceph/entrypoint-wrapper.sh
+++ b/eve/workers/ceph/entrypoint-wrapper.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+touch /artifacts/ceph.log
+mkfifo /tmp/entrypoint_output
+# We run this in the background so that we can tail the RGW log after init,
+# because entrypoint.sh never returns
+bash entrypoint.sh > /tmp/entrypoint_output &
+entrypoint_pid="$!"
+while read -r line; do
+    echo $line
+    # When we find this line server has started
+    if [ -n "$(echo $line | grep 'Creating bucket')" ]; then
+        break
+    fi
+done < /tmp/entrypoint_output
+
+# Make our buckets - CEPH_DEMO_BUCKET is set to force the "Creating bucket" message, but unused
+s3cmd mb s3://cephbucket s3://cephbucket2
+
+mkdir /root/.aws
+cat > /root/.aws/credentials <<EOF
+[default]
+aws_access_key_id = accessKey1
+aws_secret_access_key = verySecretKey1
+EOF
+
+# Enable versioning on them
+for bucket in cephbucket cephbucket2; do
+    echo "Enabling versiong for $bucket"
+    aws --endpoint http://127.0.0.1:8001 s3api  put-bucket-versioning --bucket $bucket --versioning Status=Enabled
+done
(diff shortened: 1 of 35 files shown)


eve/workers/ceph/entrypoint-wrapper.sh
```diff
@@ -0,0 +1,33 @@
...
+while read -r line; do
+    echo $line
+    # When we find this line server has started
+    if [ -n "$(echo $line | grep 'Creating bucket')" ]; then
+        break
+    fi
+done < /tmp/entrypoint_output
+
+# Make our buckets - CEPH_DEMO_BUCKET is set to force the "Creating bucket" message, but unused
+s3cmd mb s3://cephbucket s3://cephbucket2
```
rahulreddy, line 18: Why use `s3cmd` here instead of the `aws cli`?

tests/functional/aws-node-sdk/test/support/awsConfig.js
```diff
@@ -19,26 +19,40 @@ function getAwsCredentials(profile, credFile) {
 
 function getRealAwsConfig(location) {
     const { awsEndpoint, gcpEndpoint, credentialsProfile,
-        credentials: locCredentials, bucketName, mpuBucketName } =
+        credentials: locCredentials, bucketName, mpuBucketName, pathStyle } =
         config.locationConstraints[location].details;
+    const useHTTPS = config.locationConstraints[location].details.https;
```
rahulreddy, line 24: This is kind of a double negation. It's easier to read if you just rely on `https` var for logic.
```js
const { https } = config.locationConstraints[location].details;
const proto = https ? 'https' : 'http';
```

rahulreddy:
This is a partial review, but posting what I have in the interest of time.

tests/functional/aws-node-sdk/test/support/awsConfig.js
tmacro: I've cleaned up the weird ternary, which was originally to handle `details.https` being undefined (after investigating `Config.js` this isn't possible), but I've kept the useHTTPS name. Reason being `https` is already a top level variable name, and it is used inside the function precluding using block level scoping to re-use the name.

eve/workers/ceph/entrypoint-wrapper.sh
tmacro: So I can create two buckets with one command. It's included in ceph/daemon so we're not dragging in another dep just for this.

tests/functional/aws-node-sdk/test/multipleBackend/listParts/azureListParts.js
```diff
@@ -14,7 +14,8 @@ const bodySecondPart = Buffer.alloc(secondPartSize);
 let bucketUtil;
 let s3;
 
-describeSkipIfNotMultiple('List parts of MPU on Azure data backend', () => {
+describeSkipIfNotMultipleorCeph('List parts of MPU on Azure data backend',
+() => {
```
giacope, line 18: Shouldn't it be like this to be fully camelcase -> `...MultipleOrCeph`?

ssalaues (approved):

rahulreddy (approved):

PR #1611, opened 2018-11-02
Improvement/zenko 1291 ext backends keep alive

rahulreddy:
# Pull request template

## Description
Keep-Alive configuration for external backends

### Motivation and context

Socket hangups were observed when keepalive is enabled (especially to AWS). This PR makes keep alive configuration per backend tunable and set keep alive disabled for AWS.

#### Commits
- improvement: ZENKO-1291 configurable httpAgent options
> This commit adds the ability to configure httpAgent options for external backends.
Currently only AWS and GCP are suppored. Azure is not supported as there is no
straight forward way to set a custom httpAgent in the Azure SDK. The defaults are
expressed to be sensible and explicit.
The default maxSockets is inifinity which is expressed as `null` and other values
are inspired by node.js defaults.
This configuration is applied globally for all locations of the same type of external backend.

- improvement: ZENKO-1291 disable keepAlive for AWS
> This commit disables keep-alive on connections to AWS as connection
reuse showed intermittent socket hang ups.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ _build
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
+yarn.lock
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -150,6 +150,41 @@ fi
 if test -v INITIAL_INSTANCE_ID && test -v S3METADATAPATH && ! test -f ${S3METADATAPATH}/uuid ; then
     echo -n ${INITIAL_INSTANCE_ID} > ${S3METADATAPATH}/uuid
 fi
+# external backends http(s) agent config
+
+# AWS
+if [[ "$AWS_S3_HTTPAGENT_KEEPALIVE" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .externalBackends.aws_s3.httpAgent.keepAlive=$AWS_S3_HTTPAGENT_KEEPALIVE"
+fi
+
+if [[ "$AWS_S3_HTTPAGENT_KEEPALIVE_MS" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .externalBackends.aws_s3.httpAgent.keepAliveMsecs=$AWS_S3_HTTPAGENT_KEEPALIVE_MS"
+fi
+
+if [[ "$AWS_S3_HTTPAGENT_KEEPALIVE_MAX_SOCKETS" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .externalBackends.aws_s3.httpAgent.maxSockets=$AWS_S3_HTTPAGENT_KEEPALIVE_MAX_SOCKETS"
+fi
+
+if [[ "$AWS_S3_HTTPAGENT_KEEPALIVE_MAX_FREE_SOCKETS" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .externalBackends.aws_s3.httpAgent.maxFreeSockets=$AWS_S3_HTTPAGENT_KEEPALIVE_MAX_FREE_SOCKETS"
+fi
+
+#GCP
+if [[ "$GCP_HTTPAGENT_KEEPALIVE" ]]; then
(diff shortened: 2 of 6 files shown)


rahulreddy:
@bert-e wait

lib/Config.js
```diff
@@ -958,6 +985,40 @@ class Config extends EventEmitter {
...
+            const extBackendsConfig = Object.keys(config.externalBackends);
+            extBackendsConfig.forEach(b => {
+                // assert that it's a valid backend
+                assert(validExternalBackends[b] !== undefined,
+                    `bad config: ${b} is not one of valid external backends: ` +
+                    `${Object.keys(validExternalBackends).join(', ')}`);
+
+                const { httpAgent } = config.externalBackends[b];
+                assert(typeof httpAgent === 'object',
+                    `bad confg: ${b} must have httpAgent object defined`);
```
alexanderchan-scality: typo: `confg` => `config`
same typos  in `asserts` below; otherwise, lgtm

lib/Config.js
```diff
@@ -9,13 +9,40 @@ const uuid = require('node-uuid');
...
 
 // whitelist IP, CIDR for health checks
 const defaultHealthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
 
 const defaultLocalCache = { host: '127.0.0.1', port: 6379 };
+const defaultExternalBackendsConfig = {
+    // eslint-disable-next-line camelcase
+    aws_s3: {
+        httpAgent: {
+            keepAlive: true,
```
vrancurel: why true here ?

lib/Config.js
```diff
@@ -9,13 +9,40 @@ const uuid = require('node-uuid');
...
+            keepAliveMsecs: 1000,
+            maxFreeSockets: 256,
+            maxSockets: null,
+        },
+    },
+    azure: {
```
JianqinWang: In the PR comment, it says that azure is not supported, do we still need this segment?

lib/Config.js
rahulreddy: Fixed.

lib/Config.js
rahulreddy: Updated.

lib/Config.js
rahulreddy: Ah good catch. I forgot to remove it while rebasing.

jonathan-gramain (approved):

JianqinWang (approved):

.gitignore
```diff
@@ -28,3 +28,4 @@ _build
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
+yarn.lock
```
bennettbuchanan, line 31: Was this meant to be committed?

rahulreddy: Yes. I should have used a different commit though.

docker-entrypoint.sh
```diff
@@ -150,6 +150,41 @@ fi
...
+
+if [[ "$AWS_S3_HTTPAGENT_KEEPALIVE_MS" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .externalBackends.aws_s3.httpAgent.keepAliveMsecs=$AWS_S3_HTTPAGENT_KEEPALIVE_MS"
+fi
+
+if [[ "$AWS_S3_HTTPAGENT_KEEPALIVE_MAX_SOCKETS" ]]; then
+    JQ_FILTERS_CONFIG="$JQ_FILTERS_CONFIG | .externalBackends.aws_s3.httpAgent.maxSockets=$AWS_S3_HTTPAGENT_KEEPALIVE_MAX_SOCKETS"
+fi
+
+if [[ "$AWS_S3_HTTPAGENT_KEEPALIVE_MAX_SOCKETS" ]]; then
```
bennettbuchanan: Looks like this is missing the 'FREE' for `...MAX_FREE_SOCKETS...`.

i.e. `AWS_S3_HTTPAGENT_KEEPALIVE_MAX_SOCKETS` => `AWS_S3_HTTPAGENT_KEEPALIVE_MAX_FREE_SOCKETS`

At least, that would follow the same pattern as the GCP http agent config naming scheme added just below.

bennettbuchanan (changes requested):
Other than that one typo in the config property, this LGTM.

docker-entrypoint.sh
rahulreddy: That was a typo, copy-pasta failure. I fixed it, thanks.

rahulreddy:
@bert-e reset

bennettbuchanan (approved):

PR #1621, opened 2018-11-14
Bugfix/zenko 1311 ci fixes

ssalaues:
### Motivation and context

- Fixes false positives on linting
- Fixes linting errors
- Adds back the unit_coverage
- Adds automated post-merge docker hub builds https://eve.devsca.com/github/scality/cloudserver/#/builders/3/builds/4628

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -129,6 +129,7 @@ stages:
       - ShellCommand:
           name: Linting
           command: |
+            set -ex
             npm run --silent lint -- --max-warnings 0
             npm run --silent lint_md
             flake8 $(git ls-files "*.py")
@@ -136,6 +137,7 @@ stages:
       - ShellCommand:
           name: Unit Coverage
           command: |
+            set -ex
             mkdir -p $CIRCLE_TEST_REPORTS/unit
             npm run test
             npm run test_legacy_location
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -265,6 +265,12 @@ stages:
             -u '%(secret:private_registry_username)s'
             -p '%(secret:private_registry_password)s'
             '%(secret:private_registry_url)s'
+      - ShellCommand:
+          name: Dockerhub Login
+          command: >
+            docker login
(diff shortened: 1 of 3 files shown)


ssalaues:
@bert-e

anurag4DSB (approved):

eve/main.yml
```diff
@@ -126,20 +126,27 @@ stages:
...
             npm run --silent lint -- --max-warnings 0
             npm run --silent lint_md
             flake8 $(git ls-files "*.py")
             yamllint $(git ls-files "*.yml")
       - ShellCommand:
           name: Unit Coverage
           command: |
+            set -ex
             mkdir -p $CIRCLE_TEST_REPORTS/unit
+            npm run unit_coverage
```
JianqinWang, line 143: I think `unit_coverage` runs the same tests as `test` and `unit_coverage_legacy_location` runs the same as `test_legacy_location`. Does eve store the test output somewhere as an artifact when using the `unit_coverage` options? If so, we can probably remove `npm run test` and `npm run test_legacy_location`.

ssalaues: That makes sense. I wasn't sure about that so I left them in. But yeah the logs get output on the `Unit Coverage logs` step seen below. If preferred, I can have them uploaded as artifacts.

https://eve.devsca.com/github/scality/cloudserver/#/builders/11/builds/672

JianqinWang: Cool! I definitely glazed over `Unit Coverage logs` because the format is so different than the console output when you run locally, I think that's fine. I think we can remove the `npm run test` and `npm run test_legacy_location` then, it was primarily added to help spot issues with the verbose output.

ssalaues:
@bert-e wait

eve/main.yml
ssalaues: Sounds good, I'll do that. Thanks!

JianqinWang (approved):

rahulreddy (approved):

PR #1630, opened 2018-11-15
ZENKO-1337

ssalaues:
### Motivation and context

Fixes a silent issue where the docker tags need escape characters to properly build.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -281,9 +281,15 @@ stages:
           name: Build docker image
           command: >-
             docker build
+            --no-cache
             -t %(prop:docker_image_name)s
-            -t zenko/cloudserver:latest-%(prop:product_version)s
             .
+      - ShellCommand:
+          name: Tag images
+          command: |
+            docker tag %(prop:docker_image_name)s zenko/cloudserver:$TAG
+          env:
+            TAG: "latest-%(prop:product_version)s"
       - ShellCommand:
           name: Push image
           command: |


rahulreddy (approved):
Ok wow.

flavien-scality:
Good catch !
However, if that PR is not urgent, let me double check this with the masters @bertranddemiddelaer @tcarmet @Zempashi : I feel it weird to have to quote slashes inside strings. May be it's a "feature" in eve that we don't want ?

eve/main.yml
```diff
@@ -280,9 +280,9 @@ stages:
       - ShellCommand:
           name: Build docker image
           command: >-
-            docker build
-            -t %(prop:docker_image_name)s
-            -t zenko/cloudserver:latest-%(prop:product_version)s
+            docker build --no-cache
```
tcarmet: There's a funny thing here. You're using a worker of type `local` on this stage. This means (you guys already know) that the `ShellCommands` are executed on the same host as the buildbot/eve backends. Lucky you there's docker installed and everything is good to go for all the commands to work. But the thing is, knowing what's in there, I wouldn't recommend to do this anymore, there's so much custom scripts around the docker cli itself (which I'm trying to remove but that's another story) and if one day by any chance we decide to change things inside the backends like removing docker or this kind of stuff (won't happen but still) well in this case your stage will break.

What I would recommend here is execute this stage on worker of type `kube_pod` with docker cli installed and bind the docker socket. (the same way that we do for zenko or backbeat, if I remember well).
Now I know that's not a good solution in the long term, and I'll work on that.

As per the `--no-cache` I think it's good to leave it there, everybody have some cache layers of cloudserver hanging and their computer, it's good to have something that build it from scratch :smile:

eve/main.yml
```diff
@@ -280,9 +280,9 @@ stages:
...
+            -t "%(prop:docker_image_name)s"
+            -t "zenko\/cloudserver:latest-%(prop:product_version)s"
```
tcarmet: wow if we reached to this point, maybe it will be better to put those commands on a script/makefile?

tcarmet (commented):
Very sorry for the trouble this caused... and good catch! I think we need to improve the ease of docker builds inside the CI and provide a cleaner way for you guys!

Sorry for the long comment! :smile:

eve/main.yml
ssalaues: I think you're right, maybe the `kube_pod` would be a better approach and backbeat is already doing it this way. 

Thanks for the input!

ssalaues:
@bert-e wait

ssalaues:
@tcarmet I updated the PR with a simple fix that seems to work well. Let me know what you think

tcarmet (approved):
LGTM

A little note: I still think we should remove the docker build from the local worker BUT that's not the problem we're trying to fight here. And I think there's some work to be done on RELENG side so let's leave it that way :)

ssalaues:
@tcarmet Thanks yeah I wanted to keep it simple and have the quick win so this isn't stuck in limbo since I haven't had time to setup the switch over to the Kube worker and test.

ssalaues:
@bert-e approve

tcarmet:
@ssalaues yeah, I agree.

PR #1638, opened 2018-11-27
feat(PfsClient): add delete functionality

giacope:
This PR:
- Adds delete support for the PfsClient.
- Allows you to configure the PfsDaemon for ReadOnly mode.
- Deletes unnecessary and unused properties.

Depends on https://github.com/scality/Arsenal/pull/609.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -669,6 +669,8 @@ class Config extends EventEmitter {
             this.pfsDaemon.dataPath =
                 process.env.PFSD_MOUNT_PATH ?
                 process.env.PFSD_MOUNT_PATH : `${__dirname}/../localPfs`;
+            this.pfsDaemon.isReadOnly =
+                process.env.PFSD_READONLY === 'true';
         }
 
         if (config.metadataDaemon) {
--- a/lib/data/external/PfsClient.js
+++ b/lib/data/external/PfsClient.js
@@ -8,13 +8,8 @@ class PfsClient {
         const { host, port } = config.endpoint;
 
         this.clientType = 'pfs';
-        this.type = 'PFS';
-        this._bucketName = config.bucketName;
         this._bucketMatch = config.bucketMatch;
-        this._serverSideEncryption = config.serverSideEncryption;
         this._dataStoreName = config.dataStoreName;
-        this._supportsVersioning = config.supportsVersioning;
-        this._mountPath = config.mountPath;
         this._restClient = new arsenal.network.rest.RESTClient({
             host,
             port,
@@ -72,9 +67,16 @@ class PfsClient {
 
     delete(objectGetInfo, reqUids, callback) {
         const log = createLogger(reqUids);
-        logHelper(log, 'error', 'Not implemented', errors.NotImplemented,
+        const key = typeof objectGetInfo === 'string' ? objectGetInfo :
+            objectGetInfo.key;
+        this._restClient.delete(key, reqUids, err => {
+            if (err) {
+                logHelper(log, 'error', 'err from data backend', err,
                 this._dataStoreName, this.clientType);
-        return callback(errors.NotImplemented);
+                return callback(err);
+            }
+            return callback();
+        });
     }
 
     // TODO: Implement a healthcheck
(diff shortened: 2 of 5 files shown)


pfsserver.js
```diff
@@ -11,6 +11,7 @@ const pfsServer = new arsenal.network.rest.RESTServer({
         dataPath: config.pfsDaemon.dataPath,
         log: config.log,
         isPassthrough: true,
+        isReadOnly: true,
```
alexanderchan-scality:
```suggestion
        isReadOnly: config.pfsDaemon.isReadOnly,
```

lib/Config.js
```diff
@@ -662,6 +662,7 @@ class Config extends EventEmitter {
             this.pfsDaemon.dataPath =
                 process.env.PFSD_MOUNT_PATH ?
                 process.env.PFSD_MOUNT_PATH : `${__dirname}/../localPfs`;
+            this.pfsDaemon.isReadOnly = process.env.PFSD_READONLY || true;
```
jonathan-gramain: That would most likely not work as expected to allow writing, because `PFSD_READONLY` is a string coming straight from the environment, except maybe if an empty string it would be interpreted as truish.

To avoid the double-negation, i.e. asking the user to set a falsy env var to disable the read-only mode, I would recommend `PFSD_RDWR` or similar, so that the user can set it to any truish value, which is more common, i.e. enable write rather than disable no-write-allowed.

lib/data/external/PfsClient.js
```diff
@@ -72,9 +67,15 @@ class PfsClient {
 
     delete(objectGetInfo, reqUids, callback) {
         const log = createLogger(reqUids);
-        logHelper(log, 'error', 'Not implemented', errors.NotImplemented,
+        const key = objectGetInfo.key ? objectGetInfo.key : objectGetInfo;
```
jonathan-gramain: Can `objectGetInfo` be a string? In which case I think it has to be checked explicitly, e.g. with `typeof objectGetInfo === "string"` to avoid attempting to access a field in a string.

lib/Config.js
giacope: Oh yeah, my bad. It should probably check if the string === 'true' and set the default to false.

However, I prefer to stick with `PFSD_READONLY` to match Kubernetes' NFS Volume spec `readOnly` attribute, since it's connected to PFSD in the end.
And because we don't support writes other than a delete.

lib/data/external/PfsClient.js
giacope: Probably, since that's how the AzureClient's delete handles it. Just updated the code to check for that.

tmacro (approved):

lib/data/external/PfsClient.js
```diff
@@ -72,9 +67,16 @@ class PfsClient {
 
     delete(objectGetInfo, reqUids, callback) {
         const log = createLogger(reqUids);
-        logHelper(log, 'error', 'Not implemented', errors.NotImplemented,
+        const key = typeof objectGetInfo === 'string' ? objectGetInfo :
```
bennettbuchanan, line 70: If not a string, is `objectGetInfo` guaranteed to be an object?

bennettbuchanan (commented):
Nice, just one clarifying question.

lib/data/external/PfsClient.js
giacope: Well, it's javascript 😅, but it seems like it is. It's usually created by a `toObjectGetInfo()` function. Azure handles it exactly the same way -> https://github.com/scality/cloudserver/blob/development/8.1/lib/data/external/AzureClient.js#L226

bennettbuchanan: I see, LGTM. Just wanted to confirm that. 😄

bennettbuchanan (approved):

giacomoguiulfo:
@bert-e wait

rahulreddy (approved):

giacope:
@bert-e approve

PR #1642, opened 2018-12-03
Improvement/zenko 1045 ceph e2 e tests

tmacro:
## Description
Allow forcing path-style buckets, and http for AWS locations using the pre-existing `CI` environment variable.

### Motivation and context
Required for Ceph in Zenko's CI. At present there is no way for a backend to be added as AWS and to also use path-style bucket, and http if configured through Orbit (or in our case orbit-simulator). Currently, Barring running our own DNS server, or resorting to kube-dns hacks, we are forced to use path-style buckets for "mocked" backends in our CI. For https we would need either external access for ACME (LetsEncrypt), external hosted DNS with an API for ACMEv2 DNS challenges, or to add a self-signed cert to our cloudserver docker images.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -237,16 +237,23 @@ function locationConstraintAssert(locationConstraints) {
             assert(typeof details.credentials.secretKey === 'string',
                 'bad config: credentials must include secretKey as string');
         }
-        if (details.https !== undefined) {
+        if (process.env.CI === 'true') {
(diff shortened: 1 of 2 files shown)


rahulreddy (approved):

tmacro:
@bert-e approve

ssalaues (approved):

PR #1652, opened 2018-12-11
feature: ZENKO-733 Lifecycle non-versioned buckets

bennettbuchanan:
Lifecycle transitions requires fetching object metadata through the backbeat routes such that the replication of data can be performed. Transitions can occur on non-versioned buckets, so we do not want to return an `InvalidBucket` error in that case.

### DIFF
--- a/lib/api/bucketPutVersioning.js
+++ b/lib/api/bucketPutVersioning.js
@@ -14,6 +14,10 @@ const monitoring = require('../utilities/monitoringHandler');
 const externalVersioningErrorMessage = 'We do not currently support putting ' +
 'a versioned object to a location-constraint of type Azure or GCP.';
 
+const replicationVersioningErrorMessage = 'A replication configuration is ' +
+'present on this bucket, so you cannot change the versioning state. To ' +
+'change the versioning state, first delete the replication configuration.';
+
 /**
  * Format of xml request:
 
@@ -75,6 +79,18 @@ function _checkBackendVersioningImplemented(bucket) {
     return true;
 }
 
+function _isValidReplicationVersioning(result, bucket) {
+    if (result.VersioningConfiguration &&
+        result.VersioningConfiguration.Status) {
+        // Is there a replication configuration set on the bucket and is the
+        // user attempting to suspend versioning?
+        if (bucket.getReplicationConfiguration()) {
+            return result.VersioningConfiguration.Status[0] !== 'Suspended';
+        }
+    }
+    return true;
+}
+
 /**
  * Bucket Put Versioning - Create or update bucket Versioning
  * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
@@ -115,6 +131,15 @@ function bucketPutVersioning(authInfo, request, log, callback) {
                     error: errors.NotImplemented });
                 const error = errors.NotImplemented.customizeDescription(
                     externalVersioningErrorMessage);
+                return next(error, bucket);
+            }
+            if (!_isValidReplicationVersioning(result, bucket)) {
+                log.debug(replicationVersioningErrorMessage, {
+                    method: 'bucketPutVersioning',
+                    error: errors.InvalidBucketState,
+                });
+                const error = errors.InvalidBucketState
+                    .customizeDescription(replicationVersioningErrorMessage);
                 return next(error, bucket);
             }
             const versioningConfiguration = {};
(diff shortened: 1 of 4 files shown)


bennettbuchanan:
@bert-e wait

nicolas2bert (approved):

lib/routes/routeBackbeat.js
```diff
@@ -851,7 +851,11 @@ function routeBackbeat(clientIP, request, response, log) {
                 return backbeatRoutes[request.method][request.resourceType]
                     [request.query.operation](request, response, log, next);
             }
-            if (!bucketInfo.isNFS() && !bucketInfo.isVersioningEnabled()) {
+            const allowNonVersionedBucket =
```
jonathan-gramain: Do we enforce the versioning check later when real CRR is done? I think we should still make sure that we cannot CRR from a nonversioned bucket even if a transition policy is setup, which would mean that we would have to distinguish replication tasks triggered by either, and accept transition-triggered ones while refusing CRR-triggered ones. It seems to me this change, although correct, does not look sufficient to ensure this.

jonathan-gramain: If we already enforce this when a replication configuration is set, maybe we could as well remove this check altogether?

bennettbuchanan: I agree. We do enforce versioning when setting a replication config, but there is no check in case versioning is suspended later on. Ultimately, I think the best option is to remove this check and add a check for suspending versioning on buckets with a replication config enabled. That way we can ensure that any bucket with replication will meet this condition, and lifecycle operations can work regardless.

AWS does not allow suspending versioning on replication enabled buckets, so that would be a good compatibility improvement also.

tests/unit/api/bucketPutVersioning.js
```diff
@@ -29,9 +31,26 @@ const locConstraintNonVersioned =
...
+            '<Bucket>arn:aws:s3:::destination-bucket</Bucket>' +
+            '<StorageClass>us-east-2</StorageClass>' +
+        '</Destination>' +
+    '</Rule>' +
+'</ReplicationConfiguration>';
+
 const externalVersioningErrorMessage = 'We do not currently support putting ' +
 'a versioned object to a location-constraint of type Azure or GCP.';
 
+const replicationVersioningErrorMessage = 'A replication configuration is ' +
```
jonathan-gramain: It's ok to test an error message but I'm not a big fan as it makes it more clumsy to change the wording (have to change the test as well), I usually prefer to just check the error code or type.

lib/api/bucketPutVersioning.js
```diff
@@ -75,6 +79,16 @@ function _checkBackendVersioningImplemented(bucket) {
     return true;
 }
 
+function _isValidReplicationVersioning(result, bucket) {
+    const status = result.VersioningConfiguration.Status[0];
```
jonathan-gramain: Is it guaranteed here that `VersioningConfiguration` always exist in the result? Even if so I would rather check it to be sure.

tests/unit/api/bucketPutVersioning.js
bennettbuchanan: Ok, sounds good to me. I was following the pattern of this test file, but I think that's a good enough reason to change it up.

lib/api/bucketPutVersioning.js
```diff
@@ -80,11 +80,12 @@ function _checkBackendVersioningImplemented(bucket) {
 }
 
 function _isValidReplicationVersioning(result, bucket) {
-    const status = result.VersioningConfiguration.Status[0];
-    // Is there a replication configuration set on the bucket and the user is
-    // attempting to suspend versioning?
-    if (bucket.getReplicationConfiguration()) {
-        return status !== 'Suspended';
+    if (result.VersioningConfiguration.Status) {
```
jonathan-gramain: Don't you want to check `result.VersioningConfiguration` as well?

bennettbuchanan: I'm assuming it's safe because just after this the same condition is used. However, it's never really a good assumption to make in any case and I agree it's better to be safe as possible.

jonathan-gramain (approved):

dora-korpar (approved):

bennettbuchanan:
@bert-e approve

PR #1656, opened 2018-12-13
bugfix: ZENKO-1365 Orbit set https for RING+S3C

dora-korpar:
Add HTTPS support for RING+S3C endpoints added through Orbit.

### DIFF
--- a/lib/management/configuration.js
+++ b/lib/management/configuration.js
@@ -128,7 +128,7 @@ function patchConfiguration(newConf, log, cb) {
                         if (awsEndpoint.includes('://')) {
                             const url = new URL(awsEndpoint);
                             awsEndpoint = url.host;
-                            https = url.scheme === 'https';
+                            https = url.protocol.includes('https');
                         }
 
                         location.details = {
--- a/tests/unit/management/configuration.js
+++ b/tests/unit/management/configuration.js
@@ -168,6 +168,19 @@ describe('patchConfiguration', () => {
                     sizeLimitGB: 0,
                     details: {},
                 },
+                'httpsawsbackendtest': {
+                    name: 'httpsawsbackendtest',
+                    objectId: 'httpsawsbackendtest',
+                    locationType: 'location-scality-ring-s3-v1',
+                    details: {
+                        bucketMatch: 'rings3bucketmatch',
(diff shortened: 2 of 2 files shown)


dora-korpar:
@bert-e reset

lib/management/configuration.js
```diff
@@ -116,6 +116,19 @@ function patchConfiguration(newConf, log, cb) {
                     break;
                 case 'location-scality-ring-s3-v1':
                     pathStyle = true; // fallthrough
```
alexanderchan-scality: no longer a `fallthrough`

rachedbenmustapha (commented):
This is a lot of copy-pasted code that was not needed before, can we track down the regression and fix that instead?

lib/management/configuration.js
```diff
@@ -116,6 +116,19 @@ function patchConfiguration(newConf, log, cb) {
...
+                    let https = false;
+                    let ringEndpoint = l.details.endpoint ||
+                        's3.amazonaws.com';
+                    if (ringEndpoint.includes('://')) {
+                        const url = new URL(awsEndpoint);
+                        ringEndpoint = url.host;
+                        https = url.scheme === 'https';
+                    }
+                    location.details = {
+                        awsEndpoint: ringEndpoint,
```
alexanderchan-scality: missing properties? `credentials`, `bucketName`, `bucketMatch`, sse, `supportsVersioning`, `pathStyle`, `https`.

dora-korpar: Are these set by default somewhere if nothing is specified for location.details? Cos what is being set doesn't seem consistent across the aws-derived locations

rachedbenmustapha: They were set in the fallthrough code path

alexanderchan-scality: the defaults are empty.

dora-korpar:
@bert-e reset

dora-korpar:
It turned out the url object was being used incorrectly

dora-korpar:
@bert-e reset

philipyoo (approved):

rachedbenmustapha (approved):

dora-korpar:
@bert-e approve

alexanderchan-scality (approved):

PR #1673, opened 2018-12-28
feat: add nfs-mount-v1 support

giacope:


lib/management/configuration.js
```diff
@@ -183,6 +183,19 @@ function patchConfiguration(newConf, log, cb) {
...
                     }
                     break;
+                case 'location-nfs-mount-v1':
+                    location.type = 'pfs';
+                    if (l.details) {
+                        location.details = {
+                            supportsVersioning: false,
+                            bucketMatch: true,
+                            pfsDaemonEndpoint: {
+                                host: l,
```
ssalaues: shouldn't it be something like `l.name` + `cosmos-pfsd`?

giacope: Yeah, my bad. However, that won't work if the location name is `cosmos` (I added the change anyway). But I'm don't like the idea of hard-coding `cosmos-pfsd`. Maybe we should use an env variable?

ssalaues: Yeah it I was just thinking about that too because of the templating on the chart. What would the env var define? As in it would just define the suffix?

giacope: Yeah, just the suffix. We can't still have a location called cosmos though.

alexanderchan-scality (approved):

jonathan-gramain (approved):

giacope:
@bert-e approve

rahulreddy (approved):

PR #1687, opened 2019-01-09
Bugfix/zenko 1435 pathstyle https

rahulreddy:
- bugfix: ZENKO-1435 remove coverage commands
> This switches up the coverage cmds of istanbul in favor of vanilla
mocha output for test results accuracy.

- bugfix: ZENKO-1435 use CI_CEPH has test flag
> The usage of CI env as a test flag for configuring https, pathStyle
options breaks other tests, so the env is being changed to be CI_CEPH
confining the scope to only CEPH backend tests.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -229,7 +229,7 @@ stages:
         env:
           <<: *multiple-backend-vars
           <<: *global-env
-          CI_CEPH: enabled
+          CI_CEPH: "true"
           MPU_TESTING: "yes"
           S3_LOCATION_FILE: tests/locationConfig/locationConfigCeph.json
     steps:
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -159,8 +159,8 @@ stages:
           command: |
             set -ex
             mkdir -p $CIRCLE_TEST_REPORTS/unit
-            npm run unit_coverage
-            npm run unit_coverage_legacy_location
+            npm test
+            npm run test_legacy_location
           env: &shared-vars
             <<: *global-env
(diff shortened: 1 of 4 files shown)


lib/Config.js
```diff
@@ -237,7 +237,7 @@ function locationConstraintAssert(locationConstraints) {
             assert(typeof details.credentials.secretKey === 'string',
                 'bad config: credentials must include secretKey as string');
         }
-        if (process.env.CI === 'true') {
+        if (process.env.CI_CEPH === 'true') {
```
ssalaues, line 240: Should update this to `'true'`
https://github.com/scality/cloudserver/blob/33a8e1c4799834f8ef1c75fe82f41920218c6597/eve/main.yml#L232

ssalaues (changes requested):

lib/Config.js
rahulreddy: Damn, yes.

alexanderchan-scality (approved):

eve/main.yml
```diff
@@ -229,7 +229,7 @@ stages:
         env:
           <<: *multiple-backend-vars
           <<: *global-env
-          CI_CEPH: enabled
+          CI_CEPH: true
```
ssalaues: I think this will need to be quoted because it registers as a bool instead of a string. I checked on the eve build and ends up like `CI_CEPH=True` in the build environment.

dora-korpar (approved):

ssalaues (approved):

rahulreddy:
@bert-e approve

PR #1702, opened 2019-02-06
feature: S3C-1182 add versioning support for GCP backend

rahulreddy:
## Description
Supports versioning with GCP backend

### DIFF
--- a/lib/data/external/GcpClient.js
+++ b/lib/data/external/GcpClient.js
@@ -6,7 +6,8 @@ const { GCP, GcpUtils } = require('./GCP');
 const { createMpuKey } = GcpUtils;
 const AwsClient = require('./AwsClient');
 const { prepareStream } = require('../../api/apiUtils/object/prepareStream');
-const { logHelper, removeQuotes } = require('./utils');
+const createLogger = require('../multipleBackendLogger');
+const { logHelper, removeQuotes, trimXMetaPrefix } = require('./utils');
 const { config } = require('../../Config');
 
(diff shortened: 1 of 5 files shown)


lib/data/external/GcpClient.js
```diff
@@ -80,9 +81,34 @@ class GcpClient extends AwsClient {
...
+            if (headResp.err) {
+                return cb(null, headResp);
+            }
+            let bucketResp;
+            return this._client.getBucketVersioning(
+            { Bucket: headResp.gcpBucket }, (err, data) => {
+                bucketResp = { gcpBucket: headResp.gcpBucket };
+                if (err) {
+                    bucketResp.error = err;
+                } else if (!data.Status || data.Status === 'Suspended') {
```
jonathan-gramain, line 95: I'm not sure but it might be possible that a status exists and is `'Disabled'`, in which case it could be useful to add it to the condition, or reverse it as `data.Status !== 'Enabled'`.

lib/data/external/GcpClient.js
```diff
@@ -97,6 +123,70 @@ class GcpClient extends AwsClient {
         });
     }
 
+    put(stream, size, keyContext, reqUids, callback) {
```
jonathan-gramain, line 126: If the `AwsClient.put()` implementation can be considered the default implementation for backends (same for `_createAwsKey`), it seems you could just set `this._supportsVersioning = true` and keep using it to avoid code duplication, but I may miss something.

PR #1719, opened 2019-02-20
feature: ZENKO-1420 return locations from backbeat PUT routes

jonathan-gramain:
Return the locations array in the API of putObject and completeMPU
multiple backend routes, so that transition policies can use it to
update metadata with the new location as-is.

I'm considering testing this change as part of backbeat, as the routes are already used and tested there.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -5,9 +5,10 @@ const httpProxy = require('http-proxy');
 const backbeatProxy = httpProxy.createProxyServer({
     ignorePath: true,
 });
-const { auth, errors, s3middleware, models } = require('arsenal');
+const { auth, errors, s3middleware, models, storage } = require('arsenal');
 const { responseJSONBody } = require('arsenal').s3routes.routesUtils;
 const { getSubPartIds } = s3middleware.azureHelper.mpuUtils;
+const { skipMpuPartProcessing } = storage.data.external.backendUtils;
 const vault = require('../auth/vault');
 const { data } = require('../data/wrapper');
 const metadata = require('../metadata/wrapper');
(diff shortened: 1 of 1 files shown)


bennettbuchanan (approved):

lib/routes/routeBackbeat.js
```diff
@@ -402,9 +402,26 @@ function putObject(request, response, log, callback) {
             if (contentMD5 !== md5) {
                 return callback(errors.BadDigest);
             }
+            // stealing some code from createAndStoreObject.js to
+            // create a valid locations array
+            const { key, dataStoreName, dataStoreType, dataStoreETag,
+                    dataStoreVersionId } = retrievalInfo;
+            const prefixedDataStoreETag = dataStoreETag
+                      ? `1:${dataStoreETag}`
```
nicolas2bert: Just curious, why `dataStoreETag` has to be updated with `1:`  as a prefix?

nicolas2bert (approved):

alexanderchan-scality (approved):

lib/routes/routeBackbeat.js
ploki: If I remember correctly, it is related to the put object copy part, to keep track of what is really a part locally (think nesting). I don't remember and I have to dig deeper to find out if it was only to support the NFS server's operation or to have the correct behavior with cloudserver as well

lib/routes/routeBackbeat.js
```diff
@@ -184,9 +185,66 @@ function getPartList(parts, objectKey, uploadId, storageLocation) {
...
+
     return partList;
 }
 
+function generateMpuAggregateInfo(parts) {
+    let aggregateSize = 0;
+
+    parts.forEach(part => {
+        // CopyLocationTask does transmit a size, MultipleBackendTask
+        // does not, so check if size is defined.
```
bennettbuchanan: Thinking out loud: can we break after we know that no size is transmitted?

bennettbuchanan (approved):

jonathan-gramain:
Test failed because of flaky test.

jonathan-gramain:
@bert-e approve

rahulreddy (approved):
Exceptional approval

rahulreddy:
@bert-e bypasss_build_status

rahulreddy:
@bert-e bypass_build_status

PR #1728, opened 2019-03-05
bugfix: restrict metadata proxy routes to policy

rahulreddy:
## Description
bugfix: restrict metadata proxy routes to policy
### Motivation and context

These changes ensure that the routes are available only to account credentials
or credentials of users who are assigned metadata:admin policy

### DIFF
--- a/lib/routes/routeMetadata.js
+++ b/lib/routes/routeMetadata.js
@@ -1,13 +1,12 @@
 const url = require('url');
 const { waterfall } = require('async');
 const httpProxy = require('http-proxy');
-const { auth, errors, s3routes } = require('arsenal');
+const { auth, errors, s3routes, policies } = require('arsenal');
 const { config } = require('../Config');
 const constants = require('../../constants');
 const vault = require('../auth/vault');
-const prepareRequestContexts = require(
-    '../api/apiUtils/authorization/prepareRequestContexts');
 
+const RequestContext = policies.RequestContext;
 const { responseJSONBody } = s3routes.routesUtils;
 const metadataProxy = httpProxy.createProxyServer({ ignorePath: true });
 auth.setHandler(vault);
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
   },
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
-    "arsenal": "scality/Arsenal#879823c4",
+    "arsenal": "scality/Arsenal#feature/S3C-2002-admin-service",
```
JianqinWang: are we going back to referencing branches instead of commit hashes?

lib/routes/routeMetadata.js
```diff
@@ -37,21 +37,35 @@ function routeMetadata(clientIP, request, response, log) {
...
             if (err) {
                 log.debug('authentication error', {
                     error: err,
                     method: request.method,
                     bucketName: request.bucketName,
                     objectKey: request.objectKey,
                 });
+                return next(err);
+            }
+            if (authRes && !authRes.isAllowed) {
```
jonathan-gramain: Looking at cloudserver lib/api/api.js, the returned `authRes` there is an array, I'm assuming it's because there is a 1:1 mapping between input requestContexts and output authorization results, so it seems the proper check would be `authRes && authRes[0].isAllowed`

jonathan-gramain (changes requested):

lib/routes/routeMetadata.js
rahulreddy: Ah, you are right. I overlooked the response.

package.json
rahulreddy: No, this change crept in. I am going to add the commit hash

rahulreddy: But yes, for S3C we are going to do branches. I will have to do those changes after the release.

rahulreddy:
@bert-e reset

jonathan-gramain (approved):

rahulreddy:
@bert-e reset

JianqinWang (approved):

rahulreddy:
@bert-e approve

philipyoo (approved):

rahulreddy:
@bert-e reset

PR #1754, opened 2019-03-14
bf: ZENKO 1402 location constraint update event

dora-korpar:
Fixes the issue of Zenko's location constraints not updating correctly after a 'location-constraint-update' event.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -3,7 +3,7 @@ const https = require('https');
 const cluster = require('cluster');
 const { series } = require('async');
 const arsenal = require('arsenal');
-const { RedisClient, StatsClient } = require('arsenal').metrics;
+const { RedisClient, StatsClient } = arsenal.metrics;
 const monitoringClient = require('./utilities/monitoringHandler');
 
 const logger = require('./utilities/logger');
@@ -12,7 +12,7 @@ const { clientCheck } = require('./utilities/healthcheckHandler');
 const _config = require('./Config').config;
 const { blacklistedPrefixes } = require('../constants');
 const api = require('./api/api');
-const { client, implName } = require('./data/wrapper');
+const dataWrapper = require('./data/wrapper');
 const kms = require('./kms/wrapper');
 const locationStorageCheck =
     require('./api/apiUtils/object/locationStorageCheck');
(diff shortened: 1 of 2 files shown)


lib/server.js
```diff
@@ -25,14 +25,24 @@ const {
...
 function updateAllEndpoints() {
     allEndpoints = Object.keys(_config.restEndpoints);
 }
 _config.on('rest-endpoints-update', updateAllEndpoints);
 updateAllEndpoints();
+_config.on('location-constraints-update', () => {
+    if (implName === 'multipleBackends') {
+        const clients = parseLC(_config, vault);
+        client = new MultipleBackendGateway(
+            clients, metadata, locationStorageCheck);
```
alexanderchan-scality, line 43: should call client setter here for `dataWrapper`?
correct me if I am wrong, the event handler seems to be redundant  (https://github.com/scality/cloudserver/pull/1754/files#diff-e670b3417c2253a33d4d979c93a02692R51)

dora-korpar: Yeah, our codepath for 'GETs' is a little different from the rest, and why we need another event handler. We're not actually instantiating a DataWrapper here because it would lose its scope reference for `this` being passed around through Arsenal, so the event handler is the only way to make sure the location constraint list is updated for the DataWrapper that's instantiated like 5 functions down the line here https://github.com/scality/Arsenal/blob/development/8.1/lib/s3routes/routesUtils.js#L325

alexanderchan-scality: I see. Thanks for the explanation. I had overlooked that the `clients` in `server.js` and `data/wrapper.js` do not reference the same object.

alexanderchan-scality (approved):

rahulreddy (changes requested):
I am curious as to why this does not show up in our testing. I am thinking we have a code path here that is not being tested which is why we never detected it in the first place. I would recommend figuring out a test that fails without this commit and then passes after this commit is added.

rahulreddy (approved):

rahulreddy:
@bert-e bypass_build_status

JianqinWang (approved):

dora-korpar:
@bert-e approve

PR #1758, opened 2019-03-15
feature: ZENKO-1585 configure replication group id

rahulreddy:
## Description
feature: ZENKO-1585 configure replication group id
### Motivation and context

Replication group id is used in the calculation of version ids for objects
on buckets that have versioning enabled. It has many benefits including avoiding
collisions in an active-active scenario, identifying distinct entries from an
instance etc.
This allows commit allows configuring a custom replication group id for an instance,
ensuring that the version ids generated by the instance are unique.


anurag4DSB (approved):

philipyoo (approved):

rahulreddy:
@bert-e approve

alexanderchan-scality (approved):

rahulreddy:
Damn, I missed a file from Config.

PR #1759, opened 2019-03-18
feature: S3C-2035: KMIP, TLS files from env

ploki:
Load the content of the TLS certificates and key when the configuration is provided through the environment variables.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -341,6 +341,27 @@ class Config extends EventEmitter {
         });
     }
 
+    _loadTlsFile(tlsFileName) {
+        if (!tlsFileName) {
+            return undefined;
+        }
+        if (typeof tlsFileName !== 'string') {
+            throw new Error(
+                'bad config: TLS file specification must be a string');
+        }
+        const tlsFilePath = (tlsFileName[0] === '/')
+              ? tlsFileName
+              : path.join(this._basepath, tlsFileName);
+        let tlsFileContent;
+        try {
+            tlsFileContent = fs.readFileSync(tlsFilePath);
+        } catch (err) {
+            throw new Error(`Could not load tls file '${tlsFileName}':` +
+                            ` ${err.message}`);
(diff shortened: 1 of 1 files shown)


alexanderchan-scality (approved):

lib/Config.js
```diff
@@ -814,11 +835,12 @@ class Config extends EventEmitter {
                      * must become 'hosts'
                      */
                     host: process.env.S3KMIP_HOSTS,
-                    key: process.env.S3KMIP_KEY || undefined,
-                    cert: process.env.S3KMIP_CERT || undefined,
-                    ca: process.env.S3KMIP_CA
-                        ? process.env.S3KMIP_CA.split(',')
-                        : undefined,
+                    key: this._loadTlsFile(process.env.S3KMIP_KEY || undefined),
```
alexanderchan-scality, line 838: ~~should `_loadTlsFile` be called here? I may be wrong, but it seems like loading the file here will cause the call on https://github.com/scality/cloudserver/pull/1759/files#diff-9280903abc69f09eff9ab36070b6d751R888 to throw an error~~ confused this.kmip with config.kmip

tmacro (approved):

lib/Config.js
ploki: I don't see why it would throw an error, can you be more specific?

ploki: does this comment apply to the other members of the tls struct? (e.g. ca and cert)

alexanderchan-scality: you are correct. upon double checking, there won't be an error. I have mixed up `this.kmip` vs `config.kmip`

ploki: I did too, the reason for this PR...

dora-korpar (approved):

ploki:
@bert-e approve

ploki:
@bert-e reset

PR #1763, opened 2019-03-21
bugfix: ZENKO-1583 Non-current version transition

bennettbuchanan:


bennettbuchanan:
Closing in favor of correct JIRA tracking number.

PR #1767, opened 2019-03-26
ft(ZENKO-1616): Add fixes for new ObjectMD properties

tmacro:
Add special handling for the creation-time property

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -120,6 +120,22 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             objectKey, bucketMD, false, size, null, null, authInfo),
         log,
     };
+
+    // For Azure BlobStorage API compatability
+    // If an object already exists copy/repair creation-time
+    // creation-time must remain static after an object is created
+    //  --> EVEN FOR VERSIONS <--
+    if (objMD) {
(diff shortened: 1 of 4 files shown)


JianqinWang (approved):

ploki (commented):
:+1: for the core logic. (we might not want to push the package.json stuff as-is though)

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -111,6 +111,22 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             objectKey, bucketMD, false, size, null, null, authInfo),
         log,
     };
+
+    // For Azure BlobStorage API compatability
+    // If an object already exists copy/repair creation-time
+    // creation-time must remain static after an object is created
+    //  --> EVEN FOR VERSIONS <--
+    if (objMD) {
+        if (objMD['creation-time']) {
```
jonathan-gramain, line 129: IMHO this is the kind of compat layer that could be done directly in `ObjectMD` class.

tmacro: That would be ideal, however there isn't really a place to put it where all ops touching MD would pass through. This area of the codebase is definitely in need of a refactor.

ploki (approved):

tmacro:
@bert-e approve

dora-korpar (approved):

PR #1768, opened 2019-03-27
ft: ZENKO-1661 add ingestion data to reporthandler

philipyoo:
Update the report handler to get ingestion location statuses and metrics.

There was an issue in determining ingestion-only locations. Pause/resume routes will return ingestion-only locations, so rely on first get pause statuses, and then for each of those locations, get metrics.

Changed "stats" -> "metrics" because naming between "stats" and "states" or "status" seemed too similar.

### DIFF
--- a/lib/utilities/reportHandler.js
+++ b/lib/utilities/reportHandler.js
@@ -17,6 +17,9 @@ const REQ_PATHS = {
     crrSchedules: '/_/crr/resume/all',
     crrStatus: '/_/crr/status',
     crrMetricPrefix: '/_/metrics/crr',
+    ingestionSchedules: '/_/ingestion/resume/all',
+    ingestionStatus: '/_/ingestion/status',
+    ingestionMetricPrefix: '/_/metrics/ingestion',
 };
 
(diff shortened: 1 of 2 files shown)


lib/utilities/reportHandler.js
```diff
@@ -323,6 +475,8 @@ function reportHandler(clientIP, req, res, log) {
                 repStatus: results.getReplicationStates,
                 config: cleanup(config),
                 capabilities: getCapabilities(),
+                ingestStats: results.getIngestionInfo.metrics,
```
philipyoo, line 478: if on error, do you expect the value here to be an empty object or undefined?

philipyoo: returning empty object

rahulreddy: I believe empty object if you see other responses.

JianqinWang (approved):

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #1769, opened 2019-03-27
bugfix: ZENKO-1583 Add if-unmodified-since check

bennettbuchanan:
Depends on https://github.com/scality/cloudserver/pull/1769.

Functional tests forthcoming.

High-level description:
* Add a multiplebackend metadata route for fetching a cloud object `LastModified` value.
* Add conditional batch delete support using `IfUnmodifiedSince` logic.
* For Azure locations, use conditional delete with access condition `DateUnModifiedSince `.
* For all other locations (which do not support conditional deletes), if the object has changed since the transition was initiated, we do nothing. Otherwise set a tag on the object (the tag key-value is sent by the garbage collector) indicating that the object has transitioned to a new location and should be deleted. Note: we do not carry out the delete operation ourselves due to a race condition which could result in lost data.

The `IfUnmodifiedSince` value will only be sent by the backbeat garbage collector if the source object is on a non-versioned public cloud.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -811,9 +811,11 @@ function _azureConditionalDelete(request, response, log, cb) {
     const objectGetInfo = _createAzureConditionalDeleteObjectGetInfo(request);
     const reqUids = log.getSerializedUids();
     return dataClient.delete(objectGetInfo, reqUids, err => {
-        if (err && err.statusCode === 412) {
+        if (err && err.code === 412) {
             log.info('precondition for Azure deletion was not met', {
                 method: '_azureConditionalDelete',
+                key: request.objectKey,
+                bucket: request.bucketName,
             });
             return cb(err);
         }
@@ -853,6 +855,8 @@ function _conditionalTagging(request, response, locations, log, cb) {
                 method: '_conditionalTagging',
                 ifUnmodifiedSince,
                 lastModified,
+                key: request.objectKey,
+                bucket: request.bucketName,
             });
             return _respond(response, null, log, cb);
         }
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -63,8 +63,12 @@ function _normalizeBackbeatRequest(req) {
 }
 
 function _isObjectRequest(req) {
-    return ['data', 'metadata', 'multiplebackenddata']
-        .includes(req.resourceType);
+    return [
+        'data',
+        'metadata',
+        'multiplebackenddata',
+        'multiplebackendmetadata',
+    ].includes(req.resourceType);
 }
 
 function _respond(response, payload, log, callback) {
@@ -305,8 +309,44 @@ POST /_/backbeat/multiplebackenddata/<bucket name>/<object key>
     ?operation=completempu
 POST /_/backbeat/multiplebackenddata/<bucket name>/<object key>
     ?operation=puttagging
+GET /_/backbeat/multiplebackendmetadata/<bucket name>/<object key>
 POST /_/backbeat/batchdelete
 */
+
+function _getLastModified(locations, log, cb) {
+    const reqUids = log.getSerializedUids();
+    return dataClient.head(locations, reqUids, (err, data) => {
+        if (err) {
+            log.error('head object request failed', {
+                method: 'headObject',
+                error: err,
+            });
+            return cb(err);
+        }
+        return cb(null, data.LastModified || data.lastModified);
+    });
+}
+
+function headObject(request, response, log, cb) {
+    let locations;
+    try {
+        locations = JSON.parse(request.headers['x-scal-locations']);
+    } catch (e) {
+        const msg = 'x-scal-locations header is invalid';
+        return cb(errors.InvalidRequest.customizeDescription(msg));
+    }
+    if (!Array.isArray(locations)) {
+        const msg = 'x-scal-locations header is invalid';
+        return cb(errors.InvalidRequest.customizeDescription(msg));
+    }
+    return _getLastModified(locations, log, (err, lastModified) => {
+        if (err) {
+            return cb(err);
+        }
+        const dataRetrievalInfo = { lastModified };
+        return _respond(response, dataRetrievalInfo, log, cb);
+    });
+}
 
 function putData(request, response, bucketInfo, objMd, log, callback) {
     let errMessage;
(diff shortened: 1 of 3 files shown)


dora-korpar (approved):
lgtm 😸

philipyoo (approved):

lib/routes/routeBackbeat.js
```diff
@@ -305,9 +309,36 @@ POST /_/backbeat/multiplebackenddata/<bucket name>/<object key>
     ?operation=completempu
 POST /_/backbeat/multiplebackenddata/<bucket name>/<object key>
     ?operation=puttagging
+GET /_/backbeat/multiplebackenddata/<bucket name>/<object key>
```
jonathan-gramain: I think you wanted to write `multiplebackendmetadata`

lib/routes/routeBackbeat.js
```diff
@@ -883,18 +947,19 @@ function routeBackbeat(clientIP, request, response, log) {
         }, 's3', requestContexts);
     }
 
-    const useMultipleBackend = request.resourceType === 'multiplebackenddata';
+    const useMultipleBackend = request.resourceType.includes('multiplebackend');
```
jonathan-gramain: minor: `startswith` could be used instead for more specific check

lib/routes/routeBackbeat.js
```diff
@@ -305,9 +309,36 @@ POST /_/backbeat/multiplebackenddata/<bucket name>/<object key>
...
 POST /_/backbeat/multiplebackenddata/<bucket name>/<object key>
     ?operation=puttagging
+GET /_/backbeat/multiplebackenddata/<bucket name>/<object key>
 POST /_/backbeat/batchdelete
 */
 
+function headObject(request, response, log, callback) {
+    let locations;
+    try {
+        locations = JSON.parse(request.headers['x-scal-locations']);
```
jonathan-gramain, line 333: That should work, as long as we make sure we don't pass a full array of sproxyd locations here (which could be in MBs while HTTP header size is usually limited to a few KB at most).

lib/routes/routeBackbeat.js
```diff
@@ -754,22 +785,54 @@ function deleteObjectTagging(request, response, log, callback) {
...
+    request, response, locations, log, callback) {
+        return async.waterfall([
+            next => {
+                const reqUids = log.getSerializedUids();
+                return dataClient.head(locations, reqUids, next);
+            },
+            (data, next) => {
+                const { headers } = request;
+                if (headers['if-unmodified-since'] !== data.LastModified) {
+                    return next(null);
```
jonathan-gramain: `return next()`

It could be appropriate to add an info log with object info and both LastModified and if-unmodified-since values, so we know that this condition occurred on this object, since we silently skip but it's not a regular operation.

lib/routes/routeBackbeat.js
bennettbuchanan: Okay, I can add a check for that before backbeat sends the request. It appears that sending in the body (as is done with batch-delete) is not supported by the AWS schema when using GET.

bennettbuchanan: Just to follow-up on this. I've updated backbeat to only send one array entry as this is sufficient for the data wrapper to perform the head request.

dora-korpar (approved):

lib/routes/routeBackbeat.js
```diff
@@ -754,22 +794,94 @@ function deleteObjectTagging(request, response, log, callback) {
...
+            },
+        },
+    });
+}
+
+function _azureConditionalDelete(request, response, log, cb) {
+    const objectGetInfo = _createAzureConditionalDeleteObjectGetInfo(request);
+    const reqUids = log.getSerializedUids();
+    return dataClient.delete(objectGetInfo, reqUids, err => {
+        if (err) {
```
jonathan-gramain, line 822: Could check specifically for the precondition error on delete, and log with info level instead of error in such case.

lib/routes/routeBackbeat.js
```diff
@@ -754,22 +794,94 @@ function deleteObjectTagging(request, response, log, callback) {
...
+    });
+}
+
+function _conditionalTagging(request, response, locations, log, cb) {
+    return _getLastModified(locations, log, (err, lastModified) => {
+        if (err) {
+            return cb(err);
+        }
+        const ifUnmodifiedSince = request.headers['if-unmodified-since'];
+        if (ifUnmodifiedSince !== lastModified) {
```
jonathan-gramain: That should work for our purpose, but in theory we should check `ifUnmodifiedSince < lastModified`, to be correct regarding HTTP spec.

lib/routes/routeBackbeat.js
```diff
@@ -754,22 +794,94 @@ function deleteObjectTagging(request, response, log, callback) {
...
             // FIXME: add error type MalformedJSON
             return callback(errors.MalformedPOSTRequest);
         }
-        if (!request || !Array.isArray(request.Locations)) {
+        if (!parsedPayload || !Array.isArray(parsedPayload.Locations)) {
             return callback(errors.MalformedPOSTRequest);
         }
-        const locations = request.Locations;
+        const locations = parsedPayload.Locations;
+        if (request.headers['if-unmodified-since']) {
```
jonathan-gramain: It seems if we ask to delete a specific version but also provide the "if-unmodified-since" header, it will delete the master version instead. I think it should not happen for transition policies but is potentially dangerous, we should always pass the version to delete if it was specified in the location IMHO.

jonathan-gramain: Ignore this, it should be fine here as we pass the locations array, but I think I'll have another related comment regarding versioning and put tagging :)

lib/routes/routeBackbeat.js
```diff
@@ -754,22 +794,94 @@ function deleteObjectTagging(request, response, log, callback) {
...
+        const ifUnmodifiedSince = request.headers['if-unmodified-since'];
+        if (ifUnmodifiedSince !== lastModified) {
+            log.info('object has been modified, skipping tagging operation', {
+                method: '_conditionalTagging',
+                ifUnmodifiedSince,
+                lastModified,
+            });
+            return cb();
+        }
+        return handleTaggingOperation(
```
jonathan-gramain: When we reach this point, are we sure we're going to tag only if the bucket is non-versioned? I don't see where we take this decision explicitly (I would say, relying on the fact that `if-unmodified-since` header is provided is not a proper check for this, I would rely on a version ID being passed in the location instead.

bennettbuchanan: It's conceived that the version check happens in backbeat, so this route would only be called with the `if-unmodified-since` header in the case that there is no version id from the source.

lib/routes/routeBackbeat.js
bennettbuchanan: Okay, that works. I'll still send the error back to backbeat, in which case it sounds like we'd want to do the same in backbeat.

lib/routes/routeBackbeat.js
bennettbuchanan: It's conceived that the version check happens in backbeat. Sound like what you're suggesting is an explicit safety check for a version id in the `locations`? That makes sense to me, but just want to confirm.

lib/routes/routeBackbeat.js
bennettbuchanan: Good point. This is too transition policies specific. Updating.

lib/routes/routeBackbeat.js
```diff
@@ -811,6 +811,12 @@ function _azureConditionalDelete(request, response, log, cb) {
     const objectGetInfo = _createAzureConditionalDeleteObjectGetInfo(request);
     const reqUids = log.getSerializedUids();
     return dataClient.delete(objectGetInfo, reqUids, err => {
+        if (err && err.statusCode === 412) {
+            log.info('precondition for Azure deletion was not met', {
```
jonathan-gramain, line 815: I suggest to log some object info (key, bucket, version) so that the log is more useful for troubleshooting.

lib/routes/routeBackbeat.js
bennettbuchanan: Updated to decouple logic from the two services (i.e. backbeat client and cloudserver's route backbeat API), as per the suggestion.

lib/routes/routeBackbeat.js
```diff
@@ -754,22 +794,121 @@ function deleteObjectTagging(request, response, log, callback) {
...
+}
+
+function _conditionalTagging(request, response, locations, log, cb) {
+    return _getLastModified(locations, log, (err, lastModified) => {
+        if (err) {
+            return cb(err);
+        }
+        const ifUnmodifiedSince = request.headers['if-unmodified-since'];
+        if (new Date(ifUnmodifiedSince) < new Date(lastModified)) {
+            log.info('object has been modified, skipping tagging operation', {
```
jonathan-gramain, line 854: same here to add object info in the log

lib/routes/routeBackbeat.js
```diff
@@ -754,22 +794,121 @@ function deleteObjectTagging(request, response, log, callback) {
...
+    // Azure supports a conditional delete operation.
+    if (location && location.type === 'azure') {
+        return _azureConditionalDelete(request, response, log, cb);
+    }
+    // Other clouds do not support a conditional delete. Instead, we
+    // conditionally put tags to indicate if it should be deleted by the user.
+    return _conditionalTagging(request, response, locations, log, cb);
+}
+
+function _shouldConditionallyDelete(request, locations) {
```
jonathan-gramain, line 885: :+1:

jonathan-gramain (approved):

lib/routes/routeBackbeat.js
bennettbuchanan: Good call, I've added the key and bucket. However there isn't a version id associated with this request as it is already checked for in `_shouldConditionallyDelete`.

lib/routes/routeBackbeat.js
bennettbuchanan: Updated.

bennettbuchanan:
@bert-e approve

PR #1774, opened 2019-04-05
bugfix: ZENKO-1606 Replicate tags from object PUT

bennettbuchanan:
If tags are passed in the header from backbeat during a put object request to the public cloud, we should also send that to the destination cloud.


JianqinWang (approved):

dora-korpar (approved):

bennettbuchanan:
@bert-e approve

alexanderchan-scality (approved):

anurag4DSB (approved):

PR #1779, opened 2019-04-09
bugfix: ZENKO-1606 MPU tagging during replication

bennettbuchanan:
Update MPU route requests to set tags on a target public cloud object.

Depends on https://github.com/scality/Arsenal/pull/773.


dora-korpar (approved):

jonathan-gramain (approved):
As discussed, you can comment on why we need to pass tags both at createMPU and completeMPU stages.

bennettbuchanan:
@bert-e approve

PR #1795, opened 2019-04-20
bf: ZENKO-1585 Zenko userMD header zenko-instance-id

philipyoo:
Apply zenko user metadata header called "x-amz-meta-zenko-instance-id"
on the following operations that create objects or new versions
    - objectPut
    - multiObjectDelete
    - objectCopy
    - initiateMPU

Value of the header is the Zenko deployment hashed instance id

### DIFF
--- a/lib/api/apiUtils/object/applyZenkoUserMD.js
+++ b/lib/api/apiUtils/object/applyZenkoUserMD.js
@@ -1,0 +1,17 @@
+const { zenkoIDHeader } = require('arsenal').constants;
+
+/**
+ * applyZenkoUserMD - if request is within a Zenko deployment, apply user
+ * metadata called "zenko-source" to the object
+ * @param {Object} metaHeaders - user metadata object
+ * @return {undefined}
+ */
+function applyZenkoUserMD(metaHeaders) {
+    if (process.env.REMOTE_MANAGEMENT_DISABLE === '0' &&
+        !metaHeaders[zenkoIDHeader]) {
+        // eslint-disable-next-line no-param-reassign
+        metaHeaders[zenkoIDHeader] = 'zenko';
+    }
+}
+
+module.exports = applyZenkoUserMD;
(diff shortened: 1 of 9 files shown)


lib/api/apiUtils/object/applyZenkoUserMD.js
```diff
@@ -0,0 +1,19 @@
...
+const _config = require('../../../Config').config;
+
+/**
+ * applyZenkoUserMD - if request is within a Zenko deployment, apply user
+ * metadata called "zenko-source" to the object
+ * @param {Object} metaHeaders - user metadata object
+ * @return {undefined}
+ */
+function applyZenkoUserMD(metaHeaders) {
+    if (process.env.REMOTE_MANAGEMENT_DISABLE === '0' &&
```
JianqinWang, line 12: by default, isn't `REMOTE_MANAGEMENT_DISABLE` undefined and unspecified? Maybe it'd be better to check if someone explicitly set `REMOTE_MANAGEMENT_DISABLE=1` and work from there

JianqinWang (commented):
just 1 quick comment, otherwise LGTM

dora-korpar:
What about objectPutCopyPart? x)

lib/api/apiUtils/object/applyZenkoUserMD.js
philipyoo: I think `REMOTE_MANAGEMENT_DISABLE` will only be equal to '0' when orbit is enabled. You're right about this being undefined for cases like running file or mem backend, but we can avoid setting this user md field in those cases

philipyoo:
@dora-korpar I think only the initiateMPU is the one that needs to have the user md field set. It looked like putPart/copyPutPart ops will add to the mpu object created with initiateMPU, and on completeMPU, the user md headers defined on the mpu will be sent to the storage location

JianqinWang (approved):

dora-korpar:
I somehow read objectPut as objectPutPart 🙃

dora-korpar (approved):

philipyoo:
ty @JianqinWang and @dora-korpar :)

philipyoo:
@bert-e approve

PR #1796, opened 2019-04-24
bf: S3C 2120 abort mpu timeout

dora-korpar:
For the abortMPU, listParts, and objectPutCopyPart APIs we did not have a safeguard for checking if the original MPU had been initiated on a legacy version of CloudServer, before location constraints were introduced, which led to typeErrors.
The putPart and completeMPU APIs did have a safeguard implemented. I just added that to others.

### DIFF
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -3,13 +3,14 @@ const async = require('async');
 const { config } = require('../../../Config');
 const constants = require('../../../../constants');
 const data = require('../../../data/wrapper');
+const locationConstraintCheck = require('../object/locationConstraintCheck');
 const { metadataValidateBucketAndObj } =
     require('../../../metadata/metadataUtils');
 const multipleBackendGateway = require('../../../data/multipleBackendGateway');
 const services = require('../../../services');
 
 function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
-    callback) {
+    callback, request) {
     const metadataValMPUparams = {
         authInfo,
         bucketName,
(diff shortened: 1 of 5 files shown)


dora-korpar:
@bert-e reset

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -56,8 +57,22 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+                let location;
+                // if controlling location constraint is not stored in object
+                // metadata, mpu was initiated in legacy S3C, so need to
+                // determine correct location constraint
+                if (!mpuOverviewObj.controllingLocationConstraint) {
+                    const backendInfoObj = locationConstraintCheck(request,
+                        null, destBucket, log);
+                    if (backendInfoObj.err) {
+                        return process.nextTick(() => {
+                            callback(backendInfoObj.err);
```
rahulreddy: This should be calling `next` callback which shortcuts the waterfall to end.

lib/api/listParts.js
```diff
@@ -138,7 +140,22 @@ function listParts(authInfo, request, log, callback) {
...
+                let location;
+                // if controlling location constraint is not stored in object
+                // metadata, mpu was initiated in legacy S3C, so need to
+                // determine correct location constraint
+                if (!mpuOverviewObj.controllingLocationConstraint) {
+                    const backendInfoObj = locationConstraintCheck(request,
+                        null, destBucket, log);
+                    if (backendInfoObj.err) {
+                        return process.nextTick(() => {
+                            callback(backendInfoObj.err);
```
rahulreddy: Same thing, need to call `next` here.

jonathan-gramain: I think you have to call `next(backendInfoObj.err)` to end the async waterfall properly. Also in other parts of the waterfall I see `return next(err, destBucket)`, there seem to be a use for the extra argument in case of error, so likely better to follow the pattern.

jonathan-gramain (changes requested):

tests/multipleBackend/multipartUpload.js
```diff
@@ -1,17 +1,18 @@
+/* eslint-disable indent */
```
rahulreddy: It doesn't look like you have any linter issues, why are you disabling this rule?

rahulreddy (changes requested):

lib/api/listParts.js
dora-korpar: 👍 good catch

rahulreddy:
You can split the commits into tests and fix as separate commits. The first one will prove that the test indeed breaks and the second commit will make sure that the tests pass with the fix.

tests/multipleBackend/multipartUpload.js
dora-korpar: ah meant to remove that. my local linter is going insane 🤷‍♀️

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e force_reset

rahulreddy (approved):

jonathan-gramain (approved):

dora-korpar:
@bert-e approve

PR #1805, opened 2019-05-01
bugfix: S3C-1959 Sanity check for data deletion

rahulreddy:
## Description
Add sanity check for deletion of data keys
### Motivation and context

When metadata service is unstable it's possible that an object has successfully
completed composing, final object written to the destination bucket but metadata
service returned HTTP code 500. In this case, the server returns HTTP code 500 to the client and
if the client retries the complete mpu request, it tries to compose the object again but this
time accidentally deleting data keys from the previous request as it assumes that it is overwriting
an existing object where in reality it is deleting the very own keys of the object it is composing.
This check ensures that the current object being composed does not
accidentally delete current keys.

### Other commits accompanying this PR

- test: process mpu parts
- improvement: avoid coverage use npm test instead
- bugfix: fix typo in test setup

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -140,8 +140,8 @@ stages:
           command: |
             set -ex
             mkdir -p $CIRCLE_TEST_REPORTS/unit
-            npm run unit_coverage
-            npm run unit_coverage_legacy_location
+            npm test
+            npm run test_legacy_location
           env: &shared-vars
             <<: *global-env
             S3_LOCATION_FILE: tests/locationConfig/locationConfigTests.json
--- a/lib/api/apiUtils/object/locationKeysSanityCheck.js
+++ b/lib/api/apiUtils/object/locationKeysSanityCheck.js
@@ -1,0 +1,24 @@
+/**
+* Check keys that exist in the current list which will be used in composing
+* object. This method checks against accidentally removing data keys due to
+* instability from the metadata layer. The check returns true if there was no
+* match and false if at least one key from the previous list exists in the
+* current list
+* @param {array|string} prev - list of keys from the object being overwritten
+* @param {array} curr - list of keys to be used in composing current object
+* @returns {array} list of keys that can be deleted
+*/
+function locationKeysSanityCheck(prev, curr) {
+    if (!prev || prev.length === 0) {
+        return true;
+    }
+    // backwards compatibility check if object is of model version 2
+    if (typeof prev === 'string') {
+        return curr.every(v => v.key !== prev);
+    }
+    const keysMap = {};
+    prev.forEach(v => { keysMap[v.key] = true; });
+    return curr.every(v => !keysMap[v.key]);
+}
+
+module.exports = locationKeysSanityCheck;
--- a/tests/unit/api/apiUtils/processMpuParts.js
+++ b/tests/unit/api/apiUtils/processMpuParts.js
@@ -1,0 +1,45 @@
+const assert = require('assert');
+const { Logger } = require('werelogs');
+const log = new Logger('S3').newRequestLogger();
+const { validateAndFilterMpuParts } =
+    require('../../../../lib/api/apiUtils/object/processMpuParts');
+
+let storedParts;
+let jsonList;
+
+// r - number of parts to remove
+function _buildExpectedResult(r) {
+    const result = [];
+    for (let i = 0; i < r; i++) {
+        jsonList.Part.shift();
+        result.push(...storedParts[i].value.partLocations);
+    }
+    return result;
+}
+
+describe('processMpuParts::validateAndFilterMpuParts', () => {
+    let mpuOverviewKey;
+    let splitter;
+    beforeEach(() => {
+        mpuOverviewKey =
+            '"overview..|..fred..|..8e51eecb51ca4caa96dc4ebd51514f2a"';
+        splitter = '..|..';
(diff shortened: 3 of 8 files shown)


rahulreddy:
I suggest reviewing commit by commit for context.

lib/api/apiUtils/object/locationKeysSanityCheck.js
```diff
@@ -0,0 +1,20 @@
...
+* @param {array} curr - list of keys to be used in composing current object
+* @returns {array} list of keys that can be deleted
+*/
+function locationKeysSanityCheck(prev, curr) {
+    // backwards compatibility check if object is of model version 2
+    if (typeof prev === 'string') {
+        return curr.every(v => v.key !== prev);
+    }
+    const keysList = prev.map(v => v.key);
+    return curr.every(v => !keysList.includes(v.key));
```
jonathan-gramain: Looks like this line has an O(n^2) complexity in the number of parts, I fear it could cause long lock-ups of the process with 1-10K MPUs being completed on top of an existing 1-10K MPU, as the node process could loop up to 100M times without giving hand to the event loop.

I suggest to store all prev keys in a js object first instead of using `Array.map`, to benefit from an efficient lookup.

jonathan-gramain (commented):
I'm good with the philosophy of the fix, just have a technical concern.

lib/api/apiUtils/object/locationKeysSanityCheck.js
rahulreddy: I debated it but I didn't think of the case where a 10K mpu could replace a 10K mpu.

lib/api/apiUtils/object/locationKeysSanityCheck.js
```diff
@@ -0,0 +1,20 @@
...
+* @param {array|string} prev - list of keys from the object being overwritten
+* @param {array} curr - list of keys to be used in composing current object
+* @returns {array} list of keys that can be deleted
+*/
+function locationKeysSanityCheck(prev, curr) {
+    // backwards compatibility check if object is of model version 2
+    if (typeof prev === 'string') {
+        return curr.every(v => v.key !== prev);
+    }
+    const keysList = prev.map(v => v.key);
```
bennettbuchanan: I'm not sure if this case can occur, but we could make an optimization by returning `true` if `keysList.length === 0`.

rahulreddy: I can add a check at the top.

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e reset

lib/api/apiUtils/object/locationKeysSanityCheck.js
```diff
@@ -0,0 +1,24 @@
...
+*/
+function locationKeysSanityCheck(prev, curr) {
+    if (!prev || prev.length === 0) {
+        return true;
+    }
+    // backwards compatibility check if object is of model version 2
+    if (typeof prev === 'string') {
+        return curr.every(v => v.key !== prev);
+    }
+    const keysList = {};
```
bennettbuchanan: ~keysList~ keysHash

eve/main.yml
```diff
@@ -140,8 +140,8 @@ stages:
           command: |
             set -ex
             mkdir -p $CIRCLE_TEST_REPORTS/unit
-            npm run unit_coverage
-            npm run unit_coverage_legacy_location
+            npm test
```
jonathan-gramain, line 143: Was this change intentional?

tests/unit/api/apiUtils/processMpuParts.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    [0, 2, 4].forEach(n => {
+        it(`should filter ${n} parts that are not used in complete mpu`,
+            () => {
+                const expected = _buildExpectedResult(n);
+                const result = validateAndFilterMpuParts(storedParts, jsonList,
+                    mpuOverviewKey, splitter, log);
+                // console.log(expected, result);
+                assert(expected, result);
```
bennettbuchanan: Looks like this is meant to be `assert.deepStrictEqual`?

tests/unit/api/apiUtils/processMpuParts.js
```diff
@@ -0,0 +1,44 @@
...
+        splitter = null;
+    });
+
+    [0, 2, 4].forEach(n => {
+        it(`should filter ${n} parts that are not used in complete mpu`,
+            () => {
+                const expected = _buildExpectedResult(n);
+                const result = validateAndFilterMpuParts(storedParts, jsonList,
+                    mpuOverviewKey, splitter, log);
+                // console.log(expected, result);
```
bennettbuchanan: Extra comment here.

jonathan-gramain: debug comment to be removed

eve/main.yml
rahulreddy: Yes, we removed this in 8.0, 8.1 but did not backport it.

rahulreddy:
@bert-e reset

bennettbuchanan (approved):

jonathan-gramain (approved):

rahulreddy:
@bert-e reset

rahulreddy:
@bert-e approve

PR #1815, opened 2019-05-07
ft: ZENKO-1755 Add hdclient

benoit-a:
## Description

Add hdclient support

### Motivation and context

Hyperdrive is a new storage system by Scality. hdclient is a JS module to access it.

This PR adds "hdclient" as new possible backend for storage.

### Related issues

ZENKO-1755

### DIFF
--- a/lib/utilities/reportHandler.js
+++ b/lib/utilities/reportHandler.js
@@ -39,6 +39,7 @@ function getCapabilities() {
         locationTypeSproxyd: true,
         locationTypeNFS: true,
         locationTypeCephRadosGW: true,
+        locationTypeHyperdriveV2: true,
         preferredReadLocation: true,
         managedLifecycle: true,
         managedLifecycleTransition: true,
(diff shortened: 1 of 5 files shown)


lib/utilities/reportHandler.js
```diff
@@ -39,6 +39,7 @@ function getCapabilities() {
         locationTypeSproxyd: true,
         locationTypeNFS: true,
         locationTypeCephRadosGW: true,
+        locationTypeHyperdrive: true,
```
rachedbenmustapha: We may have over-engineered this but as practice for incomatible format upgrades, we went ahead with full-v2 breakage :) So the Orbit stack will check for the `locationTypeHyperdriveV2` capability in order to show the new version of the hyperdrive settings.

benoit-a: thx, will fix

rachedbenmustapha (approved):
LGTM, provided the hdclient backend itself honors versioning requirements as the location constraint advertises

rahulreddy (approved):

fferrandis (approved):

benoit-a:
@bert-e approve

PR #1831, opened 2019-05-14
bugfix: S3C-2105 Add utapi CRR conditions

bennettbuchanan:
When backbeat updates metadata upon putting a replica during CRR of a new object version or delete marker, push the following metrics:

* storageUtilized
* numberOfObjects
* incomingBytes

No metrics should be pushed for replication of tagging or ACLs.

### DIFF
--- a/lib/routes/utilities/pushReplicationMetric.js
+++ b/lib/routes/utilities/pushReplicationMetric.js
@@ -1,0 +1,35 @@
+const assert = require('assert');
+
+const { ObjectMD } = require('arsenal').models;
+const { pushMetric } = require('../../utapi/utilities');
+
+function shouldPushMetric(prevObjectMD, newObjectMD) {
+    // We only want to update metrics for a destination bucket.
+    if (newObjectMD.getReplicationStatus() !== 'REPLICA') {
+        return false;
+    }
+    try {
+        // Replication of object tags and ACLs should not increment metrics.
+        assert.deepStrictEqual(prevObjectMD.getAcl(), newObjectMD.getAcl());
+        assert.deepStrictEqual(prevObjectMD.getTags(), newObjectMD.getTags());
+    } catch (e) {
+        return false;
+    }
+    return true;
+}
+
+function pushReplicationMetric(prevValue, newValue, bucket, key, log) {
+    const prevObjectMD = new ObjectMD(prevValue);
+    const newObjectMD = new ObjectMD(newValue);
+    if (!shouldPushMetric(prevObjectMD, newObjectMD)) {
+        return undefined;
+    }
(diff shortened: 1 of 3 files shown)


tmacro (approved):

lib/routes/utilities/pushReplicationMetric.js
```diff
@@ -0,0 +1,33 @@
...
+const { ObjectMD } = require('arsenal').models;
+const { pushMetric } = require('../../utapi/utilities');
+
+function shouldPushMetric(objectMD) {
+    if (objectMD.getReplicationStatus() !== 'REPLICA') {
+        return false;
+    }
+    try {
+        const newObjectMD = new ObjectMD();
+        assert.deepStrictEqual(objectMD.getAcl(), newObjectMD.getAcl());
```
jonathan-gramain: Could be worth commenting both that comparing against a pristine object is to check that there is no tag nor ACL present in the object, as well as explaining why we don't want to push metrics in such case. I'm suspecting this is to distinguish normal PUT from PUT_TAGGING&co, in any case it does not seem obvious and deserves a comment I think.

If my assumption applies, wondering if it is possible to end up in a situation where we could count the same object twice, say by putting an empty tag set or ACL, or by putting then clearing them afterwards (not with DELETE_TAGGING but putting an empty set, for example) in which case it would be a problem to solve.

bennettbuchanan: Nice catch, thank you. I've added a comment and instead compared against the previous object metadata. That way we catch any case in which the ACL or tagging is changing (i.e. tagging or ACLs are being replicated).

lib/routes/utilities/pushReplicationMetric.js
```diff
@@ -0,0 +1,35 @@
+const assert = require('assert');
+
+const { ObjectMD } = require('arsenal').models;
+const { pushMetric } = require('../../utapi/utilities');
+
+function shouldPushMetric(prevObjectMd, newObjectMD) {
```
jonathan-gramain: just for the sake of consistency and least surprise, you might want to use `MD` suffix on both variables (case difference here)

jonathan-gramain (approved):

lib/routes/utilities/pushReplicationMetric.js
bennettbuchanan: I agree, updated.

dora-korpar (approved):

bennettbuchanan:
@bert-e help

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #1832, opened 2019-05-15
bugfix: ZENKO-1809 replication from nfs

jonathan-gramain:
* bugfix: ZENKO-1809 flag nfs location type as supporting versioning

Change the supportsVersioning flag to true for NFS locations, as
although versioning is set initially by cloudserver, they should also
support new calls to putBucketVersioning() with Status=Enabled. This
partially fixes replication support by making sure the orbit workflow
can be applied to NFS locations.

* bugfix: ZENKO-1809 allow replication from md-ingestion account

Allow md-ingestion service account to trigger replication on new
objects, so that NFS ingestion buckets are allowed to replicate to
cloud locations.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -116,7 +116,16 @@ const constants = {
     // user metadata header to set object locationConstraint
     objectLocationConstraintHeader: 'x-amz-meta-scal-location-constraint',
     legacyLocations: ['sproxyd', 'legacy'],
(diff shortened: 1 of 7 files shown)


dora-korpar (approved):

JianqinWang (approved):

jonathan-gramain:
@bert-e approve

PR #1925, opened 2019-06-08
bugfix: S3C-2076 Logical operator line break error

bennettbuchanan:
Line break before the logical AND operator prevents proper object attribute reassignment.


bennettbuchanan:
Closing this, and please disregard—this is not a real bug. The configuration had not taken affect for other reasons in my environment.

PR #1930, opened 2019-06-10
ci: only generate version on specific branches

ssalaues:
Only generate product version for appropriate branches all others default to 0.0.0. This is needed because the version has to pass `^([0-9]+(\.[0-9]+){1,3}$)` regex otherwise it will cause issues with CI artifact uploads which uses this for the name generation.

### DIFF
--- a/eve/get_product_version.sh
+++ b/eve/get_product_version.sh
@@ -1,3 +1,13 @@
 #!/bin/sh
 
-cat .git/HEAD | sed 's/.*\///'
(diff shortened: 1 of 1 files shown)


rahulreddy (approved):

JianqinWang (approved):

ssalaues:
@bert-e wait

ssalaues:
@bert-e approve

PR #1934, opened 2019-06-11
bugfix: fix utapi redis configuration parsing

JianqinWang:
Zenko does not have `config.utapi.redis` set by default. 

Fixes https://scality.atlassian.net/browse/ZENKO-1883

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -847,47 +847,50 @@ class Config extends EventEmitter {
             this.utapi.localCache = config.localCache;
             assert(config.redis, 'missing required property of utapi ' +
                 'configuration: redis');
-            if (config.utapi.redis.sentinels) {
-                this.utapi.redis = { sentinels: [], name: null };
+            if (config.utapi.redis) {
+                if (config.utapi.redis.sentinels) {
+                    this.utapi.redis = { sentinels: [], name: null };
 
-                assert(typeof config.utapi.redis.name === 'string',
-                    'bad config: redis sentinel name must be a string');
-                this.utapi.redis.name = config.utapi.redis.name;
+                    assert(typeof config.utapi.redis.name === 'string',
+                        'bad config: redis sentinel name must be a string');
+                    this.utapi.redis.name = config.utapi.redis.name;
 
-                assert(Array.isArray(config.utapi.redis.sentinels),
-                    'bad config: redis sentinels must be an array');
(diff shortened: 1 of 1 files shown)


JianqinWang:
https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/8889

dora-korpar (approved):

lib/Config.js
```diff
@@ -847,47 +847,50 @@ class Config extends EventEmitter {
             this.utapi.localCache = config.localCache;
             assert(config.redis, 'missing required property of utapi ' +
                 'configuration: redis');
-            if (config.utapi.redis.sentinels) {
-                this.utapi.redis = { sentinels: [], name: null };
+            if (config.utapi.redis) {
```
bennettbuchanan, line 850: Just thinking out loud here. 

I'm not sure this is needed. I believe it was intended that the redis and utapi configs were separated to allow other components to access the redis config.

Still, if the utapi config is defined, there also must be a redis config defined (hence [this](https://github.com/scality/cloudserver/pull/1934/files#diff-9280903abc69f09eff9ab36070b6d751R848) assertion). `redis` is just no longer a sub-object of `utapi`.

The redis configuration will be passed to Utapi [here](https://github.com/scality/cloudserver/blob/development/8.0/lib/utapi/utilities.js#L10).

rahulreddy: Although it looks like the config is implying as if utapi will be enabled without Redis, it's not. It's merely checking for redis existence as this particular test runs without enabling Utapi https://github.com/scality/cloudserver/blob/62df2c09c94ac1ab1d3bd48b4cd8c840e03f901d/lib/api/apiUtils/object/locationStorageCheck.js#L4.
Given that we are moving to using Utapi standalone in Zenko, I wouldn't worry too much about it.

rahulreddy (approved):

lib/Config.js
bennettbuchanan: I don’t think Utapi will be enabled without Redis. I mean to point out that any values set here in `config.utapi.redis` will be overwritten by `config.redis` in any case.

Yes, Utapi will be standalone in Zenko—I think that’s another reason adding in `config.utapi.redis` isn’t needed.

JianqinWang:
@bert-e approve

PR #1936, opened 2019-06-12
bugfix: S3C-2243 fix check for location type

rahulreddy:
This fixes the check where the logic should be looking at the type
of location instead of the name to leverage batch delete. It also fixes
the format sent to the sproxydclient which expects and object with keys
as an attribute whose value is an array of sproxyd keys.


anurag4DSB (approved):

alexanderchan-scality (approved):

rahulreddy:
@bert-e approve

PR #1956, opened 2019-06-22
ZENKO-1840 fix secure MongoDB access

nicolas2bert:
The env variable lookup for creds should not be in the entrypoint anymore but in cloudserver directly.

The operator is deploying mounting configmaps (read-only) as config files, containing correct values already, except for these credentials that will come from secrets and will be injected through env vars.


tmacro (approved):

JianqinWang (approved):

ssalaues (approved):

dora-korpar (approved):

anurag4DSB (approved):

nicolas2bert:
@bert-e approve

PR #1975, opened 2019-07-12
bugfix: S3C-1959 add sanity check for data deletion

rahulreddy:
## Description
Add sanity check for data deletion
### Motivation and context

When metadata service is unstable it's possible that an object has successfully
completed composing, final object written to the destination bucket but metadata
service returned HTTP code 500. In this case, the server returns HTTP code 500 to the client and
if the client retries the complete mpu request, it tries to compose the object again but this
time accidentally deleting data keys from the previous request as it assumes that it is overwriting
an existing object where in reality it is deleting the very own keys of the object it is composing.
This check ensures that the current object being composed does not
accidentally delete current keys.

(cherry picked from commit 111f90e500c89477d951ab0b663ba0ec29aaa98d)

### DIFF
--- a/lib/api/apiUtils/object/locationKeysSanityCheck.js
+++ b/lib/api/apiUtils/object/locationKeysSanityCheck.js
@@ -1,0 +1,22 @@
+/**
+* Check keys that exist in the current list which will be used in composing
+* object. This method checks against accidentally removing data keys due to
+* instability from the metadata layer. The check returns true if there was no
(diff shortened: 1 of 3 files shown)


anurag4DSB (approved):

JianqinWang (approved):

rahulreddy:
@bert-e approve

jonathan-gramain (approved):

PR #1980, opened 2019-07-17
improvement: S3C-2293 upgrade NodeJS to v10

anurag4DSB:
## Description
Upgrade NodeJS

Upgrade node as node 6 is not supported anymore
### Related issues
[S3C-2113](https://scality.atlassian.net/browse/S3C-2113)


anurag4DSB:
test

JianqinWang (approved):

jonathan-gramain (approved):

anurag4DSB:
@bert-e force_reset

anurag4DSB:
test

anurag4DSB:
ping

Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN apt-get update \
     && npm install --production \
     && apt-get autoremove --purge -y python git build-essential \
     && rm -rf /var/lib/apt/lists/* \
-    && npm cache clear \
```
tcarmet, line 14: I was the one that added this with @Katherinelaue1 but does anyone knows/understand what it does ? :smile:

ssalaues: I think it's to make the images smaller.

tmacro (approved):

Dockerfile
dora-korpar: It just makes sure that all of the packages are pulled fresh instead of using any versions that might be cached (sometimes even if you upgrade a package version npm will still install whatever version is cached 👎)

ssalaues (approved):

Dockerfile
tcarmet: sorry, I pointed the comment on the wrong line, I was referring to the `--force` option we added.

bennettbuchanan: I believe this is added to resolve an issue installing level-down bindings, it had been added to the README at one point [here](https://github.com/scality/cloudserver#install-js-dependencies) (though it's possible that is now outdated).

In any case, as of npm v5 running `npm cache clear` gives this output:

```
npm ERR! As of npm@5, the npm cache self-heals from corruption issues and data extracted from the cache is guaranteed to be valid. If you want to make sure everything is consistent, use 'npm cache verify' instead. On the other hand, if you're debugging an issue with the installer, you can use `npm install --cache /tmp/empty-cache` to use a temporary cache instead of nuking the actual one.
npm ERR!
npm ERR! If you're sure you want to delete the entire cache, rerun this command with --force.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/bennettbuchanan/.npm/_logs/2019-07-19T18_38_00_157Z-debug.log
```

anurag4DSB: The npm cache self-heals from corruption issues and data extracted from the cache is guaranteed to be valid.
to force a clean we often use npm cache clear --force

tcarmet: oooh ok ok I see. Thanks for sharing, all :)

anurag4DSB:
@bert-e approve

dora-korpar (approved):

PR #1995, opened 2019-07-24
RELENG-2986 applying a 0 burst policy

tcarmet:
TL;DR bumping cpu and memory resources on the CI

On CI workloads, you cannot guess the state of the node you're going
to schedule in, therefore, for functional tests it is advised to not do
any burst on resources.

Here you'll see the requested VS what is actually beeing used for CPU resources on one build
image
note that we use a lot more than we ask for, which is the info that the kubernetes node scheduler use to schedule the pods, and what I'm afraid is that we might schedule in some builds our workload on a already stressed node.

Therefore the 0 burst policy, we request what we need, nothing more, nothing less.

This is what the same chart looks like with the builds from now on.
image

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -155,8 +155,8 @@ stages:
         aggressor: eve/workers/build
         s3: "."
       vars:
-        aggressorMemLimit: "2Gi"
-        s3MemLimit: "2Gi"
+        aggressorMem: "2Gi"
+        s3Mem: "2Gi"
         env:
           <<: *multiple-backend-vars
           <<: *global-env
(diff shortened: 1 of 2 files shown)


tcarmet:
@bert-e wait

jonathan-gramain (approved):

JianqinWang (approved):

tmacro (approved):

ssalaues (approved):

anurag4DSB (approved):

tcarmet:
@bert-e approve

dora-korpar (approved):

PR #1999, opened 2019-07-24
ft: S3C-2277 put bucket policy api

dora-korpar:
Add the basic put bucket policy API and dummies for get and delete (in PRs to follow). Includes updated Arsenal routes, but won't function until a few more changes are made in Arsenal and BUCKET_POLICY env var requirement is removed.

### DIFF
--- a/lib/api/bucketPutPolicy.js
+++ b/lib/api/bucketPutPolicy.js
@@ -1,0 +1,70 @@
+const async = require('async');
+const { BucketPolicy } = require('arsenal').models;
+
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const metadata = require('../metadata/wrapper');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const { errors } = require('arsenal');
+
+/**
+ * bucketPutPolicy - create or update a bucket policy
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+function bucketPutPolicy(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketPutPolicy' });
+
+    const { bucketName } = request;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketOwnerAction',
+    };
+
+    if (!process.env.BUCKET_POLICY) {
+        return callback(errors.NotImplemented);
+    }
+
+    return async.waterfall([
+        next => {
+            const policyJSON = JSON.parse(request.post);
+            const bucketPolicy = new BucketPolicy(policyJSON);
+            // if there was an error getting bucket policy,
+            // returned policyObj will contain 'error' key
+            process.nextTick(() => {
+                const policyObj = bucketPolicy.getBucketPolicy();
+                if (policyObj.error) {
+                    return next(policyObj.error);
+                }
+                return next(null, policyObj);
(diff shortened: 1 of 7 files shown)


lib/api/bucketPutPolicy.js
```diff
@@ -0,0 +1,71 @@
...
+        },
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        if (err) {
+            log.trace('error processing request',
+                { error: err, method: 'bucketPutPolicy' });
+            return callback(err, corsHeaders);
+        }
+        pushMetric('putBucketPolicy', log, {
```
bennettbuchanan: Could be worth commenting out as a TODO or just removing this until the metric is supported.

lib/api/bucketPutPolicy.js
```diff
@@ -0,0 +1,71 @@
...
+
+    return async.waterfall([
+        next => {
+            const policyJSON = JSON.parse(request.post);
+            const bucketPolicy = new BucketPolicy(policyJSON);
+            // if there was an error getting bucket policy,
+            // returned configObj will contain 'error' key
+            process.nextTick(() => {
+                const policyObj = bucketPolicy.getBucketPolicy();
+                return next(policyObj.error || null, policyObj);
```
bennettbuchanan: If the policy object has an error, I'm wondering whether you'll need to find a way to pass a `BucketInfo` instance as the second argument since it will go immediately to the final callback in the waterfall. I'm thinking that `collectCorsHeaders` will then error when trying to call `getCors` [here](https://github.com/scality/cloudserver/blob/development/8.1/lib/utilities/collectCorsHeaders.js#L24). I may be mistaken about something here, so please disregard otherwise.

dora-korpar: That's a good catch, but the `collectCorsHeaders` function just returns an empty object if no bucket is passed to it :)

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

lib/api/bucketPutPolicy.js
bennettbuchanan: Isn't `policyObj` passed as `bucket` to that function in the error condition here?

dora-korpar: No, if there's no error, `policyObj` is just passed to the next function in the waterfall. The var name changes to `bucketPolicy` in the next function, but then that's where the `bucket` var is created

dora-korpar:
@bert-e reset

lib/api/bucketPutPolicy.js
```diff
@@ -0,0 +1,67 @@
...
+        (bucket, bucketPolicy, next) => {
+            bucket.setBucketPolicy(bucketPolicy);
+            metadata.updateBucket(bucket.getName(), bucket, log,
+                err => next(err, bucket));
+        },
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        if (err) {
+            log.trace('error processing request',
```
JianqinWang, line 61: is there a reason why we aren't using `log.error`

lib/api/bucketPutPolicy.js
bennettbuchanan: I mean if there is an error. As in, `policyObj.error` is truthy.

tmacro (approved):

dora-korpar:
@bert-e reset

bennettbuchanan (approved):

dora-korpar:
@bert-e approve

PR #2011, opened 2019-07-26
Feature/s3 c 2283 bucket policy get

dora-korpar:
Add the basic get bucket policy API. Includes updated Arsenal routes, but won't function until a few more changes are made in Arsenal and BUCKET_POLICY env var requirement is removed.

Depends on https://github.com/scality/cloudserver/pull/1999

PR #2012, opened 2019-07-27
Feature/s3 c 2284 bucket policy delete

dora-korpar:
Add the basic delete bucket policy API. Includes updated Arsenal routes, but won't function until a few more changes are made in Arsenal and BUCKET_POLICY env var requirement is removed.

Depends on #1999 and #2011

### DIFF
--- a/lib/api/bucketDeletePolicy.js
+++ b/lib/api/bucketDeletePolicy.js
@@ -1,7 +1,49 @@
-const { errors } = require('arsenal');
+const metadata = require('../metadata/wrapper');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 
+/**
+ * bucketDeletePolicy - Delete the bucket policy
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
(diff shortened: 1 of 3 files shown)


lib/api/bucketDeletePolicy.js
```diff
@@ -0,0 +1,51 @@
...
+        }
+        if (!bucket.getBucketPolicy()) {
+            log.trace('no existing bucket policy', {
+                method: 'bucketDeletePolicy',
+            });
+            // TODO: implement Utapi metric support
+            return callback(null, corsHeaders);
+        }
+        log.trace('deleting bucket policy in metadata');
+        bucket.setBucketPolicy(null);
```
bennettbuchanan, line 38: Thinking out loud here: Maybe we can introduce a new method `deleteBucketPolicy` to `BucketInfo`. I know we've already been passing in `null` to the setter in other APIs so I could be persuaded on the merits of consistency, but it seems more intuitive than passing `null` to the setter to have the effect of deletion.

bennettbuchanan (approved):
LGTM! Left one consideration to turn over 😄

lib/api/bucketDeletePolicy.js
dora-korpar: yeah, it's not a bad idea, i do think it's just keeping the consistency with our other apis and also the convention of just having getters/setters on classes 🤷‍♀

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e approve

JianqinWang (approved):

PR #2013, opened 2019-07-29
improvement: increase the limit of num. of allowed tags

rahulreddy:
## Description
improvement: increase the limit of num. of allowed tags
### Motivation and context

This increases the limit of number of allowed tags on an object
from 10 to 50. This is to be inline and retain compatibility with
AWS S3.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "s3",
-  "version": "7.4.0",
+  "version": "7.4.5",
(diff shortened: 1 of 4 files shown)


JianqinWang (approved):

bennettbuchanan (approved):

anurag4DSB (approved):

rahulreddy:
@bert-e approve

PR #1609, opened 2018-11-01
feat(ZENKO-1270): pfs daemon and backend

giacope:
Check https://github.com/scality/Arsenal/pull/598 for reference.

This PR adds the `Passthrough FileSystem Daemon` (pfsserver.js) and the `PfsClient` external backend. 

See https://github.com/scality/Zenko-RFCs/pull/5.

PR #2025, opened 2019-07-31
ZENKOIO-61 - ft: allow meta-mdonly put for all backends

ssalaues:
Moves the meta-mdonly check into Cloudserver allowing for any supported backend to allow for metadata only operations. This also adds three tests along with enabling the `ft_awssdk_external_backends` test suite to enable the tests in CI.

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -167,11 +167,29 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
 
     const requestLogger =
         logger.newRequestLoggerFromSerializedUids(log.getSerializedUids());
+    const mdOnlyHeader = request.headers['x-amz-meta-mdonly'];
+    const mdOnlySize = request.headers['x-amz-meta-size'];
     return async.waterfall([
         function storeData(next) {
             if (size === 0 && !dontSkipBackend[locationType]) {
                 metadataStoreParams.contentMD5 = constants.emptyFileMd5;
                 return next(null, null, null);
+            }
+            // Handle mdOnlyHeader as a metadata only operation. If
+            // the object in question is actually 0 byte or has a body size
+            // then handle normally.
+            if (mdOnlyHeader === 'true' && mdOnlySize > 0 && size === 0) {
+                log.debug('metadata only operation x-amz-meta-mdonly');
+                const md5 = new Buffer(request.headers
+                    ['x-amz-meta-md5chksum'], 'base64').toString('hex');
+                const dataGetInfo = {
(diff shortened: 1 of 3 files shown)


tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
```diff
@@ -73,7 +73,7 @@ function awsGetCheck(objectKey, s3MD5, awsMD5, location, cb) {
     });
 }
 
-describe('MultipleBackend put object', function testSuite() {
+describe.only('MultipleBackend put object', function testSuite() {
```
ssalaues, line 76: This is because the `ft_awssdk_external_backends` has many failing tests when I added it to the CI. I created ticket ZENKO-2032 to address these failures separately but in the mean time the `.only` is necessary to enable these tests in CI.

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -196,13 +214,11 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
                 dataGetInfoArr[0].cipheredDataKey =
                     cipherBundle.cipheredDataKey;
             }
-            if (dataStoreSize !== undefined) {
```
ssalaues, line 199: This and the `dataStoreMD5` conditions were added specifically for the `mdonly` header use originally but are consolidated with these code changes.
Original PR in which these conditions were added:
https://github.com/scality/cloudserver/pull/1609/files#diff-e45defa8c8c53634533cf165cc9232ceR196

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -196,13 +214,11 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
...
-                metadataStoreParams.size = dataStoreSize;
+            if (mdOnlyHeader === 'true') {
+                metadataStoreParams.size = mdOnlySize;
```
dora-korpar, line 218: does `metadataStoreParams.contentMD5` not need to be updated to `md5`?

ssalaues: Since the MD5 gets passed along in the previous function (as the `calculatedHash` in this function), it gets handled by line 221 `metadataStoreParams.contentMD5 = calculatedHash;`

dora-korpar (approved):

tmacro (approved):

rahulreddy:
@bert-e bypass_build_status

ssalaues:
@bert-e approve

rahulreddy:
I will follow up with a PR that enables the tests.

PR #2044, opened 2019-08-07
Feature/s3 c 2371 apis check bucket policy

dora-korpar:
The first commit adds functionality for checking bucket policies. I removed the request type of `'bucketOwnerAction'`, creating an array of those actions instead so the proper method could be compared to the bucket policy.

The second commits updates each api to include the check for bucket policies. The apis that don't call `isBucketAuthorized` directly use the metadata function `metadataValidateBucket`, which calls the `isBucketAuthorized` function.


lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -0,0 +1,302 @@
...
+    if (Array.isArray(principal.AWS) && principal.AWS.includes(arn)) {
+        return true;
+    }
+    return false;
+}
+
+function checkBucketPolicy(policy, requestType, canonicalID, arn) {
+    let allowed = false;
+    let explicitDeny = false;
+    while (!explicitDeny && policy.Statement.length > 0) {
```
alexanderchan-scality: minor: performance-wise, it should better not to perform `Array.shift`.

alexanderchan-scality (commented):
first pass

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -0,0 +1,302 @@
...
+    if (explicitDeny) {
+        return 'explicitDeny';
+    }
+    return allowed;
+}
+
+function isBucketAuthorized(bucket, requestType, canonicalID, arn) {
+    // Check to see if user is authorized to perform a
+    // particular action on bucket based on ACLs.
+    // TODO: Add IAM checks and bucket policy checks.
```
jonathan-gramain: Can this TODO be removed now?

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -0,0 +1,302 @@
...
+function checkBucketPolicy(policy, requestType, canonicalID, arn) {
+    let allowed = false;
+    let explicitDeny = false;
+    while (!explicitDeny && policy.Statement.length > 0) {
+        const s = policy.Statement[0];
+        const principalMatch = _checkPrincipal(canonicalID, arn, s.Principal);
+        const actionMatch = _checkAction(requestType, s.Action);
+
+        if (principalMatch && actionMatch && s.Effect === 'Deny') {
+            explicitDeny = true;
```
jonathan-gramain: Minor debatable style comments:

- you might instead directly `return 'explicitDeny'` here, and get rid of the `explicitDeny` variable, that would work since you're in a regular `while` loop.

- I personally don't like too much mixing returned types, I think it can be confusing and easy to misinterpret when checking from the calling function (this preference might come from my C programmer background :wink:) instead when a function is not really returning a boolean value, it might make more sense to assign distinct values to all returned cases, which can be strings for example, so you could set the returned value to one of `'allowed'`, `'notAllowed'` or `'denied'` for example, making the contract clearer.

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -0,0 +1,302 @@
...
+function isObjAuthorized(bucket, objectMD, requestType, canonicalID, arn) {
+    const bucketOwner = bucket.getOwner();
+    if (!objectMD) {
+        return false;
+    }
+    if (objectMD['owner-id'] === canonicalID) {
+        return true;
+    }
+    // account is authorized if:
+    // - requesttype is "bucketOwnerAction" (example: for objectTagging) and
```
jonathan-gramain: comment to be updated to reflect the new reality

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -0,0 +1,302 @@
...
+
+// the resources specified in the bucket policy should contain the bucket name
+function validatePolicyResource(bucket) {
+    const bucketName = bucket.getName();
+    const policy = bucket.getBucketPolicy();
+    const bucketArn = `arn:aws:s3:::${bucketName}`;
+
+    return policy.Statement.every(s => {
+        if (Array.isArray(s.Resource)) {
+            return s.Resource.every(r => {
```
jonathan-gramain: you can create a local helper function for the check of a single resource, to factorize between array and string cases.

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -0,0 +1,302 @@
...
+                if (r === bucketArn) {
+                    return true;
+                } else if (r.includes('/')) {
+                    const rSubs = r.split('/');
+                    return rSubs.includes(bucketArn);
```
jonathan-gramain: This check seems to succeed if any of the sub-path components in `r` *is* the bucket ARN, is this intended? I'm surprised as I would expect the sub-path components to have different semantic, and only the first one to represent the bucket ARN, the following ones could be to select objects in the bucket etc.

tmacro (approved):

lib/api/apiUtils/authorization/permissionChecks.js
dora-korpar: Yeah, I could be more explicit and say `rSubs[0]` 👍

lib/api/apiUtils/authorization/permissionChecks.js
dora-korpar: Yeah, I'm not a huge fan of mixing return types either, but I thought it would simplify the return of the calling function. I'll change it to keep the return consistent

jonathan-gramain (approved):

dora-korpar:
@bert-e approve

dora-korpar:
@bert-e reset

tmacro (approved):

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e force_reset

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e force_reset

PR #2055, opened 2019-08-12
bugfix: S3C-2052 Delete orphaned data

bennettbuchanan:
**Context**
At a high-level, the following steps occur in sequence when uploading a MPU part:

1. store the data
2. store the metadata

**Problem**
When step 2 fails, an error is sent to the client and the part isn't included when performing a listing of that MPU. However the data associated with the attempted upload remains stored in the backend: i.e. a data orphan.

**Proposed Change**
When step 2 fails, an attempt is made to rollback the data put operation by sending a delete to the backend with the sproxyd key.

### DIFF
--- a/lib/api/objectPutPart.js
+++ b/lib/api/objectPutPart.js
@@ -7,6 +7,7 @@ const { BackendInfo } = require('./apiUtils/object/BackendInfo');
 const constants = require('../../constants');
 const data = require('../data/wrapper');
 const { dataStore } = require('./apiUtils/object/storeObject');
+const { dataDelete } = require('./apiUtils/object/deleteObject');
 const { isBucketAuthorized } = require('./apiUtils/authorization/aclChecks');
 const kms = require('../kms/wrapper');
 const metadata = require('../metadata/wrapper');
@@ -322,7 +323,12 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                             error: err,
                             method: 'objectPutPart::metadata.putObjectMD',
                         });
-                        return next(err, destinationBucket);
+                        // Make a best effort single attempt to cleanup the
+                        // orphaned data while returning the original error from
+                        // the metadata layer.
+                        return dataDelete(dataGetInfo, log, () => {
+                            next(err, destinationBucket);
+                        });
                     }
                     return next(null, oldLocations, objectLocationConstraint,
                         destinationBucket, hexDigest, prevObjectSize);
--- a/tests/multipleBackend/objectPutPart.js
+++ b/tests/multipleBackend/objectPutPart.js
@@ -30,7 +30,7 @@ const log = new DummyRequestLogger();
 const canonicalID = 'accessKey1';
 const authInfo = makeAuthInfo(canonicalID);
 const namespace = 'default';
-const bucketName = `bucketname-${Date.now}`;
+const bucketName = `bucketname-${Date.now()}`;
 
 const body1 = Buffer.from('I am a body', 'utf8');
 const body2 = Buffer.from('I am a body with a different ETag', 'utf8');
--- a/lib/api/apiUtils/object/deleteObject.js
+++ b/lib/api/apiUtils/object/deleteObject.js
@@ -1,0 +1,16 @@
+const data = require('../../../data/wrapper');
+
+function dataDelete(objectGetInfo, log, cb) {
+    data.delete(objectGetInfo, log, err => {
+        if (err) {
+            log.error('error deleting object data', {
+                error: err,
+                method: 'dataDelete',
+            });
+            return cb(err);
+        }
+        return cb();
+    });
+}
+
+module.exports = { dataDelete };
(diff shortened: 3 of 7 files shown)


lib/api/objectPutPart.js
```diff
@@ -315,18 +345,17 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
-                        log.error('error putting object in mpu bucket', {
-                            error: err,
-                            method: 'objectPutPart::metadata.putObjectMD',
-                        });
-                        return next(err, destinationBucket);
-                    }
-                    return next(null, oldLocations, objectLocationConstraint,
-                        destinationBucket, hexDigest, prevObjectSize);
-                });
+            if (process.env.TEST_API_DATA === 'true') {
```
jonathan-gramain: A less intrusive way to achieve the same could be to instrument the memory backend itself to e.g. set a flag to ask it to delete or not store the next metadata blob to be put, called from the test code itself with a specific API exposed by the memory backend, so that a subsequent GET would fail. This way you're only instrumenting test-related code and no production code needs to be instrumented, and you don't need any extra env variable.

tests/multipleBackend/objectPutPart.js
```diff
@@ -30,7 +30,7 @@ const log = new DummyRequestLogger();
 const canonicalID = 'accessKey1';
 const authInfo = makeAuthInfo(canonicalID);
 const namespace = 'default';
-const bucketName = `bucketname-${Date.now}`;
+const bucketName = `bucketname-${Date.now()}`;
```
jonathan-gramain, line 33: good catch :)

lib/api/objectPutPart.js
```diff
@@ -315,18 +345,17 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
+                // Simulates a metadata failure for testing purposes.
+                memory.buckets.delete(mpuBucketName);
+            }
+            return safePutObjectMD(mpuBucketName, partKey, omVal, log, err => {
```
jonathan-gramain: minor: for clarity I would personally prefer keeping the code flow as-is and adding the rollback code directly in the callback of `putObjectMD`, because it's part of the global `objectPutPart` workflow rather than part of the `putObjectMD` operation. In other words it's hard to define what is a "safe" `putObjectMD` without knowing the context in which it is done (putting data first that might need a rollback).

rahulreddy: I agree and would like a generic method that takes a list of keys that need to be deleted. This can be re-used with the apis that are impacted Object PUT, Object PUT Copy Part, Object PUT Part, Object COPY.

bennettbuchanan: I can agree @jonathan-gramain that it is a bit better to be within the workflow, though it feels more readable to have the steps broken down into smaller pieces. If adding directly to the global `objectPutPart` workflow then, I suppose this means it won't be a common utility.

lib/api/objectPutPart.js
bennettbuchanan: I like that idea! 😄 I've made it so the unit test defines an error that the API will return. It is a small change and simple interface that gives a lot of control for testing purposes. What do you think?

lib/api/objectPutPart.js
jonathan-gramain: The actual deletion function can be common to all backends (the one you created `deleteData` serves this purpose, though it might be augmented to fit other use cases regarding Rahul's comment), but it can be called directly from the workflow code in each place where it's needed, that's what I meant.

jonathan-gramain (approved):

lib/api/objectPutPart.js
```diff
@@ -322,7 +323,12 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                             error: err,
                             method: 'objectPutPart::metadata.putObjectMD',
                         });
-                        return next(err, destinationBucket);
+                        // Make a best effort single attempt to cleanup the
+                        // orphaned data while returning the original error from
+                        // the metadata layer.
+                        return dataDelete(dataGetInfo, log, () => {
```
rahulreddy, line 329: I would still keep this sync, as-in wait for callback. There were reasons before but given that batch delete is now in place we no longer should be adopting this pattern.

rahulreddy (changes requested):

lib/api/objectPutPart.js
bennettbuchanan: The waterfall callback `next` is only called after the `dataDelete` function's callback is invoked. I've preserved the metadata failure error here.

rahulreddy: Thanks, I overlooked that.

rahulreddy (approved):

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #2074, opened 2019-08-14
bugfix: S3C-2052 Delete orphaned data in APIs

bennettbuchanan:
Continuation of https://github.com/scality/cloudserver/pull/2055 for the following APIs:
* objectPut
* objectCopy
* objectCopyPart

**Context**
At a high-level, the following steps occur in sequence when putting an object or part:

1. store the data
2. store the metadata

**Problem**
When step 2 fails, an error is sent to the client and the object or part isn't included when performing a listing. However the data associated with the attempted put remains stored in the backend: i.e. a data orphan.

**Proposed Change**
When step 2 fails, an attempt is made to rollback the data put operation by sending a delete to the backend with the sproxyd key.

### DIFF
--- a/lib/api/apiUtils/object/deleteObject.js
+++ b/lib/api/apiUtils/object/deleteObject.js
@@ -1,9 +1,25 @@
 const data = require('../../../data/wrapper');
 
-function dataDelete(objectGetInfo, log, cb) {
-    data.delete(objectGetInfo, log, err => {
+function dataDelete(locations, method, log, cb) {
+    if (!Array.isArray(locations) || locations.length === 0) {
+        return process.nextTick(() => cb());
+    }
+    if (locations.length === 1) {
+        return data.delete(locations[0], log, err => {
+            if (err) {
+                log.error('error deleting object data', {
+                    error: err,
+                    method: 'dataDelete',
+                });
+                return cb(err);
+            }
+            return cb();
+        });
+    }
+    const dataStoreName = locations[0].dataStoreName;
+    return data.batchDelete(locations, method, dataStoreName, log, err => {
         if (err) {
-            log.error('error deleting object data', {
+            log.error('error batch deleting object data', {
                 error: err,
                 method: 'dataDelete',
             });
--- a/lib/api/objectPutCopyPart.js
+++ b/lib/api/objectPutCopyPart.js
@@ -10,6 +10,7 @@ const { pushMetric } = require('../utapi/utilities');
 const logger = require('../utilities/logger');
 const services = require('../services');
 const setUpCopyLocator = require('./apiUtils/object/setUpCopyLocator');
+const { dataDelete } = require('./apiUtils/object/deleteObject');
 const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
 
 const versionIdUtils = versioning.VersionID;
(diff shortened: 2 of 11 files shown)


lib/api/objectPutCopyPart.js
```diff
@@ -261,7 +262,12 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                     if (err && !err.NoSuchKey) {
                         log.debug('error getting current part (if any)',
                         { error: err });
-                        return next(err);
+                        // Make a best effort single attempt to cleanup the
+                        // orphaned data while returning the original error from
+                        // the metadata layer.
+                        return dataDelete(locations[0], log, () => {
```
jonathan-gramain: only one location to delete here?

lib/api/objectPutCopyPart.js
```diff
@@ -295,7 +301,12 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                     if (err) {
                         log.debug('error storing new metadata',
                         { error: err, method: 'storeNewPartMetadata' });
-                        return next(err);
+                        // Make a best effort single attempt to cleanup the
+                        // orphaned data while returning the original error from
+                        // the metadata layer.
+                        return dataDelete(locations[0], log, () => {
```
jonathan-gramain: same

jonathan-gramain (approved):

jonathan-gramain:
I would also recommend to check that the code is correct when putting or copying 0-byte objects.

lib/api/objectPutCopyPart.js
bennettbuchanan: Yes, from what I saw with this particular operation. I was wondering though what the behavior is if the source data is originally from an MPU. i.e. whether the locations array will be more than one. In which case, the `dataDelete` function could be altered to accommodate that. I'll dig a bit deeper there.

lib/api/objectPutCopyPart.js
bennettbuchanan: https://github.com/scality/cloudserver/pull/2074#discussion_r314416414

lib/api/objectPutCopyPart.js
rahulreddy: You can use the batch delete method instead if you have more than 1 location to delete.

bennettbuchanan: Found that indeed for copies where the source is an MPU object, there will be multiple locations to be cleaned up. I'm moving this over to use `batchDelete`.

lib/api/apiUtils/object/deleteObject.js
```diff
@@ -1,9 +1,25 @@
 const data = require('../../../data/wrapper');
 
-function dataDelete(objectGetInfo, log, cb) {
-    data.delete(objectGetInfo, log, err => {
+function dataDelete(locations, method, log, cb) {
+    if (!Array.isArray(locations) || locations.length === 0) {
```
jonathan-gramain, line 4: I think in earlier versions of the product, `location` in metadata could be a string instead of an array, but not sure if we have to support this here. @rahulreddy an idea?

jonathan-gramain (approved):

lib/api/apiUtils/object/deleteObject.js
rahulreddy: It's good to support the old location format as well. I think at this point it's safe but you can never be too sure.

bennettbuchanan: Okay, I'll move forward with merging and forward porting this PR while also working toward a follow-up, smaller PR to support older locations here.

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e force_reset

bennettbuchanan:
@bert-e approve

dora-korpar (approved):

PR #2091, opened 2019-08-22
bf: S3C 2412 fix bucket pol principal eval

dora-korpar:
In bucket policies, if the `Principal` key is set to an account arn or account id, any account or user arns with matching ids should be allowed.
Added a bunch of tests cases with all the combinations


JianqinWang (approved):

rahulreddy (approved):

dora-korpar:
@bert-e approve

PR #2119, opened 2019-09-09
Bugfix/s3 c 2169/invalid user metadata header

naren-rajendran:

### DIFF
--- a/lib/api/apiUtils/object/checkUserMetadataSize.js
+++ b/lib/api/apiUtils/object/checkUserMetadataSize.js
@@ -1,0 +1,38 @@
+const { maximumMetaHeadersSize,
+    invalidObjectUserMetadataHeader } = require('../../../../constants');
+
+/**
+ * Checks the size of the user metadata in the object metadata and removes
+ * them from the response if the size of the user metadata is larger than
+ * the maximum size allowed. A custome metadata key is added to the response
+ * with the number of user metadata keys not returned as its value
+ * @param {object} responseMetadata - response metadata
+ * @return {object} responseMetaHeaders headers with object metadata to include
+ * in response to client
+ */
+function checkUserMetadataSize(responseMetadata) {
+    let md = {};
+    let userMetadataSize = 0;
+    md = JSON.parse(JSON.stringify(responseMetadata));
+    // collect the user metadata keys from the object metadata
+    const userMetadataHeaders = Object.keys(md)
+        .filter(key => key.startsWith('x-amz-meta-'));
+    // compute the size of all user metadata key and its value
+    userMetadataHeaders.forEach(header => {
+        userMetadataSize += header.length + md[header].length;
+    });
+    // check the size computed against the maximum allowed
+    // if the computed size is greater, then remove all the
+    // user metadata from the response object
+    if (userMetadataSize > maximumMetaHeadersSize) {
+        userMetadataHeaders.forEach(header => {
+            delete md[header];
+        });
+        // add the prescribed/custom metadata with number of user metadata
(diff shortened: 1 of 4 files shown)


lib/api/apiUtils/object/checkUserMetadataSize.js
```diff
@@ -0,0 +1,38 @@
...
+ * the maximum size allowed. A custome metadata key is added to the response
+ * with the number of user metadata keys not returned as its value
+ * @param {object} objMD - object's metadata
+ * @return {object} responseMetaHeaders headers with object metadata to include
+ * in response to client
+ */
+function checkUserMetadataSize(objMD) {
+    let md = {};
+    let userMetadataSize = 0;
+    md = JSON.parse(JSON.stringify(objMD));
```
bennettbuchanan: I think this can be simplified a bit using the public methods exposed in `objMD`. For example, it allows retrieval of user metadata fields with `getUserMetadata` or the whole md simply with `getValue`.

naren-rajendran:
@bert-e reset

lib/api/apiUtils/object/checkUserMetadataSize.js
```diff
@@ -0,0 +1,38 @@
...
+        .filter(key => key.startsWith('x-amz-meta-'));
+    // compute the size of all user metadata key and its value
+    userMetadataHeaders.forEach(header => {
+        userMetadataSize += header.length + md[header].length;
+    });
+    // check the size computed against the maximum allowed
+    // if the computed size is greater, then remove all the
+    // user metadata from the response object
+    if (userMetadataSize > maximumMetaHeadersSize) {
+        userMetadataHeaders.forEach(header => {
```
bennettbuchanan, line 28: Similar to above, this can be simplified by using `clearMetadataValues` method. Also, I'd think doing so has the advantage of keeping the modifications of objMD unified through a single interface.

lib/api/apiUtils/object/checkUserMetadataSize.js
```diff
@@ -0,0 +1,38 @@
...
+function checkUserMetadataSize(objMD) {
+    let md = {};
+    let userMetadataSize = 0;
+    md = JSON.parse(JSON.stringify(objMD));
+    // collect the user metadata keys from the object metadata
+    const userMetadataHeaders = Object.keys(md)
+        .filter(key => key.startsWith('x-amz-meta-'));
+    // compute the size of all user metadata key and its value
+    userMetadataHeaders.forEach(header => {
+        userMetadataSize += header.length + md[header].length;
```
bennettbuchanan, line 22: Is the content stored internally UTF-8 encoded? I suppose the size could grow if it gets encoded down the line, before the response is sent. Just thinking out loud here—not sure about it.

bennettbuchanan (approved):
Overall, LGTM! Left a few considerations.

lib/api/apiUtils/object/checkUserMetadataSize.js
naren-rajendran: @bennettbuchanan from the official aws docs, https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html the acceptable encoding for user metadata is ascii/utf-8 and the 2KB restriction is for the same.

naren-scality:
bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e status

naren-rajendran:
@bert-e help

naren-rajendran:
@bert-e force_reset

PR #2123, opened 2019-09-10
bugfix: S3C-2169 filter invalid user metadata headers

naren-rajendran:

### DIFF
--- a/lib/api/apiUtils/object/checkUserMetadataSize.js
+++ b/lib/api/apiUtils/object/checkUserMetadataSize.js
@@ -1,0 +1,38 @@
+const { maximumMetaHeadersSize,
+    invalidObjectUserMetadataHeader } = require('../../../../constants');
+
+/**
+ * Checks the size of the user metadata in the object metadata and removes
+ * them from the response if the size of the user metadata is larger than
+ * the maximum size allowed. A custome metadata key is added to the response
+ * with the number of user metadata keys not returned as its value
+ * @param {object} responseMetadata - response metadata
+ * @return {object} responseMetaHeaders headers with object metadata to include
+ * in response to client
+ */
+function checkUserMetadataSize(responseMetadata) {
+    let md = {};
+    let userMetadataSize = 0;
+    md = JSON.parse(JSON.stringify(responseMetadata));
+    // collect the user metadata keys from the object metadata
+    const userMetadataHeaders = Object.keys(md)
+        .filter(key => key.startsWith('x-amz-meta-'));
+    // compute the size of all user metadata key and its value
+    userMetadataHeaders.forEach(header => {
+        userMetadataSize += header.length + md[header].length;
+    });
+    // check the size computed against the maximum allowed
+    // if the computed size is greater, then remove all the
+    // user metadata from the response object
+    if (userMetadataSize > maximumMetaHeadersSize) {
+        userMetadataHeaders.forEach(header => {
+            delete md[header];
+        });
+        // add the prescribed/custom metadata with number of user metadata
+        // as its value
+        md[invalidObjectUserMetadataHeader] = userMetadataHeaders.length;
+    }
+    return md;
+}
+
+module.exports = checkUserMetadataSize;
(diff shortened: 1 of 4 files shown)


tcarmet:
ping

bennettbuchanan (approved):

JianqinWang (approved):

dora-korpar (approved):

naren-rajendran:
@bert-e approve

tests/unit/metadata/usermetadatasize/usermetadatasize.js
```diff
@@ -0,0 +1,42 @@
...
+        Object.keys(md).forEach(key => {
+            metadata[`${userMetadataKey}${key}`] = md[key];
+        });
+        userMetadataKeys = Object.keys(metadata)
+            .filter(key => key.startsWith(userMetadataKey)).length;
+    });
+
+    it('Should return user metadata when the size is within limits', () => {
+        const responseMetadata = checkUserMetadataSize(metadata);
+        assert.notEqual(responseMetadata, undefined);
```
nicolas2bert, line 25: Should we also check that no user metadata has been removed and no prescribed/custom metadata has been added?

tests/unit/metadata/usermetadatasize/usermetadatasize.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    it('Should not return user metadata when the size exceeds limit', () => {
+        const firstMetadatKey = `${userMetadataKey}header0`;
+        // add one more byte to be over the limit
+        metadata[firstMetadatKey] = `${metadata[firstMetadatKey]}${'0'}`;
+        const responseMetadata = checkUserMetadataSize(metadata);
+        const invalidHeader
+            = responseMetadata[invalidObjectUserMetadataHeader];
+        const responseMetdataKeys = Object.keys(responseMetadata)
```
nicolas2bert, line 36: typing: responseMet**a**dataKeys :)

tests/unit/metadata/usermetadatasize/usermetadatasize.js
```diff
@@ -0,0 +1,42 @@
...
+            .filter(key => key.startsWith(userMetadataKey));
+        assert.notEqual(responseMetadata, undefined);
+        assert.strictEqual(responseMetdataKeys > 0, false);
```
nicolas2bert, line 39: Is not `responseMetdataKeys` an array?

naren-rajendran: oh yes..I was refactoring and I guess i missed 'length'.. I will make the change. Thanks.

tests/unit/metadata/usermetadatasize/usermetadatasize.js
naren-rajendran: Yes. We should. I will correct the test. Thanks.

lib/api/apiUtils/object/checkUserMetadataSize.js
```diff
@@ -0,0 +1,38 @@
...
+ * the maximum size allowed. A custome metadata key is added to the response
+ * with the number of user metadata keys not returned as its value
+ * @param {object} responseMetadata - response metadata
+ * @return {object} responseMetaHeaders headers with object metadata to include
+ * in response to client
+ */
+function checkUserMetadataSize(responseMetadata) {
+    let md = {};
+    let userMetadataSize = 0;
+    md = JSON.parse(JSON.stringify(responseMetadata));
```
jonathan-gramain, line 16: I think this can be a significant CPU eater when the metadata is not tiny. On the other hand, it seems you would only need to copy if the usermd size is above the limit, hence I recommend to do this "copy" down in the `if (userMetadataSize > maximumMetaHeadersSize)` block.

jonathan-gramain: Or even less costly would be to simply reassign headers in the new object without going through JSON serialization tricks.

dora-korpar (approved):

PR #2124, opened 2019-09-10
bf: S3C 2424 bucket policy api perms

dora-korpar:
A couple of changes to make functionality match AWS:
1. Remove a few actions that are listed in AWS documentation, but aren't actually allowed to be set in a bucket policy after manual testing
2. For the Put, Get, and DeleteBucketPolicy APIs, if the requester is not associated with the bucket account owner, the request should fail with a MethodNotAllowed error, even if permission is explicitly set in the bucket policy. Errors will be tested in end to end tests

The tests only changed to reflect the removal of a couple of actions

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -3,7 +3,6 @@ const constants = require('../../../../constants');
 
 const actionMap = {
     's3:AbortMultipartUpload': 'multipartDelete',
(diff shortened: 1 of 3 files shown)


tmacro (approved):

bennettbuchanan (approved):

dora-korpar:
@bert-e approve

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

PR #2133, opened 2019-09-12
bugfix: S3C-2169 filter invalid user metadata headers

naren-rajendran:

### DIFF
--- a/lib/api/apiUtils/object/checkUserMetadataSize.js
+++ b/lib/api/apiUtils/object/checkUserMetadataSize.js
@@ -11,28 +11,28 @@ const { maximumMetaHeadersSize,
  * in response to client
  */
 function checkUserMetadataSize(responseMetadata) {
-    let md = {};
     let userMetadataSize = 0;
-    md = JSON.parse(JSON.stringify(responseMetadata));
     // collect the user metadata keys from the object metadata
-    const userMetadataHeaders = Object.keys(md)
+    const userMetadataHeaders = Object.keys(responseMetadata)
         .filter(key => key.startsWith('x-amz-meta-'));
     // compute the size of all user metadata key and its value
     userMetadataHeaders.forEach(header => {
-        userMetadataSize += header.length + md[header].length;
+        userMetadataSize += header.length + responseMetadata[header].length;
     });
     // check the size computed against the maximum allowed
     // if the computed size is greater, then remove all the
     // user metadata from the response object
     if (userMetadataSize > maximumMetaHeadersSize) {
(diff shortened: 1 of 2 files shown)


lib/api/apiUtils/object/checkUserMetadataSize.js
```diff
@@ -11,28 +11,29 @@ const { maximumMetaHeadersSize,
...
     userMetadataHeaders.forEach(header => {
-        userMetadataSize += header.length + md[header].length;
+        userMetadataSize += header.length + responseMetadata[header].length;
     });
     // check the size computed against the maximum allowed
     // if the computed size is greater, then remove all the
     // user metadata from the response object
     if (userMetadataSize > maximumMetaHeadersSize) {
+        let md = {};
+        md = Object.assign({}, responseMetadata);
```
bennettbuchanan: Just a minor thing: this can be a constant variable despite the eventual additions or deletions of the object's properties. 😄

tests/unit/metadata/usermetadatasize/usermetadatasize.js
```diff
@@ -22,8 +22,12 @@ describe('Check user metadata size', () => {
 
     it('Should return user metadata when the size is within limits', () => {
         const responseMetadata = checkUserMetadataSize(metadata);
+        const invalidHeader
+            = responseMetadata[invalidObjectUserMetadataHeader];
         assert.notEqual(responseMetadata, undefined);
         assert.strictEqual(userMetadataKeys > 0, true);
+        assert.strictEqual(invalidHeader, undefined);
+        assert.deepStrictEqual(metadata, responseMetadata);
```
bennettbuchanan, line 29: Looks like the `deepStrictEqual` would cover the comparison of `invalidHeader`.

bennettbuchanan (approved):
LGTM! Just a couple minor things that may simplify a bit.

dora-korpar (approved):

lib/api/apiUtils/object/checkUserMetadataSize.js
naren-rajendran: Yes..missed it...it should be `const md = Object.assign({}, responseMetadata);`

tests/unit/metadata/usermetadatasize/usermetadatasize.js
naren-rajendran: Just a sanity check.

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e approve

jonathan-gramain (approved):

PR #2147, opened 2019-09-20
bugfix: S3C-2399 head-object part number size corrections

naren-rajendran:
`head-object` with `--part-number` option returns total object size in the response instead of the specific part size. This is now corrected now and `--part-number` option with value ranging between 1-10000 returns the specific part size.

New unit tests are added for these corrections.

### DIFF
--- a/lib/api/apiUtils/object/partInfo.js
+++ b/lib/api/apiUtils/object/partInfo.js
@@ -1,0 +1,50 @@
+/**
+ * Checks for partNumber in request query and returns the part number
+ * @param {object} query - request query
+ * @return {(Integer|undefined)} - part number, zero if part number is less than
+ * 1 and undefined if no partNumber is found in the request query
+ */
+function getPartNumber(query) {
+    if (query && query.partNumber !== undefined) {
+        return Number.isNaN(query.partNumber) ?
+            0 : Number.parseInt(query.partNumber, 10);
+    }
+    return undefined;
+}
+
+/**
+ * Gets the size of the requested part of the object
+ * @param {object} objMD - object metadata
+ * @param {object} partNumber - part number
+ * @return {(Integer|undefined)} - size of the part or undefined
+ */
+function getPartSize(objMD, partNumber) {
+    let size;
+    if (partNumber && objMD && objMD.location
+        && objMD.location.length >= partNumber) {
+        const locations = [];
+        for (let i = 0; i < objMD.location.length; i++) {
+            const { dataStoreETag } = objMD.location[i];
+            const locationPartNumber =
+                Number.parseInt(dataStoreETag.split(':')[0], 10);
+            // Get all parts that belong to the requested part number
+            if (partNumber === locationPartNumber) {
+                locations.push(objMD.location[i]);
+            } else if (locationPartNumber > partNumber) {
+                break;
+            }
+        }
+        if (locations.length > 0) {
+            const { start } = locations[0];
+            const endLocation = locations[locations.length - 1];
+            const end = endLocation.start + endLocation.size - 1;
+            size = end - start + 1;
+        }
+    }
+    return size;
+}
+
+module.exports = {
+    getPartNumber,
+    getPartSize,
+};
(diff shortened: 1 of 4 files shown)


naren-rajendran:
@bert-e reset

lib/api/objectHead.js
```diff
@@ -75,6 +77,18 @@ function objectHead(authInfo, request, log, callback) {
             if (headerValResult.error) {
                 return callback(headerValResult.error, corsHeaders);
             }
+            const partNumber = getPartNumber(request.query);
+            if (partNumber !== undefined) {
+                if (partNumber < 1 || partNumber > maximumAllowedPartCount) {
```
dora-korpar, line 82: `if partNumber > maximumAllowedPartCount` seems like a check we should have in a few other places too 👍

lib/api/apiUtils/object/partInfo.js
```diff
@@ -0,0 +1,50 @@
...
+    return undefined;
+}
+
+/**
+ * Gets the size of the requested part of the object
+ * @param {object} objMD - object metadata
+ * @param {object} partNumber - part number
+ * @return {(Integer|undefined)} - size of the part or undefined
+ */
+function getPartSize(objMD, partNumber) {
```
dora-korpar, line 21: This is actually a way more efficient and less confusing way of doing it, but just for future reference, we do store part info in metadata as StoredParts, returned with this function: https://github.com/scality/cloudserver/blob/development/8.2/lib/services.js#L650

dora-korpar (approved):

lib/api/apiUtils/object/partInfo.js
```diff
@@ -0,0 +1,50 @@
...
+    let size;
+    if (partNumber && objMD && objMD.location
+        && objMD.location.length >= partNumber) {
+        const locations = [];
+        for (let i = 0; i < objMD.location.length; i++) {
```
jonathan-gramain, line 26: I have a small performance concern here, I'm not sure it's worth trying to optimize this depending on the use case for this function and the associated performance hit, but want to raise the point anyway for brainstorming. This is something that might be worth testing with a specific unit test for example, or even on a lab, to assess the impact.

If the MPU is large, i.e. close to 10000 parts (which can sometimes give more than 10000 locations as we observed in the past, but should not generally go beyond 20K-30K I think), and if the request is for a part number towards the end of the locations array, we might have to loop several thousand times before reaching the part we want. Considering this blocks the JS process entirely during that time, it could be significant enough to slow down everything a little bit (depending on how many such HEAD requests arrive). Also, if one wants to do a HEAD to know the size of each single part in a large MPU for example, it might incur significant load because of this.

One way to improve might be possible by using dichotomy to search since I think the locations array will always be sorted by part number, although multiple locations can refer to the same part number. But we would have to double-check first the assumption that the array is sorted by part number.

lib/api/apiUtils/object/partInfo.js
```diff
@@ -0,0 +1,50 @@
...
+ * Gets the size of the requested part of the object
+ * @param {object} objMD - object metadata
+ * @param {object} partNumber - part number
+ * @return {(Integer|undefined)} - size of the part or undefined
+ */
+function getPartSize(objMD, partNumber) {
+    let size;
+    if (partNumber && objMD && objMD.location
+        && objMD.location.length >= partNumber) {
+        const locations = [];
```
jonathan-gramain, line 25: A more compact syntax (maybe more efficient too? not sure) could be to use `Array.filter()` to create the `locations` array with all matching locations.

lib/api/apiUtils/object/partInfo.js
```diff
@@ -0,0 +1,50 @@
...
+            const locationPartNumber =
+                Number.parseInt(dataStoreETag.split(':')[0], 10);
+            // Get all parts that belong to the requested part number
+            if (partNumber === locationPartNumber) {
+                locations.push(objMD.location[i]);
+            } else if (locationPartNumber > partNumber) {
+                break;
+            }
+        }
+        if (locations.length > 0) {
```
jonathan-gramain, line 37: Are locations refering to a single part always sorted by offset? I think they are, just want to point that it's a requirement for the method to work, and might be worth adding a comment then.

jonathan-gramain (approved):

naren-rajendran:
@bert-e approve

PR #2157, opened 2019-09-25
improvement: ingestion metrics ZENKOIO-100

ssalaues:
Adds two Prometheus gauges for Rclone specific metrics for metrics monitoring purposes. 

Metrics added:
  - a counter of the Ingested Objects 
  - total size of data ingested (not metadata size)

Parallel/related fixes:
- a minor bug with how `httpRequestSizeBytes` is calculated.
- typo causing multipart uploads to not calculate metrics

### DIFF
--- a/lib/api/objectPut.js
+++ b/lib/api/objectPut.js
@@ -103,6 +103,14 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
                     'putObject');
                 return callback(err, responseHeaders);
             }
+            // ingestSize assumes that these custom headers indicate
+            // an ingestion PUT which is a metadata only operation.
+            // Since these headers can be modified client side, they
+            // should be used with caution if needed for precise
+            // metrics.
+            const ingestSize = (request.headers['x-amz-meta-mdonly']
+                && !Number.isNaN(request.headers['x-amz-meta-size']))
+                ? Number.parseInt(request.headers['x-amz-meta-size'], 10) : null;
             const newByteLength = request.parsedContentLength;
 
             // Utapi expects null or a number for oldByteLength:
@@ -134,7 +142,8 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
                 oldByteLength: isVersionedObj ? null : oldByteLength,
             });
             monitoring.promMetrics('PUT', bucketName, '200',
-                'putObject', newByteLength, oldByteLength, isVersionedObj);
+                'putObject', newByteLength, oldByteLength, isVersionedObj,
+                null, ingestSize);
             return callback(null, responseHeaders);
         });
     });
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -11,6 +11,14 @@ const numberOfObjects = new client.Gauge({
     name: 'cloud_server_number_of_objects',
     help: 'Total number of objects',
 });
+const numberOfIngestedObjects = new client.Gauge({
+    name: 'cloud_server_number_of_ingested_objects',
+    help: 'Number of out of band ingestion',
+});
+const dataIngested = new client.Gauge({
+    name: 'cloud_server_data_ingested',
+    help: 'Cumulative size of data ingested in bytes',
+});
 const dataDiskAvailable = new client.Gauge({
     name: 'cloud_server_data_disk_available',
     help: 'Available data disk storage in bytes',
@@ -45,7 +53,7 @@ const httpResponseSizeBytes = new client.Summary({
 
 function promMetrics(requestType, bucketName, code, typeOfRequest,
     newByteLength, oldByteLength, isVersionedObj,
-    numOfObjectsRemoved) {
+    numOfObjectsRemoved, ingestSize) {
     let bytes;
 
     httpRequestsTotal.labels(requestType, 'cloud_server', code).inc();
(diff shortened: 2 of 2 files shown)


tmacro (approved):

lib/api/objectPut.js
```diff
@@ -134,7 +137,8 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
                 oldByteLength: isVersionedObj ? null : oldByteLength,
             });
             monitoring.promMetrics('PUT', bucketName, '200',
-                'putObject', newByteLength, oldByteLength, isVersionedObj);
+                'putObject', newByteLength, oldByteLength, isVersionedObj,
+                undefined, isIngest);
```
JianqinWang: I think it's generally best practice, if you're setting the value, to use `null` over `undefined`

JianqinWang (approved):

lib/api/objectPut.js
ssalaues: Ahh okay, I'll change it. Thank you

ssalaues:
@bert-e wait

lib/utilities/monitoringHandler.js
```diff
@@ -80,10 +92,11 @@ function promMetrics(requestType, bucketName, code, typeOfRequest,
     && code === '200') {
         dataDiskAvailable.inc(newByteLength);
         dataDiskFree.inc(newByteLength);
-        if (numOfObjectsRemoved) {
-            numberOfObjects.dec(numOfObjectsRemoved);
-        } else {
-            numberOfObjects.dec();
+        const objs = numOfObjectsRemoved || 1;
```
ploki, line 95: is it still correct if the Put Object is not a replacement but a creation?

lib/api/objectPut.js
```diff
@@ -103,7 +103,10 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
                     'putObject');
                 return callback(err, responseHeaders);
             }
-            const newByteLength = request.parsedContentLength;
+            const isIngest = request.headers['x-amz-meta-mdonly'] || false;
```
ploki: does it have an adverse effect if a regular S3 client uses these user metadata headers on a put object requests?

ssalaues: If a regular S3 client uses these headers and `x-amz-meta-size` is undefined, then it would cause `newByteLength` to be undefined. I can add a condition to check if the size header is defined.

lib/utilities/monitoringHandler.js
ssalaues: The behavior is the same as before. `dec()` without a value defaults to 1. In this case it only does this for `deleteObject`, `abortMultipartUpload`, `multiObjectDelete` operations so the counter should remain correct.

lib/api/objectPut.js
ploki: oh! that's a good one! I didn't see it. My original thought was about the ability a regular S3 client can have to mess with the prom metrics

ssalaues: Ah, I see your point. Yes, there is a potential to mess with the ingestion metrics if someone were to use both the `size` and `mdonly` headers in the PUT object. Fortunately, these metrics are there to simply provide estimates and are not generally valued as the "source of truth"

bennettbuchanan: The same would apply to the UTAPI metrics which is used for billing purposes. Those metrics need to be precise and probably shouldn't be alterable via user-metadata.

ploki: particularly if you can audit the code for such things

bennettbuchanan: Looks like we could have an easy win by keeping the `newByteLength` value sent to UTAPI different from that needed for the prometheus metric. That way we'd for sure avoid any potential issue with that.

ssalaues: Prometheus isn't used for billings, so I'll update the code to not modify the `newByteLength`.

ssalaues: Yeah exactly.

lib/api/objectPut.js
```diff
@@ -103,7 +103,13 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
                     'putObject');
                 return callback(err, responseHeaders);
             }
-            const newByteLength = request.parsedContentLength;
+            const isIngest = (request.headers['x-amz-meta-mdonly']
+                && !isNaN(request.headers['x-amz-meta-size'])) || false;
```
rahulreddy: `Number.isNaN` is more predictible than the global `isNaN` btw as the global one does type conversion along with testing the value.

lib/api/objectPut.js
```diff
@@ -103,7 +103,13 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
...
+
+            const newByteLength = isIngest ?
+                Number(request.headers['x-amz-meta-size'])
```
rahulreddy: This one has similar caveat. Best way to do conversion is to
```js
Number.parseInt(request.headers['x-amz-meta-size'], 10);
```

rahulreddy (changes requested):
I think we need to give it more thought on how/which metrics in Utapi should be impacted by OOB features.

bennettbuchanan (approved):

rahulreddy (approved):

ssalaues:
@bert-e approve

anurag4DSB (approved):

PR #2163, opened 2019-09-30
bugfix: S3C-2410 Unsupported operations are responded with Not Implem…

naren-rajendran:
More unsupported queries in requests are identified. These are added to the unsupported queries list.

Tests are added to check new unsupported queries.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -83,18 +83,26 @@ const constants = {
         .update('', 'binary').digest('hex'),
 
     // Queries supported by AWS that we do not currently support.
+    // Non-bucket queries
     unsupportedQueries: [
         'accelerate',
         'analytics',
+        'encryption',
         'inventory',
+        'legal-hold',
         'logging',
         'metrics',
         'notification',
+        'object-lock',
         'policy',
+        'policyStatus',
+        'publicAccessBlock',
         'requestPayment',
         'restore',
+        'retention',
         'torrent',
     ],
+
     // Headers supported by AWS that we do not currently support.
     unsupportedHeaders: [
         'x-amz-server-side-encryption',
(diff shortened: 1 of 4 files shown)


constants.js
```diff
@@ -94,7 +94,14 @@ const constants = {
         'requestPayment',
         'restore',
         'torrent',
+        'legal-hold',
```
dora-korpar, line 92: i know the diff will look weird, but can you put these in alphabetical order within the array? I think it'll be better for future code reading :)

lib/utilities/validateQueryAndHeaders.js
```diff
@@ -17,13 +17,27 @@ function _validateKeys(unsupportedKeys, obj) {
...
  * @return {object} - empty object or object with error boolean property
  */
-function validateQueryAndHeaders(reqMethod, reqQuery, reqHeaders, log) {
+function validateQueryAndHeaders(request, log) {
+    const reqMethod = request.method;
+    const reqQuery = request.query;
+    const reqHeaders = request.headers;
+    const isBucketQuery = !request.objectKey;
+    // if the request is at bucket level, check for unsupported bucket queries
+    if (isBucketQuery) {
```
dora-korpar, line 30: I don't know the answer off the top of my head, but if we're distinguishing between bucket queries and not bucket queries, should any of the queries currently in the `UnsupportedQueries` array be in `UnsupportedBucketQueries` instead so this condition is hit?

naren-rajendran: I think, ideally yes, because there are operations/queries that we support at object level and the same is not supported at bucket level, example `tagging`. Having said that I should reorganize the arrays.

naren-rajendran:
@bert-e reset

constants.js
naren-rajendran: Arrays are sorted and organized now.

dora-korpar (approved):

bennettbuchanan (approved):

JianqinWang (approved):

anurag4DSB (approved):

naren-rajendran:
@bert-e approve

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

PR #2190, opened 2019-10-07
Bugfixes: MD5 and VersionID for AWS OOB

ssalaues:
Included are two bug fixes:
- fix ingest MD5 inconsistencies
This commit fixes a bug where cloudserver can crash if there is an ingestion
request without the x-amz-meta-md5chksum header. By assigning null to the value
in the event that it is undefined, this behavior is prevented.
- fix mdOnly version IDs
This commit fixes a bug where "metadata only" puts used for ingestion do not
add the server version ID. Without this, getting older versions of an ingested
object would fail.

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -187,25 +187,28 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
     const mdOnlySize = request.headers['x-amz-meta-size'];
     return async.waterfall([
         function storeData(next) {
-            if (size === 0 && !dontSkipBackend[locationType]) {
-                metadataStoreParams.contentMD5 = constants.emptyFileMd5;
-                return next(null, null, null);
-            }
-            // Handle mdOnlyHeader as a metadata only operation. If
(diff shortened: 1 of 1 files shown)


lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -200,11 +200,12 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
                     const md5 = request.headers['x-amz-meta-md5chksum']
                         ? new Buffer(request.headers['x-amz-meta-md5chksum'],
                         'base64').toString('hex') : null;
+                    const ingestVersion = request.headers['x-amz-version-id'];
```
dora-korpar: It looks like the same header was just put into a variable here?

ssalaues: Yeah, thank you. It must have got lost in my commits for some reason. I just force pushed the correct one.

ssalaues:
@bert-e wait

JianqinWang (approved):

dora-korpar (approved):

ssalaues:
@bert-e reset

ssalaues:
@bert-e approve

PR #2194, opened 2019-10-09
bugfix: S3C-2397 Check invalid range

bennettbuchanan:
An invalid range header (i.e. not of the form `bytes=first-last`) when performing a copy object part operation results in an `InvalidArgument` error when using AWS. This PR introduces the same behavior to CloudServer for compatibility purposes.

Note that the behavior when providing an invalid range with the object GET API results in retrieving the entire object rather than an error. Thus the object GET API is not modified.


bennettbuchanan:
@bert-e reset

tmacro (approved):

dora-korpar (approved):

bennettbuchanan:
@bert-e approve

PR #2201, opened 2019-10-10
bugfix: 2206 x-forwarded-for ip extraction

naren-rajendran:
IAM policy match for IP fails if the S3 requests are made via proxies or from load balancers. `x-forwarded-for` header will be set in the request in such cases and it is used for getting the client IP. `x-forwarded-for` will contain list of IPs including the client IP and the proxy's separated by comma. Changes are made to extract only client IP and forward to vault for policy checks. Unit tests are added to check the new IP extraction utility.

### DIFF
--- a/lib/utilities/requestUtils.js
+++ b/lib/utilities/requestUtils.js
@@ -1,0 +1,35 @@
+const { ipCheck } = require('arsenal');
+const { config } = require('../Config');
+
+/**
+ * getClientIp - Gets the client IP from the request
+ * @param {object} request - http request object
+ * @param {object} s3config - (optional) s3 config
+ * @return {string} - returns client IP from the request
+ */
+function getClientIp(request, s3config) {
+    const clientIp = request.socket.remoteAddress;
+    const requestConfig = (s3config || config).requests;
+    if (requestConfig && requestConfig.viaProxy) {
+        /**
+         * if requests are configured to come via proxy,
+         * check from config which proxies are to be trusted and
+         * which header to be used to extract client IP
+         */
+        if (ipCheck.ipMatchCidrList(requestConfig.trustedProxyCIDRs,
+            clientIp)) {
+            const ipFromHeader
+            // eslint-disable-next-line operator-linebreak
+                = request.headers[requestConfig.extractClientIPFromHeader];
+            if (ipFromHeader && ipFromHeader.trim().length) {
+                return ipFromHeader.split(',')[0].trim();
+            }
+        }
+    }
+
+    return clientIp;
+}
+
+module.exports = {
+    getClientIp,
+};
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -278,6 +278,30 @@ function parseUtapiReindex({ enabled, schedule, sentinel, bucketd }) {
     }
 }
 
+function requestsConfigAssert(requestsConfig) {
+    if (requestsConfig.viaProxy !== undefined) {
+        assert(typeof requestsConfig.viaProxy === 'boolean',
+        'config: invalid requests configuration. viaProxy must be a ' +
+        'boolean');
+
+        if (requestsConfig.viaProxy) {
+            assert(Array.isArray(requestsConfig.trustedProxyCIDRs) &&
+            requestsConfig.trustedProxyCIDRs.length > 0 &&
+            requestsConfig.trustedProxyCIDRs
+                .every(ip => typeof ip === 'string'),
+            'config: invalid requests configuration. ' +
+            'trustedProxyCIDRs must be set if viaProxy is set to true ' +
+            'and must be an array');
+
+            assert(typeof requestsConfig.extractClientIPFromHeader === 'string'
+            && requestsConfig.extractClientIPFromHeader.length > 0,
+            'config: invalid requests configuration. ' +
+            'extractClientIPFromHeader must be set if viaProxy is ' +
+            'set to true and must be a string');
+        }
+    }
+}
+
 /**
  * Reads from a config file and returns the content as a config object
  */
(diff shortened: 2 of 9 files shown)


rachedbenmustapha (commented):
If `X-Forwarded-For` is now used to generally feed to IAM policy evaluation, then it should be opt-in, because anyone can forge this header. Only deployments that use a reverse proxy that is guaranteed to overwrite this header should have this enabled.

rachedbenmustapha (dismissed):
@naren-scality continuing our Slack discussion here.

So essentially, the reasoning is that it's better to refuse service to someone authorized than agree to serve someone unauthorized: denial of service is better than data leak.

The current behavior I don't consider a bug, rather an abundance of caution.

For the legitimate use case where we have a trusted reverse proxy in front of cloudserver, that is **guaranteed** to set/unset the `X-Forwarded-For` header or equivalents (like `X-Real-IP`), it's fine. Since we want to be secure by default (not exposing people who deploy cloudserver as a standalone service without the rest of Zenko), my opinion is that it needs to be opt-in.

A way to make it opt-in would be a yes/no configuration option to allow falling back on headers, that would be no by default, with yes set by our deployment tools. Another (pretty widespread, my preferred solution) way of doing it would be to have a "trusted network" setting for which you allow falling back to the forwarding headers.

As pseudo-code:
```
def getRemoteIP(request):
    trustedProxyHeaderNetwork = config.trustedProxyHeaderNetwork || '127.0.0.1/32'

    if trustedProxyHeaderNetwork.includes(request.remoteIP):
         return request.headers[X-Forwarded-For]

    return request.remoteIP
```

Then our kube deployment tools for example can set `trustedProxyHeaderNetwork` to the pod network.

It will return false-negatives if not properly configured, but that's the secure tradeoff, especially since it's to be used in policy evaluation.

Note that which header should be trusted could be configurable as well, since sometimes it can be `X-Forwarded-For`, sometimes `X-Real-IP` depending on the reverse proxy.

rahulreddy:
Thanks @RachedBenMustapha . I agree with the approach, this fix makes sense when only there is a trusted load-balancer or proxy in front of the Cloudserver and not when it is standalone.

naren-rajendran:
@bert-e reset

tmacro (approved):

config.json
```diff
@@ -77,5 +77,10 @@
 	"host": "localhost",
 	"port": 27018,
 	"database": "metadata"
+    },
+    "requests": {
+        "viaProxy": false,
+        "trustedProxyIPs": [],
```
rachedbenmustapha: `trustedProxyCIDRs` would be more appropriate since we're not going to list ip addresses one by one.

lib/utilities/requestUtils.js
```diff
@@ -0,0 +1,32 @@
...
+        request.connection.socket.remoteAddress);
+
+    const requestConfig = config.requests;
+    if (requestConfig && requestConfig.viaProxy) {
+        /**
+         * if requests are configured to come via proxy,
+         * check from config which IPs are to be trusted and
+         * which header to be used to extract client IP
+        */
+        if (requestConfig.trustedProxyIPs.includes(clientIp)) {
```
rachedbenmustapha: We'd want to test if the trusted network ranges include this ip, not if an ip list contains an ip. In dynamic environments like Kubernetes possible source ip addresses will be in the thousands.

rachedbenmustapha (changes requested):

naren-rajendran:
@bert-e reset

lib/utilities/requestUtils.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * getClientIp - Gets the client IP from the request
+ * @param {object} request - http request object
+ * @param {object} s3config - (optional) s3 config
+ * @return {string} - returns client IP from the request
+ */
+function getClientIp(request, s3config) {
+    const clientIp = (request.connection && request.connection.remoteAddress) ||
+        (request.socket && request.socket.remoteAddress) ||
```
rachedbenmustapha: `request.socket` and `request.connection` seem to be the same thing, `request.socket` being the canonical one.

lib/utilities/requestUtils.js
```diff
@@ -0,0 +1,35 @@
...
+        (request.connection && request.connection.socket &&
+            request.connection.socket.remoteAddress);
```
rachedbenmustapha: In which case would `request.connection.socket` exist?

rachedbenmustapha (changes requested):

lib/utilities/requestUtils.js
naren-rajendran: @RachedBenMustapha yes, I checked the docs they are the same. So reverting to use only `request.socket`.

lib/utilities/requestUtils.js
naren-rajendran: I am not sure. `Vault` had this. I am reverting to use only `request.socket`.

naren-rajendran:
@bert-e reset

lib/utilities/requestUtils.js
```diff
@@ -0,0 +1,31 @@
...
+    const requestConfig = (s3config || config).requests;
+    if (requestConfig && requestConfig.viaProxy) {
+        /**
+         * if requests are configured to come via proxy,
+         * check from config which proxies are to be trusted and
+         * which header to be used to extract client IP
+         */
+        if (ipCheck.ipMatchCidrList(requestConfig.trustedProxyCIDRs,
+            clientIp)) {
+            return request.headers[requestConfig.extractClientIPFromHeader]
```
rachedbenmustapha: It is still possible that the header will be missing or empty, which would crash.

naren-rajendran: I missed it. Corrected now.

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

rachedbenmustapha (approved):

naren-rajendran:
@bert-e approve

PR #2212, opened 2019-10-16
bf: S3C 2392 update aws sdk

dora-korpar:
Upgrades the `aws-sdk` package by 52 versions!
All of the dates have been forced to GMT format because the default return from the sdk is now a UTC date that is not a string, so the `toUTCSTring()` method did not work. Using the `toUTCString()` method on the return as well caused a mismatch in milliseconds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "arsenal": "github:scality/Arsenal#a7b6fc8",
     "async": "~2.5.0",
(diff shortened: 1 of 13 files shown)


JianqinWang (approved):

rahulreddy (approved):

tmacro (approved):

dora-korpar:
@bert-e approve

PR #2235, opened 2019-11-06
docs: ZENKOIO-116 small updates

dashagurova:
This fixes ZENKOIO-116

Started with several small updates to CloudServer documentation like:
- fixed broken links that led nowhere before or to a wrong place
- removed old badges for old CIs
- fixed commands that were using old docker repository - scality/cloudserver, and changed it to zenko/cloudserver
- added an environmental variable to docker run cloudserver without Orbit

We need this because we have questions coming in about cloudserver, it seems people are using it or want to use it, so docs need some love.

### DIFF
--- a/docs/DOCKER.rst
+++ b/docs/DOCKER.rst
@@ -1,13 +1,6 @@
 Docker
 ======
 
--  `Environment Variables <environment-variables>`__
--  `Tunables and setup tips <tunables-and-setup-tips>`__
--  `Examples for continuous integration with Docker 
-   <continuous-integration-with-docker-hosted-cloudserver>`__
--  `Examples for going into production with Docker 
-   <in-production-w-a-Docker-hosted-cloudserver>`__
-
 .. _environment-variables:
 
 Environment Variables
@@ -26,7 +19,7 @@ For multiple data backends, a custom locationConfig.json file is required.
 This file enables you to set custom regions. You must provide associated 
 rest_endpoints for each custom region in config.json.
 
-`Learn more about multiple-backend configurations <./GETTING_STARTED#location-configuration>`__
+`Learn more about multiple-backend configurations <GETTING_STARTED.html#location-configuration>`__
 
 If you are using Scality RING endpoints, refer to your customer documentation.
 
@@ -80,8 +73,7 @@ S3BACKEND=file
 ^^^^^^^^^^^^^^
 
 For stored file data to persist, you must mount Docker volumes
-for both data and metadata. See
-`In Production with a Docker-Hosted CloudServer <in-production-w-a-Docker-hosted-cloudserver>`__
+for both data and metadata. See :ref:`In Production with a Docker-Hosted CloudServer <in-production-w-a-Docker-hosted-cloudserver>`
 
 S3BACKEND=mem
 ^^^^^^^^^^^^^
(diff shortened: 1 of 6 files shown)


docs/DOCKER.rst
```diff
@@ -96,11 +88,22 @@ new.host.com, for example, specify the endpoint with:
 
 .. code-block:: shell
 
-    $ docker run -d --name cloudserver -p 8000:8000 -e ENDPOINT=new.host.com scality/cloudserver
+    $ docker run -d --name cloudserver -p 8000:8000 -e ENDPOINT=new.host.com zenko/cloudserver
 
 .. note:: On Unix-like systems (Linux, OS X, etc.) edit /etc/hosts
    to associate 127.0.0.1 with new.host.com.
 
+REMOTE_MANAGEMENT_DISABLE=1
```
JianqinWang: keeping in line with the other env variable title formats, we could probably just title this section as `REMOTE_MANAGEMENT_DISABLE`

docs/DOCKER.rst
```diff
@@ -96,11 +88,22 @@ new.host.com, for example, specify the endpoint with:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+CloudServer is a part of `Zenko <https://www.zenko.io/>`__. When you run CloudServer standalone it will still try to connect to Orbit by default (browser-based graphical user interface for the Zenko).
+
+Using this variable will default to accessKey1 and verySecretKey1 for credentials and disable the automatic Orbit management:
```
JianqinWang: maybe we could rephrase the opening:

Setting this variable to `1` will default to ...

docs/DOCKER.rst
dashagurova: totally

nicolas2bert:
Just a note (not sure if it should be added to this PR): The destination address of this section https://github.com/scality/cloudserver/blob/development/8.2/README.md#docker should be our Cloudserver Docker image.

JianqinWang (approved):

dashagurova:
@bert-e approve

docs/DOCKER.rst
```diff
@@ -96,11 +88,22 @@ new.host.com, for example, specify the endpoint with:
...
-    $ docker run -d --name cloudserver -p 8000:8000 -e ENDPOINT=new.host.com scality/cloudserver
+    $ docker run -d --name cloudserver -p 8000:8000 -e ENDPOINT=new.host.com zenko/cloudserver
 
 .. note:: On Unix-like systems (Linux, OS X, etc.) edit /etc/hosts
    to associate 127.0.0.1 with new.host.com.
 
+REMOTE_MANAGEMENT_DISABLE
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+CloudServer is a part of `Zenko <https://www.zenko.io/>`__. When you run CloudServer standalone it will still try to connect to Orbit by default (browser-based graphical user interface for the Zenko).
```
rachedbenmustapha: "for the Zenko" sounds weird

rachedbenmustapha (changes requested):

docs/DOCKER.rst
dashagurova: I think so too

dashagurova: done

rachedbenmustapha (approved):
@dashagurova I assume you've tested the instructions manually - since these commands (with the embedded metadata server for example) are not something we do routinely ourselves or for our customers, there could be a regression uncaught by tests.

PR #2243, opened 2019-11-18
improvement: allow sse header for encrypted buckets

naren-rajendran:
### Why is this change required? What problem does it solve?
Allows encryption header (SSE-S3) in object requests if bucket level encryption is enabled. Necessary unit tests are added/updated to support this change.

### DIFF
--- a/lib/api/apiUtils/object/checkEncryption.js
+++ b/lib/api/apiUtils/object/checkEncryption.js
@@ -1,0 +1,29 @@
+/**
+ * Checks the object encryption request against bucket encryption for matching
+ * SSE-S3 configuration
+ * @param {object} objectRequest - http request
+ * @param {object} serverSideEncryption - bucket encryption info
+ * @return {boolean} returns true if the object request has correct SSE-S3
+ * configuration
+ */
+function isValidSSES3(objectRequest, serverSideEncryption) {
+    // x-amz-server-side-encryption is allowed only if bucket
+    // encryption is enabled and if the value is AES256
+    // NOTE: object level encryption is not supported, but we allow
+    // encryption headers in the object request headers!
+    const sseHeader = objectRequest.headers['x-amz-server-side-encryption'];
+    const encryptionAlgorithm = 'AES256';
+
+    const result = ((!serverSideEncryption && sseHeader) ||
+        (serverSideEncryption && sseHeader
+            && sseHeader === encryptionAlgorithm
+            && serverSideEncryption.algorithm !== encryptionAlgorithm) ||
+        (serverSideEncryption && sseHeader
+            && sseHeader !== encryptionAlgorithm
+            && serverSideEncryption.algorithm === encryptionAlgorithm));
+    return !result;
+}
+
+module.exports = {
+    isValidSSES3,
+};
--- a/lib/api/initiateMultipartUpload.js
+++ b/lib/api/initiateMultipartUpload.js
@@ -14,6 +14,7 @@ const validateWebsiteHeader = require('./apiUtils/object/websiteServing')
     .validateWebsiteHeader;
 const { config } = require('../Config');
 const multipleBackendGateway = require('../data/multipleBackendGateway');
+const checkObjectEncryption = require('./apiUtils/object/checkEncryption');
 
 const externalVersioningErrorMessage = 'We do not currently support putting ' +
 'a versioned object to a location-constraint of type Azure.';
@@ -45,6 +46,9 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'initiateMultipartUpload' });
     const bucketName = request.bucketName;
     const objectKey = request.objectKey;
+    const invalidSSEError
+        = errors.InvalidArgument.customizeDescription('The encryption method '
+        + 'specified is not supported');
     // Note that we are using the string set forth in constants.js
     // to split components in the storage
     // of each MPU.  AWS does not restrict characters in object keys so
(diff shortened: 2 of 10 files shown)


lib/api/objectCopy.js
```diff
@@ -325,7 +330,13 @@ function objectCopy(authInfo, request, sourceBucket,
             const isVersionedObj = vcfg && vcfg.Status === 'Enabled';
             const destLocationConstraintName =
                 storeMetadataParams.dataStoreName;
-
+            if ((!serverSideEncryption && sseHeader) ||
```
nicolas2bert: You might want to add a utility function `isValidSSE()` into lib/api/apiUtils for clarity and to avoid code duplication.

tmacro (approved):

tests/functional/kmip/serverside_encryption.js
```diff
@@ -6,21 +6,28 @@ const http = require('http');
 const https = require('https');
 const assert = require('assert');
 const logger = { info: msg => process.stdout.write(`${msg}\n`) };
+const async = require('async');
```
nicolas2bert, line 9: You might want to add test failing if 'x-amz-scal-server-side-encryption' is not equal to 'AES256'.

lib/api/initiateMultipartUpload.js
```diff
@@ -164,6 +169,13 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
         const serverSideEncryption =
             destinationBucket.getServerSideEncryption();
         let cipherBundle = null;
+        if ((!serverSideEncryption && sseHeader) ||
+            (sseHeader === 'AES256' &&
```
nicolas2bert: It seems that it will not return any error if bucket encryption is enabled and `sseHeader` is not equal `'AES256'`

tests/functional/kmip/serverside_encryption.js
```diff
@@ -68,37 +75,185 @@ function _buildS3() {
...
+        Body: 'I am the best content ever',
+    };
+
+    if (encrypt) {
+        Object.assign(params, {
+            ServerSideEncryption: 'AES256',
+        });
+    }
+
+    s3.putObject(params, err => {
```
dora-korpar: super minor, but you could just pass `cb` as the final argument instead of the `err => ...` function

lib/api/initiateMultipartUpload.js
dora-korpar: I think it will, because in that case the `serverSideEncryption.algorithm !== sseHeader` condition would be true

nicolas2bert: I am probably missing something out but if `serverSideEncryption.algorithm = 'AES256' ` and `sseHeader = 'invalidheader'` would it not end up `(false && true ) || (false && true)` ?

dora-korpar: oh yeah, you right. That condition should just check if `sseHeader` exists

tests/functional/kmip/serverside_encryption.js
naren-rajendran: Allowed values for `x-amz-scal-server-side-encryption` are `AES256` and `aws:kms`, we do not allow object level encryption and the change/improvement is to allow the header in object level requests and it has to be `AES256`. I have added unit tests for the utility and it covers this case.

lib/api/apiUtils/object/checkEncryption.js
```diff
@@ -0,0 +1,32 @@
...
+    // NOTE: object level encryption is not supported, but we allow
+    // encryption headers in the object request headers!
+    const sseHeader = objectRequest.headers['x-amz-server-side-encryption'];
+    const encryptionAlgorithm = 'AES256';
+
+    const result = ((!serverSideEncryption
+        && sseHeader === encryptionAlgorithm) ||
+        (!serverSideEncryption && sseHeader
+            && sseHeader !== encryptionAlgorithm) ||
+        (serverSideEncryption && sseHeader
```
nicolas2bert, line 18: If bucket encryption is enabled and if its value is different from `AES256` (i.e. `rsa`), should we forbid any `x-amz-server-side-encryption` header? Maybe add more unit test with bucket encryption equals to rsa.

lib/api/apiUtils/object/checkEncryption.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+function isValidSSES3(objectRequest, serverSideEncryption) {
+    // x-amz-server-side-encryption is allowed only if bucket
+    // encryption is enabled and if the value is AES256
+    // NOTE: object level encryption is not supported, but we allow
+    // encryption headers in the object request headers!
+    const sseHeader = objectRequest.headers['x-amz-server-side-encryption'];
+    const encryptionAlgorithm = 'AES256';
+
+    const result = ((!serverSideEncryption
```
nicolas2bert: I think the first four lines of this conditional expression could be replaced by `(!serverSideEncryption && sseHeader) || ...`

lib/api/apiUtils/object/checkEncryption.js
naren-rajendran: We don't allow other encryption algorithms besides AES256 (from the docs I have checked it is the only one supported and I assume it is the standard). If other values are passed then encryption is not enabled but the bucket is created.

lib/api/apiUtils/object/checkEncryption.js
naren-rajendran: I had it like that..now it is just a bit more descriptive :)

nicolas2bert (approved):

dora-korpar (approved):

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e approve

PR #2247, opened 2019-11-20
bugfix: S3C-2504 rework chunked upload stream handling

jonathan-gramain:
*Note*: for branches development/8.1+, since some storage code has moved to Arsenal, part of the changes will need to be ported there (lib/data/wrapper.js and the dependency update on sproxydclient).

- make sure the V4Transform._transform() callback is always called
  even in case of error. When there is a client error (like a chunked
  stream signature mismatch), continue consuming the data but do not
  send it to sproxyd. Do not emit an 'error' event as this confuses
  the HTTP stack and retains bad client connections,

- instead of an 'error' event, emit a custom 'clientError' event on
  V4Transform in case of client error to allow a cleanup of sproxyd
  connections,

- depend on a fix of sproxydclient that aborts sproxyd requests when
  the input stream is closed.

### DIFF
--- a/lib/api/apiUtils/object/prepareStream.js
+++ b/lib/api/apiUtils/object/prepareStream.js
@@ -7,12 +7,12 @@ const V4Transform = require('../../../auth/streamingV4/V4Transform');
  * accessKey, signatureFromRequest, region, scopeDate, timestamp, and
  * credentialScope (to be used for streaming v4 auth if applicable)
  * @param {RequestLogger} log - the current request logger
- * @param {function} cb - callback containing the result for V4Transform
+ * @param {function} errCb - callback called if an error occurs
  * @return {object|null} - V4Transform object if v4 Auth request, or
  * the original stream, or null if the request has no V4 params but
  * the type of request requires them
  */
-function prepareStream(stream, streamingV4Params, log, cb) {
+function prepareStream(stream, streamingV4Params, log, errCb) {
     if (stream.headers['x-amz-content-sha256'] ===
         'STREAMING-AWS4-HMAC-SHA256-PAYLOAD') {
         if (typeof streamingV4Params !== 'object') {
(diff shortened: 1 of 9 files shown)


jonathan-gramain:
@bert-e reset

rachedbenmustapha (approved):
Changes look good, I'm wondering, in case a checksum or other error occurs:
- how long until the client is notified?
- how is the client notified?
- is AWS S3 behaving the same as us?

jonathan-gramain:
@RachedBenMustapha 

> how long until the client is notified?

He should be notified quickly after the stream issue is detected (like a bad signature), so he will not have to wait until his request payload is entirely sent. But the connection is not cut automatically so he can still send the remaining payload if he wants to, e.g. to reuse the connection afterwards.

> how is the client notified?

The standard HTTP response 403 is sent back with a short message explaining that the signature did not match (at least if a signature issue).

> is AWS S3 behaving the same as us?

I did not test this, but I think it's a good idea so will try it.

jonathan-gramain:
@RachedBenMustapha 

So AWS sends back a similar 403 message as soon as the issue is detected (with more info about when went wrong like the sent signature, string to sign etc. to help troubleshooting). But it then closes the connection.

I realized that the node.js HTTP stack was not dealing with 'error' events correctly when the upload side was attached to a stream and that stream emits an error event, it leaves the HTTP server in a bad state with persistent half-open connections (closed from client side but not server-side), apparently without means to fix the situation until the configured HTTP timeout which will close the server-side half of the connection. So it looked safer to me sending back a regular HTTP status without touching at the connection state.

jonathan-gramain:
@bert-e reset

rahulreddy (approved):

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e approve

PR #2248, opened 2019-11-20
feature: add metrics and secure channel servers

rachedbenmustapha:
This allows using specialized connections for each usage:
- configuration push
- metrics collection (to be later moved to prom scraping)
- secure channel

This way each connection handler can be restricted in what it can
do for reliability and security purposes.

Should be functionally equivalent for current deployment mode, the new scripts are to be used with Zenko Operator.

### DIFF
--- a/bin/metrics_server.js
+++ b/bin/metrics_server.js
@@ -1,0 +1,46 @@
+#!/usr/bin/env node
+'use strict'; // eslint-disable-line strict
+
+const {
+    startWSManagementClient,
+    startPushConnectionHealthCheckServer,
+} = require('../lib/management/push');
+
+const logger = require('../lib/utilities/logger');
+
+const {
+    PUSH_ENDPOINT: pushEndpoint,
+    INSTANCE_ID: instanceId,
+    MANAGEMENT_TOKEN: managementToken,
+} = process.env;
+
+if (!pushEndpoint) {
+    logger.error('missing push endpoint env var');
+    process.exit(1);
+}
+
+if (!instanceId) {
+    logger.error('missing instance id env var');
+    process.exit(1);
+}
+
+if (!managementToken) {
+    logger.error('missing management token env var');
+    process.exit(1);
+}
+
+startPushConnectionHealthCheckServer(err => {
+    if (err) {
+        logger.error('could not start healthcheck server', { error: err });
+        process.exit(1);
+    }
+    const url = `${pushEndpoint}/${instanceId}/ws?metrics=1`;
+    startWSManagementClient(url, managementToken, err => {
+        if (err) {
+            logger.error('connection failed, exiting', { error: err });
+            process.exit(1);
+        }
+        logger.info('no more connection, exiting');
+        process.exit(0);
+    });
+});
--- a/lib/management/push.js
+++ b/lib/management/push.js
@@ -5,6 +5,7 @@ const request = require('request');
 const { URL } = require('url');
 const WebSocket = require('ws');
 const assert = require('assert');
+const http = require('http');
 
 const _config = require('../Config').config;
 const logger = require('../utilities/logger');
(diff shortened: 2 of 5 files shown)


lib/management/push.js
```diff
@@ -244,7 +254,8 @@ function startWSManagementClient(pushEndpoint, instanceId, token) {
             closeChannel(message.getChannelNumber());
             break;
         case CHANNEL_PAYLOAD_MESSAGE:
-            if (_config.browserAccessEnabled) {
+            // browserAccessEnabled defaults to true unless explicitly false
+            if (_config.browserAccessEnabled === undefined || _config.browserAccessEnabled) {
```
jonathan-gramain: minor style: maybe you could write `if (_config.browserAccessEnabled !== false)`

bin/metrics_server.js
```diff
@@ -0,0 +1,46 @@
...
+        logger.error('could not start healthcheck server', { error: err });
+        process.exit(1);
+    }
+    const url = `${pushEndpoint}/${instanceId}/ws?metrics=1`;
+    startWSManagementClient(url, managementToken, err => {
+        if (err) {
+            logger.error('connection failed, exiting', { error: err });
+            process.exit(1);
+        }
+        logger.error('no more connection, exiting');
```
dora-korpar: would it be better to have these logged at the debug or info level? since it's technically not an error, right

lib/management/push.js
```diff
@@ -261,8 +272,30 @@ function addOverlayMessageListener(callback) {
     overlayMessageListener = callback;
 }
 
+function startPushConnectionHealthCheckServer(cb) {
+    const server = http.createServer((req, res) => {
+        const path = parse(req.url).pathname;
```
jonathan-gramain: I think `URL.parse` is considered legacy code (https://nodejs.org/docs/latest-v10.x/api/url.html#url_url_strings_and_url_objects), for being future-proof it might be better to use `new URL` instead and access the object attributes.

dora-korpar (approved):

bin/metrics_server.js
rachedbenmustapha: I was on the fence, because it's not supposed to not be connected and quitting will make it be restarted by the orchestrator so it's not an error at this level but it is going to be the cause of a problem higher up the stack. I'll make it `info` though to avoid confusion.

lib/management/push.js
```diff
@@ -204,15 +207,23 @@ function startWSManagementClient(pushEndpoint, instanceId, token) {
...
+        }
     });
 
     ws.on('error', err => {
+        connected = false;
         logger.error('error from push server connection', {
             error: err,
             errorMessage: err.message,
         });
+        if (cb) {
```
jonathan-gramain: I think it does not matter much because the callback is going to exit the process immediately, but it might be possible to have `cb` being called twice, once in the `error` event handler and after in the `close` event which I think is always emitted when the stream does not end in a normal manner. It could be safer to add a guard for this (e.g. using `jsutil.once` from Arsenal).

jonathan-gramain (approved):

lib/management/push.js
rachedbenmustapha: of course 🤦‍♂

lib/management/push.js
rachedbenmustapha: Yes once disconnected all bets are off, but better safe than sorry I'll wrap the cb with `once` for the benefit of future users.

lib/management/push.js
```diff
@@ -261,8 +272,30 @@ function addOverlayMessageListener(callback) {
...
+            res.write('Connected');
+        } else {
+            res.writeHead(503);
+            res.write('Not Connected');
+        }
+
+        res.end();
+    });
+
+    server.listen(_config.port, cb);
```
nicolas2bert, line 292: I am assuming "push health check" servers and cloudserver will run in different containers and so you did not have to specify new ports for them.

nicolas2bert (approved):

lib/management/push.js
rachedbenmustapha: So in classic nodejs fashion, the `url` member is not actually a valid URL but just a path (`new URL` on it throws), and I guess that's the reason the 10.x `http` module docs still use `parse`. I'll just use `req.url` directly with no parsing since it's for healthchecks that won't be adding any hashes or a query string.

lib/management/push.js
rachedbenmustapha: Yes, it's only used in the 2 new scripts that are not using the whole s3 server/routes and have a different view of what is considered healthy.

rachedbenmustapha:
@bert-e approve

tmacro (approved):

PR #2306, opened 2020-01-10
ZENKO-2277 installing curl and gpg due to parent image upgrade

tcarmet:
as the image node:10-slim received an update and removed the commands
curl and gpg, we lost the ability to build cloudserver.
We're now installing them specifically.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,11 @@ COPY ./yarn.lock /usr/src/app/
 
 WORKDIR /usr/src/app
 
+RUN apt-get update \
+    && apt-get install -y \
+    curl \
+    gnupg2
+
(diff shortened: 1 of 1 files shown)


rachedbenmustapha (approved):
If the base image is not disciplined enough to keep the same set of packages between upgrades, we may want to move away from a floating tag.

tcarmet:
Yeah, I agree with you @RachedBenMustapha 
Should we also update the parent to a more precise reference, like `10.18.0-slim` (it's the same one, but at least we don't magically get breaking changes) 

What do you think?

rachedbenmustapha:
@tcarmet Sounds good! Can be in a later PR though, no urgency.

tcarmet:
I made the change, @RachedBenMustapha
doesn't cost and we might as well fix it while we're here :)

rahulreddy (approved):

PR #2339, opened 2020-02-07
bf: S3C 2544 correct metrics if obj owner diff from delete requester

dora-korpar:
Users other than the bucket owner can put objects into a bucket. On a delete, metrics should reflect the update to the correct user's metrics, instead of only the bucket owner's.

### DIFF
--- a/tests/unit/api/multiObjectDelete.js
+++ b/tests/unit/api/multiObjectDelete.js
@@ -9,76 +9,92 @@ const { ds } = require('../../../lib/data/in_memory/backend');
 const DummyRequest = require('../DummyRequest');
 const { bucketPut } = require('../../../lib/api/bucketPut');
 const objectPut = require('../../../lib/api/objectPut');
+const bucketPutACL = require('../../../lib/api/bucketPutACL');
 
 const log = new DummyRequestLogger();
 const canonicalID = 'accessKey1';
+const altCanonicalID = 'accessKey2';
 const authInfo = makeAuthInfo(canonicalID);
+const altAuthInfo = makeAuthInfo(altCanonicalID);
 const namespace = 'default';
 const bucketName = 'bucketname';
 const postBody = Buffer.from('I am a body', 'utf8');
 const contentLength = 2 * postBody.length;
 const objectKey1 = 'objectName1';
 const objectKey2 = 'objectName2';
+const objectKey3 = 'objectName3';
 const testBucketPutRequest = new DummyRequest({
     bucketName,
     namespace,
     headers: {},
     url: `/${bucketName}`,
 });
+const testPutObjectRequest1 = new DummyRequest({
+    bucketName,
+    namespace,
+    objectKey: objectKey1,
+    headers: {},
+    url: `/${bucketName}/${objectKey1}`,
+}, postBody);
+const testPutObjectRequest2 = new DummyRequest({
+    bucketName,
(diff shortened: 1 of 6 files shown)


tests/unit/api/multiObjectDelete.js
```diff
@@ -134,54 +155,151 @@ describe('getObjMetadataAndDelete function for multiObjectDelete', () => {
...
+                assert.strictEqual(quietSetting, true);
+                assert.deepStrictEqual(errorResults, []);
+                assert.strictEqual(deletedObjStats.requesterIsObjOwner.
+                    numOfObjectsRemoved, 1);
+                assert.strictEqual(deletedObjStats.requesterNotObjOwner.
+                    numOfObjectsRemoved, 2);
+                assert.strictEqual(deletedObjStats.requesterIsObjOwner.
+                    totalContentLengthDeleted, postBody.length);
+                assert.strictEqual(deletedObjStats.requesterNotObjOwner.
+                    totalContentLengthDeleted, contentLength);
```
jonathan-gramain, line 288: minor: you could use `2 * postBody.length` here instead of `contentLength` for clarity (as in this test I don't really get what `contentLength` refers to).

tests/unit/api/multiObjectDelete.js
```diff
@@ -134,54 +155,151 @@ describe('getObjMetadataAndDelete function for multiObjectDelete', () => {
...
+                assert.strictEqual(metadata.keyMaps.get(bucketName)
+                    .has(objectKey1), false);
+                assert.strictEqual(metadata.keyMaps.get(bucketName)
+                    .has(objectKey2), false);
+                assert.strictEqual(metadata.keyMaps.get(bucketName)
+                    .has(objectKey3), false);
+                // call to delete data is async so wait 20 ms to check
```
jonathan-gramain, line 295: 20ms sounds rather short to me even if it's just to wait some node.js scheduling, maybe 200ms would reduce the chances of being flaky? Not sure just an impression since the tests will run on potentially heavily loaded environments where slowdowns are common.

jonathan-gramain (approved):

tests/unit/api/multiObjectDelete.js
dora-korpar: It does seem short lol, but I borrowed that value from the existing test for deletions, and I don't think there have been any issues with flakiness, so I think it should be ok

rahulreddy:
@dora-korpar  Have you tested this fix as a whole with Utapi? If you follow these steps as mentioned in the issue
```
1. create account1 and account2.
2. create bucket1 on account1.
3. put-bucket-acl --bucket bucket1 --grant-write account2.
4. put 10GB file to bucket1 by account2.
5. this makes storageUtilized/numberOfObjects of account2 are updated(numberOfObjects +1, storageUtilized +10GB).
6. delete 10GB file on bucket1 by account1.
7. this makes storageUtilized/numberOfObjects of account1 are updated(numberOfObjects -1, storageUtilized -10GB).
```
If numberOfObjects is 0 for both accounts and you issue `numberOfObjects -1` for account1, it will go negative. The proper way is to update both accounts when you put as well as delete an object, not just when deleting the objects. Otherwise the numbers will not add up.

dora-korpar:
@bert-e wait

tmacro (approved):

PR #2344, opened 2020-02-12
Bugfix/s3 c 2544 diff acct metrics

dora-korpar:
Fix for incorrect metric handling. Previously, metrics were collected solely based on the requester. With this fix, if an account makes a stat-changing request (i.e. puts or deletes an object) in a different account's bucket, the bucket owner's metrics will be updated instead of the requester's.

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -424,6 +424,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
         const xml = convertToXml('completeMultipartUpload', xmlParams);
         pushMetric('completeMultipartUpload', log, {
             authInfo,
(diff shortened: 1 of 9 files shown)


dora-korpar:
@bert-e reset

rahulreddy (changes requested):
Removing `authInfo` altogether means you are no longer sending userId i.e not capturing user level metrics anymore. If you keep it instead of removing it, you can make this change in pushMetric. Let me add a commit to your change.

dora-korpar:
@bert-e reset

naren-rajendran (approved):

rahulreddy (approved):

dora-korpar:
@bert-e approve

PR #2354, opened 2020-02-20
S3C-2597-bucket-policy-can-id

dora-korpar:
Previous behavior: Assumed that canonical ids were different for each IAM entity (i.e. users and accounts). Accordingly, every user within the bucket owner account was basically treated as the bucket owner and granted full permissions on an action, even if there was an explicit bucket policy denying it.

Current behavior: A user within the bucket owner's account will still be allowed all actions by default (this behavior assumes an IAM policy has already allowed the user), unless there is a bucket policy which denies the user, in which case they will receive an Access Denied error.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -206,8 +206,13 @@ function _checkPrincipals(canonicalID, arn, principal) {
     return false;
 }
 
-function checkBucketPolicy(policy, requestType, canonicalID, arn, log) {
+function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, log) {
     let permission = 'defaultDeny';
+    // if requester is user within bucket owner account, actions should be
+    // allowed unless explicitly denied (assumes allowed by IAM policy)
+    if (bucketOwner === canonicalID) {
+        permission = 'allow';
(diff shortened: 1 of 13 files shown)


jonathan-gramain (approved):

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -226,11 +231,18 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, log) {
...
     // Check to see if user is authorized to perform a
     // particular action on bucket based on ACLs.
     // TODO: Add IAM checks
-    if (bucket.getOwner() === canonicalID) {
+    let requesterIsNotUser = true;
+    let arn = null;
+    if (authInfo) {
+        requesterIsNotUser = !authInfo.isRequesterAnIAMUser();
+        arn = authInfo.getArn();
+    }
```
alexanderchan-scality, line 243: minor: duplicate code(https://github.com/scality/cloudserver/pull/2354/files#diff-c3ebea7ca58e537d85230bd2a8d75ec8R270-R274)

```suggestion
const arn = _getArn(authInfo)
```

alexanderchan-scality (approved):

dora-korpar:
@bert-e approve

PR #2454, opened 2020-04-17
bugfix:S3C-2571 enable accept ranges header

ilkescality:
HeadObject supports "Accept-ranges" header for request with `Range` (String | i.e. "bytes=0-9") param.

### DIFF
--- a/lib/api/objectHead.js
+++ b/lib/api/objectHead.js
@@ -1,5 +1,6 @@
 const { errors, s3middleware } = require('arsenal');
 const validateHeaders = s3middleware.validateConditionalHeaders;
+const { parseRange } = require('arsenal/lib/network/http/utils');
 
 const { decodeVersionId } = require('./apiUtils/object/versioning');
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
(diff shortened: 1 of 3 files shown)


lib/utilities/collectResponseHeaders.js
```diff
@@ -45,6 +45,9 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
         responseMetaHeaders['x-amz-server-side-encryption-aws-kms-key-id']
             = objectMD['x-amz-server-side-encryption-aws-kms-key-id'];
     }
+    if (objectMD['accept-ranges']) {
```
jonathan-gramain: Why is this part of `objectMD`? For me it's a server-side capability that is not related to object internal metadata.

ilkescality: I thought so, but wasn't sure, I tried to leave a comment on my original PR to check, removing it then.

tests/unit/api/objectHead.js
```diff
@@ -137,6 +137,27 @@ describe('objectHead API', () => {
         });
     });
 
+    it('should return Accept-Ranges header if request header includes' +
```
miniscruff: your code changes introduced a total of 4 new paths through the method but you only added one new test. Not a requirement but consider adding tests for all paths. Just my general practice anyway.

lib/api/objectHead.js
```diff
@@ -77,8 +78,47 @@ function objectHead(authInfo, request, log, callback) {
             if (headerValResult.error) {
                 return callback(headerValResult.error, corsHeaders);
             }
+            const responseHeaders = collectResponseHeaders(objMD, corsHeaders,
```
miniscruff, line 81: This method is quite long now ( roughly 80 lines ) and although eslint doesn't require it I would suggest splitting it up if possible. I generally recommend no more than about a "screen" tall. In practice this is varied but 20-40 lines is a good range in most cases for me but would, of course, adjust based on language or needs.

tests/unit/api/objectHead.js
ilkescality: done; added new unit tests and all the paths should be covered now

lib/api/objectHead.js
ilkescality: I agree about this; for now, I copied the logic from the objectGet API as it is, in order to to keep those consistent and not to impact the scope of this PR.
I plan to create a follow-up ticket to submit a refactor PR to address your comment as well as some other closely related issues that I encountered.

tmacro (approved):

lib/api/objectHead.js
```diff
@@ -87,10 +127,8 @@ function objectHead(authInfo, request, log, callback) {
                     return callback(errors.InvalidRange, corsHeaders);
                 }
                 // eslint-disable-next-line no-param-reassign
-                objMD['content-length'] = partSize;
+                responseHeaders['content-length'] = partSize;
```
dora-korpar, line 129: minor, but you I don't think you need the eslint comment above this line anymore

ilkescality: good catch

nicolas2bert (approved):

ilkescality:
@bert-e reset

ilkescality:
@bert-e force_reset

ilkescality:
@bert-e approve

rahulreddy (approved):

PR #2472, opened 2020-04-20
bf: S3C 2392 mpu tagging tests

dora-korpar:
Decided a functional test would be enough for this feature (instead of e2e), so adding it.


dora-korpar:
@bert-e reset

PR #2489, opened 2020-04-21
bf: S3C-2801 mpu tagging tests

dora-korpar:
Adds correct error handling for tags put with initiate mpu (previously, tags were evaluated on mpu completion), and adds tests.

### DIFF
--- a/tests/functional/aws-node-sdk/lib/utility/tagging.js
+++ b/tests/functional/aws-node-sdk/lib/utility/tagging.js
@@ -12,6 +12,15 @@ const taggingTests = [
     },
 ];
 
+function generateMultipleTagQuery(numberOfTag) {
+    const tagsArray = [];
+    for (let i = 0; i < numberOfTag; i++) {
+        tagsArray.push(`key${i}=value${i}`);
+    }
+    return tagsArray.join('&');
+}
+
 module.exports = {
     taggingTests,
+    generateMultipleTagQuery,
 };
--- a/tests/functional/aws-node-sdk/test/object/completeMPU.js
+++ b/tests/functional/aws-node-sdk/test/object/completeMPU.js
@@ -1,4 +1,5 @@
 const assert = require('assert');
+const async = require('async');
 const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
 const {
@@ -6,6 +7,7 @@ const {
     versioningEnabled,
     versioningSuspended,
 } = require('../../lib/utility/versioning-util.js');
+const { taggingTests } = require('../../lib/utility/tagging');
 
 const date = Date.now();
 const bucket = `completempu${date}`;
@@ -136,6 +138,71 @@ describe('Complete MPU', () => {
             it('should complete an MPU with fewer parts than were ' +
                 'originally put and should not return a version id', done => {
                 _completeMpuAndCheckVid(uploadId, eTag, undefined, done);
+            });
+        });
+
+        describe('with tags set on initiation', () => {
+            const tagKey = 'keywithtags';
+
+            taggingTests.forEach(test => {
+                it(test.it, done => {
+                    const [key, value] =
+                        [test.tag.key, test.tag.value].map(encodeURIComponent);
+                    const tagging = `${key}=${value}`;
+
+                    async.waterfall([
+                        next => s3.createMultipartUpload({
+                            Bucket: bucket,
+                            Key: tagKey,
+                            Tagging: tagging,
+                        }, (err, data) => {
+                            if (test.error) {
+                                assert.strictEqual(err.code, test.error);
+                                assert.strictEqual(err.statusCode, 400);
+                                next('expected');
+                            }
+                            next(null, data.UploadId);
+                        }),
+                        (uploadId, next) => s3.uploadPart({
+                            Bucket: bucket,
+                            Key: tagKey,
+                            PartNumber: 1,
+                            UploadId: uploadId,
+                            Body: 'foo',
+                        }, (err, data) => {
+                            next(err, data.ETag, uploadId);
+                        }),
+                        (eTag, uploadId, next) => s3.completeMultipartUpload({
+                            Bucket: bucket,
+                            Key: tagKey,
+                            UploadId: uploadId,
+                            MultipartUpload: {
+                                Parts: [{
+                                    ETag: eTag,
+                                    PartNumber: 1,
+                                }],
+                            },
+                        }, next),
+                    ], err => {
+                        if (err === 'expected') {
+                            done();
+                        } else {
+                            assert.ifError(err);
+                            s3.getObjectTagging({
+                                Bucket: bucket,
+                                Key: tagKey,
+                            }, (err, tagData) => {
+                                assert.ifError(err);
+                                assert.deepStrictEqual(tagData.TagSet,
+                                    [{
+                                        Key: test.tag.key,
+                                        Value: test.tag.value,
+                                    }]);
+                                done();
+                            });
+                        }
+                    });
+                });
             });
         });
     });
--- a/tests/functional/aws-node-sdk/test/object/initiateMPU.js
+++ b/tests/functional/aws-node-sdk/test/object/initiateMPU.js
@@ -5,6 +5,7 @@ const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
 const genMaxSizeMetaHeaders
     = require('../../lib/utility/genMaxSizeMetaHeaders');
+const { generateMultipleTagQuery } = require('../../lib/utility/tagging');
 
 const bucket = `initiatempubucket${Date.now()}`;
 const key = 'key';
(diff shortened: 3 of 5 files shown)


dora-korpar:
@bert-e reset

tests/functional/aws-node-sdk/lib/utility/tagging.js
```diff
@@ -28,6 +28,19 @@ const taggingTests = [
...
 ];
 
+function generateMultipleTagQuery(numberOfTag) {
+    let tags = '';
+    let and = '';
+    for (let i = 0; i < numberOfTag; i++) {
+        if (i !== 0) {
+            and = '&';
+        }
+        tags = `key${i}=value${i}${and}${tags}`;
```
jonathan-gramain: minor style: you could also construct an array of query params first, then use `.join('&')` to create the string.

tests/functional/aws-node-sdk/test/object/completeMPU.js
```diff
@@ -138,5 +140,67 @@ describe('Complete MPU', () => {
...
+                        next => s3.createMultipartUpload({
+                            Bucket: bucket,
+                            Key: tagKey,
+                            Tagging: tagging,
+                        }, (err, data) => {
+                            if (test.error) {
+                                assert(err, 'Expected err but found none');
+                                assert.strictEqual(err.code, test.error);
+                                assert.strictEqual(err.statusCode, 400);
+                                done();
```
jonathan-gramain: I don't think it's allowed to shunt the waterfall by calling `done()` directly here, as it would cause the waterfall to be still active, causing potential leaks or timeouts etc., I'd rather call `next(someError)` where `someError` is a specific "error" that is used only to break from the waterfall.

tests/functional/aws-node-sdk/test/object/completeMPU.js
```diff
@@ -138,5 +140,67 @@ describe('Complete MPU', () => {
...
+                            UploadId: uploadId,
+                            MultipartUpload: {
+                                Parts: [{
+                                    ETag: eTag,
+                                    PartNumber: 1,
+                                }],
+                            },
+                        }, next),
+                    ], err => {
+                        assert.equal(err, null, 'Error in mpu setup: ' +
```
jonathan-gramain: You can use instead `assert.ifError(err)`, also this would allow `err` to be `undefined` which is another common case.

tests/functional/aws-node-sdk/test/object/completeMPU.js
```diff
@@ -138,5 +140,67 @@ describe('Complete MPU', () => {
...
+                        `${JSON.stringify(err)}`);
+                        s3.getObjectTagging({
+                            Bucket: bucket,
+                            Key: tagKey,
+                        }, (err, tagData) => {
+                            assert.equal(err, null, 'Expected success, ' +
```
jonathan-gramain: same here

tests/functional/aws-node-sdk/test/object/completeMPU.js
```diff
@@ -138,5 +140,67 @@ describe('Complete MPU', () => {
...
+                                `got error ${JSON.stringify(err)}`);
+                            assert.deepStrictEqual(tagData.TagSet[0], {
```
jonathan-gramain: You could widen the test coverage by checking `tagData.TagSet` itself against the expected array of a single tag element.

tests/functional/aws-node-sdk/test/object/initiateMPU.js
```diff
@@ -65,5 +66,120 @@ describe('Initiate MPU', () => {
                 done();
             });
         });
+
+        describe('with tag set', () => {
```
jonathan-gramain, line 70: I'm thinking those tests could be shared between MPU and non-MPU by adding them in the lib/utility/tagging.js file?

ilkescality (approved):

tests/functional/aws-node-sdk/lib/utility/tagging.js
miniscruff: Agree, would not suggest string concatenations and use array joins instead.

dora-korpar:
@bert-e reset

tests/functional/aws-node-sdk/lib/utility/tagging.js
dora-korpar: any specific reason why?

miniscruff: Mostly I do it for efficiency, every iteration creates a brand new combined string whereas an array + join will only create each item. Each loop will create a new string that is then tossed aside to later be picked up by the garbage collector. We can not really avoid that but we can make it smaller.
compare concating 1-5 with a join.
```
1
12
123
1234
12345
compared to:
1
2
3
4
5
```

miniscruff: Note that this isn't that big a deal in small and infrequent things, but I come from game programming, and running 60 FPS means string work is always bad as it creates a lot of garbage that runs the collector more often. And the more garbage it needs to go through the longer the collection takes which can cause frames to drop. Especially on mobile.

tests/functional/aws-node-sdk/lib/utility/tagging.js
```diff
@@ -12,6 +12,18 @@ const taggingTests = [
     },
 ];
 
+function generateMultipleTagQuery(numberOfTag) {
+    let tagsArray = [];
+    for (let i = 0; i < numberOfTag; i++) {
+        if (i !== 0) {
+            and = '&';
+        }
```
miniscruff: no need for this now

tests/functional/aws-node-sdk/test/object/completeMPU.js
```diff
@@ -138,5 +140,71 @@ describe('Complete MPU', () => {
...
             });
         });
+
+        describe('with tags set on initiation', () => {
+            const tagKey = 'keywithtags';
+
+            taggingTests.forEach(test => {
+                it(test.it, done => {
+                    const key = encodeURIComponent(test.tag.key);
+                    const value = encodeURIComponent(test.tag.value);
```
miniscruff: probably not any better readability wise but you could ( but probably wouldn't want to ) use a map + array destructure
```suggestion
                    const [key, value] = [test.tag.key, test.tag.value].map(encodeURIComponent)
```

tests/functional/aws-node-sdk/test/object/completeMPU.js
```diff
@@ -138,5 +140,71 @@ describe('Complete MPU', () => {
...
+                    const tagging = `${key}=${value}`;
+
+                    async.waterfall([
+                        next => s3.createMultipartUpload({
+                            Bucket: bucket,
+                            Key: tagKey,
+                            Tagging: tagging,
+                        }, (err, data) => {
+                            if (test.error) {
+                                assert(err, 'Expected err but found none');
```
miniscruff: I tend to not include asserts like this that would be checked anyway by the following asserts. Meaning, if this would fail, so would the next check so we can skip it.
Like if we want `a` to equal 5, we do not need to check that its a number first just that its 5.

tests/functional/aws-node-sdk/test/object/initiateMPU.js
```diff
@@ -65,5 +66,118 @@ describe('Initiate MPU', () => {
...
+                }, err => {
+                    assert.ifError(err);
+                    done();
+                });
+            });
+
+            it('should return BadRequest if putting more that 50 tags',
+            done => {
+                const taggingConfig = generateMultipleTagQuery(51);
+                s3.createMultipartUpload({
```
miniscruff, line 99: Is this a promise? I would use assert.rejects if so. Assert.throws would be good if it threw an error instead of just returning it.

tests/functional/aws-node-sdk/test/object/initiateMPU.js
```diff
@@ -65,5 +66,118 @@ describe('Initiate MPU', () => {
...
+                    Bucket: bucket,
+                    Key: key,
+                    Tagging: taggingConfig,
+                }, err => {
+                    assert(err, 'Expected err but did not find one');
```
miniscruff, line 104: If assert.rejects doesn't work, I would remove this assert as it is already checked by the following assert.
This is similar to trying to assert `a` is 5 by first checking `a` is a number and then 5 when just asserting it is 5 would work.

miniscruff (changes requested):
Minor comments, I do get confused on async/await, promise or callbacks though so some stuff may be incorrect.

tests/functional/aws-node-sdk/test/object/initiateMPU.js
dora-korpar: Tis not a promise

miniscruff: dang

miniscruff (approved):

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e approve

PR #2467, opened 2020-04-20
feature: S3C-2773 object lock design document

rahulreddy:
Introduces design document for Object locking feature. The doc is currently not connected to the Read the Docs tree but will be cleaned up later.

PR #2498, opened 2020-04-22
Feature/s3 c 2777 object lock test plan

dora-korpar:
Test plan for Object Lock feature

### DIFF
--- a/docs/OBJECT_LOCK_TEST_PLAN.md
+++ b/docs/OBJECT_LOCK_TEST_PLAN.md
@@ -1,0 +1,161 @@
+# Object Lock Feature Test Plan
+
+## Feature Component Description
+
+Implementing Object Lock will introduce six new APIs:
+
+- putObjectLockConfiguration
+- getObjectLockConfiguration
+- putObjectRetention
+- getObjectRetention
+- putObjectLegalHold
+- getObjectLegalHold
+
+Along with these APIs, putBucket, putObject, deleteObject, and multiObjectDelete
+be affected. In Arsenal, both the BucketInfo and ObjectMD models will be
+updated. Bucket policy and IAM policy permissions will be updated to include
+the new API actions.
+
+## Functional Tests
+
+### putBucket tests
+
+- passing option to enable object lock updates bucket metadata and enables
+    bucket versioning
+
+### putBucketVersioning tests
+
+- suspending versioning on bucket with object lock enabled returns error
+
+### putObject tests
+
+- putting retention configuration on object should be allowed
+- putting invalid retention configuration returns error
+
+### getObject tests
+
+- getting object with retention information should include retention information
+
+### copyObject tests
+
+- copying object with retention information should include retention information
+
+### initiateMultipartUpload tests
+
+- mpu object initiated with retention information should include retention
+    information
+
+### putObjectLockConfiguration tests
+
+- putting configuration as non-bucket-owner user returns AccessDenied error
+- disabling object lock on bucket created with object lock returns error
+- enabling object lock on bucket created without object lock returns
+    InvalidBucketState error
+- enabling object lock with token on bucket created without object lock succeeds
+- putting valid object lock configuration when bucket does not have object
+    lock enabled returns error (InvalidRequest?)
+- putting valid object lock configuration updates bucket metadata
+- putting invalid object lock configuration returns error
+    - ObjectLockEnabled !== "Enabled"
+    - Rule object doesn't contain DefaultRetention key
+    - Mode !== "GOVERNANCE" or "COMPLIANCE"
+    - Days are not an integer
+    - Years are not an integer
+
+### getObjectLockConfiguration tests
+
+- getting configuration as non-bucket-owner user returns AccessDenied error
+- getting configuration when none is set returns
+    ObjectLockConfigurationNotFoundError error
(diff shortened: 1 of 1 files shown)


docs/OBJECT_LOCK_TEST_PLAN.md
```diff
@@ -0,0 +1,157 @@
...
+    - Ensure object cannot be overwritten.
+
+- Create bucket with object locked enabled and put configuration. Put object.
+Put new retention period with date past expiration on object.
+
+    - Ensure object can be deleted.
+    - Ensure object can be overwritten.
+
+- Create bucket with object lock enabled and put configuration with date past
+expiration. Put object. Put legal hold on object.
```
ilkescality: Putting configuration with date past expiration isn't allowed (both in our implementation or per AWS) so this sounds bit misleading, I think could be better to tweak this test a bit such as: 
`Put object in an object lock enabled bucket which has object lock configuration with past expiration date. Put legal hold on object.`

docs/OBJECT_LOCK_TEST_PLAN.md
```diff
@@ -0,0 +1,157 @@
...
+
+    - Ensure object cannot be deleted.
+    - Ensure object cannot be overwritten.
+
+- Create bucket with object lock enabled and put configuration. Put object.
+Check object retention. Change bucket object lock configuration.
+
+    - Ensure object retention period has not changed with bucket configuration.
```
ilkescality, line 152: That's a good one 👍

docs/OBJECT_LOCK_TEST_PLAN.md
```diff
@@ -0,0 +1,157 @@
...
+    - Ensure object can be overwritten.
+
+- Create bucket with object lock enabled and put configuration with date past
+expiration. Put object. Put new retention period on object.
+
+    - Ensure object cannot be deleted.
+    - Ensure object cannot be overwritten.
+
+- Create bucket with object locked enabled and put configuration. Put object.
+Put new retention period with date past expiration on object.
```
ilkescality: I think this test could be misleading as well, as putting object with retention period with date past expiration returns `An error occurred (InvalidArgument) when calling the PutObject operation: The retain until date must be in the future!` error by AWS.

docs/OBJECT_LOCK_TEST_PLAN.md
```diff
@@ -0,0 +1,157 @@
...
+
+    - Ensure user without permission cannot delete object
+    - Ensure user without permission cannot overwrite object
+    - Ensure user with permission can delete object
+    - Ensure user with permission can overwrite object
+    - Ensure user with permission can lengthen retention period
+    - Ensure user with permission cannot shorten retention period
+
+- Create bucket with object lock enabled and put configuration with date past
+expiration. Put object.
```
ilkescality: I think this is another case (putting configuration with date past expiration) which we don't make possible in the first place if I'm not mistaken.

docs/OBJECT_LOCK_TEST_PLAN.md
```diff
@@ -0,0 +1,157 @@
...
+- getObjectRetention
+- putObjectLegalHold
+- getObjectLegalHold
+
+Along with these APIs, putBucket, putObject, deleteObject, and multiObjectDelete
+be affected. In Arsenal, both the BucketInfo and ObjectMD models will be
+updated. Bucket policy and IAM policy permissions will be updated to include
+the new API actions.
+
+## Functional Tests
```
ilkescality, line 19: It'd be good to add functional tests to `getObject` and `copyObject` as well, e.g:
- `getObject` returns the object with `ObjectLockMode` and `ObjectLockRetainUntilDate` if the object has valid object lock configuration in place.
- `copyObject` should copy the object with its object lock configuration if the object has any.

docs/OBJECT_LOCK_TEST_PLAN.md
```diff
@@ -0,0 +1,157 @@
...
+### Scenarios
+
+- Create bucket with object lock enabled. Put object. Put object lock
+configuration. Put another object.
+
+    - Ensure object put before configuration does not have retention period set
+    - Ensure object put after configuration does have retention period set
+
+- Create bucket without object lock. Put object. Enable object lock with token
+and put object lock configuration. Put another object.
```
ilkescality: Are we going to make it possible to be enabled by `token`? In [here](https://github.com/scality/cloudserver/pull/2467/files#diff-f5bc3c260b0f1e4f3bd8617dd2c3a292R187) it suggests that ` we will use a tool to enable object lock by modifying the bucket's configuration` (I'm not sure about what kind of tool though :))

docs/OBJECT_LOCK_TEST_PLAN.md
dora-korpar: yknow you're right. we should still be testing the scenarios that have an expired retention period 🤔 it'll just require manually updating the metadata instead of going through the api. but good to note

ilkescality: Yeah that's right; we'd still wanna test it to verify the behavior in that scenario :)

ilkescality (approved):

rahulreddy (approved):

dora-korpar:
@bert-e approve

PR #2518, opened 2020-05-04
Feature/s3 c 2789 put object lock configuration

dora-korpar:
Implements PutObjectLockConfiguration API. Depends on https://github.com/scality/Arsenal/pull/992, https://github.com/scality/Arsenal/pull/995, & https://github.com/scality/Arsenal/pull/996 in Arsenal.
Functional tests to follow in another PR

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -22,6 +22,7 @@ const bucketPutWebsite = require('./bucketPutWebsite');
 const bucketPutReplication = require('./bucketPutReplication');
 const bucketPutLifecycle = require('./bucketPutLifecycle');
 const bucketPutPolicy = require('./bucketPutPolicy');
+const bucketPutObjectLock = require('./bucketPutObjectLock');
 const bucketGetReplication = require('./bucketGetReplication');
 const bucketDeleteReplication = require('./bucketDeleteReplication');
 const corsPreflight = require('./corsPreflight');
@@ -187,6 +188,7 @@ const api = {
     bucketPutPolicy,
     bucketGetPolicy,
     bucketDeletePolicy,
+    bucketPutObjectLock,
     corsPreflight,
     completeMultipartUpload,
     initiateMultipartUpload,
--- a/lib/api/bucketPutObjectLock.js
+++ b/lib/api/bucketPutObjectLock.js
@@ -1,0 +1,82 @@
+const { waterfall } = require('async');
+const arsenal = require('arsenal');
+
+const errors = arsenal.errors;
+const ObjectLockConfiguration = arsenal.models.ObjectLockConfiguration;
+
+const parseXML = require('../utilities/parseXML');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const metadata = require('../metadata/wrapper');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const { pushMetric } = require('../utapi/utilities');
+
+/**
+ * Bucket Put Object Lock - Create or update bucket object lock configuration
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+
+function bucketPutObjectLock(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketPutObjectLock' });
+
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketPutObjectLock',
+    };
+    return waterfall([
+        next => parseXML(request.post, log, next),
+        (parsedXml, next) => {
+            const lockConfigClass = new ObjectLockConfiguration(parsedXml);
+            // if there was an error getting object lock configuration,
+            // returned configObj will contain 'error' key
+            process.nextTick(() => {
+                const configObj = lockConfigClass.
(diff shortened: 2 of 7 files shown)


lib/api/api.js
```diff
@@ -22,6 +22,7 @@ const bucketPutWebsite = require('./bucketPutWebsite');
 const bucketPutReplication = require('./bucketPutReplication');
 const bucketPutLifecycle = require('./bucketPutLifecycle');
 const bucketPutPolicy = require('./bucketPutPolicy');
+const bucketPutObjectLock = require('./bucketPutObjectLock');
```
ilkescality, line 25: nit: could be moved above `bucketPutPolicy` to keep the order

ilkescality, line 25: nit: could be moved above `bucketPutPolicy` to keep the alphabetical order

package.json
```diff
@@ -19,9 +19,9 @@
   },
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
-    "arsenal": "github:scality/Arsenal#2b9ac57",
+    "arsenal": "github:scality/Arsenal#c5e21e0",
     "async": "~2.5.0",
-    "aws-sdk": "2.178.0",
+    "aws-sdk": "^2.178.0",
```
ilkescality: Noticed the `^` - I believe it accidentally came from my PR, and we should be good as we're gonna merge the aws-sdk version PR first.

tests/unit/api/bucketPutObjectLock.js
```diff
@@ -0,0 +1,84 @@
...
+
+const expectedObjectLockConfig = {
+    rule: {
+        mode: 'GOVERNANCE',
+        days: 1,
+    },
+};
+
+describe('putBucketObjectLock API', () => {
+    before(() => cleanup());
```
ilkescality, line 43: This could be just `before(cleanup)`

tests/unit/api/bucketPutObjectLock.js
```diff
@@ -0,0 +1,84 @@
...
+
+    describe('without Object Lock enabled on bucket', () => {
+        beforeEach(done => bucketPut(authInfo, bucketPutRequest, log, done));
+        afterEach(() => cleanup());
```
ilkescality, line 47: same here also; could be just `afterEach(cleanup)`

tests/unit/api/bucketPutObjectLock.js
```diff
@@ -0,0 +1,84 @@
...
+            });
+        });
+    });
+
+    describe('with Object Lock enabled on bucket', () => {
+        const bucketObjLockRequest = Object.assign({}, bucketPutRequest,
+            { headers: { 'x-amz-bucket-object-lock-enabled': true } });
+
+        beforeEach(done => bucketPut(authInfo, bucketObjLockRequest, log, done));
+        afterEach(() => cleanup());
```
ilkescality, line 62: similarly `afterEach(cleanup)`

ilkescality (approved):

naren-rajendran (approved):

lib/api/bucketPutObjectLock.js
```diff
@@ -0,0 +1,82 @@
...
+    return waterfall([
+        next => parseXML(request.post, log, next),
+        (parsedXml, next) => {
+            const lockConfigClass = new ObjectLockConfiguration(parsedXml);
+            // if there was an error getting object lock configuration,
+            // returned configObj will contain 'error' key
+            process.nextTick(() => {
+                const configObj = lockConfigClass.
+                    getValidatedObjectLockConfiguration();
+                return next(configObj.error || null, configObj);
```
jonathan-gramain, line 40: If there is an error, you're supposed to return a `bucket` object as argument to the callback, not `configObj`, correct? In which case, maybe you could execute `metadataValidateBucket` first so you have the `bucket` object ready for error cases.

tests/unit/api/bucketPutObjectLock.js
```diff
@@ -0,0 +1,84 @@
...
+
+        it('should update a bucket\'s metadata with object lock config', done => {
+            bucketPutObjectLock(authInfo, putObjLockRequest, log, err => {
+                if (err) {
```
jonathan-gramain: Since this is a test, you could simply do `assert.ifError(err)`, mocha should log the error properly if it fails.

dora-korpar:
@bert-e wait

lib/api/bucketPutObjectLock.js
dora-korpar: Hm that's true. I put that part before the `metadataValidateBucket` because AWS returns MalformedXML and InvalidRequest errors before any permissions errors that would be returned by the validation function, but I do need the bucket object.

dora-korpar: We use this same order of operations for all of the other put config APIs tho, like putBucketReplication and putBucketLifecycle too

dora-korpar: @jonathan-gramain what do you think would be the best way to handle it?

jonathan-gramain: What do the other operations do, they validate the config first, and do not return a `bucket` object in case of error? If so, this is fine to do the same. In doubt I would be consistent with the other operations to avoid surprises.

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e force_reset

PR #2542, opened 2020-05-19
bugfix: S3C-2959 fix objectLockEnabled and NFSEnabled locations

ilkescality:
Fixes the locations of the `NFSEnabled` and `objectLockEnabled` params. 

I realized `NFSEnabled` is off by one while fixing the location of `objectLockEnabled`; now both should be correct according to model: https://github.com/scality/Arsenal/blob/development/8.1/lib/models/BucketInfo.js#L66

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -174,8 +174,8 @@ function createBucket(authInfo, bucketName, headers,
     const objectLockEnabled = headers['x-amz-bucket-object-lock-enabled'];
     const bucket = new BucketInfo(bucketName, canonicalID, ownerDisplayName,
         creationDate, BucketInfo.currentModelVersion(), null, null, null, null,
-        null, null, null, null, null, null, null, null, isNFSEnabled,
-        objectLockEnabled);
+        null, null, null, null, null, null, null, null, null, isNFSEnabled,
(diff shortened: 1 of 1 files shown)


jonathan-gramain (approved):
Having that many arguments to a function is calling for trouble... not blaming this PR, just a general statement. The params should be passed with an object instead, but it would likely be a significant refactor though considering the number of users of this class but might be worth it in the long run.

jonathan-gramain:
One way to transition this smoothly could be to have a `create` class function (aka. static) taking an object param and constructing and returning a BucketInfo, so the compatibility is retained but we can start using the new less error-prone constructing method.

Not asking to do this now, just a hint.

dora-korpar (approved):

ilkescality:
> Having that many arguments to a function is calling for trouble... not blaming this PR, just a general statement. The params should be passed with an object instead, but it would likely be a significant refactor though considering the number of users of this class but might be worth it in the long run.

Yeah this is bad, and very error-prone indeed. That's actually one of the biggest reasons why I love Typescript :) What you suggested sounds good, I will create a ticket to make sure it wouldn't get forgotten as it'd be really beneficial for us to refactor this (in not so distant future hopefully).

I might volunteer to take care of it sometime after this Object Lock feature.

ilkescality:
@bert-e approve

PR #2559, opened 2020-05-29
feature: S3C-2944 put object legal hold api

ilkescality:
Includes PUT Object legal hold API and unit tests.

### DIFF
--- a/lib/api/objectPutLegalHold.js
+++ b/lib/api/objectPutLegalHold.js
@@ -1,0 +1,116 @@
+const async = require('async');
+const { errors, s3middleware } = require('arsenal');
+
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const { decodeVersionId, getVersionIdResHeader } =
+  require('./apiUtils/object/versioning');
+const getReplicationInfo = require('./apiUtils/object/getReplicationInfo');
+const metadata = require('../metadata/wrapper');
+const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
+const { pushMetric } = require('../utapi/utilities');
+
+const { parseLegalHoldXml } = s3middleware.objectLegalHold;
+
+const REPLICATION_ACTION = 'PUT_LEGAL_HOLD';
+
+/**
+ * Object Put Legal Hold - Sets legal hold status of object
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+function objectPutLegalHold(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'objectPutLegalHold' });
+
+    const { bucketName, objectKey } = request;
+
+    const decodedVidResult = decodeVersionId(request.query);
+    if (decodedVidResult instanceof Error) {
+        log.trace('invalid versionId query', {
+            versionId: request.query.versionId,
+            error: decodedVidResult,
+        });
+        return process.nextTick(() => callback(decodedVidResult));
+    }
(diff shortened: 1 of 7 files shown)


lib/api/objectPutLegalHold.js
```diff
@@ -0,0 +1,124 @@
...
+            // eslint-disable-next-line no-param-reassign
+            objectMD.legalHold = legalHold;
+            const params = objectMD.versionId ?
+                { versionId: objectMD.versionId } : {};
+            const replicationInfo = getReplicationInfo(objectKey, bucket, true,
+                0, REPLICATION_ACTION, objectMD);
+            if (replicationInfo) {
+                // eslint-disable-next-line no-param-reassign
+                objectMD.replicationInfo = Object.assign({},
+                    objectMD.replicationInfo, replicationInfo);
```
ilkescality, line 91: @dora-korpar In here I used similar logic with your PUT object retention PR to set the replication info but couldn't be sure if we'd necessarily need it for this one as well. What do you think?

tests/unit/api/objectPutLegalHold.js
```diff
@@ -0,0 +1,102 @@
...
+
+const putObjectRequest = new DummyRequest({
+    bucketName,
+    namespace,
+    objectKey: objectName,
+    headers: {},
+    url: `/${bucketName}/${objectName}`,
+}, postBody);
+
+const objectLegalHoldXml = status => '<LegalHold ' +
```
miniscruff, line 32: Can combine all these lines into one backtick string

miniscruff (approved):

tests/unit/api/objectPutLegalHold.js
ilkescality: I made this change after your comment but just reverted back after finding out whitespace errors with XMLs due to backticks adding line breaks and extra whitespace (an alternative could be writing it on one line but I thought that would probably be less ideal).

objectLegalHold_js_—_cloudserver

miniscruff: Backticks will take the string verbatim including newlines and spaces. It is valid XML but your test does not include the newlines or spaces so is causing a not equal assertion. This is one of many reasons I think string to string assertions are cumbersome but meh. Building an XML file or object from actual strings is generally not ideal IMO. It is like you generally don't build JSON from strings and instead use an object and stringify it. In doing so you avoid string specific errors, especially around ordering and whitespace. Regardless, valid XML is valid XML and I have no concern with your string here. I would recommend finding a way to generate XML from a more static object such as an object or JSON like structure for sanity purposes.

lib/api/objectPutLegalHold.js
dora-korpar: yeah, I borrowed that logic from the `putObjectTagging` api, but actually I don't see any reason to include it. I think we should both remove it from our apis

dora-korpar (approved):

tests/unit/api/objectPutLegalHold.js
ilkescality: Yeah I thought of adjusting the test instead as an alternative so they would match but couldn't be sure if line breaks or whitespace would cause any unforeseen issues.

I cannot agree more about building XML from actual strings, like 110% :)
I think writing API utils to generate XML from JSON (and vice-versa) is a good idea. It could be really beneficial to have a standardized method for it to be utilized across the other APIs too. Just having a method to generate the XML from the object or JSON would be good enough for a start I think. Then we can start refactoring the existing APIs gradually as well. I'll create a ticket for it so it wouldn't get missed out or forgotten.

miniscruff: Might want to search NPM for an existing package, I can't imagine it doesn't exist already.

ilkescality:
@bert-e approve

PR #2573, opened 2020-06-01
feature: S3C-2945 get object legal hold api

ilkescality:
Includes GET Object legal hold API and unit tests.

### DIFF
--- a/lib/api/apiUtils/authorization/prepareRequestContexts.js
+++ b/lib/api/apiUtils/authorization/prepareRequestContexts.js
@@ -12,6 +12,7 @@ const apiMethodWithVersion = {
     objectPutTagging: true,
     objectGetTagging: true,
     objectDeleteTagging: true,
+    objectGetLegalHold: true,
     objectPutLegalHold: true,
 };
 
--- a/lib/api/objectGetLegalHold.js
+++ b/lib/api/objectGetLegalHold.js
@@ -1,0 +1,104 @@
+const async = require('async');
+const { errors, s3middleware } = require('arsenal');
+
+const { decodeVersionId, getVersionIdResHeader }
+    = require('./apiUtils/object/versioning');
+
+const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
+const { pushMetric } = require('../utapi/utilities');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+
+const { convertToXml } = s3middleware.objectLegalHold;
+
+/**
+ * Returns legal hold status of object
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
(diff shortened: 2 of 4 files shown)


lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -4,9 +4,17 @@ const { config } = require('../../../Config');
 const RequestContext = policies.RequestContext;
 const requestUtils = policies.requestUtils;
 let apiMethodAfterVersionCheck;
-const apiMethodWithVersion = { objectGetACL: true, objectPutACL: true,
-    objectGet: true, objectDelete: true, objectPutTagging: true,
-    objectGetTagging: true, objectDeleteTagging: true };
+const apiMethodWithVersion = {
```
miniscruff: I feel like I am seeing this diff for the third time @dora-korpar had a PR with it before cause I commented on using a Set. Is this in like 5 places or what?

lib/api/objectGetLegalHold.js
```diff
@@ -0,0 +1,104 @@
...
+
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        objectKey,
+        requestType: 'objectGetLegalHold',
+        versionId,
+    };
+
+    return async.waterfall([
```
miniscruff, line 44: no need to return here, async.waterfall will return undefined every time anyway ( from the docs anyway )

lib/api/objectPutLegalHold.js
```diff
@@ -0,0 +1,116 @@
...
+
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        objectKey,
+        requestType: 'objectPutLegalHold',
+        versionId,
+    };
+
+    return async.waterfall([
```
miniscruff: Same as above, async.waterfall returns undefined so just not returning has the same affect.

miniscruff (approved):
Seems good to me, from the JS side anyway.

naren-rajendran (approved):

lib/api/apiUtils/authorization/prepareRequestContexts.js
ilkescality: It's the same place but we both needed to add there for our PRs, once they get merged, it's gonna be only in one place.

lib/api/objectPutLegalHold.js
ilkescality: You're right about this and `return` was redundant, so initially I made the change but just reverted back due to linter warning failing CI for `consistent-return` "Expected to return a value at the end of this function". (It's interesting as there're bunch of functions with `async.waterfall` without `return` though)

miniscruff: We might want to take a look at the linter settings in more detail, more than a few of my comments have been struck down to "linter caused an error" for, of course IMO, no reason. This will only increase as we move to more recent javascript standards like async/await and generators.

ilkescality:
@bert-e reset

dora-korpar (approved):

ilkescality:
@bert-e approve

PR #2590, opened 2020-06-03
feature: S3C-2946 get and head object support object lock

ilkescality:
Extends GET and HEAD Object APIs to return object lock related information for the object if any set on the object's metadata.

When an object version have object lock related information stored in its metadata, GET Object and HEAD Object APIs should return these in the response; sample GET/HEAD Object response for a locked object:
```
{   
    "AcceptRanges": "bytes",
    "LastModified": "2020-06-03T03:26:39+00:00",
    "ContentLength": 0,
    "VersionId": "39383430383834353230303133383939393939395247303031202030",
    "Metadata": {},
    "ObjectLockMode": "GOVERNANCE",
    "ObjectLockRetainUntilDate": "2021-12-12T00:00:34+00:00",
    "ObjectLockLegalHoldStatus": "ON"
}
```

### DIFF
--- a/lib/utilities/collectResponseHeaders.js
+++ b/lib/utilities/collectResponseHeaders.js
@@ -75,6 +75,19 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
         responseMetaHeaders['x-amz-tagging-count'] =
           Object.keys(objectMD.tags).length;
     }
+    const hasRetentionInfo = objectMD.retentionInfo
+        && objectMD.retentionInfo.retainUntilDate
+        && objectMD.retentionInfo.mode;
+    if (hasRetentionInfo) {
+        responseMetaHeaders['x-amz-object-lock-retain-until-date']
+            = objectMD.retentionInfo.retainUntilDate;
+        responseMetaHeaders['x-amz-object-lock-mode']
+            = objectMD.retentionInfo.mode;
+    }
+    if (objectMD.legalHold !== undefined) {
+        responseMetaHeaders['x-amz-object-lock-legal-hold']
+            = objectMD.legalHold ? 'ON' : 'OFF';
+    }
     if (objectMD.replicationInfo && objectMD.replicationInfo.status) {
         responseMetaHeaders['x-amz-replication-status'] =
             objectMD.replicationInfo.status;
--- a/tests/unit/api/objectGet.js
+++ b/tests/unit/api/objectGet.js
@@ -75,6 +75,136 @@ describe('objectGet API', () => {
         });
     });
 
+    const testPutBucketRequestObjectLock = {
+        bucketName,
+        namespace,
+        headers: {
+            'host': `${bucketName}.s3.amazonaws.com`,
+            'x-amz-bucket-object-lock-enabled': true,
+        },
+        url: `/${bucketName}`,
+    };
+
+    const createPutDummyRetention = (date, mode) => new DummyRequest({
+        bucketName,
+        namespace,
+        objectKey: objectName,
+        headers: {
+            'x-amz-object-lock-retain-until-date': date,
+            'x-amz-object-lock-mode': mode,
+            'content-length': '12',
+        },
+        parsedContentLength: 12,
+        url: `/${bucketName}/${objectName}`,
+    }, postBody);
+
+    const testDate = new Date(2022, 6, 3).toISOString();
+
+    it('should get the object metadata with valid retention info', done => {
+        bucketPut(authInfo, testPutBucketRequestObjectLock, log, () => {
+            const request = createPutDummyRetention(testDate, 'GOVERNANCE');
+            objectPut(authInfo, request, undefined,
+                log, (err, headers) => {
+                    assert.ifError(err);
+                    assert.strictEqual(headers.ETag, `"${correctMD5}"`);
+                    const req = testGetRequest;
+                    objectGet(authInfo, req, false, log, (err, r, headers) => {
+                        assert.ifError(err);
+                        assert.strictEqual(
+                            headers['x-amz-object-lock-retain-until-date'],
+                            testDate);
+                        assert.strictEqual(
+                            headers['x-amz-object-lock-mode'],
+                            'GOVERNANCE');
+                        assert.strictEqual(headers.ETag,
+                            `"${correctMD5}"`);
+                        done();
+                    });
+                });
+        });
+    });
+
+    const createPutDummyLegalHold = legalHold => new DummyRequest({
+        bucketName,
+        namespace,
+        objectKey: objectName,
+        headers: {
+            'x-amz-object-lock-legal-hold': legalHold,
+            'content-length': '12',
+        },
+        parsedContentLength: 12,
+        url: `/${bucketName}/${objectName}`,
+    }, postBody);
+
+    const testStatuses = ['ON', 'OFF'];
+    testStatuses.forEach(status => {
+        it(`should get object metadata with legal hold ${status}`, done => {
+            bucketPut(authInfo, testPutBucketRequestObjectLock, log, () => {
+                const request = createPutDummyLegalHold(status);
+                objectPut(authInfo, request, undefined, log,
+                    (err, resHeaders) => {
+                        assert.ifError(err);
+                        assert.strictEqual(resHeaders.ETag, `"${correctMD5}"`);
+                        objectGet(authInfo, testGetRequest, false, log,
+                            (err, res, headers) => {
+                                assert.ifError(err);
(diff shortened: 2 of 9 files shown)


lib/utilities/collectResponseHeaders.js
```diff
@@ -75,6 +75,17 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
         responseMetaHeaders['x-amz-tagging-count'] =
           Object.keys(objectMD.tags).length;
     }
+    if (objectMD.retentionInfo && (objectMD.retentionInfo.retainUntilDate
```
miniscruff: the parenthesis around conditions is not required here as left to right ANDS are the same.

lib/utilities/collectResponseHeaders.js
```diff
@@ -75,6 +75,17 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
...
+        && objectMD.retentionInfo.mode)) {
+            responseMetaHeaders['x-amz-object-lock-retain-until-date']
```
miniscruff: Having these on two lines is hard to read, I am sure eslint will complain at you if you put it on one line but like, is there something we can do here?

tests/unit/api/objectGet.js
```diff
@@ -75,6 +75,155 @@ describe('objectGet API', () => {
...
+    it('should get the object metadata with valid retention info', done => {
+        bucketPut(authInfo, testPutBucketRequestObjectLock, log, () => {
+            const request = testPutObjectReqRetention(testDate, 'GOVERNANCE');
+            objectPut(authInfo, request, undefined,
+                log, (err, headers) => {
+                    assert.ifError(err);
+                    assert.strictEqual(headers.ETag, `"${correctMD5}"`);
+                    objectGet(authInfo, testGetRequest, false, log,
+                        (err, res, headers) => {
+                            assert.ifError(err);
```
miniscruff: can anything be done about the amount of whitespace on these lines? The best I could tell was merging lines 106/107 and 110/111 but even then its quite a lot of empty space. Makes it a little hard to read.

tests/unit/api/objectGet.js
```diff
@@ -75,6 +75,155 @@ describe('objectGet API', () => {
...
+        bucketName,
+        namespace,
+        headers: {
+            host: `${bucketName}.s3.amazonaws.com`,
+            'x-amz-object-lock-enabled': true,
+        },
+        url: `/${bucketName}`,
+    };
+
+    const testPutObjectReqRetention = (date, mode) => new DummyRequest({
```
miniscruff: Maybe a name that dictates it just creates a dummy request like, `createPutObjectDummyRequest` or `createPutDummy` or `createPutRequest`. I thought this was an actual test the first time I read it.

tests/unit/api/objectGet.js
```diff
@@ -75,6 +75,155 @@ describe('objectGet API', () => {
...
+                                'GOVERNANCE');
+                            assert.strictEqual(headers.ETag,
+                                `"${correctMD5}"`);
+                            done();
+                        });
+                });
+        });
+    });
+
+    const testPutObjectReqLegalHold = legalHold => new DummyRequest({
```
miniscruff: Same as above.

miniscruff (approved):
Looks fine but small comments might improve it "might".

lib/utilities/collectResponseHeaders.js
ilkescality: At first I did without those, then I confused myself :) gonna remove them back.

lib/utilities/collectResponseHeaders.js
ilkescality: Yeah I agree about this, eslint will definitely complain but I'll see if there's anything that can be done to make this look better - either bypassing lint or I will think of something else.

tests/unit/api/objectGet.js
ilkescality: I'll work on something about this, bothered me while writing it as well tbh.

lib/utilities/collectResponseHeaders.js
rahulreddy: What I did when I ran into this was to move the conditional to a variable, for example,
```js
const objectLocked = objectMD.retentionInfo && objectMD.retentionInfo.retainUntilDate
        && objectMD.retentionInfo.mode;
if (objectLocked) {
  // do the logic
}
```

miniscruff: I am a huge fan that bools should always represent the "positive" form. It overall simplifies the logic and terms. The positive form is not always the default but generally is. For example, even though all this logic has to do with the objects being locked. The bool that dictates if the object is in fact locked might want to be false if it is. Something like `objectOpen` or `objectEditable` ( kinda guessing since I do not know any more terms here that might be more standard ). Note that "unlocked" is not a good name as its effectively a double negative. 

Consider a variable that dictates if a user can go through a portal ( some game idk ).
```js
if (!portalClosed) {
  // go in
}
// OR
if (portalOpen) {
  // go in
}
```

The first says we can go in the portal if it is not closed. Or the second is just it is open. The positive form here is using the portal for portal things.

This is a little difficult at times for conditions that are sort of "blocker" conditions. Just as in this case the purpose of a lock is to block something from happening. If a door had a lock and a deadbolt we might consider the lock "engaged" to be the lock would block the door from opening. So the door could be opened if all locks are not engaged. However, since the primary function of an actual lock is to prevent things from happening its positive value is true if the lock is engaged.

```javascript
const canOpenDoor = !deadbolt.engaged && !lock.engaged;
if (canOpenDoor) {
  // do good things
}
// or
const doorLocked = deadbolt.engaged || lock.engaged;
if (!doorLocked) {
  // double negative
}
```

Also prefixing booleans with is, can, has or similar word is also very powerful as it makes conditions turn almost into sentences. Compare "if I can open the door then open the door" to "if not door locked open door". 

tl;dr use positive forms for booleans, rename locked to something like open or writable or something in the positive form. 

This might be a lot of changes since so much work was put in to add "locking" but just like mentioning this sort of stuff. I may also have some terms wrong and the lock here is actually the positive case for something in which case carry on...

anurag4DSB (approved):
I like the intensional good namings of variables and methods :)

lib/utilities/collectResponseHeaders.js
anurag4DSB: +1

rahulreddy: I agree on the usage of booleans to be in positive form, it avoids double negations and what not. I was just trying to illustrate that you can extract the evaluation logic into a var and use that for conditionals. I will let Ilke chose the naming, I used `lock` as we are working with object locking mechanism.

miniscruff: oh yea my response wasn't an alternative just an additional comment on booleans in general.

tests/unit/api/objectGet.js
miniscruff: @ilkescality this was not addressed?

lib/utilities/collectResponseHeaders.js
ilkescality: @miniscruff @rahulreddy @anurag4DSB thanks for the reviews
@miniscruff thanks for the thorough review; great suggestions. I realized I forgot to reply here after making the changes yesterday. I made the change to make positive use of booleans and named as `hasRetentionInfo` (I thought of doing it like this initially but then ended up keeping it similar with other if statements in the file, definitely looks a lot more readable now though).

In terms of renaming `locked` to something like open/editable / writable / something in the positive form; I think it’s a good practice in general too but for the object lock context, as we inevitably use the term `lock / locked` in bunch of places, (such as error messages, headers, API method names etc.) I’m guessing introducing another term to be a synonym for ‘lock’ might be confusing overall (in terms of the object lock feature at a holistic level) so I think it might probably worth to keep it.

tests/unit/api/objectGet.js
ilkescality: Oh yeah seems like I missed that one, will change this too.

lib/utilities/collectResponseHeaders.js
miniscruff: That's fair, I couldn't really come up with anything better than locked in my examples. Just know that at some points when working with locking you may need to swap the bool to make sense. But that is fine, its more a general idea and not a rule.

dora-korpar (approved):

ilkescality:
@bert-e approve

ilkescality:
@bert-e reset

ilkescality:
@bert-e approve

PR #2613, opened 2020-06-08
feature: S3C-2791 extend put object for object lock

ilkescality:
Extends PUT Object API to put object with object lock retention configuration and/or legal hold.
- If PUT Object request has legal hold header, stores legal hold status in object's metadata.
- If PUT Object request has retention headers, stores`retentionInfo` in object's metadata.
- If PUT Object request does not have retention headers but the bucket has object lock configuration, calculates the `retainUntilDate` and stores the retentionInfo with `retainUntilDate` in object's metadata.

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -148,6 +148,12 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             removeAWSChunked(request.headers['content-encoding']);
         metadataStoreParams.expires = request.headers.expires;
         metadataStoreParams.tagging = request.headers['x-amz-tagging'];
+        const defaultObjectLockConfiguration
+            = bucketMD.getObjectLockConfiguration();
+        if (defaultObjectLockConfiguration) {
+            metadataStoreParams.defaultRetention
+                = defaultObjectLockConfiguration;
+        }
     }
 
     // if creating new delete marker and there is an existing object, copy
--- a/lib/api/apiUtils/object/objectLockHelpers.js
+++ b/lib/api/apiUtils/object/objectLockHelpers.js
@@ -1,0 +1,97 @@
+const { errors } = require('arsenal');
+const moment = require('moment');
+/**
+ * Calculates retain until date for the locked object version
+ * @param {object} retention - includes days or years retention period
+ * @return {object} the date until the object version remains locked
+ */
+function calculateRetainUntilDate(retention) {
+    const { days, years } = retention;
+    const date = moment();
+    // Calculate the number of days to retain the lock on the object
+    const retainUntilDays = days || years * 365;
+    const retainUntilDate
+        = date.add(retainUntilDays, 'Days');
+    return retainUntilDate.toISOString();
+}
+/**
+ * Validates object lock headers
+ * @param {object} bucket - bucket metadata
+ * @param {object} headers - request headers
+ * @param {object} log - the log request
+ * @return {object} - object with error if validation fails
+ */
+function validateHeaders(bucket, headers, log) {
+    const bucketObjectLockEnabled = bucket.isObjectLockEnabled();
+    const objectLegalHold = headers['x-amz-object-lock-legal-hold'];
+    const objectLockDate = headers['x-amz-object-lock-retain-until-date'];
+    const objectLockMode = headers['x-amz-object-lock-mode'];
+    // If retention headers or legal hold header present but
+    // object lock is not enabled on the bucket return error
+    if ((objectLockDate || objectLockMode || objectLegalHold)
+        && !bucketObjectLockEnabled) {
+        log.trace('bucket is missing ObjectLockConfiguration');
+        return errors.InvalidRequest.customizeDescription(
+            'Bucket is missing ObjectLockConfiguration');
+    }
+    if ((objectLockMode || objectLockDate) &&
+        !(objectLockMode && objectLockDate)) {
+        return errors.InvalidArgument.customizeDescription(
+            'x-amz-object-lock-retain-until-date and ' +
+            'x-amz-object-lock-mode must both be supplied'
+        );
+    }
+    const validModes = new Set(['GOVERNANCE', 'COMPLIANCE']);
+    if (objectLockMode && !validModes.has(objectLockMode)) {
+        return errors.InvalidArgument.customizeDescription(
+            'Unknown wormMode directive');
+    }
+    const validLegalHolds = new Set(['ON', 'OFF']);
+    if (objectLegalHold && !validLegalHolds.has(objectLegalHold)) {
+        return errors.InvalidArgument.customizeDescription(
+            'Legal hold status must be one of "ON", "OFF"');
+    }
+    const currentDate = new Date().toISOString();
+    if (objectLockMode && objectLockDate <= currentDate) {
+        return errors.InvalidArgument.customizeDescription(
+            'The retain until date must be in the future!');
+    }
+    return null;
+}
+
+/**
+ * Sets object retention ond/or legal hold information on object's metadata
+ * @param {object} headers - request headers
+ * @param {object} md - object metadata
+ * @param {(object|null)} defaultRetention - bucket retention configuration if
+ * bucket has any configuration set
+ * @return {undefined}
+ */
+function setObjectLockInformation(headers, md, defaultRetention) {
+    // Stores retention information if object either has its own retention
+    // configuration or default retention configuration from its bucket
+    const headerMode = headers['x-amz-object-lock-mode'];
+    const headerDate = headers['x-amz-object-lock-retain-until-date'];
+    const objectRetention = headers && headerMode && headerDate;
+    if (objectRetention || defaultRetention) {
+        const mode = headerMode || defaultRetention.rule.mode;
+        const date = headerDate
+            || calculateRetainUntilDate(defaultRetention.rule);
+        const retention = {
+            mode,
+            retainUntilDate: date,
+        };
+        md.setRetentionInfo(retention);
+    }
+    const headerLegalHold = headers['x-amz-object-lock-legal-hold'];
+    if (headers && headerLegalHold) {
+        const legalHold = headerLegalHold === 'ON';
+        md.setLegalHold(legalHold);
+    }
+}
+
+module.exports = {
+    calculateRetainUntilDate,
+    setObjectLockInformation,
+    validateHeaders,
+};
--- a/lib/services.js
+++ b/lib/services.js
@@ -10,6 +10,8 @@ const constants = require('../constants');
 const data = require('./data/wrapper');
 const metadata = require('./metadata/wrapper');
 const logger = require('./utilities/logger');
+const { setObjectLockInformation }
+    = require('./api/apiUtils/object/objectLockHelpers');
 const removeAWSChunked = require('./api/apiUtils/object/removeAWSChunked');
 const { parseTagFromQuery } = s3middleware.tagging;
 const { config } = require('./Config');
@@ -95,7 +97,7 @@ const services = {
             contentType, cacheControl, contentDisposition, contentEncoding,
             expires, multipart, headers, overrideMetadata, log,
             lastModifiedDate, versioning, versionId, tagging, taggingCopy,
-            replicationInfo, dataStoreName } = params;
+            replicationInfo, defaultRetention, dataStoreName } = params;
         log.trace('storing object in metadata');
         assert.strictEqual(typeof bucketName, 'string');
         const md = new ObjectMD();
@@ -127,6 +129,18 @@ const services = {
         }
         if (headers && headers['x-amz-website-redirect-location']) {
             md.setRedirectLocation(headers['x-amz-website-redirect-location']);
+        }
+        if (headers) {
+            // Stores retention information if object either has its own retention
+            // configuration or default retention configuration from its bucket
+            const headerMode = headers['x-amz-object-lock-mode'];
+            const headerDate = headers['x-amz-object-lock-retain-until-date'];
+            const headerLegalHold = headers['x-amz-object-lock-legal-hold'];
+            const objectRetention = headers && headerMode && headerDate;
+            const objectLegalHold = headers && headerLegalHold;
+            if (objectRetention || defaultRetention || objectLegalHold) {
+                setObjectLockInformation(headers, md, defaultRetention);
+            }
         }
         if (replicationInfo) {
             md.setReplicationInfo(replicationInfo);
--- a/tests/unit/api/apiUtils/objectLockHelpers.js
+++ b/tests/unit/api/apiUtils/objectLockHelpers.js
@@ -1,0 +1,178 @@
+const assert = require('assert');
+const moment = require('moment');
+const { errors } = require('arsenal');
+const BucketInfo = require('arsenal').models.BucketInfo;
+const { DummyRequestLogger } = require('../../helpers');
+const {
+    calculateRetainUntilDate,
+    validateHeaders,
+} = require('../../../../lib/api/apiUtils/object/objectLockHelpers');
+
+const mockName = 'testbucket';
+const mockOwner = 'someCanonicalId';
+const mockOwnerDisplayName = 'accountDisplayName';
+const mockCreationDate = new Date().toJSON();
+
+const bucketInfo = new BucketInfo(
+    mockName, mockOwner, mockOwnerDisplayName, mockCreationDate,
+    null, null, null, null, null, null, null, null, null, null,
+    null, null, true);
+
+const objLockDisabledBucketInfo = new BucketInfo(
+    mockName, mockOwner, mockOwnerDisplayName, mockCreationDate,
+    null, null, null, null, null, null, null, null, null, null,
+    null, null, false);
+
+const log = new DummyRequestLogger();
+
+describe('objectLockHelpers: validateHeaders', () => {
+    it('should fail if object lock is not enabled on the bucket', () => {
+        const headers = {
+            'x-amz-object-lock-retain-until-date': '2050-10-12',
+            'x-amz-object-lock-mode': 'COMPLIANCE',
+        };
+        const objectLockValidationError
+            = validateHeaders(objLockDisabledBucketInfo, headers, log);
+        const expectedError = errors.InvalidRequest.customizeDescription(
+            'Bucket is missing ObjectLockConfiguration');
+        assert.strictEqual(objectLockValidationError.InvalidRequest, true);
+        assert.strictEqual(objectLockValidationError.description,
+            expectedError.description);
+    });
+
+    it('should pass with valid retention headers', () => {
+        const headers = {
+            'x-amz-object-lock-retain-until-date': '2050-10-12',
+            'x-amz-object-lock-mode': 'COMPLIANCE',
+        };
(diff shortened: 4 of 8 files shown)


lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {object} retention - includes days or years retention period
+ * @return {object} the date until the object version remains locked
+ */
+function calculateRetainUntilDate(retention) {
+    const { days, years } = retention;
+    const date = new Date();
+    // Calculate the number of days to retain the lock on the object
+    const retainUntilDays = days || years * 365;
+    const retainUntilDate
+        = date.setDate(date.getDate() + parseInt(retainUntilDays, 10));
```
miniscruff: do you need the parseInt here?

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -0,0 +1,64 @@
...
+            'x-amz-object-lock-retain-until-date and ' +
+            'x-amz-object-lock-mode must both be supplied'
+        );
+    }
+    const validModes = new Set(['GOVERNANCE', 'COMPLIANCE']);
+    const validLegalHolds = new Set(['ON', 'OFF']);
+    if ((objectLockMode && !validModes.has(objectLockMode))
+        || (objectLegalHold && !validLegalHolds.has(objectLegalHold))) {
+        validation.error = errors.InvalidArgument.customizeDescription(
+            'Unknown wormMode directive');
```
miniscruff, line 47: Is this the right error message? wormMode?

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -148,6 +148,10 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             removeAWSChunked(request.headers['content-encoding']);
         metadataStoreParams.expires = request.headers.expires;
         metadataStoreParams.tagging = request.headers['x-amz-tagging'];
+        if (bucketMD.getObjectLockConfiguration()) {
```
miniscruff: This method sounds like it returns some sort of object or structure but is being used as a conditional. I just want to make sure this is in fact what you want to do and it isn't some sort of typo.

tests/unit/api/objectPut.js
```diff
@@ -269,6 +429,30 @@ describe('objectPut API', () => {
...
+                'x-amz-object-lock-retain-until-date': mockDate,
+                'x-amz-object-lock-mode': 'GOVERNANCE',
+            },
+            url: `/${bucketName}/${objectName}`,
+            calculatedHash: 'vnR+tLdVF79rPPfF+7YvOg==',
+        }, postBody);
+
+        bucketPut(authInfo, testPutBucketRequest, log, () => {
+            objectPut(authInfo, testPutObjectRequest, undefined, log, err => {
+                assert.deepStrictEqual(err, errors.InvalidRequest
```
miniscruff, line 450: String comparing the error description is probably a bit much. Altering descriptions like this should not typically break tests.

lib/api/apiUtils/object/objectLockHelpers.js
dora-korpar: it actually is. It's a weird one for sure

miniscruff: ok, just wanted to make sure...

lib/services.js
```diff
@@ -128,6 +130,26 @@ const services = {
         if (headers && headers['x-amz-website-redirect-location']) {
             md.setRedirectLocation(headers['x-amz-website-redirect-location']);
         }
+        // Stores retention information if object either has its own retention
+        // configuration or default retention configuration from its bucket
+        const headerMode = headers['x-amz-object-lock-mode'];
```
dora-korpar: could you pull this logic (evaluating headers vs defaultRetention) into another method in `objectLockHelpers`? then we'll be able to use it for the intitiateMPU update :)

lib/api/apiUtils/object/objectLockHelpers.js
dora-korpar: @ilkescality idk if this is true for put object, but when i was testing for intitiateMPU, I found that AWS doesn't actually return an error if the legalHold value isn't `ON` or `OFF`. I think it's better that we return an error so people expecting their objects to have legal hold know their setting didn't work, but I think it should be a different error than the mode one.  'Unknown legal hold value' or something like that

lib/services.js
ilkescality: sure thing :)

lib/api/apiUtils/object/objectLockHelpers.js
ilkescality: I don't really - it can never be float so parseInt is kinda excessive; gonna remove it.

lib/api/apiUtils/object/createAndStoreObject.js
ilkescality: Yup, it's intentional although I can see why it might not look quite right, I'm gonna tweak this a bit.

ilkescality:
@dora-korpar @miniscruff I pushed the changes, should be good for re-reviews.

lib/api/apiUtils/object/objectLockHelpers.js
ilkescality: @dora-korpar I just noticed this comment of yours, I think that's a good idea to return a more meaningful error such as 'Unknown legal hold value' for the case of the invalid legal hold values (unlike aws which seems to be silently failing which is not ideal). I will take care of it.

tests/unit/api/objectPut.js
miniscruff: my comment still stands here, any thoughts @ilkescality ?

tests/unit/api/objectPut.js
```diff
@@ -162,6 +174,154 @@ describe('objectPut API', () => {
...
+        }, postBody);
+
+        bucketPut(authInfo, testPutBucketRequestLock, log, () => {
+            objectPut(authInfo, request, undefined, log, (err, headers) => {
+                assert.ifError(err);
+                assert.strictEqual(headers.ETag, `"${correctMD5}"`);
+                metadata.getObjectMD(bucketName, objectName, {}, log,
+                    (err, md) => {
+                        assert.ifError(err);
+                        assert(md);
```
miniscruff: this assertion is unnecessary since the next like would also check this one. Not sure it really adds much here.

tests/unit/api/objectPut.js
```diff
@@ -162,6 +174,154 @@ describe('objectPut API', () => {
...
+            type: 'Days',
+        },
+        {
+            mode: 'GOVERNANCE',
+            val: 5,
+            type: 'Years',
+        },
+    ];
+    testObjectLockConfigs.forEach(config => {
+        const { mode: testMode, type, val } = config;
```
miniscruff: could avoid mapping mode > testMode by just using testMode in the object above

tests/unit/api/objectPut.js
```diff
@@ -162,6 +174,154 @@ describe('objectPut API', () => {
...
+                                        = md.retentionInfo;
+                                    const date = new Date();
+                                    const days
+                                        = type === 'Days' ? val : val * 365;
+                                    const expectedDate
+                                        = date.setDate(date.getDate() + days);
+                                    assert.ifError(err);
+                                    assert(md);
+                                    assert.strictEqual(mode, testMode);
+                                    assert.strictEqual(retainDate.toString(),
```
miniscruff: not sure you need to convert the dates to strings here

tests/unit/api/apiUtils/objectLockHelpers.js
```diff
@@ -0,0 +1,169 @@
...
+    it('should calculate retainUntilDate for config with days', () => {
+        const mockConfigWithDays = {
+            mode: 'GOVERNANCE',
+            days: 90,
+        };
+        const date = new Date();
+        const expectedRetainUntilDate
+            = date.setDate(date.getDate() + mockConfigWithDays.days);
+        const retainUntilDate = calculateRetainUntilDate(mockConfigWithDays);
+        assert.strictEqual(retainUntilDate.toString().slice(0, 24),
```
miniscruff: instead of to string + slice you could use a fixed date above and with your expected.

tests/unit/api/apiUtils/objectLockHelpers.js
```diff
@@ -0,0 +1,168 @@
...
+        };
+        const objectLockValidation = validateHeaders(bucketInfo, headers, log);
+        const expectedError = errors.InvalidArgument.customizeDescription(
+            'Unknown wormMode directive');
+        assert.strictEqual(objectLockValidation.error.InvalidArgument, true);
+        assert.strictEqual(objectLockValidation.error.description,
+            expectedError.description);
+    });
+
+    it('should pass with invalid retention period header', () => {
```
miniscruff: this looks like it fails not passes

tests/unit/api/apiUtils/objectLockHelpers.js
```diff
@@ -0,0 +1,168 @@
...
+        const objectLockValidation = validateHeaders(bucketInfo, headers, log);
+        const expectedError = errors.InvalidArgument.customizeDescription(
+            'x-amz-object-lock-retain-until-date and x-amz-object-lock-mode ' +
+            'must both be supplied');
+        assert.strictEqual(objectLockValidation.error.InvalidArgument, true);
+        assert.strictEqual(objectLockValidation.error.description,
+            expectedError.description);
+    });
+
+    it('should pass with past retention date header', () => {
```
miniscruff: also looks like it fails with errors

tests/unit/api/apiUtils/objectLockHelpers.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    it('should pass with both legal hold and retention headers', () => {
+        const headers = {
+            'x-amz-object-lock-retain-until-date': '2050-10-12',
+            'x-amz-object-lock-mode': 'GOVERNANCE',
+            'x-amz-object-lock-legal-hold': 'ON',
+        };
+        const objectLockValidation = validateHeaders(bucketInfo, headers, log);
+        assert.strictEqual(objectLockValidation.error, undefined);
```
miniscruff: it might make more sense when validate headers fails it returns the error itself and when it passes it returns null. Instead of an extra layer of error: {stuff} it would just be {stuff} or null.

tests/unit/api/apiUtils/objectLockHelpers.js
```diff
@@ -0,0 +1,168 @@
...
+const mockCreationDate = new Date().toJSON();
+
+const bucketInfo = new BucketInfo(
+    mockName, mockOwner, mockOwnerDisplayName, mockCreationDate,
+    null, null, null, null, null, null, null, null, null, null,
+    null, null, true);
+
+const objLockDisabledBucketInfo = new BucketInfo(
+    mockName, mockOwner, mockOwnerDisplayName, mockCreationDate,
+    null, null, null, null, null, null, null, null, null, null,
```
miniscruff, line 23: uhh, probably unrelated to this PR but this is a bit much...

tests/unit/api/objectPut.js
ilkescality: (seems like I missed this one earlier) yeah it's not really a must-have; it's just we have many variations of `InvalidRequest` so I thought it might worth to have an extra check, but can remove it if it looks excessive.

lib/services.js
ilkescality: @dora-korpar done; now it's a separate method that can be reused.

tests/unit/api/objectPut.js
ilkescality: It doesn't add much; I suspect it might be linter related - just kept it similar with the other tests (guessing that the linter might fail for unused param) but will try removing it to see if it works.

tests/unit/api/apiUtils/objectLockHelpers.js
ilkescality: typo

tests/unit/api/apiUtils/objectLockHelpers.js
ilkescality: another typo

tests/unit/api/apiUtils/objectLockHelpers.js
ilkescality: yeahh, we're gonna refactor the BucketInfo constructor to get rid of this structure for good.

tests/unit/api/objectPut.js
miniscruff: My general rule here is that tests test functionality, what the error says doesn't matter as much as an error was thrown ( or in this case returned ). If the specific error was part of the functionality than turning it into some sort of object comparison would be ideal. A typo or updated message should not break a test basically.

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Sets object retention ond/or legal hold information on object's metadata
+ * @param {object} headers - request headers
+ * @param {object} md - object metadata
+ * @param {(object|null)} defaultRetention - bucket retention configuration if
+ * bucket has any configuration set
+ * @return {undefined}
+ */
+function setObjectLockInformation(headers, md, defaultRetention) {
```
dora-korpar, line 70: 💃

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Calculates retain until date for the locked object version
+ * @param {object} retention - includes days or years retention period
+ * @return {object} the date until the object version remains locked
+ */
+function calculateRetainUntilDate(retention) {
+    const { days, years } = retention;
+    const date = new Date();
+    // Calculate the number of days to retain the lock on the object
+    const retainUntilDays = days || years * 365;
```
dora-korpar, line 12: Does this line need parentheses here
```suggestion
    const retainUntilDays = days || (years * 365);
```
Not sure if without them this line might be read as 
```suggestion
    const retainUntilDays = (days || years) * 365;
```

miniscruff: image


Seems fine without parenthesis but I had to check

ilkescality: yup, conditional evaluation has higher precedence, so at first `days` gets evaluated, if it's not truthy then moves on to right part of `||` and sets the value to `years * 365` but I can add the parentheses still if it'd feel/look safer?

tests/unit/api/objectPut.js
ilkescality: (verified that removing it doesn't fail linter, gonna remove it with the next set of updates.)

lib/api/apiUtils/object/objectLockHelpers.js
dora-korpar: nah, it's cool. thanks for verifying

tests/unit/api/objectPut.js
ilkescality: That makes sense and I agree in the general context, although for our case, the error types are very generic (I.e. InvalidRequest or InvalidArgument) that returning the correct type doesn’t mean much/enough; in a sense, the correct error description is part of the functionality, so I thought it might still worth to keep it to verify the behavior by checking that we’re actually sending the right messages.

tests/unit/api/objectPut.js
ilkescality: I got misleading error such as below when I did without toString initially:
```
+expected -actual
-2025-06-11T04:41:28.633Z
+2025-06-11T04:41:28.634Z
```
That’s why I added it to bypass the false precision errors as making them string converts `2025-06-11T04:41:28.633Z` into `Tue Jun 10 2025 21:41:28 GMT-0700` getting rid of the last precision (`.633Z` and `.634Z` parts) but now .toISOString() takes care of it after switching to use `moment`.

tests/unit/api/apiUtils/objectLockHelpers.js
ilkescality: I don't think I could cause `calculateRetainUntilDate` calculates the date based on the current date which is why I needed to mimic the same logic (unless I misunderstood your comment 🤔 ).

tests/unit/api/objectPut.js
miniscruff: hmm I agree that those error types are too generic, and definitely not very helpful. Would it be possible ( maybe in the future ) to create more specific errors to return? Like MissingObjectLockConfigError, if not an instance of the generic error + description we can use to compare?

tests/unit/api/apiUtils/objectLockHelpers.js
miniscruff: oooh, I see, yea this would be much easier to test if `calculateRetainUntilDate` was a pure method and took in a date value. Because it has the side effect of getting the current date on its own, there will be nanosecond variations between the one calculated in the method and our test. Alternatively, you could use Sinon to mock the moment call to a specific date but pure functions are great regardless.

tests/unit/api/objectPut.js
miniscruff: This is fair, nothing I can see to get around the nanoseconds difference. Maybe mocking but not worth it for the difference.

tests/unit/api/objectPut.js
ilkescality: oh yeah I think that’s a great idea. The thing is, seems like we (unnecessarily) mimicked AWS logic internally as well and are only differentiating the errors on the description part, but, would be a lot better to have custom errors instead. I also see a lot of errors being returned with `.customizeDescription('some custom description')` on the fly with the exact same descriptions; in the same process, they could also be refactored to use that error type and extract the description from there which could be another improvement I think.

ilkescality: hmm actually the moment I wrote this I realized we already have `arsenalErrors.json` which could serve for this. I guess I could just add those errors with customized descriptive names and descriptions and use from there; e.g:
```
"InvalidRetentionDate": {
    "code": 400,
    "description": "The retain until date must be in the future!"
  }
```
It might get bit out of time (due to Arsenal involvement as well) for that PR but I think it's worth as a follow-up afterward.

tests/unit/api/apiUtils/objectLockHelpers.js
ilkescality: hmm actually I like the idea of using Sinon to mock the moment call to a specific date; not sure how I never thought of Sinon here at all lol (weird cause I'm used to using it a lot). Might be bit tight for this PR but I'll have some follow-up refactors (such as adding custom errors etc.) anyway so gonna give this a try to see if it's better mocked instead.

dora-korpar (approved):
good for me after a squash 👍

tests/unit/api/objectPut.js
miniscruff: Yup exactly the sort of stuff I was thinking, will resolve this comment but we should do something like that.

tests/unit/api/apiUtils/objectLockHelpers.js
miniscruff: Sounds good to me, resolve this if you are going to push it back.

miniscruff (approved):

ilkescality:
@bert-e approve

ilkescality:
@bert-e approve

PR #2629, opened 2020-06-10
ft:S3C-2792-mpu-object-retention

dora-korpar:
Updates initiateMPU API to accept object lock headers

### DIFF
--- a/lib/api/apiUtils/object/checkObjectLock.js
+++ b/lib/api/apiUtils/object/checkObjectLock.js
@@ -1,0 +1,47 @@
+const { errors } = require('arsenal');
+
+function checkObjectLock(bucket, headers) {
+    const retentionObj = {};
+
+    if (!bucket.isObjectLockEnabled()) {
+        retentionObj.error = errors.InvalidRequest.customizeDescription(
+            'Bucket is missing ObjectLockConfiguration');
+        return retentionObj;
+    }
+    const headerMode = headers['x-amz-object-lock-mode'];
+    const headerRetainDate = headers['x-amz-object-lock-retain-until-date'];
+    const headerLegalHold = headers['x-amz-object-lock-legal-hold-status'];
+    if ((headerMode && !headerRetainDate) || (!headerMode && headerRetainDate)) {
+        retentionObj.error = errors.InvalidArgument.customizeDescription(
+            'x-amz-object-lock-retain-until-date and x-amz-object-lock-mode ' +
+            'must both be supplied');
+        return retentionObj;
+    }
+    const validModes = new Set(['GOVERNANCE', 'COMPLIANCE']);
+    if (headerMode && !validModes.has(headerMode)) {
+        retentionObj.error = errors.InvalidArgument.customizeDescription(
+            'Unknown wormMode directive');
+        return retentionObj;
+    }
+    if (headerLegalHold && headerLegalHold !== 'ON') {
+        // AWS behavior does not return an error in this case, but proceeds
+        // without setting a legal hold. Returning an error will inform the
+        // customer
+        retentionObj.error = errors.InvalidArgument.customizeDescription(
+            'Legal hold value must be "ON". To disable legal hold, omit the ' +
+            'header');
+        return retentionObj;
+    }
+    const objectLockConfig = bucket.getObjectLockConfiguration();
+    const bucketMode = objectLockConfig.retentionInfo.mode;
+    const bucketRetainDate = objectLockConfig.retentionInfo.retainDate;
+    const bucketLegalHold = objectLockConfig.legalHold;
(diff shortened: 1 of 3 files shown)


lib/api/initiateMultipartUpload.js
```diff
@@ -189,6 +190,17 @@ function initiateMultipartUpload(authInfo, request, log, callback) {
...
+        const retentionObj = checkObjectLock(destinationBucket, request.headers);
+        if (retentionObj.error) {
+            return callback(retentionObj.error);
+        }
+        if (retentionObj.retentionInfo) {
+            metadataStoreParams.retentionInfo = retentionObj.retentionInfo;
+        }
+        if (retentionObj.legalHold) {
+            metadataStoreParams.legalHold = retentionObj.legalHold;
+        }
```
miniscruff, line 202: I am pretty sure you can just merge the objects here for the same results
```suggestion
        metadataStoreParams = {...metadataStoreParams, ...retentionObj}
```

lib/api/apiUtils/object/checkObjectLock.js
```diff
@@ -0,0 +1,47 @@
+const { errors } = require('arsenal');
+
+function checkObjectLock(bucket, headers) {
```
miniscruff, line 3: I find this name a little misleading. It does check object locks sure but then returns, what I am mostly guessing on, the object lock data of retention info, and legal hold. I think a name that describes this return value and/or the possibility of the error might make more sense. Maybe like `tryGetObjectLock` or `getLockMetadata` not sure. 

Although if this format is used in a lot of places than ignore my comment lol.

lib/api/initiateMultipartUpload.js
dora-korpar: 😮 didn't know that trick, thanks!

dora-korpar: hmmm our linter seems to not like that syntax tho

miniscruff: Grrr the linter doesn't like any spread notation I guess, probably needs to be updated. Object.assign will also work but doesn't look as cool IMO.

tests/unit/api/multipartUpload.js
```diff
@@ -49,6 +55,26 @@ const initiateRequest = {
...
 };
+const retentionInitiateRequest = Object.assign(initiateRequest, {});
+retentionInitiateRequest.headers['x-amz-object-lock-mode'] = 'GOVERNANCE';
+retentionInitiateRequest.headers['x-amz-object-lock-retain-until-date'] =
+    futureDate;
+retentionInitiateRequest.bucketName = lockedBucket;
+const legalHoldInitiateRequest = Object.assign(initiateRequest, {});
+legalHoldInitiateRequest.headers['x-amz-object-lock-mode'] = 'GOVERNANCE';
+legalHoldInitiateRequest.headers['x-amz-object-lock-retain-until-date'] =
+    futureDate;
```
miniscruff, line 66: ( without using the spread operator this is what I came up with for this block... note that spread and object.assign does not merge recursively so you have to do two steps )
```suggestion
amzHeaders = {
  ['x-amz-object-lock-mode']: 'GOVERNANCE',
  ['x-amz-object-lock-retain-until-date']: futureData,
}
const legalHoldInitiateRequest = Object.assign(initiateRequest, {});
legalHoldInitiateRequest.headers = Object.assign(legalHoldInitiateRequest.headers, amzHeaders);
const retentionInitiateRequest = Object.assign(legalHoldInitiateRequest, {bucketName: lockedBucket});
```

lib/api/apiUtils/object/checkObjectLock.js
```diff
@@ -0,0 +1,47 @@
...
+    const retentionObj = {};
+
+    if (!bucket.isObjectLockEnabled()) {
+        retentionObj.error = errors.InvalidRequest.customizeDescription(
+            'Bucket is missing ObjectLockConfiguration');
+        return retentionObj;
+    }
+    const headerMode = headers['x-amz-object-lock-mode'];
+    const headerRetainDate = headers['x-amz-object-lock-retain-until-date'];
+    const headerLegalHold = headers['x-amz-object-lock-legal-hold-status'];
```
miniscruff, line 13: It might be worth moving these header strings into a const.js or something, they are used quite a bit and typos will be hard to find.

miniscruff (changes requested):
Just doing request changes to prevent merges before comments are resolved ( wonder if GitHub has a "do not merge if issues unresolved" flag... )

dora-korpar:
@miniscruff I'm going to close this and rebase my code from this logic: https://github.com/scality/cloudserver/pull/2613/files since it's doing the same thing

PR #2633, opened 2020-06-12
feature: S3C-2791 extend copy object for object lock

ilkescality:
Extends COPY Object API to support object lock feature. An object should be copied from a source to a destination with additional `x-amz-object-lock-retain-until-date`, `x-amz-object-lock-mode`, `x-amz-object-lock-legal-hold` headers.

Tests cases covering the scenarios:

1. COPY object request with retention headers made to an object lock disabled destination bucket, should return `InvalidRequest` error.
`"An error occurred (InvalidRequest) when calling the CopyObject operation: Bucket is missing ObjectLockConfiguration".`
2. COPY object request without retention headers made to an object lock enabled destination bucket with default retention should copy the object with default retention of its destination bucket.
3. COPY object request without retention headers but with legal hold header made to an object lock enabled bucket with default retention should copy the object with default retention of its destination bucket.
4. COPY object request with retention headers made to an object lock enabled destination bucket with default retention should copy the object with its own retention.
5. COPY object request without retention headers made to an object lock enabled bucket without default retention should copy the object without any retention.
6. If the source object has retention configuration, the retention configuration should not be copied. (Unless either COPY object request has retention headers or destination bucket has default retention, the object should be not copied with retention configuration.)

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/objectCopy.js
+++ b/tests/functional/aws-node-sdk/test/object/objectCopy.js
@@ -1,5 +1,4 @@
 const assert = require('assert');
-
 const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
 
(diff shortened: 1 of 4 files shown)


tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -1197,3 +1224,128 @@ describe('Object Copy', () => {
...
+        }).then(res => {
+            lastModified = res.LastModified;
+        }));
+
+        afterEach(() => bucketUtil.empty(sourceBucketName)
+            .then(() => bucketUtil.empty(destBucketName)));
+
+        after(() => bucketUtil.deleteMany([sourceBucketName, destBucketName]));
+
+        it('should not copy default retention info of the destination ' +
```
dora-korpar, line 1272: I think this functionality from AWS is so weird, but good catch

dora-korpar (approved):

anurag4DSB (approved):

ilkescality:
@bert-e approve

PR #2784, opened 2020-07-01
Bugfix/s3 c 3106 multiobj delete obj lock error

dora-korpar:
First commit changes a helper function for object lock tests and updates all the tests using it. Second commit fixes a bug with a missing reference and adds tests.

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/deleteObject.js
+++ b/tests/functional/aws-node-sdk/test/object/deleteObject.js
@@ -3,7 +3,7 @@ const moment = require('moment');
 const Promise = require('bluebird');
 const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
-const removeObjectLock = require('../../lib/utility/objectLock-util');
+const changeObjectLock = require('../../../../utilities/objectLock-util');
 
 const bucketName = 'testdeletempu';
 const objectName = 'key';
@@ -217,12 +217,12 @@ describe('DELETE object', () => {
                     VersionId: versionIdTwo,
                 }, err => {
                     assert.strictEqual(err.code, 'AccessDenied');
-                    removeObjectLock(
+                    changeObjectLock(
                         [{
                             bucket: bucketName,
                             key: objectNameTwo,
                             versionId: versionIdTwo,
-                        }], done);
+                        }], '', done);
                 });
             });
         });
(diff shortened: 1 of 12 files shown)


lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -123,10 +123,13 @@ function isObjectLocked(bucket, objectMD, headers) {
         if (objectLegalHold) {
             return true;
         }
-        const objectRetention = objectMD.retentionInfo;
+        let objectRetention = objectMD.retentionInfo;
```
rahulreddy: I see a lot of objectMD usage where properties are accessed directly. The reason we have ObjectMD is so that we avoid this and set/get properties using setters/getters. It is also the root cause of this bug.

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -123,10 +123,13 @@ function isObjectLocked(bucket, objectMD, headers) {
...
         if (!objectRetention) {
             return false;
         }
+        if (!objectRetention.retention) {
+            objectRetention = { retention: objectRetention };
```
rahulreddy: Isn't this a circular reference?
```js
let objectRetention = objectMD.retentionInfo;
objectRetention = { retention: objectRetention };
```
Also, `retention` is not a property in ObjectMD as I see here https://github.com/scality/Arsenal/blob/development/7.7/lib/models/ObjectMD.js#L911

rahulreddy (changes requested):
I didn't look at the tests but I have a lot of questions on how properties are being used here.

lib/api/apiUtils/object/objectLockHelpers.js
dora-korpar: resolved with the great flattening https://github.com/scality/cloudserver/pull/2794

dora-korpar:
@bert-e reset

lib/api/apiUtils/object/objectLockHelpers.js
dora-korpar: As discussed, how we use objectMD is flawed throughout Cloudserver. It would take a good bit of refactoring to make using getters/setters viable, which is outside the scope of this pr

ilkescality (approved):

lib/api/apiUtils/object/objectLockHelpers.js
rahulreddy: Thanks for looking into it @dora-korpar . Let's follow up as a tech debt to address.

rahulreddy (approved):

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e approve

PR #2826, opened 2020-07-02
bugfix: S3C-3121 fix put bucket api to create unlocked bucket

ilkescality:
Fixes bucket creation with object lock; AWS accepts booleans both as string or boolean, as a result, when `false` passed as a string, previous implementation interpreted it as truthy therefore buckets were always getting created with object lock enabled.

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -165,7 +165,9 @@ function createBucket(authInfo, bucketName, headers,
     const ownerDisplayName =
         authInfo.getAccountDisplayName();
     const creationDate = new Date().toJSON();
-    const objectLockEnabled = headers['x-amz-bucket-object-lock-enabled'];
+    const headerObjectLock = headers['x-amz-bucket-object-lock-enabled'];
+    const objectLockEnabled
+        = headerObjectLock && headerObjectLock.toLowerCase() === 'true';
     const bucket = new BucketInfo(bucketName, canonicalID, ownerDisplayName,
         creationDate, BucketInfo.currentModelVersion(), null, null, null, null,
         null, null, null, null, null, null, null, objectLockEnabled);
--- a/tests/functional/aws-node-sdk/test/bucket/put.js
+++ b/tests/functional/aws-node-sdk/test/bucket/put.js
@@ -194,15 +194,34 @@ describe('PUT Bucket - AWS.S3.createBucket', () => {
         });
 
         describe('bucket creation success with object lock', () => {
-            function _testObjectLock(name, done) {
+            function _testObjectLockEnabled(name, done) {
                 bucketUtil.s3.createBucket({
                     Bucket: name,
                     ObjectLockEnabledForBucket: true,
                 }, (err, res) => {
                     assert.ifError(err);
+                    assert.strictEqual(res.Location, `/${name}`,
+                    'Wrong Location header');
+                    bucketUtil.s3.getObjectLockConfiguration({ Bucket: name }, (err, res) => {
+                        assert.ifError(err);
+                        assert.deepStrictEqual(res.ObjectLockConfiguration,
+                            { ObjectLockEnabled: 'Enabled' });
+                    });
+                    bucketUtil.deleteOne(name).then(done).catch(done);
+                });
+            }
+            function _testObjectLockDisabled(name, done) {
+                bucketUtil.s3.createBucket({
+                    Bucket: name,
+                    ObjectLockEnabledForBucket: false,
+                }, (err, res) => {
+                    assert.ifError(err);
                     assert(res.Location, 'No Location in response');
                     assert.strictEqual(res.Location, `/${name}`,
                         'Wrong Location header');
+                    bucketUtil.s3.getObjectLockConfiguration({ Bucket: name }, err => {
+                        assert.strictEqual(err.code, 'ObjectLockConfigurationNotFoundError');
+                    });
                     bucketUtil.deleteOne(name).then(() => done()).catch(done);
                 });
             }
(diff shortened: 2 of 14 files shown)


lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -165,7 +165,9 @@ function createBucket(authInfo, bucketName, headers,
     const ownerDisplayName =
         authInfo.getAccountDisplayName();
     const creationDate = new Date().toJSON();
-    const objectLockEnabled = headers['x-amz-bucket-object-lock-enabled'];
+    const headerObjectLock = headers['x-amz-bucket-object-lock-enabled'];
+    const objectLockEnabled = headerObjectLock === 'true'
+        || headerObjectLock === true;
```
rahulreddy: I don't think this can be a boolean. All header values are strings.

ilkescality: You're right, just verified with AWS cli and sdk as well; unit test failures tricked me into it but I just edited those tests so should be good now. :)

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -165,7 +165,8 @@ function createBucket(authInfo, bucketName, headers,
     const ownerDisplayName =
         authInfo.getAccountDisplayName();
     const creationDate = new Date().toJSON();
-    const objectLockEnabled = headers['x-amz-bucket-object-lock-enabled'];
+    const objectLockEnabled
+        = headers['x-amz-bucket-object-lock-enabled'] === 'true';
```
miniscruff: I thought headers was a string>string map so it would always be a string. Edit: Same as what Rahul said above...

Also should you add a toLower here in case of "TRUE" or "True"?

nicolas2bert: What about "on", "yes" and "1"?

miniscruff: I would only include what amazon is also expecting, so if they support any sort of truthy string such as on, yes, etc then I agree. If it explicitly only supports true/false then I wouldn't include anymore.

nicolas2bert: Yes, indeed, it needs to be first checked against AWS to see what it supports.

ilkescality: @miniscruff that's a good catch; seems like both 'true' or 'True' are covered by AWS, @nicolas2bert those are not supported by AWS.

miniscruff: For string comparisons, I always recommend case insensitive or locale comparisons.

ilkescality: Yeah, I can see how it makes sense; i.e. when I digged deeper I realized that when the request made via CLI with just the presence of the `--object-lock-enabled-for-bucket` flag without any value (this is how CLI command works), when looked into how it works under the hood, seems like it converts it's value as `x-amz-bucket-object-lock-enabled: True` whereas a request via sdk it only allows boolean and converts it lowercase as `x-amz-bucket-object-lock-enabled: true`

tests/unit/api/bucketGetObjectLock.js
```diff
@@ -20,7 +20,7 @@ const testBucketPutReqWithObjLock = {
     bucketName,
     headers: {
         'host': `${bucketName}.s3.amazonaws.com`,
-        'x-amz-bucket-object-lock-enabled': true,
+        'x-amz-bucket-object-lock-enabled': 'true',
```
miniscruff: does not need to be this one, but can you set one of the `true` values to different cases like `True`?

ilkescality: good idea, just updated while I was on it

miniscruff: As a partially unrelated note, header keys should also be case insensitive such that the key `X-AMZ-bucket-object-lock-ENABLED` should also still work. Spec is here https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

rahulreddy: Node.js takes care of the casing and we can rely on it being lowercased https://nodejs.org/dist/latest-v10.x/docs/api/http.html#http_message_headers

miniscruff: ok gotcha, thats good. Not all tools do that for you. Also, not everyone knows that headers are handled that way.

tests/functional/aws-node-sdk/test/bucket/put.js
```diff
@@ -194,15 +194,35 @@ describe('PUT Bucket - AWS.S3.createBucket', () => {
...
         describe('bucket creation success with object lock', () => {
-            function _testObjectLock(name, done) {
+            function _testObjectLockEnabled(name, done) {
                 bucketUtil.s3.createBucket({
                     Bucket: name,
                     ObjectLockEnabledForBucket: true,
+                }, (err, res) => {
+                    assert.ifError(err);
+                    assert(res.Location, 'No Location in response');
+                    assert.strictEqual(res.Location, `/${name}`,
```
miniscruff: First assert is redundant as it can not be falsy and also strictly equal.
```suggestion
                    assert.strictEqual(res.Location, `/${name}`,
```

tests/functional/aws-node-sdk/test/bucket/put.js
```diff
@@ -194,15 +194,35 @@ describe('PUT Bucket - AWS.S3.createBucket', () => {
...
+                    'Wrong Location header');
+                    bucketUtil.s3.getObjectLockConfiguration({ Bucket: name }, (err, res) => {
+                        assert.ifError(err);
+                        assert.deepStrictEqual(res.ObjectLockConfiguration,
+                            { ObjectLockEnabled: 'Enabled' });
+                    });
+                    bucketUtil.deleteOne(name).then(() => done()).catch(done);
```
miniscruff, line 210: Can just pass in done
```suggestion
                    bucketUtil.deleteOne(name).then(done).catch(done);
```

tests/unit/api/bucketPutObjectLock.js
```diff
@@ -56,7 +56,7 @@ describe('putBucketObjectLock API', () => {
 
     describe('with Object Lock enabled on bucket', () => {
         const bucketObjLockRequest = Object.assign({}, bucketPutRequest,
-            { headers: { 'x-amz-bucket-object-lock-enabled': true } });
+            { headers: { 'x-amz-bucket-object-lock-enabled': 'true' } });
```
miniscruff, line 59: For sanity, it may be worth adding a few different casings just to be complete.

miniscruff (approved):
Small optional changes for your consideration

tests/unit/api/bucketPutObjectLock.js
ilkescality: done ☑️

rahulreddy (approved):

PR #2849, opened 2020-07-07
bf: S3C-3145 - add objectLocked field to AccessDenied errors

dora-korpar:
Removes excessive log output for objectLock in multiObjectDelete API, and adds `objectLocked: true` to output of both multiObjectDelete and objectDelete APIs.

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -187,6 +187,7 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
     let totalContentLengthDeleted = 0;
     let numOfObjectsRemoved = 0;
     const skipError = new Error('skip');
+    const objectLockedError = new Error('object locked');
 
     // doing 5 requests at a time. note that the data wrapper
     // will do 5 parallel requests to data backend to delete parts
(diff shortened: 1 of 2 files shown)


lib/api/multiObjectDelete.js
```diff
@@ -255,6 +256,9 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
         ], (err, objMD, deleteInfo, versionId) => {
             if (err === skipError) {
                 return moveOn();
+            } else if (err === objectLockedError) {
```
naren-rajendran, line 259: Is the comparison valid?

dora-korpar: what do you mean? the error is returned as `objectLockedError` on line 233 above. and it passed manual testing :)

naren-rajendran: What I meant was comparing error objects, I thought it would fail, but it works.

naren-rajendran (approved):

lib/api/multiObjectDelete.js
jonathan-gramain: Should work since `objectLockedError` is a singleton instance, it's always referring to the same object.

lib/api/multiObjectDelete.js
```diff
@@ -255,6 +256,9 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
         ], (err, objMD, deleteInfo, versionId) => {
             if (err === skipError) {
```
jonathan-gramain, line 257: A minor style comment: the code flow could be slightly improved with the following structure to avoid repeating `return moveOn()`:

```
if (err) {
    if (err === skipError) {
    } else if (...) {
    } else {
        //generic error
    }
    return moveOn();
}
```

jonathan-gramain (approved):

dora-korpar:
@bert-e approve

PR #2912, opened 2020-07-31
Feature/s3 c 3132 utapi v2 push metric

dora-korpar:
Updates pushMetric function and calls in each API to provide more information for pushMetric v2

### DIFF
--- a/lib/utapi/utilities.js
+++ b/lib/utapi/utilities.js
@@ -2,7 +2,7 @@ const http = require('http');
 const https = require('https');
 const commander = require('commander');
 const { auth } = require('arsenal');
-const { UtapiClient } = require('utapi');
+const { UtapiClient, utapiVersion } = require('utapi');
 const logger = require('../utilities/logger');
 const _config = require('../Config').config;
 // setup utapi client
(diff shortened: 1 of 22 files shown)


lib/utapi/utilities.js
```diff
@@ -206,8 +223,40 @@ function listMetrics(metricType) {
...
+
+    if (utapiVersion === 2) {
+        const utapiObj = {
+            operationId: action,
+            timestamp:
+            bucket,
+            location,
+            objectDelta: numberOfObjects,
+            sizeDelta: byteLength,
+            incomingBytes: -byteLength || newByteLength,
```
tmacro: I'm not sure if these are right. `incomingBytes` and `outgoingBytes` represents the amount of data received or sent by the server during the operation.

tmacro (approved):

lib/utapi/utilities.js
tmacro: I meant to comment and not approve, but I can't figure out how to change it

naren-rajendran (approved):

lib/utapi/utilities.js
miniscruff: You can just re-review and it overrides your previous.

dora-korpar: oh so they're not a running total? would the incomingBytes on a delete operation just be zero then?

tmacro: yes & yes, the event reflects how an op would change the totals

dora-korpar:
@bert-e approve

PR #2931, opened 2020-08-11
bf: S3C-3246 getObject pushMetric missing key

dora-korpar:
Adds missing `keys` to pushMetric call in getObject, and adds check in conditional so similar error is not thrown for operations that do not include `keys`.


tmacro (approved):

miniscruff (approved):

ilkescality (approved):

dora-korpar:
@bert-e approve

PR #2934, opened 2020-08-11
ft: S3C-3238 add bucket notification config

dora-korpar:
Adds assertions for bucket notification configuration

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -296,7 +296,7 @@ function bucketNotifAssert(bucketNotifConfig) {
         assert(typeof auth === 'object',
             'bad config: bucket notification auth must be an object');
         const { cert, user, password } = auth;
-        if ((cert && user && password) || (!cert && !user && !password)) {
+        if ((cert && (user || password)) || (!cert && !user && !password)) {
             throw new Error('bad config: bucket notification configuration ' +
                 'auth should contain either cert or user and password');
         }
@@ -307,7 +307,7 @@ function bucketNotifAssert(bucketNotifConfig) {
                 `File not found or unreachable: ${certpath}`);
             // eslint-disable-next-line no-param-reassign
             c.auth.cert = certpath;
-        } else if (!(user && password)) {
+        } else if (!user || !password) {
             throw new Error('bad config: bucket notification configuration ' +
                 'auth should contain both user and password if not using cert');
         } else {
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -280,6 +280,46 @@ function requestsConfigAssert(requestsConfig) {
     }
 }
 
+function bucketNotifAssert(bucketNotifConfig) {
+    assert(Array.isArray(bucketNotifConfig),
+        'bad config: bucket notification configuration must be an array');
+    bucketNotifConfig.forEach(c => {
+        const { resource, type, host, port, auth } = c;
+        assert(typeof resource === 'string',
+            'bad config: bucket notification configuration resource must be a string');
+        assert(typeof type === 'string',
+            'bad config: bucket notification configuration type must be a string');
+        assert(typeof host === 'string' && host !== '',
+            'bad config: hostname must be a non-empty string');
+        assert(Number.isInteger(port) && port > 0,
+            'bad config: port must be a positive integer');
+        assert(typeof auth === 'object',
+            'bad config: bucket notification auth must be an object');
+        const { cert, user, password } = auth;
+        if ((cert && user && password) || (!cert && !user && !password)) {
+            throw new Error('bad config: bucket notification configuration ' +
+                'auth should contain either cert or user and password');
+        }
+        if (cert) {
+            const certpath = (cert[0] === '/') ? cert : `${this._basePath}/${cert}`;
+            assert.doesNotThrow(() =>
+                fs.accessSync(certpath, fs.F_OK | fs.R_OK),
+                `File not found or unreachable: ${certpath}`);
+            // eslint-disable-next-line no-param-reassign
+            c.auth.cert = certpath;
+        } else if (!(user && password)) {
+            throw new Error('bad config: bucket notification configuration ' +
+                'auth should contain both user and password if not using cert');
+        } else {
+            assert(typeof user === 'string',
+                'bad config: bucket notification configuration auth user should be a string');
+            assert(typeof password === 'string',
+                'bad config: bucket notification configuration auth password should be a string');
+        }
+    });
+    return bucketNotifConfig;
+}
+
 /**
  * Reads from a config file and returns the content as a config object
  */
@@ -1060,6 +1100,9 @@ class Config extends EventEmitter {
             requestsConfigAssert(config.requests);
             this.requests = config.requests;
         }
+        if (config.bucketNotificationTargets) {
+            this.bucketNotificationTargets = bucketNotifAssert(config.bucketNotificationTargets);
+        }
     }
 
     _configureBackends() {
@@ -1256,4 +1299,5 @@ module.exports = {
     ConfigObject: Config,
     config: new Config(),
     requestsConfigAssert,
+    bucketNotifAssert,
 };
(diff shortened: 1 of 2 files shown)


lib/Config.js
```diff
@@ -280,6 +280,39 @@ function requestsConfigAssert(requestsConfig) {
     }
 }
 
+function bucketNotifAssert(bucketNotifConfig) {
```
miniscruff, line 283: Placing my comment here so replies are easier but it's more of a general comment.

Using assert in production code isn't wrong, in JS anyway, python would be as asserts are ignored under some build conditions. But my real concern is that this same logic would use JOI in other projects but uses assert here. So not sure where to have this discussion but I think using one form of validation would be ideal. Whether that is JOI/other tools, plain if statements or asserts is whatever to me as long as its, for the most part, the same.

If there are other concerns like this project is too old to use JOI or something then sure but I could not find anything in the ticket or descriptions so just pointing this out.

lib/Config.js
```diff
@@ -280,6 +280,39 @@ function requestsConfigAssert(requestsConfig) {
...
+        'bad config: bucket notification configuration resource must be a string');
+    assert(typeof type === 'string',
+        'bad config: bucket notification configuration type must be a string');
+    assert(typeof host === 'string' && host !== '',
+        'bad config: hostname must be a non-empty string');
+    assert(Number.isInteger(port) && port > 0,
+        'bad config: port must be a positive integer');
+    assert(typeof auth === 'object',
+        'bad config: bucket notification auth must be an object');
+    const { cert, user, password } = auth;
```
naren-rajendran: I am still figuring out auth part of the destinations. This may change in the future. For the time being this works.

naren-rajendran (changes requested):
Can we add the `bucketNotifAssert` function in arsenal instead of cloudserver, because I have to use the same in backbeat.

naren-rajendran (approved):
We should add the `bucketNotifAssert` function in arsenal instead of cloudserver, because the same will be used in `backbeat`. Approving this for the time being and the change will be implemented as an improvement.

dora-korpar:
@bert-e reset

lib/Config.js
dora-korpar: Yeah, that's a valid point. I don't think the project is too old to use JOI, but it just wasn't used when it was first written, and asserts have just been continually added. Converting these configs to use JOI is a side project we can tackle when we refactor Cloudserver for node12 or around then

miniscruff: Ok, that is fine. Using uniform practices will make reviewing and contributing a lot easier in the long run.

lib/Config.js
```diff
@@ -280,6 +280,46 @@ function requestsConfigAssert(requestsConfig) {
...
+        assert(typeof type === 'string',
+            'bad config: bucket notification configuration type must be a string');
+        assert(typeof host === 'string' && host !== '',
+            'bad config: hostname must be a non-empty string');
+        assert(Number.isInteger(port) && port > 0,
+            'bad config: port must be a positive integer');
+        assert(typeof auth === 'object',
+            'bad config: bucket notification auth must be an object');
+        const { cert, user, password } = auth;
+        if ((cert && user && password) || (!cert && !user && !password)) {
```
miniscruff: It probably doesn't happen often but I think the first set should be `cert && (user || password)` as it would be weird to specify a cert and a user even if you forgot the password and vice versa.

lib/Config.js
```diff
@@ -280,6 +280,46 @@ function requestsConfigAssert(requestsConfig) {
...
+                'auth should contain either cert or user and password');
+        }
+        if (cert) {
+            const certpath = (cert[0] === '/') ? cert : `${this._basePath}/${cert}`;
+            assert.doesNotThrow(() =>
+                fs.accessSync(certpath, fs.F_OK | fs.R_OK),
+                `File not found or unreachable: ${certpath}`);
+            // eslint-disable-next-line no-param-reassign
+            c.auth.cert = certpath;
+        } else if (!(user && password)) {
```
miniscruff: Not sure if this is any better looking or easier to read but you could also do:
```suggestion
        } else if (!user || !password) {
```

lib/Config.js
```diff
@@ -280,6 +280,46 @@ function requestsConfigAssert(requestsConfig) {
...
+        assert(typeof auth === 'object',
+            'bad config: bucket notification auth must be an object');
+        const { cert, user, password } = auth;
+        if ((cert && user && password) || (!cert && !user && !password)) {
+            throw new Error('bad config: bucket notification configuration ' +
+                'auth should contain either cert or user and password');
+        }
+        if (cert) {
+            const certpath = (cert[0] === '/') ? cert : `${this._basePath}/${cert}`;
+            assert.doesNotThrow(() =>
```
miniscruff, line 305: This assert doesn't feel necessary.
We try and do something and assert it doesn't fail, but when it does we throw an exception cause the assert failed.
It would be just as reasonable to just do accessSync and on failure, it will throw anyway.

tests/unit/testConfigs/bucketNotifConfigTest.js
```diff
@@ -0,0 +1,142 @@
+const assert = require('assert');
+const { bucketNotifAssert } = require('../../../lib/Config');
+
+describe('bucketNotifAssert', () => {
+    it('should not throw an error if bucket notification config is valid', () => {
+        assert.doesNotThrow(() => {
```
miniscruff: https://nodejs.org/api/assert.html#assert_assert_doesnotthrow_fn_error_message doesNotThrow doesn't really do anything as stated in the docs. If you call a method and everything is peachy then the test passes. If it fails the test will fail. No, assert required. 
Note that the extra parameter of error message is also redundant with the test name that is displayed in the logs anyway.

tests/unit/testConfigs/bucketNotifConfigTest.js
```diff
@@ -0,0 +1,142 @@
...
+        assert.throws(() => {
+            bucketNotifAssert({
+                resource: 'target1',
+                type: 'kafka',
+                host: 'localhost',
+                port: 8000,
+                auth: { user: 'user', password: 'password' },
+            });
+        },
+        '/bad config: bucket notification configuration must be an array/');
```
miniscruff, line 24: this is the part where using asserts over if's kinda sucks. Using regex to match now requires us to keep error messages in sync with test code. That is, updating an error message will break a test. This, to me anyway, is not ideal as I should be able to update "flavor" text without breaking a test. Consider if a page header change broke a test, it doesn't really add any value for that. 

Anyway, that is my rant. I would say just asserting that it throws and ignoring the error message is probably fine but this is more of a team thing so up to everyone.

miniscruff (changes requested):

tests/unit/testConfigs/bucketNotifConfigTest.js
dora-korpar: This test format was basically copy/pasted from other config unit tests. I agree that it's not super practical to use error messages in tests, but it's the only way we have to differentiate where in the code the error is thrown. Only checking for an error leaves the possibility of there being an unintentional error in a test case or something like that. I'm defo into figuring out some better way of doing it, but I think the error messages need to be included.

tests/unit/testConfigs/bucketNotifConfigTest.js
dora-korpar: I can remove the `doesNotThrow`, but see my below response about the error messages

dora-korpar:
@bert-e wait

dora-korpar:
@bert-e approve

tests/unit/testConfigs/bucketNotifConfigTest.js
miniscruff: One method would be to turn the error messages into constants and compare that way. Comparing a thrown error against the expected const error ( as a string or error object ) would be fine imo. But I will approve either way as I am very late to the party.

miniscruff (approved):
Looks good, the issue of how to test errors or not is a bigger topic so I wouldn't stop the PR for it.

PR #2946, opened 2020-08-12
Feature/s3 c 2797 put bucket notification

dora-korpar:
Adds `putBucketNotification` API.
Requires https://github.com/scality/cloudserver/pull/2934 and https://github.com/scality/Arsenal/pull/1231
to be merged before.
The new `bucketNotificationTargets` field in the config is needed to run the tests.
There was also an attempt to make this API an async function using current node async/await and promise practices, but it would require more refactoring outside the scope of this PR.

### DIFF
--- a/config.json
+++ b/config.json
@@ -77,5 +77,14 @@
         "viaProxy": false,
         "trustedProxyCIDRs": [],
         "extractClientIPFromHeader": ""
-    }
+    },
+    "bucketNotificationDestinations": [
+        {
+            "resource": "target1",
+            "type": "dummy",
+            "host": "localhost",
+            "port": 6000,
+            "auth": { "user": "user", "password": "password" }
+        }
+    ]
 }
--- a/lib/api/apiUtils/bucket/getNotificationConfiguration.js
+++ b/lib/api/apiUtils/bucket/getNotificationConfiguration.js
@@ -1,0 +1,25 @@
+const { errors, models } = require('arsenal');
+const { NotificationConfiguration } = models;
+
+const { config } = require('../../../Config');
+
+function getNotificationConfiguration(parsedXml) {
+    const notifConfig = new NotificationConfiguration(parsedXml).getValidatedNotificationConfiguration();
+    if (notifConfig.error) {
+        return notifConfig;
+    }
+    if (!config.bucketNotificationDestinations) {
+        return { error: errors.InvalidArgument.customizeDescription(
+            'Unable to validate the following destination configurations') };
+    }
+    const targets = new Set(config.bucketNotificationDestinations.map(t => t.resource));
+    const notifConfigTargets = notifConfig.queueConfig.map(t => t.queueArn.split(':')[5]);
+    if (!notifConfigTargets.every(t => targets.has(t))) {
+        return { error: errors.MalformedXML.customizeDescription(
+            'queue arn target is not included in Cloudserver bucket notification config'),
+        };
+    }
+    return notifConfig;
+}
+
+module.exports = getNotificationConfiguration;
--- a/lib/api/bucketPutNotification.js
+++ b/lib/api/bucketPutNotification.js
@@ -1,0 +1,58 @@
+const async = require('async');
+
+const parseXML = require('../utilities/parseXML');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const getNotificationConfiguration = require('./apiUtils/bucket/getNotificationConfiguration');
+const metadata = require('../metadata/wrapper');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const { pushMetric } = require('../utapi/utilities');
+
+/**
+ * Bucket Put Notification - Create or update bucket notification configuration
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+
+function bucketPutNotification(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketPutNotification' });
+
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketPutNotification',
+    };
+
+    return async.waterfall([
+        next => parseXML(request.post, log, next),
+        (parsedXml, next) => {
+            const notificationConfig = getNotificationConfiguration(parsedXml);
+            process.nextTick(() => next(notificationConfig.error, notificationConfig));
+        },
+        (notifConfig, next) => metadataValidateBucket(metadataValParams, log,
+            (err, bucket) => next(err, bucket, notifConfig)),
+        (bucket, notifConfig, next) => {
+            bucket.setNotificationConfiguration(notifConfig);
+            metadata.updateBucket(bucket.getName(), bucket, log,
+                err => next(err, bucket));
+        },
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        if (err) {
+            log.trace('error processing request', { error: err,
+                method: 'bucketPutNotification' });
+            return callback(err, corsHeaders);
+        }
+        pushMetric('putBucketNotification', log, {
+            authInfo,
+            bucket: bucketName,
+        });
+        return callback(null, corsHeaders);
+    });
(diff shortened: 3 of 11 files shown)


config.json
```diff
@@ -77,5 +77,14 @@
         "viaProxy": false,
         "trustedProxyCIDRs": [],
         "extractClientIPFromHeader": ""
-    }
+    },
+    "bucketNotificationTargets": [
```
naren-rajendran: The property is renamed to `bucketNotificationDestinations`, ref https://github.com/scality/Federation/blob/4bf06519734ea45479ecb60dbd0a1cb86e4223a9/roles/run-s3/templates/config.json.j2#L139

naren-rajendran (approved):
Just the property name change, otherwise looks good to me.

dora-korpar:
@bert-e reset

lib/api/apiUtils/bucket/getNotificationConfiguration.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+    const notifConfig = new NotificationConfiguration(parsedXml).getValidatedNotificationConfiguration();
+    if (notifConfig.error) {
+        return notifConfig;
+    }
+    const notifConfigTargets = [];
+    notifConfig.queueConfig.forEach(t => {
+        const arnArray = t.queueArn.split(':');
+        notifConfigTargets.push(arnArray[5]);
+    });
```
miniscruff: for each and push is just a map.
```suggestion
    const notifConfigTargets = notifConfig.queueConfig.map(t => t.queueArn.split(':')[5]);
```

lib/api/apiUtils/bucket/getNotificationConfiguration.js
```diff
@@ -0,0 +1,33 @@
...
+
+function getNotificationConfiguration(parsedXml) {
+    const targets = [];
+    if (!config.bucketNotificationDestinations) {
+        return { error: errors.InvalidArgument.customizeDescription(
+            'Unable to validate the following destination configurations') };
+    }
+    config.bucketNotificationDestinations.forEach(t => {
+        targets.push(t.resource);
+    });
```
miniscruff: Another for each and push.
```suggestion
    if (!config.bucketNotificationDestinations) {
        return { error: errors.InvalidArgument.customizeDescription(
            'Unable to validate the following destination configurations') };
    }
    const targets = config.bucketNotificationDestinations.map(t => t.resource);
```

lib/api/apiUtils/bucket/getNotificationConfiguration.js
```diff
@@ -0,0 +1,33 @@
...
+    const notifConfig = new NotificationConfiguration(parsedXml).getValidatedNotificationConfiguration();
+    if (notifConfig.error) {
+        return notifConfig;
+    }
```
miniscruff: I would move this to the top as its basically a guard clause.

lib/api/apiUtils/bucket/getNotificationConfiguration.js
miniscruff: Alternatively, targets can be a set to make the lookup below O(1) instead of O(n).

```suggestion
    if (!config.bucketNotificationDestinations) {
        return { error: errors.InvalidArgument.customizeDescription(
            'Unable to validate the following destination configurations') };
    }
    const targets = new Set(config.bucketNotificationDestinations.map(t => t.resource));
```

lib/api/apiUtils/bucket/getNotificationConfiguration.js
```diff
@@ -0,0 +1,33 @@
...
+    notifConfig.queueConfig.forEach(t => {
+        const arnArray = t.queueArn.split(':');
+        notifConfigTargets.push(arnArray[5]);
+    });
+    notifConfigTargets.forEach(t => {
+        if (!targets.includes(t)) {
+            notifConfig.error = errors.MalformedXML.customizeDescription(
+                'queue arn target is not included in Cloudserver bucket notification config');
+        }
+    });
```
miniscruff: for each and includes could be every or some depending on the condition.
This assumes using a set above.

```suggestion
    if (!notifConfigTargets.every(t => targets.has(t))) {
        return { error: errors.MalformedXML.customizeDescription(
            'queue arn target is not included in Cloudserver bucket notification config')
        };
    }
```

lib/api/bucketPutNotification.js
```diff
@@ -0,0 +1,67 @@
...
+                    return next(notificationConfig.error);
+                }
+                return next(null, notificationConfig);
+            });
+        },
+        (notifConfig, next) => metadataValidateBucket(metadataValParams, log, (err, bucket) => {
+            if (err) {
+                return next(err, bucket);
+            }
+            return next(null, bucket, notifConfig);
```
miniscruff: I think I mentioned this somewhere else but I do not think the if check is required as undefined/null error will work the same.
```suggestion
            return next(err, bucket, notifConfig);
```

miniscruff (changes requested):

lib/api/bucketPutNotification.js
```diff
@@ -0,0 +1,63 @@
...
+
+    return async.waterfall([
+        next => parseXML(request.post, log, next),
+        (parsedXml, next) => {
+            const notificationConfig = getNotificationConfiguration(parsedXml);
+            process.nextTick(() => {
+                if (notificationConfig.error) {
+                    return next(notificationConfig.error);
+                }
+                return next(null, notificationConfig);
```
miniscruff: Can remove the if check
```suggestion
                return next(notificationConfig.error, notificationConfig);
```

tests/functional/aws-node-sdk/test/bucket/putBucketNotification.js
```diff
@@ -0,0 +1,111 @@
...
+describe('aws-sdk test put notification configuration', () => {
+    let s3;
+    let otherAccountS3;
+
+    before(done => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+        otherAccountS3 = new BucketUtility('lisa', {}).s3;
+        return done();
+    });
```
miniscruff: Shouldn't need `done` if not using callbacks.
```suggestion
    before(() => {
        const config = getConfig('default', { signatureVersion: 'v4' });
        s3 = new S3(config);
        otherAccountS3 = new BucketUtility('lisa', {}).s3;
    });
```

tests/unit/api/bucketPutNotification.js
```diff
@@ -0,0 +1,70 @@
...
+const putNotifConfigRequest = {
+    bucketName,
+    headers: {
+        host: `${bucketName}.s3.amazonaws.com`,
+    },
+    post: notifXml,
+};
+
+describe('putBucketNotification API', () => {
+    before(() => cleanup());
```
miniscruff:
```suggestion
    before(cleanup);
```

tests/unit/api/bucketPutNotification.js
```diff
@@ -0,0 +1,70 @@
...
+    beforeEach(done => bucketPut(authInfo, bucketPutRequest, log, done));
+    afterEach(() => cleanup());
```
miniscruff:
```suggestion
    afterEach(cleanup);
```

tests/unit/api/bucketPutNotification.js
```diff
@@ -0,0 +1,70 @@
...
+
+    it('should update a bucket\'s metadata with bucket notification obj',
```
miniscruff, line 52: If you want to use a quote, then use double quotes to avoid escaping. Or can use backticks.

```suggestion
    it("should update a bucket's metadata with bucket notification obj",
```

tests/unit/api/bucketPutNotification.js
```diff
@@ -0,0 +1,70 @@
...
+    done => {
+        bucketPutNotification(authInfo, putNotifConfigRequest, log, err => {
+            if (err) {
+                process.stdout.write(`Err putting bucket notification ${err}`);
```
miniscruff: was this meant to stay?

tests/unit/api/bucketPutNotification.js
```diff
@@ -0,0 +1,70 @@
...
+                return done(err);
```
miniscruff: does `done(err)` do an assertion? Usually I see `assert.ifError(err)`

miniscruff (changes requested):

miniscruff (approved):

dora-korpar:
@bert-e approve

PR #2964, opened 2020-08-24
bf(S3C-1232): Validate passed account ids for list_metrics.js

tmacro:
Allow only valid account ids to be passed

### DIFF
--- a/lib/utapi/utilities.js
+++ b/lib/utapi/utilities.js
@@ -156,6 +156,19 @@ function listMetrics(metricType) {
         }
     });
 
+    // The string `commander[metric]` is a comma-separated list of resources
+    // given by the user.
+    const resources = commander[metric].split(',');
+
+    // Validate passed accounts to remove any canonicalIDs
+    if (metric === 'accounts') {
+        const invalid = resources.filter(r => !/^\d{12}$/.test(r));
+        if (invalid.length > 0) {
+            logger.error(`Invalid account ID: ${invalid.join(', ')}`);
+            process.exit(1);
+        }
+    }
+
     const timeRange = [];
     // If recent listing, we disregard any start or end option given
     if (!recent) {
@@ -178,9 +191,7 @@ function listMetrics(metricType) {
             timeRange.push(numEnd);
         }
     }
-    // The string `commander[metric]` is a comma-separated list of resources
-    // given by the user.
-    const resources = commander[metric].split(',');
+
     _listMetrics(host, port, metric, resources, timeRange, accessKey, secretKey,
         verbose, recent, ssl);
 }


dora-korpar (approved):

lib/utapi/utilities.js
```diff
@@ -156,6 +156,20 @@ function listMetrics(metricType) {
...
+    const resources = commander[metric].split(',');
+
+    // Validate passed accounts to remove any canonicalids
+    if (metric === 'accounts') {
+        resources.forEach(r => {
+            if (!/^\d{12}$/.test(r)) {
+                logger.error(`${r} is not a valid account id`);
+                process.exit(1);
+            }
+        });
```
miniscruff: If instead of using forEach you could use filter and error on any non empty array. Could report all invalid account id's instead of just one at a time.
Something like:
```suggestion
        const badAccounts = resources.filter(r => !/^\d{12}$/.test(r));
        if (badAccounts) {
            logger.error(`${badAccounts} are not a valid account ids`);
            process.exit(1);
        }
```

I tried to do `resources.filter(/d/.test)` directly but doesn't seem like that is allowed.

miniscruff: I find that returning all possible concurrent errors better than one at a time when possible, but maybe personal preference.

lib/utapi/utilities.js
```diff
@@ -156,6 +156,19 @@ function listMetrics(metricType) {
...
 
+    // The string `commander[metric]` is a comma-separated list of resources
+    // given by the user.
+    const resources = commander[metric].split(',');
+
+    // Validate passed accounts to remove any canonicalIDs
+    if (metric === 'accounts') {
+        const invalid = resources.filter(r => !/^\d{12}$/.test(r));
+        if (invalid.length > 0) {
+            logger.error(`Invalid account ID: ${invalid.join(', ')}`);
```
miniscruff, line 167: ooh yeah, using .join here is much better than my suggestion lol.

lib/utapi/utilities.js
```diff
@@ -156,6 +156,19 @@ function listMetrics(metricType) {
         }
     });
 
+    // The string `commander[metric]` is a comma-separated list of resources
+    // given by the user.
+    const resources = commander[metric].split(',');
+
+    // Validate passed accounts to remove any canonicalIDs
+    if (metric === 'accounts') {
+        const invalid = resources.filter(r => !/^\d{12}$/.test(r));
```
miniscruff, line 165: I know you want to invert the filter but isn't it lame you can't pass in `/regex/.test` directly to filter?

miniscruff (approved):

lib/utapi/utilities.js
tmacro: Yeah coming from python the prototype nature and how js handles namespaces is a super pain

PR #2984, opened 2020-08-31
bf: S3C-3301 bucket notif config fix

naren-rajendran:
Bucket notification configuration properties 'port' and 'auth' are optional and should not stop cloudserver to start. Corrected bucket notification configuration checks to allow optional 'port' and 'auth' properties. Necessary tests are updated.

### DIFF
--- a/config.json
+++ b/config.json
@@ -82,9 +82,7 @@
         {
             "resource": "target1",
             "type": "dummy",
-            "host": "localhost",
-            "port": 6000,
-            "auth": { "user": "user", "password": "password" }
+            "host": "localhost:6000"
         }
     ]
 }
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -291,30 +291,13 @@ function bucketNotifAssert(bucketNotifConfig) {
             'bad config: bucket notification configuration type must be a string');
         assert(typeof host === 'string' && host !== '',
             'bad config: hostname must be a non-empty string');
-        assert(Number.isInteger(port) && port > 0,
-            'bad config: port must be a positive integer');
-        assert(typeof auth === 'object',
-            'bad config: bucket notification auth must be an object');
-        const { cert, user, password } = auth;
-        if ((cert && (user || password)) || (!cert && !user && !password)) {
-            throw new Error('bad config: bucket notification configuration ' +
-                'auth should contain either cert or user and password');
+        if (port) {
+            assert(Number.isInteger(port, 10) && port > 0,
+                'bad config: port must be a positive integer');
         }
-        if (cert) {
-            const certpath = (cert[0] === '/') ? cert : `${this._basePath}/${cert}`;
-            assert.doesNotThrow(() =>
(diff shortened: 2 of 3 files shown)


config.json
```diff
@@ -80,11 +80,11 @@
...
         {
-            "resource": "target1",
-            "type": "dummy",
+            "resource": "scality-test-destination",
+            "type": "kafka",
             "host": "localhost",
-            "port": 6000,
-            "auth": { "user": "user", "password": "password" }
+            "port": "6000",
+            "auth": null
```
dora-korpar: if there's no auth, i'd just remove the key altogether

config.json
```diff
@@ -80,11 +80,11 @@
...
     "bucketNotificationDestinations": [
         {
-            "resource": "target1",
-            "type": "dummy",
+            "resource": "scality-test-destination",
+            "type": "kafka",
             "host": "localhost",
-            "port": 6000,
-            "auth": { "user": "user", "password": "password" }
+            "port": "6000",
```
dora-korpar: throughout the rest of the config, we have port values as ints, not strings, i feel like it would be better to keep that consistent

lib/Config.js
```diff
@@ -291,30 +291,13 @@ function bucketNotifAssert(bucketNotifConfig) {
...
             'bad config: hostname must be a non-empty string');
-        assert(Number.isInteger(port) && port > 0,
-            'bad config: port must be a positive integer');
-        assert(typeof auth === 'object',
-            'bad config: bucket notification auth must be an object');
-        const { cert, user, password } = auth;
-        if ((cert && (user || password)) || (!cert && !user && !password)) {
-            throw new Error('bad config: bucket notification configuration ' +
-                'auth should contain either cert or user and password');
+        if (port) {
```
dora-korpar, line 294: a default port still need to be set, even if it's not included in the config right?

jonathan-gramain (approved):

lib/Config.js
naren-rajendran: Currently we support only kafka as a destination type and the user can choose to provide 'host:port' or just 'some-host-value-without-port' under `host` and leave `port` empty.

config.json
naren-rajendran: Sure, will change and also update the examples in federation to keep it integer.

lib/Config.js
dora-korpar: if that's the case, I think there should be a check in Config.js to make sure that if a `port` key isn't included, that `host` actually contains both host and port

naren-rajendran: I should have explained it better :) . `port` is not mandatory at all, the `host` property can be supplied with just a host name or host name with port or comma separated list of hosts/hosts-with-port.

dora-korpar (approved):

naren-scality:
@bert-e approve

naren-rajendran:
@bert-e approve

PR #3005, opened 2020-09-12
bugfix: S3C-3042 prevent suspending version on replication buckets

rahulreddy:
This fix which is compatibility with AWS S3, ensures that versioning cannot be suspended on buckets that have replication enabled

### DIFF
--- a/lib/api/bucketPutVersioning.js
+++ b/lib/api/bucketPutVersioning.js
@@ -12,7 +12,9 @@ const { config } = require('../Config');
 
 const externalVersioningErrorMessage = 'We do not currently support putting ' +
 'a versioned object to a location-constraint of type Azure.';
-
+const invalidBucketStateMessage = 'A replication configuration is present on ' +
+ 'this bucket, so you cannot change the versioning state. To change the ' +
+ 'versioning state, first delete the replication configuration.';
 /**
  * Format of xml request:
 
@@ -116,6 +118,18 @@ function bucketPutVersioning(authInfo, request, log, callback) {
             return next(null, bucket, versioningConfiguration);
         }),
         (bucket, versioningConfiguration, next) => {
+            // check if replication is enabled if versioning is being suspended
+            const replicationConfig = bucket.getReplicationConfiguration();
+            const invalidAction =
+                versioningConfiguration.Status === 'Suspended'
+                && replicationConfig
+                && replicationConfig.rules
+                && replicationConfig.rules.some(r => r.enabled);
+            if (invalidAction) {
+                next(errors.InvalidBucketState
+                    .customizeDescription(invalidBucketStateMessage));
+                return;
+            }
             bucket.setVersioningConfiguration(versioningConfiguration);
             // TODO all metadata updates of bucket should be using CAS
             metadata.updateBucket(bucket.getName(), bucket, log, err =>
--- a/tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
+++ b/tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
@@ -1,0 +1,80 @@
+const assert = require('assert');
+const async = require('async');
+
+const withV4 = require('../support/withV4');
+const BucketUtility = require('../../lib/utility/bucket-util');
+
+const bucketName = `versioning-bucket-${Date.now()}`;
+
+
+function checkError(err, code) {
+    assert.notEqual(err, null, 'Expected failure but got success');
+    assert.strictEqual(err.code, code);
+}
+
+function checkNoError(err) {
+    assert.ifError(err, `Expected success, got error ${JSON.stringify(err)}`);
+}
+
+function testVersioning(s3, versioningStatus, replicationStatus, cb) {
+    const versioningParams = { Bucket: bucketName,
+        VersioningConfiguration: { Status: versioningStatus } };
+    const replicationParams = {
+        Bucket: bucketName,
+        ReplicationConfiguration: {
+            Role: 'arn:aws:iam::123456789012:role/examplerole,' +
+            'arn:aws:iam::123456789012:role/examplerole',
+            Rules: [
+                {
+                    Destination: {
+                        Bucket: 'arn:aws:s3:::destinationbucket',
+                        StorageClass: 'STANDARD',
+                    },
+                    Prefix: '',
+                    Status: replicationStatus,
+                },
+            ],
+        },
+    };
+    async.waterfall([
+        cb => s3.putBucketReplication(replicationParams, e => cb(e)),
+        cb => s3.putBucketVersioning(versioningParams, e => cb(e)),
+    ], cb);
+}
+
+describe('Versioning on a replication source bucket', () => {
+    withV4(sigCfg => {
+        const bucketUtil = new BucketUtility('default', sigCfg);
+        const s3 = bucketUtil.s3;
+
+        beforeEach(done => {
+            async.waterfall([
+                cb => s3.createBucket({ Bucket: bucketName }, e => cb(e)),
+                cb => s3.putBucketVersioningPromise({
+                    Bucket: bucketName,
+                    VersioningConfiguration: {
+                        Status: 'Enabled',
+                    },
+                }, err => cb(err)),
+            ], done);
+        });
+
+        afterEach(done => s3.deleteBucket({ Bucket: bucketName }, done));
+
+        it('should not be able to disable versioning if replication enabled',
+        done => {
+            testVersioning(s3, 'Suspended', 'Enabled', err => {
+                checkError(err, 'InvalidBucketState');
+                done();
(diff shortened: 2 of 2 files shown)


rachedbenmustapha (approved):

lib/api/bucketPutVersioning.js
```diff
@@ -12,7 +12,9 @@ const { config } = require('../Config');
 
 const externalVersioningErrorMessage = 'We do not currently support putting ' +
 'a versioned object to a location-constraint of type Azure.';
-
+const invalidBucketStateMessage = `A replication configuration is present on
+ this bucket, so you cannot change the versioning state. To change the
+ versioning state, first delete the replication configuration.`;
```
miniscruff: Newlines are maintained in backtick strings, might want to use string concats like the variable above.

lib/api/bucketPutVersioning.js
```diff
@@ -116,10 +118,22 @@ function bucketPutVersioning(authInfo, request, log, callback) {
...
+                replicationEnabled) {
+                next(errors.InvalidBucketState
+                    .customizeDescription(invalidBucketStateMessage));
+                return;
+            }
             bucket.setVersioningConfiguration(versioningConfiguration);
             // TODO all metadata updates of bucket should be using CAS
             metadata.updateBucket(bucket.getName(), bucket, log, err =>
                 next(err, bucket));
+            return;
```
miniscruff: not necessary as it is the last line of the function anyway

lib/api/bucketPutVersioning.js
```diff
@@ -116,10 +118,22 @@ function bucketPutVersioning(authInfo, request, log, callback) {
...
+            const replicationConfig = bucket.getReplicationConfiguration();
+            const replicationEnabled = replicationConfig
+                && replicationConfig.rules
+                && replicationConfig.rules.some(r => r.enabled);
+            if (versioningConfiguration.Status === 'Suspended' &&
+                replicationEnabled) {
+                next(errors.InvalidBucketState
+                    .customizeDescription(invalidBucketStateMessage));
+                return;
+            }
```
miniscruff: I would just merge all these checks into one variable or one if condition instead of doing 3 parts in a variable and then another one in the if condition. Considering all of the replication config work is irrelevant if versioning configuration status is not suspended as it would early evaluate in the if block.

```suggestion
            const replicationConfig = bucket.getReplicationConfiguration();
            if (versioningConfiguration.Status === 'Suspended' &&
                replicationConfig &&
                replicationConfig.rules &&
                replicationConfig.rules.some(r => r.enabled)) {
                next(errors.InvalidBucketState
                    .customizeDescription(invalidBucketStateMessage));
                return;
            }
```

Can't wait for optional chaining though...
```javascript
if (versioningConfiguration.Status === 'Suspended' &&
    bucket.getReplicationConfiguration()?.rules?.some(r => r.enabled)) {
```

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
```diff
@@ -0,0 +1,80 @@
...
+const bucketName = `versioning-bucket-${Date.now()}`;
+
+
+function checkError(err, code) {
+    assert.notEqual(err, null, 'Expected failure but got success');
+    assert.strictEqual(err.code, code);
+}
+
+function checkNoError(err) {
+    assert.ifError(err, `Expected success, got error ${JSON.stringify(err)}`);
```
miniscruff, line 16: This method is basically just a wrapper around ifError that stringifies the error, is that all you wanted to do? The default of ifError is to print the error anyway just not stringified I guess.

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
```diff
@@ -0,0 +1,80 @@
...
+                    },
+                    Prefix: '',
+                    Status: replicationStatus,
+                },
+            ],
+        },
+    };
+    async.waterfall([
+        cb => s3.putBucketReplication(replicationParams, e => cb(e)),
+        cb => s3.putBucketVersioning(versioningParams, e => cb(e)),
```
miniscruff, line 41: Might not need the extra wrapper method on the callbacks, sometimes you do with dealing with `this`...

```suggestion
        cb => s3.putBucketReplication(replicationParams, cb),
        cb => s3.putBucketVersioning(versioningParams, cb),
```

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
```diff
@@ -0,0 +1,80 @@
...
+                done();
+            });
+        });
+
+        it('should be able to disable versioning if replication disabled',
+        done => {
+            testVersioning(s3, 'Suspended', 'Disabled', err => {
+                checkNoError(err);
+                done();
+            });
```
miniscruff, line 77: You can just pass in done here, it will automatically fail if done is passed something that is not null or undefined. Also should apply to the other test.

```suggestion
            testVersioning(s3, 'Suspended', 'Disabled', done);
```

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
```diff
@@ -0,0 +1,80 @@
...
+
+const withV4 = require('../support/withV4');
+const BucketUtility = require('../../lib/utility/bucket-util');
+
+const bucketName = `versioning-bucket-${Date.now()}`;
+
+
+function checkError(err, code) {
+    assert.notEqual(err, null, 'Expected failure but got success');
+    assert.strictEqual(err.code, code);
```
miniscruff, line 12: Small thing that can be ignored, but the second test will fail if the first one fails as well. As if it was equal null would not have a code attribute. So you could remove the first assert and still cover all the points. But sometimes the error message might be a bit harder to debug. As it would just say "null does not have an attribute code" instead of "code does not match: code".

miniscruff (commented):
Totally skippable comments

lib/api/bucketPutVersioning.js
rahulreddy: True, I don't know what I was thinking

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
rahulreddy: It's exactly why it is. Even though the second test will fail if the first one fails, when the test is run in isolation it is helpful.

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
rahulreddy: I can't remember exactly when but it was to make the errors readable by a tool, `istanbul` may be? This is the reason you see this pattern across the test suite. It's worth revisiting though if it's a moot point now.

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
rahulreddy: Ah this, it's needed as this is a waterfall and I would have to feed both result and response to the next step

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
rahulreddy: This was done for readability, again you will see this across the test suite as mocha was not good at spitting out what went wrong. In the previous one we are testing for the error, so if we pass the error to done the test will fail, while we want it to pass when there is an error.

lib/api/bucketPutVersioning.js
rahulreddy: true

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
miniscruff: ok good to know

lib/api/bucketPutVersioning.js
rahulreddy: Thanks, I collapsed it into a single conditional. Once we upgrade to Node 14 we will have optional chaining.

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
miniscruff: Will resolve but maybe take a not of it somewhere else to come back to

tests/functional/aws-node-sdk/test/versioning/replicationBucket.js
miniscruff: oh whoops read that backwards, ignore this one lol

miniscruff (approved):

miniscruff:
The build is broken, if any significant changes are made to fix it you can request another review from me. But usually, it's just a flaky test or a typo.

dora-korpar (approved):

PR #3014, opened 2020-09-14
bugfix: S3C-2775 update logic for counting object replicas

jonathan-gramain:
Before, we were relying on the presence of ACL change to know if the
operation is metadata-only. This was not correct, as a new object can
be put with specific ACL too.

Instead, consider that if there is already an object present with the
same version, we are necessarily updating this object hence the object
counter should not be updated.

NOTE: this change will come along with another change in backbeat: the
PUT /_/backbeat/metadata route will be provided the version ID of the
object to PUT, so that the validation of the existing object (through
the metadataValidateBucketAndObj helper) checks the specific version
instead of checking the master key (which for backbeat replication
purpose is also incorrect). Without this backbeat change, the
replicated object count metric becomes wrong when an object is
rewritten.

# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/routes/utilities/pushReplicationMetric.js
+++ b/lib/routes/utilities/pushReplicationMetric.js
@@ -1,5 +1,3 @@
-const assert = require('assert');
-
 const { ObjectMD } = require('arsenal').models;
 const { pushMetric } = require('../../utapi/utilities');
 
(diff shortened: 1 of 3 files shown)


lib/routes/utilities/pushReplicationMetric.js
```diff
@@ -8,18 +6,15 @@ function shouldPushMetric(prevObjectMD, newObjectMD) {
...
-        assert.deepStrictEqual(prevObjectMD.getAcl(), newObjectMD.getAcl());
-        assert.deepStrictEqual(prevObjectMD.getTags(), newObjectMD.getTags());
-    } catch (e) {
-        return false;
-    }
-    return true;
+    // The rule for a REPLICA is simple: push object metrics if it is
+    // a new object version i.e. there is no existing object of the
+    // same version, which happens when updating ACLs and/or tags
+    // in-place, in which case we don't want to update object metrics.
```
miniscruff: I read an article before I joined scality about writing docs and what you might want to not do. ( Will try and find it as well ). One of the things it said was not to use were phrases like "simple" or "its easy" as it makes those who do not understand it feel worse. Which could lead to lots of things, imposter syndrome is a common one. I know this is sort of inline code docs but I think it still applies.

miniscruff: This might be the article I am referring to... https://www.knowledgeowl.com/home/dont-say-simply-jim-fisher

jonathan-gramain: Ha, good one, I can change this easily :slightly_smiling_face:

jonathan-gramain: I removed "simple" and rephrased a bit the comment.

tmacro (approved):

miniscruff (approved):

anurag4DSB (approved):

dora-korpar (approved):

PR #3034, opened 2020-09-17
Feature/s3 c 3177 policy condition keys

dora-korpar:
Necessary changes for evaluation of IAM policies containing tag condition keys. An explanation of the new workflow can be found in this PR: https://github.com/scality/Arsenal/pull/1265

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -60,6 +60,7 @@ const websiteHead = require('./websiteHead');
 const writeContinue = require('../utilities/writeContinue');
 const validateQueryAndHeaders = require('../utilities/validateQueryAndHeaders');
 const parseCopySource = require('./apiUtils/object/parseCopySource');
+const tagConditionKeyAuth = require('./apiUtils/authorization/tagConditionKeys');
 
 auth.setHandler(vault);
 
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -124,10 +124,7 @@ const api = {
             }
             // issue 100 Continue to the client
             writeContinue(request, response);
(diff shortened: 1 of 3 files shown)


lib/api/api.js
```diff
@@ -100,34 +101,35 @@ const api = {
                 log.trace('authentication error', { error: err });
                 return callback(err);
             }
-            if (authorizationResults) {
+            function ifAuthResults(authResults) {
```
anurag4DSB, line 104: Just curious, what is this way of declaring a function?
Like are we declaring it inside the variable api?
Never seen this trick. How does this work?

lib/api/objectPutTagging.js
```diff
@@ -46,6 +46,8 @@ function objectPutTagging(authInfo, request, log, callback) {
         versionId: reqVersionId,
     };
 
+    const util = require('util');
+    console.log(`\n\n------- REQUEST POST IN PUT OBJ TAGGING::::::::; ${util.inspect(request.post, false, null)}`);
```
anurag4DSB, line 50: I know this PR is still just review only, but just may be

lib/api/objectPutTagging.js
```diff
@@ -75,6 +77,8 @@ function objectPutTagging(authInfo, request, log, callback) {
         },
         (bucket, tags, objectMD, next) => {
             // eslint-disable-next-line no-param-reassign
+            const util = require('util');
+            console.log(`\n\n----------PARSED TAG XML::::: ${util.inspect(tags, false, null)}\n\n`);
```
anurag4DSB, line 81: same here.

PR #3127, opened 2020-10-31
Bugfix/s3 c 2687 acl invalid

dora-korpar:
Tightens up our canonical id regex evaluation.

### DIFF
--- a/lib/utilities/aclUtils.js
+++ b/lib/utilities/aclUtils.js
@@ -188,7 +188,7 @@ aclUtils.parseGrant = function parseGrant(grantHeader, grantType) {
 };
 
 aclUtils.isValidCanonicalId = function isValidCanonicalId(canonicalID) {
-    return /^[A-Za-z0-9]{64}$/.test(canonicalID);
+    return /^(?=.*?[a-f])(?=.*?[0-9])[a-f0-9]{64}$/.test(canonicalID);
 };
 
 aclUtils.reconstructUsersIdentifiedByEmail =
--- a/tests/unit/api/bucketPutACL.js
+++ b/tests/unit/api/bucketPutACL.js
@@ -24,6 +24,18 @@ const canonicalIDforSample1 =
 const canonicalIDforSample2 =
     '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2bf';
 
+const invalidIds = {
+    'too short': 'id="invalid_id"',
+    'too long':
+        'id=79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2aaa',
+    'only numbers':
+        'id=0000000000000000000000000000000000000000000000000000000000000000',
+    'only letters':
+        'id=abcdefabcdefabcdefabcdefabcdefacbdefabcdefabcdefabcdefabcdefabcd',
+    'non-hex letters':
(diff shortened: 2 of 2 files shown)


tests/unit/api/bucketPutACL.js
```diff
@@ -247,24 +259,23 @@ describe('putBucketACL API', () => {
...
-                'x-amz-grant-full-control': invalidCanonicalID,
-            },
-            url: '/?acl',
-            query: { acl: '' },
-        };
-
-        return bucketPutACL(authInfo, testACLRequest, log, err => {
-            assert.deepStrictEqual(err, errors.InvalidArgument);
-            done();
+    Object.keys(invalidIds).forEach(idType => {
```
miniscruff, line 262: Could use Object.entries here

tests/unit/api/bucketPutACL.js
```diff
@@ -24,6 +24,18 @@ const canonicalIDforSample1 =
 const canonicalIDforSample2 =
     '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2bf';
 
+const invalidIds = {
```
miniscruff, line 27: It seems to be the standard to place these sorts of test arrays/objects at the top of the file which by all means be consistent, but I would say when reading the test it is very convenient having them together. 200 lines is more than most monitors so it would require scrolling to see or update the test values. 

My usual method is to just use the object directly like...
```javascript
for (const [testName, testValue] of Object.entries({
  'sample test': 'blah',
  'another test': 'blah 2',
}) {
    it(`should do a thing ${testName}`, () => assert(testValue))
}
```

Just an idea to keep tests as close together as possible so scrolling around isn't required to look at the tests.

lib/utilities/aclUtils.js
```diff
@@ -188,7 +188,7 @@ aclUtils.parseGrant = function parseGrant(grantHeader, grantType) {
 };
 
 aclUtils.isValidCanonicalId = function isValidCanonicalId(canonicalID) {
-    return /^[A-Za-z0-9]{64}$/.test(canonicalID);
+    return /^(?=.*?[a-f])(?=.*?[0-9])[a-f0-9]{64}$/.test(canonicalID);
```
miniscruff, line 191: 0-9 could just be \d ( for digit ) although I have seen some prefer 0-9.
I kinda wish there was \h for hex but doesn't seem like it.

```suggestion
    return /^(?=.*?[a-f])(?=.*?\d)[a-f\d]{64}$/.test(canonicalID);
```

miniscruff (approved):

lib/utilities/aclUtils.js
dora-korpar: yeah, i found a couple of options for hex characters, but they didn't do exactly what i needed. good to know about the `/d`, but i'm gonna keep the `[0-9]` cos i think it's a clearer at a quick glance

dora-korpar:
@bert-e reset

naren-rajendran (approved):

lib/utilities/aclUtils.js
miniscruff: yea \d isn't super clear on its own, it goes well if you use all the shorthands like \w and \s.

dora-korpar:
@bert-e approve

PR #3182, opened 2020-12-02
improvement: S3C-3653 add server ip, port to log

rahulreddy:
These fields are added to support transformation to LEEF log format
for S3C. It's redundant to the fields address and port but they are not
removed to maintain backwards compatibility.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -124,8 +124,9 @@ class S3Server {
                 address: ipAddress || '[::]',
                 port,
             };
-            logger.info('server started', { address: addr.address,
-                port: addr.port, pid: process.pid });
+            const { address } = addr;
+            logger.info('server started', { address, port,
+                pid: process.pid, serverIP: address, serverPort: port });
         });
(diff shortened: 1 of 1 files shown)


alexanderchan-scality (approved):

lib/server.js
```diff
@@ -124,8 +124,9 @@ class S3Server {
                 address: ipAddress || '[::]',
                 port,
             };
-            logger.info('server started', { address: addr.address,
-                port: addr.port, pid: process.pid });
+            const { address } = addr;
```
ilkescality, line 127: not sure if I'm missing anything but shouldn't `port` be added in here as well? (in order to support being referred directly as `port` instead of `addr.port` on the below lines)
```suggestion
            const { address, port } = addr;
```

ilkescality: nvm, just read again.

ilkescality (approved):

dora-korpar (approved):

PR #3191, opened 2020-12-07
Bugfix/s3 c 3313 fix storageused

dora-korpar:
Fix `storageUtilized` Utapi metric when bucket is deleted with ongoing MPUs.

### DIFF
--- a/lib/api/apiUtils/bucket/bucketDeletion.js
+++ b/lib/api/apiUtils/bucket/bucketDeletion.js
@@ -3,6 +3,7 @@ const async = require('async');
 const { errors } = require('arsenal');
 
 const abortMultipartUpload = require('../object/abortMultipartUpload');
+const { pushMetric } = require('../../../utapi/utilities');
 
 const { splitter, oldSplitter, mpuBucketPrefix } =
     require('../../../../constants');
@@ -29,7 +30,15 @@ function _deleteOngoingMPUs(authInfo, bucketName, mpus, log, cb) {
         // `overview${splitter}${objectKey}${splitter}${uploadId}
         const [, objectKey, uploadId] = mpu.key.split(splitterChar);
         abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
-            next);
+            (err, destBucket, partSizeSum) => {
+                pushMetric('abortMultipartUpload', log, {
+                    authInfo,
+                    bucket: bucketName,
+                    keys: [objectKey],
+                    byteLength: partSizeSum,
+                });
+                next(err);
+            });
     }, cb);
 }
 /**


lib/api/bucketDelete.js
```diff
@@ -42,13 +42,14 @@ function bucketDelete(authInfo, request, log, cb) {
...
             return deleteBucket(authInfo, bucketMD, bucketName,
-                authInfo.getCanonicalID(), log, err => {
+                authInfo.getCanonicalID(), log, err, totalPartSizeSum => {
                     if (err) {
                         return cb(err, corsHeaders);
                     }
                     pushMetric('deleteBucket', log, {
                         authInfo,
                         bucket: bucketName,
+                        byteLength: totalPartSizeSum,
```
rahulreddy: This should be negative, otherwise we would be incrementing storage instead of decrementing it.

rahulreddy (commented):
Also check if Utapi.deletebucket takes bytelength into account and sets the metrics

lib/api/bucketDelete.js
tmacro: also v2 only looks at `newByteLength/oldByteLength` to compute `sizeDelta` https://github.com/scality/cloudserver/blob/cd3aac2d8ddbc7ecaa00496b9fc80300873e3fd6/lib/utapi/utilities.js#L262

dora-korpar: That's what I thought too, but looking at the other APIs that should decrement storageUtilized, they pass the number as a positive integer, like here https://github.com/scality/cloudserver/blob/development/8.2/lib/api/multipartDelete.js#L53 (partSizeSum is positive) and here https://github.com/scality/cloudserver/blob/development/8.2/lib/api/multipartDelete.js#L53

dora-korpar: Which looking at the utapi functions for those apis, they use the `decr` command. Delete bucket uses the generic push metric call, so would it be too hacky to pass a negative value to the `incr` command? Otherwise, will require some changes in utapi

tmacro: I just looked at that and it seems like its more a bug in the `sizeDelta` bit then. `multiObjectDelete` passes a `isDelete` flag that could be used for v2's `sizeDelta`
https://github.com/scality/cloudserver/blob/cd3aac2d8ddbc7ecaa00496b9fc80300873e3fd6/lib/api/multiObjectDelete.js#L508

dora-korpar: yeah, I have a feeling that for v1 tho, anything used with the `incrby` command needs to be positive. do you know if that's right?

miniscruff (approved):

ilkescality (approved):

naren-rajendran (approved):

lib/api/bucketDelete.js
rahulreddy: @dora-korpar You are right, I forgot about that part. So the methods issue `incr` or `decr` depending on the api call.

PR #3282, opened 2021-01-25
bugfix: S3C-3860: object put write continue too early

alexandre-merle:
## Description
Continue message was send before checking auths, which could lead
to api calls to send the object data, while cloudserver will
check auth and send back an error, closing the connection.
All this was leading to an EPIPE error from the client
perspective.

## Related issues

Fixes S3C-3860
Fixes S3C-3063

### DIFF
--- a/lib/api/objectPutPart.js
+++ b/lib/api/objectPutPart.js
@@ -16,7 +16,7 @@ const { config } = require('../Config');
 const multipleBackendGateway = require('../data/multipleBackendGateway');
 const locationConstraintCheck
     = require('./apiUtils/object/locationConstraintCheck');
-
+const writeContinue = require('../utilities/writeContinue');
 const skipError = new Error('skip');
 
 // We pad the partNumbers so that the parts will be sorted in numerical order.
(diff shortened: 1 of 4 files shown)


anurag4DSB:
ping

alexandre-merle:
@bert-e reset

alexandre-merle:
ping

alexandre-merle:
ping

anurag4DSB (approved):

lib/api/objectPutPart.js
```diff
@@ -113,6 +113,7 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                 log.debug('access denied for user on bucket', { requestType });
                 return next(errors.AccessDenied, destinationBucket);
             }
+            writeContinue(request, request._response);
```
nicolas2bert: To avoid EPIPE error, should we `writeContinue` right before the data from the response is consumed.
What if, for example, the cipher bundle fails right after we indicate to our client that the request body should be sent? Won't it also lead to a client EPIPE error?

alexandre-merle: it's true, this can happen, still will be trigger in response to an internal error so less problematic, but you are right, i will move the call

naren-rajendran (approved):

lib/api/objectPutPart.js
alexandre-merle: modified :)

nicolas2bert (approved):

alexandre-merle:
@bert-e reset

alexandre-merle:
ping

alexandre-merle:
ping

PR #3302, opened 2021-01-28
ft: S3C-2767 bucket policy not implemented error

dora-korpar:
Makes PutBucketPolicy API return a NotImplemented error if the user attempts to put a condition, federated principal, or service principal in the bucket policy instead of MalformedXML.

Not the absolute most elegant solution, but this way nothing in Arsenal is impacted (which would have required enough work that we could have just implemented conditions x) ), and it will be easy to remove once those keys are implemented.

### DIFF
--- a/lib/api/bucketPutPolicy.js
+++ b/lib/api/bucketPutPolicy.js
@@ -9,6 +9,20 @@ const { validatePolicyResource } =
 const { BucketPolicy } = models;
 
 /**
+ * _checkNotImplementedPolicy - some bucket policy features have not been
+ * implemented and should return NotImplemented error
+ * @param {string} policyString - string bucket policy
+ * @return {boolean} - returns true if policy contains not implemented elements
+ */
+function _checkNotImplementedPolicy(policyString) {
+    // bucket names and key names cannot include "", so including those
+    // isolates not implemented keys
+    return policyString.includes('"Condition"')
+    || policyString.includes('"Service"')
+    || policyString.includes('"Federated"');
+}
+
+/**
  * bucketPutPolicy - create or update a bucket policy
  * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
  * @param {object} request - http request object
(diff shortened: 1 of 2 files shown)


ilkescality (approved):

lib/api/bucketPutPolicy.js
```diff
@@ -8,6 +8,23 @@ const { validatePolicyResource } =
...
+ */
+function _checkNotImplementedPolicy(policyString) {
+    // bucket names and key names cannot include "", so including those
+    // isolates not implemented keys
+    if (policyString.includes('"Condition"')
+    || policyString.includes('"Service"')
+    || policyString.includes('"Federated"')) {
+        return true;
+    }
+    return false;
```
miniscruff: Can just use return

```suggestion
    return policyString.includes('"Condition"')
        || policyString.includes('"Service"')
        || policyString.includes('"Federated"')
```

tests/unit/api/bucketPutPolicy.js
```diff
@@ -73,4 +75,33 @@ describe('putBucketPolicy API', () => {
...
+
+    it('should return error if policy contains conditions', done => {
+        expectedBucketPolicy.Statement[0].Condition =
+            { StringEquals: { 's3:x-amz-acl': ['public-read'] } };
+        bucketPutPolicy(authInfo, getPolicyRequest(expectedBucketPolicy), log,
+        err => {
+            assert.strictEqual(err.NotImplemented, true);
+            return done();
+        });
+    });
```
miniscruff: Seems like these three tests can be parameterized pretty easily

tests/unit/api/bucketPutPolicy.js
```diff
@@ -73,4 +75,33 @@ describe('putBucketPolicy API', () => {
...
         });
     });
+
+    it('should return error if policy contains conditions', done => {
+        expectedBucketPolicy.Statement[0].Condition =
+            { StringEquals: { 's3:x-amz-acl': ['public-read'] } };
+        bucketPutPolicy(authInfo, getPolicyRequest(expectedBucketPolicy), log,
+        err => {
+            assert.strictEqual(err.NotImplemented, true);
+            return done();
```
miniscruff: no need to return here as well

naren-rajendran (approved):

miniscruff (approved):

PR #3312, opened 2021-02-02
bugfix: S3C-3860: fix corner case multiple backend

alexandre-merle:
## Description
in case of multiple backend, we were writing twice
'continue' in the socket, leading to econnreset from the client
side

## Related issues

S3C-3860

### DIFF
--- a/lib/api/objectPutPart.js
+++ b/lib/api/objectPutPart.js
@@ -199,6 +199,12 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                     // eslint-disable-next-line no-param-reassign
                     objectLocationConstraint = backendInfoObj.controllingLC;
                 }
+                if (!multipleBackendGateway.isClientHandleMpu(
+                        objectLocationConstraint)) {
+                    // if data backend doesn't handle MPU, continue waterfall
+                    return next(null, destinationBucket,
+                    objectLocationConstraint, cipherBundle, splitter, null);
+                }
                 writeContinue(request, request._response);
                 return multipleBackendGateway.uploadPart(request,
                 streamingV4Params, null, size, objectLocationConstraint,
@@ -221,9 +227,14 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                             objectLocationConstraint, cipherBundle, splitter,
                             partInfo);
                     }
-                    // if data backend doesn't handle MPU, continue waterfall
-                    return next(null, destinationBucket,
-                    objectLocationConstraint, cipherBundle, splitter, null);
+                    let msg = 'backend is managing MPU but was';
+                    msg += ' not handle after uploadPart';
+                    log.error(msg, {
+                        error: errors.InternalError,
+                        method:
+                        'objectPutPart::multipleBackendGateway.uploadPart',
+                    });
+                    return next(errors.InternalError, destinationBucket);
                 });
             }
             return next(null, destinationBucket, objectLocationConstraint,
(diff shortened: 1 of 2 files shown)


alexandre-merle:
ping

lib/api/objectPutPart.js
```diff
@@ -221,9 +227,14 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                             objectLocationConstraint, cipherBundle, splitter,
                             partInfo);
                     }
-                    // if data backend doesn't handle MPU, continue waterfall
-                    return next(null, destinationBucket,
-                    objectLocationConstraint, cipherBundle, splitter, null);
+                    let msg = 'backend is managing MPU but was';
+                    msg += ' not handle after uploadPart';
```
ilkescality, line 231: could also use constant but it's okay either way
```suggestion
                    const msg = 'backend is managing MPU but was'
                        + ' not handle after uploadPart';
```

ilkescality (approved):

lib/data/multipleBackendGateway.js
```diff
@@ -183,6 +183,14 @@ const multipleBackendGateway = {
         return cb();
     },
 
+    isClientHandleMpu: location => {
+        const client = clients[location];
+        if (client.uploadPart) {
+            return true;
+        }
+        return false;
```
miniscruff, line 191:
```suggestion
        return clients[location].uploadPart;
```

miniscruff (changes requested):
Is there a test to go with the change or is this part of your "fix tests" branch?

lib/data/multipleBackendGateway.js
alexandre-merle: yes, an integration tests was already failing because of this

alexandre-merle: clients[location].uploadPart is a function, we don't want to return it :) we just check to validate if the backend is manage MPU (as uploadPart function in it)

miniscruff:
```suggestion
        return client[location].uploadPart !== undefined;
```
How about checking if it is undefined?

alexandre-merle: we could, but doing the all bert-e reset for a cosmetic change is maybe too much :D

jonathan-gramain (approved):

lib/data/multipleBackendGateway.js
miniscruff: I think this is not purely cosmetic as it is checking explicitly for undefined instead of just falsy. But either way, having bert-e do more work should not dictate what does and doesn't go into PRs. 

I think there is also added value to reducing the line count as it is generally a smoother read.

alexandre-merle: hardening the checks make sense i agree, we could even better put `typeof client[location].uploadPart === 'function'` , but i'm focusing on bugfixes here, i'll prefer to have a ticket to hardened the checks in different parts of cloudserver, not only this place (the point was not really bert-e, but the fact that i need to validate again the fix with all branches after it)

miniscruff: I prefer to handle bugfixes with some hardening when possible, but the extra time to validate does not seem worth it considering the number of fixes you already have. We should push for some hardening time for this type of effort soon.

miniscruff (approved):

lib/data/multipleBackendGateway.js
alexandre-merle: i agree on that, hardening will be the next maintenance priority i think

PR #3392, opened 2021-03-07
bf: S3C-4055 cloudserver metrics computation correction

naren-rajendran:
Metrics like incomingBytes, sizeDelta were not calculated correctly for actions like getObject and it is corrected now.


naren-rajendran:
Wrong base branch.

PR #3394, opened 2021-03-07
bf: S3C-4055 cloudserver metrics computation correction

naren-rajendran:
metrics like incomingBytes, sizeDelta were not calculated correctly for actions like getObject and it is corrected now.


rahulreddy:
@naren-scality Can you please change the base branch to `stabilization/7.9.0`? It needs to be picked up for 7.9.0

tests/utapi/utilities.js
```diff
@@ -0,0 +1,282 @@
...
+
+    testEvents.forEach(event => {
+        it(`should compute and push metrics for ${event.action}`, () => {
+            const eventPushed = pushMetric(event.action, log, event.metrics);
+            assert(eventPushed);
+            const compareObj = computeMetrics(event.action, event.metrics);
+            assert(compareObj);
+            Object.keys(compareObj).forEach(key => {
+                assert.strictEqual(eventPushed[key], compareObj[key]);
+            });
```
miniscruff: I think this can be assert.deepStrictEqual instead.

tests/utapi/utilities.js
```diff
@@ -0,0 +1,282 @@
...
+    const incomingBytes = action === 'getObject' ? 0 : newByteLength;
+    const sizeDelta = oldByteLength ? newByteLength - oldByteLength :
+        incomingBytes;
+    return {
+        objectDelta: isDelete ? -numberOfObjects : numberOfObjects,
+        sizeDelta: isDelete ? -byteLength : sizeDelta,
+        incomingBytes,
+        outgoingBytes: action === 'getObject' ? newByteLength : 0,
+    };
+}
```
miniscruff: Duplicating the logic into the test as the code itself isn't the best method of writing tests. For one it ties our tests to the internal logic which we should not do, we just want to test the observable result (eg the output). Additionally, this just proves that what comes in matches what comes out and not that what comes out is correct. For example, if you wrote the test by copy pasting the broken code, the test would still pass even though it was wrong.

Instead, create `expected` objects to compare to in your test array.

Basically it would be the same as doing: 
```js
const jsonString = '{"a": 5}';
assert.strictEqual(JSON.stringify(JSON.parse(jsonString)), jsonString);
```

miniscruff (changes requested):

tests/utapi/utilities.js
naren-rajendran: I am still working on this. This is not quite up for review yet, will convert this to draft. I have the correction but did not push it yet... too soon @miniscruff :)

naren-rajendran:
@bert-e reset

anurag4DSB (commented):
Do we not have any tests yet?

naren-rajendran:
> Do we not have any tests yet?

@anurag4DSB On cloudserver, no, we do not! We have integration tests for utapi v1 but not for v2.

anurag4DSB (approved):

tmacro (approved):

naren-rajendran:
@bert-e reset

tests/utapi/utilities.js
```diff
@@ -0,0 +1,471 @@
...
+        sinon.restore();
+    });
+
+    testEvents.forEach(event => {
+        it(`should compute and push metrics for ${event.action}`, () => {
+            const eventPushed = pushMetric(event.action, log, event.metrics);
+            assert(eventPushed);
+            Object.keys(event.expected).forEach(key => {
+                assert.deepStrictEqual(eventPushed[key], event.expected[key]);
+            });
```
miniscruff: I think just one assertion will work here.
```suggestion
            assert.deepStrictEqual(eventPushed, event.expected);
```

tests/utilities/mock/Utapi.js
```diff
@@ -0,0 +1,82 @@
+const express = require('express');
```
miniscruff, line 1: This feels like a lot of code for a mock. Feels like we would need to add tests here as it has a fair bit of logic.

naren-rajendran:
@bert-e reset

tests/utapi/awsNodeSdk.js
```diff
@@ -0,0 +1,375 @@
+const async = require('async');
```
miniscruff, line 1: I feel like this isn't wrong but a bit much. Both with the mock UTAPI server with express and recreating a bunch of events to send to UTAPI. Part of my concern here is too much testing as well as potentially duplicate testing.
It seems to me our unit testing and functional testing overlap here as well as what would probably be included in UTAPI v2. 

I don't really want to block the PR to come up with some sort of test plan or Venn Diagram of what is tested where but I don't want to blindly review all this code.

miniscruff (changes requested):

tests/utapi/utilities.js
naren-rajendran: Not all properties of the published event are compared, only metrics.

miniscruff (approved):

naren-rajendran:
@bert-e approve

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e help

naren-rajendran:
@bert-e status

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e help

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e force_reset

rahulreddy:
I am going to break 8.1 and 8.2 branches by merging this to allow 7.9 RC4 to be built. @naren-scality  will follow up with a fix for 8.1 and 8.2 branches.

PR #3507, opened 2021-04-07
bf(S3C-4245): Don't allow COMPLIANCE mode to be shortened or disabled

tmacro:
Add several checks to disallow changing mode or shortening the retention period while in COMPLIANCE mode

### DIFF
--- a/lib/api/apiUtils/object/objectLockHelpers.js
+++ b/lib/api/apiUtils/object/objectLockHelpers.js
@@ -144,14 +144,18 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
-function validateObjectLockUpdate(objectMD, retentionInfo) {
+function validateObjectLockUpdate(objectMD, retentionInfo, bypassGovernance) {
     const { retentionMode: existingMode, retentionDate: existingDateISO } = objectMD;
-    if (!existingMode || existingMode === 'GOVERNANCE') {
+    if (!existingMode) {
         return null;
     }
 
     const existingDate = new Date(existingDateISO);
     const isExpired = existingDate < Date.now();
+
+    if (existingMode === 'GOVERNANCE' && !isExpired && !bypassGovernance) {
+        return errors.AccessDenied;
+    }
 
     if (retentionInfo.mode === 'GOVERNANCE' && !isExpired) {
         return errors.AccessDenied;
--- a/lib/api/apiUtils/object/objectLockHelpers.js
+++ b/lib/api/apiUtils/object/objectLockHelpers.js
@@ -144,10 +144,31 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
+function validateObjectLockUpdate(objectMD, retentionInfo) {
+    const { retentionMode: existingMode, retentionDate: existingDateISO } = objectMD;
+    if (!existingMode || existingMode === 'GOVERNANCE') {
+        return null;
+    }
+
+    const existingDate = new Date(existingDateISO);
+    const isExpired = existingDate < Date.now();
+
+    if (retentionInfo.mode === 'GOVERNANCE' && !isExpired) {
+        return errors.AccessDenied;
+    }
+
+    if (new Date(retentionInfo.date) < existingDate) {
+        return errors.AccessDenied;
+    }
+
+    return null;
+}
+
 module.exports = {
     calculateRetainUntilDate,
     compareObjectLockInformation,
     setObjectLockInformation,
     isObjectLocked,
     validateHeaders,
+    validateObjectLockUpdate,
 };
--- a/lib/api/objectPutRetention.js
+++ b/lib/api/objectPutRetention.js
@@ -3,7 +3,8 @@ const { errors, s3middleware } = require('arsenal');
 
 const { decodeVersionId, getVersionIdResHeader } =
   require('./apiUtils/object/versioning');
-
+const { validateObjectLockUpdate } =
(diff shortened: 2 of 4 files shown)


lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -144,10 +144,25 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
+function validateObjectLockUpdate(objectMD, retentionInfo) {
+    const { retentionMode: existingMode, retentionDate: existingDate } = objectMD;
+    const isExpired = existingDate ? new Date(existingDate) < Date.now() : true;
+    if (existingMode === 'COMPLIANCE') {
+        if (retentionInfo.mode === 'GOVERNANCE' && !isExpired) {
```
jonathan-gramain: good catch, did not think about the case when a COMPLIANCE lock is expired and thus allowed to be set to GOVERNANCE

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -144,10 +144,25 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
+function validateObjectLockUpdate(objectMD, retentionInfo) {
+    const { retentionMode: existingMode, retentionDate: existingDate } = objectMD;
+    const isExpired = existingDate ? new Date(existingDate) < Date.now() : true;
```
jonathan-gramain: I'd rather move this line inside the `if (existingMode === 'COMPLIANCE')` to avoid unnecessary conversion to Date object in other cases. Also, the `new Date(existingDate)` conversion could be done once in a variable to avoid duplicate conversion (likely an insignificant performance hit but easy to change).

jonathan-gramain (approved):

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -144,10 +144,26 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
+function validateObjectLockUpdate(objectMD, retentionInfo) {
+    const { retentionMode: existingMode, retentionDate: existingDate } = objectMD;
+    if (existingMode === 'COMPLIANCE') {
```
miniscruff: I would flip this as a guard clause to reduce indents

lib/api/objectPutRetention.js
```diff
@@ -78,6 +79,13 @@ function objectPutRetention(authInfo, request, log, callback) {
             parseRetentionXml(request.post, log,
                 (err, retentionInfo) => next(err, bucket, retentionInfo, objectMD));
         },
+        (bucket, retentionInfo, objectMD, next) => {
+            const validationError = validateObjectLockUpdate(objectMD, retentionInfo);
+            if (validationError) {
+                return next(validationError, bucket, objectMD);
```
miniscruff, line 86: Would this be more appropriate as just ?
```suggestion
                return next(validationError);
```

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -144,10 +144,26 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
+function validateObjectLockUpdate(objectMD, retentionInfo) {
+    const { retentionMode: existingMode, retentionDate: existingDate } = objectMD;
```
miniscruff: could existing date string be `existingDateISO` so it is clear it is not a Date already?

tests/unit/api/apiUtils/objectLockHelpers.js
```diff
@@ -175,4 +176,96 @@ describe('objectLockHelpers: calculateRetainUntilDate', () => {
         assert.strictEqual(retainUntilDate.slice(0, 16),
             expectedRetainUntilDate.toISOString().slice(0, 16));
     });
+
+    describe('objectLockHelpers: validateObjectLockUpdate', () => {
```
miniscruff, line 180: test suite looks really good :+1:

lib/api/objectPutRetention.js
naren-rajendran: It's `async.waterfall`

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -144,10 +144,31 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
+function validateObjectLockUpdate(objectMD, retentionInfo) {
+    const { retentionMode: existingMode, retentionDate: existingDateISO } = objectMD;
+    if (!existingMode || existingMode === 'GOVERNANCE') {
+        return null;
+    }
+
+    const _existingDate = new Date(existingDateISO);
```
miniscruff: Now that the other date is a separate name you can remove the underscore :smirk:

miniscruff (approved):

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -144,10 +144,31 @@ function isObjectLocked(bucket, objectMD, headers) {
...
+    }
+
+    const existingDate = new Date(existingDateISO);
+    const isExpired = existingDate < Date.now();
+
+    if (retentionInfo.mode === 'GOVERNANCE' && !isExpired) {
+        return errors.AccessDenied;
+    }
+
+    if (new Date(retentionInfo.date) < _existingDate) {
```
jonathan-gramain: `_existingDate` -> `existingDate`

PR #3512, opened 2021-04-09
bf(S3C-4276): Update KMIP test certificates

tmacro:
Regenerate kmip test certificates. Previously we bumped the script to use an expiry of 100 years, but committed certificates expiring in 1. Also adds the rust toolchain to the pykmip image as its now required by the python `cryptography` module.


eve/workers/pykmip/Dockerfile
```diff
@@ -1,7 +1,10 @@
 FROM python:3-alpine
 
+ADD https://sh.rustup.rs /tmp/rustup.sh
```
miniscruff: What is this used for?

tmacro: `cryptography`

miniscruff (approved):

naren-rajendran (approved):

eve/workers/pykmip/Dockerfile
rahulreddy: I think there is a way to do this outside of Dockerfile to leverage the CI proxy cache, we would run into rate limits easily otherwise.
This is how @rdebay-scality did it in Federation PR https://github.com/scality/Federation/pull/3526/files, can't really tell how to do it in eve/main.yaml or elsewhere. @tcarmet may have an idea

rdebay-scality: The proxy-cache is setuped in the env of the CI.

the get_url module I used has an option `use_proxy` set to yes:

`use_proxy | no | yes | yesno | if no, it will not use a proxy, even if one is defined in an environment variable on the target hosts.
`

tcarmet: Did we ran into any rate limiting on this particular image build?

Those images are already cached by the registry, (Eve does some work towards that if it builds images) and if they don't change, they're not build on every single build. (Or at least they shouldn't be). Therefore I don't see the point of adding the proxy here.

tmacro: Turns out I had an old image for `python:3-alpine` (and thus alpine itself).I thought the version in the apk repos was too old, but after updating it works

PR #3530, opened 2021-04-13
bf(S3C-4245): Enforce x-amz-bypass-governance-retention header and s3:BypassGovernanceRetention user policy

tmacro:
Second half of the object lock fix

### DIFF
--- a/lib/api/apiUtils/object/objectLockHelpers.js
+++ b/lib/api/apiUtils/object/objectLockHelpers.js
@@ -144,21 +144,27 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
-function validateObjectLockUpdate(objectMD, retentionInfo) {
+function validateObjectLockUpdate(objectMD, retentionInfo, bypassGovernance) {
     const { retentionMode: existingMode, retentionDate: existingDateISO } = objectMD;
-    if (!existingMode || existingMode === 'GOVERNANCE') {
+    if (!existingMode) {
         return null;
     }
 
     const existingDate = new Date(existingDateISO);
     const isExpired = existingDate < Date.now();
 
-    if (retentionInfo.mode === 'GOVERNANCE' && !isExpired) {
+    if (existingMode === 'GOVERNANCE' && !isExpired && !bypassGovernance) {
         return errors.AccessDenied;
     }
 
-    if (new Date(retentionInfo.date) < existingDate) {
-        return errors.AccessDenied;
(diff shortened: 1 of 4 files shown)


lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -144,20 +144,24 @@ function isObjectLocked(bucket, objectMD, headers) {
...
     const existingDate = new Date(existingDateISO);
     const isExpired = existingDate < Date.now();
 
-    if (retentionInfo.mode === 'GOVERNANCE' && !isExpired) {
+    if (existingMode === 'GOVERNANCE' && !isExpired && !bypassGovernance) {
         return errors.AccessDenied;
     }
 
-    if (new Date(retentionInfo.date) < existingDate) {
+    if (existingMode === 'COMPLIANCE' && retentionInfo.mode === 'GOVERNANCE' && !isExpired) {
```
jonathan-gramain: Just for clarity, could benefit to group compliance checks in a block, or maybe even also grouping governance checks with a switch-like:

```javascript
if (existingMode === 'GOVERNANCE') {
    // governance checks
} else if (existingMode === 'COMPLIANCE') {
    // compliance checks
}
```

lib/api/objectPutRetention.js
```diff
@@ -80,7 +82,51 @@ function objectPutRetention(authInfo, request, log, callback) {
...
+                        generalResource: bucketName,
+                        specificResource: { key: objectKey },
+                        requesterIp: ip,
+                        sslEnabled: request.connection.encrypted,
+                        apiMethod: 'bypassGovernanceRetention',
+                        awsService: 's3',
+                        locationConstraint: bucket.getLocationConstraint(),
+                        requesterInfo: authInfo,
+                        signatureVersion: authParams.params.data.authType,
+                        authType: authParams.params.data.signatureVersion,
```
rachedbenmustapha: Swap here

jonathan-gramain: makes me think that some specific tests to catch this particular typo issue could be useful (not sure what to test exactly, but I'm surprised that Vault would return a success if we mess with those input parameters)

tmacro: After some testing, the value of these params doesn't matter as the values aren't used (though they still need to be present or vault will reject the request with `ValidationError`). The same error also exists in bucketPut and multiObjectDelete.

jonathan-gramain (approved):

rachedbenmustapha (approved):

PR #3541, opened 2021-04-21
ft(S3C-3752): Add GetBucketEncryption handler

tmacro:


lib/api/bucketGetEncryption.js
```diff
@@ -0,0 +1,84 @@
...
+            log.debug('access denied for account on bucket', {
+                requestType,
+                method: 'bucketGetEncryption',
+            });
+            return callback(errors.AccessDenied, null, corsHeaders);
+        }
+
+        const sseInfo = bucket.getServerSideEncryption();
+
+        // If ssInfo is present but the `mandatory` flag is not set
```
jonathan-gramain: if ss**e**Info

lib/api/bucketGetEncryption.js
```diff
@@ -0,0 +1,84 @@
...
+            return callback(errors.ServerSideEncryptionConfigurationNotFoundError);
+        }
+
+        const xml = [];
+        xml.push(
+            '<?xml version="1.0" encoding="UTF-8"?>',
+            '<ServerSideEncryptionConfiguration>',
+            '<Rule>',
+            '<ApplyServerSideEncryptionByDefault>',
+            `<KMSMasterKeyID>${escapeForXml(sseInfo.masterKeyId)}</KMSMasterKeyID>`,
```
jonathan-gramain: There are two different supported schemes:
- the `aws:kms` mode of encryption: one where the PutBucketEncryption contains a specified master key ID to use for the bucket, in which case it should be returned in the KMSMasterKeyID field
- the `AES256` mode where it uses a managed key (a Scality key) and the user should not know about it as it is meant to be transparent, then we should not return the "KMSMasterKeyID" field even though the bucket has an internal managed key that it uses to encrypt objects.

cf. design doc in citadel and https://docs.aws.amazon.com/AmazonS3/latest/API/API_ServerSideEncryptionByDefault.html

tests/functional/aws-node-sdk/test/bucket/getBucketEncryption.js
```diff
@@ -0,0 +1,83 @@
...
+describe('aws-sdk test get bucket encryption', () => {
+    let s3;
+
+    before(done => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+        return done();
+    });
+
+    it('should return NoSuchBucket error if bucket does not exist', done => {
```
jonathan-gramain, line 33: I'd rather put this in a separate test with a dummy bucket name, than in the `before` section

tests/functional/aws-node-sdk/test/bucket/getBucketEncryption.js
```diff
@@ -0,0 +1,83 @@
...
+                assert.ifError(err);
+                s3.getBucketEncryption({Bucket: bucketName}, (err, res) => {
+                    assert.ifError(err);
+                    assert.deepStrictEqual(res, {
+                        ServerSideEncryptionConfiguration: {
+                            Rules: [
+                                {
+                                    ApplyServerSideEncryptionByDefault: {
+                                        SSEAlgorithm: 'AES256',
+                                        KMSMasterKeyID: '12345'
```
jonathan-gramain: As said above, this is invalid because `KMSMasterKeyID` is not allowed with `SSEAlgorithm: 'AES256'` (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ServerSideEncryptionByDefault.html)

lib/api/bucketGetEncryption.js
```diff
@@ -0,0 +1,84 @@
...
+            });
+            return callback(errors.AccessDenied, null, corsHeaders);
+        }
+
+        const sseInfo = bucket.getServerSideEncryption();
+
+        // If ssInfo is present but the `mandatory` flag is not set
+        // then this info was not created using bucketPutEncryption
+        // and should not be returned
+        if (sseInfo === null || !sseInfo.mandatory) {
```
jonathan-gramain, line 57: In the design doc, I suggested we could return the existing "legacy" encryption config via GetBucketEncryption. Not sure if it's a good idea or not but let's discuss it and make sure we agree on what to do here

jonathan-gramain (commented):
First round looks a good start, please review the design doc in citadel and comment/suggest changes/approve so we are aligned on some implementation details.

lib/api/bucketGetEncryption.js
tmacro: As I understand it bucket created using the `x-amz-scal-server-side-encryption` header are in essence the same as a bucket with a default scheme defined using `PutBucketEncryption`

lib/api/bucketGetEncryption.js
```diff
@@ -0,0 +1,84 @@
...
+
+    return metadata.getBucket(bucketName, log, (err, bucket) => {
+        if (err) {
+            log.debug('metadata getbucket failed', { error: err });
+            return callback(err);
+        }
+        if (bucketShield(bucket, requestType)) {
+            return callback(errors.NoSuchBucket);
+        }
+        log.trace('found bucket in metadata');
```
anurag4DSB: Should we add the method here?

lib/api/bucketGetEncryption.js
jonathan-gramain: Yes, those should be compatible regarding metadata, I was thinking of using the `mandatory` flag as a way to know if the policy is applied to the bucket (so either via PutBucketEncryption or via the legacy x-scal header) or if it only contains the default encryption params used to encrypt objects that explicitly contain the `x-amx-server-side-encryption` header, i.e. the master key ID and cryptoScheme, since ideally a bucket always uses the same master key by default (except if specified explicitly). I cannot tell for sure what was the original intent of `mandatory`, but I think it is somehow in the same spirit, except that without the backward-compat thing I'd rather call it `applyByDefault` but we have to keep the name. This way we can also support object-level encryption with minimal changes to the metadata format. I can change this if it does not fit well.

anurag4DSB (approved):
code looks good. I am reviewing the design PR. I Will report back if I find anything odd.

lib/api/bucketGetEncryption.js
```diff
@@ -0,0 +1,93 @@
...
+                requestType,
+                method: 'bucketGetEncryption',
+            });
+            return callback(errors.AccessDenied, null, corsHeaders);
+        }
+
+        const sseInfo = bucket.getServerSideEncryption();
+
+        // If sseInfo is present but the `mandatory` flag is not set
+        // then this info was not created using bucketPutEncryption
```
jonathan-gramain, line 54: Maybe you could mention that it was also not created by the `x-amz-scal-server-side-encryption` header at bucket creation.

lib/api/bucketGetEncryption.js
```diff
@@ -0,0 +1,93 @@
...
+            `<SSEAlgorithm>${escapeForXml(sseInfo.algorithm)}</SSEAlgorithm>`,
+        );
+
+        if (sseInfo.configuredMasterKeyId) {
+            xml.push(`<KMSMasterKeyID>${escapeForXml(sseInfo.masterKeyId)}</KMSMasterKeyID>`);
+        }
+
+        xml.push(
+            '</ApplyServerSideEncryptionByDefault>',
+            '<BucketKeyEnabled>false</BucketKeyEnabled>',
```
jonathan-gramain, line 79: I fixed the design to remove this unsupported tag, I don't know if we should keep it and return `false` or drop the tag completely, it should not hurt to return it but if we can drop it, it would avoid implying that we support the bucket key feature.

tests/functional/aws-node-sdk/test/bucket/getBucketEncryption.js
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    it('should return ServerSideEncryptionConfigurationNotFoundError if no sse configured', done => {
+        s3.getBucketEncryption({ Bucket: bucketName }, err => {
+            checkError(err, 'ServerSideEncryptionConfigurationNotFoundError', 404);
+            done();
+        });
+    });
+
+    it('should return ServerSideEncryptionConfigurationNotFoundError `mandatory` flag not set', done => {
```
jonathan-gramain: **if** mandatory flag not set

jonathan-gramain (approved):
looks good, just minor comments

lib/api/bucketGetEncryption.js
```diff
@@ -0,0 +1,93 @@
...
+        xml.push(
+            '<?xml version="1.0" encoding="UTF-8"?>',
+            '<ServerSideEncryptionConfiguration>',
+            '<Rule>',
+            '<ApplyServerSideEncryptionByDefault>',
+            `<SSEAlgorithm>${escapeForXml(sseInfo.algorithm)}</SSEAlgorithm>`,
+        );
+
+        if (sseInfo.configuredMasterKeyId) {
+            xml.push(`<KMSMasterKeyID>${escapeForXml(sseInfo.masterKeyId)}</KMSMasterKeyID>`);
```
jonathan-gramain: you want to return `configuredMasterKeyId` here instead

tests/functional/aws-node-sdk/test/bucket/getBucketEncryption.js
```diff
@@ -0,0 +1,101 @@
...
+            assert.ifError(err);
+            s3.getBucketEncryption({ Bucket: bucketName }, err => {
+                checkError(err, 'ServerSideEncryptionConfigurationNotFoundError', 404);
+                done();
+            });
+        });
+    });
+
+    it('should include KMSMasterKeyID if algo is aws:kms', done => {
+        setEncryptionInfo({ cryptoScheme: 1, algorithm: 'aws:kms', masterKeyId: '12345', mandatory: true }, err => {
```
jonathan-gramain: same here, pass `configuredMasterKeyId`

jonathan-gramain: Normally you would also have a `masterKeyId` that is generated and different from the configured one.

PR #3560, opened 2021-04-29
ft(S3C-3749): Add PutBucketEncryption Handler

tmacro:
Add `PutBucketEncryption` API handler and bumps Arsenal to pick up routing changes.

### DIFF
--- a/lib/api/apiUtils/bucket/bucketEncryption.js
+++ b/lib/api/apiUtils/bucket/bucketEncryption.js
@@ -1,0 +1,114 @@
+const { errors } = require('arsenal');
+const { parseString } = require('xml2js');
+
+/**
+ * parseEncryptionXml - Parses and validates a ServerSideEncryptionConfiguration xml document
+ * @param {object} xml - ServerSideEncryptionConfiguration doc
+ * @param {object} log - logger
+ * @param {function} cb - callback
+ * @returns {undefined}
+ */
+function parseEncryptionXml(xml, log, cb) {
+    return parseString(xml, (err, parsed) => {
+        if (err) {
+            log.trace('xml parsing failed', {
+                error: err,
+                method: 'parseEncryptionXml',
+            });
+            log.debug('invalid xml', { xml });
+            return cb(errors.MalformedXML);
+        }
+
+        if (!parsed
+            || !parsed.ServerSideEncryptionConfiguration
+            || !parsed.ServerSideEncryptionConfiguration.Rule) {
+            log.trace('error in sse config, invalid ServerSideEncryptionConfiguration section', {
+                method: 'parseEncryptionXml',
+            });
+            return cb(errors.MalformedXML);
+        }
+
+        const { Rule } = parsed.ServerSideEncryptionConfiguration;
+
+        if (!Array.isArray(Rule)
+            || Rule.length > 1
+            || !Rule[0]
+            || !Rule[0].ApplyServerSideEncryptionByDefault
+            || !Rule[0].ApplyServerSideEncryptionByDefault[0]) {
+            log.trace('error in sse config, invalid ApplyServerSideEncryptionByDefault section', {
+                method: 'parseEncryptionXml',
+            });
+            return cb(errors.MalformedXML);
+        }
+
+        const [encConfig] = Rule[0].ApplyServerSideEncryptionByDefault;
+
+        if (!encConfig.SSEAlgorithm || !encConfig.SSEAlgorithm[0]) {
+            log.trace('error in sse config, no SSEAlgorithm provided', {
+                method: 'parseEncryptionXml',
+            });
+            return cb(errors.MalformedXML);
+        }
+
+        const [algorithm] = encConfig.SSEAlgorithm;
+
+        if (algorithm !== 'AES256' && algorithm !== 'aws:kms') {
+            log.trace('error in sse config, unknown SSEAlgorithm', {
+                method: 'parseEncryptionXml',
+            });
+            return cb(errors.MalformedXML);
+        }
+
+        const result = { algorithm };
+
+        if (encConfig.KMSMasterKeyID) {
+            if (algorithm === 'AES256') {
+                log.trace('error in sse config, can not specify KMSMasterKeyID when using AES256', {
+                    method: 'parseEncryptionXml',
+                });
+                return cb(errors.InvalidArgument.customizeDescription(
+                    'a KMSMasterKeyID is not applicable if the default sse algorithm is not aws:kms'));
+            }
+
+            if (!encConfig.KMSMasterKeyID[0] || typeof encConfig.KMSMasterKeyID[0] !== 'string') {
+                log.trace('error in sse config, invalid KMSMasterKeyID', {
+                    method: 'parseEncryptionXml',
+                });
+                return cb(errors.MalformedXML);
+            }
+
+            result.configuredMasterKeyId = encConfig.KMSMasterKeyID[0];
+        }
+        return cb(null, result);
+    });
+}
+
+/**
+ * parseBucketEncryptionHeaders - retrieves bucket level sse configuration from request headers
+ * @param {object} headers - Request headers
+ * @returns {object} SSE configuration
+ */
+function parseBucketEncryptionHeaders(headers) {
+    const sseAlgorithm = headers['x-amz-scal-server-side-encryption'];
+    const configuredMasterKeyId = headers['x-amz-scal-server-side-encryption-aws-kms-key-id'] || null;
+
+    if (sseAlgorithm === 'AES256') {
+        return {
+            algorithm: sseAlgorithm,
+        };
+    } else if (sseAlgorithm === 'aws:kms') {
+        return {
+            algorithm: sseAlgorithm,
+            configuredMasterKeyId,
+        };
+    }
+
+    return {
+        algorithm: null,
+    };
+}
+
+module.exports = {
+    parseEncryptionXml,
+    parseBucketEncryptionHeaders,
+};
--- a/lib/api/bucketPutEncryption.js
+++ b/lib/api/bucketPutEncryption.js
@@ -1,0 +1,84 @@
+const async = require('async');
+
+const { parseEncryptionXml } = require('./apiUtils/bucket/bucketEncryption');
+const metadata = require('../metadata/wrapper');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const kms = require('../kms/wrapper');
+const { pushMetric } = require('../utapi/utilities');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+
+/**
+ * Bucket Put Encryption - Put bucket SSE configuration
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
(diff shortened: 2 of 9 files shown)


lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -0,0 +1,114 @@
...
+    } else if (sseAlgorithm === 'aws:kms') {
+        return {
+            algorithm: sseAlgorithm,
+            configuredMasterKeyId,
+        };
+    }
+
+    return {
+        algorithm: null,
+        configuredMasterKeyId: null,
```
jonathan-gramain: For consistency with what is returned in the AES256 block i.e. only the algorithm, you could leave out the `configuredMasterKeyId` here too I believe (if we consider the `configuredMasterKeyId` to be a side attribute of `aws:kms` it may not make sense to return it without an algorithm set).

lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -0,0 +1,114 @@
...
+            || !parsed.ServerSideEncryptionConfiguration.Rule) {
+            log.trace('error in sse config, invalid ServerSideEncryptionConfiguration section', {
+                method: 'parseEncryptionXml',
+            });
+            return cb(errors.MalformedXML);
+        }
+
+        const { Rule } = parsed.ServerSideEncryptionConfiguration;
+
+        if (!Rule instanceof Array
```
jonathan-gramain: We might also want to return an error if there is more than one item in the `Rule` array

lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -0,0 +1,114 @@
...
+            return cb(errors.MalformedXML);
+        }
+
+        const [encConfig] = Rule[0].ApplyServerSideEncryptionByDefault;
+
+        if (!encConfig.SSEAlgorithm || !encConfig.SSEAlgorithm[0]) {
+            log.trace('error in sse config, no SSEAlgorithm provided', {
+                method: 'parseEncryptionXml',
+            });
+        }
```
jonathan-gramain, line 51: You likely want to return the error here

lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -0,0 +1,114 @@
...
+            return cb(errors.MalformedXML);
+        }
+
+        if (!parsed
+            || !parsed.ServerSideEncryptionConfiguration
+            || !parsed.ServerSideEncryptionConfiguration.Rule) {
+            log.trace('error in sse config, invalid ServerSideEncryptionConfiguration section', {
+                method: 'parseEncryptionXml',
+            });
+            return cb(errors.MalformedXML);
```
jonathan-gramain, line 28: I believe it could be beneficial to debugging if we also return a customized description of the error each time there is an invalid XML for some reason (basically a copy of the trace level message)

lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            if (!encConfig.KMSMasterKeyID[0] || typeof encConfig.KMSMasterKeyID[0] !== 'string') {
+                log.trace('error in sse config, invalid KMSMasterKeyID', {
+                    method: 'parseEncryptionXml',
+                });
+                return cb(errors.MalformedXML);
+            }
+
+            const [configuredMasterKeyId] = encConfig.KMSMasterKeyID;
```
jonathan-gramain: it could be done with `result.configuredMasterKeyId = encConfig.KMSMasterKeyId[0]`

lib/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,107 @@
...
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+
+function bucketPutEncryption(authInfo, request, log, callback) {
+    const bucketName = request.bucketName;
+    const canonicalID = authInfo.getCanonicalID();
+
+    return async.waterfall([
+        next => metadata.getBucket(bucketName, log, (err, bucket) => {
```
jonathan-gramain: Is there a reason not to use the `metadataValidateBucket` helper instead? I was wondering about this while reading the GetBucketEncryption implementation.

lib/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,107 @@
...
+                bucket: bucketName,
+                method: 'bucketPutEncryption',
+            });
+
+            if (!isBucketAuthorized(bucket, requestType, canonicalID, authInfo, log)) {
+                log.debug('access denied for account on bucket', {
+                    requestType,
+                    method: 'bucketPutEncryption',
+                });
+                return next(errors.AccessDenied, null);
```
jonathan-gramain: is it needed to pass `null` as second argument? (assuming it's the bucket name for cors checking)

lib/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,107 @@
...
+                updatedConfig.configuredMasterKeyId = configuredMasterKeyId;
+            }
+
+            return next(null, bucket, updatedConfig);
+        },
+        (bucket, updatedConfig, next) => {
+            bucket.setServerSideEncryption(updatedConfig);
+            metadata.updateBucket(bucket.getName(), bucket, log, err => next(err, bucket));
+        },
+
```
jonathan-gramain: extra newline

lib/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,107 @@
...
+    ],
+    (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
+        if (err) {
+            log.trace('error processing request', { error: err, method: 'putBucketEncryption' });
```
jonathan-gramain: the method name is `bucketPutEncryption`

tests/unit/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,310 @@
...
+    return {
+        bucketName,
+        headers: { host: `${bucketName}.s3.amazonaws.com` },
+        post,
+    };
+}
+
+function getSSEConfig(cb) {
+    return metadata.getBucket(bucketName, log, (err, md) => {
+        if (err) return cb(err);
```
jonathan-gramain: style: missing newline before `return`

tests/unit/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,310 @@
...
+        return cb(null, md.getServerSideEncryption());
+    });
+}
+
+describe.only('bucketPutEncryption API', () => {
```
jonathan-gramain: remove `.only` ;)

tests/unit/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,310 @@
...
+    before(() => cleanup());
```
jonathan-gramain, line 59: Is the cleanup needed before starting the tests?

tests/unit/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,310 @@
...
+        it('should reject a config with SSEAlgorithm == AES256 and a provided KMSMasterKeyID', done => {
+            const post = templateSSEConfig({ algorithm: 'AES256', keyId: '12345' });
+            bucketPutEncryption(authInfo, templateRequest({ post }), log, err => {
+                assert.strictEqual(err.InvalidArgument, true);
+                done();
+            });
+        });
+    });
+
+    describe('test setting config without a prvious one', () => {
```
jonathan-gramain: typo: pr**e**vious

tests/unit/api/bucketPutEncryption.js
```diff
@@ -0,0 +1,310 @@
...
+        it('should apply a config with SSEAlgorithm == AES256', done => {
+            const post = templateSSEConfig({ algorithm: 'AES256' });
+            bucketPutEncryption(authInfo, templateRequest({ post }), log, err => {
+                assert.ifError(err);
+                return getSSEConfig((err, sseInfo) => {
+                    assert.ifError(err);
+                    assert.notStrictEqual(sseInfo, null);
```
jonathan-gramain: Those two first checks are not needed since there is a `deepStrictEqual` afterwards, regarding ease of debugging it should be fine since `deepStrictEqual` should show the actual diff between what's expected and what's obtained. Same comment for following tests.

lib/api/apiUtils/bucket/bucketEncryption.js
tmacro: I investigated each of these error cases and matched the desctription returned by AWS. I've seen issues raised in the past when wording didn't match so I wanted to be strict here.

lib/api/bucketPutEncryption.js
tmacro: I was basing these handlers off of `bucketGetLocation` which doesn't use the helper. I'll switch over as it looks like it covers a bunch of this.

lib/api/bucketPutEncryption.js
tmacro: Same as above

tests/unit/api/bucketPutEncryption.js
tmacro: I was expecting `deepStrictEqual` to assert that these objects were the same `
```
{
    mandatory: true,
    algorithm: 'AES256',
    cryptoScheme: 1,
}
```
```
{
    mandatory: true,
    algorithm: 'AES256',
    cryptoScheme: 1,
    masterKeyId: undefined,
}
```

tmacro: I forgot that an explicit value of `undefined` is different than the key not existing

tests/unit/api/bucketPutEncryption.js
tmacro: in case the previous tests failed to cleanup (errors etc), Internally it consists of a `this._db = {}` so there's functionally no extra overhead

lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -0,0 +1,114 @@
...
+            || !parsed.ServerSideEncryptionConfiguration.Rule) {
+            log.trace('error in sse config, invalid ServerSideEncryptionConfiguration section', {
+                method: 'parseEncryptionXml',
+            });
+            return cb(errors.MalformedXML);
+        }
+
+        const { Rule } = parsed.ServerSideEncryptionConfiguration;
+
+        if (!(Rule instanceof Array)
```
miniscruff: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray can use this method to check if something is an array. It says it is preferred over instanceof but the context is very irrelevant here so take that for what it is. Edit: If you want to get rid of the nested paranthesis.

lib/api/bucketPutEncryption.js
jonathan-gramain: Do you plan to use `metadataValidateBucket` finally? Wondering since you mentioned you would, not that it's blocking for me.

lib/api/bucketPutEncryption.js
jonathan-gramain: looks like you forgot to change this one :slightly_smiling_face:

jonathan-gramain (approved):

miniscruff (approved):

PR #3584, opened 2021-05-07
bf: S3C-3061 allow bucket acl via predefined groups

naren-rajendran:
Corrections to bucket ACL using s3 predefined groups. 

Ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html

### DIFF
--- a/lib/api/objectDelete.js
+++ b/lib/api/objectDelete.js
@@ -22,6 +22,10 @@ const versionIdUtils = versioning.VersionID;
  */
 function objectDelete(authInfo, request, log, cb) {
     log.debug('processing request', { method: 'objectDelete' });
+    if (authInfo.isRequesterPublicUser()) {
+        log.debug('operation not available for public user');
+        return cb(errors.AccessDenied);
+    }
     const bucketName = request.bucketName;
     const objectKey = request.objectKey;
 
@@ -50,12 +54,6 @@ function objectDelete(authInfo, request, log, cb) {
             (err, bucketMD, objMD) => {
                 if (err) {
                     return next(err, bucketMD);
-                }
-                // based on AWS's behavior, object delete can be performed by
-                // either bucket owner or the object owner
-                if ((objMD && objMD['owner-id'] !== canonicalID)
-                    && bucketMD.getOwner() !== canonicalID) {
-                    return next(errors.AccessDenied);
                 }
                 const versioningCfg = bucketMD.getVersioningConfiguration();
                 if (!objMD) {
--- a/lib/api/objectDelete.js
+++ b/lib/api/objectDelete.js
@@ -22,10 +22,6 @@ const versionIdUtils = versioning.VersionID;
  */
 function objectDelete(authInfo, request, log, cb) {
     log.debug('processing request', { method: 'objectDelete' });
-    if (authInfo.isRequesterPublicUser()) {
-        log.debug('operation not available for public user');
-        return cb(errors.AccessDenied);
-    }
     const bucketName = request.bucketName;
     const objectKey = request.objectKey;
 
@@ -54,6 +50,12 @@ function objectDelete(authInfo, request, log, cb) {
             (err, bucketMD, objMD) => {
                 if (err) {
                     return next(err, bucketMD);
+                }
+                // based on AWS's behavior, object delete can be performed by
+                // either bucket owner or the object owner
+                if ((objMD && objMD['owner-id'] !== canonicalID)
+                    && bucketMD.getOwner() !== canonicalID) {
+                    return next(errors.AccessDenied);
                 }
                 const versioningCfg = bucketMD.getVersioningConfiguration();
                 if (!objMD) {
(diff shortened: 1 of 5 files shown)


naren-rajendran:
Note to reviewers, the tests have to be refactored.

tmacro (approved):

tests/functional/aws-node-sdk/test/bucket/aclUsingPredefinedGroups.js
```diff
@@ -0,0 +1,634 @@
...
+    function unauthenticatedRequest(operation, params, callback) {
+        const bucketUtil = new BucketUtility('default', sigCfg);
+        const request = bucketUtil.s3[operation](params);
+        request.removeListener('validate', VALIDATE_CREDENTIALS);
+        request.removeListener('sign', SIGN);
+        request.send(callback);
+    }
+
+    // tests for authenticated user(signed) and anonymous user(unsigned)
+    [true, false].forEach(auth => {
```
miniscruff, line 47: splitting all these tests with a bool here is making it quite hard to read and adds at least one if check to every test, I think it would be better to create two describes instead.

tests/functional/aws-node-sdk/test/bucket/aclUsingPredefinedGroups.js
```diff
@@ -0,0 +1,634 @@
...
+                s3.putBucketAcl({
+                    Bucket: testBucket,
+                    GrantRead: grantUri,
+                }, err => {
+                    assert.ifError(err);
+                    const param = { Bucket: testBucket };
+                    const cb = er => {
+                        assert.ifError(er);
+                        done();
+                    };
```
miniscruff: Instead of including this each time you could create a helper method

```js
function wrapCallback(done) {
  return err => {
    assert.ifError(err);
    done();
  };
}
```
Then use `wrapCallback(done)` in place of `cb` in all successful future cases.

tests/functional/aws-node-sdk/test/bucket/aclUsingPredefinedGroups.js
```diff
@@ -0,0 +1,634 @@
...
+                }
+                s3.putBucketAcl(aclParam, err => {
+                    assert.ifError(err);
+                    const param = {
+                        Bucket: testBucket,
+                        Body: testBody,
+                        Key: testKey,
+                    };
+                    const cb = er => {
+                        assert(er.code, errors.AccessDenied);
```
miniscruff: Did you mean to use `assert.strictEqual`? This will return true if er.code is not 0, then use access denied as the message.

tests/functional/aws-node-sdk/test/bucket/aclUsingPredefinedGroups.js
naren-rajendran: The tests are the same for both the cases and that is why I had to do this. I originally had separate tests.

tests/functional/aws-node-sdk/test/bucket/aclUsingPredefinedGroups.js
naren-rajendran: Yep, it was a copy-paste.

tests/functional/aws-node-sdk/test/bucket/aclUsingPredefinedGroups.js
miniscruff: I would not consider the tests the same if you need an if/else in the test based on the condition. It is not like a table test where you are testing many inputs against the same method and output. `auth` is used as a conditional check a lot.

lib/api/objectDelete.js
```diff
@@ -73,6 +69,12 @@ function objectDelete(authInfo, request, log, cb) {
                     // versioning has been configured
                     return next(null, bucketMD, objMD);
                 }
+                // based on AWS's behavior, object delete can be performed by
+                // either bucket owner or the object owner
+                if (objMD['owner-id'] !== canonicalID
```
jonathan-gramain: The difference is subtle, but if you move this check above the `if (!objMD)` and add the `objMD && ` condition at the beginning (similar to what is done for `objectPut`), it allows to return `AccessDenied` whether the object exists or not, without leaking info about the object existence for unauthorized users. I don't know what AWS does in this respect but it could be good to check.

naren-rajendran: Yes, corrected. Checked AWS behavior, you get access denied on non existing object.

tests/functional/aws-node-sdk/test/bucket/aclUsingPredefinedGroups.js
naren-rajendran: Refactored a bit.

miniscruff: I will accept it, but I tend to dislike conditionals in tests as I think a straight top-down given-when-then flow works more consistently.

miniscruff (approved):

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e force_reset

PR #3612, opened 2021-05-13
Feature/s3 c 3755 add bucket delete encryption handler

tmacro:
~Note for reviewers: Arsenal has been bumped to the hash of the Arsenal PR branch, and will be changed after it has merged.~

### DIFF
--- a/lib/api/apiUtils/bucket/bucketEncryption.js
+++ b/lib/api/apiUtils/bucket/bucketEncryption.js
@@ -1,5 +1,6 @@
 const { errors } = require('arsenal');
 const { parseString } = require('xml2js');
+const vault = require('../../../auth/vault');
 
 /**
  * parseEncryptionXml - Parses and validates a ServerSideEncryptionConfiguration xml document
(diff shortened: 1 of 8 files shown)


lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -108,7 +109,31 @@ function parseBucketEncryptionHeaders(headers) {
...
+function checkExpectedBucketOwner(headers, bucket, log, cb) {
+    const expectedOwner = headers['x-amz-expected-bucket-owner'];
+    if (expectedOwner === undefined) {
+        return cb();
+    }
+
+    const bucketOwner = bucket.getOwner();
+    return vault.getAccountIds([bucketOwner], log, (error, res) => {
+        if (error) {
+            log.error('error fetch accountId from vault', {
```
jonathan-gramain, line 121:
```suggestion
            log.error('error fetching accountId from vault', {
```

lib/api/apiUtils/bucket/bucketEncryption.js
```diff
@@ -108,7 +109,31 @@ function parseBucketEncryptionHeaders(headers) {
     };
 }
 
+function checkExpectedBucketOwner(headers, bucket, log, cb) {
```
jonathan-gramain, line 112: Why do we check this header specifically for delete-bucket-encryption? According to https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-owner-condition.html, it looks to me this header is generic to pretty much all object and bucket operations, and optional, so kinda looks orthogonal to bucket encryption.

tests/unit/api/bucketDeleteEncryption.js
```diff
@@ -0,0 +1,80 @@
...
+                return getSSEConfig(bucketName, (err, sseInfo) => {
+                    assert.ifError(err);
+                    assert.strictEqual(sseInfo.mandatory, false);
+                    done();
+                });
+            });
+        });
+    });
+
+    it('should disable mandatory sse for aws:kms', done => {
```
jonathan-gramain: You might want to merge this test with the preceeding one and a `forEach` on the algorithm since the logic is essentially the same

lib/api/apiUtils/bucket/bucketEncryption.js
tmacro: It was accidentally skipped in put/get. I've created a ticket to add support for them.

jonathan-gramain: OK sounds good

Ideally I think it should not be defined in a bucket-encryption-specific file but it should be part of the common checks path for all request types.

jonathan-gramain (approved):

lib/api/apiUtils/bucket/bucketEncryption.js
tmacro: Added a note in the ticket to look at moving it.

naren-rajendran (approved):

PR #3620, opened 2021-05-17
S3C-2663: fix missing object access error response

alexanderchan-scality:
Add checks to return proper error response in the case of object 404s; `AccessDenied` when not granted bucket read vs `NoSuchKey` when granted bucket read.

note for reviewers: another PR will be created to handle bucket policies in development/7.10 (https://github.com/scality/cloudserver/pull/3622)

### DIFF
--- a/lib/api/apiUtils/authorization/aclChecks.js
+++ b/lib/api/apiUtils/authorization/aclChecks.js
@@ -86,11 +86,56 @@ function isBucketAuthorized(bucket, requestType, canonicalID) {
     requestType === 'objectGet' || requestType === 'objectHead');
 }
 
+function _isPermissionGranted(permissionList, canonicalID) {
+    if (permissionList.indexOf(publicId) > -1) {
+        return true;
+    }
+
+    if (permissionList.indexOf(allAuthedUsersId) > -1 &&
+        canonicalID !== publicId) {
+        return true;
+    }
+
+    if (permissionList.indexOf(canonicalID) > -1) {
+        return true;
+    }
+
+    return false;
+}
+
+function hasBucketReadAccess(bucket, requestType, canonicalID) {
+    const bucketAcl = bucket.getAcl();
+    const bucketOwner = bucket.getOwner();
+
+    if (canonicalID === bucketOwner ||
+        _isPermissionGranted(bucketAcl.FULL_CONTROL, canonicalID) ||
+        _isPermissionGranted(bucketAcl.READ, canonicalID)) {
+        return true;
+    }
+
+    if (bucketAcl.Canned === 'public-read' ||
+        bucketAcl.Canned === 'public-read-write' ||
+        (bucketAcl.Canned === 'authenticated-read' &&
+         canonicalID !== publicId)) {
+        return true;
+    }
+
+    return false;
+}
+
 function isObjAuthorized(bucket, objectMD, requestType, canonicalID) {
     const bucketOwner = bucket.getOwner();
     if (!objectMD) {
-        return false;
+        if (requestType === 'objectPut' || requestType === 'objectDelete') {
+            return true;
+        }
+
+        // if read access is granted, return true to have the api handler
+        // respond accordingly to the missing object metadata
+        // if read access is not granted, return false for AccessDenied
+        return hasBucketReadAccess(bucket, requestType, canonicalID);
     }
+
     if (objectMD['owner-id'] === canonicalID) {
         return true;
     }
--- a/tests/unit/api/objectACLauth.js
+++ b/tests/unit/api/objectACLauth.js
@@ -207,3 +207,98 @@ describe('object authorization for objectPutACL and objectGetACL', () => {
         assert.strictEqual(authorizedResult, true);
     });
 });
+
+describe('without object metadata', () => {
+    afterEach(() => {
+        bucket.setFullAcl({
+            Canned: 'private',
+            FULL_CONTROL: [],
+            WRITE: [],
+            WRITE_ACP: [],
+            READ: [],
+            READ_ACP: [],
+        });
+    });
+
+    const requestTypes = [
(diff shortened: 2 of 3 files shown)


lib/api/apiUtils/authorization/aclChecks.js
```diff
@@ -86,11 +86,58 @@ function isBucketAuthorized(bucket, requestType, canonicalID) {
...
+    return false;
+}
+
+function hasBucketReadAccess(bucket, requestType, canonicalID) {
+    const bucketAcl = bucket.getAcl();
+    const bucketOwner = bucket.getOwner();
+
+    // User is already authorized on the bucket for FULL_CONTROL or WRITE or
+    // bucket has canned ACL public-read-write
+    if (requestType === 'objectPut' || requestType === 'objectDelete') {
```
jonathan-gramain: I find it confusing to have this check done in this helper function, i.e. checking write requests in a "has read access?" checker. A couple questions then:
- do we need this check, or could it be already covered by the `canonicalID === bucketOwner` + `bucketAcl.Canned === 'public-read-write'` checks below?
- wondering if the user is not the bucket owner but has `WRITE` access without `READ`, what should we return in this case?
- otherwise, if the check on `requestType` is needed, would it be better to do this check earlier, before calling this function?

lib/api/apiUtils/authorization/aclChecks.js
```diff
@@ -86,11 +86,58 @@ function isBucketAuthorized(bucket, requestType, canonicalID) {
     requestType === 'objectGet' || requestType === 'objectHead');
 }
 
+function _isPermissionGranted(permissionList, canonicalID) {
+    if (permissionList.indexOf(publicId) > -1) {
```
naren-rajendran, line 90: Do we need multiple `if`s, can these be combined?

tests/unit/api/objectACLauth.js
```diff
@@ -207,3 +207,110 @@ describe('object authorization for objectPutACL and objectGetACL', () => {
...
+            bucket.setCannedAcl(value.canned);
+            const results = requestTypes.map(type =>
+                isObjAuthorized(bucket, null, type, value.id));
+            assert.deepStrictEqual(results, value.response);
+            done();
+        });
+    });
+
+
+    it('should allow access to anyone since checks ' +
```
naren-rajendran: I am not sure if this test is needed, this is a false positive.

lib/api/apiUtils/authorization/aclChecks.js
alexanderchan-scality: this definitely could be condensed into a single return statement

lib/api/apiUtils/authorization/aclChecks.js
alexanderchan-scality: > I find it confusing to have this check done in this helper function, i.e. checking write requests in a "has read access?" checker

I can see how that can be confusing;~the idea is that if this wasn't blocked at the bucket level, the user should have read access, but this would not be true if the user was only granted `WRITE` permissions.~ will fix

lib/api/apiUtils/authorization/aclChecks.js
jonathan-gramain: I personally prefer the separate `if` style whenever merging into a single statement would require extra parenthesis or if it would rely on operator precedence rules to avoid them. I just think it makes the code easier to follow this way. This is just my personal preference, I let you decide if you prefer to merge or not :)

lib/api/apiUtils/authorization/aclChecks.js
alexanderchan-scality: moved `objectPut` and `objectDelete` check out of the `read` check;

> wondering if the user is not the bucket owner but has WRITE access without READ, what should we return in this case?

verified against S3 and `objectPut`/`objectDelete` are only blocked if writes are not granted, and this check is handled on the bucket level check

tests/unit/api/objectACLauth.js
alexanderchan-scality: indeed. in both cases of `objectMD`, the result would be true.

lib/api/apiUtils/authorization/aclChecks.js
alexanderchan-scality: kept as separate conditions, I find that it reads easier that way as well

tests/unit/api/objectACLauth.js
```diff
@@ -207,3 +207,98 @@ describe('object authorization for objectPutACL and objectGetACL', () => {
...
+            response: allowedAccess,
+        },
+        {
+            it: 'should not allow public if canned private',
+            canned: 'private', id: constants.publicId,
+            aclParam: null,
+            response: deniedAccess,
+        },
+        {
+            it: 'should not other accounts if canned private',
```
jonathan-gramain:
```suggestion
            it: 'should not allow other accounts if canned private',
```

tests/unit/api/objectACLauth.js
```diff
@@ -207,3 +207,98 @@ describe('object authorization for objectPutACL and objectGetACL', () => {
...
+            aclParam: null,
+            response: allowedAccess,
+        },
+        {
+            it: 'should allow account if granted bucket READ',
+            canned: '', id: accountToVet,
+            aclParam: ['READ', accountToVet],
+            response: allowedAccess,
+        },
+        {
```
jonathan-gramain, line 284: I suggest adding a test for `WRITE` ACL as well, if you have time

jonathan-gramain (approved):

tests/unit/api/objectACLauth.js
naren-rajendran: `WRITE` is controlled or checked at bucket level.

naren-rajendran (approved):

PR #3659, opened 2021-06-09
S3C-4110 Backport lifecycle expiration compatibility improvement

nicolas2bert:
Add tests + Update Arsenal with:
- Include the Prefix in the response even if it is an empty string.

- Check the max lengths of certain values (i.e. filter prefix length) to avoid DDOS-like attack surfaces where a very large value could be accepted by the API and pushed to Kafka.

- When setting multiple lifecycle filter tags, we previously only applied one tag.
This fix should apply all unique tags. If all tags are not unique, should error out.


rahulreddy:
Issuing a bypass a Azure is flaky
/bypass_build_status

alexanderchan-scality (approved):

naren-rajendran (approved):

nicolas2bert:
@bert-e approve

PR #3660, opened 2021-06-10
ft: S3C-4551 tiny version IDs

vrancurel:
The tiny mode is NOT selected by default when enabling versioning. It is for now configurable for deployments that need it.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -14,6 +14,9 @@ const { buildAuthDataAccount } = require('./auth/in_memory/builder');
 const externalBackends = require('../constants').externalBackends;
 const { azureAccountNameRegex, base64Regex } = require('../constants');
 const { utapiVersion } = require('utapi');
+const { versioning } = require('arsenal');
+
+const versionIdUtils = versioning.VersionID;
 
 // whitelist IP, CIDR for health checks
 const defaultHealthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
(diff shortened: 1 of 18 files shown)


vrancurel:
There is probably much better/simpler/more elegant approaches. Please suggest.

lib/Config.js
```diff
@@ -1284,11 +1287,32 @@ class Config extends EventEmitter {
     }
 }
 
+// get the version ID encoding type
+function getVersionIdEncType(verCfg) {
+    if (verCfg === undefined) {
+        throw new Error('Invalid version configuration');
```
jonathan-gramain: it looks dangerous to me to throw here, e.g. at https://github.com/scality/cloudserver/pull/3660/files#diff-d9d6c7d27a37666321fe0d79f3d76b7349a03856a0933381651431fb4e5cff62R341 are we sure that there is always a defined versioning configuration?

vrancurel: I realize that `decode()` is sometimes called without a versioningConfiguration. Problematic.

vrancurel: So if you look carefully I somehow managed to call `getEncType()` only when needed and it does not assert any more.

lib/Config.js
```diff
@@ -1284,11 +1287,27 @@ class Config extends EventEmitter {
     }
 }
 
+// get the version ID encoding type
+function getVersionIdEncType() {
+    let encType;
+    if (process.env.VERSION_ID_ENC_TYPE !== undefined) {
```
anurag4DSB: are the possible values 0 or 1?
If yes, will note for documentation purposes.

anurag4DSB (commented):
What should we do about older objects?
will we create a migration tool?

lib/Config.js
vrancurel: Yes, but I think we need set ENC_TYPE_HEX by default. I am just trying to pass the tests with BASE62 then I will set HEX by default (for now).

vrancurel:
> What should we do about older objects?
> will we create a migration tool?

This is the same binary format. Only the presentation changes.

But we would need a flag per bucket to state if we shall use the old or new representation implicitly or explicitly.
This state can be managed by a bucket property "BucketFeature" that we can use for other breaking changes.

saitharunthodupunuru (approved):
LGTM

miniscruff (approved):

saitharunthodupunuru (approved):

anurag4DSB (approved):

rahulreddy (approved):

rahulreddy:
Bypassing as Azure flakiness work is coming

PR #3663, opened 2021-06-11
bugfix: S3C-4324 bucket policy allows notification actions

jonathan-gramain:
* bugfix: S3C-4324 shared action map to check bucket policy actions

  When checking bucket policy actions, instead of using a hard-coded local action map which was missing a few items, check against the shared bucket policy action map in arsenal.

* bugfix: S3C-4324 deny any unsupported action

  As a security feature rather than a functional change, all action types that are not explicitly known are denied instead of allowed (or subject to further non-action policy controls).

Integration run: https://eve.devsca.com/github/scality/integration/#/builders/11/builds/3432 (fails on node5 due to a known issue in backbeat, unrelated)


naren-rajendran (approved):

mrepop (approved):
LGTM

PR #3682, opened 2021-06-24
add x-amz-bucket-region header on bucket HEAD response

watsaqat:


lib/api/bucketHead.js
```diff
@@ -30,7 +30,8 @@ function bucketHead(authInfo, request, log, callback) {
             authInfo,
             bucket: bucketName,
         });
-        return callback(null, corsHeaders);
+        const headers = { 'x-amz-bucket-region': bucket._locationConstraint };
```
tmacro: logic looks good, however the model provides a getter for this.
https://github.com/scality/Arsenal/blob/6347358cc22c4f2703f0ff309b006f1842bd1e40/lib/models/BucketInfo.js#L535

tmacro (changes requested):

lib/api/bucketHead.js
watsaqat: thanks, didn't find it at first

jonathan-gramain (approved):
Whenever you're ready to have the code merged, you can squash the commits into one, add the ticket number in the commit summary, before doing `/approve`

tmacro (approved):

PR #3739, opened 2021-07-30
bugfix: CLDSRV-5 Remove extra headers from 304 responses

miniscruff:
Note: This will fail until the arsenal version is updated from https://github.com/scality/Arsenal/pull/1526

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/get.js
+++ b/tests/functional/aws-node-sdk/test/object/get.js
@@ -658,13 +658,19 @@ describe('GET object', () => {
             it('If-None-Match & If-Modified-Since: returns NotModified when ' +
                 'Etag does not match and lastModified is greater',
                 done => {
-                    requestGet({
+                    const req = s3.getObject({
+                        Bucket: bucketName,
+                        Key: objectName,
                         IfNoneMatch: etagTrim,
                         IfModifiedSince: dateFromNow(-1),
                     }, err => {
                         checkError(err, 'NotModified');
                         done();
                     });
+                    req.on('httpHeaders', (code, headers, response, status) => {
+                        assert(!headers.hasOwnProperty('content-type'));
+                        assert(!headers.hasOwnProperty('content-length'));
+                    })
                 });
 
             it('If-None-Match not match & If-Modified-Since not match',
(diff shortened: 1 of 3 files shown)


naren-rajendran (approved):

saitharunthodupunuru (approved):

tests/functional/aws-node-sdk/test/object/get.js
```diff
@@ -658,13 +658,19 @@ describe('GET object', () => {
...
+                        Bucket: bucketName,
+                        Key: objectName,
                         IfNoneMatch: etagTrim,
                         IfModifiedSince: dateFromNow(-1),
                     }, err => {
                         checkError(err, 'NotModified');
                         done();
                     });
+                    req.on('httpHeaders', (code, headers, response, status) => {
+                        assert(!headers.hasOwnProperty('content-type'));
```
saitharunthodupunuru: Change this to

```suggestion
                       assert(!headers.hasOwnProperty('content-type') || !headers.hasOwnProperty('Content-Type'));
```

tests/functional/aws-node-sdk/test/object/get.js
```diff
@@ -658,13 +658,19 @@ describe('GET object', () => {
...
+                        assert(!headers.hasOwnProperty('content-length'));
```
saitharunthodupunuru: same as above

saitharunthodupunuru (changes requested):

tests/functional/aws-node-sdk/test/object/get.js
miniscruff: Hmm, I know from the outputs the results have all lower case keys so not sure it is worth adding the other case.

saitharunthodupunuru: Hmm That is fine too. If the output has always lower case

saitharunthodupunuru (approved):

tests/functional/aws-node-sdk/test/object/get.js
rahulreddy: No need to check for case sensitiveness. Node.js converts all headers to lower case https://nodejs.org/api/http.html#http_message_headers
Also a check like `!headers['content-type']` is sufficient.

PR #3747, opened 2021-08-02
VAULT-7 unexpected 403 errors

nicolas2bert:
When a user, with bucket read access, requests a non-existing object (head object, get object tagging...) we are returning a **403** instead of a **404** error.

`hasBucketReadAccess` and `isBucketAuthorized` (with `bucketGet` as the "request type" argument) seems to have the same logic. 

`hasBucketReadAccess` was missing the `bucket.getOwner() === canonicalID` condition that will allow user (with the right user policy) to access its account bucket's objects without any extra bucket policy or bucket ACL needed.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -278,85 +278,17 @@ function isBucketAuthorized(bucket, requestType, canonicalID, authInfo, log) {
     return (aclPermission || (bucketPolicyPermission === 'allow'));
 }
 
-function _isPermissionGranted(permissionList, canonicalID) {
-    if (permissionList.indexOf(publicId) > -1) {
-        return true;
-    }
-
-    if (permissionList.indexOf(allAuthedUsersId) > -1 &&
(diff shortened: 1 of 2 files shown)


jonathan-gramain (commented):
Isn't this PR for 7.10?

jonathan-gramain (commented):
@nicolas2bert can you ask Anurag to review this instead? I'm not an expert in bucket/IAM policies and have a hard time understanding what this PR attempts to do. If you want to spend time explaining it more in detail to me, I may be able to review it more thoroughly then.

nicolas2bert:
@jonathan-gramain yes, it should be against `development/7.10`. Thank you. I will also improve the description.

jonathan-gramain:
Aren't VAULT tickets supposed to be used only for PRs in the Vault project?

nicolas2bert:
@jonathan-gramain we first thought that this issue was a Vault issue.

anurag4DSB (approved):

nicolas2bert:
Using the correct JIRA project name: https://github.com/scality/cloudserver/pull/3750

PR #3750, opened 2021-08-03
CLDSRV-6 unexpected 403 errors

nicolas2bert:
When a user, with bucket read access, requests a non-existing object (head object, get object tagging...) we are returning a **403** instead of a **404** error.

`hasBucketReadAccess` and `isBucketAuthorized` (with `bucketGet` as the "request type" argument) seems to have the same logic. 

`hasBucketReadAccess` was missing the `bucket.getOwner() === canonicalID` condition that will allow user (with the right user policy) to access its account bucket's objects without any extra bucket policy or bucket ACL needed.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -278,85 +278,17 @@ function isBucketAuthorized(bucket, requestType, canonicalID, authInfo, log) {
     return (aclPermission || (bucketPolicyPermission === 'allow'));
 }
 
-function _isPermissionGranted(permissionList, canonicalID) {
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

alexanderchan-scality:
@nicolas2bert does this also resolve the issue for role type authinfos?

nicolas2bert:
@alexanderchan-scality Yes, user that assumes `account1` role should be able to perform "role policy allowed" actions on `account1` buckets without any extra bucket policy or bucket ACL needed.

alexanderchan-scality (approved):

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #3769, opened 2021-08-12
bf CLDSRV-10 get acl with predefined groups

naren-rajendran:
Corrected listing ACLs for an object or a bucket when predefined groups are used.

### DIFF
--- a/lib/utilities/aclUtils.js
+++ b/lib/utilities/aclUtils.js
@@ -354,6 +354,8 @@ aclUtils.getCanonicalIDs = function getCanonicalIDs(acl) {
         acl.READ_ACP
     );
     const uniqueGrantees = Array.from(new Set(aclGrantees));
+    // grantees can be a mix of canonicalIDs and predefined groups in the form
+    // of uri, so filter out only canonicalIDs
     return uniqueGrantees.filter(item => item && !grantsByURI.includes(item));
 };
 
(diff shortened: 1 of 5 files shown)


naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

lib/utilities/aclUtils.js
```diff
@@ -320,4 +325,89 @@ aclUtils.checkGrantHeaderValidity = function checkGrantHeaderValidity(headers) {
...
+                    individualGrantInfo.push({
+                        ID: grantee,
+                        displayName,
+                        permission,
+                    });
+                }
+            });
+        }
+    });
+    return individualGrantInfo;
```
miniscruff: This feels overly complex but it was kinda fun to make...

```suggestion
    return Object.entries(getGrants(acl))
        .filter(([permission, grantees]) => Array.isArray(grantees))
        .map(([permission, grantees]) => {
            grantees.filter(g => emails[g])
                .filter(g => canonicalIDs.includes(g))
                .map(g => { return {
                    ID: g,
                    displayName: emails[g],
                    permission,
                }})
        })
        .flat();
```

naren-rajendran: Yeah I looked at flat and made something similar, but had to revert it because we are on node v10 and `Array.flat()` isn't supported. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

naren-rajendran: Also we can combine the two filters into one, `.filter(g => emails[g] && canonicalIDs.includes(g))`

miniscruff: Nodejs 11.0... lame.
Combining filters is fine, really just preference it won't be any faster.

naren-rajendran:
@bert-e reset

miniscruff (approved):

anurag4DSB (approved):

naren-rajendran:
@bert-e ping

naren-rajendran:
@bert-e help

PR #3782, opened 2021-08-13
bugfix: CLDSRV-11 Write ACL via predefined groups

miniscruff:


miniscruff:
ping

miniscruff:
reset

naren-rajendran (approved):

lib/api/apiUtils/authorization/aclChecks.js
```diff
@@ -73,6 +74,13 @@ function isBucketAuthorized(bucket, requestType, canonicalID) {
             || bucketAcl.FULL_CONTROL.indexOf(canonicalID) > -1
             || bucketAcl.WRITE.indexOf(canonicalID) > -1) {
             return true;
+        } else if (bucketAcl.WRITE.indexOf(publicId) > -1
```
alexanderchan-scality, line 77: both of the blocks return `true`, so they can be merged.

lib/api/objectDelete.js
```diff
@@ -55,6 +55,12 @@ function objectDelete(authInfo, request, log, cb) {
                 if (err) {
                     return next(err, bucketMD);
                 }
+                // based on AWS's behavior, object delete can be performed by
+                // either bucket owner or the object owner
+                if ((objMD && objMD['owner-id'] !== canonicalID)
+                    && bucketMD.getOwner() !== canonicalID) {
```
alexanderchan-scality, line 61: ~would this conditional affect the `public-read-write` behavior?~

lib/api/apiUtils/authorization/aclChecks.js
miniscruff: I believe this was to match the other conditionals in the method, I cherry-picked it from a commit by @naren-scality

lib/api/objectDelete.js
naren-rajendran: @alexanderchan-scality how do you think this would affect?

alexanderchan-scality: The comment is not valid. I was under the impression that the `public-read-write' should allow other users to overwrite object even when they don't own it. Which is not the case in real AWS as mentioned in the PR. double-checked against real AWS
the check is still valid.

alexanderchan-scality: this check can however added as part of the `isObjAuthorized` method as opposed to each of the api handlers

naren-rajendran: This was very specific to object delete and wanted to avoid conditionals in `isObjAuthorized`.

alexanderchan-scality: i see. does the check only occur for the `objectPut` and `objectDelete` operation? I was wondering about the other api calls that would result in object overwrites like `objectCopy`

naren-rajendran: `objectCopy` is a combo of `objectGet` and `objectPut` so the user performing the action should have permission to read and write and `complete-mpu` is same as objectPutPart which again internally is an `objectPut`

alexanderchan-scality (approved):

PR #3800, opened 2021-08-19
bugfix: CLDSRV-11 Remove ownership checks

miniscruff:
This should also resolve the integration test

### DIFF
--- a/lib/api/objectDelete.js
+++ b/lib/api/objectDelete.js
@@ -55,12 +55,7 @@ function objectDelete(authInfo, request, log, cb) {
                 if (err) {
                     return next(err, bucketMD);
                 }
-                // based on AWS's behavior, object delete can be performed by
-                // either bucket owner or the object owner
-                if ((objMD && objMD['owner-id'] !== canonicalID)
-                    && bucketMD.getOwner() !== canonicalID) {
-                    return next(errors.AccessDenied);
-                }
+
                 const versioningCfg = bucketMD.getVersioningConfiguration();
                 if (!objMD) {
                     if (!versioningCfg) {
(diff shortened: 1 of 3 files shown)


alexanderchan-scality (dismissed):

lib/api/objectDelete.js
```diff
@@ -22,10 +22,6 @@ const versionIdUtils = versioning.VersionID;
  */
 function objectDelete(authInfo, request, log, cb) {
     log.debug('processing request', { method: 'objectDelete' });
-    if (authInfo.isRequesterPublicUser()) {
```
naren-rajendran: I am still not sure that object delete can be performed by non-owners (public, authenticated and other account access), but based on the behavior we saw, this should be ok.

lib/api/objectDelete.js
```diff
@@ -22,10 +22,6 @@ const versionIdUtils = versioning.VersionID;
...
-        log.debug('operation not available for public user');
-        return cb(errors.AccessDenied);
-    }
```
alexanderchan-scality: should this be addressed in a different ticket (to make cloudserver more like aws)? it seems to be out of scope of the fix, and I believe this may be something that would warrant further discussion.

alexanderchan-scality: would be undoing a part of a related commit that has more to it: https://github.com/scality/cloudserver/commit/12b866dc680df403bb771a59426e9e9e18b101bd

miniscruff: Those 4ish checks seem like things that should be checked in the ACL stuff tbh. I can undo that work and tweak the tests so that the delete does not work anymore and save the rest for later.

lib/api/objectDelete.js
miniscruff: Will undo this change and adjust the tests accordingly

naren-rajendran (approved):

alexanderchan-scality (approved):

PR #3817, opened 2021-08-25
bf: CLDSRV-15 use updated vaultclient

naren-rajendran:
use updated vault client with the fix setting us-east-1 as default region when region is not provided.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "utapi": "scality/utapi#45d7e26",
     "utf8": "~2.1.1",
     "uuid": "^8.3.2",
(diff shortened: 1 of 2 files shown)


anurag4DSB (commented):
There is a conflict.
We need green builds for 8.x as well with the respective 8.x hash of Vaultclient.

nicolas2bert (approved):

alexanderchan-scality (approved):

naren-rajendran:
@bert-e status

naren-rajendran:
@bert-e status

PR #3818, opened 2021-08-26
CLDSRV-16: fix object-lock comparison checks

alexanderchan-scality:
adds checks to avoid referencing properties of an `undefined` retention rule.

### DIFF
--- a/lib/api/apiUtils/object/objectLockHelpers.js
+++ b/lib/api/apiUtils/object/objectLockHelpers.js
@@ -7,6 +7,11 @@ const moment = require('moment');
  */
 function calculateRetainUntilDate(retention) {
     const { days, years } = retention;
+
+    if (!days && !years) {
+        return undefined;
+    }
+
     const date = moment();
     // Calculate the number of days to retain the lock on the object
     const retainUntilDays = days || years * 365;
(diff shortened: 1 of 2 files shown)


lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -67,17 +72,25 @@ function validateHeaders(bucket, headers, log) {
...
+
     if (headers) {
         const headerMode = headers['x-amz-object-lock-mode'];
         const headerDate = headers['x-amz-object-lock-retain-until-date'];
-        const objectRetention = headerMode && headerDate;
-        if (objectRetention || defaultRetention) {
-            const mode = headerMode || defaultRetention.rule.mode;
-            const date = headerDate
-                || calculateRetainUntilDate(defaultRetention.rule);
+        if (headerMode && headerDate) {
```
jonathan-gramain, line 90: Does this change mean that both headers (mode and retain-until-date) must be passed in order to have any effect?

alexanderchan-scality: Yes, both headers must be set for there to be an effect and was the case for the original code. This change is to remove the assignment to a var that was only used once

the behaviors is also tested here: https://github.com/scality/cloudserver/blob/development/7.10/tests/functional/aws-node-sdk/test/object/put.js#L380-L414

jonathan-gramain (approved):

naren-rajendran (approved):

PR #3911, opened 2021-10-05
bugfix: CLDSRV-31 disable KMS healthchecks

jonathan-gramain:
Disable KMS healthchecks introduced with S3C-4562, because when the
KMS is unavailable, Cloudserver would not start, or if already
started, would reject all requests (including unencrypted ones).

The KMS healthchecks will be reworked later, this is a quick fix for now.

### DIFF
--- a/lib/utilities/healthcheckHandler.js
+++ b/lib/utilities/healthcheckHandler.js
@@ -3,7 +3,6 @@ const _config = require('../Config').config;
 const data = require('../data/wrapper');
 const vault = require('../auth/vault');
 const metadata = require('../metadata/wrapper');
-const kms = require('../kms/wrapper');
 const async = require('async');
 
 // current function utility is minimal, but will be expanded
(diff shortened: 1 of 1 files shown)


jonathan-gramain:
ping

lib/utilities/healthcheckHandler.js
```diff
@@ -3,7 +3,7 @@ const _config = require('../Config').config;
 const data = require('../data/wrapper');
 const vault = require('../auth/vault');
 const metadata = require('../metadata/wrapper');
-const kms = require('../kms/wrapper');
+// const kms = require('../kms/wrapper');
```
naren-rajendran: can you please remove the commented out lines? you can grab it from history!

jonathan-gramain: I can remove that one, I prefer to keep `// kms` because it gives some context on the above comment (and IMHO does not hurt in such case, since we want to implement it soon anyway).

naren-rajendran (approved):

mrepop (approved):
Provisional approval so long as we get S3C-4982 completed before delivery.  

The KMS is a in the critical path for system functionality.  IMHO, the health-checks / KMS operational availability is critical.

tmacro (approved):

PR #3923, opened 2021-10-08
bugfix: CLDSRV-30 remove old data keys on CRR target version overwrite

jonathan-gramain:
Remove data keys belonging to the old object when overwriting a CRR target object version with a new object, to avoid creating orphaned data keys.
 
The code checks whether the data keys have changed before removing the old ones, in order to cope with replays of the PUT metadata operation only. Similarly, metadata-only operations do not trigger any data key removal.

- refactor: rename locationKeysSanityCheck to locationKeysHaveChanged

The function name now reflects what it is returning (a boolean) which is clearer in the new context where it will be used. The jsdoc was also improved and fixed.


jonathan-gramain:
ping

rachedbenmustapha (approved):

nicolas2bert (approved):

tmacro (approved):

jonathan-gramain:
ping

PR #3931, opened 2021-10-11
CLDSRV-34 - fix time computation for test comparison

lavigne958:
need a new date that is 30 minutes in the past.
Computation is based on a date already 30 minutes in the future,
take that date and rewind it by 60 minutes
- 30 to go back to the original data
- 30 to go back 30 minutes in the past

# Pull request template

### Motivation and context

This change is required as it brakes a unit test

### DIFF
--- a/tests/unit/api/objectHead.js
+++ b/tests/unit/api/objectHead.js
@@ -17,9 +17,10 @@ const postBody = Buffer.from('I am a body', 'utf8');
 const correctMD5 = 'be747eb4b75517bf6b3cf7c5fbb62f3a';
 const incorrectMD5 = 'fkjwelfjlslfksdfsdfsdfsdfsdfsdj';
 const objectName = 'objectName';
-const date = new Date();
-const laterDate = date.setMinutes(date.getMinutes() + 30);
-const earlierDate = date.setMinutes(date.getMinutes() - 30);
+const laterDate = new Date();
+laterDate.setMinutes(laterDate.getMinutes() + 30);
+const earlierDate = new Date();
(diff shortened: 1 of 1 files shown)


lavigne958:
Hey bert-e

tests/unit/api/objectHead.js
```diff
@@ -19,7 +19,7 @@ const incorrectMD5 = 'fkjwelfjlslfksdfsdfsdfsdfsdfsdj';
 const objectName = 'objectName';
 const date = new Date();
 const laterDate = date.setMinutes(date.getMinutes() + 30);
-const earlierDate = date.setMinutes(date.getMinutes() - 30);
+const earlierDate = date.setMinutes(date.getMinutes() - 60);
```
miniscruff: I think the problem here is that date is being changed with each method as setMinutes updates in place.
You probably want to do `laterDate = new Date().setMinutes(...);` and so on.

miniscruff (changes requested):

tests/unit/api/objectHead.js
miniscruff: Actually on second thought I think just doing two calls is better.
```js
const earlierDate = new Date();
earlierDate.setMinutes(...);
```

lavigne958: Right it actually looks easier to read and understand.

miniscruff (approved):

XinLi0207 (approved):

lavigne958:
squashed commits and rebased PR.

PR #3948, opened 2021-10-14
[BACKPORT 7.4] bugfix: S3C-2775 update logic for counting object replicas

jonathan-gramain:
Before, we were relying on the presence of ACL change to know if the
operation is metadata-only. This was not correct, as a new object can
be put with specific ACL too.

Instead, consider that if there is already an object present with the
same version, we are necessarily updating this object hence the object
counter should not be updated.

NOTE: this change will come along with another change in backbeat: the
PUT /_/backbeat/metadata route will be provided the version ID of the
object to PUT, so that the validation of the existing object (through
the metadataValidateBucketAndObj helper) checks the specific version
instead of checking the master key (which for backbeat replication
purpose is also incorrect). Without this backbeat change, the
replicated object count metric becomes wrong when an object is
rewritten.

(cherry picked from commit 69e6d678a010dc8db1cf171a83d08430991369d3)

# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/routes/utilities/pushReplicationMetric.js
+++ b/lib/routes/utilities/pushReplicationMetric.js
@@ -1,5 +1,3 @@
-const assert = require('assert');
-
 const { ObjectMD } = require('arsenal').models;
 const { pushMetric } = require('../../utapi/utilities');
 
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
ping

jonathan-gramain:
ping

tmacro (approved):

naren-rajendran (approved):

PR #3975, opened 2021-10-22
Allow configuring CLDSRV-41 capability

francoisferrand:
Issue: CLDSRV-41
Issue: ARTESCA-2190

### DIFF
--- a/lib/utilities/reportHandler.js
+++ b/lib/utilities/reportHandler.js
@@ -34,6 +34,7 @@ function hasWSOptionalDependencies() {
 }
 
 function getCapabilities() {
+    const localVolumeCap = process.env.LOCAL_VOLUME_CAPABILITY;
     return {
         locationTypeDigitalOcean: true,
         locationTypeS3Custom: true,
@@ -41,6 +42,7 @@ function getCapabilities() {
         locationTypeNFS: true,
         locationTypeCephRadosGW: true,
         locationTypeHyperdriveV2: true,
+        locationTypeLocal: localVolumeCap === undefined ? true : JSON.parse(localVolumeCap),
         preferredReadLocation: true,
         managedLifecycle: true,
         managedLifecycleTransition: true,
--- a/lib/utilities/reportHandler.js
+++ b/lib/utilities/reportHandler.js
@@ -34,7 +34,7 @@ function hasWSOptionalDependencies() {
 }
 
 function getCapabilities() {
-    const localVolumeCap = process.env.LOCAL_VOLUME_CAPABILITY;
+    const localVolumeCap = process.env.LOCAL_VOLUME_CAPABILITY || 'true';
     return {
         locationTypeDigitalOcean: true,
         locationTypeS3Custom: true,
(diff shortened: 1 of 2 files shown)


francoisferrand:
fixed version in ticket

tmacro (approved):

lib/utilities/reportHandler.js
```diff
@@ -34,13 +34,15 @@ function hasWSOptionalDependencies() {
 }
 
 function getCapabilities() {
+    const localVolumeCap = process.env.LOCAL_VOLUME_CAPABILITY;
```
miniscruff: Not a fan of using JSON.parse to get a bool, I would just do a string check with two changes.
```suggestion
    const localVolumeCap = process.env.LOCAL_VOLUME_CAPABILITY || 'true';
```
Default to a true string

lib/utilities/reportHandler.js
```diff
@@ -34,13 +34,15 @@ function hasWSOptionalDependencies() {
...
     return {
         locationTypeDigitalOcean: true,
         locationTypeS3Custom: true,
         locationTypeSproxyd: true,
         locationTypeNFS: true,
         locationTypeCephRadosGW: true,
         locationTypeHyperdriveV2: true,
+        locationTypeLocal: localVolumeCap === undefined ? true : JSON.parse(localVolumeCap),
```
miniscruff: No more need to check undefined just compare the string, I always add a toLower/toUpper to prevent some typos
```suggestion
        locationTypeLocal: localVolumeCap.toLowerCase() === 'true',
```

tests/unit/management/secureChannel.js
```diff
@@ -11,9 +11,21 @@ describe('report handler', () => {
         assert.strictEqual(c.locationTypeS3Custom, true);
         assert.strictEqual(c.locationTypeSproxyd, true);
         assert.strictEqual(c.locationTypeHyperdriveV2, true);
+        assert.strictEqual(c.locationTypeLocal, true);
         assert.strictEqual(c.preferredReadLocation, true);
         assert.strictEqual(c.managedLifecycle, true);
         assert.strictEqual(c.secureChannelOptimizedPath, true);
         assert.strictEqual(c.s3cIngestLocation, true);
     });
+    it('should allow configure local file system capability', () => {
```
miniscruff: I would make this multiple tests, and also test a completely random string value like 'not-a-bool' to determine what happens ( you never know ).

miniscruff (changes requested):

lib/utilities/reportHandler.js
alexanderchan-scality: can this just be set to a `bool` without the parsing?
also would prefer a specific value vs all truthy

```suggestion
    const localVolumeCap = process.env.LOCAL_VOLUME_CAPABILITY === '1'; // or 'true'
```"

miniscruff: yea but I prefer to be a little more lenient to account for capitals such that 'True' and 'true' are both ok, some people like TRUE or true for values. I know python has a isTruthyVar or something that supports most common true and false values like true, True, 1, yes, y, etc. Could also work. Found it: https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

alexanderchan-scality: a truthy set can work too. the specific case I wanted to avoid was '0', which would resolve to true.
~I believe in this project, we are using `0` or `1`~ (it's a mix)

miniscruff: 0 would still be false when you take into account my other suggestion of `.toLowerCase() === 'true'`

miniscruff (approved):

PR #4025, opened 2021-11-04
Expose cloudserver metrics on separate port and /metrics path

francoisferrand:
This reduces the attack surface, and ensure we do not leak (metrics) information via the `_/monitoring/metrics` path.

Issue: CLDSRV-51

### DIFF
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -161,11 +161,11 @@ function routeHandler(req, res, log, cb) {
  * Checks if client IP address is allowed to make http request to
  * S3 server. Defines function 'montiroingEndHandler', which is
  * called if IP not allowed or passed as callback.
- * @param {object} clientIP - IP address of client
- * @param {object} req - http request object
- * @param {object} res - http response object
- * @param {object} log - werelogs logger instance
- * @return {undefined}
+ * @param {string | undefined} clientIP - IP address of client
+ * @param {http.IncomingMessage} req - http request object
+ * @param {http.ServerResponse} res - http response object
+ * @param {RequestLogger} log - werelogs logger instance
+ * @return {void}
  */
 function monitoringHandler(clientIP, req, res, log) {
     function monitoringEndHandler(err, results) {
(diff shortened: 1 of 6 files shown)


jbertran (approved):

lavigne958:
~you don't have any ticket number in your PR/commits :thinking: 
you include commits from CLDSRV-51 from @jbertran 
should this PR need to cleaning first ? wait for the CLDSRV-51 to pass first ?~
got it, sorry

lavigne958 (approved):

lib/utilities/monitoringHandler.js
```diff
@@ -167,7 +179,7 @@ function monitoringHandler(clientIP, req, res, log) {
     if (req.method !== 'GET') {
         return monitoringEndHandler(res, errors.MethodNotAllowed);
     }
-    const monitoring = (req.url === '/_/monitoring/metrics');
+    const monitoring = (req.url === '/_/metrics');
```
alexanderchan-scality: small change: 
the constant defined here can be used.
https://github.com/scality/Arsenal/blob/development/8.1/lib/network/probe/ProbeServer.js#L7

alexanderchan-scality (approved):

nicolas2bert (approved):

lib/utilities/monitoringHandler.js
francoisferrand: This is pending discussion on Confluence: https://scality.atlassian.net/wiki/spaces/OS/pages/1819148292/Strategy+for+exposing+metrics (where we consider using a simpler `/metrics` path, since we are already moved to a different port)

Also, importing from Arsenal does not seem look so appropriate with the current Arsenal implementation : this constant is present indeed, but deep inside the code of ProbeServer, and looks more like an implementation detail than an "established" constant we should rely on.

miniscruff: The constant was meant to be relied upon regardless of usage to probe server, as probe server itself is optional for metrics/probes as your service may already have an established server. It is of course not "established" because it is new and only used a couple of times in backbeat as that is where it was initialized built for.

francoisferrand: in that case, should we not move the constant in arsenal's `constants` ?

miniscruff: Uh maybe? Our use of constant.js files is a little confusing to me and I tend to go with more contextual constants rather than one mega file with everything under the sun. If you were discouraged to use the value because it was not placed in a constants file I guess we should move it. I got discouraged when I would use backbeat and there are multiple constants ( one in lib/ and one in each extension ).

PR #4060, opened 2021-11-12
Feature/cldsrv 55/utapi event filtering

tmacro:


tests/utapi/awsNodeSdk.js
```diff
@@ -471,4 +471,26 @@ describe('utapi v2 metrics incoming and outgoing bytes', function t() {
             next => deleteBucket(bucket, next),
         ], done);
     });
+
+    it('should not push a metric for a filtered bucket', done => {
+        const bucket = 'utapi-event-filter-deny-bucket';
+        const objectSize = 1024 * 1024;
+        const objSize = objectSize * 2;
```
jonathan-gramain:
```suggestion
        const objSize = 2 * 1024 * 1024;
```

tests/utapi/awsNodeSdk.js
```diff
@@ -471,4 +471,26 @@ describe('utapi v2 metrics incoming and outgoing bytes', function t() {
...
+        const key = '1.txt';
+        async.series([
+            next => createBucket(bucket, next),
+            next => putObject(bucket, key, objSize, next),
+            next => wait(WAIT_MS, next),
+            next => wait(WAIT_MS, () => {
```
jonathan-gramain:
```suggestion
            next => wait(2 * WAIT_MS, () => {
```
or (if WAIT_MS is enough)

```suggestion
            next => wait(WAIT_MS, () => {
```

eve/main.yml
```diff
@@ -305,6 +305,7 @@ stages:
         env:
           ENABLE_UTAPI_V2: t
           S3BACKEND: mem
+          BUCKET_DENY_FILTER: utapi-event-filter-deny-bucket
```
jonathan-gramain, line 308: I'd suggest having two buckets in this list and check each of them in turn in the tests, to cover parsing of the env var for multiple buckets

jonathan-gramain (approved):

eve/main.yml
jonathan-gramain:
```suggestion
          UTAPI_BUCKET_DENY_FILTER: utapi-event-filter-deny-bucket
```

tmacro: I don't think it will support multiple buckets without some more sed magic. This was only added for test setup, so maybe somewhere else would be easier to add multiple buckets. I'll see what I can do

jonathan-gramain: Ah, I though we were setting this variable in the same way than in the production environment, if it's just a test helper, then no need to change it since it would not increase the test coverage.

naren-rajendran (approved):

PR #4081, opened 2021-11-20
improvement: CLDSRV-60 cleanup replay IDs of null versions

jonathan-gramain:
* cleanup replay IDs of null versions

Replace the previous way of passing replayId option to metadata DELETE by the result of the preprocessingVersioningDelete function, since it handles more cases like when a bucket with versioning suspended has to delete the previous null version, in which case, the null version's replay IDs have to be cleaned up.

* process upload IDs in versioning helpers

In versioning helpers, make sure we pass on the nullUploadId/replayId to the metadataStoreObject() or deleteObject() function, so that they can update ObjectMD and pass the appropriate replay options to metadata

### DIFF
--- a/tests/unit/api/apiUtils/versioning.js
+++ b/tests/unit/api/apiUtils/versioning.js
@@ -44,6 +44,24 @@ describe('versioning helpers', () => {
                 },
             },
             {
+                // prior MPU object non-null version exists
+                objMD: {
+                    versionId: 'v1',
+                    uploadId: 'fooUploadId',
+                },
+                versioningEnabledExpectedRes: {
+                    options: {
+                        versioning: true,
+                    },
+                },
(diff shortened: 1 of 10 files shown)


watsaqat (approved):

tests/unit/api/apiUtils/versioning.js
```diff
@@ -43,6 +43,24 @@ describe('versioning helpers', () => {
                     },
                 },
             },
+            {
+                // prior MPU object non-null version exists
```
watsaqat: instead of commenting the description of the test case, it might be better to do:
`{ description: 'prior MPU object non-null version exists }`
and incorporate this into `it('..${ testCase.description }..'` in forEach

tmacro (approved):

PR #4086, opened 2021-11-22
bf(CLDSRV-66): Don't pass `oldByteLength` for completeMPU if versioned bucket

tmacro:


lib/api/completeMultipartUpload.js
```diff
@@ -447,7 +447,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
         xmlParams.eTag = `"${aggregateETag}"`;
         const xml = convertToXml('completeMultipartUpload', xmlParams);
         pushMetric('completeMultipartUpload', log, {
-            oldByteLength,
+            oldByteLength: !generatedVersionId ? oldByteLength : null,
```
jonathan-gramain: I think that checking whether the backend generated a version ID is not enough to know if we created a new version *without* removing an old version, due to the "null" version handling on versioning-suspended buckets.

Looking at objectPut, this is what is done, before checking `isVersionedObj` to tell whether to pass `oldByteLength` or not:
```
            const vcfg = bucket.getVersioningConfiguration();
            const isVersionedObj = vcfg && vcfg.Status === 'Enabled';
```

jonathan-gramain (changes requested):

lib/api/completeMultipartUpload.js
jonathan-gramain: At the end of `services.metadataStoreObject`, we have this, which tends to confirm that a null version has a versionID (`'null'`) returned.

```
            // if versioning is enabled, data will be returned from metadata
            // as JSON containing a versionId which some APIs will need sent
            // back to them
            let versionId;
            if (data) {
                if (params.isNull && params.isDeleteMarker) {
                    versionId = 'null';
                } else if (!params.isNull) {
                    versionId = JSON.parse(data).versionId;
                }
            }
            return cb(err, { contentMD5, versionId });
```

jonathan-gramain: Actually, only sends `null` versionId if a delete marker, read too quickly :confused: 

If you can check this anyway, and feel free to keep or amend the code accordingly, I'm not sure I understand this logic fully but it can be tricky so worth checking what happens when null versions are deleted or overwritten.

jonathan-gramain (approved):
Approving but please check correctness regarding null version overwrite

lib/api/completeMultipartUpload.js
tmacro: Did some checking, for version suspended buckets the `generatedVersionId` here is `undefined`.
I'll swap to the other method though to keep the logic consistent.

rachedbenmustapha (approved):

PR #4094, opened 2021-11-24
Add http metrics

francoisferrand:
Issue: CLDSRV-63

### DIFF
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -38,6 +38,12 @@ const httpRequestsTotal = new client.Counter({
     name: 'http_requests_total',
     help: 'Total number of HTTP requests',
 });
+const httpRequestDurationSeconds = new client.Histogram({
+    name: 'http_request_duration_seconds',
+    help: 'Duration of HTTP requests in seconds',
+    labelNames,
+    buckets: [0.0001, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5], // buckets for response time from 0.1ms to 500ms
+});
 
 const httpRequestSizeBytes = new client.Summary({
     labelNames,
@@ -192,4 +198,5 @@ module.exports = {
     collectDefaultMetrics,
     promMetrics,
     crrCacheToProm,
+    httpRequestDurationSeconds,
 };
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -32,7 +32,7 @@ const dataDiskTotal = new client.Gauge({
     help: 'Total data disk storage in bytes',
 });
 
-const labelNames = ['method', 'route', 'code'];
+const labelNames = ['method', 'action', 'code'];
 const httpRequestsTotal = new client.Counter({
     labelNames,
     name: 'http_requests_total',
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -177,7 +177,7 @@ async function routeHandler(req, res, cb) {
  * @param {RequestLogger} log - werelogs logger instance
  * @return {void}
  */
-async function monitoringHandler(clientIP, req, res, log) {
+function monitoringHandler(clientIP, req, res, log) {
     function monitoringEndHandler(err, results) {
         writeResponse(res, err, results, error => {
             if (error) {
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -155,14 +155,16 @@ function writeResponse(res, error, results, cb) {
 }
 
 
-function routeHandler(req, res, cb) {
+async function routeHandler(req, res, cb) {
     if (req.method !== 'GET') {
         return cb(errors.BadRequest, []);
     }
-    const promMetrics = client.register.metrics();
+    const promMetrics = await client.register.metrics();
     const contentLen = Buffer.byteLength(promMetrics, 'utf8');
-    res.setHeader('content-length', contentLen);
-    res.setHeader('content-type', client.register.contentType);
+    res.writeHead(200, {
+        'Content-Length': contentLen,
+        'Content-Type': client.register.contentType
+    });
     res.end(promMetrics);
     return undefined;
 }
@@ -177,7 +179,7 @@ function routeHandler(req, res, cb) {
  * @param {RequestLogger} log - werelogs logger instance
  * @return {void}
  */
-function monitoringHandler(clientIP, req, res, log) {
+async function monitoringHandler(clientIP, req, res, log) {
     function monitoringEndHandler(err, results) {
         writeResponse(res, err, results, error => {
             if (error) {
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -32,40 +32,40 @@ const dataDiskTotal = new client.Gauge({
     help: 'Total data disk storage in bytes',
 });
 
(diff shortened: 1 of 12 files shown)


tests/functional/report/monitoring.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    async function getMetrics() {
+        return new Promise(resolve => {
+            http.get({ host: conf.ipAddress, path: '/metrics', port: 8002 }, res => {
+                assert.strictEqual(res.statusCode, 200);
+
+                let body = '';
+                res.on('data', chunk => { body += chunk; });
+                res.on('end', () => resolve(body));
```
miniscruff: Does not really matter much for tests but I recommend using an array and push, with a join in end. Slightly more efficient. We also use this pattern a lot so it would be consistent. 

```suggestion
                let body = [];
                res.on('data', chunk => { body.push(chunk); });
                res.on('end', () => resolve(body.join('')));
```

tests/functional/report/monitoring.js
```diff
@@ -0,0 +1,106 @@
...
+            });
+        });
+    }
+
+    function parseMetric(metrics, name, labels) {
+        const labelsString = Object.entries(labels !== undefined ? labels : {
+            method: 'GET', code: '200', route: 'healthcheck',
+        }).map(e => `${e[0]}="${e[1]}"`).join(',');
```
miniscruff, line 33:
```suggestion
        }).map(e => e.join('=')).join(',');
```

tests/functional/report/monitoring.js
```diff
@@ -0,0 +1,106 @@
...
+        ['/_/healthcheck/deep',
+                            { method: 'GET', code: '200', route: 'deepHealthcheck' }],
+        ['/_/report',       { method: 'GET', code: '200', route: 'report' }],
+        ['/_/backbeat',     { method: 'GET', code: '405', route: 'routeBackbeat' }],
+        ['/_/metadata',     { method: 'GET', code: '403', route: 'routeMetadata' }],
+        ['/_/workflow-engine-operator',
+                            { method: 'GET', code: '405', route: 'routeWorkflowEngineOperator' }],
+    ].forEach(testcase => {
+        const path = testcase[0];
+        const labels = testcase[1];
```
miniscruff: Does this work in node 10?
```suggestion
    ].forEach([path, labels] => {
```

tests/functional/report/monitoring.js
```diff
@@ -0,0 +1,106 @@
...
+        it(`it should count http ${labels.method} requests metrics on ${path}`, async () => {
```
miniscruff: You don't need to include it in descriptions.
```suggestion
        it(`should count http ${labels.method} requests metrics on ${path}`, async () => {
```

tests/unit/utils/monitoring.js
```diff
@@ -0,0 +1,69 @@
...
+
+    it('it should return an error is path is not /metrics', async () => {
+        await fetchMetrics({ method: 'GET', url: '/foo' }, res);
+        assert(res.writeHead.calledOnceWith(405));
+        assert(res.end.calledOnce);
+    });
+
+    it('it should return some metrics', async () => {
+        await fetchMetrics({ method: 'GET', url: '/metrics' }, res);
+        assert(res.writeHead.calledOnceWith(200) || !res.writeHead.called);
```
miniscruff: using `||` for assertions feels weird, we should know which one it is

tests/unit/utils/monitoring.js
```diff
@@ -0,0 +1,69 @@
...
+        assert(res.end.calledOnce);
+
+        // Check that some "system" metrics is present
+        assert(res.end.args[0][0].includes('\nnodejs_active_handles_total '));
+    });
+
+    it('it should have http duration histogram metrics', async () => {
```
miniscruff, line 52: I feel like these last 3 tests can be another table test

lib/utilities/monitoringHandler.js
```diff
@@ -167,23 +177,23 @@ function routeHandler(req, res, log, cb) {
  * @param {RequestLogger} log - werelogs logger instance
  * @return {void}
  */
-function monitoringHandler(clientIP, req, res, log) {
+async function monitoringHandler(clientIP, req, res, log) {
```
ghost: If we don't await any of the returned/called function, why do we want this function to be async?

ghost (commented):
Quick review

tests/functional/report/monitoring.js
francoisferrand: if you also put parenthesis `([path, labels])`, then it works indeed.
thanks!

tests/unit/utils/monitoring.js
francoisferrand: i agree, though i did this to match the behavior of http module... (and thus not depend on the actual implementation).
But I fixed the code to explicitely call `writeHead`, so no reason to keep it indeed.

lib/utilities/monitoringHandler.js
francoisferrand: it calls `routeHandler()`, which is async because it calls/awaits' promClient's `metrics()` function...
but it cannot be awaited anyway, i'll just try to remote it.

tests/functional/report/monitoring.js
francoisferrand: This does not work, we need to add the double-quotes in the string

lib/utilities/monitoringHandler.js
ghost: Unless you are awaiting an async function, you should be able to call async function from sync ones, it will behave as expected :smile:

tests/unit/utils/monitoring.js
francoisferrand: i fear this may increase the tests duration & flakiness...

lib/utilities/monitoringHandler.js
```diff
@@ -44,6 +44,10 @@ const httpRequestDurationSeconds = new client.Histogram({
     labelNames,
     buckets: [0.0001, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5], // buckets for response time from 0.1ms to 500ms
 });
+const httpActiveRequests = new client.Gauge({
+    name: 'http_active_requests',
+    help: 'Total number of HTTP requests',
```
lavigne958: This is miss-leading, it should be something that explicitly describe the value is a gauge and only contains currently active requests.

lib/utilities/monitoringHandler.js
```diff
@@ -169,21 +179,21 @@ function routeHandler(req, res, log, cb) {
  */
 function monitoringHandler(clientIP, req, res, log) {
     function monitoringEndHandler(err, results) {
-        writeResponse(res, err, log, results, error => {
+        writeResponse(res, err, results, error => {
```
ghost, line 182: Is this function called with a `results` parameter that is not equal to `[]`? The `routeHandler` function never calls this function if there is no error.

lib/utilities/monitoringHandler.js
francoisferrand: Bad copy-paste indeed, good catch! 👍

lavigne958 (approved):

lib/utilities/monitoringHandler.js
francoisferrand: correct, I don't think it is really used at the moment.
I don't know why it's there, but there may be a reason why it was added, i'd prefer not rushing to clean this thing..

ghost (approved):

francoisferrand:
fixed version in ticket

PR #4116, opened 2021-12-09
Add cloudserver alerting rules

francoisferrand:
The alerts will be deployed by Zenko-Operator as PrometheusRule objects, after "resolving" the input variables.

Issue: CLDSRV-70

### DIFF
--- a/monitoring/alerts.yaml
+++ b/monitoring/alerts.yaml
@@ -1,0 +1,91 @@
+# Variables which should be replaced. Similar to grafana dashboards' __inputs section
+x-inputs:
+  - name: namespace
+    type: constant
+    value: zenko
+  - name: service
+    type: constant
+    value: artesca-data-connector-s3api-metrics
+
+groups:
+- name: CloudServer
+  rules:
+
+  # As a platform admin I want to be alerted (warning) when the system errors are more than 3% of
+  # all the response codes
+  - alert: SystemErrorsWarning
+    expr: |
+      sum by(pod) (rate(http_requests_total{namespace="${namespace}", service="${service}", code=~"5.."}[1m]))
+          / sum by(pod) (rate(http_requests_total{namespace="${namespace}", service="${service}"}[1m]))
+        >= 0.03
+    for: 5m
+    labels:
+      severity: warning
+    annotations:
+      summary: "High ratio of system erors on {{ $labels.pod }}"
+
+  # As a platform admin I want to be alerted (critical) when the system errors are more than 5% of
+  # all the response codes
+  - alert: SystemErrorsCritical
+    expr: |
+      sum by(pod) (rate(http_requests_total{namespace="${namespace}", service="${service}", code=~"5.."}[1m]))
+          / sum by(pod) (rate(http_requests_total{namespace="${namespace}", service="${service}"}[1m]))
+        >= 0.05
+    for: 5m
+    labels:
+      severity: critical
+    annotations:
+      summary: "Very high ratio of system erors on {{ $labels.pod }}"
+
+  # As a platform admin I want to be alerted (warning) when a listing operation latency or a
+  # version listing operation latency is more than 300ms
+  - alert: ListingLatencyWarning
+    expr: |
+      rate(http_request_duration_seconds_sum{namespace="${namespace}",service="${service}",action="getService|listBucket"}[1m])
+          / rate(http_request_duration_seconds_count{namespace="${namespace}",service="${service}",action="getService|listBucket"}[1m])
+        >= 0.300
+    for: 5m
+    labels:
+      severity: warning
+    annotations:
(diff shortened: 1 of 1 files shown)


monitoring/alerts.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - alert: SystemErrorsWarning
+    expr: |
+      sum (http_requests_total{namespace="${namespace}", service="${service}", code=~"5.."})
+          / sum (http_requests_total{namespace="${namespace}", service="${service}"})
+        >= 0.03
+    for: 5m
+    labels:
+      severity: warning
+    annotations:
+      summary: "High ratio of system erors on {{ $labels.pod }}"
```
lavigne958, line 25: Be careful if you apply a `sum(...)` you might loose you labels which makes that message incomplete

monitoring/alerts.yaml
```diff
@@ -0,0 +1,91 @@
...
+  # As a platform admin I want to be alerted (critical) when the system errors are more than 5% of
+  # all the response codes
+  - alert: SystemErrorsCritical
+    expr: |
+      sum (http_requests_total{namespace="${namespace}", service="${service}", code=~"5.."})
+          / sum (http_requests_total{namespace="${namespace}", service="${service}"})
+        >= 0.05
+    for: 5m
+    labels:
+      severity: warning
```
lavigne958: severity should be `critical` here

monitoring/alerts.yaml
```diff
@@ -0,0 +1,91 @@
...
+  # As a platform admin I want to be alerted (critical) when a delete operation latency is more
+  # than 1s
+  - alert: DeleteLatencyCritical
+    expr: |
+      rate(http_request_duration_seconds_sum{namespace="${namespace}",service="${service}",action="deleteObject|deleteBucket|multiObjectDelete"}[1m])
+          / rate(http_request_duration_seconds_count{namespace="${namespace}",service="${service}",action="deleteObject|deleteBucket|multiObjectDelete"}[1m])
+        >= 1.000
+    for: 5m
+    labels:
+      severity: warning
```
lavigne958: should be `critical`

monitoring/alerts.yaml
francoisferrand: good catch, thanks

monitoring/alerts.yaml
francoisferrand: I need to test this indeed, it is not clear to me how this would work... In "EXPLORE" i can choose to get either a series (with labels it seems) or a single value, and I don't know how it goes with alerts ; but it labels get removed, I don't see the point having these Go templates in the first place :-/

Anyway, I will look into this.

lavigne958: if you `expr` is based on a metric with some labels you can get the labels in the description. if you expression results in a value with no labels (like in the case of a `sum(...)`) you loose all your labels. you have access to all the labels you can see if you put your expression in the prometheus UI or in grafana

francoisferrand: Ok, should work with `sum by(pod)`

monitoring/alerts.yaml
```diff
@@ -0,0 +1,91 @@
...
+- name: CloudServer
+  rules:
+
+  # As a platform admin I want to be alerted (warning) when the system errors are more than 3% of
+  # all the response codes
+  - alert: SystemErrorsWarning
+    expr: |
+      sum by(pod) (rate(http_requests_total{namespace="${namespace}", service="${service}", code=~"5.."}[1m]))
+          / sum by(pod) (rate(http_requests_total{namespace="${namespace}", service="${service}"}[1m]))
+        >= 0.03
```
lavigne958, line 20: be aware you are looking at the rate of a counter, and your sampling is over 1 minute (the `[1m]`) so the resulting graph can be very _spiky_ which will make the alert trigger from standby mode to firing mode very often without over-passing the threshold of 5 minutes.

 I am wondering (I have no clue if it is better) if we should have the rate sampling aligned with the `for` condition :thinking:

francoisferrand: we are not actually looking at the rate : since we are dividing rate by rate, this is just a mean to "scope" the computation to the last minute only. The point is indeed to have it spiky, e.g. measure "instantaneous error rate", and than "filter" spikes by adding the duration threshold.

I have been wondering the same, but I don't think there is an actual reason to use the same value for averaging and filtering. That said, I have no idea what the values of these 2 parameters should be.

The `5m` filter is what seems to be used in Ring ; and the `1m` is the value that pops' up when looking for this kind "error ratios" on google. But I have no science to backup these...

lavigne958: :laughing:  let's keep it this way. sounds good to me :-)

lavigne958 (approved):

jbertran (approved):

PR #4119, opened 2021-12-13
Feature/cldsrv-65 create dashboard

lavigne958:
Create and deploy dashboards for grafana.
CLDSRV-65

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -498,6 +498,19 @@ stages:
           command: docker push ${DEVELOPMENT_DOCKER_IMAGE_NAME}
           haltOnFailure: true
           env: *docker_env
+      - ShellCommand:
+          name: Oras login
+          command: |
+            oras login --username "${HARBOR_LOGIN}" --password "${HARBOR_PASSWORD}" registry.scality.com
+          env:
+            HARBOR_LOGIN: '%(secret:harbor_login)s'
+            HARBOR_PASSWORD: '%(secret:harbor_password)s'
+      - ShellCommand:
+          name: push dashboards to the development namespace
+          command: |
+            ./push-to-registry.sh cloudserver-dashboard:latest dashboard.json "application/grafana-dashboard+json"
+          workdir: build/monitoring/
+
 
   # This stage can be used to release your Docker image.
   # To use this stage:
--- a/eve/workers/docker/Dockerfile
+++ b/eve/workers/docker/Dockerfile
@@ -18,8 +18,16 @@ RUN yum install -y epel-release \
     python36-pip \
     git \
     docker-ce-cli-18.09.6 \
+    which \
     && adduser -u 1042 --home /home/eve eve --groups docker \
     && chown -R eve:eve /home/eve \
     && pip3 install buildbot-worker==${BUILDBOT_VERSION}
+
+
+ARG ORAS_VERSION=0.12.0
+RUN curl -LO https://github.com/oras-project/oras/releases/download/v${ORAS_VERSION}/oras_${ORAS_VERSION}_linux_amd64.tar.gz && \
+    mkdir -p oras-install/ && \
+    tar -zxf oras_${ORAS_VERSION}_*.tar.gz -C /usr/local/bin oras && \
+    rm -rf oras_${ORAS_VERSION}_*.tar.gz oras-install/
 
 CMD buildbot-worker create-worker . ${BUILDMASTER}:${BUILDMASTER_PORT} ${WORKERNAME} ${WORKERPASS} && buildbot-worker start --nodaemon
--- a/monitoring/dashboard.json
+++ b/monitoring/dashboard.json
@@ -1,0 +1,614 @@
+{
+    "__inputs": [
+        {
+            "name": "DS_PROMETHEUS",
+            "label": "Prometheus",
+            "description": "",
+            "type": "datasource",
+            "pluginId": "prometheus",
+            "pluginName": "Prometheus"
+        },
+        {
+            "name": "namespace",
+            "type": "constant",
+            "label": "namespace",
+            "description": "Namespace associated with the Zenko instance",
+            "value": "zenko"
+        },
+        {
+            "name": "job",
+            "type": "constant",
+            "label": "job",
+            "description": "Name of the Cloudserver job, used to filter only the Cloudserver instances.",
+            "value": "artesca-data-connector-s3api-metrics"
+        }
+    ],
+    "editable": true,
+    "gnetId": null,
+    "graphTooltip": 0,
+    "id": 33,
+    "links": [],
+    "panels": [
+        {
+            "collapsed": false,
+            "datasource": null,
+            "gridPos": {
+                "h": 1,
+                "w": 24,
+                "x": 0,
+                "y": 0
+            },
+            "id": 2,
+            "panels": [],
+            "title": "Response codes",
+            "type": "row"
+        },
+        {
+            "datasource": "${DS_PROMETHEUS}",
+            "fieldConfig": {
+                "defaults": {
+                    "color": {
+                        "mode": "palette-classic"
+                    },
+                    "custom": {
+                        "axisLabel": "",
+                        "axisPlacement": "auto",
+                        "barAlignment": 0,
+                        "drawStyle": "line",
+                        "fillOpacity": 0,
+                        "gradientMode": "none",
+                        "hideFrom": {
+                            "legend": false,
+                            "tooltip": false,
+                            "viz": false
+                        },
+                        "lineInterpolation": "linear",
+                        "lineWidth": 1,
+                        "pointSize": 5,
+                        "scaleDistribution": {
+                            "type": "linear"
+                        },
+                        "showPoints": "auto",
+                        "spanNulls": false,
+                        "stacking": {
+                            "group": "A",
+                            "mode": "none"
+                        },
+                        "thresholdsStyle": {
+                            "mode": "off"
+                        }
+                    },
+                    "mappings": [],
+                    "thresholds": {
+                        "mode": "absolute",
+                        "steps": [
+                            {
+                                "color": "green",
+                                "value": null
+                            },
+                            {
+                                "color": "red",
+                                "value": 80
+                            }
+                        ]
+                    }
+                },
+                "overrides": []
+            },
+            "gridPos": {
+                "h": 8,
+                "w": 10,
+                "x": 0,
+                "y": 1
+            },
+            "id": 4,
+            "options": {
+                "legend": {
+                    "calcs": [],
+                    "displayMode": "list",
+                    "placement": "bottom"
+                },
+                "tooltip": {
+                    "mode": "single"
+                }
+            },
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum by (code) (rate(http_requests_total{namespace=\"${namespace}\", job=\"${job}\"}[$__rate_interval]))",
+                    "interval": "",
+                    "legendFormat": "{{code}}",
+                    "refId": "A"
+                }
+            ],
+            "title": "response codes",
+            "type": "timeseries"
+        },
+        {
+            "datasource": "${DS_PROMETHEUS}",
+            "fieldConfig": {
+                "defaults": {
+                    "color": {
+                        "mode": "thresholds"
+                    },
+                    "mappings": [],
+                    "max": 100,
+                    "min": 0,
+                    "thresholds": {
+                        "mode": "percentage",
+                        "steps": [
+                            {
+                                "color": "red",
+                                "value": null
+                            },
+                            {
+                                "color": "orange",
+                                "value": 80
+                            },
+                            {
+                                "color": "green",
+                                "value": 90
+                            }
+                        ]
+                    },
+                    "unit": "percent"
+                },
+                "overrides": []
+            },
+            "gridPos": {
+                "h": 8,
+                "w": 4,
+                "x": 10,
+                "y": 1
+            },
+            "id": 8,
+            "options": {
+                "orientation": "auto",
+                "reduceOptions": {
+                    "calcs": [
+                        "lastNotNull"
+                    ],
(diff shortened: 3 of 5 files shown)


eve/main.yml
```diff
@@ -498,6 +498,19 @@ stages:
...
+          name: Oras login
+          command: |
+            oras login --username "${HARBOR_LOGIN}" --password "${HARBOR_PASSWORD}" registry.scality.com
+          env:
+            HARBOR_LOGIN: '%(secret:harbor_login)s'
+            HARBOR_PASSWORD: '%(secret:harbor_password)s'
+      - ShellCommand:
+          name: push dashboards to the development namespace
+          command: |
+            ./push-to-registry.sh cloudserver-dashboard:latest dashboard.json "application/grafana-dashboard+json"
```
francoisferrand, line 511: looking at the script: not sure it was really worth the effort, calling `oras` directly is just as short/long as calling the script...

eve/workers/docker/Dockerfile
```diff
@@ -18,8 +18,16 @@ RUN yum install -y epel-release \
...
     docker-ce-cli-18.09.6 \
+    which \
     && adduser -u 1042 --home /home/eve eve --groups docker \
     && chown -R eve:eve /home/eve \
     && pip3 install buildbot-worker==${BUILDBOT_VERSION}
 
+
+RUN curl -LO https://github.com/oras-project/oras/releases/download/v0.12.0/oras_0.12.0_linux_amd64.tar.gz && \
+    mkdir -p oras-install/ && \
+    tar -zxf oras_0.12.0_*.tar.gz -C oras-install/ && \
```
francoisferrand: you should extract the version with a (dockerfile) buildarg, to make easier to update the version & avoid repetition:

```
ARG ORAS_VERSION=0.12.0
RUN curl -LO https://github.com/oras-project/oras/releases/download/v${ORAS_VERSION}/oras_${ORAS_VERSION}_linux_amd64.tar.gz && \
    mkdir -p oras-install/ && \
    tar -zxf oras_${ORAS_VERSION}_*.tar.gz -C oras-install/ && \
    mv oras-install/oras /usr/local/bin/ && \
    rm -rf oras_${ORAS_VERSION}_*.tar.gz oras-install/

File: eve/workers/docker/Dockerfile [hunk 3]
```diff
@@ -18,8 +18,16 @@ RUN yum install -y epel-release \
...
+    mv oras-install/oras /usr/local/bin/ && \
```
francoisferrand: you can avoid this with a single call to `tar` : e.g. by simply specifying that you want to extract only "oras".

```
tar -zxf oras_${ORAS_VERSION}_*.tar.gz -C /usr/local/bin oras
```

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+            "type": "constant",
+            "label": "namespace",
+            "description": "Namespace associated with the Zenko instance",
+            "value": "zenko"
+        },
+        {
+            "name": "job",
+            "type": "constant",
+            "label": "job",
+            "description": "Name of the Redis job, used to filter only the Redis instances.",
```
francoisferrand: should not be 'redis'...

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                    "placement": "bottom"
+                },
+                "tooltip": {
+                    "mode": "single"
+                }
+            },
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum by (code) (rate(http_requests_total{namespace=\"${namespace}\", job=\"${job}\"}[10m]))",
```
francoisferrand: should this rate here be computed over a fixed 10m window (or smaller window? i used 1 min for alerts, but i don't know what value is appropriate, as we discussed on alerts PR...), or use the window that is specified in grafana?

(the answer may be more obvious with the dashboard in front, but looking at the JSON it is not very clear to me, sorry...)

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                "x": 0,
+                "y": 0
+            },
+            "id": 2,
+            "panels": [],
+            "title": "Response codes",
+            "type": "row"
+        },
+        {
+            "datasource": null,
```
francoisferrand: should not be NULL : this is there is a 'prometheus' datasource variable

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                    "interval": "",
+                    "legendFormat": "Success rate",
+                    "refId": "A"
+                }
+            ],
+            "title": "Success rate",
+            "type": "gauge"
+        },
+        {
+            "datasource": null,
```
francoisferrand: shoudl use variable

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                "x": 0,
+                "y": 9
+            },
+            "id": 10,
+            "panels": [],
+            "title": "Latency",
+            "type": "row"
+        },
+        {
+            "datasource": null,
```
francoisferrand: should use variable

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                    "placement": "bottom"
+                },
+                "tooltip": {
+                    "mode": "single"
+                }
+            },
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "(sum(rate(http_request_duration_seconds_sum{namespace=\"${namespace}\", job=\"${job}\"}[10m]) * 1000)) / (sum(rate(http_request_duration_seconds_count{namespace=\"${namespace}\", job=\"${job}\"}[10m]) * 1000))",
```
francoisferrand: same thing about 10m.
also, just thinking: should we use the same averaging window for dashboard and alerts?

Note that we may use a variable to avoid repeating this value, if it makes sense....

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                    "placement": "bottom"
+                },
+                "tooltip": {
+                    "mode": "single"
+                }
+            },
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(rate(http_request_duration_seconds_sum{namespace=\"${namespace}\", job=\"${job}\", action=~\"objectPut|objectPutPart|objectCopy|objectPutCopyPart\"}[10m]) * 1000) /\nsum(rate(http_request_duration_seconds_count{namespace=\"${namespace}\", job=\"${job}\", action=~\"objectPut|objectPutPart|objectCopy|objectPutCopyPart\"}[10m]) * 1000)",
```
francoisferrand: not sure about the `sum` here: the latency is just the `http_request_duration_seconds_sum / http_request_duration_seconds_count` (with `rate` to scope it to 10 minute), making the sum afterwards seems weird...

if needed at all (e.g. not handled by rate) for "merging" the different objectPut/PutPart/...", should we not do the sum() before getting the rate?

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                "x": 0,
+                "y": 18
+            },
+            "id": 16,
+            "panels": [],
+            "title": "Data rate",
+            "type": "row"
+        },
+        {
+            "datasource": null,
```
francoisferrand: use datasource variable

monitoring/dashboard.json
```diff
@@ -0,0 +1,627 @@
...
+                    "placement": "bottom"
+                },
+                "tooltip": {
+                    "mode": "single"
+                }
+            },
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(rate(http_response_size_bytes_sum{namespace=\"${namespace}\", job=\"${job}\"}[10m]))",
```
francoisferrand: what is the purpose of the sum here?

jbertran (approved):

eve/main.yml
lavigne958: it is easier to copy/paste it in other repos (vault, backbeat).
if we need to push more files it makes it easier I think.

eve/workers/docker/Dockerfile
lavigne958: will do.

eve/workers/docker/Dockerfile
lavigne958: sure ? i thought you must pass name filename from the archive, then rename the extracted file. I will try that

lavigne958: does not work, you can't rename the file (or this might be a different command).

```
# tar cvf test.tar titi
# tar xvf test.tar -C ./bin toto                                                                                                                                
tar: toto: Not found in archive
tar: Exiting with failure status due to previous errors
```

monitoring/dashboard.json
lavigne958: in fact 10 minutes is quick short, previously in xcore for a graph over 7 days we used a 5 hours window, it give the tendency of the (bandwidth|latency|....)

here I used 10 minutes but later with some real workload will change it to a longer time range. 

we don´t want to use the grafana value for the window the user is looking at because with a window of 7 days for the graph we want something like 1/10th of it for the rate window or something similar.

monitoring/dashboard.json
lavigne958: this is odd I have `"datasource": null` everywhere :open_mouth: 
but It works fine when I created the dashboard

monitoring/dashboard.json
lavigne958: no for alerts we want to have the _spikes_ when some latency reach them and we want them for a short period of time, here people may look at some graphs over a week period, or day period and so we want at very large value here (from minutes to hours).

eve/main.yml
francoisferrand: once we are moved to github-actions, we may be able to write a dedicated and re-usable "oras" github action, which will take care of that. But that will need to wait a bit, we are not there yet...

eve/workers/docker/Dockerfile
francoisferrand: you are not renaming the file (`mv oras-install/oras /usr/local/bin/` is just moving the file `oras` from `oras/install` to `/usr/local/bin`) so it should work...

(in your exemple, you must pass the name of the file which you want to extract from the archive: you added 'titi', so you must request to extract 'titi' as well)

monitoring/dashboard.json
francoisferrand: looking at the doc, we should thus use `$__rate_interval` to handle all cases: https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/

miniscruff:
Can we get a screenshot of the dashboard to reference in the JSON

eve/workers/docker/Dockerfile
lavigne958: alright, I see, sorry I did not get your point at first now I get it, yes I will update so we extract the file in the right place in once command.

monitoring/dashboard.json
lavigne958: looking at the article now, thanks did not know about that.

monitoring/dashboard.json
lavigne958: you can't just use `http_request_duration_seconds_sum / http_request_duration_seconds_count` because both metrics are counter, so they never go down, they always go up. This means the resulting value from the division here will only rise and never go down.

This is why you want to use the `rate()` function, this way you can have a graph that goes up and down regarding if your latency goes up or goes down.

you cannot get a range vector (aka: the syntax `metric{n="toto"}[10m]`) from the resulting `sum` method. you must apply it to your metric, then call functions on the top of it. The `sum` method does not accept range vector so we have only one order of function call: metric -> range vector -> rate -> sum

What can be done is multiplying the result by 1000 after the sum instead of before :thinking:

monitoring/dashboard.json
lavigne958: I use a sum to aggregate the different pods

lavigne958:
> Can we get a screenshot of the dashboard to reference in the JSON

Sure, it looks like this:
screenshot_2021-12-14_12-08-04_045656212

monitoring/dashboard.json
francoisferrand: `datasource: null` will always use the "default" datasource, which is actually prometheus...

monitoring/dashboard.json
```diff
@@ -0,0 +1,614 @@
...
+                    "placement": "bottom"
+                },
+                "tooltip": {
+                    "mode": "single"
+                }
+            },
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "(sum(rate(http_request_duration_seconds_sum{namespace=\"${namespace}\", job=\"${job}\"}[$__rate_interval]) * 1000)) / (sum(rate(http_request_duration_seconds_count{namespace=\"${namespace}\", job=\"${job}\"}[$__rate_interval]) * 1000))",
```
francoisferrand: * Should not multiply the _count by `1000` in divisor
* Should not multiply the _sym by `1000` in dividend, and instead specify the correct unit in grafana (`"unit": "s"`) ?

lavigne958: done, removed the `* 1000` and use unit `s`.

francoisferrand:
after reviewing "external" dashboards from Rahul, I think we should add some extra indicators on this dashboard (if we can get that information):
* Up/Down status indicator
* Number of pods/replicas
* Number and size of objects (over time)

(e.g. an extra "summary" row, on top)

francoisferrand (approved):

lavigne958:
squashed commits.

lavigne958:
> after reviewing "external" dashboards from Rahul, I think we should add some extra indicators on this dashboard (if we can get that information):
> 
>     * Up/Down status indicator
> 
>     * Number of pods/replicas
> 
>     * Number and size of objects (over time)
> 
> 
> (e.g. an extra "summary" row, on top)

I'll create a new ticket with this information for later to update the dashboard.

PR #4120, opened 2021-12-15
CLDSRV-73 - convert nanoseconds to seconds

lavigne958:
during metric measurement we get the request time in nanoseconds.
Convert it to seconds, as per the standard from prometheus, by
dividing it by 1000000000.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -122,7 +122,7 @@ class S3Server {
             monitoringClient.httpRequestsTotal.labels(labels).inc();
             monitoringClient.httpRequestDurationSeconds
                 .labels(labels)
-                .observe(responseTimeInMs / 1000000);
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -122,7 +122,7 @@ class S3Server {
             monitoringClient.httpRequestsTotal.labels(labels).inc();
             monitoringClient.httpRequestDurationSeconds
                 .labels(labels)
-                .observe(responseTimeInMs / 1000000);
+                .observe(responseTimeInMs / 1000000000);
```
jbertran: Use exponent notation so we don't make that mistake again :slightly_smiling_face: 

```
❯ node
> 1e9
1000000000
```

francoisferrand (approved):

jbertran (approved):

PR #4131, opened 2021-12-22
improvement: CLDSRV-79-support checkPolicy for session user

XinLi0207:
### Motivation and context

Currently in cloudserver, for some APIs, if it's an account user, we don't check any permission, we only check the permission of iam users. And also we assume only either it's an account user or it's iam user. This caused an issue after we introduced role session user in Vault, we should not use the logic that if it's not iam user then it must be an account user and we don't check anything for it. We need to support checking if it's a role session user for these APIs.

APIs affected are below: createBucket, multipleObjectDelete, objectPutCopyPart, objectPutPart, objectPutRetention

### Related issues

https://scality.atlassian.net/browse/CLDSRV-79

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -65,6 +65,7 @@ const writeContinue = require('../utilities/writeContinue');
 const validateQueryAndHeaders = require('../utilities/validateQueryAndHeaders');
 const parseCopySource = require('./apiUtils/object/parseCopySource');
 const { tagConditionKeyAuth } = require('./apiUtils/authorization/tagConditionKeys');
+const { isRequesterASessionUser } = require('./apiUtils/authorization/permissionChecks');
 
 const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
 
@@ -173,6 +174,9 @@ const api = {
             const authNames = { accountName: userInfo.getAccountDisplayName() };
             if (userInfo.isRequesterAnIAMUser()) {
                 authNames.userName = userInfo.getIAMdisplayName();
+            }
+            if (isRequesterASessionUser(userInfo)) {
+                authNames.sessionName = userInfo.getShortid().split(':')[1];
             }
             log.addDefaultFields(authNames);
             if (tagAuthResults) {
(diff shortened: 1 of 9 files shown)


lib/api/api.js
```diff
@@ -174,6 +175,9 @@ const api = {
             if (userInfo.isRequesterAnIAMUser()) {
                 authNames.userName = userInfo.getIAMdisplayName();
             }
+            if (isRequesterASessionUser(userInfo)) {
+                authNames.sessionName = userInfo.getShortid().split(':')[1];
```
ghost, line 179: Do we use this `sessionName` somewhere?

ghost (commented):
Small comment, otherwise LGTM, I could test the PR locally and confirm that it fixes the issue.

lib/api/api.js
XinLi0207: no, actually it's just for log printing.

ghost (approved):
LGTM, two potential open questions to discuss:
- What about the other APIs for which we never check policies?
- Should we add this `isRequesterASessionUser` to the arsenal (I think not, but it is worth discussing)?

ghost:
> LGTM, two potential open questions to discuss:
> 
> * What about the other APIs for which we never check policies?
> * Should we add this `isRequesterASessionUser` to the arsenal (I think not, but it is worth discussing)?

I can imagine adding to arsenal would be a good long term idea but I might not have a large enough context knowledge to say.

ghost (approved):
LGTM

XinLi0207:
> LGTM, two potential open questions to discuss:
> 
> * What about the other APIs for which we never check policies?
> * Should we add this `isRequesterASessionUser` to the arsenal (I think not, but it is worth discussing)?

I agree with you that assume role only used in cloudserver and vault, adding `isRequesterASessionUser` to arsenal seems a little bit too customized. 
but in this case, `isRequesterASessionUser` are not consistent with the existing function `isRequesterAnIAMUser`, looks awkward.

for the other APIs, if we need to do checkPolicies for them, I think we need to discuss it with @anurag4DSB and hippo team, let's keep this open topic here and discuss it later.

PR #4138, opened 2021-12-29
Improve cloudserver dashboards

francoisferrand:
* Add overview panels (number of buckets, number of objects, number of Up replicas)
* Add missing "Top10 buckets with errors" panels
* Add request duration heatmap
* Add upload/download chunk size distribution
* Fix unit for http responses

Issue: CLDSRV-74

Screenshot 2021-12-29 at 16 24 06

### DIFF
--- a/monitoring/dashboard.json
+++ b/monitoring/dashboard.json
@@ -1264,7 +1264,7 @@
                             }
                         ]
                     },
-                    "unit": "s"
+                    "unit": "binBps"
                 },
                 "overrides": []
             },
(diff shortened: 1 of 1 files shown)


monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
-    "gnetId": null,
-    "graphTooltip": 0,
-    "id": 33,
-    "links": [],
-    "panels": [
+            "name": "job",
+            "type": "constant",
+            "label": "job",
+            "description": "Name of the Cloudserver job, used to filter only the Cloudserver instances.",
+            "value": "artesca-data-connector-s3api-metrics"
```
ghost, line 31: Just out of curiosity, is it a good thing to hardcode the artesca prefix here? How is the testing of this feature within other CIs if the prefix is different?

ghost (commented):
A quick comment before approving :slightly_smiling_face: 
Also, why does the latency on your bottom right screen go from 500 ms to +Inf? Is this on purpose? Maybe a better display would make more sense, replacing for example +Inf by 600ms.

francoisferrand:
> why does the latency on your bottom right screen go from 500 ms to +Inf? Is this on purpose? Maybe a better display would make more sense, replacing for example +Inf by 600ms.

There is a known bug in the metrics : the values are completely wrong, by a factor of 1000... So normally we would have much smaller values.

+Inf is the upper value of the range of the last bucket... I am not sure I can have much impact on how/what grafana will display in that case :-/ (though hopefully it will look better with the metrics fix, as it does in vault dashboard)

monitoring/dashboard.json
francoisferrand: it is not really "hard-coded" : in production, the value is actually computed by Zenko-operator, and this value is actually not used.

However, having the value here makes it easier to load (manually) the dashboard in an artesca setup : grafana will prompt to get the values for the __inputs, but it will pre-populate with the values specified here...

ghost: I see, thanks! :)

ghost (approved):

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                    "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(http_requests_total{namespace=\"${namespace}\", job=\"${job}\"})",
```
lavigne958: I disagree with this formula. I see no points watching a counter growing.
If we want to measure how many requests we are currently receiving we must use something like:
`sum(rate(http_requests_total{namespace="${namespace}", job="${job}"}[1h]))`

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                    "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(cloud_server_number_of_buckets{namespace=\"${namespace}\",job=\"${job}\"})",
```
lavigne958, line 281: I currently don't have a running platform but, this is a gauge right ?

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                    "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(cloud_server_number_of_objects{namespace=\"${namespace}\",job=\"${job}\"})",
```
lavigne958, line 338: same here this is a gauge right ?

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                    "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(http_requests_total{namespace=\"${namespace}\",job=\"${job}\",code=\"200\"})",
```
lavigne958: same as `http_requests_total` their is no point displaying the counter itself.

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                    "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(http_requests_total{namespace=\"${namespace}\", job=\"${job}\",code=~\"4..\"})",
```
lavigne958: same here, no need to display a pure counter.
I know I am annoying :-) with this

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                    "values": false
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(http_requests_total{namespace=\"${namespace}\", job=\"${job}\",code=~\"5..\"})",
```
lavigne958: same here, no need to display a pure counter without a formula.

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(http_active_requests{namespace=\"${namespace}\", job=\"${job}\"})",
+                    "instant": false,
```
lavigne958, line 665: This should have the value `instant=true`, because we only care about the last non null value only.

monitoring/dashboard.json
```diff
@@ -16,35 +24,1214 @@
...
+                },
+                "text": {},
+                "textMode": "auto"
+            },
+            "pluginVersion": "8.0.6",
+            "targets": [
+                {
+                    "exemplar": true,
+                    "expr": "sum(up{namespace=\"${namespace}\", job=\"${job}\"})",
+                    "instant": false,
```
lavigne958, line 99: same here I believe the `instant=true` should be set, we only want the last non null value.

monitoring/dashboard.json
francoisferrand: yes and no :-)

This panel is a **stat**, showing the "Total requests" : soi the value we want to display is the current one.
In addition, the value computed with `rate` (or actually, `increase`) is not correct  : seems to be some rounding error, and we don't reach the same value...

monitoring/dashboard.json
francoisferrand: yes

monitoring/dashboard.json
francoisferrand: yes

monitoring/dashboard.json
francoisferrand: not really : since we have the "background" graph : showing the history of this gauge

monitoring/dashboard.json
francoisferrand: not really : since we have the "background" graph : showing the history of this gauge

monitoring/dashboard.json
lavigne958: I understand and you currently display the latest recorded value for that counter, but as life itself will go on, the counter will only increase, so what we read here is the number of served requests since the platform has been setup. this is nice for breaking records but I don't see the use of it in monitoring :open_mouth: I don't mind keeping it though

lavigne958 (approved):

PR #4147, opened 2022-01-05
CLDSRV-89 - update kubernetes readiness/liveness probe route

lavigne958:
Update the readiness and liveness route for kubernetes.
The operator will configure the pod to check for readiness
and liveness on the same port as for the metrics.

In this commit allow the new routes on the metric code path.

### DIFF
--- a/Healthchecks.md
+++ b/Healthchecks.md
@@ -1,6 +1,7 @@
 # S3 Healthcheck
 
-Scality S3 exposes a healthcheck route `/live` which returns a
+Scality S3 exposes a healthcheck route `/live` on the port used
+for the metrics (defaults to port 8002) which returns a
 response with HTTP code
 
 - 200 OK
--- a/Healthchecks.md
+++ b/Healthchecks.md
@@ -1,6 +1,6 @@
 # S3 Healthcheck
 
-Scality S3 exposes a healthcheck route `/healthcheck` which returns a
+Scality S3 exposes a healthcheck route `/live` which returns a
 response with HTTP code
 
 - 200 OK
--- a/Healthchecks.md
+++ b/Healthchecks.md
@@ -1,6 +1,6 @@
 # S3 Healthcheck
 
-Scality S3 exposes a healthcheck route `/_/healthcheck` which returns a
+Scality S3 exposes a healthcheck route `/healthcheck` which returns a
 response with HTTP code
 
 - 200 OK
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -2,7 +2,7 @@
 version: 0.2
 
 branches:
-  feature/*, documentation/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*, dependabot/*:
+  feature/*, documentation/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*, dependabot/*, user/*:
     stage: pre-merge
 
 models:
--- a/lib/management/push.js
+++ b/lib/management/push.js
@@ -277,7 +277,7 @@ function addOverlayMessageListener(callback) {
 
 function startPushConnectionHealthCheckServer(cb) {
     const server = http.createServer((req, res) => {
-        if (req.url !== '/healthcheck') {
+        if (req.url !== '/_/healthcheck') {
             res.writeHead(404);
             res.write('Not Found');
         } else if (connected) {
--- a/lib/management/push.js
+++ b/lib/management/push.js
@@ -277,7 +277,7 @@ function addOverlayMessageListener(callback) {
 
 function startPushConnectionHealthCheckServer(cb) {
     const server = http.createServer((req, res) => {
-        if (req.url !== '/_/healthcheck') {
+        if (req.url !== '/healthcheck') {
             res.writeHead(404);
             res.write('Not Found');
         } else if (connected) {
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -189,10 +189,7 @@ function monitoringHandler(clientIP, req, res, log) {
     if (req.method !== 'GET') {
         return monitoringEndHandler(errors.MethodNotAllowed, []);
     }
-    const monitoring = (req.url === '/metrics');
-    const healthcheck = (req.url === '/healthcheck');
-    const healthcheckdeep = (req.url === '/healthcheck/deep');
-    if (!monitoring && !healthcheck && !healthcheckdeep) {
+    if (req.url !== '/metrics') {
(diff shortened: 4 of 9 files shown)


lib/utilities/monitoringHandler.js
```diff
@@ -190,7 +190,9 @@ function monitoringHandler(clientIP, req, res, log) {
         return monitoringEndHandler(errors.MethodNotAllowed, []);
     }
     const monitoring = (req.url === '/metrics');
-    if (!monitoring) {
+    const healthcheck = (req.url === '/healthcheck');
```
jbertran: Minor: this is a switch with fallthrough on every case, and `MethodNotAllowed` as default.

lib/utilities/monitoringHandler.js
```diff
@@ -190,7 +190,9 @@ function monitoringHandler(clientIP, req, res, log) {
...
+    const healthcheckdeep = (req.url === '/healthcheck/deep');
+    if (!monitoring && !healthcheck && !healthcheckdeep) {
         return monitoringEndHandler(errors.MethodNotAllowed, []);
     }
     return routeHandler(req, res, monitoringEndHandler);
```
francoisferrand, line 195: I don't think this works.
`monitoringHandler()` function is not expected to route traffic : it actually just handles the /metrics path, and will always return the metrics.

(granted, maybe this `routeHandler()` function is not clearly named; but it is a local/private function, and just means "handle this route")

Healthchecks are handled by `utilities/healthcheckHandler`, and should not go through monitoring : this "dispatch" may be done in `routeAdminRequest()` in `server.js`.

francoisferrand (changes requested):
Please also update documentation (`Healthchecks.md`), existing tests (at least metrics test rely on this) and add specific tests for this feature, if not already present.

lib/utilities/monitoringHandler.js
lavigne958: will do,.

lib/utilities/monitoringHandler.js
lavigne958: don't need anymore

lib/utilities/monitoringHandler.js
lavigne958: done.

lavigne958:
read to review.

Healthchecks.md
```diff
@@ -1,6 +1,6 @@
 # S3 Healthcheck
 
-Scality S3 exposes a healthcheck route `/_/healthcheck` which returns a
+Scality S3 exposes a healthcheck route `/healthcheck` which returns a
```
francoisferrand: should indicate this is not on the same port as s3, actually.

eve/main.yml
```diff
@@ -2,7 +2,7 @@
 version: 0.2
 
 branches:
-  feature/*, documentation/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*, dependabot/*:
+  feature/*, documentation/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*, dependabot/*, user/*:
```
francoisferrand, line 5: should this be commited?

lib/management/push.js
```diff
@@ -277,7 +277,7 @@ function addOverlayMessageListener(callback) {
 
 function startPushConnectionHealthCheckServer(cb) {
     const server = http.createServer((req, res) => {
-        if (req.url !== '/_/healthcheck') {
+        if (req.url !== '/healthcheck') {
```
francoisferrand: not sure this is related.
I don't know what this "push connection" is, but it seems to be a dedicated server : so it is not needed for our purpose I think, and changing it here would break the other end of the push channel?

Healthchecks.md
lavigne958: actually not, in the ticket it is defined that we put it on a different port, it should be on the metrics port, but sure we should indicate the port to use.

eve/main.yml
lavigne958: yep, so everyone can create a `user/` branch to test things on the CI and not have to make `feature/` or `documentation/` branches.

Healthchecks.md
francoisferrand: yes exactly my point : looking at it here we may think that healthcheck is on the same port as S3, which is not the case. So we should indicate it is on this metrics port (or admin, not sure how it is called in the config).

lib/management/push.js
lavigne958: it seems to be a server that has listent to a websocket and pushes some metrics somewhere. it has the healthcheck route and I assumed that route is checked. if it is not, we could leave it here untouched.

Healthchecks.md
lavigne958: will do.

lib/management/push.js
francoisferrand: i am not sure, but seeing the same value is not proof enough : since this `/_/healthcheck` is used everywhere...
hopefully someone knowledgeable on the subject will be able to confirm either way....

lavigne958: right, wait for confirmation from someone else then regarding the outcome I will push a new commit.
I'll try to limit the number of commits as it is very hard to make them land and have a green build

lavigne958:
Updated the route name to match new names:
- liveness probe: `/live`
- readiness probe: `/ready`

lib/management/push.js
lavigne958: After looking around, this piece of code is not called by cloudserver. it is a dedicated server that can be run manually (manually => yarn run start_metrics_server) may be it is used by other components or simply unused at all.

lavigne958: I think I should update the route to makes sure it matches the same route name as expected by zenko in general

Healthchecks.md
lavigne958: done.

miniscruff (approved):

lib/management/push.js
francoisferrand: i confirmed with Rached that this is not used in Artesca (and should generally be removed, it was related to Orbit and the "cloud management" feature or Zenko 1)

francoisferrand (approved):

lib/management/push.js
lavigne958: I am not too sure, this is run in artesca, or at least when I run a local cluster I have a pod running, I will try on a fresh artesca, but I think it runs. we could update it so it uses the `/live` route even though it will be removed for now it is here and should follow the update ?

lavigne958:
revert the changes on the push server.
Waitting for green CI

lavigne958:
Needed to rebase my branch to include the fix to pass the CI.

PR #4183, opened 2022-01-18
bugfix: CLDSRV-47 check http header size

alexanderchan-scality:
return error if http headers size is too large

### DIFF
--- a/lib/api/apiUtils/object/checkHttpHeadersSize.js
+++ b/lib/api/apiUtils/object/checkHttpHeadersSize.js
@@ -1,0 +1,25 @@
+const { errors } = require('arsenal');
+const { maxHttpHeadersSize } = require('../../../../constants');
+
+/**
+ * Checks the size of the HTTP headers
+ * @param {object} requestHeaders - HTTP request headers
+ * @return {object} object with error or null
+ */
+function checkHttpHeadersSize(requestHeaders) {
+    let httpHeadersSize = 0;
+
+    Object.keys(requestHeaders).forEach(header => {
+        httpHeadersSize += Buffer.byteLength(header, 'utf8') +
+            Buffer.byteLength(requestHeaders[header], 'utf8');
+    });
+
+    if (httpHeadersSize > maxHttpHeadersSize) {
+        return {
+            httpHeadersSizeError: errors.HttpHeadersTooLarge,
+        };
+    }
+    return {};
+}
+
+module.exports = checkHttpHeadersSize;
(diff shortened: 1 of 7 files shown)


lib/api/apiUtils/object/checkHttpHeadersSize.js
```diff
@@ -0,0 +1,25 @@
+const { errors } = require('arsenal');
+const { maxHttpHeadersSize } = require('../../../../constants');
+
+/**
+ * Checks the size of the size of the HTTP headers
```
miniscruff:
```suggestion
 * Checks the size of the HTTP headers
```

lib/api/apiUtils/object/checkHttpHeadersSize.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {object} requestHeaders - HTTP request headers
+ * @return {object} object with error or null
+ */
+function checkHttpHeadersSize(requestHeaders) {
+    let httpHeadersSize = 0;
+
+    Object.keys(requestHeaders).forEach(header => {
+        httpHeadersSize += Buffer.byteLength(header, 'utf8') +
```
miniscruff, line 13: Just want to clarify this simply takes the number of bytes for the header and value, but does not count any separators or quotes we would need if we were to send the headers as a request. Not sure if that makes any sense...

miniscruff (changes requested):

lib/api/apiUtils/object/checkHttpHeadersSize.js
alexanderchan-scality: I am not sure I understand the issue being raised. Wouldn't quotes/separators be included as part of the header values (`;` or any escaped quotes)?  
I would need to check how nodejs validates the request header size

alexanderchan-scality: (https://scality.atlassian.net/browse/S3C-5072?focusedCommentId=293869) describes the rationale behind the change

miniscruff: So in the request you see the headers as:
`url?h1=val1&h2=val2` etc
Just want to make sure the "header size" is not the total size of that string and is instead just the key and values. That is, the equals and ampersand are ignored.

alexanderchan-scality: I might be wrong, but I don't think the request querystring size is factored into the header size count.

miniscruff (approved):

alexanderchan-scality:
ping

rahulreddy (approved):

PR #4188, opened 2022-01-20
AbortMPU writes an empty object with isAborted and uploadId set

tmacro:
# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -32,7 +32,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             metadataValidateBucketAndObj(metadataValParams, log,
                 (err, destinationBucket) => {
                     if (err) {
-                        return next(err, destinationBucket, null);
+                        return next(err, destinationBucket);
                     }
                     if (destinationBucket.policies) {
                         // TODO: Check bucket policies to see if user is granted
@@ -58,7 +58,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 });
         },
         function ifMultipleBackend(mpuBucket, mpuOverviewObj, destBucket,
-            next) {
+        next) {
             if (config.backends.data === 'multiple') {
                 let location;
                 // if controlling location constraint is not stored in object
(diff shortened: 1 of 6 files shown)


jonathan-gramain (commented):
At first glance, looks good to me, but we need to be careful about keeping any existing object (MPU or not), since the abortMPU applies to an MPU in progress, not on a validated object that may already exist with the same key. That's why we use a conditional PUT, just for reference.

As we discussed the need to create a stub request is not ideal but if it solves the issue, we can keep it.

tmacro:
Ah, so for preexisting objects only the `isAborted` and `uploadId` flags should be updated, rather than writing a new object?

jonathan-gramain:
@tmacro there seems to be an overlook in the proposal that is described in the ticket, because we should be able to abort a *new* MPU on top of an existing, valid object (MPU or non-MPU). It is not obvious because on non-versioned buckets we only store a single entry per object, where as Vianney mentioned, ideally we should deal with this issue by versioning. But it should be doable anyway.

Here's a suggestion, @vrancurel can you please review it as well? The main difference with the original design is that we add a new field `abortedUploadId` that we can keep along with an `uploadId` instead of replacing it.

- `abortMPU`: send a special PUT request containing a single field: `abortedUploadId`. Four cases to handle in metadata (to ensure atomicity):
  - if an object already exists that is either a non-MPU or an MPU which **`uploadId` does not match the `abortedUploadId`**, **merge** the `abortedUploadId` into the existing metadata of the object. If there is already an `abortedUploadId`, then overwrite it with the new value. Return success.
  - if an object already exists that is an MPU which **`uploadId` matches the `abortedUploadId`**, **replace** the whole object metadata by a special abort placeholder (see next case). Return success.
  - if **no object exists**, **create** a special placeholder value `{"isAbort": true, "abortedUploadId": "someUploadId"}`, like already planned. This entry must be ignored on GET and listings. Return success.
  - if an `"isAbort": true` entry is already there i.e. **an abort placeholder exists**, **replace** its `abortedUploadId` with the new value. Return success
- `completeMPU`: do a conditional PUT on the object, with the following condition:
  - if no object exists, do the PUT and return success
  - if an object exists, whether it's a regular object or a `isAbort` placeholder, check its `abortedUploadId` value.
    -  If there is no `abortedUploadId` or if it is different than the `uploadId` of the MPU to complete, do the PUT and return success
    - if the `abortedUploadId` matches the `uploadId` of the object to complete, fail the condition check and return a precondition check error. Cloudserver should return the same error than when trying to complete a MPU with an invalid uploadID

vrancurel:
The suggestion looks good to me. Please be careful that for versioned buckets we have to update 2 objects: the lastVersion and the master, but normally the condition / merge logic shall be before the double update in Repd.

PR #4210, opened 2022-01-28
bf: CLDSRV-111 log correct client ip

naren-rajendran:
Log correct client ip.


nicolas2bert:
Could you add a test here?

naren-rajendran:
> Could you add a test here?

Yes, but not until I use updated arsenal.

naren-rajendran:
Rebasing to 7.4 and opening a new PR.

PR #4224, opened 2022-02-02
Feature/cldsrv 112 add mdsearch api

XinLi0207:
1. add new metadataSearch API (copy from bucketGet API)
2. remove search functionality from bucketGetAPI
3. didn't add unit test for this API cause in-memory db doesn't support metadata search
4. functional tests for md search already defined in https://github.com/scality/cloudserver/tree/development/8.3/tests/functional/aws-node-sdk/test/mdSearch, no need to change
5. unit tests for testing all search queries are defined in https://github.com/scality/cloudserver/blob/development/8.3/tests/unit/utils/validateSearch.js

please feel free to tell us if you feel like more tests are needed

### DIFF
--- a/lib/api/metadataSearch.js
+++ b/lib/api/metadataSearch.js
@@ -1,0 +1,149 @@
+const { errors, versioning } = require('arsenal');
+const constants = require('../../constants');
+const services = require('../services');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const { pushMetric } = require('../utapi/utilities');
+const validateSearchParams = require('../api/apiUtils/bucket/validateSearch');
+const parseWhere = require('../api/apiUtils/bucket/parseWhere');
+const versionIdUtils = versioning.VersionID;
+const monitoring = require('../utilities/monitoringHandler');
+const { decryptToken }
+    = require('../api/apiUtils/object/continueToken');
+const { config } = require('../Config');
+const { processVersions, processMasterVersions } = require('./bucketGet');
+
+
+function handleResult(listParams, requestMaxKeys, encoding, authInfo,
+                      bucketName, list, corsHeaders, log, callback) {
+    // eslint-disable-next-line no-param-reassign
+    listParams.maxKeys = requestMaxKeys;
+    // eslint-disable-next-line no-param-reassign
+    listParams.encoding = encoding;
+    let res;
+    if (listParams.listingType === 'DelimiterVersions') {
+        res = processVersions(bucketName, listParams, list,
+            config.versionIdEncodingType);
+    } else {
+        res = processMasterVersions(bucketName, listParams, list);
+    }
+    pushMetric('metadataSearch', log, { authInfo, bucket: bucketName });
+    monitoring.promMetrics('GET', bucketName, '200', 'metadataSearch');
+    return callback(null, res, corsHeaders);
+}
+
+/**
+ * metadataSearch - Return list of objects in bucket that meet the search query, supports v1 & v2
+ * @param  {AuthInfo} authInfo - Instance of AuthInfo class with
+ *                               requester's info
+ * @param  {object} request - http request object
+ * @param  {function} log - Werelogs request logger
+ * @param  {function} callback - callback to respond to http request
+ *  with either error code or xml response body
+ * @return {undefined}
+ */
+function metadataSearch(authInfo, request, log, callback) {
+    const params = request.query;
+    const bucketName = request.bucketName;
+    const v2 = params['list-type'];
+    if (v2 !== undefined && Number.parseInt(v2, 10) !== 2) {
+        return callback(errors.InvalidArgument.customizeDescription('Invalid ' +
+            'List Type specified in Request'));
+    }
+    log.debug('processing request', { method: 'metadataSearch' });
+    const encoding = params['encoding-type'];
+    if (encoding !== undefined && encoding !== 'url') {
+        monitoring.promMetrics(
+            'GET', bucketName, 400, 'metadataSearch');
+        return callback(errors.InvalidArgument.customizeDescription('Invalid ' +
+            'Encoding Method specified in Request'));
+    }
+    const requestMaxKeys = params['max-keys'] ?
+        Number.parseInt(params['max-keys'], 10) : 1000;
+    if (Number.isNaN(requestMaxKeys) || requestMaxKeys < 0) {
+        monitoring.promMetrics(
+            'GET', bucketName, 400, 'metadataSearch');
+        return callback(errors.InvalidArgument);
+    }
+    // AWS only returns 1000 keys even if max keys are greater.
+    // Max keys stated in response xml can be greater than actual
+    // keys returned.
+    const actualMaxKeys = Math.min(constants.listingHardLimit, requestMaxKeys);
+
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'metadataSearch',
+        request,
+    };
+    const listParams = {
+        listingType: 'DelimiterMaster',
+        maxKeys: actualMaxKeys,
+        delimiter: params.delimiter,
+        prefix: params.prefix,
+    };
+    try {
+        const validatedAst = validateSearchParams(params.search).ast;
+        listParams.mongifiedSearch = parseWhere(validatedAst);
+    } catch (err) {
+        log.debug(err.message, {
+            stack: err.stack,
+        });
+        monitoring.promMetrics(
+            'GET', bucketName, 400, 'metadataSearch');
+        return callback(errors.InvalidArgument
+            .customizeDescription('Invalid sql where clause ' +
+                'sent as search query'));
+    }
+    if (v2) {
+        listParams.v2 = true;
+        listParams.startAfter = params['start-after'];
+        listParams.continuationToken =
+            decryptToken(params['continuation-token']);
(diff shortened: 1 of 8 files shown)


lib/api/metadataSearch.js
```diff
@@ -0,0 +1,410 @@
...
+        <Key>my-image.jpg</Key>
+        <LastModified>2009-10-12T17:50:30.000Z</LastModified>
+        <ETag>&quot;fba9dede5f27731c9771645a39863328&quot;</ETag>
+        <Size>434234</Size>
+        <StorageClass>STANDARD</StorageClass>
+    </Contents>
+</ListBucketResult>
+*/
+
+/*	Sample XML response for GET bucket objects:
```
KillianG:
```suggestion
/* Sample XML response for GET bucket objects:
```

lib/api/metadataSearch.js
```diff
@@ -0,0 +1,410 @@
...
+    = require('../api/apiUtils/object/continueToken');
+const { config } = require('../Config');
+
+// do not url encode the continuation tokens
+const skipUrlEncoding = new Set([
+    'ContinuationToken',
+    'NextContinuationToken',
+]);
+
+/* Sample XML response for GET bucket objects V2:
```
ghost: Do we need to keep these comments? Looks like they are equal to thoses in the bucketGet file. Or maybe we can document here the metadata-related calls and expected responses?

lib/api/metadataSearch.js
```diff
@@ -0,0 +1,410 @@
...
+        </Owner>
+    </DeleteMarker>
+    <CommonPrefixes>
+        <Prefix>photos/</Prefix>
+    </CommonPrefixes>
+</ListVersionsResult>
+*/
+/* eslint-enable max-len */
+
+function processVersions(bucketName, listParams, list, encType) {
```
ghost: The functions `processVersions`, `processMasterVersions`, `handleResult` are already in bucketGet.js, can we reuse them, so that we don't have to update 2 function when we update the GetBucket API?

ghost: Also, because `skipUrlEncoding` is used in `processMasterVersions` only, no need to redefine it in this file.

lib/api/metadataSearch.js
```diff
@@ -0,0 +1,410 @@
...
+ * bucketGet - Return list of objects in bucket, supports v1 & v2
+ * @param  {AuthInfo} authInfo - Instance of AuthInfo class with
+ *                               requester's info
+ * @param  {object} request - http request object
+ * @param  {function} log - Werelogs request logger
+ * @param  {function} callback - callback to respond to http request
+ *  with either error code or xml response body
+ * @return {undefined}
+ */
+function metadataSearch(authInfo, request, log, callback) {
```
ghost, line 45: Quick note on this, if reviewers are wondering where is the logic calling this function instead of `bucketGet`, it is in Arsenal: https://github.com/scality/Arsenal/commit/68c5b42e6f8994c2d9ba10d91b70d9a01bbcd759#diff-b2fc5116de013086739badfff8f252c51deb78bffd5b24fe117f1edb1fc16b2eR103

tests/unit/api/metadataSearch.js
```diff
@@ -0,0 +1,258 @@
...
+}, postBody);
+
+const baseGetRequest = {
+    bucketName,
+    namespace,
+    headers: { host: '/' },
+};
+const baseUrl = `/${bucketName}`;
+
+const tests = [
```
ghost: So, we are not testing the "AccessDenied" case (i.e., whether Vault successfully handle this new API and permissions or not) here, right?

KillianG (approved):
lgtm

tests/unit/api/metadataSearch.js
ghost: Also, what about the `tests/functional/aws-node-sdk/test/mdSearch/basicSearch.js` file that implements some md tests? Is it related to this API?

ghost (changes requested):
First iteration, mostly comment and a request for change (regarding the reuse of already-defined functions)

tests/unit/api/metadataSearch.js
XinLi0207: > So, we are not testing the "AccessDenied" case (i.e., whether Vault successfully handle this new API and permissions or not) here, right?

I think no, interacting with Vault happens outside of this metadataSearch() function

XinLi0207: > Also, what about the `tests/functional/aws-node-sdk/test/mdSearch/basicSearch.js` file that implements some md tests? Is it related to this API?

this file is still related to this API, because we didn't change the way of requesting metadata search, it's still /<bucketname>?search=<search query> https://github.com/scality/cloudserver/blob/8fbba4e3c4f1ce2f31eddc0b57bd809eb1f95207/tests/functional/aws-node-sdk/test/mdSearch/utils/helpers.js#L27
so this test file is still suitable

lib/api/metadataSearch.js
XinLi0207: solved https://github.com/scality/cloudserver/pull/4224/commits/8a91dfef09659d6282a64e47ccc0a2200b3c3d82

lib/api/metadataSearch.js
XinLi0207: solved: https://github.com/scality/cloudserver/pull/4224/commits/8a91dfef09659d6282a64e47ccc0a2200b3c3d82

tests/unit/api/metadataSearch.js
ghost: Thanks!

lib/api/metadataSearch.js
```diff
@@ -0,0 +1,159 @@
...
+            const emptyList = {
+                CommonPrefixes: [],
+                Contents: [],
+                Versions: [],
+                IsTruncated: false,
+            };
+            return handleResult(listParams, requestMaxKeys, encoding, authInfo,
+                bucketName, emptyList, corsHeaders, log, callback);
+        }
+        if (params.search !== undefined) {
```
ghost: `params.search` is equal to `request.query.search`, because we already check that in arsenal, this condition should not be needed here

lib/api/metadataSearch.js
```diff
@@ -0,0 +1,159 @@
...
+                log.debug(err.message, {
+                    stack: err.stack,
+                });
+                monitoring.promMetrics(
+                    'GET', bucketName, 400, 'metadataSearch');
+                return callback(errors.InvalidArgument
+                    .customizeDescription('Invalid sql where clause ' +
+                        'sent as search query'));
+            }
+        }
```
ghost: Because here we are always running this code, since the actual function will only be called with `params.search` set, then maybe we can move this block of code above, so that we don't do unnecessary calculations if the `parseWhere` throws an error? (Maybe not, just suggesting!)

anurag4DSB (changes requested):
Haven't reviewed code yet.
But as this is for ARTESCA 1.4, it should go to Cloudserver development/8.4
Please rebase to development/8.4
If development/8.4 is not created, then we need to create it through RELENG.

tests/unit/api/metadataSearch.js
```diff
@@ -0,0 +1,258 @@
...
+        request: Object.assign({ query: {}, url: baseUrl }, baseGetRequest),
+        assertion: result => {
+            assert.strictEqual(result.ListBucketResult.Contents[1].Key[0],
+                objectName1);
+            assert.strictEqual(result.ListBucketResult.Contents[2].Key[0],
+                objectName2);
+            assert.strictEqual(result.ListBucketResult.Contents[0].Key[0],
+                objectName3);
+        },
+    },
```
ghost: As far as I understand this, we are re-defining tests we already have in `tests/unit/api/bucketGet.js`. While some are using search, most of them are not. Can we clean the `tests/unit/api/bucketGet.js` to remove the search tests and this file to remove the non-search tests?

See: https://github.com/scality/cloudserver/blob/8a91dfef09659d6282a64e47ccc0a2200b3c3d82/tests/unit/api/bucketGet.js#L71

ghost (changes requested):
2nd iteration!

lib/api/metadataSearch.js
XinLi0207: solved https://github.com/scality/cloudserver/pull/4224/commits/826e7bd7b4ad400430f80a5eec1d6651b1df38dc

lib/api/metadataSearch.js
XinLi0207: solved https://github.com/scality/cloudserver/pull/4224/commits/826e7bd7b4ad400430f80a5eec1d6651b1df38dc

tests/unit/api/metadataSearch.js
XinLi0207: thanks to your comment, I found in-memory db doesn't support metadata search. since we already have[ functional tests ](https://github.com/scality/cloudserver/tree/development/8.3/tests/functional/aws-node-sdk/test/mdSearch)and [search query unit tests](https://github.com/scality/cloudserver/blob/development/8.3/tests/unit/utils/validateSearch.js). I just removed the whole unit test file for metadata search api, there is no use to unit test things other than metadata search for this api.

ghost: Nice finding, and agree :slightly_smiling_face:

ghost (approved):
LGTM

ghost (approved):

anurag4DSB (approved):

XinLi0207:
ping

erwan-b:
ping

PR #4275, opened 2022-02-17
ft(CLDSRV-102): Add Aborted MPU PUT

tmacro:


lib/metadata/bucketfile/backend.js
```diff
@@ -190,6 +190,9 @@ class BucketFileInterface {
             if (err) {
                 return cb(err);
             }
+            if (params && params.isAbort && params.replayId) {
```
jonathan-gramain: Maybe comment here that we don't handle abort specifically but just ignore it.

Also any abort could be ignored, there should not be a need to check `params.replayId` (even though we normally always pass that attribute, it's an implementation detail).

jonathan-gramain (approved):

jonathan-gramain (approved):

naren-rajendran (approved):

PR #4328, opened 2022-03-07
Migrate dashboard to grafanalib

francoisferrand:
Issue: CLDSRV-147

### DIFF
--- a/monitoring/dashboard.py
+++ b/monitoring/dashboard.py
@@ -112,7 +112,8 @@ dataDiskStorage = TimeSeries(
 )
 
 
-def http_status_panel(title: str, code: str) -> Stat:
+def http_status_panel(title, code):
+    # type: (str, str) -> Stat
     return Stat(
         title=title,
         dataSource="${DS_PROMETHEUS}",
@@ -187,14 +188,16 @@ httpStatusCodes = TimeSeries(
 )
 
 
-def http_aggregated_request_target(title: str, code: str) -> Target:
+def http_aggregated_request_target(title, code):
+    # type: (str, str) -> Target
     return Target(
         expr='sum(increase(http_requests_total{namespace="${namespace}", job="${job}", code=' + code + "}[$__rate_interval]))",  # noqa: E501
         legendFormat=title,
     )
 
 
-def color_override(name: str, color: str):
+def color_override(name, color):
+    # type: (str, str) -> dict
     return {
         "matcher": {"id": "byName", "options": name},
         "properties": [{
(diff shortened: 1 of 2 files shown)


monitoring/dashboard.py
```diff
@@ -0,0 +1,412 @@
...
+activeRequests = Stat(
+    title="Active requests",
+    dataSource="${DS_PROMETHEUS}",
+    reduceCalc="lastNotNull",
+    targets=[Target(
+        expr='sum(http_active_requests{namespace="${namespace}", job="${job}"})',
+    )],
+    thresholds=[
+        Threshold("green", 0, 0.0),
+        Threshold("red",   1, 80.0),
```
lavigne958: Why 80  as a "red" value ?

monitoring/dashboard.py
```diff
@@ -0,0 +1,412 @@
...
+def average_latency_target(title: str, action: str = "") -> Target:
+    extra = ', action=' + action if action else ""
+    return Target(
+        expr="\n".join([
+            'sum(rate(http_request_duration_seconds_sum{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))",
+            "   /",
+            'sum(rate(http_request_duration_seconds_count{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))",
+        ]),
+        legendFormat=title,
+    )
```
lavigne958, line 242: a new line here could help code readability

monitoring/dashboard.py
```diff
@@ -0,0 +1,412 @@
...
+)
+
+def axisPlacement_override(name: str, mode: str):
+    return {
+        "matcher": {"id": "byName", "options": name},
+        "properties": [{
+            "id": "custom.axisPlacement", 
+            "value": mode,
+        }],
+    }
```
lavigne958, line 288: same here a new line could help to read code

monitoring/dashboard.py
francoisferrand: this is same as Vault, just an arbitrary limit I guess
It should generally stay low since it counts "in flight" requests only; but maybe 80 is normal in a loaded system.
Since it just changes the color of the panel, I guess this should be fine anyway : what do you think?

lavigne958 (approved):

jbertran (approved):

monitoring/dashboard.py
```diff
@@ -0,0 +1,452 @@
...
+            ConstantInput(
+                name="namespace",
+                label="namespace",
+                description="Namespace associated with the Zenko instance",
+                value="zenko",
+            ),
+            ConstantInput(
+                name="job",
+                label="job",
+                description="Name of the Vault job, used to filter the Vault "
```
alexanderchan-scality: the description mentions vault. is this correct?

francoisferrand: good catch, thanks!

PR #4342, opened 2022-03-16
Feature: CLDSRV-134 integrate new bucket format

welansari:
- bump arsenal to integrate new bucket format for mongo metadata backend
- Added CI steps to run functional tests in both bucket version formats
- Added test for when having buckets with different vFormats at the same time

### DIFF
--- a/tests/functional/metadata/MixedVersionFormat.js
+++ b/tests/functional/metadata/MixedVersionFormat.js
@@ -1,0 +1,196 @@
+const assert = require('assert');
+const async = require('async');
+const withV4 = require('../aws-node-sdk/test/support/withV4');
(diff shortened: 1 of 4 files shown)


francoisferrand (approved):

tests/functional/metadata/MixedVersionFormat.js
```diff
@@ -0,0 +1,196 @@
...
+                await updateBucketVFormat('v1-bucket', 'v1');
+            })
+            .catch(err => {
+                process.stdout.write('Error in afterEach');
+                throw err;
+            });
+        });
+
+        afterEach(() => {
+            process.stdout.write('Emptying buckets');
```
jbertran, line 94: Not sure we still need these `stdout`s

jbertran (approved):

PR #4350, opened 2022-03-17
Fix data-content overview panels

francoisferrand:
- Remove _local data storage_ panel, which is not relevant
- Use object/bucket count from report-handler
- Show status of report-handler as well as time elapsed since last report
- Show ingestion rate in addition to OOB-ingestion rate
- Use piecharts for _TOP10 buckets with errors_ panels
- Add description to summary panels

Screenshot 2022-03-17 at 15 46 40
Screenshot 2022-03-17 at 09 38 27

Issue: CLDSRV-149

### DIFF
--- a/monitoring/dashboard.json
+++ b/monitoring/dashboard.json
@@ -286,7 +286,7 @@
             "targets": [
                 {
                     "exemplar": true,
-                    "expr": "-sum(delta(cloud_server_data_disk_available{namespace=\"${namespace}\", job=\"${job}\"}[$__rate_interval]))",
+                    "expr": "-sum(deriv(cloud_server_data_disk_available{namespace=\"${namespace}\", job=\"${job}\"}[$__rate_interval]))",
                     "interval": "",
                     "legendFormat": "",
                     "refId": "A"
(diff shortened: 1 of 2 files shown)


monitoring/dashboard.json
```diff
@@ -285,23 +286,24 @@
             "targets": [
                 {
                     "exemplar": true,
-                    "expr": "sum(cloud_server_number_of_buckets{namespace=\"${namespace}\",job=\"${reportJob}\"})",
+                    "expr": "-sum(delta(cloud_server_data_disk_available{namespace=\"${namespace}\", job=\"${job}\"}[$__rate_interval]))",
```
francoisferrand: should be `deriv` instead of `delta`

francoisferrand:
@bert-e create_pull_requests

anurag4DSB (commented):
How can I try them out?
Can I just use the hash as tag for cloudserver in artesca 1.4.0?

anurag4DSB (approved):

ghost (approved):

PR #4361, opened 2022-03-23
Fix last report counter

francoisferrand:
- Formula must account for the date of last successful `count-items` cronjob execution. The expression is a bit complex, because:
   - we need to remove labels (hence the `max by()` and `on()`) so that prometheus can combine the metrics;
   - there is no "max" function between series, so we use `>0` as a filter to add a conditional operand.
- Unit is actually second.
- Display last value, even if null.

Issue: CLDSRV-149


ghost (approved):
LGTM

welansari (approved):

PR #4394, opened 2022-03-30
Fix success rate gauge

francoisferrand:
- Filtering out period with no http requests (NaN)
- Display greyed-out "-" when no data
- Compute mean value over current range
- Also fix Stat-panel value of http request counter

Issue: CLDSRV-164

### DIFF
--- a/monitoring/dashboard.py
+++ b/monitoring/dashboard.py
@@ -38,7 +38,8 @@ httpRequests = Stat(
     dataSource="${DS_PROMETHEUS}",
     decimals=0,
     format=UNITS.SHORT,
-    reduceCalc="last",
+    noValue="0",
+    reduceCalc="sum",
     targets=[Target(
         expr='sum(increase(http_requests_total{namespace="${namespace}", job="${job}"}[$__rate_interval]))',  # noqa: E501
     )],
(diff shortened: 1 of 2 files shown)


monitoring/dashboard.py
```diff
@@ -50,23 +51,24 @@
...
         ]),
-        instant=True,
         legendFormat="Success rate",
     )],
     thresholds=[
-        Threshold("red",    0, 0.0),
-        Threshold("orange", 1, 80.0),
-        Threshold("green",  2, 90.0),
+        Threshold("#808080", 0, 0.0),
+        Threshold("red",     1, 0.0),
```
ghost, line 69: Just to make sure I understand, the index 0 will be used when the value is `null` according to https://github.com/scality/cloudserver/pull/4394/files#diff-636a1e6d6da977cd13d38ab0bce8f93afcf8de9e18aae154dd76588edb205977R254, so why is the percentage set as 0 here?

francoisferrand: This is the "old" Grafana threshold configuration. First entry (index 0) is the default color, and the "value" is not used.

In this case, we set index 1 from 0.0 : so that grey is used when there is no value (null), ie not any query for some time (or since boot?)

ghost: Thanks! I understand

ghost (approved):

welansari (approved):

PR #4397, opened 2022-03-31
feature: CLDSRV-162 use metadata and data from arsenal backport

miniscruff:
Ready to review, forward branches have not been touched but will fix soon, can review though


lib/api/apiUtils/object/locationStorageCheck.js
```diff
@@ -0,0 +1,48 @@
...
+ * PUT operations, or will update metric on DELETE
+ * NOTE: storage limit may not be exactly enforced in the case of concurrent
+ * requests when near limit
+ * @param {string} location - name of location to check quota
+ * @param {number} updateSize - new size to check against quota in bytes
+ * @param {object} log - werelogs logger
+ * @param {function} cb - callback function
+ * @return {undefined}
+ */
+function locationStorageCheck(location, updateSize, log, cb) {
```
miniscruff, line 22: @vrancurel does this go into armory or stay in cloudserver? I am not sure it should go into armory as it would bind armory to utapi but also not sure it belongs here as it is strictly related to the multi backend data wrapper that armory should control?

vrancurel: if it brings armory to utapi then no.

nicolas2bert (approved):

nicolas2bert (changes requested):
Before merging, let's make sure the Integration build passes with this change and the Arsenal ones.

lib/api/multiObjectDelete.js
```diff
@@ -386,7 +386,7 @@ function multiObjectDelete(authInfo, request, log, callback) {
             return vault.checkPolicies(requestContextParams, authInfo.getArn(),
                 log, (err, authorizationResults) => {
                     // there were no policies so received a blanket AccessDenied
-                    if (err && err.AccessDenied) {
+                    if (err && err.is.AccessDenied) {
```
nicolas2bert, line 389: Since `VaultClient` repo has not been updated to use the latest arsenal errors, would this one throw error: `err.is is undefined`? @jbertran @miniscruff

nicolas2bert (commented):
This might not be the only place.

lib/api/multiObjectDelete.js
jbertran: I would expect it to. As a stopgap (with a ticket tracking it), you might write `err && err.code === errors.AccessDenied.code`

However, we might not catch these if we mock VaultClient behaviour in Cloudserver tests

vrancurel (approved):

jbertran (approved):
I've reviewed as much as I can, but I'm afraid I won't catch most of the changes that subsequently get lost in the noise of the move to Armory - one commit for all the changes is what I was trying to avoid with a single ticket/PR specifically for the breaking change catchup :slightly_frowning_face: 

Regardless, we need to move forward. We can pick up issues as they crop up, if they crop up.

miniscruff:
> I've reviewed as much as I can, but I'm afraid I won't catch most of the changes that subsequently get lost in the noise of the move to Armory - one commit for all the changes is what I was trying to avoid with a single ticket/PR specifically for the breaking change catchup slightly_frowning_face
> 
> Regardless, we need to move forward. We can pick up issues as they crop up, if they crop up.

Sorry for the one massive commit, I made so many tiny ones testing things that it was just a message to rebase any better than one massive squash. Ideally this would of been split up a little more.

miniscruff:
Duplicate of #4548

PR #4411, opened 2022-04-06
S3C-5390 s3api head-object with part-number 1 on empty file fails: ht…

watsaqat:
s3api head-object with part-number 1 on empty file fails: httpCode 416

### DIFF
--- a/lib/api/objectHead.js
+++ b/lib/api/objectHead.js
@@ -124,10 +124,12 @@ function objectHead(authInfo, request, log, callback) {
                     return callback(errors.BadRequest, corsHeaders);
                 }
                 const partSize = getPartSize(objMD, partNumber);
-                if (!partSize) {
+                const isEmptyObject = objLength === 0;
+                if (!partSize && !isEmptyObject) {
                     return callback(errors.InvalidRange, corsHeaders);
                 }
-                responseHeaders['content-length'] = partSize;
+
+                responseHeaders['content-length'] = isEmptyObject ? 0 : partSize;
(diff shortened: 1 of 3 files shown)


tests/functional/aws-node-sdk/test/object/getPartSize.js
```diff
@@ -26,128 +15,154 @@ function checkNoError(err) {
...
-                        PartNumber: partNumber + 1,
-                        UploadId: this.currentTest.UploadId,
-                        Body: generateContent(partNumber + 1),
-                    };
-
-                    return s3.uploadPart(uploadPartParams,
-                        (err, data) => {
-                            if (err) {
-                                return callback(err);
+describe.only('Part size tests with object head', () => {
```
rahulreddy: Please remove `.only`

tmacro (approved):

alexanderchan-scality (approved):

lib/api/objectHead.js
```diff
@@ -124,10 +124,12 @@ function objectHead(authInfo, request, log, callback) {
                     return callback(errors.BadRequest, corsHeaders);
                 }
                 const partSize = getPartSize(objMD, partNumber);
-                if (!partSize) {
+                const isEmptyObject = objLength === 0;
+                if (!partSize && !isEmptyObject) {
```
jonathan-gramain, line 128: Maybe we should additionally enforce that the requested part is 1 in the empty object case, and return 416 otherwise (to match AWS behavior which is akin to considering an empty object as having a single empty part)

PR #4417, opened 2022-04-11
bugfix: CLDSRV-170 skip orphan cleanup in UploadPart[Copy]

jonathan-gramain:
Do not delete orphan data in UploadPart/UploadPartCopy on overwrite
iff a CompleteMPU of the target MPU is already in progress.

This is to prevent a race condition where a CompleteMPU is running
while UploadPart is uploading a part for the same MPU.

It leaves an orphan in storage since only one of the upload data will
be present in the finished MPU, but the window is limited to the
CompleteMPU execution and should only occur when there are retries of
UploadPart due to prior stuck requests, or with broken clients
misusing the MPU API, so it should be acceptable.

Implementation details:

- set a flag in the MPU overview key when starting the CompleteMPU
  process, before listing the parts from metadata to construct the
  manifest

- in UploadPart/UploadPartCopy, after the part metadata is written and
  if the same part already existed, re-fetch the MPU overview key to
  check the flag: if set, skip the deletion of the old data of this
  part, since the CompleteMPU process in progress may choose either
  part data depending on the exact timing of the listing vs. the
  part overwrite.

### DIFF
--- a/lib/api/objectPutPart.js
+++ b/lib/api/objectPutPart.js
@@ -12,6 +12,7 @@ const kms = require('../kms/wrapper');
 const metadata = require('../metadata/wrapper');
 const { pushMetric } = require('../utapi/utilities');
 const logger = require('../utilities/logger');
+const services = require('../services');
 const { config } = require('../Config');
 const multipleBackendGateway = require('../data/multipleBackendGateway');
 const locationConstraintCheck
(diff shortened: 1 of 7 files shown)


lib/api/objectPutPart.js
```diff
@@ -336,19 +337,55 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                         });
                         return next(err, destinationBucket);
                     }
-                    return next(null, oldLocations, objectLocationConstraint,
+                    return next(null, partLocations, oldLocations, objectLocationConstraint,
+                        destinationBucket, hexDigest, prevObjectSize, splitter);
+                });
+        },
+        (partLocations, oldLocations, objectLocationConstraint, destinationBucket,
```
tmacro, line 345: It might make sense to move this into a helper so that we can us it in PutCopyPart as well rather than duplicating the logic. The wrapper code that would be needed may outweigh the benefits though.

tmacro (approved):

lib/api/objectPutPart.js
jonathan-gramain: Yeah, I can do a `isCompleteMPUInProgress` in services for example. I would likely still have to duplicate the "not deleting locations" log since creating a helper to encapsulate the deletion of the parts would be too convoluted in the current logic or would require a larger refactor to encapsulate more logic from `objectPutPart`/`objectPutCopyPart`.

jonathan-gramain: done

vrancurel (approved):
do we check the Etags at CompleteMPU ?

jonathan-gramain:
@vrancurel yes, it's already done in the current implementation (and there are tests for it as well)

jonathan-gramain:
ping

tcarmet:
ping

jonathan-gramain:
ping

PR #4444, opened 2022-04-13
CLDSRV-173 DeleteMarkers created by Lifecycle should not be replicated

nicolas2bert:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-what-is-isnot-replicated.html
Changes needed to be compliant with Amazon:
"if AWS S3 deletes an object due to a lifecycle action, the delete marker is not replicated to the destination buckets."

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -1,7 +1,8 @@
 const { evaluators, actionMaps, RequestContext } = require('arsenal').policies;
 const constants = require('../../../../constants');
 
-const { allAuthedUsersId, bucketOwnerActions, logId, publicId } = constants;
+const { allAuthedUsersId, bucketOwnerActions, logId, publicId,
+    assumedRoleArnResourceType, backbeatLifecycleSessionName } = constants;
 
 // whitelist buckets to allow public read on objects
 const publicReadBuckets = process.env.ALLOW_PUBLIC_READ_BUCKETS ?
@@ -364,10 +365,34 @@ function validatePolicyResource(bucketName, policy) {
     });
 }
 
+/** isLifecycleSession - check if it is the Lifecycle assumed role session arn.
+ * @param {string} arn - Amazon resource name - example:
+ * arn:aws:sts::257038443293:assumed-role/rolename/backbeat-lifecycle
+ * @return {boolean} true if Lifecycle assumed role session arn, false if not.
+ */
+function isLifecycleSession(arn) {
+    if (!arn) {
+        return false;
+    }
+
+    const arnSplits = arn.split(':');
+    const service = arnSplits[2];
+
+    const resourceNames = arnSplits[arnSplits.length - 1].split('/');
+
+    const resourceType = resourceNames[0];
+    const sessionName = resourceNames[resourceNames.length - 1];
+
+    return (service === 'sts' &&
+        resourceType === assumedRoleArnResourceType &&
+        sessionName === backbeatLifecycleSessionName);
+}
+
 module.exports = {
     isBucketAuthorized,
     isObjAuthorized,
     checkBucketAcls,
     checkObjectAcls,
     validatePolicyResource,
+    isLifecycleSession,
 };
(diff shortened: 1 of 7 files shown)


nicolas2bert:
ping

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -364,10 +365,36 @@ function validatePolicyResource(bucketName, policy) {
...
+ */
+function isLifecycleSession(arn) {
+    if (!arn) {
+        return false;
+    }
+
+    const arnSplits = arn.split(':');
+    const service = arnSplits[2];
+
+    const ressourceNames = arnSplits[arnSplits.length - 1].split('/');
```
miniscruff:
```suggestion
    const resourceNames = arnSplits[arnSplits.length - 1].split('/');
```
one too many s's

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -364,10 +365,36 @@ function validatePolicyResource(bucketName, policy) {
...
+
+    const ressourceType = ressourceNames[0];
```
miniscruff:
```suggestion
    const resourceType = ressourceNames[0];
```

miniscruff (changes requested):

lib/api/apiUtils/authorization/permissionChecks.js
nicolas2bert: Thanks :)

miniscruff (approved):

lib/api/apiUtils/object/getReplicationInfo.js
```diff
@@ -63,14 +64,21 @@ function _getReplicationInfo(rule, replicationConfig, content, operationType,
...
  * @return {undefined}
  */
 function getReplicationInfo(objKey, bucketMD, isMD, objSize, operationType,
-    objectMD) {
+    objectMD, authInfo) {
     const content = isMD || objSize === 0 ? ['METADATA'] : ['DATA', 'METADATA'];
     const config = bucketMD.getReplicationConfiguration();
     // If bucket does not have a replication configuration, do not replicate.
     if (config) {
+        // If delete an object due to a lifecycle action,
```
rachedbenmustapha, line 77: Shouldn't we check for operation type? In theory nothing lifecycle does should be replicated (unless configured, seems AWS has that option now), but we're targeting deletes from expiration right now - while an object that we PUT from a transition workflow would be expected to replicate if a policy is present.

nicolas2bert: I think Lifecycle Transition is using backbeatclient.multipleBackendPutObject() request to “PUT object”. 
The request then gets proxied to CloudServer and  Cloudserver uses its own “multiple backend” clients to put the object in the right transitioned location. 
So Cloudserver won’t be able to retrieve the "lifecycle assumed role" session name in that case and the PUT should pass the check and succeed.
Also, I am not sure if it is necessary to add a safeguarding check that checks for “DELETE” method or checks that `isDeleteMarker` property is true.

nicolas2bert: https://github.com/scality/backbeat/blob/development/8.3/extensions/replication/tasks/CopyLocationTask.js#L289

naren-rajendran (approved):

lib/api/apiUtils/object/getReplicationInfo.js
rachedbenmustapha: I don't know if the backbeat-specific handlers on cloudserver would understand any types of credentials besides old service account (assuming they would), but backbeat's transition is only using old-style credentials because we haven't made the change yet, in the next few weeks in artesca we'll do the same for transition as we did for expiration to use assumed role credentials. So in theory they would match lifecycle actions.

For DELETE op vs delete marker, I'm not sure, it seems plain DELETEs should never be replicated, and delete markers should be replicated only if not from lifecycle (described [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-what-is-isnot-replicated.html#replication-what-is-replicated) as we don't use the new-format replication configuration) - so checking for `isDeleteMarker` in conjunction with lifecycle would be enough.

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #4451, opened 2022-04-15
CLDSRV-167: implement put bucket tagging

KillianG:
Checking order has been changed to fit with AWS one, meaning that we
raise the first error we face

- Checking every tag one by one
- Checking that the tag key is unique, the error is raised only if we
  already seen a tag with the same key before
- Checking that the key length is less than 128
- Checking that the tag value length is less than 256
- After seeing more than 50 tags, if there is one more, raise the too
  many tags error


XinLi0207 (commented):
LGTM, just a small question: Is there a max number of tags that can be added to a bucket or it's unlimited?

lib/api/bucketPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+function checkTags(tagSet, bucket, log, cb) {
+    if (tagSet === undefined) {
+        return cb(null, { tags: [], bucket });
+    }
+    const tags = Array.isArray(tagSet) ? tagSet : [tagSet];
+    const unique = [...new Set(tags.map(item => item.Key))];
+    let errMsg = null;
+    if (unique.length !== tags.length) {
+        errMsg = tagKeyNotUnique;
+    }
```
ghost: Not tested, but I don't think it's needed to convert the Set to array just to get its length:
```suggestion
    let errMsg = null;
    if (new Set(tags.map(item => item.Key)).size !== tags.length) {
        errMsg = tagKeyNotUnique;
    }
```

lib/api/bucketPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+        bucketName,
+        requestType: 'bucketPutTagging',
+    };
+    return waterfall([
+        next => _parseXML(request, log, next, parser),
+        (result, next) => metadataValidateBucket(metadataValParams, log,
+            (err, bucket) => next(err, { result, bucket })),
+        ({ result, bucket }, next) => checkTags(result.Tagging.TagSet.Tag, bucket, log, next),
+        ({ tags, bucket }, next) => {
+            bucket.setTags(tags);
```
ghost: Random question, the only place I found this function was in ObjectMD (in Arsenal), but the function overwrites all the previously-set tags with the new tags. In your tests, can you try to put two tags (one by one) and check, after the two PutBucketTagging, that both tags are in the bucket MD?

ghost:
@XinLiScality yes, according to https://docs.aws.amazon.com/AmazonS3/latest/userguide/tagging-managing.html, it's 10. Do we have the same limits in the requirements?
Edit: maybe it's only for object tagging?

tests/functional/aws-node-sdk/test/bucket/putBucketTagging.js
```diff
@@ -0,0 +1,177 @@
...
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+        return done();
+    });
+
+    beforeEach(done => s3.createBucket({ Bucket: bucket }, done));
+
+    afterEach(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+    it('should not add tag if tagKey not unique', done => {
```
XinLi0207, line 109: do you think it's worth testing if a bucket tagging can overwrite the one with the same key?

lib/api/bucketPutTagging.js
ghost: Didn't see the api note: "When this operation sets the tags for a bucket, it will overwrite any current tags the bucket already has. You cannot use this operation to add tags to an existing list of tags.", so all good

tests/functional/aws-node-sdk/test/bucket/putBucketTagging.js
ghost: A PutBucketTagging will always overwrite any previous tags. See note in https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-bucket-tagging.html

We can test that scenario, maybe?

lib/api/bucketPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+            metadata.updateBucket(bucket.getName(), bucket, log, err =>
+                next(err, bucket));
+        },
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        return callback(err, corsHeaders);
+    });
+}
```
ghost:
```suggestion
    ], (err, bucket) => {
        const corsHeaders = collectCorsHeaders(request.headers.origin,
            request.method, bucket);
        pushMetric('putBucketTagging', log, {
            authInfo,
            bucket: bucketName,
        });
        return callback(err, corsHeaders);
    });
}
```

ghost (commented):
Suggest adding the pushMetric call in the API

lib/api/bucketPutTagging.js
```diff
@@ -0,0 +1,104 @@
...
+        errMsg = tagKeyNotUnique;
+    }
+    tags.forEach(item => {
+        if (item.Key.length > 128) {
+            errMsg = tagKeyInvalid;
+        }
+        if (item.Value.length > 256) {
+            errMsg = tagValueInvalid;
+        }
+    });
```
anurag4DSB: As per this we check all the tagsets and return the last known error.
In AWS, I tried the same, it returns on the first error that was found.


For example, I tried the below JSON:
```json
{
    "TagSet": [
      {
        "Key": "organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1",
        "Value": "marketing"
      },
      {
        "Key": "o121",
        "Value": "rganization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1organization1"
      }
    ]
 }
```

Cloudserver (current PR code) gives me invalid value:
```sh
❯ aws s3api put-bucket-tagging --bucket test1 --tagging file://tagset.json --profile vault --endpoint-url http://localhost:8000

An error occurred (InvalidTag) when calling the PutBucketTagging operation: The TagValue you have provided is invalid
```

Where as AWS gives me invalid Tag Key:
```sh
❯ aws s3api put-bucket-tagging --bucket anuragplatform --tagging file://tagset.json

An error occurred (InvalidTag) when calling the PutBucketTagging operation: The TagKey you have provided is invalid
```

lib/api/bucketPutTagging.js
ghost: PutBucketTagging is meant to do exactly that: overwrite all tags each time.

tests/functional/aws-node-sdk/test/bucket/putBucketTagging.js
ghost: It does this by default, all tags are overwritten every time a tagset is uploaded

anurag4DSB (changes requested):
I also tried uploading a TagSet of more than 50 tags, AWS sent me back an error, but cloudserver did not
```
❯ aws s3api put-bucket-tagging --bucket anuragplatform --tagging file://tagset.json

An error occurred (BadRequest) when calling the PutBucketTagging operation: Bucket tag count cannot be greater than 50
❯ aws s3api put-bucket-tagging --bucket test1 --tagging file://tagset.json --profile vault --endpoint-url http://localhost:8000
```
We should adhere to the restrictions mentioned in requirements: https://github.com/scality/citadel/blob/development/1.0/docs/requirements/bucket-tagging-s3c.md#restrictions-on-tagging

lib/api/bucketPutTagging.js
ghost: you're right, done dee6e21fdf6c2fef04ad5784060fb2be8d5ef149

lib/api/bucketPutTagging.js
ghost: done dee6e21fdf6c2fef04ad5784060fb2be8d5ef149

lib/api/bucketPutTagging.js
ghost: for the tag limit, that was indeed something Killian and I missed in the code, added here dee6e21fdf6c2fef04ad5784060fb2be8d5ef149

As for the ordering of things, there's some evaluation to do.
For example, AWS sends the tags back in what seems to be a random but consistent order, whereas right now we send the tags back in the order they were received.

For the ordering of Errors, I'll look into it.

lib/api/bucketPutTagging.js
```diff
@@ -0,0 +1,121 @@
...
+
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const metadata = require('../metadata/wrapper');
+const { pushMetric } = require('../utapi/utilities');
+
+const tagKeyNotUnique = 'Cannot provide multiple Tags with the same key';
+const tagKeyInvalid = 'The TagKey you have provided is invalid';
+const tagValueInvalid = 'The TagValue you have provided is invalid';
+const tooManyTags = 'Bucket tag count cannot be greater than 50'
```
ghost: Nit:

```suggestion
const tooManyTags = 'Bucket tag count cannot be greater than 50';
```

ghost: Otherwise LGTM (maybe the `50` could be stored as a variable though)

ghost: Should've done a lint before, apologies. Will fix in next push

ghost:
As per what I've tested with AWS, error prioritization is now as I've observed:
- failing on the first error it comes across as it goes through through tags one by one (including unique keys: a dupe key-pair could be at positions 1 and 49, but the error received would be TagValue invalid if an invalid tag is at position 48)
- THEN checking tag count once it's evaluated all the tags (no matter array size, as inefficient as that is...)

So as much as I hate pushing such inefficient code, here we are 090e8a290b6adf2d69bce53dc98eefe539751b33

Please do make any suggestions to optimize this !

lib/api/bucketPutTagging.js
ghost: See comment here for how I fixed error prioritisation https://github.com/scality/cloudserver/pull/4451#issuecomment-1108687268

lib/api/bucketPutTagging.js
ghost: done 090e8a290b6adf2d69bce53dc98eefe539751b33

XinLi0207 (approved):

ghost (approved):

anurag4DSB:
> As per what I've tested with AWS, error prioritization is now as I've observed:
> 
> * failing on the first error it comes across as it goes through through tags one by one (including unique keys: a dupe key-pair could be at positions 1 and 49, but the error received would be TagValue invalid if an invalid tag is at position 48)
> * THEN checking tag count once it's evaluated all the tags (no matter array size, as inefficient as that is...)
> 
> So as much as I hate pushing such inefficient code, here we are [090e8a2](https://github.com/scality/cloudserver/commit/090e8a290b6adf2d69bce53dc98eefe539751b33)
> 
> Please do make any suggestions to optimize this !

@WillToozsScality Do you see similar for AWS-SDK as well. There is a possibility that AWS CLI is doing something than what just BOTO3 does. 
Also, as per the current code we will go through all possible tags, which is expensive ofcourse like you mentioned.
I tried AWS and they do have some kind of rate limiting post 3500 hundred tags(depends on the content of key-value).

Either we figure out what AWS does for rate-limiting(more below, might need some sort of design if not already present), or we put the condition of checking the number of tags first(error out if more than 50) and move the rest of the checking logic after that.


AWS behavior with 3500+ tags:

```
❯ aws s3api put-bucket-tagging --bucket anuragplatform --tagging file://tagset2.json

An error occurred (MalformedXML) when calling the PutBucketTagging operation: The XML you provided was not well-formed or did not validate against our published schema
```

TagSet used:
```sh
{
    "TagSet":
	[{"Key":"a0","Value":"a0"},{"Key":"a1","Value":"a1"},{"Key":"a2","Value":"a2"},{"Key":"a3","Value":"a3"},{"Key":"a4","Value":"a4"},{"Key":"a5","Value":"a5"},{"Key":"a6","Value":"a6"},{"Key":"a7","Value":"a7"},{"Key":"a8","Value":"a8"},{"Key":"a9","Value":"a9"},{"Key":"a10","Value":"a10"},
.
.
.
{"Key":"a3545","Value":"a3545"},{"Key":"a3546","Value":"a3546"},{"Key":"a3547","Value":"a3547"},{"Key":"a3548","Value":"a3548"},{"Key":"a3549","Value":"a3549"},{"Key":"a3550","Value":"a3550"},{"Key":"a3551","Value":"a3551"},{"Key":"a3552","Value":"a3552"},{"Key":"a3553","Value":"a3553"},{"Key":"a3554","Value":"a3554"},{"Key":"a3555","Value":"a3555"},{"Key":"a3556","Value":"a3556"},{"Key":"a3557","Value":"a3557"},{"Key":"a3558","Value":"a3558"},{"Key":"a3559","Value":"a3559"},{"Key":"a3560","Value":"a3560"}]
}
```

For the same tagSet we check everything:
```
❯ aws s3api put-bucket-tagging --bucket test1 --tagging file://tagset2.json --endpoint-url http://localhost:8000 --profile vault

An error occurred (BadRequest) when calling the PutBucketTagging operation: Bucket tag count cannot be greater than 50
```

anurag4DSB (changes requested):
Requesting for last bit of changes.
Would recommend, we put the condition of checking the number of tags first(error out if more than 50) and move the rest of the checking logic after that.
I know this is not what AWS does, but post a certain value, they do some rate limiting.

Lastly, AWSCLI provides this option: `[--expected-bucket-owner <value>]`
Do we support that? If not then we need to do that as well.

KillianG:
> Requesting for last bit of changes. Would recommend, we put the condition of checking the number of tags first(error out if more than 50) and move the rest of the checking logic after that. I know this is not what AWS does, but post a certain value, they do some rate limiting.

This is done now [here](https://github.com/scality/cloudserver/pull/4451/commits/aab111506a9d3e68df36cf9c058b8301f24fc960)

> Lastly, AWSCLI provides this option: `[--expected-bucket-owner <value>]` Do we support that? If not then we need to do that as well.

We do support that now [here](https://github.com/scality/cloudserver/pull/4451/commits/b29756602ba8c5671672f8b0b4d1c25d34cc80c4)

lib/api/bucketPutTagging.js
```diff
@@ -58,6 +58,16 @@ function checkTags(tagSet, bucket, log, cb) {
...
+        if (i > 50) {
+            log.debug(tooManyTags,
+                {
+                    method: 'bucketPutTagging',
+                    error: errors.BadRequest,
+                });
+            const error = errors.BadRequest.customizeDescription(
+                tooManyTags);
+            return cb(error);
+        }
```
ghost: While in the end this will still be an improvement, when Anurag asked to put it before the other tests, he likely meant before the loop. Either way I'm sure it's not a big deal at this point.

ghost (approved):

lib/api/bucketPutTagging.js
KillianG: I checked, and AWS raises the first error it faces, so if I check for 50 tags before anything else we might end up showing "error more than 50 tags" before showing "invalid tag".

By doing this way, we check all errors on the first 50 tags and if there is no errors, but there is more than 50 tags, we raise the "more than 50 tags error" without going further than 50, so if there is 999.999.999.999 tags, and they are all valid, we only check the first 50

lib/api/bucketPutTagging.js
```diff
@@ -0,0 +1,129 @@
...
```
anurag4DSB: This logic deserves an in-code comment.
And a commit message story(while squashing commits)
A small example: https://github.com/scality/Vault/commit/7c73cc53b673665344329bd83ff5633f29c6e138
But you can even make bullet points. The only requirement is the first line should comply with bert-e requirements.

anurag4DSB (changes requested):
the code LGTM, but we need more in code comments and a bit more detailed story in the commit message.

lib/api/bucketPutTagging.js
anurag4DSB: feel free to squash for the next review :)

KillianG: Squashed and fixed [here](https://github.com/scality/cloudserver/pull/4451/commits/ea107a50e15842ac0a046b10bcb55ff4bc2e773d)

anurag4DSB (approved):
LGTM, I would still say deserves a in-code comment :) 
Specially where we do not comply with AWS

KillianG:
> LGTM, I would still say deserves a in-code comment :) Specially where we do not comply with AWS

AFAIK, we totally comply with AWS.. @WillToozsScality can you confirm ?

PR #4458, opened 2022-04-19
feature: CLDSRV-175 add support for s3:ObjectACL:Put notification

welansari:
Issue: [CLDSRV-175](https://scality.atlassian.net/browse/CLDSRV-175)

### DIFF
--- a/lib/metadata/acl.js
+++ b/lib/metadata/acl.js
@@ -12,12 +12,40 @@ const acl = {
         metadata.updateBucket(bucket.getName(), bucket, log, cb);
     },
 
+    /**
+     * returns true if the specified ACL grant is unchanged
+     * @param {string} grant name of the grant
+     * @param {object} oldAcl old acl config
+     * @param {object} newAcl new acl config
+     * @returns {bool} is the grant the same
+     */
+    _aclGrantDidNotChange(grant, oldAcl, newAcl) {
+        if (grant === 'Canned') {
+            return oldAcl.Canned === newAcl.Canned;
+        }
+        /**
+         * An ACL grant is in form of an array of strings
+         * An ACL grant is considered unchanged when both the old and new one
+         * contain the same number of elements, and all elements from one
+         * grant are incuded in the other grant
+         */
+        return oldAcl[grant].length === newAcl[grant].length
+            && oldAcl[grant].every(value => newAcl[grant].includes(value));
+    },
+
     addObjectACL(bucket, objectKey, objectMD, addACLParams, params, log, cb) {
         log.trace('updating object acl in metadata');
-        // eslint-disable-next-line no-param-reassign
-        objectMD.acl = addACLParams;
-        metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params, log,
-            cb);
+        const isAclUnchanged = Object.keys(objectMD.acl).length === Object.keys(addACLParams).length
+            && Object.keys(objectMD.acl).every(grant => this._aclGrantDidNotChange(grant, objectMD.acl, addACLParams));
+        if (!isAclUnchanged) {
+            /* eslint-disable no-param-reassign */
+            objectMD.acl = addACLParams;
+            objectMD.originOp = 's3:ObjectAcl:Put';
+            /* eslint-disable no-param-reassign */
+            return metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params, log,
+                cb);
+        }
+        return cb();
     },
 
     parseAclFromHeaders(params, cb) {
@@ -129,3 +157,4 @@ const acl = {
 };
 
 module.exports = acl;
+
--- a/tests/unit/api/objectPutACL.js
+++ b/tests/unit/api/objectPutACL.js
@@ -461,4 +461,116 @@ describe('putObjectACL API', () => {
                 });
         });
     });
+
+    [
+        { headers: { 'x-amz-acl': 'public-read-write' }, type: 'ACL' },
+        { headers: { 'x-amz-grant-full-control': 'emailaddress="sampleaccount1@sampling.com"' }, type: 'FULL_CONTROL' },
+        { headers: { 'x-amz-grant-read': `uri=${constants.logId}` }, type: 'READ' },
+        { headers: { 'x-amz-grant-read-acp': `id=${ownerID}` }, type: 'READ_ACP' },
+        { headers: { 'x-amz-grant-write-acp': `id=${anotherID}` }, type: 'WRITE_ACP' },
+        { headers: {
+            'x-amz-grant-full-control': 'emailaddress="sampleaccount1@sampling.com"',
+            'x-amz-grant-read': `uri=${constants.logId}`,
+            'x-amz-grant-read-acp': `id=${ownerID}`,
+            'x-amz-grant-write-acp': `id=${anotherID}`,
+        }, type: 'ALL' },
+        { headers: {
+            'x-amz-grant-read': `uri=${constants.logId}`,
+            'x-amz-grant-read-acp': `id=${ownerID}`,
+            'x-amz-grant-write-acp': `id=${anotherID}`,
+        }, type: 'READ/READ_ACP/WRITE_ACP' },
+        { headers: {
+            'x-amz-grant-full-control': 'emailaddress="sampleaccount1@sampling.com"',
+            'x-amz-grant-read-acp': `id=${ownerID}`,
+            'x-amz-grant-write-acp': `id=${anotherID}`,
+        }, type: 'FULL/READ_ACP/WRITE_ACP' },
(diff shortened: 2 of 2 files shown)


lib/metadata/acl.js
```diff
@@ -14,10 +14,21 @@ const acl = {
...
-        objectMD.acl = addACLParams;
-        metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params, log,
-            cb);
+        function _didNotChange(grant) {
+            return objectMD.acl[grant].length === addACLParams[grant].length
+                && objectMD.acl[grant].every(value => addACLParams[grant].includes(value));
+        }
+        const cannedChanged = objectMD.acl.Canned !== addACLParams.Canned;
+        if (cannedChanged || !_didNotChange('FULL_CONTROL') || !_didNotChange('WRITE_ACP')
+            || !_didNotChange('READ') || !_didNotChange('READ_ACP')) {
```
francoisferrand: We should not hardcode a check for each grant.
If there is a list of all "allowed" grants, we iterate over it ; however, since this function does not do any such check, it would be better to just compare the keys in both "objectMD.acl" and "addACLParams" : if both don't have the same keys (whatever they may be), do the update ; else check for each key it still matches.

lib/metadata/acl.js
```diff
@@ -14,10 +14,21 @@ const acl = {
...
+            /* eslint-disable no-param-reassign */
+            objectMD.acl = addACLParams;
```
francoisferrand, line 42: this _replaces_ the ACL : so it can actually also remove grants.... and is not consistent with the check in line 19 (`newAcl.includes(oldAcl)`)

lib/metadata/acl.js
```diff
@@ -14,10 +14,21 @@ const acl = {
 
     addObjectACL(bucket, objectKey, objectMD, addACLParams, params, log, cb) {
         log.trace('updating object acl in metadata');
-        // eslint-disable-next-line no-param-reassign
-        objectMD.acl = addACLParams;
-        metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params, log,
-            cb);
+        function _didNotChange(grant) {
+            return objectMD.acl[grant].length === addACLParams[grant].length
+                && objectMD.acl[grant].every(value => addACLParams[grant].includes(value));
```
francoisferrand: not sure this condition is right : if the new ACL includes the old one, it does not mean it did not change, just that nothing was removed...

if the function did only "adding" ACL (and we were to merge the existing ACL with the new one), I would expect the condition to be the other way round : no change if the new ACL are "included" in the existing ones already...

That said, I am not sure what is expected exacty of this API : so i don't know which behavior should be expected.

tests/unit/api/objectPutACL.js
```diff
@@ -461,4 +461,58 @@ describe('putObjectACL API', () => {
...
     });
+
+    it('should set originOp to s3:ObjectAcl:Put when ACL is changed', done => {
+        const testObjACLRequest = {
+            bucketName,
+            namespace,
+            objectKey: objectName,
+            headers: { 'x-amz-acl': 'public-read-write' },
+            url: `/${bucketName}/${objectName}?acl`,
+            query: { acl: '' },
```
francoisferrand: we may want to run this test with different kind of updates : i.e. testing all paths of the condition on line 22 in acl.js :
```
if (cannedChanged || !_didNotChange('FULL_CONTROL') || !_didNotChange('WRITE_ACP')
            || !_didNotChange('READ') || !_didNotChange('READ_ACP')) 
```

lib/metadata/acl.js
Kerkesni: We have five cases here for each grant:
- new ACL adds new grantees => line 18 would catch that (we check the length of both arrays)
- new ACL removes new grantees => line 18 would also catch that
- new ACL does not change anything => both tests will pass
- new ACL replaces grantees (both addition and deletion is made) => we may have the same number of grantees in some cases but the values would be different and so this will be caught in line 19

welansari: if both grants (old and new one) contain the same number of elements and if all elements in one of them are included in the second, that means both grants contain the same elements, so if we add new grantees this will be caught in line 18

francoisferrand: indeed, you are right.

that said, if we need to "think" if the condition is right, then maybe there is something to improve : make implementation easier to read and/or at least a comment...

lib/metadata/acl.js
```diff
@@ -15,12 +15,14 @@ const acl = {
...
+            if (grant === 'Canned') {
+                return objectMD.acl.Canned === addACLParams.Canned;
+            }
             return objectMD.acl[grant].length === addACLParams[grant].length
                 && objectMD.acl[grant].every(value => addACLParams[grant].includes(value));
         }
-        const cannedChanged = objectMD.acl.Canned !== addACLParams.Canned;
-        if (cannedChanged || !_didNotChange('FULL_CONTROL') || !_didNotChange('WRITE_ACP')
-            || !_didNotChange('READ') || !_didNotChange('READ_ACP')) {
+        const isAclUnchanged = Object.keys(objectMD.acl).every(grant => _didNotChange(grant));
```
francoisferrand: that may not be correct: e.g. if `addACLParams` contains grants which are not present in `objectMD.acl`...
(until we have typescript to prove otherwise)

francoisferrand: also, are we sure this field is always set (e.g. can `objectMD.acl` be null) ?

tests/unit/api/objectPutACL.js
```diff
@@ -462,30 +462,38 @@ describe('putObjectACL API', () => {
...
-                            assert.strictEqual(md.originOp,
-                            's3:ObjectAcl:Put');
-                            done();
+    [
+        { headers: { 'x-amz-acl': 'public-read-write' }, type: 'ACL' },
+        { headers: { 'x-amz-grant-full-control': 'emailaddress="sampleaccount1@sampling.com"' }, type: 'FULL_CONTROL' },
+        { headers: { 'x-amz-grant-read': `uri=${constants.logId}` }, type: 'READ' },
+        { headers: { 'x-amz-grant-read-acp': `id=${ownerID}` }, type: 'READ_ACP' },
+        { headers: { 'x-amz-grant-write-acp': `id=${anotherID}` }, type: 'WRITE_ACP' },
+    ].forEach(params => {
```
francoisferrand, line 521: don't we need to also check combinations of grants ?

note that maybe it would be easier to test the various conditions/paths by calling `addObjectACL` directly (while keeping a separate test for the objectPutACL API itself, to check if the API does the correct parsing/plumbing/error handling/... ?

lib/metadata/acl.js
welansari: When parsing the params from headers the function only considers the supported grants https://github.com/scality/cloudserver/blob/fe923f679decd55ffd0db7455910cc33adc0c2c1/lib/api/objectPutACL.js#L101-L109

Otherwise when the config is in a json format the function skips unsupported grants https://github.com/scality/cloudserver/blob/fe923f679decd55ffd0db7455910cc33adc0c2c1/lib/api/objectPutACL.js#L187-L189

And `objectMD.acl` is always initialised, and can never be removed, we always have all the grant fields even when empty

francoisferrand: This function is not objectPutACL, but in a separate module : so we should not make expectations on the calling code.
The only expectation is on the "contract" of the module (its interface) : e.g. what it expects, ...

For a typed-language, this is easy ; obviously here with javascript this is all blurry... The code in https://github.com/scality/cloudserver/blob/fe923f679decd55ffd0db7455910cc33adc0c2c1/lib/api/objectPutACL.js#L93-L99 kind of introduces such a type, but this is not used in other places... and there is always a risk of mistake (e.g. typo, or "new" field added in one place only...)

This will be much safer once we move to typescript, but until them, better avoid making too many assumptions, esp. when it can be fixed easily: e.g. if you just do an extra check on the number of keys we should be safe in any case...

> And objectMD.acl is always initialised, and can never be removed, we always have all the grant fields even when empty

fair enough.

lib/metadata/acl.js
```diff
@@ -14,10 +14,30 @@ const acl = {
 
     addObjectACL(bucket, objectKey, objectMD, addACLParams, params, log, cb) {
         log.trace('updating object acl in metadata');
-        // eslint-disable-next-line no-param-reassign
-        objectMD.acl = addACLParams;
-        metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params, log,
-            cb);
+        function _didNotChange(grant) {
```
jbertran: This function can be defined at class level - no need to redefine the function at every `addObjectACL` call.

jbertran (approved):

francoisferrand (approved):

PR #4460, opened 2022-04-19
bugfix: CLDSRV-177 fix crash with empty object replication

jonathan-gramain:
Fix a case of crash when a replication occurs with an empty object
over a non-empty object.

It is not clear how this happens in practice but there can be some
corner cases with race conditions between object replication and
versioning getting suspended on the target bucket at the same time,
that could lead to this situation, as the check between replication
configuration and actual replication is not atomic.


jonathan-gramain:
ping

watsaqat (approved):

francoisferrand (approved):

jonathan-gramain:
ping

PR #4480, opened 2022-04-22
Improvement/cldsrv 168 delete bucket tagging

ghost:
Monitoring not supported on 7.x branches.

### DIFF
--- a/lib/api/bucketDeleteTagging.js
+++ b/lib/api/bucketDeleteTagging.js
@@ -1,0 +1,50 @@
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const { pushMetric } = require('../utapi/utilities');
+const metadata = require('../metadata/wrapper');
+const util = require('node:util');
+
+/**
+ * Bucket Delete Tagging - Delete a bucket's Tagging
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+async function bucketDeleteTagging(authInfo, request, log, callback) {
+    const bucketName = request.bucketName;
+    let error = null;
+    log.debug('processing request', { method: 'bucketDeleteTagging', bucketName });
+
+    let bucket;
+    const metadataValidateBucketPromise = util.promisify(metadataValidateBucket);
+    let updateBucketPromise = util.promisify(metadata.updateBucket);
+    // necessary to bind metadata as updateBucket calls 'this', causing undefined otherwise
+    updateBucketPromise = updateBucketPromise.bind(metadata);
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketDeleteTagging',
+    };
+
+    try {
+        bucket = await metadataValidateBucketPromise(metadataValParams, log);
+        bucket.setTags([]);
+        // eslint-disable-next-line no-unused-expressions
+        await updateBucketPromise(bucket.getName(), bucket, log);
+        pushMetric('deleteBucketTagging', log, {
+            authInfo,
+            bucket: bucketName,
+        });
+    } catch (err) {
+        error = err;
+        log.error('error processing request', { error: err,
+            method: 'deleteBucketTagging', bucketName });
+    }
+    const corsHeaders = collectCorsHeaders(request.headers.origin,
+        request.method, bucket);
+    return callback(error, corsHeaders);
+}
+
+module.exports = bucketDeleteTagging;
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -24,6 +24,7 @@ const bucketPutACL = require('./bucketPutACL');
 const bucketPutCors = require('./bucketPutCors');
 const bucketPutVersioning = require('./bucketPutVersioning');
 const bucketPutTagging = require('./bucketPutTagging');
+const bucketDeleteTagging = require('./bucketDeleteTagging');
 const bucketPutWebsite = require('./bucketPutWebsite');
 const bucketPutReplication = require('./bucketPutReplication');
 const bucketPutLifecycle = require('./bucketPutLifecycle');
@@ -281,6 +282,7 @@ const api = {
     bucketPutPolicy,
     bucketGetPolicy,
     bucketDeletePolicy,
+    bucketDeleteTagging,
     bucketPutObjectLock,
     bucketPutNotification,
     bucketGetNotification,


ghost (approved):
LGTM

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,39 @@
...
+        (bucket, next) => {
+            bucket.setTags([]);
+            metadata.updateBucket(bucket.getName(), bucket, log, err =>
+                next(err, bucket));
+        },
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        return callback(err, corsHeaders);
+    });
```
ghost:
```suggestion
    ], (err, bucket) => {
        const corsHeaders = collectCorsHeaders(request.headers.origin,
            request.method, bucket);
        pushMetric('deleteBucketTagging', log, {
            authInfo,
            bucket: bucketName,
        });
        return callback(err, corsHeaders);
    });
```

ghost (commented):
Suggested a missing pushMetrics call

KillianG (changes requested):
I think you are missing tests for getBucketTagging no ?

ghost:
@KillianG they are in the GET PR AFAIK, as Get is needed in the tests.

KillianG:
> @KillianG they are in the GET PR AFAIK, as Get is needed in the tests.

Oops I meant get bucket tagging, edited

nicolas2bert (commented):
Why is there no test?

ghost:
@nicolas2bert The tests are added in [getBucketTagging](https://github.com/scality/cloudserver/pull/4481) as they would be incomplete otherwise

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,44 @@
...
+    return waterfall([
+        next => metadataValidateBucket(metadataValParams, log,
+            (err, bucket) => next(err, bucket)),
+        (bucket, next) => {
+            bucket.setTags([]);
+            metadata.updateBucket(bucket.getName(), bucket, log, err =>
+                next(err, bucket));
+        },
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
```
nicolas2bert: log the error in case of failure?

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,44 @@
...
+            request.method, bucket);
+        pushMetric('deleteBucketTagging', log, {
```
nicolas2bert, line 36: Should we push UTAPI metrics even if the call failed?

KillianG (approved):

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Bucket Delete Tagging - Delete a bucket's Tagging
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+function bucketDeleteTagging(authInfo, request, log, callback) {
```
nicolas2bert: Is there a plan for monitoring this call?

lib/api/bucketDeleteTagging.js
ghost: Is this ok? https://github.com/scality/cloudserver/pull/4480/commits/94700611b434aa8bb2dc5c21bf7071f44d328a89#diff-ecb34075d5365a2b604ea4bcb794846a9c3352a7bdbcd5caebbb8a2c7862466eR37

lib/api/bucketDeleteTagging.js
ghost: moved out of error scope, this ok? https://github.com/scality/cloudserver/pull/4480/commits/94700611b434aa8bb2dc5c21bf7071f44d328a89#diff-ecb34075d5365a2b604ea4bcb794846a9c3352a7bdbcd5caebbb8a2c7862466eR41

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,49 @@
...
+            bucket.setTags([]);
+            metadata.updateBucket(bucket.getName(), bucket, log, err =>
+                next(err, bucket));
+        },
+    ], (err, bucket) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        if (err) {
+            log.trace('error processing request', { error: err,
+                method: 'deleteBucketTagging' });
```
ghost: I feel like we should use a `log.error` here, as the error can be a metadata error. What do you think?

ghost: https://github.com/scality/cloudserver/pull/4480/commits/8e1820362ca258dae30885496ef79cb79282f0d0#diff-ecb34075d5365a2b604ea4bcb794846a9c3352a7bdbcd5caebbb8a2c7862466eR37

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,50 @@
...
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketDeleteTagging',
+    };
+
+    try {
+        let promisifyFunc = util.promisify(metadataValidateBucket);
+        bucket = await promisifyFunc(metadataValParams, log);
```
ghost: Maybe you already discussed that with @miniscruff but I am not a big fan of try/catch to catch Promise errors...
Why not trying to stick with golang-like logic using https://www.npmjs.com/package/await-to-js ?

ghost: Otherwise LGTM

miniscruff: Hmm, not sure the issue with try/catch but for JS it is definitely the standard and most common approach. I think we might want to make our try catches smaller, as in one around each thing that can error. But I would not be a fan of using that package, don't get me wrong, I really like go syntax but for JS it will put us on an island as it is not standard at all.

lib/api/bucketDeleteTagging.js
miniscruff: Resolving Nicolas's comments as he is on a long vacation.

package.json
```diff
@@ -42,7 +42,7 @@
   },
   "devDependencies": {
     "bluebird": "^3.3.1",
-    "eslint": "^2.4.0",
+    "eslint": "^3.6.0",
```
miniscruff: not sure why this change was added to this PR, adds a bit of fluff

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,50 @@
...
+    let bucket;
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketDeleteTagging',
+    };
+
+    try {
+        let promisifyFunc = util.promisify(metadataValidateBucket);
```
miniscruff: Same comment from the other PR

miniscruff (changes requested):

package.json
ghost: because otherwise [this line ](https://github.com/scality/cloudserver/pull/4480/files#diff-ecb34075d5365a2b604ea4bcb794846a9c3352a7bdbcd5caebbb8a2c7862466eR33) triggers the linter. Do you suggest putting an ignore comment instead?

lib/api/bucketDeleteTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4480/files#diff-ecb34075d5365a2b604ea4bcb794846a9c3352a7bdbcd5caebbb8a2c7862466eR19

KillianG (approved):

package.json
miniscruff: What is wrong with that line? I think an ignore is easier if it can't be fixed with a follow up later. Generally upgrading eslint gives a lot of changes, this one was sort of small but still a bit.

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,50 @@
...
+        bucket = await metadataValidateBucketPromise(metadataValParams, log);
+        bucket.setTags([]);
+        await updateBucketPromise(bucket.getName(), bucket, log);
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        pushMetric('deleteBucketTagging', log, {
+            authInfo,
+            bucket: bucketName,
+        });
+        return callback(null, corsHeaders);
```
miniscruff: Not really asking for change here as we do it all the time but the return is not required. We use it as an early return and in this case it won't matter. Although, if our callback throws an error we might do something we don't want to.
Generally I prefer to do
```js
callback(...);
return;
```
But its fine.

miniscruff (approved):

package.json
jonathan-gramain: I agree with @miniscruff, in this PR's scope it's better not to upgrade the linter version and just add an ignore rule if the line is legit/needed and cannot easily be changed to please the linter.

If we decide that we should upgrade the linter version, then we can do it with a separate ticket and PR.

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Bucket Delete Tagging - Delete a bucket's Tagging
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+async function bucketDeleteTagging(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketDeleteTagging' });
```
jonathan-gramain: This log may be more useful if also logging some request info, at least the bucket name and maybe also the owner ID

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,50 @@
...
+            request.method, bucket);
+        pushMetric('deleteBucketTagging', log, {
+            authInfo,
+            bucket: bucketName,
+        });
+        return callback(null, corsHeaders);
+    } catch (err) {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        log.error('error processing request', { error: err,
```
jonathan-gramain, line 42: same here with logging the bucket name

lib/api/bucketDeleteTagging.js
jonathan-gramain: I'm not sure it is a real issue here, but at least something to check: we might want to be wary of exceptions that may be raised from within the callback, in order to avoid them being caught and the callback being called again with an error. In doubt, I suggest to move the callback after the `try...catch` block, and `return` in the `catch` block.

In the callback-based code base, we usually do not throw exceptions on purpose, and if they occur anyway, we do not catch them so they end up crashing the worker. If we make use of exceptions and attempt to catch them, this kind of issue of double callback becomes a possibility.

jonathan-gramain, line 40: I'm not sure it is a real issue here, but at least something to check: we might want to be wary of exceptions that may be raised from within the callback, in order to avoid them being caught and the callback being called again with an error. In doubt, I suggest to move the callback after the `try...catch` block, and `return` in the `catch` block.

In the callback-based code base, we usually do not throw exceptions on purpose, and if they occur anyway, we do not catch them so they end up crashing the process. If we make use of exceptions and attempt to catch them, this kind of issue of double callback becomes a possibility.

miniscruff: oh I mentioned something similar in one of my comments, so at least I am on the same page sometimes. :smirk:

watsaqat (approved):

watsaqat (approved):

lib/api/bucketDeleteTagging.js
ghost: done 684e92cd5c1651787050c6a7bb8e8f440cb8ce2f

lib/api/bucketDeleteTagging.js
ghost: done 684e92cd5c1651787050c6a7bb8e8f440cb8ce2f

lib/api/bucketDeleteTagging.js
ghost: done 684e92cd5c1651787050c6a7bb8e8f440cb8ce2f

ghost:
ping

PR #4481, opened 2022-04-22
Improvement/cldsrv 169 get bucket tagging

ghost:
Monitoring not supported on 7.x branches.

### DIFF
--- a/lib/api/bucketGetTagging.js
+++ b/lib/api/bucketGetTagging.js
@@ -1,0 +1,108 @@
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const util = require('node:util');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const { checkExpectedBucketOwner } = require('./apiUtils/authorization/bucketOwner');
+const { pushMetric } = require('../utapi/utilities');
+const { errors, s3middleware } = require('arsenal');
+const escapeForXml = s3middleware.escapeForXml;
+
+//  Sample XML response:
+/*
+    <Tagging>
+        <TagSet>
+            <Tag>
+                <Key>string</Key>
+                <Value>string</Value>
+            </Tag>
+            <Tag>
+                <Key>string</Key>
+                <Value>string</Value>
+            </Tag>
+        </TagSet>
+    </Tagging>
+*/
+
+/**
+ * @typedef Tag
+ * @type {object}
+ * @property {string} Value - Value of the tag.
+ * @property {string} Key - Key of the tag.
+ */
+/**
+ * Convert Versioning Configuration object of a bucket into xml format.
+ * @param {array.<Tag>} tags - set of bucket tag
+ * @return {string} - the converted xml string of the versioning configuration
+ */
+function tagsToXml(tags) {
+    const xml = [];
+
+    xml.push('<?xml version="1.0" encoding="UTF-8" standalone="yes"?><Tagging><TagSet>');
+
+    tags.forEach(tag => {
+        xml.push('<Tag>');
+        xml.push(`<Key>${escapeForXml(tag.Key)}</Key>`);
+        xml.push(`<Value>${escapeForXml(tag.Value)}</Value>`);
+        xml.push('</Tag>');
+    });
+
+    xml.push('</TagSet></Tagging>');
+
+    return xml.join('');
+}
+
+/**
+ * bucketGetVersioning - Return Versioning Configuration for bucket
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to respond to http request
+ *  with either error code or xml response body
+ * @return {undefined}
+ */
+async function bucketGetTagging(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketGetTagging' });
+
+    const { bucketName, headers } = request;
+    const metadataValidateBucketPromise = util.promisify(metadataValidateBucket);
+    const checkExpectedBucketOwnerPromise = util.promisify(checkExpectedBucketOwner);
+
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketGetTagging',
+        request,
+    };
+
+    let bucket;
+    let xml = null;
+
+    try {
+        bucket = await metadataValidateBucketPromise(metadataValParams, log);
+        // eslint-disable-next-line no-unused-expressions
+        await checkExpectedBucketOwnerPromise(headers, bucket, log);
+        const tags = bucket.getTags();
+        if (!tags || !tags.length) {
+            log.debug('bucket TagSet does not exist', {
+                method: 'bucketGetTagging',
+            });
+            throw errors.NoSuchTagSet;
+        }
+        xml = tagsToXml(tags);
+        pushMetric('getBucketTagging', log, {
+            authInfo,
+            bucket: bucketName,
+        });
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        return callback(null, xml, corsHeaders);
+    } catch (err) {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        log.debug('error processing request',
+            { method: 'bucketGetTagging', error: err });
+
+        return callback(err, corsHeaders);
+    }
+}
+
+module.exports = bucketGetTagging;
--- a/tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
+++ b/tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
@@ -1,0 +1,84 @@
+const assert = require('assert');
+const { S3 } = require('aws-sdk');
+const async = require('async');
+const assertError = require('../../../../utilities/bucketTagging-util');
+
+const getConfig = require('../support/config');
+
+const bucket = 'policyputtaggingtestbucket';
+
+const validTagging = {
+    TagSet: [
+        {
+            Key: 'key1',
+            Value: 'value1',
+        },
+        {
+            Key: 'key2',
+            Value: 'value2',
+        },
+    ],
+};
+
+describe('aws-sdk test delete bucket tagging', () => {
+    let s3;
+
+    before(() => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+    });
+
+    beforeEach(done => s3.createBucket({ Bucket: bucket }, done));
+
+    afterEach(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+    it('should delete tag', done => {
+        async.series([
+            next => s3.putBucketTagging({
+                AccountId: s3.AccountId,
+                Tagging: validTagging, Bucket: bucket,
+            }, (err, res) => next(err, res)),
+            next => s3.getBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, (err, res) => {
+                assert.deepStrictEqual(res, validTagging);
+                next(err, res);
+            }),
+            next => s3.deleteBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, (err, res) => next(err, res)),
+            next => s3.getBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, next),
+        ], (err) => {
+            assertError(err, 'NoSuchTagSet');
+            done();
+        });
+    });
+
+    it('should make no change when deleting tags on bucket with no tags', done => {
+        async.series([
+            next => s3.getBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, (err) => {
+                assertError(err, 'NoSuchTagSet');
+                next();
+            }),
+            next => s3.deleteBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, (err, res) => next(err, res)),
+            next => s3.getBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, (err) => {
+                assertError(err, 'NoSuchTagSet');
+                next();
+            }),
+        ], done);
+    });
+});
--- a/tests/functional/aws-node-sdk/test/bucket/getBucketTagging.js
+++ b/tests/functional/aws-node-sdk/test/bucket/getBucketTagging.js
@@ -1,0 +1,48 @@
+const assertError = require('../../../../utilities/bucketTagging-util');
+const { S3 } = require('aws-sdk');
+const async = require('async');
+
+const getConfig = require('../support/config');
+
+const bucket = 'gettaggingtestbucket';
+
+describe('aws-sdk test get bucket tagging', () => {
+    let s3;
+
+    before(() => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+    });
+
+    beforeEach(done => s3.createBucket({ Bucket: bucket }, done));
+
+    afterEach(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+    it('should return accessDenied if expected bucket owner does not match', done => {
+        async.waterfall([
+            next => s3.getBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+                ExpectedBucketOwner: '944690102203',
+            },
+                (err, res) => {
+                    next(err, res);
+                }),
+        ], (err) => {
+            assertError(err, 'AccessDenied');
+            done();
+        });
+    });
+
+    it('should not return accessDenied if expected bucket owner matches', done => {
+        async.series([
+            next => s3.getBucketTagging({ AccountId: s3.AccountId, Bucket: bucket, ExpectedBucketOwner: s3.AccountId },
+                (err, res) => {
+                    next(err, res);
+                }),
+        ], (err) => {
+            assertError(err, 'NoSuchTagSet');
+            done();
+        });
+    });
+});
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -25,6 +25,7 @@ const bucketPutCors = require('./bucketPutCors');
 const bucketPutVersioning = require('./bucketPutVersioning');
 const bucketPutTagging = require('./bucketPutTagging');
 const bucketDeleteTagging = require('./bucketDeleteTagging');
+const bucketGetTagging = require('./bucketGetTagging');
 const bucketPutWebsite = require('./bucketPutWebsite');
 const bucketPutReplication = require('./bucketPutReplication');
 const bucketPutLifecycle = require('./bucketPutLifecycle');
@@ -272,6 +273,7 @@ const api = {
     bucketPutCors,
     bucketPutVersioning,
     bucketPutTagging,
+    bucketGetTagging,
     bucketPutWebsite,
     bucketPutReplication,
     bucketGetReplication,
--- a/tests/functional/aws-node-sdk/test/bucket/putBucketTagging.js
+++ b/tests/functional/aws-node-sdk/test/bucket/putBucketTagging.js
@@ -1,7 +1,7 @@
 const assert = require('assert');
-const { errors } = require('arsenal');
 const { S3 } = require('aws-sdk');
 const async = require('async');
+const assertError = require('../../../../utilities/bucketTagging-util');
 
 const getConfig = require('../support/config');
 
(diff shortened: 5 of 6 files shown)


tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
```diff
@@ -0,0 +1,94 @@
...
+        assert.strictEqual(err.code, expectedErr, 'incorrect error response ' +
+            `code: should be '${expectedErr}' but got '${err.code}'`);
+        assert.strictEqual(err.statusCode, errors[expectedErr].code,
+            'incorrect error status code: should be  ' +
+            `${errors[expectedErr].code}, but got '${err.statusCode}'`);
+    }
+    cb();
+}
+
+describe.only('aws-sdk test delete bucket tagging', () => {
```
ghost:
```suggestion
describe('aws-sdk test delete bucket tagging', () => {
```

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
```diff
@@ -0,0 +1,94 @@
...
+            next => s3.getBucketTagging({
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, (err, res) => {
+                next(err, res);
+            }),
+        ], (err) => {
+            assertError(err, 'NoSuchTagSet', done);
+        });
+    });
```
ghost: I don't understand what is tested here, but I think we should check at each step that the operation returned the right results, i.e. that the first GET returns something, and the second returns an empty array.

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
```diff
@@ -0,0 +1,94 @@
...
+                Bucket: bucket,
+            }, (err, res) => {
+                next(err, res);
+            }),
+        ], (err) => {
+            assert.ifError(err);
+            done(err);
+        });
+    });
+});
```
ghost: Same here: "no change" but we are not testing the returned values. Can you do a GET, store the result, then DELETE, and then GET and compare with the first result?

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,90 @@
...
+        const tags = bucket.getTags();
+        if (!tags || !tags.length) {
+            log.debug('bucket TagSet does not exist', {
+                method: 'bucketGetTagging',
+            });
+            return callback(errors.NoSuchTagSet, null,
+                corsHeaders);
+        }
+        const xml = convertToXml(tags);
+        pushMetric('bucketGetTagging', log, {
```
ghost:
```suggestion
        pushMetric('getBucketTagging', log, {
```

ghost (commented):
Some improvements I think we can do in the PR

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
ghost: I probably should've been clearer in the PRs notes: this PR's tests are unfinished until I can get this change through https://github.com/scality/Arsenal/pull/1814

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
ghost: In theory this should come out to the same, as we know at the start the bucket has no tags, so this test is mostly to see no error arises from trying to delete tags on a bucket that has none. If I try a getBucketTagging here, it'll send me into the error callback for 'NoSuchTags' each time, so I'd have to rework this differently. Maybe I should redo this file in async/await to be able to do all these tests cleanly?

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,90 @@
...
+const { errors } = require('arsenal');
+
+//  Sample XML response:
+/*
+    <Tagging>
+        <TagSet>
+            <Tag>
+                <Key>string</Key>
+                <Value>string</Value>
+            </Tag>
```
KillianG, line 16: Maybe it can be worth adding more than one tag in the example tagset ? What do you think ?

KillianG:
I think you are missing tests for get bucket tagging no ?

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
KillianG: You can maybe "shadow" the error by checking that the error is NoSuchtagSet and if so, return a next(null, null) or simply next(), otherwise continue returning next(err, res) to catch a possible error

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,90 @@
...
+        </TagSet>
+    </Tagging>
+*/
+
+/**
+ * Convert Versioning Configuration object of a bucket into xml format.
+ * @param {object} tags - set of bucket tags
```
miniscruff: this is listed as an object but is used with foreach as an array

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,44 @@
...
+function bucketDeleteTagging(authInfo, request, log, callback) {
+    log.debug('processing request', { method: 'bucketDeleteTagging' });
+
+    const bucketName = request.bucketName;
+    const metadataValParams = {
+        authInfo,
+        bucketName,
+        requestType: 'bucketDeleteTagging',
+    };
+    return waterfall([
```
miniscruff: we can use async/await instead of waterfall for new code

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
ghost: I honestly don't feel on this particular case this is necessary... the default is that the bucket has no tags, and we know that. This test is to check that the delete doesn't give any error when run against a bucket with no tags.
I'll try returning an empty next() in any case

lib/api/bucketGetTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/commits/bc8c6f527ec83cfcc7cb7b2bcf2d83ff9b35609b#diff-889d6969db5c5fda428f1b3f508e237cb4bc1419949c78bc0a146bcca7d9209bR14

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/commits/bc8c6f527ec83cfcc7cb7b2bcf2d83ff9b35609b#diff-76b4a70b2a7470e6cde6671151107006543e8d06514f148d131f7f86ea0007b3R80

ghost:
> I think you are missing tests for get bucket tagging no ?

@KillianG I considered that get was amply tested in the put and delete suites but we may need more tests for 'expected bucket owner' specifically for get. I'll add them.

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,94 @@
...
+    const xml = [];
+
+    xml.push('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>',
+             '<Tagging>',
+             '<TagSet>'
+    );
+
+    tags.forEach(tag => {
+        xml.push('<Tag>');
+        xml.push(`<Key>${tag.Key}</Key>`);
```
KillianG: Maybe you can use escapeForXml to avoid xml injection in tag key and value
```suggestion
        xml.push(`<Key>${escapeForXml(tag.Key)}</Key>`);
```

lib/api/bucketDeleteTagging.js
ghost: After looking at the issue and discussion with @anurag4DSB , this would require promisifying some elements and will take some time we aren't ready to spend right now given it would be better done in a general code cleanup; will create a ticket.

lib/api/bucketGetTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/commits/8c19871aabb66c070212762aa0f1bc9df9d5de7e

lib/api/bucketGetTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/commits/f045bcd4bb3100b1f0412d671d72871a3a920cf1

ghost:
> @KillianG I considered that get was amply tested in the put and delete suites but we may need more tests for 'expected bucket owner' specifically for get. I'll add them.

done https://github.com/scality/cloudserver/pull/4481/commits/f045bcd4bb3100b1f0412d671d72871a3a920cf1

KillianG (approved):

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,106 @@
...
+            const corsHeaders = collectCorsHeaders(request.headers.origin,
+                request.method, bucket);
+            return callback(null, xml, corsHeaders);
+        },
+    ], (err) => {
+        const corsHeaders = collectCorsHeaders(request.headers.origin,
+            request.method, bucket);
+        log.debug('error processing request',
+            { method: 'bucketGetTagging', error: err });
+        return callback(err, corsHeaders);
```
ghost: Is it expected to always log an error?
```suggestion
        const corsHeaders = collectCorsHeaders(request.headers.origin,
            request.method, bucket);
        if (err) {
            log.debug('error processing request',
                { method: 'bucketGetTagging', error: err });
        }
        return callback(err, corsHeaders);
```

lib/api/bucketDeleteTagging.js
miniscruff: You mean this guy? https://nodejs.org/api/util.html#utilpromisifyoriginal 
Promisifying can be done very easily and with no effort using the above util method.

lib/api/bucketGetTagging.js
miniscruff: I will let it be for now but "array of what?"

lib/api/bucketGetTagging.js
ghost: Oops I misread my work on this one, fixed 9d9afd022f7963994ba2a1572bc8afab39d7941c

ghost (approved):
LGTM

lib/api/bucketDeleteTagging.js
ghost: Used async await, let me know what you think https://github.com/scality/cloudserver/pull/4480/commits/13907cb2dd25f0511c3929f5987e111a68487c02

lib/api/bucketGetTagging.js
ghost: Is this better?
https://github.com/scality/cloudserver/pull/4481/files#diff-889d6969db5c5fda428f1b3f508e237cb4bc1419949c78bc0a146bcca7d9209bR27

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,105 @@
...
+        bucketName,
+        requestType: 'bucketGetTagging',
+        request,
+    };
+
+    let bucket;
+    let xml = null;
+
+    try {
+        let promisifyFunc = util.promisify(metadataValidateBucket);
```
miniscruff: I would create a separate value for each promisified(?) function, maybe even at the top level if you want.
Such as `const metadataValidateBucketPromise = util.promisify(...);`

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,105 @@
...
+
+    tags.forEach(tag => {
+        xml.push('<Tag>');
+        xml.push(`<Key>${escapeForXml(tag.Key)}</Key>`);
+        xml.push(`<Value>${escapeForXml(tag.Value)}</Value>`);
+        xml.push('</Tag>');
+    });
+
+    xml.push('</TagSet>',
+        '</Tagging>');
```
miniscruff: Super nit picky but if you just have a constant value you don't need two entries it can just be one. Separating entries is so we can handle conditional or formatted options. Also counts for the first one if you want.
```suggestion
    xml.push('</TagSet></Tagging>');
```

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,105 @@
...
+            </Tag>
+        </TagSet>
+    </Tagging>
+*/
+
+/**
+ * Convert Versioning Configuration object of a bucket into xml format.
+ * @param {array} tags - set of bucket tag keys and values
+ * @return {string} - the converted xml string of the versioning configuration
+ */
```
miniscruff: New comment, I would either extend array to include the specific data structure or create a typedef of what a tag is.
https://jsdoc.app/tags-type.html

```suggestion
/**
 * @typedef Tag
 * @type {object}
 * @property {string} Value - Value of our tag.
 * @property {string} Key - Key of the tag.
 */
/**
 * Convert Versioning Configuration object of a bucket into xml format.
 * @param {array.<Tag>} tags - set of bucket tags
 * @return {string} - the converted xml string of the versioning configuration
 */
```
Ideally if you added more details as this was mostly an example, but now we get full typing.

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,105 @@
...
+function convertToXml(tags) {
```
miniscruff: Now that I think about it, this function name can be better, something like `tagsToXml`, as the name itself doesn't reference what its converting, so it could be anything.

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,105 @@
...
+        let promisifyFunc = util.promisify(metadataValidateBucket);
+        bucket = await promisifyFunc(metadataValParams, log);
+        promisifyFunc = util.promisify(checkExpectedBucketOwner);
+        await promisifyFunc(headers, bucket, log);
+        const tags = bucket.getTags();
+        if (!tags || !tags.length) {
+            log.debug('bucket TagSet does not exist', {
+                method: 'bucketGetTagging',
+            });
+            throw (errors.NoSuchTagSet);
```
miniscruff:
```suggestion
            throw errors.NoSuchTagSet;
```

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
```diff
@@ -0,0 +1,101 @@
...
+    if (expectedErr === null) {
+        assert.strictEqual(err, null, `expected no error but got '${err}'`);
+    } else {
+        assert.strictEqual(err.code, expectedErr, 'incorrect error response ' +
+            `code: should be '${expectedErr}' but got '${err.code}'`);
+        assert.strictEqual(err.statusCode, errors[expectedErr].code,
+            'incorrect error status code: should be  ' +
+            `${errors[expectedErr].code}, but got '${err.statusCode}'`);
+    }
+    cb();
```
miniscruff: This callback seems completely separate from the rest of the function and as nothing is async doesn't seem required.

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
```diff
@@ -0,0 +1,101 @@
...
+        s3 = new S3(config);
+        return done();
+    });
+
+    beforeEach(done => s3.createBucket({ Bucket: bucket }, done));
+
+    afterEach(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+    it('should delete tag', done => {
+        async.series([
```
miniscruff, line 36: we might need to update to v3 of the aws sdk but we should at some point and use async await for this sort of stuff.

tests/functional/aws-node-sdk/test/bucket/getBucketTagging.js
```diff
@@ -0,0 +1,62 @@
...
+const { errors } = require('arsenal');
+const { S3 } = require('aws-sdk');
+const async = require('async');
+
+const getConfig = require('../support/config');
+
+const bucket = 'gettaggingtestbucket';
+
+// Check for the expected error response code and status code.
+function assertError(err, expectedErr, cb) {
```
miniscruff: wait is this the same as the other one? I also feel like I have seen this around before.

tests/functional/aws-node-sdk/test/bucket/getBucketTagging.js
```diff
@@ -0,0 +1,62 @@
...
+    cb();
+}
+
+describe('aws-sdk test get bucket tagging', () => {
+    let s3;
+
+    before(done => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+        return done();
```
miniscruff: Doesn't seem like we need done here

miniscruff (changes requested):

lib/api/bucketGetTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/files#diff-889d6969db5c5fda428f1b3f508e237cb4bc1419949c78bc0a146bcca7d9209bR66

lib/api/bucketGetTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/files#diff-889d6969db5c5fda428f1b3f508e237cb4bc1419949c78bc0a146bcca7d9209bR48

lib/api/bucketGetTagging.js
ghost: done? https://github.com/scality/cloudserver/pull/4481/files#diff-889d6969db5c5fda428f1b3f508e237cb4bc1419949c78bc0a146bcca7d9209bR41

lib/api/bucketGetTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/files#diff-889d6969db5c5fda428f1b3f508e237cb4bc1419949c78bc0a146bcca7d9209bR36

lib/api/bucketGetTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/files#diff-889d6969db5c5fda428f1b3f508e237cb4bc1419949c78bc0a146bcca7d9209bR87

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
ghost: done [tests/utilities/bucketTagging-util.js](https://github.com/scality/cloudserver/pull/4481/files#diff-5f6289713c22c427fbe9c6a9db3190cf73c0637ebda0a9a92f8c660ce9ac6ac1R1)

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
ghost: Going to suppose this doesn't need immediate changes

tests/functional/aws-node-sdk/test/bucket/getBucketTagging.js
ghost: Yes, done https://github.com/scality/cloudserver/pull/4481/files#diff-5f6289713c22c427fbe9c6a9db3190cf73c0637ebda0a9a92f8c660ce9ac6ac1R1

tests/functional/aws-node-sdk/test/bucket/getBucketTagging.js
ghost: done https://github.com/scality/cloudserver/pull/4481/files#diff-5e6446c3e90586b538e1b00f9605b791ca303183acc5533411925d0e8be625d7R12

KillianG (approved):

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
miniscruff: Yea I was just mentioning it cause we should use async await but in v2 its a bit harder.

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
miniscruff: I think you linked the this in the wrong comment

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
```diff
@@ -0,0 +1,88 @@
...
+    ],
+};
+
+describe('aws-sdk test delete bucket tagging', () => {
+    let s3;
+
+    before(done => {
+        const config = getConfig('default', { signatureVersion: 'v4' });
+        s3 = new S3(config);
+        return done();
```
miniscruff: This done is not needed either

tests/functional/aws-node-sdk/test/bucket/deleteBucketTagging.js
```diff
@@ -0,0 +1,88 @@
...
+                AccountId: s3.AccountId,
+                Bucket: bucket,
+            }, (err) => {
+                assertError(err, 'NoSuchTagSet');
+                next();
+            }),
+        ], (err) => {
+            assert.ifError(err);
+            done(err);
+        });
```
miniscruff: passing err to done does the same thing as the ifError, can just do:
```suggestion
        ], done);
```

miniscruff (approved):
Two minor things but looks good from me otherwise.

PR #4466, opened 2022-04-20
CLDSRV-174 bump arsenal to error breaking change

jbertran:
### Motivation and context

Typescript migration introduced a breaking change in arsenal errors. This PR updates Arsenal and all Arsenal error uses to conform with the new API.

## Checklist

- [x] bump Arsenal
- [x] unit tests
- [x] functional tests

PR #4493, opened 2022-05-03
Improvement/cldsrv 191 bump arsenal post breaking changes dev8x

XinLi0207:
use an arsenal commit hashtag for testing, will replace it with 8.1.45 once it's released.

this PR is mainly doing two things:
- update new arsenal error type: 
  - use `err.is.Error` instead of  `=== `error equal, 
  - use `assert.strictEqual(err.is.Error, true) ` instead of `deepStrictEqual(err, errors.Error)`
- after[ this change](https://github.com/scality/Arsenal/commit/b3ce76d7d8d2f62e7ebf8dbee46221923bac2c76) in arsenal, we can no longer use a random string to represent a non existing versionId for testing, it won't decode it successfully so will return `InvalidArgument` error instead of `NoSuchVersion` error that we expect. So even for non existing versionId, we have to give it a standard formated versionId, so I replaced several places where we use `000000000` for non existing versionId with a meaningful but also non existing versionId. https://github.com/scality/cloudserver/pull/4493/commits/b1d824c9aa573dfdd0b9d7b1509e7c18fbbd05c1

green build in zenko https://eve.devsca.com/github/scality/zenko/#/builders/4/builds/21334

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -24,7 +24,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
 
     // Get new format usersBucket to see if it exists
     return metadata.getBucket(usersBucket, log, (err, usersBucketAttrs) => {
-        if (err && !err.NoSuchBucket && !err.BucketAlreadyExists) {
+        if (err && !err.is.NoSuchBucket && !err.is.BucketAlreadyExists) {
             return cb(err);
         }
         const splitter = usersBucketAttrs ?
@@ -41,7 +41,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
             usersBucket : oldUsersBucket;
         return metadata.putObjectMD(usersBucketBeingCalled, key,
             omVal, {}, log, err => {
-                if (err && err.NoSuchBucket) {
+                if (err?.is.NoSuchBucket) {
                     // There must be no usersBucket so createBucket
                     // one using the new format
                     log.trace('users bucket does not exist, ' +
@@ -62,8 +62,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
                             // error with respect
                             // to the usersBucket.
                             if (err &&
-                                err !==
-                                    errors.BucketAlreadyExists) {
+                                !err.is.BucketAlreadyExists) {
                                 log.error('error from metadata', {
                                     error: err,
                                 });
@@ -224,7 +223,7 @@ function createBucket(authInfo, bucketName, headers,
         },
         getAnyExistingBucketInfo: function getAnyExistingBucketInfo(callback) {
             metadata.getBucket(bucketName, log, (err, data) => {
-                if (err && err.NoSuchBucket) {
+                if (err?.is.NoSuchBucket) {
                     return callback(null, 'NoBucketYet');
                 }
                 if (err) {
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -214,12 +214,12 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
             (versionId, callback) => metadataGetObject(bucketName, entry.key,
                 versionId, log, (err, objMD) => {
                     // if general error from metadata return error
-                    if (err && !err.NoSuchKey) {
+                    if (err && !err.is.NoSuchKey) {
                         monitoring.promMetrics('DELETE', bucketName, err.code,
                             'multiObjectDelete');
                         return callback(err);
                     }
-                    if (err && err.NoSuchKey) {
+                    if (err?.is.NoSuchKey) {
                         const verCfg = bucket.getVersioningConfiguration();
                         // To adhere to AWS behavior, create a delete marker
                         // if trying to delete an object that does not exist
@@ -396,7 +396,8 @@ function multiObjectDelete(authInfo, request, log, callback) {
             return vault.checkPolicies(requestContextParams, authInfo.getArn(),
                 log, (err, authorizationResults) => {
                     // there were no policies so received a blanket AccessDenied
-                    if (err && err.AccessDenied) {
+                    log.info('please see here: ', { err, authorizationResults });
+                    if (err?.is.AccessDenied) {
                         objects.forEach(entry => {
                             errorResults.push({
                                 entry,
--- a/tests/functional/aws-node-sdk/test/bucket/deleteBucketReplication.js
+++ b/tests/functional/aws-node-sdk/test/bucket/deleteBucketReplication.js
@@ -81,7 +81,7 @@ describe('aws-node-sdk test deleteBucketReplication', () => {
             }),
             next => deleteReplicationAndCheckResponse(bucket, next),
             next => s3.getBucketReplication({ Bucket: bucket }, err => {
-                assert(errors.ReplicationConfigurationNotFoundError[err.code]);
+                assert(errors.ReplicationConfigurationNotFoundError.is[err.code]);
                 return next();
             }),
         ], done));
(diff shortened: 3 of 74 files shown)


miniscruff:
Would it be possible to target 7.10 with this one?

XinLi0207:
> Would it be possible to target 7.10 with this one?

Hi Ronnie, @miniscruff 
there is already one pr for dev/7.10 https://github.com/scality/cloudserver/pull/4466
and also considering dev/7.x and dev/8.x use different arsenal versions, so I didn't target 7.x and 8.x in one pr. how do you think?

tests/functional/aws-node-sdk/test/bucket/deleteBucketReplication.js
```diff
@@ -81,7 +81,7 @@ describe('aws-node-sdk test deleteBucketReplication', () => {
             }),
             next => deleteReplicationAndCheckResponse(bucket, next),
             next => s3.getBucketReplication({ Bucket: bucket }, err => {
-                assert(errors.ReplicationConfigurationNotFoundError[err.code]);
+                assert(errors.ReplicationConfigurationNotFoundError.is[err.code]);
```
ghost, line 84:
```suggestion
                assert(errors.ReplicationConfigurationNotFoundError.is[err.type]);
```

`code` is the http status code

tests/functional/aws-node-sdk/test/bucket/getBucketReplication.js
```diff
@@ -45,7 +45,7 @@ describe('aws-node-sdk test getBucketReplication', () => {
     it("should return 'ReplicationConfigurationNotFoundError' if bucket does " +
     'not have a replication configuration', done =>
         s3.getBucketReplication({ Bucket: bucket }, err => {
-            assert(errors.ReplicationConfigurationNotFoundError[err.code]);
+            assert(errors.ReplicationConfigurationNotFoundError.is[err.code]);
```
ghost, line 48:
```suggestion
            assert(errors.ReplicationConfigurationNotFoundError.is[err.type]);
```

tests/unit/api/objectPutTagging.js
```diff
@@ -38,7 +38,7 @@ const testPutObjectRequest = new DummyRequest({
 function _checkError(err, code, errorName) {
     assert(err, 'Expected error but found none');
     assert.strictEqual(err.code, code);
```
ghost, line 40: should be type as well, right?
```suggestion
    assert.strictEqual(err.type, code);
```
image


You also need to update all `_checkError` functions in multiple places (you can use the `assert.strictEqual(err.code, code);` filter):
image

tests/functional/aws-node-sdk/test/object/getObjectLegalHold.js
```diff
@@ -85,11 +89,11 @@ describeSkipIfCeph('GET object legal hold', () => {
             });
         });
 
-        it('should return NoSuchVersion error if version does not exist', done => {
+        it('should return NoSuchVersion error if version does not exist1', done => {
```
ghost:
```suggestion
        it('should return NoSuchVersion error if version does not exist', done => {
```

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -41,7 +41,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
             usersBucket : oldUsersBucket;
         return metadata.putObjectMD(usersBucketBeingCalled, key,
             omVal, {}, log, err => {
-                if (err && err.NoSuchBucket) {
+                if (err && err.is.NoSuchBucket) {
```
ghost: We could do something like that everywhere we have the `err && err.is` pattern:

```suggestion
                if (err?.is.NoSuchBucket) {
```

ghost (approved):
LGTM, just two little improvements we can do.

tests/functional/aws-node-sdk/test/bucket/deleteBucketReplication.js
jbertran: Alternatively, this is just `err.is.ReplicationConfigurationNotFoundError`, unless this error comes from another component and/or wasn't rebuilt into an ArsenalError.

tests/multipleBackend/objectPutCopyPart.js
```diff
@@ -300,7 +299,7 @@ function testSuite() {
     itSkipCeph('should return error 403 AccessDenied copying part to a ' +
     'different AWS location without object READ access',
     done => {
-        const errorPutCopyPart = { code: 'AccessDenied', statusCode: 403 };
+        const errorPutCopyPart = errors.AccessDenied;
```
jbertran: We can skip the `const` declaration here and just shove `errors.AccessDenied` in the function call

jbertran (commented):
Nice work!

You can kill `ec68150f33babaad08395fa13cfd0d133d5c356d` and squash `d0bf45eccc09591634e3e2067e94564996ef0c36` into `3b2725f209e572231f1aadb0c3c588063fa4b269` :slightly_smiling_face: 

We should get some green Zenko smoke tests before merging this, to be extra safe.

lib/api/apiUtils/bucket/bucketCreation.js
XinLi0207: ah, linter doesn't allow TT

francoisferrand:
should this PR not target 8.2 ?
(the ticket is _migrate latest arsenal to cloudserver dev/8.x_)

XinLi0207:
> should this PR not target 8.2 ? (the ticket is _migrate latest arsenal to cloudserver dev/8.x_)

Hi @francoisferrand , we are going to open dev/8.5 branch very soon, we want to maintain only the latest two branches, so I only target dev/8.4

francoisferrand:
Maintaining 2 versions means 8.3 and 8.4 in addition to the Dev branch (8.5)

But my point is that the fix will land in 8.2, due to waterflow : so best to avoid extra/redundant work, and just target 8.2

Changes from your PR should be minimal I guess, and there is no extra work with federation/integration in this case.

XinLi0207:
> Maintaining 2 versions means 8.3 and 8.4 in addition to the Dev branch (8.5)
> 
> But my point is that the fix will land in 8.2, due to waterflow : so best to avoid extra/redundant work, and just target 8.2
> 
> Changes from your PR should be minimal I guess, and there is no extra work with federation/integration in this case.

yeah, I understand. But I am not sure I can do this, rebasing to 8.2 will need some extra time, I am not sure I can still lose time for cold storage, I will check with PO quickly.

XinLi0207:
> Maintaining 2 versions means 8.3 and 8.4 in addition to the Dev branch (8.5)
> 
> But my point is that the fix will land in 8.2, due to waterflow : so best to avoid extra/redundant work, and just target 8.2
> 
> Changes from your PR should be minimal I guess, and there is no extra work with federation/integration in this case.

hi @francoisferrand, sorry, I may not be able to do it this time, targeting 8.2 and 8.3 will bring a lot of extra work like rebasing and especially testing end2end in zenko, we don't want to add more risks to cold storage. But I will help to review the PR targeting 7.10 and also its integration PRs, hope can accelerate the migrating speed :)

KillianG (approved):

lib/api/apiUtils/bucket/bucketCreation.js
XinLi0207: fixed linter

lib/api/multiObjectDelete.js
```diff
@@ -396,7 +396,8 @@ function multiObjectDelete(authInfo, request, log, callback) {
             return vault.checkPolicies(requestContextParams, authInfo.getArn(),
                 log, (err, authorizationResults) => {
                     // there were no policies so received a blanket AccessDenied
-                    if (err.is.AccessDenied) {
+                    log.info('please see here: ', { err, authorizationResults });
```
jbertran, line 399: Stray log?

jbertran (approved):

PR #4502, opened 2022-05-09
CLDSRV-192 Introduce s3:PutObjectVersion permission

nicolas2bert:


naren-rajendran (approved):

lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -118,6 +118,21 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
               generateRequestContext('objectPutACL');
             requestContexts.push(putAclRequestContext);
         }
+        // if put object with version
+        if (request.headers['x-scal-s3-version-id']) {
```
rachedbenmustapha: This evaluates to `false` if the given version id is empty, but the request would still have some "special" work to do, is it intended?

nicolas2bert: I was planning on using "null" to specify the "non-version object" version id.

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #4504, opened 2022-05-10
Feature/cldsrv 187 update head object to support cold storage

ghost:
This PR updates HEAD object:

* Add a new output field StorageClass to reflect storage location if the object is in cold storage/tape backend.
* Add a new output filed Restore with ongoing-request="true" if the object is currently being restored.

The API will also rely on the new 'archive' metadata field of the object to dynamically compute the right state and avoid race conditions if a HEAD request is sent right after a `RestoreObject` API completes, but before the backbeat job detects it.


lib/api/apiUtils/coldStorage.js
```diff
@@ -0,0 +1,33 @@
...
+    let value;
+    if (objMD.archive &&
+        objMD.archive.restoreRequestedAt &&
+        !objMD.archive.restoreCompletedAt) {
+        // Avoid race condition by relying on the `archive` MD of the object
+        // and return the right header after a RESTORE request.
+        // eslint-disable-next-line
+        value = `ongoing-request="true"`;
+    } else if (objMD['x-amz-restore']) {
+        const utcDateTime = new Date(objMD['x-amz-restore']['expiry-date']).toUTCString();
```
miniscruff: Move this inside the next if

lib/api/apiUtils/coldStorage.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+/**
+ * Get response header "x-amz-restore"
+ * Be called by objectHead.js
+ * @param {object} objMD - object's metadata
+ * @returns {string} x-amz-restore
+ */
+function getAmzRestoreResHeader(objMD) {
+    let value;
```
miniscruff: This var isn't required and I would just return instead of setting it. Use `if` instead of `else if` as well

lib/api/objectHead.js
```diff
@@ -90,6 +92,10 @@ function objectHead(authInfo, request, log, callback) {
             if (headerValResult.error) {
                 return callback(headerValResult.error, corsHeaders);
             }
+            if (locationConstraints[bucket.getLocationConstraint()]?.cold === true) {
```
miniscruff:
```suggestion
            if (locationConstraints[bucket.getLocationConstraint()]?.isCold) {
```
I think the value is going to be `isCold` @naren-rajendran 
Also no need for === true

miniscruff (changes requested):

lib/api/apiUtils/coldStorage.js
ghost: Done: https://github.com/scality/cloudserver/pull/4504/commits/a30ae19648cf97efb17ed42ef27c84bbf91b3fff#diff-5c1a41d999742cdc5b48a81be655c072599261c1755387ef030e91f2f54105b6R23

lib/api/apiUtils/coldStorage.js
ghost: I now return the value, or undefined: https://github.com/scality/cloudserver/pull/4504/commits/a30ae19648cf97efb17ed42ef27c84bbf91b3fff#diff-5c1a41d999742cdc5b48a81be655c072599261c1755387ef030e91f2f54105b6R19

lib/api/objectHead.js
ghost: Done and updated in the test conf files: https://github.com/scality/cloudserver/pull/4504/commits/a30ae19648cf97efb17ed42ef27c84bbf91b3fff#diff-bf4e2086fcee49bbecaf2f0643e61782339817ae23eb92a6fe54de17c2654be6R95

ghost: I also updated the same pattern in the GetObject PR: https://github.com/scality/cloudserver/pull/4506/commits/bbca89a890b79380cf128b44173164a2cc011c09

lib/api/apiUtils/coldStorage.js
```diff
@@ -0,0 +1,33 @@
+/*
```
XinLi0207: I realized that our `coldStorage.js` files are not under the same folder, mine is under `lib/api/apiUtils/object/coldStorage.js`, which one do you think makes more sense?

ghost: I chose to use the same location as yours (and Yutaka) : https://github.com/scality/cloudserver/pull/4504/commits/acef14d85ebf63b394b7a0bfb1c37f83428e9b50
I also rebased the other PR.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,33 @@
+/*
+ * Code based on Yutaka Oishi (Fujifilm) contributions
+ * Date: 11 Sep 2020
+ */
+
+/**
+ * Get response header "x-amz-restore"
+ * Be called by objectHead.js
+ * @param {object} objMD - object's metadata
+ * @returns {string} x-amz-restore
```
miniscruff: string or undefined

miniscruff (changes requested):
Marking as needs work mostly cause its using a custom arsenal branch

ghost (approved):
lgtm

lib/api/apiUtils/object/coldStorage.js
ghost: Thanks, I will make sure to udpate this when I will use a proper Arsenal version. I will ping you at that time

ghost: Done: https://github.com/scality/cloudserver/pull/4504/commits/790dd5a55224adb4489b89e5a22211bb7608e931

XinLi0207 (approved):

miniscruff (approved):

PR #4506, opened 2022-05-10
Feature/cldsrv 186 get object to support restore object

ghost:
GetObject was updated to:

* **(done by HeadObject)** Add a new output field StorageClass to reflect storage location if the restored object is in cold storage/tape backend.
* **(done by HeadObject)** Add a new output filed Restore with `ongoing-request="false", expiry-date: ...` if the object is restored.
* Check object status, return `InvalidObjectStateError` if object is archived or `RestoreAlreadyInProgress` if the object is currently being restored.

The object status is computed from the `archive` field of the object metadata.

### DIFF
--- a/tests/unit/api/objectGet.js
+++ b/tests/unit/api/objectGet.js
@@ -404,7 +404,7 @@ describe('objectGet API', () => {
         });
     });
 
-    it('should return RestoreAlreadyInProgress if trying to GET the object in cold storage', done => {
+    it('should return InvalidObjectState if trying to GET the object in cold storage being restored', done => {
         const testGetRequest = {
             bucketName,
             namespace,
@@ -415,7 +415,7 @@ describe('objectGet API', () => {
         mdColdHelper.putBucketMock(bucketName, null, () => {
             mdColdHelper.putObjectMock(bucketName, objectName, mdColdHelper.getArchiveOngoingRequestMD(), () => {
                 objectGet(authInfo, testGetRequest, false, log, err => {
-                    assert.strictEqual(err.is.RestoreAlreadyInProgress, true);
+                    assert.strictEqual(err.is.InvalidObjectState, true);
                     done();
                 });
             });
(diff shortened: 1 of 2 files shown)


miniscruff (approved):

tests/unit/api/objectGet.js
```diff
@@ -384,4 +385,101 @@ describe('objectGet API', () => {
...
+            mdColdHelper.putObjectMock(bucketName, objectName, mdColdHelper.getArchiveArchivedMD(), () => {
+                objectGet(authInfo, testGetRequest, false, log, err => {
+                    assert.strictEqual(err.is.InvalidObjectState, true);
+                    done();
+                });
+            });
+        });
+    });
+
+    it('should return RestoreAlreadyInProgress if trying to GET the object in cold storage', done => {
```
XinLi0207: in aws, if the object is currently being restore, we still got `InvalidObjectState` error instead of `RestoreAlreadyInProgress` for get-object api. See below, for the same object that is currently being restored:
```
➜  aws s3api restore-object --bucket xinli-bucket-test1 --key file2 --restore-request Days=12 --profile aws

An error occurred (RestoreAlreadyInProgress) when calling the RestoreObject operation: Object restore is already in progress

➜  ~ aws s3api get-object --bucket xinli-bucket-test1 --key file2 --profile aws outfile

An error occurred (InvalidObjectState) when calling the GetObject operation: The operation is not valid for the object's storage class
```

ghost: I think it is fine to comply with AWS on this, even if it is different from what we said in the design. I am making the according changes. Thanks!
See: https://github.com/scality/cloudserver/pull/4506/commits/3de2f0ca90d980a0221b9722c13e3ecfe1e33526

@RachedBenMustapha , does it make sense to you?

XinLi0207 (approved):

tests/unit/api/objectGet.js
rachedbenmustapha: It was implied that `RestoreAlreadyInProgress` was the same as what AWS returned, I didn't specifically check. If that code doesn't exist on AWS then of course `InvalidObjectState` should be used.

ghost: Thank you!

PR #4507, opened 2022-05-11
Feature/cldsrv 183 implement restore object api

XinLi0207:
commit1: bump arsenal (based on this pr https://github.com/scality/Arsenal/pull/1888)
commit2: add new API entrance (from Yutaka Oishi)
commit3: add general cold storage util functions (from Yutaka Oishi)
commit4: remove `restore` from unsupported API list
commit5: implement restoreObject api logic(workflow like the most of the other APIs): 
- decode versionId
- validate bucket and object existance
- check if isDeleteMarker, return error
- parse restore request XML (from arsenal, in this pr https://github.com/scality/Arsenal/pull/1888)
- check if tier param in request is supported
- startResotre
  - check error cases
  - update objectMD
commit 6: unit test

### DIFF
--- a/lib/api/apiUtils/object/coldStorage.js
+++ b/lib/api/apiUtils/object/coldStorage.js
@@ -2,6 +2,10 @@
  * Code based on Yutaka Oishi (Fujifilm) contributions
  * Date: 11 Sep 2020
  */
+const ObjectMDArchive = require('arsenal').models.ObjectMDArchive;
+const errors = require('arsenal').errors;
+const { config } = require('../../../Config');
+const { locationConstraints } = config;
 
 /**
  * Get response header "x-amz-restore"
@@ -28,6 +32,130 @@ function getAmzRestoreResHeader(objMD) {
     return undefined;
 }
 
+
+/**
+ * Check if restore can be done.
+ *
+ * @param {ObjectMD} objectMD - object metadata
+ * @param {object} log - werelogs logger
+ * @return {ArsenalError|undefined} - undefined if the conditions for RestoreObject are fulfilled
+ */
+function _validateStartRestore(objectMD, log) {
+    const isLocationCold = locationConstraints[objectMD.dataStoreName]?.isCold;
+    if (!isLocationCold) {
+        // return InvalidObjectState error if the object is not in cold storage,
+        // not in cold storage means either location cold flag not exists or cold flag is explicit false
+        log.debug('The bucket of the object is not in a cold storage location.',
+            {
+                isLocationCold,
+                method: '_validateStartRestore',
+            });
+        return errors.InvalidObjectState;
+    }
+    if (objectMD.archive?.restoreCompletedAt
+        && new Date(objectMD.archive?.restoreWillExpireAt) < new Date(Date.now())) {
+        // return InvalidObjectState error if the restored object is expired
+        // but restore info md of this object has not yet been cleared
+        log.debug('The restored object already expired.',
+            {
+                archive: objectMD.archive,
+                method: '_validateStartRestore',
+            });
+        return errors.InvalidObjectState;
+    }
+    if (objectMD.archive?.restoreRequestedAt && !objectMD.archive?.restoreCompletedAt) {
+        // return RestoreAlreadyInProgress error if the object is currently being restored
+        // check if archive.restoreRequestAt exists and archive.restoreCompletedAt not yet exists
+        log.debug('The object is currently being restored.',
+            {
+                archive: objectMD.archive,
+                method: '_validateStartRestore',
+            });
+        return errors.RestoreAlreadyInProgress;
+    }
+    return undefined;
+}
+
+/**
+ * Check if the object is already restored
+ *
+ * @param {ObjectMD} objectMD - object metadata
+ * @param {object} log - werelogs logger
+ * @return {boolean} - true if the object is already restored
+ */
+function isObjectAlreadyRestored(objectMD, log) {
+    //  check if restoreCompletedAt field exists
+    //  and archive.restoreWillExpireAt > current time
+    const isObjectAlreadyRestored = objectMD.archive?.restoreCompletedAt
+        && new Date(objectMD.archive?.restoreWillExpireAt) >= new Date(Date.now());
+    log.debug('The restore status of the object.',
+        {
+            isObjectAlreadyRestored,
+            method: 'isObjectAlreadyRestored'
+        });
+    return isObjectAlreadyRestored;
+}
+
+/**
+ * update restore expiration date.
+ *
+ * @param {ObjectMD} objectMD - objectMD instance
+ * @param {object} restoreParam - restore param
+ * @param {object} log - werelogs logger
+ * @return {ArsenalError|undefined} internal error if object MD is not valid
+ *
+ */
+function _updateRestoreInfo(objectMD, restoreParam, log) {
+    if (!objectMD.archive) {
+        log.debug('objectMD.archive doesn\'t exits', {
+            objectMD,
+            method: '_updateRestoreInfo'
+        });
+        return errors.InternalError.customizeDescription('Archive metadata is missing.');
+    }
+    // eslint-disable-next-line no-param-reassign
+    objectMD.archive.restoreRequestedAt = new Date();
+    // eslint-disable-next-line no-param-reassign
+    objectMD.archive.restoreRequestedDays = restoreParam.days;
+    if (!ObjectMDArchive.isValid(objectMD.archive)) {
+        log.debug('archive is not valid', {
+            archive: objectMD.archive,
+            method: '_updateRestoreInfo'
+        });
+        return errors.InternalError.customizeDescription('Invalid archive metadata.');
+    }
+    return undefined;
+}
+
+/**
+ * start to restore object.
+ * If not exist x-amz-restore, add it to objectMD.(x-amz-restore = false)
+ * calculate restore expiry-date and add it to objectMD.
+ * Be called by objectRestore.js
+ *
+ * @param {ObjectMD} objectMD - objectMd instance
+ * @param {object} restoreParam - bucket name
+ * @param {object} log - werelogs logger
+ * @param {function} cb - bucket name
+ * @return {undefined}
+ *
+ */
+function startRestore(objectMD, restoreParam, log, cb) {
+    log.info('Validating if restore can be done or not.');
+    const checkResultError = _validateStartRestore(objectMD, log);
+    if (checkResultError) {
+        return cb(checkResultError);
+    }
+    log.info('Updating restore information.');
+    const updateResultError = _updateRestoreInfo(objectMD, restoreParam, log);
+    if (updateResultError) {
+        return cb(updateResultError);
+    }
+    return cb(null, isObjectAlreadyRestored(objectMD, log));
+}
+
+
 module.exports = {
-    getAmzRestoreResHeader,
+    startRestore,
+    getAmzRestoreResHeader
 };
--- a/lib/api/apiUtils/object/objectRestore.js
+++ b/lib/api/apiUtils/object/objectRestore.js
@@ -1,0 +1,153 @@
+const async = require('async');
+const { errors, s3middleware } = require('arsenal');
+
+const { allowedRestoreObjectRequestTierValues } = require('../../../../constants');
+const coldStorage = require('./coldStorage');
+const monitoring = require('../../../utilities/monitoringHandler');
+const { pushMetric } = require('../../../utapi/utilities');
+const { decodeVersionId } = require('./versioning');
+const collectCorsHeaders = require('../../../utilities/collectCorsHeaders');
+const { parseRestoreRequestXml } = s3middleware.objectRestore;
+
+
+/**
+ * Check if tier is supported
+ * @param {object} restoreInfo - restore information
+ * @returns {ArsenalError|undefined} return NotImplemented error if tier not support
+ */
+function checkTierSupported(restoreInfo) {
+    if (!allowedRestoreObjectRequestTierValues.includes(restoreInfo.tier)) {
+        return errors.NotImplemented;
+    }
+    return undefined;
+}
+
+/**
+ * POST Object restore process
+ *
+ * @param {MetadataWrapper} metadata - metadata wrapper
+ * @param {object} mdUtils - utility object to treat metadata
+ * @param {AuthInfo} userInfo - Instance of AuthInfo class with requester's info
+ * @param {IncomingMessage} request - request info
+ * @param {object} log - Werelogs logger
+ * @param {function} callback callback function
+ * @return {undefined}
+ */
+function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
+    const METHOD = 'objectRestore';
+
+    const { bucketName, objectKey } = request;
+
+    log.debug('processing request', { method: METHOD });
+
+    const decodedVidResult = decodeVersionId(request.query);
+    if (decodedVidResult instanceof Error) {
+        log.trace('invalid versionId query',
+            {
+                method: METHOD,
+                versionId: request.query.versionId,
+                error: decodedVidResult,
+            });
+        return process.nextTick(() => callback(decodedVidResult));
+    }
+
+    let isObjectRestored = false;
+
+    const mdValueParams = {
+        authInfo: userInfo,
+        bucketName,
+        objectKey,
+        versionId: decodedVidResult,
+        requestType: 'restoreObject',
+    };
+
+    return async.waterfall([
+            // get metadata of bucket and object
+            function validateBucketAndObject(next) {
+                return mdUtils.metadataValidateBucketAndObj(mdValueParams, log, (err, bucketMD, objectMD) => {
+                    if (err) {
+                        log.trace('request authorization failed', { method: METHOD, error: err });
+                        return next(err);
+                    }
+                    // Call back error if object metadata could not be obtained
+                    if (!objectMD) {
+                        const err = decodedVidResult ? errors.NoSuchVersion : errors.NoSuchKey;
+                        log.trace('error no object metadata found', { method: METHOD, error: err });
+                        return next(err, bucketMD);
+                    }
+                    // If object metadata is delete marker,
+                    // call back NoSuchKey or MethodNotAllowed depending on specifying versionId
+                    if (objectMD.isDeleteMarker) {
+                        let err = errors.NoSuchKey;
+                        if (decodedVidResult) {
+                            err = errors.MethodNotAllowed;
+                        }
+                        log.trace('version is a delete marker', { method: METHOD, error: err });
+                        return next(err, bucketMD, objectMD);
+                    }
+                    log.info('it acquired the object metadata.', {
+                        'method': METHOD,
+                    });
+                    return next(null, bucketMD, objectMD);
+                });
+            },
+
+            // generate restore param obj from xml of request body and check tier validity
+            function parseRequestXmlAndCheckTier(bucketMD, objectMD, next) {
+                log.trace('parsing object restore information');
+                return parseRestoreRequestXml(request.post, log, (err, restoreInfo) => {
+                    if (err) {
+                        return next(err, bucketMD, objectMD, restoreInfo);
+                    }
+                    log.info('it parsed xml of the request body.', { method: METHOD, value: restoreInfo });
+                    const checkTierResult = checkTierSupported(restoreInfo);
+                    if (checkTierResult instanceof Error) {
+                        return next(checkTierResult);
+                    }
+                    return next(null, bucketMD, objectMD, restoreInfo);
+                });
+            },
+            // start restore process
+            function startRestore(bucketMD, objectMD, restoreInfo, next) {
+                return coldStorage.startRestore(objectMD, restoreInfo, log,
+                    (err, _isObjectRestored) => {
+                        isObjectRestored = _isObjectRestored;
+                        return next(err, bucketMD, objectMD);
+                    });
+            },
+            function updateObjectMD(bucketMD, objectMD, next) {
+                const params = objectMD.versionId ? { versionId: objectMD.versionId } : {};
+                metadata.putObjectMD(bucketMD.getName(), objectKey, objectMD, params,
+                    log, err => next(err, bucketMD, objectMD));
+            },
+        ],
+        (err, bucketMD) => {
+            // generate CORS response header
+            const responseHeaders = collectCorsHeaders(request.headers.origin, request.method, bucketMD);
+            if (err) {
+                log.trace('error processing request',
+                    {
+                        method: METHOD,
+                        error: err,
+                    });
+                monitoring.promMetrics(
+                    'POST', bucketName, err.code, 'restoreObject');
+                return callback(err, err.code, responseHeaders);
+            }
+            pushMetric('restoreObject', log, {
+                userInfo,
+                bucket: bucketName,
+            });
+            if (isObjectRestored) {
+                monitoring.promMetrics(
+                    'POST', bucketName, '200', 'restoreObject');
+                return callback(null, 200, responseHeaders);
+            }
+            monitoring.promMetrics(
+                'POST', bucketName, '202', 'restoreObject');
+            return callback(null, 202, responseHeaders);
+        });
+}
+
+
+module.exports = objectRestore;
--- a/lib/api/objectRestore.js
+++ b/lib/api/objectRestore.js
@@ -1,0 +1,29 @@
+/*
+ * Code based on Yutaka Oishi (Fujifilm) contributions
+ * Date: 11 Sep 2020
+ *
+ * This module handles POST Object restore.
+ *
+ * @module api/objectRestore
+ */
+
+const metadata = require('../metadata/wrapper');
+const metadataUtils = require('../metadata/metadataUtils');
+
+const sdtObjectRestore = require('./apiUtils/object/objectRestore');
+
+/**
+ * Process POST Object restore request.
+ *
+ * @param {AuthInfo} userInfo Instance of AuthInfo class with requester's info
+ * @param {object} request http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+function objectRestore(userInfo, request, log, callback) {
+    return sdtObjectRestore(metadata, metadataUtils, userInfo, request,
+        log, callback);
+}
+
+module.exports = objectRestore;
--- a/tests/unit/api/objectRestore.js
+++ b/tests/unit/api/objectRestore.js
@@ -1,0 +1,159 @@
+const assert = require('assert');
+
+const { bucketPut } = require('../../../lib/api/bucketPut');
+const objectPut = require('../../../lib/api/objectPut');
+const objectRestore = require('../../../lib/api/objectRestore');
+const { cleanup, DummyRequestLogger, makeAuthInfo } = require('../helpers');
+const mdColdHelper = require('./utils/metadataMockColdStorage');
+const DummyRequest = require('../DummyRequest');
+const metadata = require('../metadataswitch');
+
+const log = new DummyRequestLogger();
+const authInfo = makeAuthInfo('accessKey1');
+const namespace = 'default';
+const bucketName = 'bucketname';
+const objectName = 'objectName';
+const postBody = Buffer.from('I am a body', 'utf8');
+const restoreDays = 1;
+
+const bucketPutRequest = {
+    bucketName,
+    headers: { host: `${bucketName}.s3.amazonaws.com` },
+    url: '/',
+};
+
+const putObjectRequest = new DummyRequest({
+    bucketName,
+    namespace,
+    objectKey: objectName,
+    headers: {},
+    url: `/${bucketName}/${objectName}`,
+}, postBody);
+
+const objectRestoreXml = '<RestoreRequest ' +
+    'xmlns="http://s3.amazonaws.com/doc/2006-03-01/">' +
+    `<Days>${restoreDays}</Days>` +
+    '<Tier>Standard</Tier>' +
+    '</RestoreRequest>';
+
+const objectRestoreXmlBulkTier = '<RestoreRequest ' +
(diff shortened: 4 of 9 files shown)


lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Check if restore can be done.
+ *
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - object metadata
+ * @return {Error|boolean} - true if the object is already restored
+ */
+function _validateStartRestore(bucketMD, objectMD) {
+    const isLocationCold = locationConstraints[bucketMD.getLocationConstraint()]?.cold;
```
rachedbenmustapha: I think the flag in the configuration is `isCold` (confirm with @naren-rajendran).

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,107 @@
+const ObjectMDArchive = require('arsenal').models.ObjectMDArchive;
+const errors = require('arsenal').errors;
+const { config } = require('../../../Config');
+const { locationConstraints } = config;
+
+/**
+ * Check restore status.
+ *
+ * @param {ObjectMD} objectMD - object metadata
+ * @return {undefined}
```
rachedbenmustapha: The return type doesn't seem to be `undefined`.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,107 @@
...
+ */
+function _checkRestoreStatus(objectMD) {
```
rachedbenmustapha: APIs with this kind of mixed return types are not very developer-friendly and are extremely error-prone (the fact that javascript allows it should not be seen as an encouragement).

It seems to me that this whole logic is just a yes/no question: can we initiate a new restore or not? There's no real need for an error (or even different variations of the status) when doing the computation, the error is just part of the response to the user.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,107 @@
...
+        return errors.InvalidObjectState;
+    }
+}
+
+/**
+ * Check if restore can be done.
+ *
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - object metadata
+ * @return {Error|boolean} - true if the object is already restored
```
rachedbenmustapha: Same issue here with the mixed return types. There doesn't seem to be a need for a boolean, as the different errors (of lack of error) already conveys the information.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,107 @@
...
+function _updateRestoreInfo(bucketMD, objectMD, restoreParam, log) {
+    try {
+        // eslint-disable-next-line no-param-reassign
+        objectMD.archive.restoreRequestedAt = new Date();
+        // eslint-disable-next-line no-param-reassign
+        objectMD.archive.restoreRequestedDays = restoreParam.days;
+        if (!ObjectMDArchive.isValid(objectMD.archive)) {
+            throw new Error('archive object is not valid');
+        }
+    } catch (e) {
```
rachedbenmustapha: Why throw then catch within the same scope?

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,107 @@
...
+        log.debug(e.message, { md: objectMD.archive });
+        return errors.InternalError;
```
rachedbenmustapha: I'm not sure that `InternalError` is the right kind of error (we use it too much to the point of not being very helpful).

tests/unit/api/objectRestore.js
```diff
@@ -0,0 +1,170 @@
...
+    it('should update the expiry time and return OK ' +
+        'while restoring an object from cold storage ' +
+        'and the object have a restored copy in bucket', done => {
+        mdColdHelper.putBucketMock(bucketName, null, () => {
+            mdColdHelper.putObjectMock(bucketName, objectName, mdColdHelper.getArchiveRestoredMD(), () => {
+                objectRestore(authInfo, objectRestoreRequest(objectRestoreXml), log, (err, statusCode) => {
+                    assert.ifError(err);
+                    assert.strictEqual(statusCode, 200);
+                    metadata.getObjectMD(bucketName, objectName, {}, log, (err, md) => {
+                            assert.strictEqual(md.archive.restoreRequestedDays, restoreDays);
```
rachedbenmustapha: Assertions on the requestedAt date would be useful as well.

rachedbenmustapha (changes requested):

lib/api/apiUtils/object/coldStorage.js
XinLi0207: agree with you. changed to use one function to only check if we can initiate a restore(return error on failure, return undefined on success)

lib/api/apiUtils/object/coldStorage.js
XinLi0207: removed the try catch.

lib/api/apiUtils/object/coldStorage.js
XinLi0207: yeah, actually here `ObjectMDArchive.isValid(objectMD.archive)` won't be false because we already validated these params [here](https://github.com/scality/cloudserver/pull/4507/files#diff-6c0b3b4adb50fc9f073fc3ec5b60328a076859fbe8d1917857f08bee6e1f7e89R90)

just to be extra safe, if we got false here that means there is an issue in our code, so I use InternalError here

tests/unit/api/objectRestore.js
XinLi0207: solved: https://github.com/scality/cloudserver/pull/4507/commits/e89def3e17f447da1eb9bbbcafba9d274f10160d

lib/api/apiUtils/object/coldStorage.js
ghost: Maybe the `BadMetadataDocument` can be used here?

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,99 @@
...
+const { config } = require('../../../Config');
+const { locationConstraints } = config;
+
+
+/**
+ * Check if restore can be done.
+ *
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - object metadata
+ * @return {Error|boolean} - true if the object is already restored
```
ghost: How about something like this? The function does not return any boolean anymore
```suggestion
 * @return {ArsenalError|undefined} - undefined if the conditions for RestoreObject are fulfilled
```

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+function _validateStartRestore(bucketMD, objectMD) {
+    const isLocationCold = locationConstraints[bucketMD.getLocationConstraint()]?.isCold;
+    if (isLocationCold === undefined || isLocationCold === false) {
```
ghost: A falsy value could be simplified with:
```suggestion
    if (!isLocationCold) {
```

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * update restore expiration date.
+ *
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - objectMD instance
+ * @param {object} restoreParam - restore param
+ * @param {object} log - werelogs logger
+ * @return {undefined|Error} internal error if object MD is not valid
```
ghost:
```suggestion
 * @return {ArsenalError|undefined} internal error if object MD is not valid
```

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+function startRestore(bucketMD, objectMD, restoreParam, log, cb) {
+    const checkResult = _validateStartRestore(bucketMD, objectMD);
+    if (checkResult instanceof Error) {
+        return cb(checkResult);
+    }
+    const updateResult = _updateRestoreInfo(bucketMD, objectMD, restoreParam, log);
+    if (checkResult instanceof Error) {
+        return cb(updateResult);
+    }
```
ghost: I think it should be the `updateResult` here.

```suggestion
    if (updateResult instanceof Error) {
        return cb(updateResult);
    }
```

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,146 @@
...
+
+const { allowedRestoreObjectRequestTierValues } = require('../../../../constants');
+const coldStorage = require('./coldStorage');
+const monitoring = require('../../../utilities/monitoringHandler');
+const { pushMetric } = require('../../../utapi/utilities');
+const { parseRestoreRequestXml } = s3middleware.objectRestore;
+
+const METHOD = 'objectRestore';
+
+
```
ghost:
```suggestion

```

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,146 @@
+const async = require('async');
+const { errors, s3middleware } = require('arsenal');
+
+const { allowedRestoreObjectRequestTierValues } = require('../../../../constants');
+const coldStorage = require('./coldStorage');
+const monitoring = require('../../../utilities/monitoringHandler');
+const { pushMetric } = require('../../../utapi/utilities');
+const { parseRestoreRequestXml } = s3middleware.objectRestore;
+
+const METHOD = 'objectRestore';
```
ghost: I would suggest putting this const in the `objectRestore` function, no need to have it as a global variable.

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,146 @@
...
+ * @param {function(object):string|Error} func.decodeVersionId - decodeVersionId function
+ * @param {function(object, string, BucketInfo):object} func.collectCorsHeaders - collectCorsHeaders function
+ * @param {function(object, object):string} func.getVersionIdResHeader - getVersionIdResHeader function
+ * @param {AuthInfo} userInfo - Instance of AuthInfo class with requester's info
+ * @param {IncomingMessage} request - request info
+ * @param {object} log - Werelogs logger
+ * @param {function} callback callback function
+ * @return {undefined}
+ */
+function objectRestore(metadata, mdUtils, func, userInfo, request, log, callback) {
```
ghost: Can we use a more explicit name for the `func` parameter? 
As the functions in the object are coming from imports in `lib/api/objectRestore.js`, maybe this parameter can be deleted and you can `require` the functions in this file directly.

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,146 @@
...
+                method: METHOD,
+                versionId: request.query.versionId,
+                error: decodedVidResult,
+            });
+        return process.nextTick(() => callback(decodedVidResult));
+    }
+
+    let isObjectRestored = false;
+
+    const reqVersionId = decodedVidResult;
```
ghost: Why not just use `decodedVidResult` if you assign its value to a const?

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,146 @@
...
+            // generate CORS response header
+            const responseHeaders = func.collectCorsHeaders(request.headers.origin, request.method, bucketMD);
+            if (err) {
+                log.trace('error processing request',
+                    {
+                        method: METHOD,
+                        error: err,
+                    });
+                monitoring.promMetrics(
+                    'PUT', bucketName, err.code, 'restoreObject');
```
ghost:
```suggestion
                    'POST', bucketName, err.code, 'restoreObject');
```

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,146 @@
...
+                return callback(err, err.code, responseHeaders);
+            }
+            pushMetric('restoreObject', log, {
+                userInfo,
+                bucket: bucketName,
+            });
+            if (isObjectRestored) {
+                monitoring.promMetrics(
+                    'PUT', bucketName, '200', 'restoreObject');
```
ghost:
```suggestion
                    'POST', bucketName, '200', 'restoreObject');
```

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,146 @@
...
+                return callback(null, 200, responseHeaders);
+            }
+            monitoring.promMetrics(
+                'PUT', bucketName, '202', 'restoreObject');
```
ghost:
```suggestion
                'POST', bucketName, '202', 'restoreObject');
```

ghost (commented):
Some code-related comments :slightly_smiling_face: 
I will try to re-review the logic tomorrow

ghost (approved):
LGTM

lib/api/apiUtils/object/coldStorage.js
rachedbenmustapha: At least `InternalError.customizeDescription(...)` would be more useful.

lib/api/objectRestore.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * This module handles POST Object restore.
+ *
+ * @module api/objectRestore
+ */
+
+const metadata = require('../metadata/wrapper');
+const metadataUtils = require('../metadata/metadataUtils');
+
+const sdtObjectRestore = require('./apiUtils/object/objectRestore');
```
miniscruff, line 13: What does sdt stand for here?

XinLi0207: actually, it's from Fujifilm's contribution, we thought of keeping the most of their code, so didn't change here

lib/api/apiUtils/object/coldStorage.js
XinLi0207: done: https://github.com/scality/cloudserver/pull/4507/commits/a7016fef2534bc31a385438e9cd86745b1eee72e

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,141 @@
...
+            // start restore process
+            function startRestore(bucketMD, objectMD, restoreInfo, next) {
+                return coldStorage.startRestore(bucketMD, objectMD, restoreInfo, log,
+                    (err, _isObjectRestored) => {
+                        isObjectRestored = _isObjectRestored;
+                        return next(err, bucketMD, objectMD);
+                    });
+            },
+            function updateObjectMD(bucketMD, objectMD, next) {
+                metadata.putObjectMD(bucketMD.getName(), objectKey, objectMD, { archive: objectMD.archive },
```
nicolas2bert: The fourth argument of the `putObjectMD`  is the "versioning parameters". The archive's information has already been updated in `startRestore`.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +33,99 @@ function getAmzRestoreResHeader(objMD) {
...
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - objectMD instance
+ * @param {object} restoreParam - restore param
+ * @param {object} log - werelogs logger
+ * @return {ArsenalError|undefined} internal error if object MD is not valid
+ *
+ */
+function _updateRestoreInfo(bucketMD, objectMD, restoreParam, log) {
+    // eslint-disable-next-line no-param-reassign
+    objectMD.archive.restoreRequestedAt = new Date();
```
nicolas2bert, line 117: I assume objectMD.archive will always be defined since `objectMD.archive.archiveInfo` will be set during the archival process. However, I would make sure `Cannot set properties of undefined` error would not be thrown if `objectMD.archive` happens to be undefined.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +33,99 @@ function getAmzRestoreResHeader(objMD) {
...
+    // eslint-disable-next-line no-param-reassign
+    objectMD.archive.restoreRequestedDays = restoreParam.days;
+    if (!ObjectMDArchive.isValid(objectMD.archive)) {
+        log.debug('archive is not valid', { archive: objectMD.archive });
```
nicolas2bert: Please add `method: "_updateRestoreInfo"` to the log.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +33,99 @@ function getAmzRestoreResHeader(objMD) {
...
 
+
+/**
+ * Check if restore can be done.
+ *
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - object metadata
+ * @return {ArsenalError|boolean} - undefined if the conditions for RestoreObject are fulfilled
+ */
+function _validateStartRestore(bucketMD, objectMD) {
```
nicolas2bert: Adding some error logs might help to debug.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +33,99 @@ function getAmzRestoreResHeader(objMD) {
...
+ *
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - objectMd instance
+ * @param {object} restoreParam - bucket name
+ * @param {object} log - werelogs logger
+ * @param {function} cb - bucket name
+ * @return {undefined}
+ *
+ */
+function startRestore(bucketMD, objectMD, restoreParam, log, cb) {
```
nicolas2bert: This function seems to be synchronous.

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,141 @@
...
+                log.trace('parsing object restore information');
+                return parseRestoreRequestXml(request.post, log, (err, restoreInfo) => {
+                    if (err) {
+                        return next(err, bucketMD, objectMD, restoreInfo);
+                    }
+                    log.info('it parsed xml of the request body.', { method: METHOD, value: restoreInfo });
+                    return next(null, bucketMD, objectMD, restoreInfo);
+                });
+            },
+            function checkTierSupported(bucketMD, objectMD, restoreInfo, next) {
```
nicolas2bert: The waterfall is expecting asynchronous function as task. (Same for the following function)

nicolas2bert (changes requested):

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -0,0 +1,141 @@
...
+                if (!allowedRestoreObjectRequestTierValues.includes(restoreInfo.tier)) {
+                    return next(errors.NotImplemented, bucketMD, objectMD);
+                }
+                return next(null, bucketMD, objectMD, restoreInfo);
+            },
+            // start restore process
+            function startRestore(bucketMD, objectMD, restoreInfo, next) {
+                return coldStorage.startRestore(bucketMD, objectMD, restoreInfo, log,
+                    (err, _isObjectRestored) => {
+                        isObjectRestored = _isObjectRestored;
```
nicolas2bert, line 114: Should we handle "updating the restoration period" during "POST Object restore process" (updating `restoreRequestedAt`, `restoreRequestedDays`, `restoreCompletedAt` and `restoreWillExpireAt`) without going through the whole restoring flow (ie get object from tape, putObject version...)? @RachedBenMustapha 

Reference for update the restoration period:
"After restoring an archived object, you can update the restoration period by reissuing the request with a new period. Amazon S3 updates the restoration period relative to the current time and charges only for the request-there are no data transfer charges. You cannot update the restoration period when Amazon S3 is actively processing your current restore request for the object."
from https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +33,99 @@ function getAmzRestoreResHeader(objMD) {
...
+
+/**
+ * Check if restore can be done.
+ *
+ * @param {BucketInfo} bucketMD - bucket instance
+ * @param {ObjectMD} objectMD - object metadata
+ * @return {ArsenalError|boolean} - undefined if the conditions for RestoreObject are fulfilled
+ */
+function _validateStartRestore(bucketMD, objectMD) {
+    const isLocationCold = locationConstraints[bucketMD.getLocationConstraint()]?.isCold;
```
nicolas2bert: We should not be able to create a bucket in a cold storage location. Would this always be false?

XinLiScality: I think the archive process will update the locationConstraint of bucketMD to a cold storage location, so that this should be true after object being archived

lib/api/apiUtils/object/objectRestore.js
XinLi0207: done: https://github.com/scality/cloudserver/pull/4507/commits/4bf88225092246150b1f22c1fba17a36d6b0a5f1

lib/api/apiUtils/object/coldStorage.js
XinLi0207: I added a logic to check the existence of objectMD.archive if it not exist, return an InternalError
done: https://github.com/scality/cloudserver/pull/4507/commits/4bf88225092246150b1f22c1fba17a36d6b0a5f1

lib/api/apiUtils/object/coldStorage.js
XinLi0207: done https://github.com/scality/cloudserver/pull/4507/commits/4bf88225092246150b1f22c1fba17a36d6b0a5f1

lib/api/apiUtils/object/coldStorage.js
XinLi0207: done https://github.com/scality/cloudserver/pull/4507/commits/4bf88225092246150b1f22c1fba17a36d6b0a5f1

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -37,21 +35,17 @@ function getAmzRestoreResHeader(objMD) {
...
- * @param {BucketInfo} bucketMD - bucket instance
  * @param {ObjectMD} objectMD - object metadata
  * @param {object} log - werelogs logger
  * @return {ArsenalError|boolean} - undefined if the conditions for RestoreObject are fulfilled
  */
-function _validateStartRestore(bucketMD, objectMD, log) {
-    const isLocationCold = locationConstraints[bucketMD.getLocationConstraint()]?.isCold
-        || locationConstraints[objectMD.dataStoreName]?.isCold;
+function _validateStartRestore(objectMD, log) {
+    const isLocationCold = !!objectMD.archive;
```
XinLi0207: hi @nicolas2bert  @naren-rajendran, 
we changed the way to check if the object is stored in cold storage or hot. 
Since bucketMD.locationConstraint might not change when we archive an object of the bucket, I removed the line checking bucket locationConstraint.
we use `archive` field of objectMD to check, we assume if this field exists, then the object should have been archive.
Could you please confirm with us that if this check look good?

cc @williamlardier

lib/api/apiUtils/object/objectRestore.js
XinLi0207: done https://github.com/scality/cloudserver/pull/4507/commits/8a706cef41d9750b815dbb593a49b301bd271a26

lib/api/apiUtils/object/objectRestore.js
XinLi0207: done: https://github.com/scality/cloudserver/pull/4507/commits/8a706cef41d9750b815dbb593a49b301bd271a26

nicolas2bert (approved):

naren-rajendran (approved):

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +31,131 @@ function getAmzRestoreResHeader(objMD) {
     return undefined;
 }
 
+
+/**
+ * Check if restore can be done.
+ *
+ * @param {ObjectMD} objectMD - object metadata
+ * @param {object} log - werelogs logger
+ * @return {ArsenalError|boolean} - undefined if the conditions for RestoreObject are fulfilled
```
rachedbenmustapha: The return type is wrong.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +31,131 @@ function getAmzRestoreResHeader(objMD) {
...
+ * @param {object} restoreParam - bucket name
+ * @param {object} log - werelogs logger
+ * @param {function} cb - bucket name
+ * @return {undefined}
+ *
+ */
+function startRestore(objectMD, restoreParam, log, cb) {
+    log.info('Validating if restore can be done or not.');
+    const checkResult = _validateStartRestore(objectMD, log);
+    if (checkResult instanceof Error) {
```
rachedbenmustapha: This seems to be a leftover from the old signature. `checkResult` is always an error so return if it's truthy, and for clarity it can be named something that hints at an error.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +31,131 @@ function getAmzRestoreResHeader(objMD) {
...
+        return cb(checkResult);
+    }
+    log.info('Updating restore information.');
+    const updateResult = _updateRestoreInfo(objectMD, restoreParam, log);
+    if (updateResult instanceof Error) {
```
rachedbenmustapha: Same here, it's always an error.

lib/api/apiUtils/object/coldStorage.js
rachedbenmustapha: This check looks wrong, as it has become a heuristic (educated guess based on indirect information).

My testing on real AWS S3 shows that the storage class changes on the object after transition, which translates in our implementation into the new cold location as the object's storageclass. This makes testing the actual location referenced by the object for the `isCold` flag the most reliable test.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +31,131 @@ function getAmzRestoreResHeader(objMD) {
...
+ * @param {ObjectMD} objectMD - object metadata
+ * @param {object} log - werelogs logger
+ * @return {boolean} - true if the object is already restored
+ */
+function isObjectAlreadyRestored(objectMD, log) {
+    //  check if restoreCompletedAt field exists
+    //  and archive.restoreWillExpireAt > current time
+    const isObjectAlreadyRestored = objectMD.archive?.restoreCompletedAt
+        && new Date(objectMD.archive?.restoreWillExpireAt) >= new Date(Date.now());
+    log.info(`Object is ${isObjectAlreadyRestored ? '' : 'not'} already restored.`,
```
rachedbenmustapha: This is not `info` material. As an aside, the message itself should be immutable, the attached object can contain the extra info. Please follow [the guidelines](https://github.com/scality/Guidelines/blob/development/8.2/CONTRIBUTING.md#logging) for formatting.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +31,131 @@ function getAmzRestoreResHeader(objMD) {
...
+    }
+    if (objectMD.archive?.restoreRequestedAt && !objectMD.archive?.restoreCompletedAt) {
+        // return RestoreAlreadyInProgress error if the object is currently being restored
+        // check if archive.restoreRequestAt exists and archive.restoreCompletedAt not yet exists
+        log.debug('The object is currently being restored.',
+            {
+                archive: objectMD.archive,
+                method: '_validateStartRestore',
+            });
+        return errors.RestoreAlreadyInProgress;
```
rachedbenmustapha, line 74: Following the design update I thought `InvalidObjectState` was the error to return?

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +31,131 @@ function getAmzRestoreResHeader(objMD) {
...
+ * @return {ArsenalError|undefined} internal error if object MD is not valid
+ *
+ */
+function _updateRestoreInfo(objectMD, restoreParam, log) {
+    if (!objectMD.archive) {
+        log.debug('objectMD.archive doesn\'t exits', {
+            objectMD,
+            method: '_updateRestoreInfo'
+        });
+        return errors.InternalError.customizeDescription('objectMD.archive doesn\'t exit for an archived object.');
```
rachedbenmustapha: Since users could theoretically see it, they don't have to see this kind of detail, can be like "Archive metadata missing", it's enough for us to know, and doesn't expose jargon to the user.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -28,6 +31,131 @@ function getAmzRestoreResHeader(objMD) {
...
+    objectMD.archive.restoreRequestedAt = new Date();
+    // eslint-disable-next-line no-param-reassign
+    objectMD.archive.restoreRequestedDays = restoreParam.days;
+    if (!ObjectMDArchive.isValid(objectMD.archive)) {
+        log.debug('archive is not valid', {
+            archive: objectMD.archive,
+            method: '_updateRestoreInfo'
+        });
+        return errors.InternalError.customizeDescription('Invalid archive object. ' +
+            'Please check ObjectMDArchive validation function.');
```
rachedbenmustapha: Same here, the end-user of the S3 service is certainly not going to check the validation function. And the log should not be debug, but error, to help find the error cause.

If it's a serious invariant that is violated, an assert+crash would even be suitable (but not the case here).

rachedbenmustapha (changes requested):

lib/api/apiUtils/object/coldStorage.js
XinLi0207: the updated design to use `InvalidObjectState` for all cases is only for `GetObject` API,  won't change `restoreObject` API 
https://github.com/scality/citadel/pull/112/files#diff-a8bd259bf0a5bae573753085d126a597616b071b2a3422933f0da353090d886dL125

rachedbenmustapha (approved):

PR #4531, opened 2022-05-13
CLDSRV-202 Put object version with x-scal-s3-version-id

nicolas2bert:
Put object data without changing the metadata.
We recalculate content MD5, content length, ... to make sure the metadata is accurate. (Could be skipped if calculation is redundant).

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -10,6 +10,38 @@ const versionIdUtils = versioning.VersionID;
 const nonVersionedObjId =
     versionIdUtils.getInfVid(config.replicationGroupId);
 
+/** _decodePutVersionId - decode the version id from x-scal-s3-version-id header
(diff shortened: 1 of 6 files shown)


lib/api/apiUtils/object/versioning.js
```diff
@@ -370,6 +402,69 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD,
...
+
+    // CASE 2: overwriting a version id.
+    const options = {
+        versionId: decodedPutVersionId,
+    };
+
+    // NOTE: We need to retrieve the specific object version ("versionMD") to ensure
+    //       that the rest of the original object properties remain intact.
+    // NOTE: "objMD" holds the master version properties,
+    return metadata.getObjectMD(bucketName, metadataStoreParams.objectKey, options, log,
```
nicolas2bert, line 451: Currently looking for a clean way to leverage objMD and avoid having to call object metadata twice.

PR #4534, opened 2022-05-16
CLDSRV-202 Put object version with x-scal-s3-version-id

nicolas2bert:
When restoring an object, the tape library will send a S3 put object with `x-scal-s3-version-id` header.

This PR:
- supports the new header
- writes new location
- ensures that the rest of the original object properties remain intact ie last-modified date etc.

_Note: We recalculate content MD5, content length to make sure the metadata is accurate._

_Next: multi-part upload supports `x-scal-s3-version-id` header._

_Next: update object metadata restore_

### DIFF
--- a/lib/api/objectPut.js
+++ b/lib/api/objectPut.js
@@ -7,7 +7,7 @@ const { cleanUpBucket } = require('./apiUtils/bucket/bucketCreation');
 const { getObjectSSEConfiguration } = require('./apiUtils/bucket/bucketEncryption');
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const createAndStoreObject = require('./apiUtils/object/createAndStoreObject');
-const { checkQueryVersionId } = require('./apiUtils/object/versioning');
+const { checkQueryVersionId, decodeVID } = require('./apiUtils/object/versioning');
 const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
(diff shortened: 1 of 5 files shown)


miniscruff (approved):

lib/api/objectPut.js
```diff
@@ -41,6 +41,24 @@ const versionIdUtils = versioning.VersionID;
  */
 function objectPut(authInfo, request, streamingV4Params, log, callback) {
     log.debug('processing request', { method: 'objectPut' });
+
```
vrancurel, line 44: Can you also update the microVersionId ? (look at object model)

`models/ObjectMD.js: updateMicroVersionId()`

This will be used to know internally that there was an update.

vrancurel (changes requested):

lib/api/objectPut.js
nicolas2bert: `updateMicroVersionId()` has not been introduced in Cloudserver (for putObjectTagging...).
Would it not make sense to have a specific PR for `updateMicroVersionId()` ?

nicolas2bert: Update, after discussing with @vrancurel , we decided to introduce `updateMicroVersionId()` in this PR.

vrancurel (approved):

nicolas2bert:
@bert-e approve

PR #4539, opened 2022-05-18
CLDSRV-205 Update restore properties when putting object version

nicolas2bert:
Update `archive.restoreCompletedAt` and  `archive.restoreWillExpireAt`
Note: we do not use `putObjectVersion` to update the restoration period.

Design information from https://github.com/scality/citadel/blob/development/1.0/docs/design/cold-storage.md:
- In Zenko, the lifecycle transition workflow will change the object location: metadata and not its storage class (which is ignored).

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -4,6 +4,8 @@ const async = require('async');
 const metadata = require('../../../metadata/wrapper');
 const { config } = require('../../../Config');
 
+const oneDay = 24 * 60 * 60 * 1000;
+
 const versionIdUtils = versioning.VersionID;
 // Use Arsenal function to generate a version ID used internally by metadata
 // for null versions that are created before bucket versioning is configured
@@ -401,6 +403,18 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     metadataStoreParams.creationTime = objMD['creation-time'];
     metadataStoreParams.lastModifiedDate = objMD['last-modified'];
     metadataStoreParams.updateMicroVersionId = true;
+
+    // update restore
+    const days = objMD.archive?.restoreRequestedDays;
+    const now = Date.now();
+    metadataStoreParams.archive = {
+        archiveInfo: objMD.archive?.archiveInfo,
+        restoreRequestedAt: objMD.archive?.restoreRequestedAt,
+        restoreRequestedDays: objMD.archive?.restoreRequestedDays,
+        restoreCompletedAt: new Date(now),
+        restoreWillExpireAt: new Date(now + (days * oneDay)),
+    };
+
     /* eslint-enable no-param-reassign */
 
     const versionId = objMD.versionId || undefined;
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -143,10 +143,26 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 oldByteLength = objMD['content-length'];
             }
 
-            if (isPutVersion && !objMD) {
-                const err = putVersionId ? errors.NoSuchVersion : errors.NoSuchKey;
-                log.trace('error no object metadata found', { method: 'completeMultipartUpload', putVersionId });
-                return callback(err);
+            if (isPutVersion) {
+                if (!objMD) {
+                    const err = putVersionId ? errors.NoSuchVersion : errors.NoSuchKey;
+                    log.error('error no object metadata found', { method: 'completeMultipartUpload', putVersionId });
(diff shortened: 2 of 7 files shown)


lib/api/apiUtils/object/versioning.js
```diff
@@ -401,6 +403,18 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     metadataStoreParams.creationTime = objMD['creation-time'];
     metadataStoreParams.lastModifiedDate = objMD['last-modified'];
     metadataStoreParams.updateMicroVersionId = true;
+
+    // update restore
+    const days = objMD.archive?.restoreRequestedDays;
```
miniscruff, line 408: This question mark means it is optional but later on we use it as though it was required.

lib/api/apiUtils/object/versioning.js
```diff
@@ -401,6 +403,18 @@ function overwritingVersioning(objMD, metadataStoreParams) {
...
+    const now = Date.now();
+    metadataStoreParams.archive = {
+        archiveInfo: objMD.archive?.archiveInfo,
+        restoreRequestedAt: objMD.archive?.restoreRequestedAt,
+        restoreRequestedDays: objMD.archive?.restoreRequestedDays,
+        restoreCompletedAt: new Date(now),
+        restoreWillExpireAt: new Date(now + (days * oneDay)),
+    };
```
ghost, line 416: I think it would be great here to call the static `isValid` function from https://github.com/scality/Arsenal/blob/fb39a4095e3429e98e743106dfbf36c6f19682d5/lib/models/ObjectMDArchive.js#L36, to make sure the object is in a valid state. Because this is a "complex" object, we have implemented this function for such cases, to make sure that we don't write corrupted data (this shouldn't be the case here, but checking doesn't hurt!)

What do you think? :smile:

lib/api/apiUtils/object/versioning.js
nicolas2bert: Yes, earlier in the API process we check that `objMD.archive.restoreRequestedDays` is defined.
Since `overwritingVersioning()` is exported and meant to be reused, I used this question mark as a safeguard.

lib/api/apiUtils/object/versioning.js
nicolas2bert: I don't think it is needed here.
It would be redundant with: https://github.com/scality/cloudserver/pull/4539/files#diff-a76cecfcb0b408028fb3c7e02359a47730668d6424f8b02bb9522c162c21ce47R173

miniscruff (approved):

ghost (approved):

lib/api/apiUtils/object/versioning.js
ghost: Right, thanks

lib/api/apiUtils/object/versioning.js
```diff
@@ -401,6 +403,18 @@ function overwritingVersioning(objMD, metadataStoreParams) {
...
     metadataStoreParams.lastModifiedDate = objMD['last-modified'];
     metadataStoreParams.updateMicroVersionId = true;
+
+    // update restore
+    const days = objMD.archive?.restoreRequestedDays;
+    const now = Date.now();
+    metadataStoreParams.archive = {
+        archiveInfo: objMD.archive?.archiveInfo,
+        restoreRequestedAt: objMD.archive?.restoreRequestedAt,
+        restoreRequestedDays: objMD.archive?.restoreRequestedDays,
```
XinLi0207, line 413: LGTM, just a small question: can these values not exist in objectMD when put object version? do we need to throw an error if objectMD.archive doesn't exist?

nicolas2bert: We check earlier that these value does not exist: https://github.com/scality/cloudserver/pull/4539/files#diff-f2ca335d4dd243e368c6c2752d4aa2ac20fe772d94b2a21cd877a235a7ef0850R161 
Note: we do not use `putObjectVersion` to update the restoration period.

We already return an error if `objectMD.archive` doesn't exist https://github.com/scality/cloudserver/pull/4539/files#diff-f2ca335d4dd243e368c6c2752d4aa2ac20fe772d94b2a21cd877a235a7ef0850R160

XinLi0207: ok I see. right,  [this check](https://github.com/scality/cloudserver/pull/4539/files#diff-f2ca335d4dd243e368c6c2752d4aa2ac20fe772d94b2a21cd877a235a7ef0850R161) happens before here

XinLi0207 (approved):

nicolas2bert:
@bert-e approve

PR #4549, opened 2022-05-25
Remove default replication endpoint mandate

francoisferrand:
The default replication endpoint should not be needed, esp. since the
one we use (in zenko) does not relate to an existing location: it is
used only to let Cloudserver start.

Issue: CLDSRV-211

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -560,12 +560,6 @@ class Config extends EventEmitter {
             const { replicationEndpoints } = config;
             assert(replicationEndpoints instanceof Array, 'bad config: ' +
                 '`replicationEndpoints` property must be an array');
-            if (replicationEndpoints.length > 1) {
-                const hasDefault = replicationEndpoints.some(
-                    replicationEndpoint => replicationEndpoint.default);
-                assert(hasDefault, 'bad config: `replicationEndpoints` must ' +
-                    'contain a default endpoint');
-            }
             replicationEndpoints.forEach(replicationEndpoint => {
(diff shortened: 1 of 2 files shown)


nicolas2bert (commented):
From what I remember the default replication endpoint was there because:
When putBucketReplication, we specify the following parameters:
```
{
      Bucket: 'your source bucket',
      ReplicationConfiguration: {
        Role: 'roleArn',
        Rules: [{
          Destination: {StorageClass: "destination location", Bucket: "prefix of destination bucket"},
          Prefix: '',
          Status: 'Enabled',
        }],
      },
    }
```
From S3 documentation, “Rule.Destination.StorageClass” is optional, however, this information is key in the replication logic (shows in which location we replicate the objects).
If it is not provided, we use the site mentioned in replicationEndpoints with .default === true

francoisferrand:
> If it is not provided, we use the site mentioned in replicationEndpoints with .default === true

indeed, this is the code I changed in this PR.
we used to require a "default" replicationEndpoint, but ended up just creating a dummy one to make cloudserver start (and I could not get better explanation). So I updated the logic so that, when there is no explicit endpoint, we take:
* the one marked as `.default === true`, if any
* else the first one
* else return an error

miniscruff (approved):
I agree the the code does what the comment says, but can not say myself if this is the right approach. I would like a test if possible either way.

rachedbenmustapha (approved):
The change seems to make sense, but I can't be sure that no regression will come from it.

nicolas2bert (approved):

PR #4560, opened 2022-06-02
CLDSRV-215 S3.DeleteObjects should clean up data

nicolas2bert:
Deleting a version should delete its data on the backend side.

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -233,6 +233,12 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                         log.debug('trying to delete locked object');
                         return callback(objectLockedError);
                     }
+                    if (versionId && objMD.location &&
+                        Array.isArray(objMD.location) && objMD.location[0]) {
+                        // we need this information for data deletes to AWS
+                        // eslint-disable-next-line no-param-reassign
+                        objMD.location[0].deleteVersion = true;
+                    }
                     return callback(null, objMD, versionId);
                 }),
             (objMD, versionId, callback) =>
--- a/tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
+++ b/tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
@@ -76,6 +76,32 @@ function delAndAssertResult(s3, params, cb) {
     });
 }
 
+function delObjectsAndAssertResult(s3, params, cb) {
+    const { bucket, key, versionId, resultType, resultError } = params;
+    const deleteParams = {
+        Objects: [
+            {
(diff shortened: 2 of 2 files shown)


nicolas2bert:
ping

lib/api/multiObjectDelete.js
```diff
@@ -233,6 +233,12 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                         log.debug('trying to delete locked object');
                         return callback(objectLockedError);
                     }
+                    if (versionId && objMD.location &&
+                        Array.isArray(objMD.location) && objMD.location[0]) {
```
francoisferrand, line 237: nit: may be easier to read with a function "wrapping" the `objMD.location && Array.isArray(objMD.location) && objMD.location[0]` expression behind a meaningful name (like `isRemoteObject` or whatever this condition means)

lib/api/multiObjectDelete.js
```diff
@@ -233,6 +233,12 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
+                        // we need this information for data deletes to AWS
+                        // eslint-disable-next-line no-param-reassign
+                        objMD.location[0].deleteVersion = true;
```
francoisferrand, line 240: nit: probably for later, but it seems strange to store this flag only in the first location, it hints that the flag should not be in the 'location' array but separate from it, somehow... → maybe some refactoring to think of in the future.

francoisferrand (approved):

naren-rajendran (approved):

lib/api/multiObjectDelete.js
miniscruff: Small addition to that, undefined is not an array so the third check handles the second one already.

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -557,3 +583,83 @@ describeSkipIfNotMultiple('AWS backend delete object w. versioning: ' +
...
+    'using object location constraint', function testSuite() {
+    this.timeout(120000);
+    withV4(sigCfg => {
+        let bucketUtil;
+        let s3;
+        beforeEach(() => {
+            process.stdout.write('Creating bucket\n');
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
+            return s3.createBucket({ Bucket: bucket }).promise()
```
miniscruff, line 598: you could use async await here btw, quite minor but you probably wouldn't need to catch the error just to throw it back. If you let the test runner catch the error the call stack is already available.

miniscruff (approved):

nicolas2bert:
@bert-e approve

PR #4565, opened 2022-06-03
Remove left-over tiny version id

francoisferrand:
Tiny version id code has been removed, there is only Short version id,
which is handled fully in Arsenal by setting `S3_VERSION_ID_ENCODING_TYPE`
variable.

The `VersionId.encode()` function does not even actually support passing
an encoding type anymore, so the code was of no use anymore.

Issue: CLDSRV-214


ghost (approved):
LGTM

nicolas2bert (approved):

PR #4575, opened 2022-06-08
Use standard release procedure

francoisferrand:
Bump version **when** doing the release, not afterwards, as it is done almost everywhere else:

* It makes the git history harder to read, with a tag on any commit AND a separate commit (anywhere) which bumps the version
* It does not work with “yarn version” : which creates a misleadingly commit named “v8.x.x” in our repos
* It requires creating an extra ticket after the release ; and this ticket does not actually represent anything: it is just a placeholder to “please” Bert-e (whereas a ticket for the release does represent a piece of work we want to do, with relation to other pieces of work)
* It creates the risk this extra ticket/bump is done, and someone creating another bump when doing the next release (i.e. skipping a release)

Issue: CLDSRV-216

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -26,55 +26,47 @@ docker pull registry.scality.com/cloudserver/cloudserver:<tag>
 
 To release a production image:
 
-* Checkout the relevant branch. In this example,
-  we are working on development/8.3, and we want to release version `8.3.0`.
-
-```sh
-git checkout development/8.3
-```
-
-* Tag the branch with the release version. In this example, `8.3.0`
-
-```sh
-git tag -a 8.3.0
-# The message should be 'v<version>'
-v8.3.0
-```
-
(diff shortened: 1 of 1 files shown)


welansari (approved):

docs/RELEASE.md
```diff
@@ -26,55 +26,47 @@ docker pull registry.scality.com/cloudserver/cloudserver:<tag>
...
-# The message should be 'v<version>'
-v8.3.0
-```
+* Create the release on GitHub
+  
+  * Go the Release tab (https://github.com/scality/cloudserver/releases);
+  * Click on the `Draft new release button`;
+  * In the `tag` field, type the name of the release (`8.4.8`), and confirm
+    to create the tag on publish;
+  * Click on `Generate release notes` button to field the fields;
```
welansari: To **fill** the fields ?

ghost (approved):

PR #4588, opened 2022-06-14
CLDSRV-225: Implement delete bucket tagging

ghost:
Cherry-picked from 7.10 #4480.

**Monitoring for the API is added in https://github.com/scality/cloudserver/pull/4594**

### DIFF
--- a/lib/api/bucketDeleteTagging.js
+++ b/lib/api/bucketDeleteTagging.js
@@ -1,8 +1,8 @@
+const { waterfall } = require('async');
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const { metadataValidateBucket } = require('../metadata/metadataUtils');
 const { pushMetric } = require('../utapi/utilities');
 const metadata = require('../metadata/wrapper');
-const util = require('node:util');
 
 /**
  * Bucket Delete Tagging - Delete a bucket's Tagging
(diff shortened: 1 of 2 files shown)


lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Bucket Delete Tagging - Delete a bucket's Tagging
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+async function bucketDeleteTagging(authInfo, request, log, callback) {
```
miniscruff: I don't think we need a callback with async, you can just return the value as we need to await the function anyway.

lib/api/bucketDeleteTagging.js
```diff
@@ -0,0 +1,52 @@
...
+    try {
+        bucket = await metadataValidateBucketPromise(metadataValParams, log);
+        bucket.setTags([]);
+        await updateBucketPromise(bucket.getName(), bucket, log);
+        pushMetric('deleteBucketTagging', log, {
+            authInfo,
+            bucket: bucketName,
+        });
+    } catch (err) {
+        error = err;
```
miniscruff: for async functions I think we can just throw errors

miniscruff (changes requested):

lib/api/bucketDeleteTagging.js
ghost: Thanks and I agree, but removing the callback now will require code [changes in Arsenal](https://github.com/scality/Arsenal/blob/development/8.1/lib/s3routes/routes/routeDELETE.ts#L13), where cloudserver API handlers are called, with a callback... How about creating a ticket to switch the entire project from callbacks to async/await? And, if possible, move out of Arsenal the code related to cloudserver only...

If you agree on the ticket, I will create it right away and add it as a suggestion for our Q3/Q4 wish list.

lib/api/bucketDeleteTagging.js
ghost: Related to my other comment, for now, the function flow is waiting for an error to be returned as part of the callback, if any. If we throw an error here, we will crash the project, is that what we want?

lib/api/bucketDeleteTagging.js
miniscruff: If that is the case I wouldn't use async at all and just use the callback, saving the switch to async/await for the new work. I would also make sure that the "switch entire project" is split up a bit as that would be a massive PR. We also probably need to sync multiple projects and teams together, similar to the node 16 work.

lib/api/bucketDeleteTagging.js
miniscruff: Remove the async prefix on the function if we are not going to use it properly, similar to my other response.

lib/api/bucketDeleteTagging.js
ghost: Thanks, but now I have a question: you suggested these changes in Will's PR [here](https://github.com/scality/cloudserver/pull/4481#discussion_r891495697) on 7.10 branch (that I cherry-picked here, I know it's not the same branch but similar changes where requested in the delete-bucket-tagging branch as well), do you think it's better to revert these changes for now and use callbacks? if yes, I will do that tomorrow :smile:

miniscruff: Yea that was my fault, looking again I don't think we should use async and callbacks. Either use a callback, or async/await. Doing both looks weird and will be confusing and most likely cause problems. We will probably want to create a technical approach doc on switching to async await completely, issues from node 16 rings a bell.

ghost: No problem! Done here: https://github.com/scality/cloudserver/pull/4588/commits/4c6075708660c9fbe54c8b44ff888e8318188707

miniscruff (approved):

ghost (approved):
lgtm

PR #4594, opened 2022-06-15
Feature/cldsrv 224 implement get bucket tagging

ghost:
The two first commits are cherry-picked from 7.10 #4481 (already approved, but feel free to request for change if you see anything wrong).

### DIFF
--- a/lib/api/bucketGetTagging.js
+++ b/lib/api/bucketGetTagging.js
@@ -1,0 +1,108 @@
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const util = require('node:util');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const { checkExpectedBucketOwner } = require('./apiUtils/authorization/bucketOwner');
+const { pushMetric } = require('../utapi/utilities');
+const { errors, s3middleware } = require('arsenal');
+const escapeForXml = s3middleware.escapeForXml;
+
+//  Sample XML response:
+/*
+    <Tagging>
+        <TagSet>
+            <Tag>
+                <Key>string</Key>
+                <Value>string</Value>
+            </Tag>
+            <Tag>
+                <Key>string</Key>
+                <Value>string</Value>
+            </Tag>
+        </TagSet>
+    </Tagging>
+*/
+
+/**
+ * @typedef Tag
+ * @type {object}
+ * @property {string} Value - Value of the tag.
+ * @property {string} Key - Key of the tag.
+ */
+/**
+ * Convert Versioning Configuration object of a bucket into xml format.
+ * @param {array.<Tag>} tags - set of bucket tag
+ * @return {string} - the converted xml string of the versioning configuration
+ */
+function tagsToXml(tags) {
+    const xml = [];
+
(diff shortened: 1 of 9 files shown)


lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * bucketGetVersioning - Return Versioning Configuration for bucket
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to respond to http request
+ *  with either error code or xml response body
+ * @return {undefined}
+ */
+async function bucketGetTagging(authInfo, request, log, callback) {
```
miniscruff: identical to the other PR I commented on but I don't believe async and callbacks should be used together

miniscruff (changes requested):

lib/api/bucketGetTagging.js
ghost: Yes, can you let me know what's your opinion on https://github.com/scality/cloudserver/pull/4588#discussion_r901346622? :slightly_smiling_face:

miniscruff: Let you know in the other PR :smirk:

miniscruff (approved):

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,117 @@
...
+ */
+function tagsToXml(tags) {
+    const xml = [];
+
+    xml.push('<?xml version="1.0" encoding="UTF-8" standalone="yes"?><Tagging><TagSet>');
+
+    tags.forEach(tag => {
+        xml.push('<Tag>');
+        xml.push(`<Key>${escapeForXml(tag.Key)}</Key>`);
+        xml.push(`<Value>${escapeForXml(tag.Value)}</Value>`);
```
ghost, line 45: pretty sure with Ronnie we settled on naming the `escapeForXml` function `tagsToXml`

lib/api/bucketGetTagging.js
```diff
@@ -0,0 +1,117 @@
...
+                'getBucketTagging');
+            return callback(err, corsHeaders);
+        }
+        monitoring.promMetrics(
+            'GET', bucketName, '200', 'getBucketTagging');
+        pushMetric('getBucketTagging', log, {
+            authInfo,
+            bucket: bucketName,
+        });
+        return callback(null, xml, corsHeaders);
```
ghost: could we not make a single `return callback` here? Like in put

ghost: Not sure, in `GET` we have 3 returned values in case of success, only 2 when there is a failure.

Edit: it's always 3 and 7.10 only return 2 in case of error...

ghost: Surely the err will be null or undefined if there is none and it will be the same?

ghost: Done: https://github.com/scality/cloudserver/pull/4594/commits/7e6d5e5629d1412ca851d75dc760d41b720c02fa

ghost (approved):
lgtm

PR #4713, opened 2022-08-08
bugfix: CLDSRV-258 set 'originOp' in MD for tagging/ACL ops

jonathan-gramain:
Set properly the 'originOp' field in object metadata to the
corresponding notification event type when doing tagging or ACL
operations, so that bucket notifications do not take this event for an
object creation (as the original field value was left as a create-type
event).

One commit was cherry-picked from the 8.1 branch for tagging operations.


jonathan-gramain:
ping

anurag4DSB (approved):

tmacro (approved):

PR #4742, opened 2022-08-15
bugfix: CLDSRV-266 prioritize "checkTagConditions" over "isAllowed"

jonathan-gramain:
In the generic API handler callApiMethod(), authorization checks are
done via a call to a local helper checkAuthResults(). When tag
conditions need to be checked, a flag "checkTagConditions" is set by
Vault in the authorization response, however, currently the flag
"isAllowed" is also set to true (but the request is not necessarily
allowed in the end, after tag conditions are checked later on).

In order to avoid requests being allowed by default when tag
conditions are present but cannot be checked yet, we need to
prioritize checking the "checkTagConditions" flag before checking the
"isAllowed" flag, and on the other hand, have Vault return "isAllowed:
false" for such requests. This commit is doing the first part: moving
the authorization check after tag condition checks.

Also: removed an extra argument "updatedContexts" to the callback of
tagConditionKeyAuth() since the function does not return such argument
(the code ignored the argument when it was missing, so it is only a
cleanup).

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -158,35 +158,26 @@ const api = {
         return async.waterfall([
             next => auth.server.doAuth(request, log, (err, userInfo, authorizationResults, streamingV4Params) =>
                 next(err, userInfo, authorizationResults, streamingV4Params), 's3', requestContexts),
-            (userInfo, authorizationResults, streamingV4Params, next) => {
-                if (authorizationResults) {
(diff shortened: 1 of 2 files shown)


watsaqat (approved):

tmacro (approved):

jonathan-gramain:
ping

PR #4751, opened 2022-08-22
bugfix: CLDSRV-269 add policy checks for object lock in put bucket

miniscruff:
Need to check two additional permissions when creating a bucket with object lock enabled.

### DIFF
--- a/lib/api/bucketPut.js
+++ b/lib/api/bucketPut.js
@@ -97,6 +97,49 @@ function _parseXML(request, log, cb) {
     });
 }
 
+function _buildConstantParams({ request, bucketName, authInfo, authParams, ip, locationConstraint, apiMethod }) {
+    return {
+        constantParams: {
+            headers: request.headers,
+            query: request.query,
+            generalResource: bucketName,
+            specificResource: {
+                key: '',
+            },
+            requesterIp: ip,
+            sslEnabled: request.connection.encrypted,
+            awsService: 's3',
+            requesterInfo: authInfo,
+            signatureVersion: authParams.params.data.authType,
+            authType: authParams.params.data.signatureVersion,
+            signatureAge: authParams.params.data.signatureAge,
+            apiMethod,
+            locationConstraint,
+        },
+    };
+}
+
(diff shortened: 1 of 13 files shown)


lib/api/bucketPut.js
```diff
@@ -97,6 +97,49 @@ function _parseXML(request, log, cb) {
     });
 }
 
+function _buildConstantParams({ request, bucketName, authInfo, authParams, ip, locationConstraint, apiMethod }) {
```
ghost, line 100: Maybe this can go to a generic util file in the future? It should be quite similar for other API calls to Vault, although we may have only this case today where multiple permissions must be checked...

miniscruff: Maybe a builder pattern util thing in vault client or arsenal yea, I think we should do a check up on missing permissions as I believe we are missing a quite a few.

miniscruff: I have more of these to do, so I will probably consolidate the functions during those other tasks.

lib/api/bucketPut.js
```diff
@@ -97,6 +97,49 @@ function _parseXML(request, log, cb) {
...
+        },
+    };
+}
+
+function _handleAuthResults(locationConstraint, log, next) {
+    return (err, authorizationResults) => {
+        if (err) {
+            return next(err);
+        }
+        if (authorizationResults[0].isAllowed !== true) {
```
ghost: I would feel more confident if we can unit test this, or add a check to ensure `authorizationResults` is an array, to avoid errors.

Also, if we have more than one result, shouldn't we iterate over them? If we have one of the permissions not set, we should deny the overall action, right?

ghost (commented):
Just some quick comments, but the changes LGTM so far

lib/api/bucketPut.js
miniscruff: Check policies in vault has always returned an array, but I swapped out the [0] with a .every and added a few unit tests.

ghost (approved):

alexanderchan-scality (approved):

PR #4757, opened 2022-08-24
bugfix: CLDSRV-270 do not check permissions if bypass is not required

miniscruff:


ghost (approved):
LGTM, it completes CLDSRV-193 :smile:

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -158,6 +158,33 @@ function isObjectLocked(bucket, objectMD, headers) {
     return false;
 }
 
+
+/* objectLockRequiresBypass will return true if the retention info change
+ * would require a bypass governance flag to be true.
+ * In order for this to be true the action must be valid as well, so going from
```
watsaqat, line 164: should action validity be in its own function? Or this function renamed to reflect its checking action validity in addition to the bypass governance flag?

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -158,6 +158,33 @@ function isObjectLocked(bucket, objectMD, headers) {
...
+ * COMPLIANCE to GOVERNANCE would return false unless it expired.
```
watsaqat, line 165: would going from COMPLIANCE to GOVERNANCE be the only case where action is invalid or is this written as an example?

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -158,6 +158,33 @@ function isObjectLocked(bucket, objectMD, headers) {
...
+    const isExtended = new Date(retentionInfo.date) > existingDate;
+
+    if (existingMode === 'GOVERNANCE' && !isExpired) {
+        if (retentionInfo.mode === 'GOVERNANCE' && isExtended) {
+            return false;
+        }
+        return true;
+    }
+
+    // an invalid object lock state
```
watsaqat: Can this comment expand on what are the invalid object lock states?

lib/api/apiUtils/object/objectLockHelpers.js
miniscruff: action validity is part of the `validateObjectLockUpdate` method, this method just checks the specific edge cases of whether or not a bypass is required valid or not. As object lock bypass is only for a specific set of action and there states it is a bit easier to check those and have the other function deal with validity. We just return false an any non-object lock bypass required or not state and leave it up to the validate method later on in the chain.

lib/api/apiUtils/object/objectLockHelpers.js
miniscruff: This is an example, there are a few before and after states that are either invalid or just not a concern for the bypass.

lib/api/apiUtils/object/objectLockHelpers.js
miniscruff: I think more appropriately the comment should say:
```suggestion
    // an invalid retention change or unrelated to bypass
```
There are a few state changes that are invalid which would be caught by the validate later, or are irrelevant to the bypass we don't care about. One example above is going from GOVERNANCE to COMPLIANCE which does not require a bypass. Moving from COMPLIANCE to GOVERNANCE is not allowed, I can't quite remember what other before/after states are invalid.

watsaqat (approved):

PR #4767, opened 2022-08-26
CLDSRV-274: release 7.10.10

KillianG:


KillianG:
ping

jbertran (approved):

anurag4DSB (approved):
would you need a bump in cloudserver 7.70 as well as you need to release that as well?
It might be worth getting the commit in this PR. 
But if you have another PR 7.70, thats good as well.
In federation, we will need a bump in 7.10 and 9.1.

PR #4781, opened 2022-08-31
utapi-v1 delete inconsistency with versioning suspended

watsaqat:
### Motivation and context
Addresses https://scality.atlassian.net/browse/S3C-5741.
Passes in a byteLength when versioning is suspended and the object exists. This allows Utapi to properly decrement the object count and storage utilized for various metrics.

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/api/objectDelete.js
+++ b/lib/api/objectDelete.js
@@ -115,6 +115,7 @@ function objectDelete(authInfo, request, log, cb) {
             const deleteInfo = {
                 removeDeleteMarker: false,
                 newDeleteMarker: false,
+                isNull: delOptions.isNull,
             };
             if (delOptions && delOptions.deleteData) {
                 if (objectMD.isDeleteMarker) {
@@ -130,8 +131,9 @@ function objectDelete(authInfo, request, log, cb) {
             deleteInfo.newDeleteMarker = true;
             return createAndStoreObject(bucketName, bucketMD,
                 objectKey, objectMD, authInfo, canonicalID, null, request,
-                deleteInfo.newDeleteMarker, null, log, (err, newDelMarkerRes) =>
-                next(err, bucketMD, objectMD, newDelMarkerRes, deleteInfo));
+                deleteInfo.newDeleteMarker, null, log, (err, newDelMarkerRes) => {
+                    next(err, bucketMD, objectMD, newDelMarkerRes, deleteInfo);
+                });
         },
     ], (err, bucketMD, objectMD, result, deleteInfo) => {
         const resHeaders = collectCorsHeaders(request.headers.origin,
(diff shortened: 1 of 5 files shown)


alexanderchan-scality (approved):

lib/api/objectDelete.js
```diff
@@ -169,8 +170,14 @@ function objectDelete(authInfo, request, log, cb) {
                     result.versionId : versionIdUtils.encode(
                         result.versionId, config.versionIdEncodingType);
             }
+            const versioningSuspended = bucketMD
+                && bucketMD._versioningConfiguration
+                && bucketMD._versioningConfiguration.status === 'Suspended';
```
miniscruff: Cloudserver should support `?.` null chaining I believe
```suggestion
            const versioningSuspended = bucketMD?._versioningConfiguration?.status === 'Suspended';
```

lib/api/objectDelete.js
```diff
@@ -169,8 +170,14 @@ function objectDelete(authInfo, request, log, cb) {
...
+            const hasByteLength = objectMD && objectMD['content-length'];
```
miniscruff, line 177:
```suggestion
            const hasByteLength = objectMD?.['content-length'];
```
I believe this is the syntax for null chaining by index, could be wrong

miniscruff (commented):
Feel free to ignore if chaining is not supported here, for whatever reason

lib/api/objectDelete.js
```diff
@@ -169,8 +170,14 @@ function objectDelete(authInfo, request, log, cb) {
...
+            const byteLength = versioningSuspended && hasByteLength ? objectMD['content-length'] : null;
```
jonathan-gramain: At https://github.com/scality/cloudserver/pull/4781/files#diff-3e46aeb1940c95fec15aef92052edc1bdd777d5f297b41e53ef0c5c3b456b5caR195 the `content-length` value is parsed to an integer, there might be cases where it's stored as a string likely? In doubt I'd rather do the same.

lib/api/objectDelete.js
```diff
@@ -169,8 +170,14 @@ function objectDelete(authInfo, request, log, cb) {
...
             pushMetric('putDeleteMarkerObject', log, {
                 authInfo,
+                byteLength,
```
jonathan-gramain: Is `byteLength` necessarily from the null version going to be deleted? It does not seem so to me unless I'm mistaken, see what `preprocessingVersioningDelete` is doing to derive info about the null version to delete. I think it may not be what is in `objectMd`.

tmacro (approved):

watsaqat:
ping

watsaqat:
ping

PR #4795, opened 2022-09-14
feature: CLDSRV-278 add github action to release image for fed

miniscruff:


images/svc-base/Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM registry.scality.com/federation/nodesvc-base:7.10.6.0
```
Emylou-s, line 1: If you fix the version like this, you'll need to be sure you update it regularly otherwise you won't benefit from CVES update for instance.

images/svc-base/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+    && apt-get install -y --no-install-recommends \
+    jq \
+    python \
+    git \
+    build-essential \
+    ssh \
+    ca-certificates \
+    wget \
+    libffi-dev \
+    zlib1g-dev \
```
Emylou-s, line 21:
```suggestion
    zlib1g-dev \
    && apt-get clean \
```

images/svc-base/Dockerfile
Emylou-s: Maybe you could have as input federation-version and use this as an arg to the dockerfile.
Similar to what we do for leef : https://github.com/scality/Federation/blob/development/9.1/Dockerfiles/leef/Dockerfile 

And call it in a worfklow here : 
https://github.com/scality/Federation/blob/development/9.1/.github/workflows/release-all-images.yml#L50-L61

Emylou-s (changes requested):

.github/workflows/release-federation.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: images/svc-base/Dockerfile
+          push: true
+          tags: "${{ env.REGISTRY_NAME }}/${{env.PROJECT_NAME }}/${{ env.PROJECT_NAME }}:${{ github.event.inputs.tag }}-svc-base"
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
Emylou-s: You could save most of this by using shared worfklow :
https://github.com/scality/workflows/blob/main/.github/workflows/docker-build.yaml

Then it would look something like : 

```
  steps: 
    uses: scality/workflows/.github/workflows/docker-build.yaml@v1
    secrets: inherit
    with:
      push: true
      namespace: ${{ env.PROJECT_NAME }}
      name: ${{ env.PROJECT_NAME }}
      tag: "${{ github.event.inputs.tag }}-svc-base"
      context: images/svc-base/Dockerfile
      registry: ${{ env.REGISTRY_NAME }}
```

images/svc-base/Dockerfile
miniscruff: Would it be possible to create a `nodesvc-base:7.10` that is updated with each 7.10 release and we can use that? I don't think we can/should rely on either passing in a federation version or expecting the federation version to already exist. It works in leef because its inside federation so the VERSION value is always going to match.

rachedbenmustapha: Floating tags generally are to be avoided for traceability and reproducibility reasons.

miniscruff: Yea yea I know... Just seemed slightly easier then relying on a federation version and constantly updating a full version tag without automation... 😦

rachedbenmustapha: There definitely needs to be some automation.

.github/workflows/release-federation.yaml
miniscruff: This was not possible to use as the `file` input was not being used in that workflow, it is now so switching over.

images/svc-base/Dockerfile
miniscruff: Added to both federation and default docker files

images/svc-base/Dockerfile
tcarmet: @Emylou-s have this subject on the view, where maybe (still a work/design in progress) svc-base will be removed from federation. From this point it will have it's own tag which will move at it's own pace.

(Emylou feel free to correct me if mistaken)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  build-federation-image:
+    uses: scality/workflows/.github/workflows/docker-build.yaml@v1
+    secrets: inherit
+    with:
+      push: true
+      registry: ${{ env.REGISTRY_NAME }}
+      namespace: ${{env.PROJECT_NAME}}
+      name: ${{env.PROJECT_NAME}}
```
tcarmet: Argh, can't believe I'm being this person 😂 
```suggestion
      namespace: ${{env.PROJECT_NAMEv}}
      name: ${{env.PROJECT_NAME }}
```

tcarmet (approved):
I don't know if you think we should add a workflow that just build this image before merging, to ensure we're not breaking it. (I'll let you be the judge of that).

LGTM

.github/workflows/release.yaml
miniscruff: Oh missing spaces yea
```suggestion
      namespace: ${{ env.PROJECT_NAME }}
      name: ${{ env.PROJECT_NAME }}
```
You mean this?

tcarmet: Ah, I missed the ones at the beginning 😂 

Yes!

tcarmet: and I'm noticing I included a `v` ?? 🤦 My bad

tcarmet (approved):

images/svc-base/Dockerfile
Emylou-s: Indeed, that's a another why we should be doing it.
I'll add it to the pile of good arguments

https://scality.atlassian.net/browse/S3C-6432

Emylou-s (approved):

PR #4803, opened 2022-09-20
feature: CLDSRV-278 can not use env vars in shared workflow

miniscruff:
Fixes an issue with GitHub workflows, see https://github.com/scality/workflows#known-issues

Also bump version which I will release after.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,8 +12,6 @@ jobs:
   build-federation-image:
     uses: scality/workflows/.github/workflows/docker-build.yaml@v1
     secrets: inherit
-    needs:
-      - build-vars
     with:
       push: true
       registry: registry.scality.com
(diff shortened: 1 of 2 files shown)


tcarmet (approved):

rdebay-scality (approved):

naren-rajendran (approved):

.github/workflows/release.yaml
```diff
@@ -8,19 +8,17 @@ on:
...
 
-env:
-  REGISTRY_NAME: registry.scality.com
-  PROJECT_NAME: ${{ github.event.repository.name }}
-
 jobs:
   build-federation-image:
     uses: scality/workflows/.github/workflows/docker-build.yaml@v1
     secrets: inherit
+    needs:
```
naren-rajendran: Do we still need this?

miniscruff: oh no that should be removed

PR #4805, opened 2022-09-21
Improvement/cldsrv 281 complete missing permission checks

ghost:
This PR includes changes from Arsenal where we fixed the permission names to comply with AWS standard.

This PR is about adding all the missing permissions according to the work done in https://scality.atlassian.net/wiki/spaces/OS/pages/1984397327/Cloudserver+vs+AWS+permission+checks, with more than 10 APIs affected by missing or wrong permission checks,e specially for Delete APIs or common APIs where multiple headers are accepted.

### DIFF
--- a/lib/api/bucketPut.js
+++ b/lib/api/bucketPut.js
@@ -145,6 +145,10 @@ function _isObjectLockEnabled(headers) {
     return header !== undefined && header.toLowerCase() === 'true';
 }
 
+function _isAclProvided(headers) {
+    return aclUtils.getHeaders().some(h => Object.keys(headers).includes(h));
+}
+
 function authBucketPut(authParams, bucketName, locationConstraint, request, authInfo) {
     const ip = requestUtils.getClientIp(request, config);
     const baseParams = {
@@ -172,6 +176,15 @@ function authBucketPut(authParams, bucketName, locationConstraint, request, auth
             { apiMethod: 'bucketPutVersioning' },
         ));
     }
+
+    if (_isAclProvided(request.headers)) {
+        requestConstantParams.push(Object.assign(
+            {},
+            baseParams,
+            { apiMethod: 'bucketPutACL' },
+        ));
+    }
+
     return requestConstantParams;
 }
 
--- a/tests/unit/policies.js
+++ b/tests/unit/policies.js
@@ -1,6 +1,287 @@
 const assert = require('assert');
 const DummyRequest = require('./DummyRequest');
 const { authBucketPut } = require('../../lib/api/bucketPut');
+const prepareRequestContexts = require('../../lib/api/apiUtils/authorization/prepareRequestContexts.js');
+
+const sourceBucket = 'bucket';
+const sourceObject = 'object';
+const apiMatrix = [
+    {
+        name: 'multipartDelete',
+        expectedPermissions: ['s3:AbortMultipartUpload'],
+    },
+    {
+        name: 'objectCopy',
+        headers: {
+            'x-amz-tagging': true,
+            'x-amz-tagging-directive': 'REPLACE',
+        },
+        expectedPermissions: ['s3:GetObject', 's3:PutObject', 's3:PutObjectTagging'],
+    },
+    {
+        name: 'objectCopy',
+        headers: {
+            'x-amz-tagging': true,
+            'x-amz-tagging-directive': 'COPY',
+        },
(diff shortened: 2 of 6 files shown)


.eslintrc
```diff
@@ -1 +1,10 @@
-{ "extends": "scality" }
+{
```
miniscruff: I wouldn't change eslint rules for the PR, not sure why JSX would need to be true either way.

lib/api/bucketPut.js
```diff
@@ -174,60 +218,37 @@ function bucketPut(authInfo, request, log, callback) {
...
-
-            if (!_isObjectLockEnabled(request.headers)) {
-                requestConstantParams.push(Object.assign(
-                    baseParams,
-                    { apiMethod: 'bucketPutObjectLock' },
-                ));
-                requestConstantParams.push(Object.assign(
-                    baseParams,
-                    { apiMethod: 'bucketPutVersioning' },
-                ));
```
miniscruff: I think this will fix it.
```js
requestConstantParams.push(Object.assign(
    {},
    baseParams,
    { apiMethod: 'bucketPutObjectLock' },
));
```

.eslintrc
ghost: Yes thanks, with your changes it's all good! I am also based on your PR now.

tests/unit/policies.js
```diff
@@ -35,5 +381,17 @@ describe('Policies: permission checks for S3 APIs', () => {
                 ['bucketPut', 'bucketPutObjectLock', 'bucketPutVersioning'],
             );
         });
+
+        it('PutBucket: should return s3:PutBucket and s3:PutBucketObjectLockConfiguration' +
```
miniscruff: I am so sorry... your test here is the same as the one I made above but that my test name is completely wrong...

tests/unit/policies.js
```diff
@@ -1,6 +1,338 @@
...
 const DummyRequest = require('./DummyRequest');
 const { authBucketPut } = require('../../lib/api/bucketPut');
+const prepareRequestContexts = require('../../lib/api/apiUtils/authorization/prepareRequestContexts.js');
+
+const sourceBucket = 'bucket';
+const sourceObject = 'object';
+const apiMatrix = [
+    {
+        name: 'multipartDelete',
+        headers: {},
```
miniscruff: I believe the headers parameter is optional, so I would remove them.

tests/unit/policies.js
```diff
@@ -15,6 +347,20 @@ function prepareDummyRequest(headers = {}) {
...
+        if (api.name.length === 0) return;
+        const message = `should return ${api.expectedPermissions.join(', ')} in requestContextParams for ${api.name}` +
+            `${api.headers.length > 0 ? ` with headers ${api.headers.map(el => el[0]).join(', ')}` : ''}`;
+        it(message, () => {
+            const request = prepareDummyRequest(api.headers);
+            const requestContexts = prepareRequestContexts(api.name, request,
+                sourceBucket, sourceObject);
+            const listOfrequestedActions = requestContexts.map(rq => rq.getAction());
+            assert.strictEqual(listOfrequestedActions.length, api.expectedPermissions.length);
+            assert(api.expectedPermissions.every(value => listOfrequestedActions.includes(value)));
```
miniscruff:
```suggestion
            const requestedActions = requestContexts.map(rq => rq.getAction());
            assert.deepStrictEqual(requestedActions, api.expectedPermissions);
```

A deep strict equal will work here better ( in my opinion anyway ). This is cause using an every+includes will not catch if the list has more permissions then our expected. ( [a, b, c].every+includes[a, b, c, d] will be true ). This is why you also needed the length check. But because our permissions are in a deterministic order we can just skip these two checks and do a deep strict equal. Plus its less code. :smirk:

PS: requested actions is plural so we can skip `list of`

miniscruff (changes requested):

tests/unit/policies.js
ghost: Oh no problem, I will update that!

ghost: Should be done in the new commits (sorry I had to rebase after releasing Arsenal and seeing recent bumps in cloudserver). I also took care of your other comments.

ghost (approved):
lgtm

miniscruff (approved):

PR #4824, opened 2022-10-05
bf: CLDSRV-285 update dockerfile to match federation

naren-rajendran:
update dockerfile to match federation.


Emylou-s (approved):
The Dockerfile looks indeed much more similar to Federation's one.

nicolas2bert (approved):

PR #4842, opened 2022-10-24
bf(CLDSRV-289): Fix putDeleteMarkerObject metric for version suspende…

tmacro:
For versioning suspended buckets the storageUtilized should be decremented the size of the obj, with numberOfObjects remaining unchanged, Since the data is deleted, but a DeleteMarker is still present

### DIFF
--- a/lib/utapi/utilities.js
+++ b/lib/utapi/utilities.js
@@ -280,11 +280,13 @@ function pushMetric(action, log, metricObj) {
             sizeDelta = -oldByteLength;
         } else if (action === 'abortMultipartUpload' && byteLength) {
             sizeDelta = -byteLength;
+        } else if (action === 'putDeleteMarkerObject' && byteLength) {
+            sizeDelta = -byteLength;
         }
 
         let objectDelta = isDelete ? -numberOfObjects : numberOfObjects;
         // putDeleteMarkerObject does not pass numberOfObjects
-        if (action === 'putDeleteMarkerObject'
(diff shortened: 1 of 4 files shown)


lib/utapi/utilities.js
```diff
@@ -280,11 +281,13 @@ function pushMetric(action, log, metricObj) {
...
         } else if (action === 'abortMultipartUpload' && byteLength) {
             sizeDelta = -byteLength;
+        } else if (action === 'putDeleteMarkerObject' && versioningSuspended && byteLength) {
+            sizeDelta = -byteLength;
         }
 
         let objectDelta = isDelete ? -numberOfObjects : numberOfObjects;
         // putDeleteMarkerObject does not pass numberOfObjects
-        if (action === 'putDeleteMarkerObject'
+        if ((action === 'putDeleteMarkerObject' && !versioningSuspended)
```
jonathan-gramain: Shouldn't we also check `byteLength`? If it is used to know whether we are removing an object or not in the versioning-suspended case (in the sense that if we are not removing anything, we should still increment `objectDelta`)

tmacro: I believe you're right. If I bring `byteLength` into the picture I think I can drop the check of `versioningSuspended`, and just use `byteLength === null` to cover all the cases.

jonathan-gramain (approved):

watsaqat (approved):

watsaqat (approved):

jonathan-gramain (approved):

PR #4878, opened 2022-11-15
Feature: CLDSRV-304 support object restore completed notification

welansari:
Issue: [CLDSRV-304](https://scality.atlassian.net/browse/CLDSRV-304)

`overwritingVersioning()` is executed at the end of each restore operation, we set the originOp to `s3:ObjectRestore:Completed` at that level

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -405,6 +405,9 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     metadataStoreParams.lastModifiedDate = objMD['last-modified'];
     metadataStoreParams.updateMicroVersionId = true;
 
+    // set correct originOp
+    metadataStoreParams.originOp = 's3:ObjectRestore:Completed';
+
     // update restore
     const days = objMD.archive?.restoreRequestedDays;
     const now = Date.now();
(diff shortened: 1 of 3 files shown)


ghost (approved):
LGTM

lib/api/apiUtils/object/versioning.js
```diff
@@ -405,6 +405,9 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     metadataStoreParams.lastModifiedDate = objMD['last-modified'];
     metadataStoreParams.updateMicroVersionId = true;
 
+    // set correct originOp
+    metadataStoreParams.originOp = 's3:ObjectRestore:Completed';
```
ghost, line 409: Possible improvement: using consts for the string, from arsenal or an internal file.

ghost (approved):

lib/api/apiUtils/object/versioning.js
welansari: In all of Cloudserver we set the originOp manually, but i do think ideally we need to use a map of some sort.

francoisferrand: an enum from typescript would be most appropriate........ once we move to typescript :-/
since the list is in Arsenal, maybe we can define a typescript enum there, and reuse the "generated" variables here (if they exists).

Anyway i agree there is no rush to do it now, it should be a "cleanup" task to do separately and everywhere in cloudserver → @Kerkesni can you create a ticket with the TechDebt label?

welansari: Done [ZENKO-4409](https://scality.atlassian.net/browse/ZENKO-4409)

PR #4895, opened 2022-11-29
Bugfix/cldsrv 310/fix hotfix cherry pick 7.10.8

alexanderchan-scality:


rahulreddy (approved):

package.json
```diff
@@ -37,7 +37,7 @@
     "utapi": "git+https://github.com/scality/utapi#7.10.7",
     "utf8": "~2.1.1",
     "uuid": "^3.0.1",
-    "vaultclient": "git+https://github.com/scality/vaultclient#7.10.8-1",
+    "vaultclient": "scality/vaultclient#7.10.8",
```
jonathan-gramain, line 40: Technically we should always bump the version right? so bumping to `7.10.8-2` in this case, even if we have to revert some code.

naren-rajendran (approved):

PR #4929, opened 2022-12-20
bf(CLDSRV-321): Fix retention extension check to consider same date as extended

tmacro:
Fixes the check when trying to extend object lock's retention period to consider the same date as extended and allow the request to succeed.

### DIFF
--- a/tests/unit/api/objectPutRetention.js
+++ b/tests/unit/api/objectPutRetention.js
@@ -16,8 +16,8 @@ const bucketName = 'bucketname';
 const objectName = 'objectName';
 const postBody = Buffer.from('I am a body', 'utf8');
 
-const date = new Date();
-date.setDate(date.getDate() + 1);
+const expectedMode = 'GOVERNANCE';
+const expectedDate = moment().add(2, 'days').toISOString();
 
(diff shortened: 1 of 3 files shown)


jonathan-gramain (approved):

tests/unit/api/objectPutRetention.js
```diff
@@ -95,8 +101,15 @@ const putObjRetRequestGovernanceLonger = {
...
+const putObjRetRequestGovernanceShorter = {
+    bucketName,
+    objectKey: objectName,
+    headers: { host: `${bucketName}.s3.amazonaws.com` },
+    post: objectRetentionXmlGovernanceShorter,
+};
+
 const expectedMode = 'GOVERNANCE';
-const expectedDate = date.toISOString();
+const expectedDate = moment().add(2, 'days').toISOString();
```
rachedbenmustapha: isn't there a theoretical flakiness risk if there's millisec skew (or overlap) while running the test, making the dates different just because the 2 values compared are evaluated a 2 different call sites? Not a big deal, but could be painful to track down, idk.

rachedbenmustapha (approved):

PR #4942, opened 2022-12-22
Feature/cldsrv 322 introduce sosapi internal routes

ghost:
Context Veeam12 SOSAPI support. See the commit descriptions for more information.

This PR introduces a new set of APIs exposed as internal routes for veeam. The ZKOP PR handling redirection is open here: https://github.com/scality/zenko-operator/pull/355.

The high level approach of this pr is:

* Create a new router that will manbage the HTTP verbs, authentication + authorization, and then call the right API based on the input parameters (or return an error).
* A utils file to handle common operations like streaming data, or getting streamed data from the client.
* One file per API: PUT, GET, DELETE, HEAD and LIST, with support for Veeam12 files, to be stored in the bucket metadata. * The associated bucket metadata changes are done in https://github.com/scality/Arsenal/pull/2059.
* XML schema validation for Veeam12 files. If the protocol version is known, the schema is validated, otherwise, any valid XML file is accepted, with a 1MB size limit.
* This implementation covers UI, API and CLI usage.

Please let me know if you have any question on the context.

This was tested end2end in a real Artesca with the ZKOP changes:

## This was tested end2end along with Data Metrics and ZKOP custom Ingresses with success:

image

image

image

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -23,6 +23,8 @@ const { isRequesterNonAccountUser } = require('./apiUtils/authorization/permissi
 const { hasGovernanceBypassHeader, checkUserGovernanceBypass, ObjectLockInfo }
     = require('./apiUtils/object/objectLockHelpers');
 const requestUtils = policies.requestUtils;
+const { validObjectKeys } = require('../routes/routeVeeam');
+const { deleteVeeamCapabilities } = require('../routes/veeam/delete');
 
 const versionIdUtils = versioning.VersionID;
 
@@ -382,6 +384,7 @@ function multiObjectDelete(authInfo, request, log, callback) {
         return callback(errors.BadDigest);
     }
 
+    const inPlayInternal = [];
     const bucketName = request.bucketName;
     const canonicalID = authInfo.getCanonicalID();
 
@@ -481,6 +484,9 @@ function multiObjectDelete(authInfo, request, log, callback) {
                             versionId: result.versionId,
                         };
                         if (result.isAllowed) {
+                            if (validObjectKeys.includes(entry.key)) {
+                                inPlayInternal.push(entry.key);
+                            }
                             inPlay.push(entry);
                         } else {
                             errorResults.push({
@@ -532,6 +538,11 @@ function multiObjectDelete(authInfo, request, log, callback) {
                 }
                 return next(null, quietSetting, errorResults, inPlay, bucketMD);
             });
+        },
+        function handleInternalFiles(quietSetting, errorResults, inPlay, bucketMD, next) {
+            return async.each(inPlayInternal,
+                (localInPlay, next) => deleteVeeamCapabilities(bucketName, localInPlay, bucketMD, log, next),
+                err => next(err, quietSetting, errorResults, inPlay, bucketMD));
         },
         function getObjMetadataAndDeleteStep(quietSetting, errorResults, inPlay,
             bucket, next) {
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -486,8 +486,9 @@ function multiObjectDelete(authInfo, request, log, callback) {
                         if (result.isAllowed) {
                             if (validObjectKeys.includes(entry.key)) {
                                 inPlayInternal.push(entry.key);
+                            } else {
+                                inPlay.push(entry);
                             }
-                            inPlay.push(entry);
                         } else {
                             errorResults.push({
                                 entry,
--- a/lib/routes/routeVeeam.js
+++ b/lib/routes/routeVeeam.js
@@ -1,6 +1,7 @@
 const url = require('url');
 const async = require('async');
 const vault = require('../auth/vault');
+const putVeeamFile = require('./veeam/put');
 const getVeeamFile = require('./veeam/get');
 const { auth, s3routes, errors } = require('arsenal');
 const { _decodeURI, validPath } = require('./veeam/utils');
@@ -27,7 +28,7 @@ const apiToAction = {
 
 const routeMap = {
     GET: getVeeamFile,
-    // PUT: putVeeamFile,
+    PUT: putVeeamFile,
     // HEAD: headVeeamFile,
     // DELETE: deleteVeeamFile,
     // LIST: listVeeamFiles,
--- a/lib/routes/routeVeeam.js
+++ b/lib/routes/routeVeeam.js
@@ -3,6 +3,7 @@ const async = require('async');
 const vault = require('../auth/vault');
 const putVeeamFile = require('./veeam/put');
 const getVeeamFile = require('./veeam/get');
+const headVeeamFile = require('./veeam/head');
 const { auth, s3routes, errors } = require('arsenal');
 const { _decodeURI, validPath } = require('./veeam/utils');
 const { routesUtils } = require('arsenal/build/lib/s3routes');
@@ -29,7 +30,7 @@ const apiToAction = {
 const routeMap = {
     GET: getVeeamFile,
     PUT: putVeeamFile,
-    // HEAD: headVeeamFile,
+    HEAD: headVeeamFile,
     // DELETE: deleteVeeamFile,
     // LIST: listVeeamFiles,
 };
--- a/lib/routes/routeVeeam.js
+++ b/lib/routes/routeVeeam.js
@@ -1,6 +1,7 @@
 const url = require('url');
 const async = require('async');
 const vault = require('../auth/vault');
+const getVeeamFile = require('./veeam/get');
 const { auth, s3routes, errors } = require('arsenal');
 const { _decodeURI, validPath } = require('./veeam/utils');
 const { routesUtils } = require('arsenal/build/lib/s3routes');
@@ -25,7 +26,7 @@ const apiToAction = {
 };
 
 const routeMap = {
-    // GET: getVeeamFile,
+    GET: getVeeamFile,
     // PUT: putVeeamFile,
     // HEAD: headVeeamFile,
     // DELETE: deleteVeeamFile,
--- a/lib/routes/routeVeeam.js
+++ b/lib/routes/routeVeeam.js
@@ -48,6 +48,9 @@ const routeMap = {
  * @returns {Error | undefined} error or undefined
  */
 function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, log) {
+    // In case bucket name is not specified and the request contains an
+    // object key or is not a GET, then the bucket name is mandatory.
+    // Reject the request in this case.
     if (!bucketName && !(method === 'GET' && !objectKey)) {
         log.debug('empty bucket name', { method: 'checkBucketAndKey' });
         return errors.MethodNotAllowed;
@@ -67,7 +70,7 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
         if (requestQueryParams && Object.keys(requestQueryParams).length > 0
             && !(method === 'GET' && (requestQueryParams['X-Amz-Credential'] || ('tagging' in requestQueryParams)))) {
             return errors.InvalidRequest
-                .customizeDescription('The Veeam folder does not support this action.');
+                .customizeDescription('The Veeam SOSAPI folder does not support this action.');
         }
         if (typeof objectKey !== 'string' || !validObjectKeys.includes(objectKey)) {
             log.debug('invalid object name', { objectKey });
@@ -90,13 +93,12 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
  */
 function authorizationMiddleware(request, response, api, log, callback) {
     if (!api) {
-        return errors.AccessDenied;
+        return responseXMLBody(errors.AccessDenied, null, response, log);
     }
-    let bucketMd = null;
     const requestContexts = prepareRequestContexts(api, request);
     return async.waterfall([
         next => auth.server.doAuth(request, log, (err, userInfo, authorizationResults, streamingV4Params) => {
-        if (err) {
+            if (err) {
                 log.debug('authentication error', {
                     error: err,
                     method: request.method,
@@ -123,15 +125,10 @@ function authorizationMiddleware(request, response, api, log, callback) {
                 requestType,
                 request,
             };
-            return metadataValidateBucket(mdValParams, log, (err, _bucketMd) => {
-                if (err) {
-                    return next(err);
-                }
-                bucketMd = _bucketMd;
-                return next();
-            });
+            return next(null, mdValParams);
         },
-    ], err => {
+        (mdValParams, next) => metadataValidateBucket(mdValParams, log, next),
+    ], (err, bucketMd) => {
         if (err || !bucketMd) {
             return responseXMLBody(err, null, response, log);
         }
@@ -141,7 +138,11 @@ function authorizationMiddleware(request, response, api, log, callback) {
 
 function _normalizeVeeamRequest(req) {
     /* eslint-disable no-param-reassign */
-    // Rewriting the URL is needed for the V4 signature check
+    // Rewriting the URL is needed for the V4 signature check: the initial
+    // request targets https://s3.subdomain/bucketName/objectKey, but the
+    // custom ingresses and/or nginx configuration for the UI will redirect this
+    // call to .../_/veeam/bucketName/objectKey. We need to revert the custom
+    // path only used for routing before computing the V4 signature.
     req.url = req.url.replace('/_/veeam', '');
     // Assign multiple common (extracted) parameters to the request object
     const parsedUrl = url.parse(req.url, true);
--- a/lib/routes/routeVeeam.js
+++ b/lib/routes/routeVeeam.js
@@ -4,6 +4,7 @@ const vault = require('../auth/vault');
 const putVeeamFile = require('./veeam/put');
 const getVeeamFile = require('./veeam/get');
 const headVeeamFile = require('./veeam/head');
+const { deleteVeeamFile } = require('./veeam/delete');
 const { auth, s3routes, errors } = require('arsenal');
 const { _decodeURI, validPath } = require('./veeam/utils');
 const { routesUtils } = require('arsenal/build/lib/s3routes');
@@ -31,7 +32,7 @@ const routeMap = {
     GET: getVeeamFile,
     PUT: putVeeamFile,
     HEAD: headVeeamFile,
-    // DELETE: deleteVeeamFile,
+    DELETE: deleteVeeamFile,
     // LIST: listVeeamFiles,
 };
 
--- a/lib/routes/routeVeeam.js
+++ b/lib/routes/routeVeeam.js
@@ -1,0 +1,219 @@
+const url = require('url');
+const async = require('async');
+const vault = require('../auth/vault');
+const { auth, s3routes, errors } = require('arsenal');
+const { _decodeURI, validPath } = require('./veeam/utils');
+const { routesUtils } = require('arsenal/build/lib/s3routes');
+const { metadataValidateBucket } = require('../metadata/metadataUtils');
+const prepareRequestContexts = require('../api/apiUtils/authorization/prepareRequestContexts');
+
+const { responseXMLBody } = s3routes.routesUtils;
+
+auth.setHandler(vault);
+
+const validObjectKeys = [
+    `${validPath}system.xml`,
+    `${validPath}capacity.xml`,
+];
+
+const apiToAction = {
+    PUT: 'PutObject',
+    GET: 'GetObject',
+    HEAD: 'HeadObject',
+    DELETE: 'DeleteObject',
+    LIST: 'ListObjects',
+};
+
+const routeMap = {
+    // GET: getVeeamFile,
+    // PUT: putVeeamFile,
+    // HEAD: headVeeamFile,
+    // DELETE: deleteVeeamFile,
+    // LIST: listVeeamFiles,
+};
+
+/**
+ * Validator for the Veeam12 custom routes. Ensures that bucket name and
+ * object name are correct, and that the bucket exists in the DB.
+ * @param {string} bucketName  - name of the bucket
+ * @param {string} objectKey - key of the object
+ * @param {array | null} requestQueryParams - request query parameters
+ * @param {string} method - HTTP verb
+ * @param {object} log - request logger
+ * @returns {Error | undefined} error or undefined
+ */
+function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, log) {
+    if (!bucketName && !(method === 'GET' && !objectKey)) {
+        log.debug('empty bucket name', { method: 'checkBucketAndKey' });
+        return errors.MethodNotAllowed;
+    }
+    if (typeof bucketName !== 'string' || routesUtils.isValidBucketName(bucketName, []) === false) {
+        log.debug('invalid bucket name', { bucketName });
+        if (method === 'DELETE') {
+            return errors.NoSuchBucket;
+        }
+        return errors.InvalidBucketName;
+    }
+    if (method !== 'LIST') {
+        // Reject any unsupported request, but allow downloads and deletes from UI
+        // Download relies on GETs calls with auth in query parameters, that can be
+        // checked if 'X-Amz-Credential' is included.
+        // Deletion requires that the tags of the object are returned.
+        if (requestQueryParams && Object.keys(requestQueryParams).length > 0
+            && !(method === 'GET' && (requestQueryParams['X-Amz-Credential'] || ('tagging' in requestQueryParams)))) {
+            return errors.InvalidRequest
+                .customizeDescription('The Veeam folder does not support this action.');
+        }
+        if (typeof objectKey !== 'string' || !validObjectKeys.includes(objectKey)) {
+            log.debug('invalid object name', { objectKey });
+            return errors.InvalidArgument;
+        }
+    }
+    return undefined;
+}
+
+/**
+ * Query the authorization service for the request, and extract the bucket
+ * and, if applicable, object metadata according to the request method.
+ *
+ * @param {object} request - incoming request
+ * @param {object} response - response object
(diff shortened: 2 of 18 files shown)


anurag4DSB (commented):
Would it be possible to these into smaller logical PRs?
Its currently very long to review.
We can piggyback branches if needed.

package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.12.0",
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.1.77",
+    "arsenal": "git+https://github.com/scality/Arsenal#f446354c526384dc93bd3d2c79c39e50b1a8186c",
```
francoisferrand: please do not forget to update before merging

lib/routes/routeVeeam.js
```diff
@@ -0,0 +1,219 @@
...
+ * object name are correct, and that the bucket exists in the DB.
+ * @param {string} bucketName  - name of the bucket
+ * @param {string} objectKey - key of the object
+ * @param {array | null} requestQueryParams - request query parameters
+ * @param {string} method - HTTP verb
+ * @param {object} log - request logger
+ * @returns {Error | undefined} error or undefined
+ */
+function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, log) {
+    if (!bucketName && !(method === 'GET' && !objectKey)) {
```
francoisferrand, line 54: kind of hard to read (too many negations), maybe consider rewriting in a way that would be easier to read... (e.g. introducing helper function, to "name" the intent; or splitting in multiple `if` statements)

lib/routes/routeVeeam.js
```diff
@@ -0,0 +1,224 @@
...
+            if (!requestType) {
+                return next(errors.AccessDenied);
+            }
+            const mdValParams = {
+                bucketName: request.bucketName,
+                authInfo: userInfo,
+                requestType,
+                request,
+            };
+            return metadataValidateBucket(mdValParams, log, (err, _bucketMd) => {
```
francoisferrand: could use waterfall instead, to avoid nesting & uneeded error checking:

```js
    return metadataValidateBucket(mdValParams, log, next)
},
(_bucketMd, next) => {
    bucketMd = _bucketMd;
    return next();
});
```

or better yet, remove the bucketMd variable and just pass it to the final callack:
```js
return async.waterfall([
    ...
    (userInfo, next) => {
         ...
         return metadataValidateBucket(mdValParams, log, next)
    }
], (err, bucketMd) {
    if (err || !bucketMd) {
        return responseXMLBody(err, null, response, log);
    }
    return callback(request, response, bucketMd, log);
});
```

lib/routes/routeVeeam.js
```diff
@@ -0,0 +1,224 @@
...
+        },
+    ], err => {
+        if (err || !bucketMd) {
+            return responseXMLBody(err, null, response, log);
+        }
+        return callback(request, response, bucketMd, log);
+    });
+}
+
+function _normalizeVeeamRequest(req) {
```
francoisferrand, line 139: do we really _have to_ replace some fields from the req?

Apart from the (not yet relevant) potential type-ability implications it may have, this kind of pattern may leave the object in a partially updated state, and lead to hard to find issues : it is generally safer to create a new normalized object with exactly the normalizedinformation you need...

lib/routes/veeam/get.js
```diff
@@ -0,0 +1,58 @@
...
+            if (!data._capabilities?.VeeamSOSApi?.CapacityInfo) {
+                return responseXMLBody(errors.NoSuchKey, null, response, log);
+            }
+            fileToBuild = {
+                CapacityInfo: data._capabilities?.VeeamSOSApi?.CapacityInfo,
+            };
+            delete fileToBuild.CapacityInfo?.LastModified;
+        }
+        modified = fileToBuild.LastModified;
+        delete fileToBuild.LastModified;
```
francoisferrand: this is never set. fileToBuild only has a single field named either `SystemInfo` or `CapacityInfo`

lib/routes/veeam/head.js
```diff
@@ -0,0 +1,44 @@
...
+    if (!bucketMd) {
+        return responseXMLBody(errors.NoSuchBucket, null, response, log);
+    }
+    return metadata.getBucket(request.bucketName, log, (err, data) => {
+        if (err) {
+            return responseXMLBody(errors.InternalError, null, response, log);
+        }
+        const isSystemXML = request.objectKey.endsWith('/system.xml');
+        const fileToBuild = isSystemXML ? data._capabilities?.VeeamSOSApi?.SystemInfo
+            : isSystemXML && !data._capabilities?.VeeamSOSApi?.CapacityInfo;
```
francoisferrand: the 'else' branch will always evaluate to `false`, since it includes the 'root' condition...

lib/routes/veeam/head.js
```diff
@@ -0,0 +1,44 @@
...
+        if (isSystemXML && !data._capabilities?.VeeamSOSApi?.SystemInfo) {
+            return responseXMLBody(errors.NoSuchKey, null, response, log);
+        } else if (!isSystemXML && !data._capabilities?.VeeamSOSApi?.CapacityInfo) {
+            return responseXMLBody(errors.NoSuchKey, null, response, log);
+        }
```
francoisferrand:
```js
if (!fleToBuild) {
    return responseXMLBody(errors.NoSuchKey, null, response, log);
}
```

lib/routes/veeam/list.js
```diff
@@ -0,0 +1,162 @@
...
+
+function buildContent(arrayOfFiles) {
+    const baseXML = [];
+    arrayOfFiles.forEach(file => {
+        baseXML.push('<Contents>');
+        baseXML.push(`<Key>${file.name}</Key>`);
+        baseXML.push(`<LastModified>${file['Last-Modified']}</LastModified>`);
+        baseXML.push(`<ETag>${file.ETag}</ETag>`);
+        baseXML.push(`<Size>${file['Content-Length']}</Size>`);
+        baseXML.push('<StorageClass>STANDARD</StorageClass>');
```
francoisferrand: is the `storageClass` value required by Veam?
Otherwise, could we (and should we?) use some specific value, to better reflect this is a veam-specific "virtual" object ?

lib/routes/veeam/list.js
```diff
@@ -0,0 +1,162 @@
+const url = require('url');
+const xml2js = require('xml2js');
+const { errors } = require('arsenal');
+const querystring = require('querystring');
+const metadata = require('../../metadata/wrapper');
+const { responseXMLBody } = require('arsenal/build/lib/s3routes/routesUtils');
+const { respondWithData, getResponseHeader, buildHeadXML, validPath } = require('./utils');
+
+function buildContent(arrayOfFiles) {
```
francoisferrand: instead of rewriting these XML-generating functions, I think we should reuse the existing `processVersions()`, `processMasterVersions()`, [...] functions (from `bucketGet.js`): e.g. here we would just build the proper parameter to these functions.

It would be less redundant, and allow updating the xml management stuff in a single place (and ideally these API will be typed, at some point...)

lib/routes/veeam/head.js
```diff
@@ -0,0 +1,44 @@
...
+        const savedModifiedDate = fileToBuild.LastModified;
+        delete fileToBuild.LastModified;
+        // Recompute file content to generate appropriate content-md5 header
+        const builder = new xml2js.Builder({
+            headless: true,
+        });
```
francoisferrand: this whole mapping of `system.xml` to `SystemInfo` and `CapacityInfo` otherwise to generate the XML is duplicate in both Head and Get, as well as list to some extend : should be factorized/de-duplicated (for get & head definitely, not sure how it would go for list...)

lib/routes/veeam/list.js
ghost: This is not required, and you are right, it would make more sense, I'll change it, thanks.

anurag4DSB:
> Would it be possible to these into smaller logical PRs?
> 
> Its currently very long to review.
> 
> We can piggyback branches if needed.

Post discussion with @francoisferrand  one PR is good

lib/routes/routeVeeam.js
ghost: This was coming from Arsenal, see example here: https://github.com/scality/Arsenal/blob/development/8.1/lib/s3routes/routes.ts#L48

I will change the logic to make it more easy to read.

lib/routes/routeVeeam.js
ghost: Yes, the url must be updated to account for the dynamic path change from the ingress, in order to have valid V4 signatures.
This is actually something that we already do in Arsenal and other custom routes as well, I am not sure we want to change the pattern in only one place, do we?

Example from Arsenal: https://github.com/scality/Arsenal/blob/development/8.1/lib/s3routes/routesUtils.ts#L1039

lib/routes/routeVeeam.js
ghost: Is it better this way? https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-d54eeee0f403b38d9cec6ec8ab8caca914402b5705e4bfb96179c5065b8ad8a0R51

lib/routes/routeVeeam.js
ghost: Done here: https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-d54eeee0f403b38d9cec6ec8ab8caca914402b5705e4bfb96179c5065b8ad8a0R130

lib/routes/veeam/get.js
ghost: I created some util functions that now better handle this: https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-a23b06e1deac8b9332f76c3ae6723a32d6a49e7b7845e7f94f76e17190be0eb0R153

lib/routes/veeam/head.js
ghost: Updated in https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-ee4e7989b66eaaa0e7f62d59ec42d4cfa0d5c6a8bf5942a69b07110aea5555d1R24 (it relies on the new util functions, that solves the issue)

lib/routes/veeam/head.js
ghost: Done in https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-a23b06e1deac8b9332f76c3ae6723a32d6a49e7b7845e7f94f76e17190be0eb0R174

lib/routes/veeam/list.js
ghost: I tried with VIRTUAL, let me know what you think: https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-25506d7604624dfc2c3d66729bd5dffc4534b63a3bffd2d7213ffa1dcb794ae5R47

image

lib/routes/veeam/list.js
ghost: I did what you suggested, so I created a function to create the right objects before calling the usual functions from `bucketGet`: https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-25506d7604624dfc2c3d66729bd5dffc4534b63a3bffd2d7213ffa1dcb794ae5R24

Do you think the object template should be made in a separate function? I find it not really helpful here, so I'm repeating it once for the base file (the folder).

lib/routes/veeam/head.js
ghost: see https://github.com/scality/cloudserver/pull/4942/commits/3941ed635c72ef5bf941e7a729d681a09403b525#diff-a23b06e1deac8b9332f76c3ae6723a32d6a49e7b7845e7f94f76e17190be0eb0R159

lib/routes/veeam/list.js
ghost: List response looks like this (tested with Artesca):

image

lib/routes/veeam/utils.js
```diff
@@ -0,0 +1,16 @@
...
+ * Decodes an URI and return the result.
+ * Do the same decoding than in S3 server
+ * @param {string} uri - uri to decode
+ * @returns {string} -
+ */
+function _decodeURI(uri) {
+    return decodeURIComponent(uri.replace(/\+/g, ' '));
+}
+
+const validPath = '.system-d26a9498-cb7c-4a87-a44a-8ae204f5ba6c/';
```
anurag4DSB, line 151: would it be better to add this in https://github.com/scality/cloudserver/blob/development/8.6/constants.js?

lib/routes/veeam/head.js
ghost: Note: I squashed again to update the Arsenal version used and some linter issues. See https://github.com/scality/cloudserver/pull/4942/commits/067383064e138bb79da293d36b6189bd97ed3ee1

lib/routes/veeam/list.js
ghost: Note: the double-double quotes for the ETags are now solved in https://github.com/scality/cloudserver/pull/4942/commits/067383064e138bb79da293d36b6189bd97ed3ee1#diff-25506d7604624dfc2c3d66729bd5dffc4534b63a3bffd2d7213ffa1dcb794ae5R42

lib/routes/routeVeeam.js
```diff
@@ -0,0 +1,219 @@
...
+ * @param {object} request - incoming request
+ * @param {object} response - response object
+ * @param {string} api - HTTP verb
+ * @param {object} log - logger instance
+ * @param {function} callback -
+ * @returns {undefined}
+ */
+function authorizationMiddleware(request, response, api, log, callback) {
+    if (!api) {
+        return errors.AccessDenied;
```
anurag4DSB: Should we return callback here?

lib/routes/routeVeeam.js
```diff
@@ -0,0 +1,219 @@
...
+    return async.waterfall([
+        next => auth.server.doAuth(request, log, (err, userInfo, authorizationResults, streamingV4Params) => {
+        if (err) {
+                log.debug('authentication error', {
+                    error: err,
+                    method: request.method,
+                    bucketName: request.bucketName,
+                    objectKey: request.objectKey,
+                });
+            }
```
anurag4DSB: nit: identation

lib/routes/routeVeeam.js
```diff
@@ -0,0 +1,219 @@
...
+        if (err || !bucketMd) {
+            return responseXMLBody(err, null, response, log);
+        }
+        return callback(request, response, bucketMd, log);
+    });
+}
+
+function _normalizeVeeamRequest(req) {
+    /* eslint-disable no-param-reassign */
+    // Rewriting the URL is needed for the V4 signature check
```
anurag4DSB: Would be great to mention also, why, rewriting the URL is needed for the V4 signature check

lib/routes/routeVeeam.js
francoisferrand: much more readable indeed (esp. with the comment 😀), but missing the `else` to behave the same?

lib/routes/routeVeeam.js
francoisferrand: probably not now indeed... 😞

lib/routes/veeam/list.js
francoisferrand: i guess that is a question for PO / Product Director...

ghost: I will ask Thomas then, thanks!

lib/routes/routeVeeam.js
ghost: Actually the return should be in the nested if, no need for else here, and then we will achieve the same result

lib/routes/veeam/list.js
ghost: He agrees on the VIRTUAL name, to be documented in the SOSAPI documentation for Artesca.

lib/routes/routeVeeam.js
ghost: I updated witht he old way of checking with a comment in https://github.com/scality/cloudserver/pull/4942/commits/93776d6f12455799f93951ef3da1753303db1439#diff-d54eeee0f403b38d9cec6ec8ab8caca914402b5705e4bfb96179c5065b8ad8a0R51, as per discussion over Slack

lib/routes/veeam/list.js
```diff
@@ -32,7 +32,7 @@ function buildXMLResponse(request, arrayOfFiles, versioned = false) {
         Contents: [],
         CommonPrefixes: [],
     };
-    arrayOfFiles.forEach(file => list[versioned ? 'Versions' : 'Contents'].push({
+    const entries = arrayOfFiles.map(file => ({
```
ghost, line 35: Just to keep track of your comments @francoisferrand :

> may be written more expressively like this:

```
const entries = arrayOfFiles.map(file => { ..... })
// Add the folder as the base file
entries.push({ ... }
```

> and that 'same' list of entries can be used to build the appropriate `list` object afterwards, without the error-prone `list[versioned ? 'Versions' : 'Contents']`


-> It's handled here https://github.com/scality/cloudserver/pull/4942/commits/93776d6f12455799f93951ef3da1753303db1439#diff-25506d7604624dfc2c3d66729bd5dffc4534b63a3bffd2d7213ffa1dcb794ae5R35
Also tested with Artesca

anurag4DSB (approved):
Logic LGTM per design + extra enhancements for the UI.

Small comments, and mostly look good with all the tests.
I have reviewed the arsenal PR but have yet to review the ZKOP one.

lib/routes/routeVeeam.js
ghost: Actually it should return a call to `responseXMLBody` with this error, the callback is the API

lib/routes/veeam/utils.js
ghost: I understand but I feel like we shouldn't mix custom route and standard configuration... And at some point, it may even be set at the kubernetes level... I'm fine putting it in constants.js but for now I do not really see a benefit of doing so. Please feel free to challenge my point of view

lib/routes/routeVeeam.js
ghost: All good in https://github.com/scality/cloudserver/pull/4942/files#diff-d54eeee0f403b38d9cec6ec8ab8caca914402b5705e4bfb96179c5065b8ad8a0R101, my squash commits will fix it in the commit you referenced.

lib/routes/routeVeeam.js
ghost: Done: https://github.com/scality/cloudserver/pull/4942/commits/e31d948aad0278ca1d2994d0999f573c437ba724#diff-d54eeee0f403b38d9cec6ec8ab8caca914402b5705e4bfb96179c5065b8ad8a0R96

lib/routes/routeVeeam.js
ghost: I hope this is clear enough: https://github.com/scality/cloudserver/pull/4942/commits/e31d948aad0278ca1d2994d0999f573c437ba724#diff-d54eeee0f403b38d9cec6ec8ab8caca914402b5705e4bfb96179c5065b8ad8a0R141

lib/routes/veeam/put.js
```diff
@@ -0,0 +1,90 @@
...
+            const capabilities = bucketMd._capabilities || {
+                VeeamSOSApi: {},
+            };
+            // Initialize Veeam capability if not already there
+            if (!capabilities.VeeamSOSApi) {
+                capabilities.VeeamSOSApi = {};
+            }
+            // Validate the JS object schema with joi and prepare the object for
+            // further logic
+            const validateFn = isSystemXML ? parseSystemSchema : parseCapacitySchema;
```
francoisferrand: can use `isSystemXML()`

lib/routes/veeam/put.js
```diff
@@ -0,0 +1,90 @@
...
+            let validatedData = null;
+            try {
+                validatedData = validateFn(parsedXML);
+            } catch (err) {
+                log.error('xml file did not pass validation', { err });
+                return next(errors.MalformedXML);
+            }
+            // System and Capacity files also include the last-modified date used by Veeam
+            const lastModified = (new Date()).toISOString();
+            if (isSystemXML) {
```
francoisferrand: we may actually be able to use `getFileToBuild()` here from utils ?

something like :

```js
capabilities.VeeamSOSApi = getFileToBuild(request, validatedData).value
```

lib/api/multiObjectDelete.js
```diff
@@ -481,6 +484,9 @@ function multiObjectDelete(authInfo, request, log, callback) {
                             versionId: result.versionId,
                         };
                         if (result.isAllowed) {
+                            if (validObjectKeys.includes(entry.key)) {
+                                inPlayInternal.push(entry.key);
+                            }
```
francoisferrand, line 491: i am wondering about this.

When the object is a "veam" object, we add it to "inPlayInteral" so we will process it with veam-specific code (which is fine), but we also add it to inPlay:
* so we will run checkBucketMetadata, which is needed i guess?
* eventually we will call getObjMetadataAndDeleteStep() : so we will try to delete a "real" object with that name from the "real" metadata backend? which will likely fail (since it would not be present)?

Or am I missing something?

lib/routes/veeam/put.js
```diff
@@ -0,0 +1,90 @@
...
+        },
+        (value, next) => parseString(value, { explicitArray: false }, (err, parsed) => {
+            // Convert the received XML to a JS object
+            if (err) {
+                return next(errors.MalformedXML);
+            }
+            return next(null, parsed);
+        }),
+        (parsedXML, next) => {
+            const isSystemXML = request.objectKey.endsWith('/system.xml');
```
francoisferrand: isSystemXML()

lib/routes/veeam/delete.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @param {string} bucketName - bucket name
+ * @param {string} objectKey - object key to delete
+ * @param {object} bucketMd - bucket metadata from the db
+ * @param {object} log - logger object
+ * @param {function} callback - callback
+ * @returns {undefined} -
+ */
+function deleteVeeamCapabilities(bucketName, objectKey, bucketMd, log, callback) {
+    const isSystemXML = objectKey.endsWith('system.xml');
```
francoisferrand: isSystemXML()

lib/routes/veeam/put.js
ghost: Done: https://github.com/scality/cloudserver/pull/4942/commits/8c671ef0ac52fb1f0100cbc68a7f5da0eb9388d5#diff-38a3a2f10ad2b2920ef0e2d5815a184a4bdaab48fe7c82ecc0767c0be2fb8c8aR50

lib/routes/veeam/put.js
ghost: Done with error handling: https://github.com/scality/cloudserver/pull/4942/commits/ed85f0bb02689051981c529557953f13fa3ffded#diff-38a3a2f10ad2b2920ef0e2d5815a184a4bdaab48fe7c82ecc0767c0be2fb8c8aR58

Also, I cannot set to capabilities directly, it would override existing data, so I'm just overriding if needed.

lib/api/multiObjectDelete.js
ghost: You are right, I moved the inPlay in the `else` to make sure we do not duplicate the custom files
See https://github.com/scality/cloudserver/pull/4942/commits/ed85f0bb02689051981c529557953f13fa3ffded#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR489

Without this change, there is no bad effects, it just does more computations than needed.

lib/routes/veeam/put.js
ghost: See https://github.com/scality/cloudserver/pull/4942/commits/ed85f0bb02689051981c529557953f13fa3ffded#diff-38a3a2f10ad2b2920ef0e2d5815a184a4bdaab48fe7c82ecc0767c0be2fb8c8aR50

lib/routes/veeam/delete.js
ghost: See https://github.com/scality/cloudserver/pull/4942/commits/ed85f0bb02689051981c529557953f13fa3ffded#diff-3d01b0ce4989d0b3f542dc7de83f55537d4fbfa70d7d63258b7201b7dffdd95bR18

alexanderchan-scality (approved):

lib/routes/veeam/put.js
```diff
@@ -0,0 +1,83 @@
...
+            return next(null, parsed);
+        }),
+        (parsedXML, next) => {
+            const capabilities = bucketMd._capabilities || {
+                VeeamSOSApi: {},
+            };
+            // Initialize Veeam capability if not already there
+            if (!capabilities.VeeamSOSApi) {
+                capabilities.VeeamSOSApi = {};
+            }
```
francoisferrand: not really needed:
* the 'default' initi is redundant with the later 'Initialize Veeam capability if not already there'
* all these are actually not needed, since we set it anyway line 62: ` capabilities.VeeamSOSApi = { ...`

i guess we could simply do this (at line 62):
```
            capabilities.VeeamSOSApi = {
                ...(capabilities.VeeamSOSApi || {}),
                ...file.value,
            };
```

francoisferrand (approved):

lib/routes/veeam/put.js
ghost: That works too, thanks! See https://github.com/scality/cloudserver/pull/4942/commits/673177c942bb2250c7358fc2e6ca89fbb9cdf1a1#diff-38a3a2f10ad2b2920ef0e2d5815a184a4bdaab48fe7c82ecc0767c0be2fb8c8aR59

Closing as resolved.

PR #4965, opened 2023-01-12
impr: CLDSRV-328 adapt metric naming conventions

naren-rajendran:
Adapt metric naming conventions.

### DIFF
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -3,65 +3,65 @@ const client = require('prom-client');
 
 const collectDefaultMetrics = client.collectDefaultMetrics;
 const numberOfBuckets = new client.Gauge({
-    name: 'cloud_server_number_of_buckets',
+    name: 's3_cloudserver_buckets_count',
     help: 'Total number of buckets',
 });
 const numberOfObjects = new client.Gauge({
-    name: 'cloud_server_number_of_objects',
+    name: 's3_cloudserver_objects_count',
     help: 'Total number of objects',
 });
 const lastReportTimestamp = new client.Gauge({
-    name: 'cloud_server_last_report_timestamp',
+    name: 's3_cloudserver_last_report_timestamp',
     help: 'Timestamp of last object/bucket count report',
 });
 const numberOfIngestedObjects = new client.Gauge({
-    name: 'cloud_server_number_of_ingested_objects',
-    help: 'Number of out of band ingestion',
+    name: 's3_cloudserver_ingested_objects_count',
+    help: 'Total number of out of band objects ingested',
 });
 const dataIngested = new client.Gauge({
-    name: 'cloud_server_data_ingested',
+    name: 's3_cloudserver_ingested_bytes',
     help: 'Cumulative size of data ingested in bytes',
 });
 const dataDiskAvailable = new client.Gauge({
-    name: 'cloud_server_data_disk_available',
+    name: 's3_cloudserver_disk_available_bytes',
     help: 'Available data disk storage in bytes',
 });
 const dataDiskFree = new client.Gauge({
-    name: 'cloud_server_data_disk_free',
+    name: 's3_cloudserver_disk_free_bytes',
     help: 'Free data disk storage in bytes',
 });
 const dataDiskTotal = new client.Gauge({
-    name: 'cloud_server_data_disk_total',
+    name: 's3_cloudserver_disk_total_bytes',
     help: 'Total data disk storage in bytes',
 });
 
 const labelNames = ['method', 'action', 'code'];
 const httpRequestsTotal = new client.Counter({
     labelNames,
-    name: 'http_requests_total',
-    help: 'Total number of HTTP requests',
+    name: 's3_cloudserver_http_requests_total',
+    help: 'Total number of cloudserver HTTP requests',
 });
 const httpRequestDurationSeconds = new client.Histogram({
-    name: 'http_request_duration_seconds',
+    name: 's3_cloudserver_http_request_duration_seconds',
     help: 'Duration of HTTP requests in seconds',
     labelNames,
     buckets: [0.0001, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5], // buckets for response time from 0.1ms to 500ms
 });
 const httpActiveRequests = new client.Gauge({
-    name: 'http_active_requests',
-    help: 'Number of HTTP in-flight requests',
+    name: 's3_cloudserver_http_active_requests',
+    help: 'Number of cloudserver HTTP in-flight requests',
 });
 
 const httpRequestSizeBytes = new client.Summary({
     labelNames,
-    name: 'http_request_size_bytes',
-    help: 'The HTTP request sizes in bytes.',
+    name: 's3_cloudserver_http_request_size_bytes',
+    help: 'Cloudserver HTTP request sizes in bytes',
 });
 
 const httpResponseSizeBytes = new client.Summary({
     labelNames,
-    name: 'http_response_size_bytes',
-    help: 'The HTTP response sizes in bytes.',
+    name: 's3_cloudserver_http_response_size_bytes',
+    help: 'Cloudserver HTTP response sizes in bytes',
 });
 
 function promMetrics(method, bucketName, code, action,
@@ -164,7 +164,7 @@ async function routeHandler(req, res, cb) {
     const contentLen = Buffer.byteLength(promMetrics, 'utf8');
     res.writeHead(200, {
         'Content-Length': contentLen,
-        'Content-Type': client.register.contentType
+        'Content-Type': client.register.contentType,
     });
     res.end(promMetrics);
     return undefined;
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -398,7 +398,7 @@ jobs:
         gem install nokogiri:1.12.5 fog-aws:1.3.0 json mime-types:3.1 rspec:3.5
     - name: Install Java dependencies
       run: |
-        sudo apt-get install -y default-jdk maven
+        sudo apt-get update && sudo apt-get install -y --fix-missing default-jdk maven
     - name: Setup CI services
       run: docker-compose --profile ceph up -d
       working-directory: .github/docker
--- a/monitoring/alerts.test.yaml
+++ b/monitoring/alerts.test.yaml
@@ -15,9 +15,9 @@ tests:
     interval: 1m
     input_series:
       # 1% error rate for 5m, then 4% error rate for 5m, then 6% error rate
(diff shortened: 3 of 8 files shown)


lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
...
-    name: 'http_request_duration_seconds',
+    name: 's3_cloudserver_http_request_duration_seconds',
     help: 'Duration of HTTP requests in seconds',
     labelNames,
     buckets: [0.0001, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5], // buckets for response time from 0.1ms to 500ms
 });
 const httpActiveRequests = new client.Gauge({
-    name: 'http_active_requests',
-    help: 'Number of HTTP in-flight requests',
+    name: 's3_cloudserver_http_active_requests',
```
francoisferrand, line 51: This is a standard metrics, that is why a standard name is used.

This is the recommended way to use metrics, to allow "generic" metrics; as discussed with the architects months ago.

c.f. prometheus metrics & label name recommandations: https://prometheus.io/docs/practices/naming/

francoisferrand:
I am off today but will comment on this one, please don't merge just yet...

lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
 
 const collectDefaultMetrics = client.collectDefaultMetrics;
 const numberOfBuckets = new client.Gauge({
-    name: 'cloud_server_number_of_buckets',
+    name: 's3_cloudserver_buckets',
```
francoisferrand: should have a unit : `..._buckets_count`

lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
...
     help: 'Total number of buckets',
 });
 const numberOfObjects = new client.Gauge({
-    name: 'cloud_server_number_of_objects',
+    name: 's3_cloudserver_objects',
```
francoisferrand: should have a unit : `..._objects_count`

lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
...
 });
 const lastReportTimestamp = new client.Gauge({
-    name: 'cloud_server_last_report_timestamp',
+    name: 's3_cloudserver_last_report_timestamp',
     help: 'Timestamp of last object/bucket count report',
 });
 const numberOfIngestedObjects = new client.Gauge({
-    name: 'cloud_server_number_of_ingested_objects',
-    help: 'Number of out of band ingestion',
+    name: 's3_cloudserver_ingested_objects',
```
francoisferrand: should have a unit : `..._ingested_objects_count`

lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
...
 const httpRequestsTotal = new client.Counter({
     labelNames,
-    name: 'http_requests_total',
-    help: 'Total number of HTTP requests',
+    name: 's3_cloudserver_http_requests_total',
+    help: 'Total number of cloudserver HTTP requests',
 });
 const httpRequestDurationSeconds = new client.Histogram({
-    name: 'http_request_duration_seconds',
+    name: 's3_cloudserver_http_request_duration_seconds',
```
francoisferrand, line 45: This is a standard metrics, that is why a standard name is used.

This is the recommended way to use metrics, to allow "generic" metrics; as discussed with the architects months ago.

c.f. prometheus metrics & label name recommandations: prometheus.io/docs/practices/naming

lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
...
+    name: 's3_cloudserver_disk_bytes',
     help: 'Total data disk storage in bytes',
 });
 
 const labelNames = ['method', 'action', 'code'];
 const httpRequestsTotal = new client.Counter({
     labelNames,
-    name: 'http_requests_total',
-    help: 'Total number of HTTP requests',
+    name: 's3_cloudserver_http_requests_total',
```
francoisferrand, line 41: This is a standard metrics, that is why a standard name is used.

This is the recommended way to use metrics, to allow "generic" metrics; as discussed with the architects months ago.

c.f. prometheus metrics & label name recommandations: prometheus.io/docs/practices/naming

lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
...
     help: 'Available data disk storage in bytes',
 });
 const dataDiskFree = new client.Gauge({
-    name: 'cloud_server_data_disk_free',
+    name: 's3_cloudserver_disk_free_bytes',
     help: 'Free data disk storage in bytes',
 });
 const dataDiskTotal = new client.Gauge({
-    name: 'cloud_server_data_disk_total',
+    name: 's3_cloudserver_disk_bytes',
```
francoisferrand: the name is not clear at all. To be consistent with the other metrics, should be `..._disk_total_bytes`

lib/utilities/monitoringHandler.js
```diff
@@ -3,65 +3,65 @@ const client = require('prom-client');
...
+    name: 's3_cloudserver_http_request_size_bytes',
+    help: 'Cloudserver HTTP request sizes in bytes',
 });
 
 const httpResponseSizeBytes = new client.Summary({
     labelNames,
-    name: 'http_response_size_bytes',
-    help: 'The HTTP response sizes in bytes.',
+    name: 's3_cloudserver_http_response_size_bytes',
+    help: 'Cloudserver HTTP response sizes in bytes',
```
francoisferrand, line 64: This is a standard metrics, that is why a standard name is used.

francoisferrand (changes requested):
Sorry there was no such discussion on https://docs.google.com/document/d/1qw9P3p7VuFba_KbxJahPgUSxzusGSo2OPCSGLA3MTVg and it is thus coming late, but I really don't think this is the way to go.

* Metrics are somewhat intrinsic part of each component, and not really related to the domain : in RING, we may indeed have an "s3" domain (which btw is completely redundant with cloudserver), but not in artesca which is only s3
* The metrics are technical, so they should relate to the actual component: use of metrics is at a technical level, not a product level; and any user of the metrics must thus understand the architecture
* Using a prefix to distinguish multiple instances is not the way to go : this should be done with labels
* If there is a really a need to enforce specific metrics name rules on a product, this should be done at product level instead of hard-coded in a component : the metrics may be renamed via specific rules in the prometheus configuration (i.e. by updating the `__name__` label) so that the remapping affects only a specific deployment
* Handling the transition and operational impact (no access to old metrics!!) must be fully documented part of the design, including how it will be handled on *all* affected products
* Some clients of Artesca (CNES in particular) may already be 

(generally, I would _love_ to cleanup some metrics, like adding missing units and changing the cloud_server prefix to cloudserver ; but i think the operational impacts negates any benefit we get...)

lib/utilities/monitoringHandler.js
naren-rajendran: Following the suggestions here https://www.robustperception.io/on-the-naming-of-things/ and https://docs.google.com/document/d/1qw9P3p7VuFba_KbxJahPgUSxzusGSo2OPCSGLA3MTVg, avoiding the suffix because it is a gauge and not counter.

lib/utilities/monitoringHandler.js
naren-rajendran: same as above.

lib/utilities/monitoringHandler.js
naren-rajendran: same as above.

lib/utilities/monitoringHandler.js
naren-rajendran: currently in contention, should be resolved soon.

lib/utilities/monitoringHandler.js
naren-rajendran: same as above.

lib/utilities/monitoringHandler.js
naren-rajendran: same as above

lib/utilities/monitoringHandler.js
naren-rajendran: same as above

lib/utilities/monitoringHandler.js
francoisferrand: That suggestion is "do not use the `count` or `total`" suffix for gauges, fine.

Anyway my point is not about the specific word used, but that semantically it is not clear what a `s3_cloudserver_buckets` metric reprensents : and i think we would benefit from having suffix ("unit" or otherwise) to better convey what this metrics represents...

(though specifically here I would agree to disagree that `count` implies a counter : i think a user of the dashboard would never expect that the "bucket_count" is a (prometheus) counter and can never go down...)

naren-rajendran: The same convention is used in other components, meaning no suffixes such as `count` or `total` for a gauge. So not modifying it.

nicolas2bert (approved):

naren-rajendran:
@bert-e reset

lib/utilities/monitoringHandler.js
francoisferrand: not sure I understand the reasoning here : since we are currently renaming all metrics, now is the time to cleanup these as well... The goal here is to have meaningful and clear names, and the (overall) guidelines is that metrics should have a unit : so it seems there is something to do indeed...

as this been disucssed within the monitoring pod already, is there a decision there?

lib/utilities/monitoringHandler.js
francoisferrand: any progress?

naren-rajendran: I was not part of the meeting where the decision was made and it is to have the prefix `s3` followed by `component` name.

lib/utilities/monitoringHandler.js
naren-rajendran: does `s3_cloudserver_total_buckets` make sense? just trying not to use it as a suffix so that it is distinguishable from a counter!

naren-rajendran:
@bert-e reset

lib/utilities/monitoringHandler.js
naren-rajendran: updated to include `count`

naren-rajendran:
@francoisferrand I have updated the metric names per your suggestion. Http metrics carry the prefix and component names because it was decided by the team (I was not part of the decision)

francoisferrand (approved):
LGTM

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e reset

naren-rajendran:
@bert-e approve

PR #4967, opened 2023-01-13
Feature/cldsrv 329/migrate to GitHub actions 8.x

alexanderchan-scality:


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,358 @@
...
+    env:
+      S3BACKEND: mem
+      MPU_TESTING: true
+      S3METADATA: mongodb
+      S3KMS: file
+      CLOUDSERVER_IMAGE: ghcr.io/${{ github.repository }}/cloudserver:${{ github.sha }}
+      S3_LOCATION_FILE: /usr/src/app/tests/locationConfig/locationConfigTests.json
+      DEFAULT_BUCKET_KEY_FORMAT: ${{ matrix.version }}
+      METADATA_MAX_CACHED_BUCKETS: ${{ matrix.md-max-cache-buckets }}
+      JOB_NAME: ${{ github.job }}-${{ matrix.version }}
```
tcarmet: I could be misunderstanding what you're trying to achieve or missing something, but as far as I know, `include` should be used along with some additional context so that it "includes" value for some cases.
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

Here's what I think you tried to do:
```suggestion
  mongo-ft-tests:
    strategy:
      matrix:
        config: # Any key here, beside include sets a loop
        - version: v0
          md-max-cache-buckets: ''
        - version: v1
          md-max-cache-buckets: '1'
    runs-on: ubuntu-latest
    needs: build
    env:
      S3BACKEND: mem
      MPU_TESTING: true
      S3METADATA: mongodb
      S3KMS: file
      CLOUDSERVER_IMAGE: ghcr.io/${{ github.repository }}/cloudserver:${{ github.sha }}
      S3_LOCATION_FILE: /usr/src/app/tests/locationConfig/locationConfigTests.json
      DEFAULT_BUCKET_KEY_FORMAT: ${{ matrix.config.version }} # note that config is added here
      METADATA_MAX_CACHED_BUCKETS: ${{ matrix.config.md-max-cache-buckets }}
      JOB_NAME: ${{ github.job }}-${{ matrix.config.version }}
```

But worth noting that functionality wise I don't think I'm changing anything, both works the same.

nicolas2bert (commented):
Adding this as a note so we do not forget: `.github/workflows/release` would have to be updated as well.

nicolas2bert (commented):
Should it target older development/8.x branches?

.github/workflows/tests.yaml
alexanderchan-scality: split the tests into 2 separate jobs as they call different tests

nicolas2bert (approved):

.github/workflows/release.yaml
```diff
@@ -8,22 +8,43 @@ on:
...
-    uses: scality/workflows/.github/workflows/docker-build.yaml@v1
-    secrets: inherit
-    with:
-      push: true
-      registry: registry.scality.com
-      namespace: ${{ github.event.repository.name }}
-      name: ${{ github.event.repository.name }}
-      context: .
-      file: images/svc-base/Dockerfile
-      tag: ${{ github.event.inputs.tag }}-svc-base
```
tcarmet: Confirm this with @scality/sup-setup, but I think you should keep this job, even if we won't be using 8.x on federation anytime soon you can build and push both docker images on each release.

But if you don't want to keep it then delete the image as well `images/svc-base/Dockerfile`

tcarmet (approved):
LGTM, just left one thought on the docker images that are released.

PR #4988, opened 2023-01-31
CLDSRV-335: Build federation image in tests workflow

dvasilas:
When a commit is pushed (`tests` workflow) build the Federation docker image in parallel with the image used for tests.

This will allow us to build Federation with an image before doing a release. 

Issue: CLDSRV-335

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -154,6 +154,18 @@ jobs:
           cache-from: type=gha,scope=cloudserver
           cache-to: type=gha,mode=max,scope=cloudserver
 
+  build-federation-image:
+    uses: scality/workflows/.github/workflows/docker-build.yaml@v1
+    secrets: inherit
+    with:
+      push: true
+      registry: registry.scality.com
+      namespace: cloudserver-dev
+      name: cloudserver
+      context: .
+      file: images/svc-base/Dockerfile
+      tag: ${{ github.sha }}-svc-base
+
   multiple-backend:
     runs-on: ubuntu-latest
     needs: build


dvasilas:
:robot:

dvasilas:
Question to reviewers: I should revert this change in 8.x branches, right ?

.github/workflows/tests.yaml
```diff
@@ -154,6 +154,18 @@ jobs:
           cache-from: type=gha,scope=cloudserver
           cache-to: type=gha,mode=max,scope=cloudserver
 
+  build-federation-image:
```
jonathan-gramain, line 157: Instead of adding a new build stage, we probably can just add the `file` parameter to the `build` stage, and remove the `ghcr.io` tag. I don't think we need to keep the existing images for 7.x branches hence this suggestion.

We can have this forward-ported to the `development/7.70` branch but revert the changes manually on the `w/8.x` integration branches so that Artesca which does not use the same base image is not modified.

dvasilas: I agree (and tried to do it) but the existing image is used later in the workflow (we deploy a cloudserver container with `docker-compose`).

We can use the Federation image instead but it requires some adjustments to `docker-compose.yaml` (use volumes for `/conf` and `/logs` and setup the right permissions, add a `supervisord.conf`, ...).

It might be simpler to build both images, and at the end cleanup the non-Federation image as it has no use outside of the CI (at least in `developement/7.10` and `development/7.70`).
What do you think ?

dvasilas: Update: after discussion with @tcarmet , it should be fine to go with the simple approach (generate and keep both images).

jonathan-gramain (approved):

welansari (approved):

.github/workflows/tests.yaml
francoisferrand: this is also what is done in the release build, in 8.x : and it helps to avoid waterflow-related issues.

francoisferrand (approved):

dvasilas:
Change reverted in 8.X branches.

PR #5000, opened 2023-02-06
CLDSRV-338: fix v2 listing with max-keys 0

alexanderchan-scality:


lib/api/bucketGet.js
```diff
@@ -214,6 +214,8 @@ function processMasterVersions(bucketName, listParams, list) {
     xmlParams.forEach(p => {
         if (p.value && skipUrlEncoding.has(p.tag)) {
             xml.push(`<${p.tag}>${p.value}</${p.tag}>`);
+        } else if (p.tag === 'MaxKeys' && p.value >= 0) {
+            xml.push(`<${p.tag}>${escapeXmlFn(p.value)}</${p.tag}>`);
         } else if (p.value || p.tag === 'KeyCount') {
```
jonathan-gramain: We might want to group this case under the one below for `KeyCount` which looks to handle a similar case (i.e. with `tag === 'KeyCount' || tag === 'MaxKeys'`)

jonathan-gramain (approved):

alexanderchan-scality:
ping

francoisferrand (approved):
LGTM, though the description incorrectly mentions v2 : the fixes applies to both v1 and v2 (which is actually expected, i guess)

alexanderchan-scality:
@francoisferrand 
yes. the description is somewhat incomplete; fix is to be applied for both v1 and v2 listings. 
`v2` was mentioned because the error is only observed when performing v2 listing using the cli tool; the tool expects the tag to contain a number value

PR #5003, opened 2023-02-09
impr: CLDSRV-340 improve preprocessingVersioningDelete()

jonathan-gramain:
- remove dead code handling nullVersionId and when no reference to null version is found: the confusion is that the metadata is already coming from the null version fetch due to the "reqVersionId" being "null", so it can only be the null version itself.

- remove the "isNull" argument returned as it is misused (for metrics) and can be incorrect, see S3C-7440

- simplify the function by reorganizing the logic and removing the callback argument, just returning an options object

- delete replay keys on multiobject delete, instead of passing the replayId option via preprocessingVersioningDelete() (a similar change has been done in objectDelete but it resulted in duplication of replayId setting)

- remove all MPU-related tests for this helper, as the helper does not return replayId anymore, they became irrelevant


jonathan-gramain:
ping

welansari (approved):

naren-rajendran (approved):

PR #5008, opened 2023-02-14
Add azure_archive location type

KillianG:
Issue: CLDSRV-337

The purpose of this PR is to add azure_archive location type to the config

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -237,6 +237,60 @@ function hdClientLocationConstraintAssert(configHd) {
     return hdclientFields;
 }
 
+function azureArchiveLocationConstraintAssert(locationObj) {
+    const checkedFields = [
+        'azureContainerName',
+        'azureStorageEndpoint',
+    ];
+    if (Object.keys(locationObj.details).length === 0 ||
+        !checkedFields.every(field => field in locationObj.details)) {
+        return;
+    }
+    const {
+        azureContainerName,
+        azureStorageEndpoint,
+    } = locationObj.details;
+    const stringFields = [
+        azureContainerName,
+        azureStorageEndpoint,
+    ];
+    stringFields.forEach(field => {
+        assert(typeof field === 'string',
+            `bad config: ${field} must be a string`);
+    });
+
+    let hasAuthMethod = false;
+    if (locationObj.details.sasToken !== undefined) {
+        assert(typeof locationObj.details.sasToken === 'string',
+            `bad config: ${locationObj.details.sasToken} must be a string`);
+        hasAuthMethod = true;
+    }
+
+    if (locationObj.details.azureStorageAccountName !== undefined &&
+        locationObj.details.azureStorageAccessKey !== undefined) {
+        assert(typeof locationObj.details.azureStorageAccountName === 'string',
+            `bad config: ${locationObj.details.azureStorageAccountName} must be a string`);
+        assert(typeof locationObj.details.azureStorageAccessKey === 'string',
+            `bad config: ${locationObj.details.azureStorageAccessKey} must be a string`);
+        assert(!hasAuthMethod, 'Multiple authentication methods are not allowed');
+        hasAuthMethod = true;
+    }
+
+    if (locationObj.details.tenantId !== undefined &&
+        locationObj.details.clientId !== undefined &&
+        locationObj.details.clientKey !== undefined) {
+        assert(typeof locationObj.details.tenantId === 'string',
+            `bad config: ${locationObj.details.tenantId} must be a string`);
+        assert(typeof locationObj.details.clientId === 'string',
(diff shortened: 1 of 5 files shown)


lib/Config.js
```diff
@@ -280,7 +280,7 @@ function dmfLocationConstraintAssert(locationObj) {
 function locationConstraintAssert(locationConstraints) {
     const supportedBackends =
           ['mem', 'file', 'scality',
-           'mongodb', 'dmf'].concat(Object.keys(validExternalBackends));
+           'mongodb', 'dmf', 'azure_archive'].concat(Object.keys(validExternalBackends));
     assert(typeof locationConstraints === 'object',
         'bad config: locationConstraints must be an object');
     Object.keys(locationConstraints).forEach(l => {
```
welansari, line 340: Do we need to add a check like for dmf ?https://github.com/scality/cloudserver/blob/0fb54c9d3153dded1a0f45d144e1d8448eda7e75/lib/Config.js#L391-L393

francoisferrand: even though it should not really be relevant (since cloudserver is not connecting to this kind of location), for consistency i think it would be better indeed...

KillianG: Done [bc20467](https://github.com/scality/cloudserver/pull/5008/commits/bc20467fcc26794be0b5eeb153d0c4ab5b31c3ea)

welansari (approved):

ghost (approved):

lib/Config.js
```diff
@@ -237,6 +237,60 @@ function hdClientLocationConstraintAssert(configHd) {
     return hdclientFields;
 }
 
+function azureArchiveLocationConstraintAssert(locationObj) {
```
francoisferrand, line 240: please add unit tests for this piece of code

francoisferrand (changes requested):

lib/Config.js
KillianG: Done: https://github.com/scality/cloudserver/pull/5008/commits/95c50953064d4f0a5c4e79a4c4bde212aa342b04

tests/unit/Config.js
```diff
@@ -311,4 +311,208 @@ describe('Config', () => {
         const { config } = require('../../lib/Config');
         assert.strictEqual(config.overlayVersion, 0);
     });
+
+    describe('azureArchiveLocationConstraintAssert', () => {
+        const { azureArchiveLocationConstraintAssert } = require('../../lib/Config');
+
+        it('should return undefined if azureStorageEndpoint is missing', () => {
```
francoisferrand: > return undefined

this function does not actually return "anything" : so we don't really care what it returns, only that is actually succeeds (as in "does not assert")

→ should just be "should succeed when azureStorageEndpoint is missing", and no reason to have any assert (just call the function)

tests/unit/Config.js
```diff
@@ -311,4 +311,208 @@ describe('Config', () => {
...
+            const locationObj = {
+                details: {
+                    azureContainerName: 'mycontainer',
+                }
+            };
+            assert.strictEqual(azureArchiveLocationConstraintAssert(locationObj), undefined);
+        });
+
+        it('should return undefined if azureContainerName is missing', () => {
```
francoisferrand: same

tests/unit/Config.js
```diff
@@ -311,4 +311,208 @@ describe('Config', () => {
...
+                details: {
+                    azureContainerName: 'mycontainer',
+                    azureStorageEndpoint: true,
+                }
+            };
+            try {
+                azureArchiveLocationConstraintAssert(locationObj);
+            } catch (err) {
+                assert.strictEqual(err.message, 'bad config: true must be a string');
+            }
```
francoisferrand: You've got the wrong assertion here.
We may check that exception, but actually this is not the important part : above all we care that `azureArchiveLocationConstraintAssert` crashes, i.e. that we don't return from `azureArchiveLocationConstraintAssert()` call...

```suggestion
            try {
                azureArchiveLocationConstraintAssert(locationObj);
                assert.fail('azureArchiveLocationConstraintAssert should have asserted !');
            } catch (err) {
                assert.strictEqual(err.message, 'bad config: true must be a string');
            }
```

francoisferrand: (same for all followup tests)

francoisferrand: Alternatively, you may use [assert.throws](https://nodejs.org/api/assert.html#assertthrowsfn-error-message) :

```suggestion
            assert.throws(() => azureArchiveLocationConstraintAssert(locationObj));
```

(possibly with 'bad config: true must be a string' as second argument, if you want to check the error message as well)

francoisferrand (approved):

PR #5042, opened 2023-03-09
CLDSRV-314 Implement listLifecycleCurrents

nicolas2bert:
`listLifecycleCurrents` API will be used by Backbeat Lifecycle Bucket Processor to list the master/current versions before a defined date. 

GET /_/backbeat/lifecycle/bucketname?list-type=current&before-date=<time_stamp>


lib/api/apiUtils/object/lifecycle.js
```diff
@@ -0,0 +1,53 @@
...
+            Size: v.Size,
+            Owner: {
+                ID: v.Owner.ID,
+                DisplayName: v.Owner.DisplayName
+            },
+            StorageClass: v.StorageClass,
+            TagSet: _makeTags(v.tags),
+            IsLatest: true, // for compatibily
+            DataStoreName: v.dataStoreName,
+            ListType: CURRENT_TYPE,
```
rachedbenmustapha, line 43: Is there a plan to make the list type different per-entry later, or can this go in the toplevel instead?

rachedbenmustapha (approved):

lib/api/apiUtils/object/lifecycle.js
nicolas2bert: I set it at the object level so that the client (Backbeat) has a hint on what "rule's group" each object should be compared to.
```
CURRENT_TYPE -> Expiration and Transition
NON_CURRENT_TYPE -> NoncurrentVersionExpiration and NoncurrentVersionTransition
...
```
We could also decide to move this listing type to the top level and do a loop to assign the `listType` for each object on the client side.

naren-rajendran (approved):

nicolas2bert:
@bert-e approve

PR #5053, opened 2023-03-14
CLDSRV-351 setup security scan of containers

tcarmet:


.github/workflows/security.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  build:
+    uses: scality/workflows/.github/workflows/docker-build.yaml@1.1.1
+    with:
+      namespace: scality
+      name: cloudserver
+  trivy:
+    needs: build
+    uses: scality/workflows/.github/workflows/trivy.yaml@a48dbfffd32d7ecdb5c93f3d12f19a6e7c85da4a
```
gaspardmoindrot, line 30: Make a PR of your modifications on the Workflow repo, it will be better and it will avoid putting a version that we will forget to change later - https://github.com/scality/workflows/blob/bugfix/RELENG-7239-trivy-scan/.github/workflows/trivy.yaml

tcarmet: forgot to reconvert this PR as draft, it's a change I was testing due to failures raised by trivy when scanning cloud server.

.dockerignore
```diff
@@ -7,3 +7,4 @@ localMetadata/*
 .tox
 coverage
 .DS_Store
+tests
```
tcarmet:
```suggestion
```

PR #5055, opened 2023-03-14
CLDSRV-349 null version compat mode config and tests

jonathan-gramain:
- support ENABLE_NULL_VERSION_COMPAT_MODE env var

  Cloudserver sets a flag in its configuration when the ENABLE_NULL_VERSION_COMPAT_MODE environment variable is set to "true".

- [tests] func test stage for ENABLE_NULL_VERSION_COMPAT_MODE

  Turn the "file-ft-tests" job into a matrix, to duplicate the suite with and without the ENABLE_NULL_VERSION_COMPAT_MODE environment variable.

  This will make sure Cloudserver behaves correctly (versioning, null version handling etc.)  when the compatibility mode is active.

### DIFF
--- a/.github/docker/docker-compose.yaml
+++ b/.github/docker/docker-compose.yaml
@@ -34,6 +34,7 @@ services:
       - S3KMIP_KEY
       - S3KMIP_CERT
       - S3KMIP_CA
+      - ENABLE_NULL_VERSION_COMPAT_MODE
     env_file:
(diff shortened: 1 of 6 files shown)


nicolas2bert (approved):

naren-rajendran (approved):

jonathan-gramain:
ping

PR #5058, opened 2023-03-15
ft: CLDSRV-354 handle null keys in metadata GET helpers

jonathan-gramain:
Update the helpers in metadataUtils to handle null keys, as well as keeping backward compatibility with null versioned keys.

The main change in logic for null keys is that, instead of fetching first the master key then the null versioned key, we first attempt to fetch the null key, and if not found, we fetch the master key (we may then also have to fetch the null versioned key for backward compatibility).

Take the chance to reduce tech debt by reorganizing the helpers responsibilities in a better way, and by using the "validateBucket" helper.

**note for reviews**: it can make sense to review individual commits

### DIFF
--- a/lib/metadata/metadataUtils.js
+++ b/lib/metadata/metadataUtils.js
@@ -118,6 +118,48 @@ function metadataGetObject(bucketName, objectKey, versionId, log, cb) {
         });
 }
 
+/**
+ * Validate that a bucket is accessible and authorized to the user,
+ * return a specific error code otherwise
+ *
+ * @param {BucketInfo} bucket - bucket info
+ * @param {object} params - function parameters
+ * @param {AuthInfo} params.authInfo - AuthInfo class instance, requester's info
+ * @param {string} params.requestType - type of request
+ * @param {string} [params.preciseRequestType] - precise type of request
+ * @param {object} params.request - http request object
+ * @param {RequestLogger} log - request logger
+ * @return {ArsenalError|null} returns a validation error, or null if validation OK
+ * The following errors may be returned:
+ * - NoSuchBucket: bucket is shielded
+ * - MethodNotAllowed: requester is not bucket owner and asking for a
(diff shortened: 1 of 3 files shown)


watsaqat (approved):

lib/metadata/metadataUtils.js
```diff
@@ -94,30 +65,67 @@ function metadataGetBucketAndObject(requestType, bucketName, objectKey,
...
+ * - MethodNotAllowed: requester is not bucket owner and asking for a
+ *     bucket policy operation
+ * - AccessDenied: bucket is not authorized
+ */
+function validateBucket(bucket, params, log) {
+    const { authInfo, requestType, preciseRequestType, request } = params;
+    if (bucketShield(bucket, requestType)) {
+        log.debug('bucket is shielded from request', {
+            requestType,
+            method: 'metadataValidateBucketAndObj',
```
tmacro:
```suggestion
            method: 'validateBucket',
```

lib/metadata/metadataUtils.js
```diff
@@ -132,35 +140,36 @@ function metadataGetObject(bucketName, objectKey, versionId, log, cb) {
  * @return {undefined} - and call callback with params err, bucket md
  */
 function metadataValidateBucketAndObj(params, log, callback) {
-    const { authInfo, bucketName, objectKey, versionId, requestType, preciseRequestType, request } = params;
-    const canonicalID = authInfo.getCanonicalID();
+    const { authInfo, bucketName, objectKey, versionId, requestType, request } = params;
     async.waterfall([
         function getBucketAndObjectMD(next) {
```
tmacro: could also take the opportunity to update the `function ...` syntax to `(...) =>`
```suggestion
        (next) => {
```

tmacro (approved):

jonathan-gramain:
ping

PR #5059, opened 2023-03-16
impr: CLDSRV-356 enhance processVersioningState() and fix replayId

jonathan-gramain:
- enhance general flow of processVersioningState(), to make it easier to read and update for null key handling

- fix an issue related to passing the uploadId for nonversioned buckets (linked to S3C-7361): remove a check "master.isNull" to also pass the uploadId as replayId when the master is non-versioned, so that it can be deleted by passing it to the metadata DELETE request

- make processVersioningState() return a 'nullVersionId' param rather than a "storeOptions", as it is always used to copy master to a null version, it simplifies a bit the API

- remove undefined params returned by the function to have clean unit tests

### DIFF
--- a/tests/unit/api/apiUtils/versioning.js
+++ b/tests/unit/api/apiUtils/versioning.js
@@ -70,14 +70,13 @@ describe('versioning helpers', () => {
                 versioningEnabledExpectedRes: {
                     options: {
                         versioning: true,
-                        nullVersionId: 'vnull',
+                        extraMD: {
+                            nullVersionId: 'vnull',
+                        },
                     },
                     // instruct to first copy the null version onto a
                     // newly created version key preserving the version ID
-                    storeOptions: {
-                        isNull: true,
-                        versionId: 'vnull',
-                    },
+                    nullVersionId: 'vnull',
                 },
                 versioningSuspendedExpectedRes: {
                     options: {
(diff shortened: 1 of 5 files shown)


tests/unit/api/apiUtils/versioning.js
```diff
@@ -99,15 +98,14 @@ describe('versioning helpers', () => {
                 versioningEnabledExpectedRes: {
                     options: {
                         versioning: true,
-                        nullVersionId: 'vnull',
-                        nullUploadId: 'fooUploadId',
+                        extraMD: {
```
watsaqat, line 101: I'm not sure I see the point of the `extraMD` nesting.

jonathan-gramain: The idea is to try to clarify the contract between `processVersioningState` and its caller, by explicitly grouping into an object the attributes should be added to the object metadata. The rest of parameters passed in `options` are specific options to the backend and/or to the calling code. What makes it still confusing in the current state is that the contract to the service call to actually store the object like e.g. `createAndStoreObject` takes params (passed via `metadataStoreParams` usually) which is also a huge flat object that mixes attributes to add, backend options and others. I did not go as far as to refactor how parameters are passed to the service (`metadataStoreParams`) but ideally this would also be clarified the same eventually, e.g. by grouping params by family, at least here it is supposed to make it clearer at this function's level even if the same attributes are passed in a flat object to the service helper thereafter.

Also, I changed the `storeParams` return value in that function, not specific enough as it specifically means to store a null version key from the master md in practice, into a `nullVersionId` returned value, which the caller takes as the hint that it should copy the master to this null version key (granted that it's not super specific either, I could call this `nullVersionIdToStore` maybe but it's getting a bit long of a name!)

jonathan-gramain: Moreover, the upcoming update to support null keys will add a few more fields to the response, so it looks better to start tidying things up a bit.

watsaqat (approved):

tmacro (approved):

jonathan-gramain:
ping

jonathan-gramain:
I added an extra commit that was needed on top of the `w/8.5` branch: https://github.com/scality/cloudserver/commit/13a5e14da58129cdd13336c3f7da58f470e0e2be

jonathan-gramain:
ping

PR #5065, opened 2023-03-16
feat: CLDSRV-357 update versioningPreprocessing() helper for null keys

jonathan-gramain:
- Modify the code flow of versioningPreprocessing() to support null keys in addition to the legacy "null versioned keys".
    
  NOTE: The null version compatibility mode is still enforced for now until all pieces are updated to make functional tests pass without the compatibility mode.

- pass deleteNullKey param to backend

  Pass the 'deleteNullKey' param that processVersioningState() may set to the Metadata backend, which tells it to delete the null key as the PUT is executed.

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -87,6 +87,9 @@ function _storeNullVersionMD(bucketName, objKey, nullVersionId, nullVersionMD, l
 */
 function _prepareNullVersionDeletion(bucketName, objKey, options, mst, log, cb) {
     const nullOptions = {};
+    if (!options.deleteData) {
+        return process.nextTick(cb, null, nullOptions);
+    }
     if (options.versionId === mst.versionId) {
         // no need to get another key as the master is the target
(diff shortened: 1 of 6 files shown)


watsaqat (approved):

lib/api/apiUtils/object/versioning.js
```diff
@@ -83,50 +85,54 @@ function _storeNullVersionMD(bucketName, objKey, objMD, options, log, cb) {
...
-                });
-        });
+function _deleteNullVersionMD(bucketName, objKey, options, log, cb) {
+    return metadata.deleteObjectMD(bucketName, objKey, options, log, err => {
+        if (err) {
+            log.warn('metadata error deleting null versioned key',
+                     { bucketName, objKey, error: err, method: '_deleteNullVersionMD' });
+            return cb(err);
+        }
+        return cb();
```
tmacro:
```suggestion
        }
        return cb(err);
```

tmacro (approved):

jonathan-gramain:
ping

PR #5066, opened 2023-03-16
feat: CLDSRV-358 preprocessingVersioningDelete() update for null keys

jonathan-gramain:
- Add support for null keys in the preprocessingVersioningDelete() helper, essentially, set a 'isNull' boolean param that gets passed to Metadata, which tells whether the version to delete is a null version.
    
  NOTE: The null version compatibility mode is still enforced for now until all pieces are updated to make functional tests pass without the compatibility mode.

- [test] fix error code checking
    
  In functional tests of 'objectDelete', an "afterAll" cleanup can crash because it checks an error code before checking if there's an error object.
    
  It does not crash in normal circumstances because there is an actual error due to the last unit test trying to clean the bucket, but if anything changes in the unit tests that leaves the bucket existing would have triggered this issue

- [cleanup] objectDelete: remove unused assignment

  Remove unused assignment of 'deleteInfo.isNull'

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -376,11 +376,22 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
     }
     if (bucketMD.getVersioningConfiguration() && reqVersionId) {
         if (reqVersionId === 'null') {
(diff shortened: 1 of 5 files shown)


watsaqat (approved):

dvasilas (approved):

jonathan-gramain:
ping

PR #5069, opened 2023-03-17
feat: CLDSRV-355 activate null keys behavior

jonathan-gramain:
Activate the use of null keys in place of null versioned keys by Cloudserver:

- allow processVersioningState() and preprocessingVersioningDelete() helpers to return the associated fields for null key handling, which tells Cloudserver to set its behavior to create/delete null keys, via sending PUT/DELETE requests with `versionId="null"` to the Metadata backend

- pass 'isNull' parameter in version-specific requests to hint the Metadata backend on what to do (most useful for V1 backend, but also to hint V0 backend that it should handle null keys appropriately)

- set "isNull2" metadata attribute when writing a null master, for optimization purpose (allows to avoid checking the null versioned key on update)

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -62,9 +62,24 @@ function checkQueryVersionId(query) {
     return undefined;
 }
 
-function _storeNullVersionMD(bucketName, objKey, nullVersionId, nullVersionMD, log, cb) {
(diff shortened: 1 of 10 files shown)


tmacro (approved):

jonathan-gramain:
ping

watsaqat (approved):

jonathan-gramain:
ping

jonathan-gramain:
ping

PR #5070, opened 2023-03-18
feat: CLDSRV-359 pass getDeleteMarker flag to metadata when needed

jonathan-gramain:
Pass the `getDeleteMarker` flag to the Metadata backend when the Cloudserver operation requires to distinguish if the target is a delete marker or does not exist, to set response header "x-amz-delete-marker" or return a specific error code.

### DIFF
--- a/lib/api/objectDeleteTagging.js
+++ b/lib/api/objectDeleteTagging.js
@@ -42,8 +42,9 @@ function objectDeleteTagging(authInfo, request, log, callback) {
         authInfo,
         bucketName,
         objectKey,
-        requestType: 'objectDeleteTagging',
         versionId: reqVersionId,
+        getDeleteMarker: true,
+        requestType: 'objectDeleteTagging',
         request,
     };
 
@@ -65,6 +66,8 @@ function objectDeleteTagging(authInfo, request, log, callback) {
               if (objectMD.isDeleteMarker) {
                   log.trace('version is a delete marker',
                   { method: 'objectDeleteTagging' });
+                  // FIXME we should return a `x-amz-delete-marker: true` header,
+                  // see S3C-7592
                   return next(errors.MethodNotAllowed, bucket);
               }
               return next(null, bucket, objectMD);
--- a/lib/api/objectGetLegalHold.js
+++ b/lib/api/objectGetLegalHold.js
@@ -33,12 +33,14 @@ function objectGetLegalHold(authInfo, request, log, callback) {
     }
     const versionId = decodedVidResult;
 
+    // FIXME pass 'getDeleteMarker: true' option to set
+    // 'x-amz-delete-marker' header (see S3C-7592)
     const metadataValParams = {
         authInfo,
         bucketName,
         objectKey,
-        requestType: 'objectGetLegalHold',
         versionId,
+        requestType: 'objectGetLegalHold',
         request,
     };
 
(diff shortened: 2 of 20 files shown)


lib/api/objectDeleteTagging.js
```diff
@@ -65,6 +66,8 @@ function objectDeleteTagging(authInfo, request, log, callback) {
               if (objectMD.isDeleteMarker) {
                   log.trace('version is a delete marker',
                   { method: 'objectDeleteTagging' });
+                  // FIXME wee should return a `x-amz-delete-marker: true` header,
```
tmacro:
```suggestion
                  // FIXME we should return a `x-amz-delete-marker: true` header,
```

lib/api/objectPutACL.js
```diff
@@ -124,6 +125,8 @@ function objectPutACL(authInfo, request, log, cb) {
                     if (objectMD.isDeleteMarker) {
                         log.trace('delete marker detected',
                         { method: 'objectPutACL' });
+                        // FIXME wee should return a `x-amz-delete-marker: true` header,
```
tmacro:
```suggestion
                        // FIXME we should return a `x-amz-delete-marker: true` header,
```

lib/api/objectPutLegalHold.js
```diff
@@ -64,6 +65,8 @@ function objectPutLegalHold(authInfo, request, log, callback) {
             if (objectMD.isDeleteMarker) {
                 log.trace('version is a delete marker',
                     { method: 'objectPutLegalHold' });
+                // FIXME wee should return a `x-amz-delete-marker: true` header,
```
tmacro:
```suggestion
                // FIXME we should return a `x-amz-delete-marker: true` header,
```

lib/api/objectPutRetention.js
```diff
@@ -65,6 +66,8 @@ function objectPutRetention(authInfo, request, log, callback) {
             if (objectMD.isDeleteMarker) {
                 log.trace('version is a delete marker',
                     { method: 'objectPutRetention' });
+                // FIXME wee should return a `x-amz-delete-marker: true` header,
```
tmacro:
```suggestion
                // FIXME we should return a `x-amz-delete-marker: true` header,
```

lib/api/objectPutTagging.js
```diff
@@ -66,6 +67,8 @@ function objectPutTagging(authInfo, request, log, callback) {
               if (objectMD.isDeleteMarker) {
                   log.trace('version is a delete marker',
                   { method: 'objectPutTagging' });
+                  // FIXME wee should return a `x-amz-delete-marker: true` header,
```
tmacro:
```suggestion
                  // FIXME we should return a `x-amz-delete-marker: true` header,
```

tmacro (approved):

lib/api/objectGetLegalHold.js
```diff
@@ -61,10 +63,14 @@ function objectGetLegalHold(authInfo, request, log, callback) {
                     if (versionId) {
                         log.trace('requested version is delete marker',
                             { method: 'objectGetLegalHold' });
+                        // FIXME wee should return a `x-amz-delete-marker: true` header,
```
watsaqat: should this be TODOs so they are more easily searchable? also wee -> we on a few of these

tests/functional/aws-node-sdk/test/versioning/objectGet.js
```diff
@@ -137,9 +137,11 @@ describe('get behavior on versioning-enabled bucket', () => {
                 s3.getObject({
                     Bucket: bucket,
                     Key: key,
-                    VersionId: this.test.deleteVersionId },
-                err => {
+                    VersionId: this.test.deleteVersionId,
+                }, function test1(err) {
```
watsaqat, line 141: why not more descriptive names instead of test1?

lib/api/objectGetLegalHold.js
jonathan-gramain: since it's an actual bug, I prefer FIXME over TODO (why TODO is more easily searchable btw?)

Yeah I will fix the typo

jonathan-gramain:
ping

watsaqat (approved):

tests/functional/aws-node-sdk/test/versioning/objectGet.js
jonathan-gramain: I would have used anonymous functions but I need actual functions to use `this.httpResponse`, and the linter expects names for every function. I agree the names are not very helpful and could be improved, but here the functions are just to fill in each test case, which is already described in the `it` description, so I don't think it really matters as the name wouldn't bring any value in helping understanding IMHO.

PR #5072, opened 2023-03-20
bugfix: CLDSRV-361 fix exception with batch delete of null version

jonathan-gramain:
Fix an exception in Cloudserver when doing a batch delete containing a deletion of a null version that does not exist on a versioned object

### DIFF
--- a/lib/metadata/metadataUtils.js
+++ b/lib/metadata/metadataUtils.js
@@ -26,7 +26,13 @@ function getNullVersion(objMD, bucketName, objectKey, log, cb) {
         // the latest version is not the null version, but null version exists
         log.debug('null version exists, get the null version');
         options.versionId = objMD.nullVersionId;
-        return metadata.getObjectMD(bucketName, objectKey, options, log, cb);
+        return metadata.getObjectMD(bucketName, objectKey, options, log, (err, nullVersionMD) => {
+            if (err && err.is.NoSuchKey) {
+                log.debug('object does not exist in metadata');
+                return cb();
+            }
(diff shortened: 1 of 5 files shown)


jonathan-gramain:
ping

lib/metadata/metadataUtils.js
```diff
@@ -26,7 +26,13 @@ function getNullVersion(objMD, bucketName, objectKey, log, cb) {
         // the latest version is not the null version, but null version exists
         log.debug('null version exists, get the null version');
         options.versionId = objMD.nullVersionId;
-        return metadata.getObjectMD(bucketName, objectKey, options, log, cb);
+        return metadata.getObjectMD(bucketName, objectKey, options, log, (err, nullVersionMD) => {
+            if (err && err.is.NoSuchKey) {
```
tmacro: Not sure it its possible for it to be a non-Arsenal error, but better safe than sorry.
```suggestion
            if (err && err.is?.NoSuchKey) {
```

lib/metadata/metadataUtils.js
```diff
@@ -96,6 +102,10 @@ function metadataGetObject(bucketName, objectKey, versionId, log, cb) {
     };
     return metadata.getObjectMD(bucketName, objectKey, options, log,
         (err, objMD) => {
+            if (err && err.is.NoSuchKey) {
```
tmacro:
```suggestion
            if (err && err.is?.NoSuchKey) {
```

KazToozs (approved):
LGTM

tmacro (approved):

PR #5075, opened 2023-03-21
CLDSRV-362 - Throw error when copying a cold or transitioning object

welansari:
A cold object should not be allowed to get copied as the data is not accessible.

Same issue happens when copying an object that is transitioning, the data might get deleted while copying is still in progress.

Issue: CLDSRV-362

### DIFF
--- a/lib/api/apiUtils/object/coldStorage.js
+++ b/lib/api/apiUtils/object/coldStorage.js
@@ -32,7 +32,6 @@ function getAmzRestoreResHeader(objMD) {
     return undefined;
 }
 
-
 /**
  * Check if restore can be done.
  *
@@ -198,9 +197,34 @@ function startRestore(objectMD, restoreParam, log, cb) {
     return cb(null, isObjectAlreadyRestored(objectMD, log));
 }
 
+/**
+ * checks if object data is available or if it's in cold storage
+ * @param {ObjectMD} objMD Object metadata
+ * @returns {ArsenalError|null} error if object data is not available
+ */
+function verifyColdObjectAvailable(objMD) {
+    // return error when object is cold
+    if (objMD.archive &&
+        // Object is in cold backend
+        (!objMD.archive.restoreRequestedAt ||
+        // Object is being restored
+        (objMD.archive.restoreRequestedAt && !objMD.archive.restoreCompletedAt))) {
+        const err = errors.InvalidObjectState
+            .customizeDescription('The operation is not valid for the object\'s storage class');
+        return err;
+    }
+    // return error when object is transitioning to a second location
+    if (objMD['x-amz-scal-transition-in-progress']) {
+        const err = errors.InvalidObjectState
+            .customizeDescription('The operation is not valid on a transitioning object');
+        return err;
+    }
+    return null;
+}
 
 module.exports = {
     startRestore,
     getAmzRestoreResHeader,
     validatePutVersionId,
+    verifyColdObjectAvailable,
 };
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -24,6 +24,7 @@ const monitoring = require('../utilities/monitoringHandler');
 const applyZenkoUserMD = require('./apiUtils/object/applyZenkoUserMD');
 const { getObjectSSEConfiguration } = require('./apiUtils/bucket/bucketEncryption');
 const { setExpirationHeaders } = require('./apiUtils/object/expirationHeaders');
+const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
 
 const versionIdUtils = versioning.VersionID;
 const locationHeader = constants.objectLocationConstraintHeader;
@@ -289,6 +290,11 @@ function objectCopy(authInfo, request, sourceBucket,
                             errors.NoSuchKey;
                         log.debug('no source object', { sourceObject });
                         return next(err, null, destBucketMD);
+                    }
+                    // check if object data is in a cold storage
+                    const coldErr = verifyColdObjectAvailable(sourceObjMD);
+                    if (coldErr) {
+                        return next(coldErr, null);
                     }
                     if (sourceObjMD.isDeleteMarker) {
                         log.debug('delete marker on source object',
--- a/lib/api/objectPutCopyPart.js
+++ b/lib/api/objectPutCopyPart.js
@@ -14,6 +14,7 @@ const services = require('../services');
 const setUpCopyLocator = require('./apiUtils/object/setUpCopyLocator');
 const { metadataValidateBucketAndObj } = require('../metadata/metadataUtils');
 const monitoring = require('../utilities/monitoringHandler');
+const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
 
 const versionIdUtils = versioning.VersionID;
 
@@ -132,6 +133,11 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                         sourceObjMD.location[0].dataStoreName) {
                         sourceLocationConstraintName =
                             sourceObjMD.location[0].dataStoreName;
+                    }
+                    // check if object data is in a cold storage
+                    const coldErr = verifyColdObjectAvailable(sourceObjMD);
+                    if (coldErr) {
+                        return next(coldErr, null);
                     }
                     if (sourceObjMD.isDeleteMarker) {
                         log.debug('delete marker on source object',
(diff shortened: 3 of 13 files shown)


lib/api/objectCopy.js
```diff
@@ -290,6 +292,18 @@ function objectCopy(authInfo, request, sourceBucket,
                         log.debug('no source object', { sourceObject });
                         return next(err, null, destBucketMD);
                     }
+                    // return error when object is cold
+                    if (locationConstraints[sourceObjMD.dataStoreName]?.isCold) {
```
francoisferrand: that check may soon be incorrect...

at the moment, indeed the `dataStoreName` is updated when the object is restored (so it will not 'appear' as cold), but this causes other issues : I think there is WIP to avoid changing the data store when restoring an object, so that we keep track of underlying cold location anyway.... and that check would thus fail for cold location.

Not sure when that is expected though, maybe @nicolas2bert can confirm if this will conflict and/or if there is a better to check for "object is cold AND not restored"

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -1234,6 +1235,42 @@ describe('Object Copy', () => {
...
+                    Key: destObjName,
+                    CopySource: `${sourceBucketName}/${sourceObjName}`,
+                }, err => {
+                        assert.strictEqual(err.code, 'InvalidObjectState');
+                        assert.strictEqual(err.statusCode, 403);
+                        done();
+                    });
+            });
+        });
+
```
francoisferrand, line 1259: need another test, to ensure we can indeed copy a "cold & restored" object
(may be the same check for the moment, but that may not always be true...)

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -710,6 +711,72 @@ describe('Object Part Copy', () => {
...
+                    UploadId: uploadId,
+                }, err => {
+                        assert.strictEqual(err.code, 'InvalidObjectState');
+                        assert.strictEqual(err.statusCode, 403);
+                        done();
+                    });
+            });
+        });
+
+        it('should copy restored object', done => {
```
francoisferrand: nit: "should copy part from restored object"

lib/api/objectPutCopyPart.js
```diff
@@ -133,6 +133,22 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                         sourceLocationConstraintName =
                             sourceObjMD.location[0].dataStoreName;
                     }
+                    // return error when object is cold
+                    if (sourceObjMD.archive &&
+                        // Object is in cold backend
+                        (!sourceObjMD.archive.restoreRequestedAt ||
+                        // Object is being restored
+                        (sourceObjMD.archive.restoreRequestedAt && !sourceObjMD.archive.restoreCompletedAt))) {
```
francoisferrand: should we introduce a function for this? (e.g. `isObjecDatatInColdStorage`)

or possibly a helper which verifies both that data is present and object is not being transitioned and returns the appropriate error? (e.g. `VerifyColdObjectAvailable`) ?

lib/api/objectPutCopyPart.js
```diff
@@ -133,6 +133,22 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
...
+                        const err = errors.InvalidObjectState
+                            .customizeDescription('The operation is not valid for the object\'s storage class');
+                        return next(err, null);
+                    }
+                    // return error when object is transitioning to a second location
+                    if (sourceObjMD['x-amz-scal-transition-in-progress']) {
+                        const err = errors.InvalidObjectState
+                        .customizeDescription('The operation is not valid on a transitioning object');
```
francoisferrand: nit: missing indent

lib/api/objectPutCopyPart.js
```diff
@@ -133,6 +134,11 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                         sourceLocationConstraintName =
                             sourceObjMD.location[0].dataStoreName;
                     }
+                    // check if object data is in a cold storage
+                    const coldErr = verifyColdObjectAvailable(sourceObjMD);
+                    if (coldErr) {
+                        return callback(coldErr, null);
```
francoisferrand: shouldn't it be `next(coldErr, null)` ?

welansari: it should

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -710,6 +711,72 @@ describe('Object Part Copy', () => {
                 });
             });
 
+
```
francoisferrand: nit: extra empty line

tests/unit/api/objectGet.js
```diff
@@ -423,6 +423,24 @@ describe('objectGet API', () => {
...
+        mdColdHelper.putBucketMock(bucketName, null, () => {
+            mdColdHelper.putObjectMock(bucketName, objectName, mdColdHelper.getTransitionInProgressMD(), () => {
+                objectGet(authInfo, testGetRequest, false, log, err => {
+                    assert.strictEqual(err.is.InvalidObjectState, true);
+                    done();
+                });
+            });
+        });
+    });
+
```
francoisferrand, line 443: should test also the other cases? (cold, restored)

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -710,6 +711,72 @@ describe('Object Part Copy', () => {
...
+        it('should not copy a cold object', done => {
```
francoisferrand: copy part of

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -710,6 +711,72 @@ describe('Object Part Copy', () => {
...
+                    UploadId: uploadId,
+                }, err => {
+                        assert.strictEqual(err.code, 'InvalidObjectState');
+                        assert.strictEqual(err.statusCode, 403);
+                        done();
+                    });
+            });
+        });
+
+        it('should not copy an object when it\'s transitioning to cold', done => {
```
francoisferrand: copy part of

tests/unit/api/objectGet.js
welansari: test for those cases already exist

francoisferrand (approved):
LGTM

KillianG (approved):

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -198,9 +197,34 @@ function startRestore(objectMD, restoreParam, log, cb) {
...
+ * @param {ObjectMD} objMD Object metadata
+ * @returns {ArsenalError|null} error if object data is not available
+ */
+function verifyColdObjectAvailable(objMD) {
+    // return error when object is cold
+    if (objMD.archive &&
+        // Object is in cold backend
+        (!objMD.archive.restoreRequestedAt ||
+        // Object is being restored
+        (objMD.archive.restoreRequestedAt && !objMD.archive.restoreCompletedAt))) {
```
KazToozs, line 211: For my understanding, why is the existence of `objMD.archive` alone not enough to throw error? Is it maintained as an attribute on the object once a restore is successful?

lib/api/objectCopy.js
```diff
@@ -290,6 +291,11 @@ function objectCopy(authInfo, request, sourceBucket,
                         log.debug('no source object', { sourceObject });
                         return next(err, null, destBucketMD);
                     }
+                    // check if object data is in a cold storage
+                    const coldErr = verifyColdObjectAvailable(sourceObjMD);
+                    if (coldErr) {
```
KazToozs, line 296: Nit: is it necessary/good practice to create a variable? Ignore if so
```suggestion
                    if (verifyColdObjectAvailable(sourceObjMD)) {
```

KazToozs: same applies for some areas below

KazToozs (commented):
LGTM otherwise

lib/api/apiUtils/object/coldStorage.js
welansari: When an object is restored `objMD.archive` is still present in the metadata, and points to the data on the cold side.
Restoration is a temporary state that eventually expires, that's why we need to keep it.

lib/api/objectCopy.js
welansari: the function returns an error so we have to store its value somewhere to return it

KazToozs (approved):

PR #5081, opened 2023-03-27
CLDSRV-366 Clear list orphan delete markers response

nicolas2bert:
Etag, Size, StorageClass, and DataStoreName are worthless when listing orphan delete markers because:

- A delete marker does not hold any data.
- The listing of orphan delete markers is only used for expiration purposes.


alexanderchan-scality (approved):

naren-rajendran (approved):

nicolas2bert:
@bert-e approve

PR #5086, opened 2023-03-28
bugfix: CLDSRV-365 fix legal hold can be deleted issue and add more t…

XinLi0207:
the previous checking ifCanModifyObject didn't take into account legalhold case, so caused the issue that object with legalhold can still be deleted with governance bypass.
the standard is whenever the object is legalhold enabled, in any case it should not be allowed to be modified(even with governance mode and bypass header)

tested working after doing a upgrade check:
- in governance mode
  - delete no
  - multi-deletes no
  - delete with bypass yes
  - multi-deletes with bypass yes
  - delete with legalhold with bypass no
  - multi-deletes with legalhold with by pass no

- in conpliance mode
  - delete no
  - multi-deletes no
  - delete with bypass no
  - multi-deletes with bypass no
  - delete with legalhold with bypass no
  - multi-deletes with legalhold with by pass no

### DIFF
--- a/lib/api/apiUtils/object/objectLockHelpers.js
+++ b/lib/api/apiUtils/object/objectLockHelpers.js
@@ -202,7 +202,13 @@ class ObjectLockInfo {
      * @returns {bool} - True if the retention policy allows the objects data to be modified (overwritten/deleted)
      */
     canModifyObject(hasGovernanceBypass) {
-        return !this.isLocked() || (this.isGovernanceMode() && !!hasGovernanceBypass);
+        // can modify object if object is not locked
+        // cannot modify object in any cases if legal hold is enabled
+        // if no legal hold, can only modify object if bypassing governance when locked
+        if (!this.isLocked()) {
+            return true;
+        }
+        return !this.legalHold && this.isGovernanceMode() && !!hasGovernanceBypass;
     }
 
     /**
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3",
-  "version": "7.10.24",
+  "version": "7.10.27",
   "description": "S3 connector",
   "main": "index.js",
   "engines": {
--- a/tests/functional/aws-node-sdk/test/object/deleteObject.js
+++ b/tests/functional/aws-node-sdk/test/object/deleteObject.js
@@ -5,7 +5,6 @@ const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
 const changeObjectLock = require('../../../../utilities/objectLock-util');
 
-const bucketName = 'testdeletempu';
 const objectName = 'key';
 const objectNameTwo = 'secondkey';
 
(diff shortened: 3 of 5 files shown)


lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -202,7 +202,10 @@ class ObjectLockInfo {
      * @returns {bool} - True if the retention policy allows the objects data to be modified (overwritten/deleted)
      */
     canModifyObject(hasGovernanceBypass) {
-        return !this.isLocked() || (this.isGovernanceMode() && !!hasGovernanceBypass);
+        // can modify object if object is not locked
+        // cannot modify object in any cases if legal hold is enabled
+        // if not legal hold, can only modify object if bypass governance when locked
+        return !this.legalHold && (!this.isLocked() || (this.isGovernanceMode() && !!hasGovernanceBypass));
```
FCScality: isLocked function already checks the legalHold
. was wondering why keep it in the second part of the condition ?

XinLi0207: yeah isLockes contains legalhold check but if we remove the second part then `!this.legalHold && (this.isGovernanceMode() && !!hasGovernanceBypass)`  will allow any compliance or governance mode without legalhold pass which should not be allowed

XinLi0207:
ping

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -202,7 +202,10 @@ class ObjectLockInfo {
      * @returns {bool} - True if the retention policy allows the objects data to be modified (overwritten/deleted)
      */
     canModifyObject(hasGovernanceBypass) {
-        return !this.isLocked() || (this.isGovernanceMode() && !!hasGovernanceBypass);
+        // can modify object if object is not locked
+        // cannot modify object in any cases if legal hold is enabled
+        // if not legal hold, can only modify object if bypass governance when locked
```
KazToozs:
```suggestion
        // if no legal hold, can only modify object if bypassing governance when locked
```

For clarity

KazToozs (approved):
lgtm otherwise

ghost (approved):
LGTM

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -202,7 +202,10 @@ class ObjectLockInfo {
...
+        return !this.legalHold && (!this.isLocked() || (this.isGovernanceMode() && !!hasGovernanceBypass));
     }
```
tmacro: I think we should break this up as the return is getting a bit long.
```suggestion
        if (!this.isLocked()) {
            return true;
        }
        return !this.legalHold && this.isGovernanceMode() && !!hasGovernanceBypass;
    }
```

tmacro (commented):
Small comment on style. Logic and test cases lgtm.

PR #5087, opened 2023-03-28
Add origin op to all delete object calls

KillianG:
TODO: Bump arsenal when ARSN-324 is merged and arsenal is released

Add origin op to all calls to internal delete object to raise the right event instead of always `s3:ObjectRemoved:Delete`
Add new route to backbeat routes

Issue: CLDSRV-367


KazToozs (approved):

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -53,13 +53,14 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
...
  * @param {function} callback - callback function
+ * @param {string} [originOp=s3:ObjectRemoved:DeleteMarkerCreated] - origin operation
  * @return {undefined} and call callback with (err, result) -
  * result.contentMD5 - content md5 of new object or version
  * result.versionId - unencrypted versionId returned by metadata
  */
 function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         canonicalID, cipherBundle, request, isDeleteMarker, streamingV4Params,
-        log, callback) {
+        log, callback, originOp = 's3:ObjectRemoved:DeleteMarkerCreated') {
```
jonathan-gramain: From an API/doc point of view this looks odd, the `createAndStoreObject` origin op must be a PUT or something per se. I understand the need to pass it for the delete marker case but it should be named differently when passed as an option I think to make it clearer, maybe `deleteMarkerOriginOp`?

KillianG: https://github.com/scality/cloudserver/pull/5087/files#diff-5900fcd2b08c9d289eafb9136636f13e87709614176c5dc541b08039c5687fccL159

As you can see here, it was already `s3:ObjectRemoved:DeleteMarkerCreated` without any possibility of configuring it... I agree that this is a bit odd but this is probably out of scope of this ticket to change that here..

KillianG: I updated that to use a boolean as parameter instead of an hardcoded originOp, might be simpler now

welansari (approved):

lib/api/apiUtils/object/createAndStoreObject.js
francoisferrand: i fear this is still hackish... The issue is that `createAndStoreObject` can both create object (and set originOp automatically to `s3:ObjectCreated:Put`) or create delete marker, and this flag affects only the second case.

IMHO it seems the clean way would be to pass originOp indeed (with no default value), use it in both cases (both when creating delete marker and when creating version) and modify the 4 calls to specify the expected `originOp` everytime

lib/api/objectDelete.js
```diff
@@ -24,9 +24,10 @@ const objectLockedError = new Error('object locked');
  *                           includes normalized headers
  * @param {Logger} log - werelogs request instance
  * @param {function} cb - final cb to call with the result and response headers
+ * @param {boolean} [isExpiration=false] - true if the call comes from LifecycleExpiration
  * @return {undefined}
  */
-function objectDelete(authInfo, request, log, cb) {
+function objectDelete(authInfo, request, log, cb, isExpiration = false) {
```
francoisferrand: nit: callback should be then last parameter

lib/api/objectDelete.js
```diff
@@ -164,7 +165,9 @@ function objectDelete(authInfo, request, log, cb) {
                 // source does not have versioning.
                 return createAndStoreObject(bucketName, bucketMD, objectKey,
                     objectMD, authInfo, canonicalID, null, request, true, null,
-                    log, err => {
+                    log, isExpiration ?
+                    's3:LifecycleExpiration:DeleteMarkerCreated' :
+                    's3:ObjectRemoved:DeleteMarkerCreated', err => {
```
francoisferrand: nit: bad indent, this is not readable...
either you need to extra indent the ternary expressions, or introduce a variable to avoid the issue

```suggestion
                    log, isExpiration ?
                        's3:LifecycleExpiration:DeleteMarkerCreated' :
                        's3:ObjectRemoved:DeleteMarkerCreated',
                     err => {
```

lib/api/objectDelete.js
```diff
@@ -176,7 +179,7 @@ function objectDelete(authInfo, request, log, cb) {
                         return services.deleteObject(bucketName, objectMD,
                             objectKey, delOptions, log, (err, delResult) =>
                             next(err, bucketMD, objectMD, delResult,
-                                deleteInfo));
+                                deleteInfo), isExpiration);
```
francoisferrand: i am wondering about this: since `services.deleteObject()` takes a bool, this looks somewhat inconsistent between the different "branches" of this function : when doing a delete, we delegate the computation of originOp to lower layer (delete case) or do it here (delete marker).

I guess some consistency would be better, but it means either reverting to the previous code or digging deeper here...

lib/api/objectDelete.js
```diff
@@ -193,13 +196,15 @@ function objectDelete(authInfo, request, log, cb) {
...
+                    objectMD, delResult, deleteInfo), isExpiration);
             }
             // putting a new delete marker
             deleteInfo.newDeleteMarker = true;
             return createAndStoreObject(bucketName, bucketMD,
                 objectKey, objectMD, authInfo, canonicalID, null, request,
-                deleteInfo.newDeleteMarker, null, log, (err, newDelMarkerRes) => {
+                deleteInfo.newDeleteMarker, null, log, isExpiration ?
+                's3:LifecycleExpiration:DeleteMarkerCreated' :
+                's3:ObjectRemoved:DeleteMarkerCreated', (err, newDelMarkerRes) => {
```
francoisferrand, line 215: nit: same indent

francoisferrand (approved):

lib/api/objectDelete.js
```diff
@@ -176,9 +180,11 @@ function objectDelete(authInfo, request, log, cb) {
                             deleteInfo.removeDeleteMarker = true;
                         }
                         return services.deleteObject(bucketName, objectMD,
-                            objectKey, delOptions, log, (err, delResult) =>
-                            next(err, bucketMD, objectMD, delResult,
-                                deleteInfo));
+                            objectKey, delOptions, log, isExpiration ?
+                                's3:LifecycleExpiration:DeleteMarkerCreated' :
+                                's3:ObjectRemoved:DeleteMarkerCreated',
```
francoisferrand, line 185: This happens when we are _removing_ a delete marker : should this not be a `s3:ObjectRemoved:Delete` event?
(or should this be done a smarter way, to avoid generating 2 notifs?)

lib/api/objectDelete.js
```diff
@@ -194,14 +200,19 @@ function objectDelete(authInfo, request, log, cb) {
                 }
 
                 return services.deleteObject(bucketName, objectMD, objectKey,
-                    delOptions, log, (err, delResult) => next(err, bucketMD,
-                    objectMD, delResult, deleteInfo));
+                    delOptions, log, isExpiration ?
+                        's3:LifecycleExpiration:DeleteMarkerCreated' :
+                        's3:ObjectRemoved:DeleteMarkerCreated',
```
francoisferrand, line 205: this is not a putting delete marker, this is either removing data or a delete marker...

lib/api/objectDelete.js
francoisferrand: solved in https://github.com/scality/cloudserver/pull/5135

lib/api/objectDelete.js
francoisferrand: solved in https://github.com/scality/cloudserver/pull/5135

PR #5088, opened 2023-03-28
impr: CLDSRV-369 new version-specific DELETE functional test

jonathan-gramain:
Add a new functional test in the aws-node-sdk test suite, that

- creates a nonversioned key

- then enables versioning and creates 5 more versioned keys

- then deletes the 5 versioned keys in a batch delete

- it expects the null version to be the only remaining version

Its primary purpose is to check version-specific DELETE with a null version in V1 metadata format (MD-847), but it is a good addition to the existing test suite also in V0 format.


jonathan-gramain:
ping

naren-rajendran (approved):

dvasilas (approved):

PR #5089, opened 2023-03-29
CLDSRV-368: backbeat route - add indexing APIs

alexanderchan-scality:
- CLDSRV-368: add indexing routes
- bump arsenal

# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -2,12 +2,14 @@ const url = require('url');
 const async = require('async');
 const httpProxy = require('http-proxy');
 const querystring = require('querystring');
+const joi = require('@hapi/joi');
 
 const backbeatProxy = httpProxy.createProxyServer({
     ignorePath: true,
 });
 const { auth, errors, s3middleware, s3routes, models, storage } =
     require('arsenal');
+
 const { responseJSONBody } = s3routes.routesUtils;
 const { getSubPartIds } = s3middleware.azureHelper.mpuUtils;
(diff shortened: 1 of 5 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -1117,6 +1122,77 @@ function listLifecycle(request, response, userInfo, log, cb) {
     });
 }
 
+const indexesForFeature = {
```
nicolas2bert: It would be a good idea to make our API more modular to accommodate future needs. 
While we may only need to create a specific index at the moment, there may be other types of indexes that we may need to create in the future. By making our API more modular, we can ensure that it will be more flexible and adaptable to changing requirements.

Some potential benefits of making our API more modular include:

**Scalability**: By building a modular API, we can easily add new functionality as our needs evolve, without having to completely rewrite our codebase.

**Reusability**: A modular API can be more easily reused in other applications or contexts, which can save time and effort in future projects.

**Flexibility**: With a more modular API, we can change and update specific modules without affecting the rest of our codebase, which can make it easier to maintain and debug.

Regarding the security concern, please note that only users who are authenticated and authorized will have the ability to create indexes. Currently, we rely on a single permission, s3:ReplicateObject, to authorize all Backbeat API calls. However, this is only a temporary measure and we plan to implement more granular permissions in the future.

rachedbenmustapha (approved):

naren-rajendran (approved):

PR #5108, opened 2023-04-03
impr: CLDSRV-373 new func test for S3C-5139

jonathan-gramain:
Add a new functional test in aws-node-sdk that reproduces the steps that formerly triggered S3C-5139, which was silently fixed since it was opened.

It is useful also for S3C-7352 because it covers a corner case for both old and new null version handling in v0 and v1 metadata formats.


jonathan-gramain:
ping

nicolas2bert (approved):

alexanderchan-scality (approved):

PR #5109, opened 2023-04-03
CLDSRV-372 Current lifecycle versions should include version id

nicolas2bert:
The current versions listed to be lifecycle should include version id if the bucket is versioned.

The version id is not used for Expiration.

This version id is used for Transitioning the current version: The “backbeat route putMetadata“ in Cloudserver specified the version id to update its location constraint once the object is transitioned.

### DIFF
--- a/lib/api/apiUtils/object/lifecycle.js
+++ b/lib/api/apiUtils/object/lifecycle.js
@@ -16,7 +16,7 @@ function _makeTags(tags) {
     return res;
 }
 
-function processCurrents(bucketName, listParams, list) {
+function processCurrents(bucketName, listParams, isBucketVersioned, list) {
     const data = {
         Name: bucketName,
         Prefix: listParams.prefix,
@@ -46,6 +46,15 @@ function processCurrents(bucketName, listParams, list) {
             DataStoreName: v.dataStoreName,
             ListType: CURRENT_TYPE,
         };
+
+        // NOTE: The current versions listed to be lifecycle should include version id
+        // if the bucket is versioned.
+        if (isBucketVersioned) {
+            const versionId = (v.IsNull || v.VersionId === undefined) ?
+                'null' : versionIdUtils.encode(v.VersionId);
+            content.VersionId = versionId;
+        }
+
         data.Contents.push(content);
     });
 
--- a/lib/api/backbeat/listLifecycleCurrents.js
+++ b/lib/api/backbeat/listLifecycleCurrents.js
@@ -8,10 +8,10 @@ const { processCurrents } = require('../apiUtils/object/lifecycle');
 
 
 function handleResult(listParams, requestMaxKeys, authInfo,
-    bucketName, list, log, callback) {
+    bucketName, list, isBucketVersioned, log, callback) {
     // eslint-disable-next-line no-param-reassign
     listParams.maxKeys = requestMaxKeys;
-    const res = processCurrents(bucketName, listParams, list);
+    const res = processCurrents(bucketName, listParams, isBucketVersioned, list);
 
     pushMetric('listLifecycleCurrents', log, { authInfo, bucket: bucketName });
     monitoring.promMetrics('GET', bucketName, '200', 'listLifecycleCurrents');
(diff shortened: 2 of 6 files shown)


lib/api/apiUtils/object/lifecycle.js
```diff
@@ -46,6 +46,15 @@ function processCurrents(bucketName, listParams, list) {
             DataStoreName: v.dataStoreName,
             ListType: CURRENT_TYPE,
         };
+
+        // NOTE: The current versions listed to be lifecycle should include version id
```
jonathan-gramain, line 50: I think the inclusion of `VersionId` should not depend on whether the bucket is versioned or not, but only whether the API call is supposed to return them. At least this is what the AWS API does. In other words, we could return `VersionId: 'null'` for all objects if the bucket is not versioned.

nicolas2bert: If the bucket is not versioned, AWS returns objects without VersionID.

(Here is one example where using a "null" version for a non-versioned object does not seem to work.
During lifecycle transition, the object transition processor does a `putMetadata(bucket, key, version id)` request to update the object metadata `dataStoreName` (data location name).
If the bucket is not versioned and the `putMetadata` call specifies the "null" version id, a new "null" version will be created instead of updating the master version.)

jonathan-gramain: > If the bucket is not versioned, AWS returns objects without VersionID.

It does :wink: 

```
jonathan@jonathan-XPS-13-9380:~/git/home/ironman/S3C-7352$ aws --profile scality s3 mb s3://jonathan-scality-nonversioned-bucket
make_bucket: jonathan-scality-nonversioned-bucket
jonathan@jonathan-XPS-13-9380:~/git/home/ironman/S3C-7352$ aws --profile scality s3 cp /etc/hosts s3://jonathan-scality-nonversioned-bucket/hosts1
upload: ../../../../../../etc/hosts to s3://jonathan-scality-nonversioned-bucket/hosts1
jonathan@jonathan-XPS-13-9380:~/git/home/ironman/S3C-7352$ aws --profile scality s3api list-object-versions --bucket jonathan-scality-nonversioned-bucket
{
    "Versions": [
        {
            "ETag": "\"e171fed98b302ddc8b64aa03a56b2af4\"",
            "Size": 376,
            "StorageClass": "STANDARD",
            "Key": "hosts1",
            "VersionId": "null",
            "IsLatest": true,
            "LastModified": "2023-04-04T21:48:21.000Z",
            "Owner": {
                "DisplayName": "ironman-mgr",
                "ID": "b9e4605d555d11651588cfec37789266a156211aeaa7173a5abb82229348c8d9"
            }
        }
    ]
}
```

> If the bucket is not versioned and the putMetadata call specifies the "null" version id, a new "null" version will be created instead of updating the master version.

I believe the correct way in our API to do this is to pass an empty version ID with `&versionId=`. So if it is the null version, write it with an empty version ID.

I think it will be slightly cleaner to follow this scheme, as it is more aligned with the standard API. Also don't forget that the null version may not be the latest version, so updating the master is not always the right thing to do when updating a null version (but it is correct when writing a new null version).

nicolas2bert: If the bucket is not versioned, BB Lifecycle Bucket Processor does a list-objects, and no VersionId is returned.
To be compatible with the old Bucket Processos implementation, I added this check.

Also, to remove this check we need to make sure Backbeat handles "null" version appropriately, https://scality.atlassian.net/browse/CLDSRV-374 
I will first focus on the CLDSRV-374  issue.

alexanderchan-scality (approved):

lib/api/backbeat/listLifecycleCurrents.js
```diff
@@ -70,7 +70,7 @@ function listLifecycleCurrents(authInfo, request, log, callback) {
                 IsTruncated: false,
             };
             return handleResult(listParams, requestMaxKeys, authInfo,
-                bucketName, emptyList, log, callback);
+                bucketName, emptyList, false, log, callback);
```
jonathan-gramain: The bucket itself could be versioned here, although we pass `false`, not sure this is wanted?

jonathan-gramain (approved):

lib/api/backbeat/listLifecycleCurrents.js
nicolas2bert: Although the listing is empty, I will still move the versioning check earlier and introduce a variable for the possibility of using it in future implementations.

nicolas2bert:
@bert-e approve

PR #5117, opened 2023-04-10
CLDSRV-375 Exclude already transitioned keys from lifecycle listings

nicolas2bert:
A new listing parameter called "excluded-data-store-name" is added to filter out keys that have already been transitioned. 

Note that orphan delete markers do not go through a transition workflow, so no filtering based on datastore name is applied to list orphan delete markers.

Note once https://github.com/scality/Arsenal/pull/2111 is merged, update the Arsenal package dependency.

### DIFF
--- a/lib/api/backbeat/listLifecycleCurrents.js
+++ b/lib/api/backbeat/listLifecycleCurrents.js
@@ -4,7 +4,7 @@ const services = require('../../services');
 const { metadataValidateBucket } = require('../../metadata/metadataUtils');
 const { pushMetric } = require('../../utapi/utilities');
 const monitoring = require('../../utilities/monitoringHandler');
-const { processCurrents } = require('../apiUtils/object/lifecycle');
+const { getLocationConstraintErrorMessage, processCurrents } = require('../apiUtils/object/lifecycle');
 
 
 function handleResult(listParams, requestMaxKeys, authInfo,
@@ -20,15 +20,16 @@ function handleResult(listParams, requestMaxKeys, authInfo,
 
 /**
  * listLifecycleCurrents - Return list of current versions/masters in bucket
- * @param  {AuthInfo} authInfo - Instance of AuthInfo class with
- *                               requester's info
- * @param  {object} request    - http request object
- * @param  {function} log      - Werelogs request logger
- * @param  {function} callback - callback to respond to http request
- *                               with either error code or xml response body
+ * @param {AuthInfo} authInfo            - Instance of AuthInfo class with
+ *                                          requester's info
+ * @param {array} locationConstraints    - array of location contraint
+ * @param {object} request               - http request object
+ * @param {function} log                 - Werelogs request logger
+ * @param {function} callback            - callback to respond to http request
+ *                                          with either error code or xml response body
  * @return {undefined}
  */
-function listLifecycleCurrents(authInfo, request, log, callback) {
+function listLifecycleCurrents(authInfo, locationConstraints, request, log, callback) {
     const params = request.query;
     const bucketName = request.bucketName;
 
@@ -37,10 +38,19 @@ function listLifecycleCurrents(authInfo, request, log, callback) {
         Number.parseInt(params['max-keys'], 10) : 1000;
     if (Number.isNaN(requestMaxKeys) || requestMaxKeys < 0) {
         monitoring.promMetrics(
-            'GET', bucketName, 400, 'listBucket');
+            'GET', bucketName, 400, 'listLifecycleCurrents');
         return callback(errors.InvalidArgument);
     }
     const actualMaxKeys = Math.min(constants.listingHardLimit, requestMaxKeys);
+
+    const excludedDataStoreName = params['excluded-data-store-name'];
+    if (excludedDataStoreName && !locationConstraints[excludedDataStoreName]) {
+        const errMsg = getLocationConstraintErrorMessage(excludedDataStoreName);
+        log.error(`locationConstraint is invalid - ${errMsg}`, { locationConstraint: excludedDataStoreName });
+        monitoring.promMetrics('GET', bucketName, 400, 'listLifecycleCurrents');
+
+        return callback(errors.InvalidLocationConstraint.customizeDescription(errMsg));
+    }
 
     const metadataValParams = {
         authInfo,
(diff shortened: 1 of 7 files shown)


lib/api/backbeat/listLifecycleOrphanDeleteMarkers.js
```diff
@@ -19,15 +19,16 @@ function handleResult(listParams, requestMaxKeys, authInfo,
...
+ *                                         requester's info
+ * @param {array} locationConstraints    - array of location contraint
+ * @param {object} request               - http request object
+ * @param {function} log                 - Werelogs request logger
+ * @param {function} callback            - callback to respond to http request
+ *                                          with either error code or xml response body
  * @return {undefined}
  */
-function listLifecycleOrphanDeleteMarkers(authInfo, request, log, callback) {
+function listLifecycleOrphanDeleteMarkers(authInfo, locationConstraints, request, log, callback) {
```
jonathan-gramain, line 31: I believe we need to pass `locationConstraints` in the `listParams` here as well.

tests/functional/backbeat/excludedDataStoreName.js
```diff
@@ -0,0 +1,283 @@
...
+            assert(!data.NextMarker);
+            assert.strictEqual(data.MaxKeys, 1000);
+
+            const contents = data.Contents;
+            assert.strictEqual(contents.length, 2);
+
+            assert.strictEqual(contents[0].Key, 'key0');
+            assert.strictEqual(contents[0].DataStoreName, location1);
+            assert.strictEqual(contents[1].Key, 'key2');
+            assert.strictEqual(contents[1].DataStoreName, location1);
```
jonathan-gramain, line 122:
```suggestion
            assert.deepStrictEqual(data.Contents, [
                { Key: 'key0', DataStoreName: location1 },
                { Key: 'key2', DataStoreName: location1 }
            ]);
```

jonathan-gramain (changes requested):

package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.12.0",
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.92",
+    "arsenal": "git+https://github.com/scality/arsenal#improvement/ARSN-328/excludedDataStoreName",
```
nicolas2bert: TODO: need to be updated before merging.

lib/api/backbeat/listLifecycleOrphanDeleteMarkers.js
nicolas2bert: Good catch. I am wondering how the functional "listing orphan delete markers" tests did not catch this. We might want to give it a quick look. Thanks.

nicolas2bert: I don't think so. What we pass in the `listParams`  is the `excludedDataStoreName` property that will be used to exclude keys that have already been transitioned. 
It is not necessary since orphan delete markers are not transitioned.

lib/api/backbeat/listLifecycleCurrents.js
```diff
@@ -37,11 +38,20 @@ function listLifecycleCurrents(authInfo, request, log, callback) {
...
     const actualMaxKeys = Math.min(constants.listingHardLimit, requestMaxKeys);
 
+    const excludedDataStoreName = params['excluded-data-store-name'];
+    if (excludedDataStoreName && !locationConstraints[excludedDataStoreName]) {
+        const errMsg = getLocationConstraintErrorMessage(excludedDataStoreName);
+        log.error(`locationConstraint is invalid - ${errMsg}`, { locationConstraint: excludedDataStoreName });
+        monitoring.promMetrics('GET', bucketName, 400, 'listLifecycleCurrents');
+
+        return callback(errors.InvalidLocationConstraint.customizeDescription(errMsg));
+    }
```
jonathan-gramain, line 53: This check seems to be duplicated in each listing backend that uses it, why not instead validating in `listLifecycle`, before calling the backend function? This would also avoid having to pass the `locationConstraints` object around.

lib/api/backbeat/listLifecycleOrphanDeleteMarkers.js
jonathan-gramain: OK, makes sense, it would be useful to describe in the JSDoc of each listing backend function which params are supported and what they are for, to avoid surprises later if we add new params etc., otherwise one may assume that all backend functions support the same set of params, which is not the case here.

jonathan-gramain (approved):
Left some suggestions to improve a few things, not blocking the PR to be merged if in a time crush but ideally would be addressed before merging.

lib/api/backbeat/listLifecycleCurrents.js
nicolas2bert: List orphans delete markers don't use it.

nicolas2bert: The check is utilized by both the 'listLifecycleCurrents' and 'listLifecycleNonCurrents' API calls, but not by the 'listLifecycleDeleteMarkers' API call. To clarify that the check applies only to the former two calls, I moved it to the 'API level'.

jonathan-gramain: Well, still not very clear to me ;)

I think for separation of concerns, it's better if the check on the locations happens from the API caller. It's ok if it is also checked for the "listLifecycleDeleteMarkers" case. Then, we pass the excluded (and validated) location to the backend listing, and it's up to the listing to decide whether to implement this exclusion mechanism or not (and document appropriately in JSDoc).

I believe it works as is but I find it a bit confusing and less easily maintainable.

alexanderchan-scality (approved):

lib/api/backbeat/listLifecycleCurrents.js
nicolas2bert: For "listLifecycleDeleteMarkers" case, I think we should check that `params['excluded-data-store-name']` is `undefined` instead and we should return a different error than "getLocationConstraintErrorMessage" if `params['excluded-data-store-name']` is defined.

I have not implemented this check today since it is an internal API but I will create a ticket for it.

PR #5153, opened 2023-05-03
CLDSRC-370: Release dev docker image

dvasilas:
In Scuba we need: 
- To use the cloudserver's "dev" docker image for pre-merge tests. This adds the publishing of this image in the release workflow so that we can use it with a release tag. 
- To specify the `bucketd.bootstrap` config parameter when using the dev docker image

Also contains a fix for a failure to setup virtualenv in the CI ( see: https://github.com/scality/cloudserver/actions/runs/4873146866/jobs/8692634059 )
The fix is to pin virtualenv to the last version that worked.

Note: please review the integration branches. I have added a commit for the virtualenv issue ( https://github.com/scality/cloudserver/commit/6952b915391aa7d89afa3d36afca22a216e13968 )

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,6 +21,18 @@ jobs:
       file: images/svc-base/Dockerfile
       tag: ${{ github.event.inputs.tag }}-svc-base
 
+  build-dev-image:
+    uses: scality/workflows/.github/workflows/docker-build.yaml@v1
+    secrets: inherit
+    with:
+      push: true
+      registry: registry.scality.com
+      namespace: ${{ github.event.repository.name }}-dev
+      name: ${{ github.event.repository.name }}
+      context: .
+      file: Dockerfile
+      tag: ${{ github.event.inputs.tag }}
+
   github-release:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


dvasilas:
ping

dvasilas:
ping

dvasilas:
ping

.github/workflows/release.yaml
```diff
@@ -21,6 +21,18 @@ jobs:
       file: images/svc-base/Dockerfile
       tag: ${{ github.event.inputs.tag }}-svc-base
 
+  build-dev-image:
```
francoisferrand: this is not actually a "dev" image, but the real image...
and it should be pushed as:
```
      namespace: ${{ github.event.repository.name }}
      name: ${{ github.event.repository.name }}
      tag: ${{ github.event.inputs.tag }}
```

dvasilas: Just to be clear, "dev" is just the term I used.
This is the non-Federation image, which for 7/x branches is not used today in production (that's why I called it "dev").

I'm ok for pushing this image under `cloudserver/cloudserver:<tag>` when releasing (the Federation image is pushed under `cloudserver/cloudserver:<tag>-svc-base`)

francoisferrand (approved):

tmacro (approved):

PR #5158, opened 2023-05-09
Bugfix/cldsrv 389 mismatch bucket creation dates

ghost:
Ensuring that the bucket creation dates are the same, as we create multiple entries in the DB in a sequential way, leading to time difference, and causing storage metrics checks to fail.

This is only a partial fix, as we'll need to handle the previously created buckets.

Issue: CLDSRV-389

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -31,7 +31,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
             constants.splitter : constants.oldSplitter;
         let key = createKeyForUserBucket(canonicalID, splitter, bucketName);
         const omVal = {
-            creationDate: new Date().toJSON(),
+            creationDate: new Date(bucketMD.getCreationDate()).toJSON(),
             ingestion: bucketMD.getIngestion(),
         };
         // If the new format usersbucket does not exist, try to put the
@@ -48,7 +48,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
                         'creating users bucket');
                     key = `${canonicalID}${constants.splitter}` +
                         `${bucketName}`;
-                    const creationDate = new Date().toJSON();
+                    const creationDate = bucketMD.getCreationDate();
                     const freshBucket = new BucketInfo(usersBucket,
                         userBucketOwner, userBucketOwner, creationDate,
                         BucketInfo.currentModelVersion());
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@zenko/cloudserver",
-  "version": "8.7.18",
+  "version": "8.7.11",
```
francoisferrand, line 3: why the downgrade in package.json ?
8.7.18 is already released (even 8.7.19, even though it was not bump'ed)

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -31,7 +31,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
             constants.splitter : constants.oldSplitter;
         let key = createKeyForUserBucket(canonicalID, splitter, bucketName);
         const omVal = {
-            creationDate: new Date().toJSON(),
+            creationDate: new Date(bucketMD.getCreationDate()).toJSON(),
```
francoisferrand, line 34: can we just use `bucketMD.getCreationDate()` ?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -48,7 +48,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
                         'creating users bucket');
                     key = `${canonicalID}${constants.splitter}` +
                         `${bucketName}`;
-                    const creationDate = new Date().toJSON();
+                    const creationDate = bucketMD.getCreationDate();
```
francoisferrand, line 51: this is the creation date of the usersBucket itself, not sure it should have the date as the bucket...

Generally, we should avoid introducing correlation where not needed, to avoid people eventually relying on this correlation (c.f Hyrum's law)

lib/api/apiUtils/bucket/bucketCreation.js
francoisferrand: Setting the date to the bucket creation date may be OK for the "first" account, but would not work if the bucket is added to multiple accounts.

Maybe that is not supported at the moment, but as a general "design" approach it would be better to avoid relying on correlations : this usersBucket supports could work with multiple accounts 'sharing' a bucket, so let's not break it; and the precise information on the creation date of a bucket is precisely in the metastore, so we should get it from there...

As you mentioned this "fix" does not work for existing buckets, which reading the creating date from metastore does...

PR #5167, opened 2023-05-17
CLDSRV-374 putMetadata API route is not updating null version properly

nicolas2bert:
Instead of using the provided "null" value, the metadata "null version id" is used when updating the metadata of a null version.

Note: In S3C, the use of the "putMetadata" call is restricted to versioned buckets.

Issue: CLDSRV-374

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -491,14 +491,43 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 omVal[headerName] = objMd[headerName];
             });
         }
-        const versionId = decodeVersionId(request.query);
-        const options = {};
-        if (versionId || omVal.replicationInfo.isNFS) {
-            // specify both 'versioning' and 'versionId' to create a "new"
-            // version (updating master as well) but with specified
-            // versionId
-            options.versioning = bucketInfo.isVersioningEnabled();
-            options.versionId = versionId;
+
+        let versionId;
+        let versioning;
+        const decodedVidResult = decodeVersionId(request.query);
+
+        if (decodedVidResult || omVal.replicationInfo.isNFS) {
+            versionId = decodedVidResult;
+            versioning = bucketInfo.isVersioningEnabled();
+        }
+
+        if (versionId === 'null') {
+            // Retrieve the null version id from the object metadata.
+            versionId = objMd && objMd.versionId;
+            if (!versionId) {
+                if (versioning) {
+                    // If the null version does not have a version id, it is a current null version.
+                    // To update the metadata of a current version, versioning is set to false.
+
+                    // This condition is to handle the case where a single null version looks like a master
+                    // key and will not have a duplicate versioned key and no version ID.
+                    // They are created when you have a non-versioned bucket with objects,
+                    // and then convert bucket to versioned.
+                    // If no new versioned objects are added for given object(s), they look like
+                    // standalone master keys.
+                    versioning = false;
+                } else {
+                    const versioningConf = bucketInfo.getVersioningConfiguration();
+                    // The purpose of this condition is to address situations in which
+                    // - versioning is "suspended" and
+                    // - no existing object or no null version.
+                    // In such scenarios, we generate a new null version and designate it as the master version.
+                    if (versioningConf && versioningConf.Status === 'Suspended') {
+                        versionId = '';
+                        omVal.isNull = true;
+                    }
+                }
+            }
         }
 
         // If the object is from a source bucket without versioning (i.e. NFS),
@@ -506,9 +535,23 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // none was provided in the object metadata value.
         if (omVal.replicationInfo.isNFS) {
             const isReplica = omVal.replicationInfo.status === 'REPLICA';
-            options.versioning = isReplica;
+            versioning = isReplica;
             omVal.replicationInfo.isNFS = !isReplica;
         }
+
+        const options = {
+            versionId,
+        };
+
+        // NOTE: When 'versioning' is set to true and no 'versionId' is specified,
+        // it results in the creation of a "new" version, which also updates the master.
+        // NOTE: Since option fields are converted to strings when they're sent to Metadata via the query string,
+        // Metadata interprets the value "false" as if it were true.
+        // Therefore, to avoid this confusion, we don't pass the versioning parameter at all if its value is false.
+        if (versioning) {
+            options.versioning = true;
+        }
+
         log.trace('putting object version', {
             objectKey: request.objectKey, omVal, options });
         return metadata.putObjectMD(bucketName, objectKey, omVal, options, log,
--- a/tests/multipleBackend/routes/routeBackbeat.js
+++ b/tests/multipleBackend/routes/routeBackbeat.js
@@ -3,7 +3,8 @@ const AWS = require('aws-sdk');
 const async = require('async');
 const crypto = require('crypto');
 const { v4: uuidv4 } = require('uuid');
-const { versioning } = require('arsenal');
+const { models, versioning } = require('arsenal');
+const { ObjectMD } = models;
 const versionIdUtils = versioning.VersionID;
 
 const { makeid } = require('../../unit/helpers');
@@ -39,6 +40,7 @@ const TEST_BUCKET = 'backbeatbucket';
 const TEST_ENCRYPTED_BUCKET = 'backbeatbucket-encrypted';
 const TEST_KEY = 'fookey';
 const NONVERSIONED_BUCKET = 'backbeatbucket-non-versioned';
+const BUCKET_FOR_NULL_VERSION = 'backbeatbucket-null-version';
 
 const testArn = 'aws::iam:123456789012:user/bart';
 const testKey = 'testkey';
@@ -132,6 +134,18 @@ function checkObjectData(s3, bucket, objectKey, dataValue, done) {
     });
 }
 
+function checkVersionData(s3, bucket, objectKey, versionId, dataValue, done) {
+    return s3.getObject({
+        Bucket: bucket,
+        Key: objectKey,
+        VersionId: versionId,
+    }, (err, data) => {
+        assert.ifError(err);
+        assert.strictEqual(data.Body.toString(), dataValue);
+        return done();
+    });
+}
+
 /** makeBackbeatRequest - utility function to generate a request going
  * through backbeat route
  * @param {object} params - params for making request
@@ -162,6 +176,21 @@ function makeBackbeatRequest(params, callback) {
         queryObj,
     };
     makeRequest(options, callback);
+}
+
+function updateStorageClass(data, storageClass) {
+    let parsedBody;
+    try {
+        parsedBody = JSON.parse(data.body);
+    } catch (err) {
+        return { error: err };
+    }
+    const { result, error } = ObjectMD.createFromBlob(parsedBody.Body);
+    if (error) {
+        return { error };
+    }
+    result.setAmzStorageClass(storageClass);
+    return { result };
 }
 
 describe.skip('backbeat DELETE routes', () => {
@@ -224,7 +253,7 @@ function getMetadataToPut(putDataResponse) {
     return mdToPut;
 }
 
-describe.skip('backbeat routes', () => {
+describe('backbeat routes', () => {
     let bucketUtil;
     let s3;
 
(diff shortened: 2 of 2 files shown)


nicolas2bert:
ping

nicolas2bert:
ping

lib/routes/routeBackbeat.js
```diff
@@ -472,14 +472,29 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+
         const versionId = decodeVersionId(request.query);
+        const versioning = bucketInfo.isVersioningEnabled();
+
         // specify both 'versioning' and 'versionId' to create a "new"
         // version (updating master as well) but with specified
         // versionId
         const options = {
-            versioning: bucketInfo.isVersioningEnabled(),
+            versioning,
```
jonathan-gramain: It could be cleaner to
1. prepare the final `versioning` and `versionId` fields to pass in options
2. build the `options` object with those two fields

lib/routes/routeBackbeat.js
```diff
@@ -472,14 +472,29 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
             versionId,
         };
+
+        if (versionId === 'null') {
+            // Retrieve the null version id from the object metadata.
+            const nullVersionId = objMd && objMd.versionId;
+            if (versioning && !nullVersionId) {
+                // If the null version does not have a version id, it is a current null version.
+                // To update the metadata of a current version, versioning is set to false.
+                options.versioning = false;
```
jonathan-gramain: Beware of this, as option fields are converted to strings when passed to Metadata (through the query string), there's a good chance that we interpret `versioning: "false"` as being true in Metadata, I'd rather recommend to not pass `versioning` at all if it is false. Metadata should have better parsing of its options on the other hand but I don't think it's done properly now.

lib/routes/routeBackbeat.js
nicolas2bert: I did not change the existing logic much to have the change easily forwarded into development/8.x branches which have a slightly different logic.

lib/routes/routeBackbeat.js
```diff
@@ -472,22 +472,49 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
         // none was provided in the object metadata value.
         if (omVal.replicationInfo.isNFS) {
             const isReplica = omVal.replicationInfo.status === 'REPLICA';
-            options.versioning = isReplica;
+            versioning = isReplica;
             omVal.replicationInfo.isNFS = !isReplica;
         }
+
+        // NOTE: When 'versioning' is set to true and no 'versionId' is specified,
+        // it results in the creation of a "new" version, which also updates the master.
```
jonathan-gramain: Possibly this comment could fit better above `if (versioning) {`

tests/functional/raw-node/test/routes/routeBackbeat.js
```diff
@@ -172,6 +189,155 @@ describeSkipIfAWS('backbeat routes', () => {
...
     describe('backbeat PUT routes', () => {
+        describe('null version', () => {
+            beforeEach(done => s3.createBucket({ Bucket: BUCKET_FOR_NULL_VERSION }, done));
+            afterEach(done => {
+                bucketUtil.empty(BUCKET_FOR_NULL_VERSION)
+                    .then(() => s3.deleteBucket({ Bucket: BUCKET_FOR_NULL_VERSION }).promise())
+                    .then(() => done());
+            });
+
+            it('should update metadata of a current null version', done => {
```
jonathan-gramain: Would you mind adding some similar tests on versioning-suspended buckets:

- when there is an existing null version
- when there is no existing null version

and make sure we only delete the data when a prior null version exists

lib/routes/routeBackbeat.js
```diff
@@ -502,7 +529,10 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 pushReplicationMetric(objMd, omVal, bucketName, objectKey, log);
                 if (objMd &&
                     headers['x-scal-replication-content'] !== 'METADATA' &&
-                    versionId &&
+                    // - if a version ID is provided in the query string (updating MD of a specific version) OR
+                    // - if no version ID provided and versioning is not enabled (updating MD of the master key),
+                    // data locations are allowed to be removed as they will be overwritten.
+                    (versionId || !versioning) &&
```
jonathan-gramain: I have a doubt that we may remove valid data in the following case, not completely sure, to be checked:

- versioning-suspended bucket
- no existing null version

I am thinking that if `objMd` is the metadata of the current master (a real version), we may remove the master's data and create a new null version, making the old master unreadable.

tests/functional/raw-node/test/routes/routeBackbeat.js
nicolas2bert: In S3C, the use of the "putMetadata" call is restricted to versioned buckets.
I will add tests for Zenko/Artesca branches that support PUT metadata for non-versioned buckets.

lib/routes/routeBackbeat.js
nicolas2bert: If there is no existing null version, the object MD (objMd) of the null version will be undefined, resulting in the condition being false.

lib/routes/routeBackbeat.js
```diff
@@ -472,22 +472,61 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                } else {
+                    const versioningConf = bucketInfo.getVersioningConfiguration();
+                    // The purpose of this condition is to address situations in which
+                    // - versioning is "suspended" and
+                    // - no existing object or no null version.
+                    // In such scenarios, we generate a new null version and designate it as the master version.
+                    if (versioningConf && versioningConf.Status === 'Suspended') {
+                        versionId = '';
+                        omVal.isNull = true;
+                    }
```
Kerkesni, line 503: There might be a case we are missing here (maybe only applicable for the Zenko branches) :
- Create a bucket with versioning suspended
- Put an object
- Enable Versioning
- Put a second version of the same object
- Suspend versioning
- Call PutMetadata on the null version

This will cause Cloudserver to only update the master object without the versions which might break things

lib/routes/routeBackbeat.js
```diff
@@ -472,22 +472,61 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                }
+            }
+        }
```
welansari: My comment might only be applicable for Zenko, but this doesn't work in versioning suspended mode.
When having a null version created after suspending the versioning, the options will equal `{ versionId: '$VERSION_ID' }` which will create another version instead of updating the master object.

nicolas2bert: I'm unsure about which of the tests I wrote is incorrect. Please feel free to identify the test that is wrong.

nicolas2bert: I just ran the tests from https://github.com/scality/cloudserver/pull/5171 and they all passed.

welansari: Yes the tests pass as what i pointed out is not visible from the S3 API side.
Basically the object initially has only a master document in mongo and instead of only updating that, we create a new version object as well as update the mater.
I'm not entirely sure if this is an intended behaviour or not hence why i pointed it out.

nicolas2bert:
While it's not the preferred scenario, due to the urgent need to implement this fix in Artesca, we will initially integrate it into the development/8.7 branch and subsequently backport it to the development/7.x branches.

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -274,6 +303,982 @@ describe.skip('backbeat routes', () => {
             .then(() => done());
     });
 
+    describe.only('null version', () => {
```
jonathan-gramain: `.only`

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -274,6 +303,982 @@ describe.skip('backbeat routes', () => {
...
+                assert(currentVersion.IsLatest);
+                assertVersionIsNullAndUpdated(currentVersion);
+                return done();
+            });
+        });
+
+        it('should create a new null version if versioning suspended and no version', done => {
+            let objMD;
+            return async.series([
+                next => s3.putBucketVersioning({ Bucket: bucket, VersioningConfiguration: { Status: 'Enabled' } },
```
jonathan-gramain: You don't need to enable versioning before suspending it, you can just put a versioning-suspended status on a nonversioned bucket (same for the other tests).

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -274,6 +303,982 @@ describe.skip('backbeat routes', () => {
...
+                    next),
+                next => s3.putBucketVersioning({ Bucket: bucket, VersioningConfiguration: { Status: 'Suspended' } },
+                    next),
+                next => s3.putObject({ Bucket: bucket, Key: keyName, Body: new Buffer(testData) }, next),
```
jonathan-gramain, line 523: Is it useful to PUT and DELETE an object prior to using the backbeat PUT?

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -274,6 +303,982 @@ describe.skip('backbeat routes', () => {
...
+        });
+
+        it('should create a new null version if versioning suspended and delete marker null version', done => {
+            let objMD;
+            return async.series([
+                next => s3.putBucketVersioning({ Bucket: bucket, VersioningConfiguration: { Status: 'Enabled' } },
+                    next),
+                next => s3.putBucketVersioning({ Bucket: bucket, VersioningConfiguration: { Status: 'Suspended' } },
+                    next),
+                next => s3.putObject({ Bucket: bucket, Key: keyName, Body: new Buffer(testData) }, next),
```
jonathan-gramain, line 586: There's probably no need to PUT an object first before DELETE, as a DELETE would normally put a delete marker with a null version.

jonathan-gramain:
Thanks for the extensive test coverage @nicolas2bert

tests/multipleBackend/routes/routeBackbeat.js
nicolas2bert: We eliminated the '.skip' modifier in the 'describe' section of the parent. 
Initially, all tests were being skipped. 
However, I unskipped them and only executed the null test for the time being. 

To address the remaining routeBackbeat tests, I created a follow-up task CLDSRV-394, which is mentioned here: https://github.com/scality/cloudserver/pull/5167/files#diff-e346bc0177b28612b0892a923d8c1d33429b4d7a3c8bd955aadf6e3ccb701dcdR1281.

tests/multipleBackend/routes/routeBackbeat.js
nicolas2bert: Indeed, I simply prefer having it Enabled prior to Suspending it as it feels more instinctive to me. 
However, as enabling it beforehand is not obligatory, I will eliminate that step.

tests/multipleBackend/routes/routeBackbeat.js
nicolas2bert: For the test to be closer to reality and to be able "GET" the null version metadata that I will later "PUT".

tests/multipleBackend/routes/routeBackbeat.js
nicolas2bert: Same reason as earlier. 
Also, we create the object to GET its metadata. 
In my opinion, this appears to be a more realistic scenario.

tests/multipleBackend/routes/routeBackbeat.js
jonathan-gramain: It may cause confusion to have a committed `.only` test, usually we follow the pattern of skipping tests until they pass, I think we should follow this scheme here too, so skipping the rest of the tests in this suite and removing `.only` of the only test we'd like to activate may be more appropriate.

nicolas2bert: I agree.

jonathan-gramain (approved):

lib/routes/routeBackbeat.js
```diff
@@ -491,24 +491,67 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+
+        // NOTE: When 'versioning' is set to true and no 'versionId' is specified,
+        // it results in the creation of a "new" version, which also updates the master.
+        // NOTE: Since option fields are converted to strings when they're sent to Metadata via the query string,
+        // Metadata interprets the value "false" as if it were true.
+        // Therefore, to avoid this confusion, we don't pass the versioning parameter at all if its value is false.
+        if (versioning) {
+            options.versioning = true;
+        }
+
```
welansari, line 554: i can see one case where this won't work:
- create non versioned bucket
- put object
- put object versioning : Enabled
- put metadata

Replication uses the versionId in the actual MD when calling this function (on 8.x)
In the example i gave it will set `undefined` as version id which will create a version and update the master, which means that we loose the null (non versioned) version

welansari: This case actually happened with KB, as they use OOB to copy metadata from an S3C to Artesca and then replicate them, the S3C bucket they used was initially non versioned and contained non versioned objects which they then turned into a versioned bucket to work with OOB.

welansari: maybe this case should be handled in backbeat itself, i'm not really sure

nicolas2bert: Is "should update metadata of a current null version" not covering this test?

nicolas2bert: Indeed `versionId` will be `undefined`, but versioning will be `false`. Only the master should be updated.
This case is handled by setting `versioning = false;` explicitly.

welansari: > Is "should update metadata of a current null version" not covering this test?

Not really cause backbeat doesn't set `null` as version id in the query params, it'll put `undefined` in there, and this is what creates the problem

welansari: i'm resolving this comment as this is a backbeat issue and is easier to address in backbeat

welansari (approved):

lib/routes/routeBackbeat.js
nicolas2bert: If the version ID specified in the query parameters is anything other than "null," it will fail the "null" validation. https://github.com/scality/cloudserver/pull/5167/files#diff-6e9c89f5d34a58972fb42340be6bdaff29f9167883a5265dc64c2d8d384ca31eR504

nicolas2bert:
@bert-e approve

PR #5168, opened 2023-05-22
Test that delete markers are not listed for version suspended buckets

KillianG:
Issue: CLDSRV-347


tests/functional/aws-node-sdk/test/bucket/listingCornerCases.js
```diff
@@ -475,4 +475,56 @@ describe('Listing corner cases tests', () => {
             }
         });
     });
+
+    it('should not list DeleteMarkers for version suspended buckets', done => {
+        const obj = { name: 'testDeleteMarker.txt', value: 'foo' };
+        let objectCount = 0;
+        const Bucket = `versioning-bucket-test-delete-marker-${Date.now()}`;
```
jonathan-gramain: There's already a global called `Bucket`, it could be better to rename this new bucket to avoid confusion.

jonathan-gramain (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@zenko/cloudserver",
-  "version": "8.7.20",
+  "version": "8.7.21",
```
francoisferrand: 8.7.22 has been released already : please rebase & do the "expected" bump

(it is better to only add this kind of release bump at the last minute, once the PR has been approved)

tests/functional/aws-node-sdk/test/bucket/listingCornerCases.js
```diff
@@ -475,4 +475,56 @@ describe('Listing corner cases tests', () => {
...
+                    Key: obj.name,
+                    Body: obj.value,
+                }, err =>
+                next(err)),
+            next => s3.listObjectsV2({ Bucket: bucketName },
+                (err, res) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    objectCount = res.Contents.length;
```
francoisferrand, line 506: here we should check that we did indeed receive the object (testDeleteMarker.txt) : e.g. by going through the result and checking the key is found... (or that the result is "just" only that object)

tests/functional/aws-node-sdk/test/bucket/listingCornerCases.js
```diff
@@ -475,4 +475,56 @@ describe('Listing corner cases tests', () => {
...
+                    assert.strictEqual(
+                        headers['x-amz-delete-marker'], 'true');
+                    return next(err);
+                }),
+            next => s3.listObjectsV2({ Bucket: bucketName },
+                (err, res) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    assert.strictEqual(res.Contents.length, objectCount - 1);
```
francoisferrand, line 524: similarly, should have a stricter check here

package.json
KillianG: Ready to re-review, will bump and resolve conflicts when PR is approved

welansari (approved):

francoisferrand (approved):

PR #5170, opened 2023-05-23
CLDSRV390: bump mongodb to 4.4

ghost:


francoisferrand:
IMHO this does not require a new branch:
* the code is compatible, there is no actual change: i.e. mongo 4.4 was already supported (and both 4.2 and 4.0 are still supported)
* ideally we would "validate" the component against different versions of mongo, i.e. 4.0, 4.2 and 4.4 at the moment (possibly in post-merge / nightly / weekly, this is low risk i believe)
* this may be a much larger change, which does not fit this ticket ; but just bump to mongo 4.4 on existing branch is acceptable in the mean time : anyway, all components are tested together in zenko, and we don't use specific feature of mongo...

anurag4DSB (approved):

KazToozs (approved):

PR #5171, opened 2023-05-23
support putting metadata of null versions in backbeat routes

welansari:
Added support for handling the "null" version id in the putMetadata backbeat route.

Possible cases:
- null version issued from a non versioned object in a versioned or versioning suspended bucket. In this case we should only update the master object in MongoDB
- null version issued from a versioning suspended bucket. In this case we should also only update the master object as versioning suspended objects don't have version objects
- null version issued from a non versioned object in a versioned bucket with a version put on top of it. In this case the null version gets transformed into a version, we need to specify the version id to modify it.

Issue: CLDSRV-392

### DIFF
--- a/tests/functional/backbeat/putMetadata.js
+++ b/tests/functional/backbeat/putMetadata.js
@@ -1,0 +1,223 @@
+const assert = require('assert');
+const async = require('async');
+const uuid = require('uuid');
+const BucketUtility = require('../aws-node-sdk/lib/utility/bucket-util');
+const { removeAllVersions } = require('../aws-node-sdk/lib/utility/versioning-util');
+const { updateMetadata, runIfMongo } = require('./utils');
+
+const credentials = {
+    accessKey: 'accessKey1',
+    secretKey: 'verySecretKey1',
+};
+
+const destLocation = 'us-east-2';
+
+const objectName = 'key';
+
+runIfMongo('backbeat routes: putMetadata', () => {
+    let bucketUtil;
+    let s3;
+    let bucketName;
+
+    before(() => {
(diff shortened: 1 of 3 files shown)


tests/functional/backbeat/putMetadata.js
```diff
@@ -0,0 +1,223 @@
...
+            }),
+            next => s3.putBucketVersioning({
+                Bucket: bucketName,
+                VersioningConfiguration: { Status: 'Enabled' },
+            }, next),
+            next => updateMetadataAndAssertState('null', 1, next),
+        ], done);
+    });
+
+    it('should update the storage class of a null version created from non versioned object '
```
nicolas2bert, line 117: Is it possible to include the following tests as well?

1. versioning is disabled
- PutObject
- PutMetadata(null)
-> Should update the object MD
We will be able to eliminate this "isBucketVersioned" check  as well: https://github.com/scality/cloudserver/blob/development/8.7/lib/api/apiUtils/object/lifecycle.js#L52

2. versioning is "suspended" and no existing object or no null version. In such scenarios, we generate a new null version and designate it as the master version.
- PutBucketVersioning('Suspended')
- PutObject()
- GetMetadata('null')
- DeleteObject('null')
- PutMetadata('null')
-> Should create a new null version

Thanks.

nicolas2bert: Also,
- PutObject()
- PutBucketVersioning('Enabled')
- PutObject()
- PutBucketVersioning('Suspended')
- GetMetadata('null')
- PutMetadata('null')
-≥ Should update the non-latest null version

PR #5191, opened 2023-06-06
keep storage class as cold for restored objects

welansari:
To be compliant with the AWS S3 standard, the storage class of restored objects should be left as cold location

Issue: CLDSRV-400

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -449,6 +449,7 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     metadataStoreParams.creationTime = objMD['creation-time'];
     metadataStoreParams.lastModifiedDate = objMD['last-modified'];
     metadataStoreParams.updateMicroVersionId = true;
+    metadataStoreParams.amzStorageClass = objMD['x-amz-storage-class'];
 
     // set correct originOp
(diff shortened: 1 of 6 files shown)


francoisferrand (commented):
- i get this is missing some tests (in objectHead for exemple), to check he storage class is as expected when the object is 'restored
- there seems to be some extra processing of x-amz-storage-class for multipart : so we should ensure this change does not affect the behavior when restoring a multipart object (e.g. when cold backend would need to do a multipart put to restore)
- `x-amz-storage-class` seems to be used a bit everywhere: so i fear that in non-versionned (or version-suspended), the 'cold' storage class would remain set when the object is replaced... → should be tested as well?

nicolas2bert:
We use the `storageclass` to prevent an object to be lifecycled twice.
Updating `x-amz-storage-class` seems to not impact this logic, however, I would suggest you to double-check that `storageclass` metadata value is not assigned to `x-amz-storage-class` or vice versa somewhere in the code.

francoisferrand (approved):

welansari:
@nicolas2bert i've looked into backbeat and cloudserver, the only instances where i see the values of `storageclass` and `x-amz-storage-class`  getting assigned together is in the case of a lifecycle transition and expiration of a restored object.

nicolas2bert (approved):

PR #5192, opened 2023-06-06
CLDSRV-397 Introduce the time-progression-factor flag

nicolas2bert:
The "time-progression-factor" variable serves as a testing-specific feature that accelerates the progression of time within a system.
By reducing the significance of each day, it enables the swift execution of specific actions, such as expiration, transition, and object locking, which are typically associated with longer timeframes.

This capability allows for efficient testing and evaluation of outcomes, optimizing the observation of processes that would normally take days or even years.
It's important to note that this variable is intended exclusively for testing purposes and is not employed in live production environments, where real-time progression is crucial for accurate results.

### DIFF
--- a/lib/api/apiUtils/object/objectLockHelpers.js
+++ b/lib/api/apiUtils/object/objectLockHelpers.js
@@ -4,6 +4,7 @@ const moment = require('moment');
 const { config } = require('../../../Config');
 const vault = require('../../../auth/vault');
 
+const { scaledMsPerDay } = config.getTimeOptions();
 /**
  * Calculates retain until date for the locked object version
  * @param {object} retention - includes days or years retention period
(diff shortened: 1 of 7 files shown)


alexanderchan-scality (approved):

tests/unit/Config.js
```diff
@@ -254,6 +257,106 @@ describe('Config', () => {
         });
     });
 
+    describe('time options', () => {
+        afterEach(() => {
+            // Clean up the environment variables
+            delete process.env.EXPIRE_ONE_DAY_EARLIER;
```
francoisferrand: `delete` may reset an otherwise useful (for later tests) env variable...
the better way to set env variables in test is to use `setEnv()`, which will store the earlier value and automatically restore or reset at the end of the test.

lib/api/apiUtils/object/objectLockHelpers.js
```diff
@@ -19,8 +20,9 @@ function calculateRetainUntilDate(retention) {
     const date = moment();
     // Calculate the number of days to retain the lock on the object
     const retainUntilDays = days || years * 365;
```
francoisferrand, line 22: unrelated to this PR, but i wonder if the semantics of 'years' parameter here is correct with respect to leap years...

tests/unit/Config.js
nicolas2bert: Yes, agree, I updated the PR.

KazToozs (approved):

nicolas2bert:
@bert-e approve

PR #5195, opened 2023-06-08
CLDSRV-402: optimize API

ghost:
For Artesca, more changes will be introduced with a bump of arsenal, and conflicts resolution (related to code not specific to this PR).

Federation & Integration (green) builds: https://github.com/scality/Integration/actions/runs/5309192284 & https://github.com/scality/Federation/actions/runs/5309145071

Context: The Veeam Ready tests for object with/without immutability are expecting us to perform a data deletion of 4 VM of 100GB each, splitted into 1MB files, in less than 4h. Today Artesca completes this in 10h, and RIng, 6hours.

The objective of this PR is to introduce optimizations for the MultiObjectDelete API:
* With a batching of the objects metadata reads, if the backends supports it (mongodb only for now).
* With an increased parallelism (that won't scale perfectly, but it helps dividing the latency for free)
* With a single call to the storage services, to avoid unecessary iowaits in the API.
* And finally, for Artesca only, supporting the direct deletion of the object metadata, if the associated bucket permits it (i.e. no notification/lifecycle configurations).

**Suggestion for the review**: by commit. Changes in the `lib/api/multiObjectDelete.js` file are not huge, but the diff might display new code, even if only indentation changed...

With these changes:
* Artesca completes the deletion in 39min instead of 10h.
* RING to be tested...

PR for 8.5: https://github.com/scality/cloudserver/pull/5205
8.6: https://github.com/scality/cloudserver/pull/5206
8.7: https://github.com/scality/cloudserver/pull/5207

Note 2: the integration changes are around all the breaking changes we have in this API across cloudserver versions, but the implementation is common between 7.x & 8.x.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -17,6 +17,7 @@ const { azureAccountNameRegex, base64Regex,
 } = require('../constants');
 const { utapiVersion } = require('utapi');
 const { versioning } = require('arsenal');
+const constants = require('../constants');
 
 const versionIdUtils = versioning.VersionID;
 
@@ -1320,6 +1321,16 @@ class Config extends EventEmitter {
         }
 
         this.lifecycleRoleName = config.lifecycleRoleName || null;
+        this.multiObjectDeleteConcurrency = constants.multiObjectDeleteConcurrency;
+        const extractedNumber = Number.parseInt(config.multiObjectDeleteConcurrency, 10);
+        if (!isNaN(extractedNumber) && extractedNumber > 0 && extractedNumber < 1000) {
+            this.multiObjectDeleteConcurrency = extractedNumber;
+        }
+
+        this.multiObjectDeleteEnableOptimizations = true;
+        if (config.multiObjectDeleteEnableOptimizations === false) {
+            this.multiObjectDeleteEnableOptimizations = false;
+        }
     }
 
     _configureBackends() {
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -17,11 +17,14 @@ const { preprocessingVersioningDelete }
     = require('./apiUtils/object/versioning');
 const createAndStoreObject = require('./apiUtils/object/createAndStoreObject');
 const monitoring = require('../utilities/metrics');
-const { metadataGetObject } = require('../metadata/metadataUtils');
+const metadataUtils = require('../metadata/metadataUtils');
 const { config } = require('../Config');
 const { hasGovernanceBypassHeader, checkUserGovernanceBypass, ObjectLockInfo }
     = require('./apiUtils/object/objectLockHelpers');
 const requestUtils = policies.requestUtils;
+const { data } = require('../data/wrapper');
+const logger = require('../utilities/logger');
+const { _bucketRequiresOplogUpdate } = require('./apiUtils/object/deleteObject');
 
 const versionIdUtils = versioning.VersionID;
 
@@ -167,6 +170,63 @@ function _parseXml(xmlToParse, next) {
 }
 
 /**
+ * decodeObjectVersion - decode object version to be deleted
+ * @param {object} entry - entry from data model
+ * @param {function} next - callback to call with error or decoded version
+ * @return {undefined}
+ **/
+function decodeObjectVersion(entry) {
+    let decodedVersionId;
+    if (entry.versionId) {
+        decodedVersionId = entry.versionId === 'null' ?
+            'null' : versionIdUtils.decode(entry.versionId);
+    }
+    if (decodedVersionId instanceof Error) {
+        return [errors.NoSuchVersion];
+    }
+    return [null, decodedVersionId];
+}
+
+/**
+ * Initialization function for the MultiObjectDelete API that will, based on the
+ * current metadata backend, assess if metadata READ batching is supported. If
+ * yes, the initialization step will call the metadataGetObjects function from
+ * the MetadataWrapper.
+ * @param {string} bucketName - bucket name
+ * @param {string []} inPlay - list of object keys still in play
+ * @param {object} log - logger object
+ * @param {function} callback - callback to call with error or list of objects
+ * @return {undefined}
+ */
+function initializeMultiObjectDeleteWithBatchingSupport(bucketName, inPlay, log, callback) {
+    if (config.multiObjectDeleteEnableOptimizations === false) {
+        return callback(null, {});
+    }
+    // If the backend supports batching, we want to optimize the API latency by
+    // first getting all the objects metadata, stored in memory, for later use
+    // in the API. This approach does not change the API architecture, but
+    // transplants an additional piece of code that can greatly improve the API
+    // latency when the database supports batching.
+    const objectKeys = Object.values(inPlay).map(entry => {
+        const [err, versionId] = decodeObjectVersion(entry, bucketName);
+        if (err) {
+            return null;
+        }
+        return {
+            versionId,
+            inPlay: entry,
+        };
+    });
+    return metadataUtils.metadataGetObjects(bucketName, objectKeys, log, (err, cache) => {
+        // This optional step is read-only, so any error can be safely ignored
+        if (err) {
+            return callback(null, {});
+        }
+        return callback(null, cache);
+    });
+}
+
+/**
 * gets object metadata and deletes object
 * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
 * @param {string} canonicalID - canonicalId of requester
@@ -191,34 +251,18 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
     let numOfObjectsRemoved = 0;
     const skipError = new Error('skip');
     const objectLockedError = new Error('object locked');
+    let deleteFromStorage = [];
 
-    // doing 5 requests at a time. note that the data wrapper
-    // will do 5 parallel requests to data backend to delete parts
-    return async.forEachLimit(inPlay, 5, (entry, moveOn) => {
-        async.waterfall([
-            callback => {
-                let decodedVersionId;
-                if (entry.versionId) {
-                    decodedVersionId = entry.versionId === 'null' ?
-                        'null' : versionIdUtils.decode(entry.versionId);
-                }
-                if (decodedVersionId instanceof Error) {
-                    monitoring.promMetrics('DELETE', bucketName, 404,
-                        'multiObjectDelete');
-                    return callback(errors.NoSuchVersion);
-                }
-                return callback(null, decodedVersionId);
-            },
-            // for obj deletes, no need to check acl's at object level
-            // (authority is at the bucket level for obj deletes)
-            (versionId, callback) => metadataGetObject(bucketName, entry.key,
-                versionId, log, (err, objMD) => {
-                    // if general error from metadata return error
-                    if (err) {
-                        monitoring.promMetrics('DELETE', bucketName, err.code,
-                            'multiObjectDelete');
-                        return callback(err);
-                    }
+    return async.waterfall([
+        callback => initializeMultiObjectDeleteWithBatchingSupport(bucketName, inPlay, log, callback),
+        (cache, callback) => async.forEachLimit(inPlay, config.multiObjectDeleteConcurrency, (entry, moveOn) => {
+            async.waterfall([
+                callback => callback(...decodeObjectVersion(entry, bucketName)),
+                // for obj deletes, no need to check acl's at object level
+                // (authority is at the bucket level for obj deletes)
+                (versionId, callback) => metadataUtils.metadataGetObject(bucketName, entry.key,
+                    versionId, cache, log, (err, objMD) => callback(err, objMD, versionId)),
+                (objMD, versionId, callback) => {
                     if (!objMD) {
                         const verCfg = bucket.getVersioningConfiguration();
                         // To adhere to AWS behavior, create a delete marker
@@ -226,7 +270,7 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                         // when versioning has been configured
                         if (verCfg && !entry.versionId) {
                             log.debug('trying to delete specific version ' +
-                            ' that does not exist');
+                                'that does not exist');
                             return callback(null, objMD, versionId);
                         }
                         // otherwise if particular key does not exist, AWS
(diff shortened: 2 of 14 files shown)


williamlardier:
ping

ghost:
ping

ghost:
ping

lib/api/multiObjectDelete.js
```diff
@@ -166,6 +169,60 @@ function _parseXml(xmlToParse, next) {
...
+        monitoring.promMetrics('DELETE', bucketName, 404,
+            'multiObjectDelete');
+        return [errors.NoSuchVersion];
+    }
+    return [null, decodedVersionId];
+}
+
+/**
+ * Initialization function for the MultiObjectDelete API that will, based on the
+ * current metadata backend,a ssess if metadata READ batching is supported. If
```
jonathan-gramain:
```suggestion
 * current metadata backend, assess if metadata READ batching is supported. If
```

lib/api/multiObjectDelete.js
```diff
@@ -166,6 +169,60 @@ function _parseXml(xmlToParse, next) {
...
+ * yes, the initialization step will call the metadataGetObjects function from
+ * the MetadataWrapper.
+ * @param {string} bucketName - bucket name
+ * @param {string []} inPlay - list of object keys still in play
+ * @param {object} log - logger object
+ * @return {undefined}
+ */
+function shouldBatchGetObjectsFromMetadata(bucketName, inPlay, log) {
```
jonathan-gramain: It could be easier to read if the function returned a boolean, as one would expect from the function name, and just call the appropriate init function in the waterfall of `getObjectMetadataAndDelete`

lib/api/multiObjectDelete.js
```diff
@@ -166,6 +169,60 @@ function _parseXml(xmlToParse, next) {
...
+    const backendSupportsBatching = constants.supportsBatchingMethods.includes(config.backends.metadata);
+    // If the backend supports batching, we want to optimize the API latency by
+    // first getting all the objects metadata, stored in memory, for later use
+    // in the API. This approach does not change the API architecture, but
+    // transplants an additional piece of code that can greatly improve the API
+    // latency when the database supports batching.
+    if (backendSupportsBatching) {
+        const objectKeys = Object.keys(inPlay).map(entry => {
```
jonathan-gramain:
```suggestion
        const objectKeys = Object.values(inPlay).map(entry => {
```

And then use `entry` directly instead of `inPlay[entry]`

lib/api/multiObjectDelete.js
```diff
@@ -166,6 +169,60 @@ function _parseXml(xmlToParse, next) {
...
+ * @return {undefined}
+ */
+function shouldBatchGetObjectsFromMetadata(bucketName, inPlay, log) {
+    const backendSupportsBatching = constants.supportsBatchingMethods.includes(config.backends.metadata);
+    // If the backend supports batching, we want to optimize the API latency by
+    // first getting all the objects metadata, stored in memory, for later use
+    // in the API. This approach does not change the API architecture, but
+    // transplants an additional piece of code that can greatly improve the API
+    // latency when the database supports batching.
+    if (backendSupportsBatching) {
```
jonathan-gramain: I would rather have this function just do `return backendSupportsBatching` (the function could then be called the same), and add the rest of the logic to populate the cache directly in the waterfall of `getObjMetadataAndDelete`, to avoid returning closures which can be confusing to read.

lib/metadata/metadataUtils.js
```diff
@@ -84,6 +89,40 @@ function metadataGetObject(bucketName, objectKey, versionId, log, cb) {
...
+        key: objectKey ? objectKey.inPlay.key : null,
+        params: options,
+        versionId: objectKey ? objectKey.versionId : null,
+    }));
+
+    // Returned objects are following the following format: { key, doc, versionId }
+    // That is required with batching to properly map the objects
+    return metadata.getObjectsMD(bucketName, objects, log, (err, objMds) => {
+        if (err) {
+            log.debug('err getting object MD from metadata', { error: err });
```
jonathan-gramain:
```suggestion
            log.debug('error getting object MD from metadata', { error: err });
```

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
+                            return callback(err);
+                        }
+                        if (!objMD) {
+                            const verCfg = bucket.getVersioningConfiguration();
+                            // To adhere to AWS behavior, create a delete marker
+                            // if trying to delete an object that does not exist
+                            // when versioning has been configured
+                            if (verCfg && !entry.versionId) {
+                                log.debug('trying to delete specific version ' +
+                                    ' that does not exist');
```
jonathan-gramain:
```suggestion
                                    'that does not exist');
```

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
-                            ' that does not exist');
-                            return callback(null, objMD, versionId);
+    return async.waterfall([
+        callback => initialStep((err, cache) => callback(err, cache)),
+        (cache, callback) => async.forEachLimit(inPlay, 50, (entry, moveOn) => {
+            async.waterfall([
+                callback => callback(...processObjectVersion(entry, bucketName)),
+                // for obj deletes, no need to check acl's at object level
+                // (authority is at the bucket level for obj deletes)
+                (versionId, callback) => metadataGetObject(bucketName, entry.key,
```
jonathan-gramain: I would expect this `metadataGetObject` call not to happen when the entry is already in cache (due to `metadataGetObjects` called earlier to populate the cache), but I don't see the mechanism to check this and skip `metadataGetObject`.

lib/api/multiObjectDelete.js
jonathan-gramain: Sorry this kinda duplicates my earlier comment https://github.com/scality/cloudserver/pull/5195#discussion_r1231640057

lib/api/multiObjectDelete.js
ghost: I wonder if this is worth changing the API architecture for this? Originally this was something we discussed with @francoisferrand specifically not to change the API structure while benefiting from the previous cache. I am fine with both approaches, but the current way seem less risky...

We could just wrap this call and do something like:

```js
const handleObjectMetadata = (err, objMD, callback) => {
    // if general error from metadata return error
    if (err) {
        monitoring.promMetrics('DELETE', bucketName, err.code,
            'multiObjectDelete');
        return callback(err);
    }
    if (!objMD) {
        const verCfg = bucket.getVersioningConfiguration();
        // To adhere to AWS behavior, create a delete marker
        // if trying to delete an object that does not exist
        // when versioning has been configured
        if (verCfg && !entry.versionId) {
            log.debug('trying to delete specific version ' +
                ' that does not exist');
            return callback(null, objMD, versionId);
        }
        // otherwise if particular key does not exist, AWS
        // returns success for key so add to successfullyDeleted
        // list and move on
        successfullyDeleted.push({ entry });
        return callback(skipError);
    }
    if (versionId && objMD.location &&
        Array.isArray(objMD.location) && objMD.location[0]) {
        // we need this information for data deletes to AWS
        objMD.location[0].deleteVersion = true;
    }
    return callback(null, objMD, versionId);
};
...
(versionId, callback) => {
    if (cache && cache[`${entry.key}${versionId}`]) {
        return handleObjectMetadata(null, cache[`${entry.key}${versionId}`], callback);
    } else {
        return metadataGetObject(bucketName, entry.key, versionId, log, (err, objMD) => handleObjectMetadata(err, objMD, callback));
    }
...
```

lib/api/multiObjectDelete.js
ghost: I would prefer keeping this function separately, that greatly ease the unit testing (and the base function is already quite complex :smile: ). Rather than just returning a flag, I will rename it `initializeMultiObjectDeleteWithBatchingSupport` instead, but I'm still open if you think we should change it anyway

lib/metadata/metadataUtils.js
ghost: Adding a 's' also to 'object' to differentiate with the other method.

ghost:
Note for reviewers: review commits will be integrated once I get more feedback on this PR. The implementation remains the same for integration branches.

lib/api/multiObjectDelete.js
ghost: See https://github.com/scality/cloudserver/pull/5195/commits/9f7cc75b6259b61fcf0cbfa22d93fc544328f703

lib/api/multiObjectDelete.js
jonathan-gramain: Ah OK with the `cachedDocument` param, fine for me.

I don't know if it's the best approach but should work, personally I'd rather have abstracted the `metadataGetObjects` implementation itself into a single call to `getObjectsMD` or an async loop of `getObjectMD` calls (with concurrency limit), but if you want to minimize latency you may want to start processing results as soon as they arrive instead if the backend doesn't support `getObjectsMD`, however it's probably not a significant optimization in my sense.

lib/services.js
```diff
@@ -286,9 +286,12 @@ const services = {
      *                           delete a specific version of the object
      * @param {Log} log - logger instance
      * @param {function} cb - callback from async.waterfall in objectGet
+     * @param {boolean} deferDeletion - true if the object should not be removed
+     *                                  from the storage, but be returned
+     *                                  instead.
      * @return {undefined}
      */
-    deleteObject(bucketName, objectMD, objectKey, options, log, cb) {
+    deleteObject(bucketName, objectMD, objectKey, options, log, cb, deferDeletion) {
```
jonathan-gramain: Since `deferDeletion` refers to **data** locations deletion, it should I think be renamed for clarity, e.g. `deferLocationsDeletion`

jonathan-gramain (approved):

lib/api/multiObjectDelete.js
```diff
@@ -166,6 +169,60 @@ function _parseXml(xmlToParse, next) {
...
+ * current metadata backend, assess if metadata READ batching is supported. If
+ * yes, the initialization step will call the metadataGetObjects function from
+ * the MetadataWrapper.
+ * @param {string} bucketName - bucket name
+ * @param {string []} inPlay - list of object keys still in play
+ * @param {object} log - logger object
+ * @return {undefined}
+ */
+function initializeMultiObjectDeleteWithBatchingSupport(bucketName, inPlay, log) {
+    const backendSupportsBatching = constants.supportsBatchingMethods.includes(config.backends.metadata);
```
francoisferrand: why do we need this `at all` ?

the support of batching is already handled in MetadataWrapper, which checks if `getObjects` is implemented : so here we should always initialize the cache by calling `MetadataWrapper.getObjectsMD`:
* If implemented, it will return the objects and fill the cache (or possibly return errors)
* If not implemented, it will simply create an empty cache
* Later on, the algorithm will check for object in the cache else call getObject, and we will be fine

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
     let numOfObjectsRemoved = 0;
     const skipError = new Error('skip');
     const objectLockedError = new Error('object locked');
+    let deleteFromStorage = [];
+    const initialStep = initializeMultiObjectDeleteWithBatchingSupport(bucketName, inPlay, log);
```
francoisferrand: the callback initialized from the function makes the code harder to follow/read, and I don't see the benefit of the functional approach here : we may as well do what needs to be done in the waterfall, there be one "branch" anyway to handle both caching/no-caching...

lib/metadata/metadataUtils.js
```diff
@@ -62,11 +62,16 @@ function getNullVersionFromMaster(bucketName, objectKey, log, cb) {
  * @param {string} [versionId] - version of object to retrieve
  * @param {RequestLogger} log - request logger
  * @param {function} cb - callback
+ * @param {object} cachedDocument - cached version of the document used for
+ *                                  abstraction purposes
  * @return {undefined} - and call callback with err, bucket md and object md
  */
-function metadataGetObject(bucketName, objectKey, versionId, log, cb) {
+function metadataGetObject(bucketName, objectKey, versionId, log, cb, cachedDocument = null) {
```
francoisferrand: shouldn't `cb` be the last param?

lib/api/multiObjectDelete.js
francoisferrand: c.f. my next comment: I guess this would be much clearer if the `callback` was kept as last param
Having the cache after the callback makes it very hard to read the call...

lib/metadata/metadataUtils.js
francoisferrand: also IMO we should pass the whole cache to this function, and let it locate the object : if the caller already handles finding the object in the cache, then he may just as well skip the function altogether...

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
-                            return callback(null, objMD, versionId);
+    return async.waterfall([
+        callback => initialStep((err, cache) => callback(err, cache)),
+        (cache, callback) => async.forEachLimit(inPlay, 50, (entry, moveOn) => {
+            async.waterfall([
+                callback => callback(...processObjectVersion(entry, bucketName)),
+                // for obj deletes, no need to check acl's at object level
+                // (authority is at the bucket level for obj deletes)
+                (versionId, callback) => metadataGetObject(bucketName, entry.key,
+                    versionId, log, (err, objMD) => {
```
francoisferrand: should just pass 'callback' and move this anonymous function as the next stage in the `async.waterfall`

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
+                        // if general error from metadata return error
+                        if (err) {
+                            monitoring.promMetrics('DELETE', bucketName, err.code,
```
francoisferrand: should the metric not be updated in the other error cases as well, e.g. in the final 'error' handling ?

lib/api/multiObjectDelete.js
ghost: yes, I will remove it, at first the check in Arsenal was not here, now this check does not make sense anymore

lib/metadata/metadataUtils.js
ghost: > shouldn't cb be the last param?

We can make it a non optional parameter yes, and change all calls to this function...

lib/api/multiObjectDelete.js
ghost: What is the rationale? That is not what was done before, and I do not want to change the structure...

lib/api/multiObjectDelete.js
```diff
@@ -166,6 +169,60 @@ function _parseXml(xmlToParse, next) {
...
 }
 
+/**
+ * processObjectVersion - process object version to be deleted
+ * @param {object} entry - entry from data model
+ * @param {string} bucketName - bucket name
+ * @param {function} next - callback to call with error or decoded version
+ * @return {undefined}
+ **/
+function processObjectVersion(entry, bucketName) {
```
francoisferrand: not clear what this does (and it certainly does not "process" a version): and especially when reading in the waterflow, extracting this function completely hides what is done...

maybe it should just be renamed to `decodeObjectVersion` ?

lib/api/multiObjectDelete.js
```diff
@@ -166,6 +169,60 @@ function _parseXml(xmlToParse, next) {
...
+    let decodedVersionId;
+    if (entry.versionId) {
+        decodedVersionId = entry.versionId === 'null' ?
+            'null' : versionIdUtils.decode(entry.versionId);
+    }
+    if (decodedVersionId instanceof Error) {
+        monitoring.promMetrics('DELETE', bucketName, 404,
```
francoisferrand: updating the `promMetrics` does not make sense here : this is not associated to the purpose of this function (decoding version id?), but to the overal process... and so should be kept somewhere in `getObjMetadataAndDelete` (eg keep level of abstraction)

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
-                        // To adhere to AWS behavior, create a delete marker
-                        // if trying to delete an object that does not exist
-                        // when versioning has been configured
-                        if (verCfg && !entry.versionId) {
-                            log.debug('trying to delete specific version ' +
-                            ' that does not exist');
-                            return callback(null, objMD, versionId);
+    return async.waterfall([
+        callback => initialStep((err, cache) => callback(err, cache)),
+        (cache, callback) => async.forEachLimit(inPlay, 50, (entry, moveOn) => {
```
francoisferrand: the concurrency level should be a (global) constant at least, and/or maybe even a config parameter to give the possibly of (eventually) tweaking it...

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
+                    const deleteInfo = {};
+                    if (options && options.deleteData) {
+                        deleteInfo.deleted = true;
+                        if ((bucket.getLifecycleConfiguration && !bucket.getLifecycleConfiguration())
+                            && (bucket.getNotificationConfiguration && !bucket.getNotificationConfiguration())) {
+                            options.shouldOnlyDelete = true;
                         }
-                        if (error) {
-                            return callback(error);
+                        if (objMD.uploadId) {
```
francoisferrand: this condition may change in the future : i find it should be isolated in a function so that we can easily evolve and maintain it (and possibly introduce more complex conditions)

lib/api/multiObjectDelete.js
```diff
@@ -191,166 +248,187 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
-                            log.debug('user does not have BypassGovernanceRetention and object is locked', { error });
-                            return callback(objectLockedError);
+                },
+                (objMD, versionId, callback) => {
+                    const options = preprocessingVersioningDelete(
+                        bucketName, bucket, objMD, versionId, config.nullVersionCompatMode);
+                    const deleteInfo = {};
+                    if (options && options.deleteData) {
+                        deleteInfo.deleted = true;
+                        if ((bucket.getLifecycleConfiguration && !bucket.getLifecycleConfiguration())
```
francoisferrand: in case lifecycle, I think this would be needed only for _lifecycle transition to cold_ is used, and not even at the moment (we did not implement the delete event there event...) : not sure if we can/want to be more precise, or even just skip the test for now.

@nicolas2bert can you confirm if there are conditions where we need the full metadata in the oplog, for lifecycle?

lib/api/multiObjectDelete.js
ghost: Shouldn't we do that in a separate PR for all the places we are relying on such constants? I can do it for this one, we we still have raw values in our functions

lib/metadata/metadataUtils.js
```diff
@@ -84,6 +89,40 @@ function metadataGetObject(bucketName, objectKey, versionId, log, cb) {
...
+ * method uses cursors, hence is only compatible with a MongoDB DB backend.
+ * @param {string} bucketName - name of bucket
+ * @param {string} objectsKeys - name of object key
+ * @param {RequestLogger} log - request logger
+ * @param {function} cb - callback
+ * @return {undefined} - and call callback with err, bucket md and object md
+ */
+function metadataGetObjects(bucketName, objectsKeys, log, cb) {
+    const options = { getDeleteMarker: true };
+    const objects = objectsKeys.map(objectKey => ({
```
francoisferrand, line 102: there are already 2 maps in arsenal, and one more here : seems a lot, should maybe rework something to find a proper 'format' which suits all and avoids too many conversions...

lib/api/multiObjectDelete.js
francoisferrand: in case lifecycle, I think this would be needed only for _lifecycle transition to cold_ is used, and not even at the moment (we did not implement the delete event there event...) : not sure if we can/want to be more precise, or even just skip the test for now.

@nicolas2bert can you confirm if there are conditions where we need the full metadata in the oplog, for lifecycle?

lib/services.js
```diff
@@ -286,9 +286,12 @@ const services = {
...
         log.trace('deleting object from bucket');
```
francoisferrand: `cb` should be the last param...

or actually, it seems to me we should not change the prototype of this function:
* `deleteObject(deferDeletion: true)` seems oxymoric, and not very readable
* it would be better instead of refactor with a function which "deletes the metadata and returns the list of locations" : which can be used to implement `deleteObject` as well as directly called by the multi delete process.
* maybe that intermediate function is not even needed, and we should simply call `metadata.deleteObjectMD` from the multi delete process.

lib/services.js
```diff
@@ -305,12 +308,16 @@ const services = {
...
-                    deleteLog, err => {
+
+                    const locations = Array.isArray(objectMD.location)
+                        ? objectMD.location : [objectMD.location];
+
+                    if (deferDeletion) {
+                        return cb(null, locations);
+                    }
+
+                    return data.batchDelete(locations, null, null, deleteLog, err => {
```
francoisferrand: code used to call `data.delete` here instead of `batchDelete` : not sure we should change that... This is probably a minor optimisation, but still...

tests/unit/api/multiObjectDelete.js
```diff
@@ -185,3 +185,48 @@ describe('getObjMetadataAndDelete function for multiObjectDelete', () => {
             });
     });
```
francoisferrand, line 193: missing tests here, esp. to verify the behaviours which you changed

* Is the cache populated when supported (e.g. call to `getObjectsMD`)
* Is the cache used (e.g. `getObjectMD` not called when object is already in cache)
* Is it working in both "with cache" and "without cache" modes
* Proper "batch" error handling : e.g. verify that we (batch) delete the `data` for each `metadata` we deleted even if there are errors on "other" objects

lib/api/multiObjectDelete.js
francoisferrand: > Shouldn't we do that in a separate PR for all the places we are relying on such constants

sure, if there are other such constant which should be cleaned'up, please create a `TechDebt` jira ticket

> I can do it for this one

yes please, this one should be fixed (the "constant" part at least) in this PR.
Fixing the rest of the world can wait, but at least we should improve the situation bit by bit when we find an issue and modify that code.

lib/metadata/metadataUtils.js
ghost: This one is required anyway, because we need a link between the object key, its params and the version id. However I updated the function in Arsenal to remove one of the maps

lib/services.js
ghost: I ended up not changing the function at all, because the azure protection logic is related to the data services, and if we return all locations, we would end up deleting unnecessary things:
* it does not make sense to duplicate the `protectAzureBlocks` call
* yet if we call only a `deleteObjectMD` and `protectAzureBlocks` was supposed to return false, we change the behavior (even if this is harmless, this is not the objective of this PR)

I am sharing here the implementation I did before changing my mind:

```js
    /**
     * Deletes metadata of the object
     * @param {string} bucketName - bucket in which objectMD is stored
     * @param {string} objectKey - object key name
     * @param {object} options - other instructions, such as { versionId } to
     *                           delete a specific version of the object
     * @param {Log} log - logger instance
     * @param {function} cb - callback function
     * @return {undefined}
     */
    deleteMetadata(bucketName, objectKey, options, log, cb) {
        log.trace('Deleting object metadata and returning locations');

        return metadata.deleteObjectMD(bucketName, objectKey, options, log,
            (err, res) => {
                if (err) {
                    return cb(err, res);
                }

                log.trace('deleteMetadata: metadata delete OK');
                return cb(err, res);
            });
    },

    /**
     * Deletes objects from a bucket
     * @param {string} bucketName - bucket in which objectMD is stored
     * @param {object} objectMD - object's metadata
     * @param {string} objectKey - object key name
     * @param {object} options - other instructions, such as { versionId } to
     *                           delete a specific version of the object
     * @param {Log} log - logger instance
     * @param {function} cb - callback from async.waterfall in objectGet
     * @return {undefined}
     */
    deleteObject(bucketName, objectMD, objectKey, options, log, cb) {
        log.trace('deleting object from bucket');
        assert.strictEqual(typeof objectMD, 'object');
        const objGetInfo = objectMD.location;

        return services.deleteMetadata(bucketName, objectKey, options, log,
            (err, res) => {
                if (err) {
                    return cb(err, res);
                }
                log.trace('deleteObject: metadata delete OK');
                const deleteLog =
                    logger.newRequestLoggerFromSerializedUids(
                        log.getSerializedUids());
                if (objGetInfo === null) {
                    return cb(null, res);
                } else if (!Array.isArray(objGetInfo)) {
                    data.delete(objGetInfo, deleteLog);
                    return cb(null, res);
                }
                return data.batchDelete(objGetInfo, null, null,
                    deleteLog, err => {
                        if (err) {
                            return cb(err);
                        }
                        return cb(null, res);
                    });
            });
    },
```

lib/api/multiObjectDelete.js
ghost: Done in https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR201

(This is a tentative commit, it may not pass the ci, in such case I'll fix it)

lib/api/multiObjectDelete.js
ghost: I reworked the function call to directly return the callback result: https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR201 && https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR254

lib/metadata/metadataUtils.js
ghost: Done: https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-459659e46eee4c3ab38fd506823fd48fbccb62998dfb6d110d4bb7cf7c1ee036R69

lib/api/multiObjectDelete.js
ghost: Done: https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR260

lib/api/multiObjectDelete.js
ghost: Added at the end of the function: https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR424

lib/api/multiObjectDelete.js
ghost: Done: https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR178

lib/api/multiObjectDelete.js
ghost: See above comment, it's now in the last callback, in case of error, I removed this one

lib/api/multiObjectDelete.js
ghost: I added support for that in the constants,; but also in the configuration file, so we can bind it with ZKOP at some point if needed
https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-e8319ec288d282012f389d259d6468a0922b91ec7cc52f2b8c15b01ff3ba9c2aR255

lib/api/multiObjectDelete.js
ghost: I created https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-b90cf333e787e31130d0434b2e48b75b69948c9e9c594c2d5f201058535b6ef4R1 that is also used when doing single deletes

lib/services.js
ghost: Done https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-a76cecfcb0b408028fb3c7e02359a47730668d6424f8b02bb9522c162c21ce47R312

tests/unit/api/multiObjectDelete.js
ghost: I updated some unit tests: https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-c2cf43a7212798ef328432b312d2ef420db8c78323dad28d5ea930cdb251b519R227
and added a new test for the last point you mentioned: https://github.com/scality/cloudserver/pull/5195/commits/21d22ad1832687ed188f284dcafeaf6ae1c4f285#diff-c2cf43a7212798ef328432b312d2ef420db8c78323dad28d5ea930cdb251b519R195

lib/api/multiObjectDelete.js
francoisferrand: rationale is really just that this is a waterfall call : whose call is to replace the "final" callback by the next stage in waterfall...

ie instead of:
```
metadataGetObject(bucketName, entry.key, versionId, log, (err, objMD) => {
    [...]
```
we should have:
```
metadataGetObject(bucketName, entry.key, versionId, log, callback),
(err, objMD) => {
    [...]
```

(unless there are other subtleties, there may be reasons why we cannot do this...)

francoisferrand:
I wonder if we can/should introduce some flag to "disable" the optimization by default, and allow us more control when we enable it ; esp. since we are introducing this in older waterfall branch and may need to have it in patch versions of product...

in theory it should be added, but on the other hand I am not sure it is "easy" to disable (i.e. the cost and risk of allowing deactivation may not be worth the benefit) or if this is a critical/blocking fix for product (which we need to enable ASAP everywhere, and which should not ever be disabled).

ghost:
Rebased following CI fixes merged in another branch.

lib/Config.js
```diff
@@ -1320,6 +1321,18 @@ class Config extends EventEmitter {
         }
 
         this.lifecycleRoleName = config.lifecycleRoleName || null;
+        this.multiObjectDeleteConcurrency = constants.multiObjectDeleteConcurrency;
+        if (!isNaN(config.multiObjectDeleteConcurrency)) {
+            const extractedNumber = Number.parseInt(config.multiObjectDeleteConcurrency, 10);
+            if (extractedNumber > 0 && extractedNumber < 1000 && extractedNumber % 1 === 0) {
```
francoisferrand: > `extractedNumber % 1 === 0`

i don't see the point of this check, `parseInt` returns an integer anyway

lib/Config.js
```diff
@@ -1320,6 +1321,18 @@ class Config extends EventEmitter {
         }
 
         this.lifecycleRoleName = config.lifecycleRoleName || null;
+        this.multiObjectDeleteConcurrency = constants.multiObjectDeleteConcurrency;
+        if (!isNaN(config.multiObjectDeleteConcurrency)) {
```
francoisferrand: what is the point of checking for `NaN` here: `Number.parseInt` will handle it and return an error/NaN when needed anyway?

lib/Config.js
```diff
@@ -1320,6 +1321,18 @@ class Config extends EventEmitter {
...
+            const extractedNumber = Number.parseInt(config.multiObjectDeleteConcurrency, 10);
+            if (extractedNumber > 0 && extractedNumber < 1000 && extractedNumber % 1 === 0) {
+                this.multiObjectDeleteConcurrency = extractedNumber;
+            }
+        }
+
+        this.multiObjectDeleteEnableOptimizations = true;
+        if (config.multiObjectDeleteEnableOptimizations === false) {
```
francoisferrand, line 1331:
```suggestion
        if (!config.multiObjectDeleteEnableOptimizations) {
```

lib/api/multiObjectDelete.js
```diff
@@ -242,114 +286,152 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
-                    if (objMD.uploadId) {
-                        // eslint-disable-next-line
-                        options.replayId = objMD.uploadId;
+                // If error but we have objects in the list, delete them to ensure
+                // consistent state.
+                log.trace('deleting objects from data backend');
+                const deleteLog = logger.newRequestLoggerFromSerializedUids(log.getSerializedUids());
+                return data.batchDelete(deleteFromStorage, null, null, deleteLog, err => {
+                    if (err) {
+                        log.error('error deleting objects from data backend', { error: err });
```
francoisferrand: when this happens, all the "metrics" computed above may be wrong : numOfObjectsRemoved, successfullyDeleted, totalContentLengthDeleted...

* if `batchDelete` works somewhat "atomically" (e.g. it succeeds to delete all OR fails and delete nothing), then the logic can be tweak to count properly;
* otherwise, we may need that function to return the "list" of objects succesfully deleted (and their size?), so we can do the appropriate count

KazToozs (approved):
lgtm

lib/api/multiObjectDelete.js
ghost: The current behavior is that, in case of any error, the function returns, and the metrics are unchanged, as they are only computed based on (successful) operations on the metadata, not the storage. A failure in deleting the actual data would lead to the whole API to fail. The only difference I see here is the number or orphans we would get, more than an issue with the metrics. When the API fails, we do not push any metric... So to me, this is not something to handle, as we push no metrics anyway.

The main question is more about handling or orphans: this can be mitigated by pre-splitting the array from `1000` to chunks of `multiObjectDeleteConcurrency` as this will be aligned with the current parallelism, without the drawback of waiting and doing the calls one by one. I will make the associated changes.

lib/Config.js
ghost: This was kept unchanged, because we want to enable the optimization by default, and this change would disable it, as there is no such configuration by default.

williamlardier:
ping

ghost:
ping

ghost:
ping

PR #5197, opened 2023-06-09
Remove userMD counting restore retries and add tests

KillianG:
Remove 'x-amz-meta-scal-s3-restore-attempt' user metadata that is used to count the time we try a restore

Issue: CLDSRV-403

### DIFF
--- a/lib/services.js
+++ b/lib/services.js
@@ -182,6 +182,9 @@ const services = {
                 'ongoing-request': false,
                 'expiry-date': archive.restoreWillExpireAt,
             });
+            md.setUserMetadata({
+                'x-amz-meta-scal-s3-restore-attempt': undefined,
+            });
         }
 
         if (oldReplayId) {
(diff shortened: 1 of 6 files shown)


lib/services.js
```diff
@@ -182,6 +182,9 @@ const services = {
                 'ongoing-request': false,
                 'expiry-date': archive.restoreWillExpireAt,
             });
+            md.setUserMetadata({
+                'x-amz-meta-scal-s3-restore-attempt': undefined,
+            });
```
francoisferrand, line 187: * we should not "set" the user metadata : what we want to is actually to _remove_ this specific user metadata
* i don't understand how this solves the problem: here the `md` object is created from scratch (line 106), so it has no user metadata ; and existing metadata (if any) are set later (line 239)... what am i missing?

francoisferrand: Also this function has a generic name (`metadataStoreObject`) : wondering if we should fix the problem here, or if it makes sense to handle that in "higher" level function (`createAndStoreObject`) which handles the 'copy' of the metadata?

(esp. what we want to make sure we reset the flag only at the end of restore, and not on other metadata updates: setting tags, ...)

KillianG: Maybe @Kerkesni can answer this cause he has more knowledge than me on that and he told me to do that here

PR #5214, opened 2023-06-22
CLDSRV-409: use latest s3cmd with python3

ghost:
Fixes the CI, as python 2.7 is now removed from the action installing it. Instead of keeping old code, let's take advantage of the situation and upgrade everything.

This requires some minor updates in the test suite as the s3cmd output format changed a bit.


williamlardier:
ping

anurag4DSB (approved):

welansari (approved):

PR #5215, opened 2023-06-22
Fix user MD and tags getting deleted when restoring

KillianG:
Issue: CLDSRV-408

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -434,6 +434,47 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
     return options;
 }
 
+/**
+ * Keep metadatas when the object is restored from cold storage
+ * but remove the specific ones we don't want to keep
+ * @param {object} objMD - obj metadata
+ * @param {object} metadataStoreParams - custom built object containing resource details.
+ * @return {undefined}
+ */
+function restoreMetadata(objMD, metadataStoreParams) {
+    /* eslint-disable no-param-reassign */
+    const userMDToSkip = ['x-amz-meta-scal-s3-restore-attempt'];
+    // We need to keep user metadata and tags
+    Object.keys(objMD).forEach(key => {
+        if (key.startsWith('x-amz-meta-') && !userMDToSkip.includes(key)) {
+            metadataStoreParams.metaHeaders[key] = objMD[key];
+        }
+    });
+
+    if (objMD['x-amz-website-redirect-location']) {
+        if (!metadataStoreParams.headers) {
+            metadataStoreParams.headers = {};
+        }
+        metadataStoreParams.headers['x-amz-website-redirect-location'] = objMD['x-amz-website-redirect-location'];
+    }
+
+    if (objMD.replicationInfo) {
+        metadataStoreParams.replicationInfo = objMD.replicationInfo;
+    }
+
+    if (objMD.legalHold) {
+        metadataStoreParams.legalHold = objMD.legalHold;
+    }
+
+    if (objMD.acl) {
+        metadataStoreParams.acl = objMD.acl;
+    }
+
+    metadataStoreParams.creationTime = objMD['creation-time'];
+    metadataStoreParams.lastModifiedDate = objMD['last-modified'];
+    metadataStoreParams.taggingCopy = objMD.tags;
+}
+
 /** overwritingVersioning - return versioning information for S3 to handle
  * storing version metadata with a specific version id.
  * @param {object} objMD - obj metadata
@@ -445,9 +486,6 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
  *  version id of the null version
  */
 function overwritingVersioning(objMD, metadataStoreParams) {
-    /* eslint-disable no-param-reassign */
-    metadataStoreParams.creationTime = objMD['creation-time'];
-    metadataStoreParams.lastModifiedDate = objMD['last-modified'];
     metadataStoreParams.updateMicroVersionId = true;
     metadataStoreParams.amzStorageClass = objMD['x-amz-storage-class'];
 
@@ -477,6 +515,8 @@ function overwritingVersioning(objMD, metadataStoreParams) {
             nullVersionId: objMD.nullVersionId,
         };
     }
+
+    restoreMetadata(objMD, metadataStoreParams);
 
     return options;
 }
--- a/lib/services.js
+++ b/lib/services.js
@@ -260,6 +260,11 @@ const services = {
         if (legalHold) {
             md.setLegalHold(legalHold);
         }
+        if (params.acl) {
+            // In case of a restore we dont pass ACL in the headers
+            // but we take them from the old metadata
+            md.setAcl(params.acl);
+        }
 
         log.trace('object metadata', { omVal: md.getValue() });
         // If this is not the completion of a multipart upload or
--- a/tests/functional/aws-node-sdk/test/object/mpuVersion.js
+++ b/tests/functional/aws-node-sdk/test/object/mpuVersion.js
@@ -12,6 +12,7 @@ const log = new DummyRequestLogger();
 
 const bucketName = 'bucket1putversion33';
 const objectName = 'object1putversion';
+const bucketNameMD = 'restore-metadata-copy-bucket-mpu';
 const mdListingParams = { listingType: 'DelimiterVersions', maxKeys: 1000 };
 const archive = {
     archiveInfo: {},
@@ -104,22 +105,19 @@ describe('MPU with x-scal-s3-version-id header', () => {
         beforeEach(done => {
             bucketUtil = new BucketUtility('default', sigCfg);
             s3 = bucketUtil.s3;
-            return metadata.setup(() =>
-                s3.createBucket({ Bucket: bucketName }, err => {
-                    if (err) {
-                        assert.strictEqual(err, null, 'Creating bucket: Expected success, ' +
-                            `got error ${JSON.stringify(err)}`);
-                    }
-                    done();
-                }));
+            async.series([
+                next => metadata.setup(next),
+                next => s3.createBucket({ Bucket: bucketName }, next),
+                next => s3.createBucket({ Bucket: bucketNameMD, ObjectLockEnabledForBucket: true, }, next),
+            ], done);
         });
 
         afterEach(() => {
             process.stdout.write('Emptying bucket');
-            return bucketUtil.empty(bucketName)
+            return bucketUtil.emptyMany([bucketName, bucketNameMD])
             .then(() => {
                 process.stdout.write('Deleting bucket');
-                return bucketUtil.deleteOne(bucketName);
+                return bucketUtil.deleteMany([bucketName, bucketNameMD]);
             })
             .catch(err => {
                 process.stdout.write('Error in afterEach');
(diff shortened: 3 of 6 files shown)


lib/api/apiUtils/object/versioning.js
```diff
@@ -453,6 +453,13 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     // set correct originOp
     metadataStoreParams.originOp = 's3:ObjectRestore:Completed';
 
+    // We need to keep user metadata and tags
+    for (const key of Object.keys(objMD)) {
+        if (key.startsWith('x-amz-meta-')) {
```
francoisferrand: must not copy the "restore count" though ; and we should take the chance to check if there are other "scality" user metadata which should not be copied...

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,43 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+ * @param {object} objMD - obj metadata
+ * @param {object} metadataStoreParams - custom built object containing resource details.
+ * @return {undefined}
+ */
+function restoreMetadatas(objMD, metadataStoreParams) {
+    /* eslint-disable no-param-reassign */
+    const userMDNotToKeep = ['x-amz-meta-scal-s3-restore-attempt'];
+    // We need to keep user metadata and tags
+    Object.keys(objMD).forEach(key => {
+        if (key.startsWith('x-amz-meta-')  && !userMDNotToKeep.includes(key)) {
```
francoisferrand:
```suggestion
        if (key.startsWith('x-amz-meta-') && !userMDNotToKeep.includes(key)) {
```

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,43 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+/**
+ * Keep metadatas when the object is restored from cold storage
+ * but remove the specific ones we don't want to keep
+ * @param {object} objMD - obj metadata
+ * @param {object} metadataStoreParams - custom built object containing resource details.
+ * @return {undefined}
+ */
+function restoreMetadatas(objMD, metadataStoreParams) {
+    /* eslint-disable no-param-reassign */
+    const userMDNotToKeep = ['x-amz-meta-scal-s3-restore-attempt'];
```
francoisferrand: NotToKeep → Skip ?

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,43 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+    // We need to keep user metadata and tags
+    Object.keys(objMD).forEach(key => {
+        if (key.startsWith('x-amz-meta-')  && !userMDNotToKeep.includes(key)) {
+            metadataStoreParams.metaHeaders[key] = objMD[key];
+        }
+        if (key.startsWith('content-') && !userMDNotToKeep.includes(key)) {
```
francoisferrand: should not check for "userMDNotToKeep", since this is not a userMetadata...

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,43 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+ * @param {object} metadataStoreParams - custom built object containing resource details.
+ * @return {undefined}
+ */
+function restoreMetadatas(objMD, metadataStoreParams) {
+    /* eslint-disable no-param-reassign */
+    const userMDNotToKeep = ['x-amz-meta-scal-s3-restore-attempt'];
+    // We need to keep user metadata and tags
+    Object.keys(objMD).forEach(key => {
+        if (key.startsWith('x-amz-meta-')  && !userMDNotToKeep.includes(key)) {
+            metadataStoreParams.metaHeaders[key] = objMD[key];
```
francoisferrand, line 450: if we reach here, no point checking later (i.e. add `else` or `continue`)

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,43 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+        }
+        if (key.startsWith('content-') && !userMDNotToKeep.includes(key)) {
+            metadataStoreParams[key] = objMD[key];
+        }
+    });
+
+    if (objMD.replicationInfo?.status === 'COMPLETED') {
```
francoisferrand: shouldn't we copy the replication status even in the other cases?

* `PENDING` should probably not happen (as data would not have been available)... though it may be that we keep retrying or just failed but did not yet update the status → maybe we can print a warning, but we should not affect this workflow
* `FAILED` can definitely happen, and we should not make it look like anyhing changed...

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,43 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+        metadataStoreParams.replicationInfo = objMD.replicationInfo;
+    }
+
+    if (objMD.legalHold) {
+        metadataStoreParams.legalHold = objMD.legalHold;
+    }
+
+    if (objMD.acl) {
+        metadataStoreParams.oldAcl = objMD.acl;
```
francoisferrand: why `oldAcl`, and not just `acl` like the other?
(i don't see an already used `acl` field in here, but maybe i missed it)

tests/unit/api/apiUtils/versioning.js
```diff
@@ -616,4 +619,345 @@ describe('versioning helpers', () => {
...
+                'versionId': '2345678',
+                'creation-time': now,
+                'last-modified': now,
+                'originOp': 's3:PutObject',
+                archive: {
+                    'restoreRequestedDays': days,
+                    'restoreRequestedAt': now,
+                    archiveInfo
+                } },
+                expectedRes: {
```
francoisferrand: missing indent:

```suggestion
                    'versionId': '2345678',
                    'creation-time': now,
                    'last-modified': now,
                    'originOp': 's3:PutObject',
                    archive: {
                        'restoreRequestedDays': days,
                        'restoreRequestedAt': now,
                        archiveInfo
                    },
                },
                expectedRes: {
```

francoisferrand (commented):
Due to node.js "untyped" quality, the unit test do not ensure the code works : need to add extra "functional" tests in `scality/cloudserver/tests/functional/aws-node-sdk/test/object/putVersion.js`, which will ensure that:
* update versioning is indeed called
* the objMD it gets is indeed correctly filled
* the resulting "metadataObject" has the field which match what is required by `metadataStoreObject`

lib/api/apiUtils/object/versioning.js
KillianG: acl is a global variable in service.js because we require it
`const acl = require('./metadata/acl');`

I wouldn't be able to create a variable like this if the field was named acl
```
const { objectKey, authInfo, size, contentMD5, metaHeaders,
            contentType, cacheControl, contentDisposition, contentEncoding,
            expires, multipart, headers, overrideMetadata, log,
            lastModifiedDate, versioning, versionId, uploadId,
            tagging, taggingCopy, replicationInfo, defaultRetention,
            dataStoreName, creationTime, retentionMode, retentionDate,
            legalHold, originOp, updateMicroVersionId, archive, oldReplayId,
            deleteNullKey, oldAcl } = params;
```

That said, I can still name it acl and use it as `params.acl`

What do you prefer?

francoisferrand: personally, I would prefer to keep the name of the field clean (and possibly using another name when importing the package, to avoid problems)

not sure what others think though, if anyone knows "good practice" for this case, please share!

lib/services.js
```diff
@@ -128,12 +128,11 @@ const services = {
...
-            md.setAmzServerSideEncryption(cipherBundle.algorithm);
-            // configuredMasterKeyId takes precedence
-            if (cipherBundle.configuredMasterKeyId || cipherBundle.masterKeyId) {
-                md.setAmzEncryptionKeyId(cipherBundle.configuredMasterKeyId || cipherBundle.masterKeyId);
-            }
+
+        if (cipherBundle || params.cipherBundle) {
+            const bundle = cipherBundle || params.cipherBundle;
+            md.setAmzServerSideEncryption(bundle.algorithm);
+            md.setAmzEncryptionKeyId(bundle.configuredMasterKeyId || bundle.masterKeyId);
```
francoisferrand: should we not keep the test which was present before?
e.g. do not set `AmzEncryptionKeyId` to null-ish value?

francoisferrand (approved):

tests/unit/api/apiUtils/versioning.js
```diff
@@ -616,4 +619,346 @@ describe('versioning helpers', () => {
...
+                    'updateMicroVersionId': true,
+                    'originOp': 's3:ObjectRestore:Completed',
+                    'metaHeaders': {
+                        'x-amz-meta-test': 'test',
+                        'x-amz-meta-test2': 'test2',
+                    },
+                    taggingCopy: { 'testtag': 'testtag', 'testtag2': 'testtag2' },
+                    archive: {
+                        archiveInfo,
+                        'restoreRequestedDays': 3,
```
KazToozs: could we not use the `days` variable here?

tests/unit/api/apiUtils/versioning.js
```diff
@@ -616,4 +619,346 @@ describe('versioning helpers', () => {
...
+                        'restoreRequestedAt': now,
+                        'restoreCompletedAt': new Date(now),
+                        'restoreWillExpireAt': new Date(now + (days * scaledMsPerDay)),
+                    }
+                },
+            },
+            {
+                description: 'Should return nullVersionId',
```
KazToozs: I'm not sure this test title is explicit enough

francoisferrand (changes requested):
Code change looks good (just a few details, c.f. comments), but missing functional tests

lib/services.js
```diff
@@ -128,12 +128,11 @@ const services = {
...
             md.setLastModified(lastModifiedDate);
         }
-        if (cipherBundle) {
-            md.setAmzServerSideEncryption(cipherBundle.algorithm);
-            // configuredMasterKeyId takes precedence
-            if (cipherBundle.configuredMasterKeyId || cipherBundle.masterKeyId) {
-                md.setAmzEncryptionKeyId(cipherBundle.configuredMasterKeyId || cipherBundle.masterKeyId);
-            }
+
+        if (cipherBundle || params.cipherBundle) {
```
nicolas2bert: Why don't we always use the `cipherBundle` generated from the request header?
Using the "copied" `params.cipherBundle` creates data discrepancy because the data is being encrypted with the cipher bundle generated from the request header.

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,57 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+            metadataStoreParams.headers = {};
+        }
+        metadataStoreParams.headers['x-amz-website-redirect-location'] = objMD['x-amz-website-redirect-location'];
+    }
+
+    if (objMD.replicationInfo) {
+        metadataStoreParams.replicationInfo = objMD.replicationInfo;
+    }
+
+    if (objMD.legalHold) {
```
nicolas2bert, line 465: What about the retention date and mode?

tests/functional/aws-node-sdk/test/utils/init.js
```diff
@@ -79,6 +79,31 @@ function fakeMetadataArchive(bucketName, objectName, versionId, archive, cb) {
         /* eslint-disable no-param-reassign */
         objMD.dataStoreName = 'location-dmf-v1';
         objMD.archive = archive;
+        objMD['x-amz-meta-custom-user-md'] = 'custom-md';
```
nicolas2bert: acl, legalHold (and potentially retention date and mode) not tested?

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,57 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+ */
+function restoreMetadatas(objMD, metadataStoreParams) {
+    /* eslint-disable no-param-reassign */
+    const userMDToSkip = ['x-amz-meta-scal-s3-restore-attempt'];
+    // We need to keep user metadata and tags
+    Object.keys(objMD).forEach(key => {
+        if (key.startsWith('x-amz-meta-') && !userMDToSkip.includes(key)) {
+            metadataStoreParams.metaHeaders[key] = objMD[key];
+        } else if (key.startsWith('content-')) {
+            metadataStoreParams[key] = objMD[key];
```
nicolas2bert: when is `metadataStoreParams['content-xxx']` used?

lib/api/apiUtils/object/versioning.js
```diff
@@ -434,6 +434,57 @@ function preprocessingVersioningDelete(bucketName, bucketMD, objectMD, reqVersio
...
+        }
+    });
+
+    if (objMD['x-amz-server-side-encryption']) {
```
nicolas2bert: We should not "copy" any metadata that is data-related (data encryption, data size...).

lib/api/apiUtils/object/versioning.js
KillianG: Retention date and mode are bucket level so we don't need to manage them here

nicolas2bert:
Much cleaner. 
Thanks. 
Last request for me, can we add tests for `should not "copy" any metadata that is data-related (data encryption, data size...)`?

nicolas2bert (approved):

tests/functional/aws-node-sdk/test/object/putVersion.js
```diff
@@ -89,8 +89,13 @@ describe('PUT object with x-scal-s3-version-id header', () => {
                 next => s3.putObject(params, next),
                 next => fakeMetadataArchive(bucketName, objectName, undefined, archive, next),
                 next => getMetadata(bucketName, objectName, undefined, (err, objMD) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    // eslint-disable-next-line no-param-reassign
+                    objMD.legalHold = true;
```
francoisferrand: why add 'legalHold' to each object?
* we probably want to the test both cases (overwrite with and without legal hold)
* setting legal hold should be done via regular s3 api, not by tweaking the API

tests/functional/aws-node-sdk/test/object/putVersion.js
```diff
@@ -89,8 +89,13 @@ describe('PUT object with x-scal-s3-version-id header', () => {
...
                     objMDBefore = objMD;
-                    return next(err);
+                    return metadata.putObjectMD(bucketName, objectName, objMD, params, log, err => next(err));
```
francoisferrand: should not overwrite metadata manually, this breaks the semantics of the scenario, no?

tests/functional/aws-node-sdk/test/object/mpuVersion.js
```diff
@@ -847,5 +847,87 @@ describe('MPU with x-scal-s3-version-id header', () => {
...
+                    objMD['x-amz-meta-custom-user-md'] = 'custom-md';
+                    objMD['x-amz-website-redirect-location'] = 'custom-md';
+                    // data related
+                    objMD['content-length'] = 99;
+                    objMD['content-type'] = 'testtype';
+                    objMD['content-md5'] = 'testmd 5';
+                    objMD['content-encoding'] = 'testencoding';
+                    objMD['x-amz-server-side-encryption'] = 'aws:kms';
+                    /* eslint-enable no-param-reassign */
+                    return metadata.putObjectMD(bucketName, objectName, objMD, undefined, log, next);
```
francoisferrand: can we use the proper AWS api to set the metadata fields?

this is _functional_ test (not unit test), so it would be better to reproduce the 'real' use case... (though we may indeed putObjectMD to fake the 'archival' of the object)

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -1271,25 +1279,36 @@ describe('Object Copy', () => {
...
+            };
+            async.waterfall([
+                next => getMetadata(sourceBucketName, sourceObjName, undefined, (err, objMD) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    /* eslint-disable no-param-reassign */
+                    objMD.dataStoreName = 'location-dmf-v1';
+                    objMD.archive = archiveCompleted;
+                    objMD['x-amz-meta-custom-user-md'] = 'custom-md';
```
francoisferrand: same here

tests/functional/aws-node-sdk/test/object/putVersion.js
```diff
@@ -753,5 +753,88 @@ describe('PUT object with x-scal-s3-version-id header', () => {
...
+                    objMD.legalHold = true;
+                    objMD['x-amz-meta-custom-user-md'] = 'custom-md';
+                    objMD['x-amz-website-redirect-location'] = 'custom-md';
+                    // data related
+                    objMD['content-length'] = 99;
+                    objMD['content-type'] = 'testtype';
+                    objMD['content-md5'] = 'testmd 5';
+                    objMD['content-encoding'] = 'testencoding';
+                    objMD['x-amz-server-side-encryption'] = 'aws:kms';
+                    /* eslint-enable no-param-reassign */
```
francoisferrand, line 879: same

francoisferrand (approved):

PR #5231, opened 2023-07-12
CLDSRV-412 Test null version in Lifecycle list of non-current versions

nicolas2bert:
Testing this change: https://github.com/scality/Arsenal/pull/2150

### DIFF
--- a/tests/functional/backbeat/listDeleteMarker.js
+++ b/tests/functional/backbeat/listDeleteMarker.js
@@ -4,17 +4,18 @@ const BucketUtility = require('../aws-node-sdk/lib/utility/bucket-util');
 const { removeAllVersions } = require('../aws-node-sdk/lib/utility/versioning-util');
 const { makeBackbeatRequest, runIfMongoV1 } = require('./utils');
 
-const testBucket = 'bucket-for-list-lifecycle-null-tests';
-
 const credentials = {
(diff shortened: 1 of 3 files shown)


alexanderchan-scality (approved):

tests/functional/backbeat/listDeleteMarker.js
```diff
@@ -0,0 +1,208 @@
...
+});
+
+runIfMongoV1('listLifecycle with current delete marker version', () => {
+    let bucketUtil;
+    let s3;
+    let expectedVersionId;
+    const testBucket = 'bucket-for-list-lifecycle-current-dm-tests';
+    const keyName = 'key0';
+
+    before(done => {
```
naren-rajendran, line 123: small nitpick :) may be `before` and `after` sections can be reused (convert into a generic fn perhaps) for both these test groups with an option to `delete` object. This is no way a necessity, just an observation.

naren-rajendran (approved):

tests/functional/backbeat/listDeleteMarker.js
nicolas2bert: the “before” functions are different. Not sure we can “generate” them

nicolas2bert:
@bert-e approve

PR #5272, opened 2023-08-11
Bugfix/cldsrv 413/crr existing null version

tmacro:
~This uses preProcessVersioning to handle `null` versions in backbeat putMetadata route.
It handles converting pre-versioning objects to have a versionId so that they can be dealt with as normal versioned objects.~

Much simplified fix by setting `isNull` in the incoming object metadata to be put.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -482,6 +482,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             // Retrieve the null version id from the object metadata.
             versionId = objMd && objMd.versionId;
             if (!versionId) {
+                // Set isNull in the object metadata to be written.
+                // Since metadata will generate a versionId for the null version,
+                // the flag is needed to allow cloudserver to know that the version
+                // is a null version and allow access to it using the "null" versionId.
+                omVal.isNull = true;
                 if (versioning) {
                     // If the null version does not have a version id, it is a current null version.
                     // To update the metadata of a current version, versioning is set to false.
@@ -501,7 +506,6 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     // In such scenarios, we generate a new null version and designate it as the master version.
                     if (versioningConf && versioningConf.Status === 'Suspended') {
                         versionId = '';
-                        omVal.isNull = true;
                     }
                 }
             }
--- a/tests/functional/aws-node-sdk/lib/utility/bucket-util.js
+++ b/tests/functional/aws-node-sdk/lib/utility/bucket-util.js
@@ -14,6 +14,18 @@ class BucketUtility {
         });
     }
 
+    bucketExists(bucketName) {
+        return this.s3
+            .headBucket({ Bucket: bucketName }).promise()
+            .then(() => true)
+            .catch(err => {
+                if (err.code === 'NotFound') {
+                    return false;
+                }
+                throw err;
+            });
+    }
+
     createOne(bucketName) {
         return this.s3
             .createBucket({ Bucket: bucketName }).promise()
@@ -116,6 +128,24 @@ class BucketUtility {
     emptyMany(bucketNames) {
         const promises = bucketNames.map(
             bucketName => this.empty(bucketName)
+        );
+
+        return Promise.all(promises);
+    }
+
+    emptyIfExists(bucketName) {
+        return this.bucketExists(bucketName)
+            .then(exists => {
+                if (exists) {
+                    return this.empty(bucketName);
+                }
+                return undefined;
+            });
+    }
+
+    emptyManyIfExists(bucketNames) {
+        const promises = bucketNames.map(
+            bucketName => this.emptyIfExists(bucketName)
         );
 
         return Promise.all(promises);
--- a/tests/functional/raw-node/test/routes/routeBackbeat.js
+++ b/tests/functional/raw-node/test/routes/routeBackbeat.js
@@ -156,7 +156,8 @@ describeSkipIfAWS('backbeat routes', () => {
         bucketUtil = new BucketUtility(
             'default', { signatureVersion: 'v4' });
         s3 = bucketUtil.s3;
-        s3.createBucket({ Bucket: TEST_BUCKET }).promise()
+        bucketUtil.emptyManyIfExists([TEST_BUCKET, TEST_ENCRYPTED_BUCKET, NONVERSIONED_BUCKET])
+            .then(() => s3.createBucket({ Bucket: TEST_BUCKET }).promise())
             .then(() => s3.putBucketVersioning(
                 {
                     Bucket: TEST_BUCKET,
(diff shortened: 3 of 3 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -474,107 +474,98 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
 
-        if (versionId === 'null') {
-            isNull = true;
-            // Retrieve the null version id from the object metadata.
-            versionId = objMd && objMd.versionId;
-            if (!versionId) {
-                if (versioning) {
-                    // If the null version does not have a version id, it is a current null version.
-                    // To update the metadata of a current version, versioning is set to false.
+        return versioningPreprocessing(bucketName, bucketInfo, objectKey, omVal, log, (err, options) => {
```
jonathan-gramain: Normally, unless I'm mistaken, this function is never called for version-specific PUTs, it is responsible to process the versioning state for new version PUTs.

According to jsdoc:
> return versioning information for S3 to handle creation of new versions and manage deletion of old data and metadata

Indeed its side-effects are to either copy the null version to a new key or delete it, depending on circumstamces. I'm not sure this is what we want here.

I would rather only call this function if we are indeed doing a non-version-specific PUT to avoid side-effects like this (and do the right thing when indeed we are doing such operation).

Are the existing tests enough to cover this by the way?

jonathan-gramain (changes requested):

lib/routes/routeBackbeat.js
tmacro: This existing test covers it https://github.com/scality/cloudserver/blob/bfa366cd272090084e41fd0b9d37e5a32df6c57f/tests/functional/raw-node/test/routes/routeBackbeat.js#L287

tmacro: What do you think would be the correct way of handling this?

jonathan-gramain: I'm thinking of a test that
- updates a specific version of an object on a versioning-suspended bucket
- the same object also having a null version

I have the feeling in such case the null version would get deleted erroneously due to the use of `versioningPreprocessing` for a version-specific update.

I suggest to not use `versioningPreprocessing` when it is a version-specific update, because the function is not meant for this use.

By the way, is there a need to support regular versioned PUTs (or nonversioned) in the backbeat route? Maybe we don't need `versioningPreprocessing` after all?

tmacro: At least for 7.x branches the API will fail your request with a 409 for version suspended buckets, I'd need to check the behavior in 8.x.
7.x branches don't need to support non-versioned, but 8.x branches do. (based on the test comments)
So instead of `versioningPreprocessing` should I just implement the behavior in the handler directly? or perhaps break it out into `versioning.js`?

jonathan-gramain: @tmacro yes, implementing the logic directly in the handler should be fine (I'm not aware of duplicate code in Cloudserver but maybe it exists). Also good to use helpers in `versioning.js` if that makes sense and is not too specific to backbeat.

francoisferrand: We had lots of issues with null versions and various use cases on 8.x, I think now we should be good in most cases in there : so you may want to check how it was done there indeed, and possibly just backport some changes/commits...

tmacro: @jonathan-gramain I;ve pushed some changes

lib/routes/routeBackbeat.js
```diff
@@ -482,6 +482,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             // Retrieve the null version id from the object metadata.
             versionId = objMd && objMd.versionId;
             if (!versionId) {
+                omVal.isNull = true;
                 if (versioning) {
                     // If the null version does not have a version id, it is a current null version.
```
jonathan-gramain, line 491: Out of context, this comment line looks wrong since a nonversioned object is not necessarily the current version (doesn't mean the code is wrong).

lib/routes/routeBackbeat.js
```diff
@@ -482,6 +482,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             // Retrieve the null version id from the object metadata.
             versionId = objMd && objMd.versionId;
             if (!versionId) {
+                omVal.isNull = true;
```
jonathan-gramain, line 489: Could be worth commenting what this does or is meant to do as it does not look obvious

lib/routes/routeBackbeat.js
tmacro: In all the cases I could think of the non-versioned was converted to a version + master the next time an object was PUT.

jonathan-gramain: OK, but it's nevertheless incorrect to update the master if the version is not the current version, we need to guard against this.

tmacro: I've poked around a bit more and it doesn't seem to be able to have a non-versioned (internal version) object that is also not the master in a versioned bucket. In every case the object would have a versionId so the check on line 484 ensures this is the case.

jonathan-gramain: Yes that's right, the nonversioned object is appended the "infinite" version ID when new versions are created on top.

jonathan-gramain: > Out of context, this comment line looks wrong since a nonversioned object is not necessarily the current version (doesn't mean the code is wrong).

So regarding my statement above, indeed it's the current version so the code comment is right, but we could maybe go further and say that it can only be a *unique* version because as soon as new versions are created, Cloudserver converts the nonversioned object to a versioned one with an infinite version ID and the `isNull` flag.

lib/routes/routeBackbeat.js
tmacro: I agree, I'm just still looking for the piece of code that this change affects downstream.

lib/routes/routeBackbeat.js
rachedbenmustapha: And just to be sure, the only way, ever, to get an infinite version id, is the explicit conversion of an object from nonversioned-style to versioned-style through a manual tool?

tmacro: @RachedBenMustapha you get them during normal operation.
You get one for example if you
- create bucket
- put object
- enable bucket versioning
- put object again

The initial object will be assigned the infinite versionId during the second PUT

lib/routes/routeBackbeat.js
tmacro: Added a comment, but posting it here as well for visibility.
The piece of code that changes behavior is [here](https://github.com/scality/cloudserver/blob/328b7bc335f10b8e88dd9553884507ad96df77c5/lib/metadata/metadataUtils.js#L20). During a get object for the "null" versionId cloudserver fetches the master version and determines whether it is a null version using the check
```
objMD.isNull || !objMD.versionId
``` 
Without this fix after after the update you would be left with an object that now has an versionId with `isNull` as undefined.
This would prevent cloudserver from recognizing this as a null version and cause it to return a 404.

rachedbenmustapha (approved):

lib/routes/routeBackbeat.js
```diff
@@ -482,6 +482,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             // Retrieve the null version id from the object metadata.
             versionId = objMd && objMd.versionId;
             if (!versionId) {
+                // Set isNull in the object metadata to be written.
+                // Since metadata will generate a versionId for the null version,
+                // the flag is need to allow cloudserver to know that the version
```
jonathan-gramain:
```suggestion
                // the flag is needed to allow cloudserver to know that the version
```

jonathan-gramain (approved):

PR #5284, opened 2023-08-18
Test interactions with sproxyd and integrate SPRXCLT-12

fmonjalet:
This PR sets up and uses an `sproxyd` container for cloudserver's multiple backend tests. This allows basic cloudserver / sproxyd integration testing at this level, and includes a non regression test for `SPRXCLT-12`.

It also reenables a number of disabled tests.

### DIFF
--- a/.github/docker/sproxyd/conf/fastcgi_params
+++ b/.github/docker/sproxyd/conf/fastcgi_params
@@ -1,0 +1,26 @@
+fastcgi_param  QUERY_STRING       $query_string;
+fastcgi_param  REQUEST_METHOD     $request_method;
+fastcgi_param  CONTENT_TYPE       $content_type;
+fastcgi_param  CONTENT_LENGTH     $content_length;
+
+#fastcgi_param  SCRIPT_NAME        $fastcgi_script_name;
+fastcgi_param  SCRIPT_NAME        /var/www;
+fastcgi_param  PATH_INFO          $document_uri;
+
+fastcgi_param  REQUEST_URI        $request_uri;
+fastcgi_param  DOCUMENT_URI       $document_uri;
+fastcgi_param  DOCUMENT_ROOT      $document_root;
+fastcgi_param  SERVER_PROTOCOL    $server_protocol;
+fastcgi_param  HTTPS              $https if_not_empty;
+
+fastcgi_param  GATEWAY_INTERFACE  CGI/1.1;
+fastcgi_param  SERVER_SOFTWARE    nginx/$nginx_version;
+
+fastcgi_param  REMOTE_ADDR        $remote_addr;
+fastcgi_param  REMOTE_PORT        $remote_port;
+fastcgi_param  SERVER_ADDR        $server_addr;
+fastcgi_param  SERVER_PORT        $server_port;
+fastcgi_param  SERVER_NAME        $server_name;
+
+# PHP only, required if PHP was built with --enable-force-cgi-redirect
+fastcgi_param  REDIRECT_STATUS    200;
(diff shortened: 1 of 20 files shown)


.github/docker/sproxyd/conf/fastcgi_params
francoisferrand, line 1: shoud we add all these config file to a "sproxyd test" image, so that is easy as well to deploy locally?

could be just a simple Dockerfile:

```
FROM sproxyd
ADD factcgi_params
ADD nginx.conf
RUN chown root:root ...
```

and we can add it as well as `build:` line in the docker-compose (in addition to the `image:`), to allow autoamtically building when needed, even if not pushed to the registry

tests/multipleBackend/objectPut.js
```diff
@@ -71,74 +74,99 @@ function put(bucketLoc, objLoc, requestHost, cb, errorDescription) {
...
-        put(null, null, 'localhost', () => {
-            assert.deepStrictEqual(ds, []);
-            done();
+    putCases.forEach(testCase => {
+        it(`should put an object to ${testCase.name}`, done => {
+            async.series([
+                next => put(testCase.bucketLoc, testCase.objLoc, 'localhost', () => {
+                    checkPut(testCase);
+                    next();
+                }),
```
francoisferrand: nit:
```suggestion
                next => put(testCase.bucketLoc, testCase.objLoc, 'localhost', next),
                next => {
                    checkPut(testCase);
                    next();
                }),
```

(and we may then merge it with the next `setTimeout` to simplify further)

francoisferrand (approved):

KillianG (approved):

tests/multipleBackend/objectPut.js
```diff
@@ -75,78 +75,78 @@ function put(bucketLoc, objLoc, requestHost, cb, errorDescription) {
...
-    });
-
-    it('should put an object to Azure based on bucket location', done => {
-        put('azurebackend', null, 'localhost', () => {
-            assert.deepStrictEqual(ds, []);
-            done();
-        });
-    });
-
-    it('should put an object to Azure based on object location', done => {
```
jonathan-gramain, line 130: Not sure if this was wanted but this test doesn't seem to be carried over

tests/multipleBackend/objectPut.js
```diff
@@ -157,6 +157,10 @@ describeSkipIfE2E('objectPutAPI with multiple backends', function testSuite() {
                     checkPut(testCase);
                     next();
                 }),
+                // Increase the probability of the first request having released
+                // the socket, so that it can be reused for the next request.
+                // This tests how HTTP connection reuse behaves.
+                next => setTimeout(next, 10),
```
jonathan-gramain: I am a bit surprised that 10ms is enough for the socket to be closed, but maybe :)

jonathan-gramain (approved):

.github/docker/sproxyd/conf/fastcgi_params
fmonjalet: Good idea, I did something in this spirit, tell me if it is what you expected.

tests/multipleBackend/objectPut.js
fmonjalet: I believe I just moved it somewhere else for consistency, it should be the `'azure'` entry of `putCases`. Tell me if I missed anything.

tests/multipleBackend/objectPut.js
fmonjalet: It's not about closing the socket, but having the first request not use the socket anymore, so that the second can use it. It's pure js logic, not OS. This very small timeout seems to leave the time for the code responsible for this to be scheduled and release the socket, therefore making socket reuse much more frequent (checked by looking at nginx logs). Otherwise, the second request often uses a new socket instead of the existing one, defeating the purpose of the test.

PR #5211, opened 2023-06-20
Task/s3 c 7845 poc bucket policies

ghost:
WIP

PR #5291, opened 2023-08-23
CLDSRV-426: update ACL permission checks for implicitDeny logic

KazToozs:
# Pull request template

### Motivation and context

Bucket policies are not correctly interpreted, this is part of the following epic to fix that: https://github.com/scality/Arsenal/pull/2162

Related issues
PRs providing implicit Deny logic to CS for processing in this PR
https://github.com/scality/Arsenal/pull/2162
https://github.com/scality/Vault/pull/2134
https://github.com/scality/cloudserver/pull/5288

### Sign your work

as soon as we can.


lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -8,13 +8,33 @@ const { allAuthedUsersId, bucketOwnerActions, logId, publicId,
...
     }
+    // Backward compatibility
+    const arrayOfAllowed = [
+        'objectGetTagging',
+        'objectPutTagging',
+        'objectPutLegalHold',
+        'objectPutRetention',
+    ];
+    if (mainApiCall === 'objectGet') {
+        if (arrayOfAllowed.includes(requestTypeParsed)) {
```
francoisferrand: if mainApiCall is objectGet, we should only accept requestTypeParsed == objectGetTagging, right?

(having the list for checking the different types of requests for objectPut is fine, this is really just about the "mixing" the Get and Put...)

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -8,13 +8,33 @@ const { allAuthedUsersId, bucketOwnerActions, logId, publicId,
...
 
-function checkBucketAcls(bucket, requestType, canonicalID) {
+function checkBucketAcls(bucket, requestType, canonicalID, mainApiCall) {
+    // Same logic applies on the Versioned APIs, so let's simplify it.
+    const requestTypeParsed = requestType.endsWith('Version') ?
+        requestType.slice(0, -7) : requestType;
     if (bucket.getOwner() === canonicalID) {
         return true;
     }
+    // Backward compatibility
```
francoisferrand, line 18: what do you mean by "Backward compatibility": are you talking about keeping the "older" behavior (but i don't see the removed/updated code which would do it) of this code?

francoisferrand (commented):
a few cases were changed, but i don't see matching tests update : please complete the tests to fully cover this.

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Per this comment in the original PoC
https://github.com/scality/cloudserver/pull/5211/files#diff-773344fcfc0f64145ff95eb6eccd556588c19449dc2c263a2532d4030bf00dffR425

These APIs which have extra permissions are allowed by default to maintain backward compatibility with current behaviour as I understand it

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Yes seperating these makes sense, I'll change this.
92cbbdb44f32f738006ff3523641e2de7c65dd9d

KazToozs:
@francoisferrand I don't see existant tests for the changes made to these functions specifically: given they are always called within the contexts of `isBucketAuthorized`/`isObjectAuthorized`, does it not make more sense to apply the changes to those tests in the following PR where changes are made to those functions?

francoisferrand:
> I don't see existant tests for the changes made to these functions specifically: given they are always called within the contexts of `isBucketAuthorized`/`isObjectAuthorized`, does it not make more sense to apply the changes to those tests in the following PR where changes are made to those functions?

Testing the lower layers first before building on top is the way to go : this way you ensure things will not crumble once you build the next layer, you can test all corner cases, and you make sure _this PR_ works (vs making sure the code eventually works after a few PR)

It is unfortunate that there are not tests already, but then that is an opportunity to fill the gap and reduce (ever so slightly) the test debt... and you can do it in a test-first manner : add the tests on the existing code, then make your changes to the target function and confirm it works/see what changes.

KazToozs:
I added test cases, particularly for the new changes, let me know if they're sufficient

PR #5341, opened 2023-09-15
CLDSRV-447:Test add for principal && arsn bump

benzekrimaha:
### Motivation and context

In this PR the principal item user arn length issue is solved to match with vault's

*Currently arsenal version is the commit hash on arsenal's PR - needs to be changed before merge*
 
Ticket Link : https://scality.atlassian.net/browse/CLDSRV-447

New test added


benzekrimaha:
ping

PR #5354, opened 2023-09-25
CLDSRV-449: overheadField support

tmacro:
This adds support for a new parameter overheadField` to passed to the metadata backend.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -185,6 +185,12 @@ const constants = {
         ORPHAN_DM_TYPE: 'orphan',
     },
     maxScannedLifecycleListingEntries: 10000,
+    overheadField: [
+        'content-length',
+        'owner-id',
+        'versionId',
+        'isNull',
+    ],
 };
 
 module.exports = constants;
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -77,6 +77,7 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
  * @param {(object|null)} streamingV4Params - if v4 auth, object containing
  * accessKey, signatureFromRequest, region, scopeDate, timestamp, and
  * credentialScope (to be used for streaming v4 auth if applicable)
+ * @param {(object|null)} overheadField - fields to be included in metadata overhead
  * @param {RequestLogger} log - logger instance
  * @param {function} callback - callback function
  * @return {undefined} and call callback with (err, result) -
@@ -85,7 +86,7 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
  */
 function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         canonicalID, cipherBundle, request, isDeleteMarker, streamingV4Params,
-        log, callback) {
+        overheadField, log, callback) {
     const size = isDeleteMarker ? 0 : request.parsedContentLength;
     // although the request method may actually be 'DELETE' if creating a
     // delete marker, for our purposes we consider this to be a 'PUT'
@@ -137,6 +138,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         headers,
         isDeleteMarker,
         replicationInfo: getReplicationInfo(objectKey, bucketMD, false, size, null, null, authInfo, isDeleteMarker),
+        overheadField,
         log,
     };
 
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -25,6 +25,7 @@ const requestUtils = policies.requestUtils;
 const { data } = require('../data/wrapper');
 const logger = require('../utilities/logger');
 const { _bucketRequiresOplogUpdate } = require('./apiUtils/object/deleteObject');
+const { overheadField } = require('../../constants');
 
 const versionIdUtils = versioning.VersionID;
 
@@ -334,6 +335,7 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                         bucketName, bucket, objMD, versionId, config.nullVersionCompatMode);
                     const deleteInfo = {};
                     if (options && options.deleteData) {
+                        options.overheadField = overheadField;
                         deleteInfo.deleted = true;
                         if (!_bucketRequiresOplogUpdate(bucket)) {
                             options.doesNotNeedOpogUpdate = true;
@@ -357,7 +359,7 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                     // This call will create a delete-marker
                     return createAndStoreObject(bucketName, bucket, entry.key,
                         objMD, authInfo, canonicalID, null, request,
-                        deleteInfo.newDeleteMarker, null, log, (err, result) =>
+                        deleteInfo.newDeleteMarker, null, overheadField, log, (err, result) =>
                         callback(err, objMD, deleteInfo, result.versionId));
                 },
             ], (err, objMD, deleteInfo, versionId) => {
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -314,6 +314,10 @@ function objectCopy(authInfo, request, sourceBucket,
                         dataStoreContext.metaHeaders =
                           storeMetadataParams.metaHeaders;
                     }
+
+                    // eslint-disable-next-line no-param-reassign
+                    storeMetadataParams.overheadField = constants.overheadField;
+
                     let dataLocator;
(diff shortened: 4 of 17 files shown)


lib/api/multiObjectDelete.js
```diff
@@ -27,7 +27,10 @@ const logger = require('../utilities/logger');
 const { _bucketRequiresOplogUpdate } = require('./apiUtils/object/deleteObject');
 
 const versionIdUtils = versioning.VersionID;
-
+const overheadField = [
```
jonathan-gramain: I think this list should go in this file https://github.com/scality/cloudserver/blob/development/7.70/constants.js and imported from all locations that need it.

lib/api/multiObjectDelete.js
```diff
@@ -357,7 +361,7 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                     // This call will create a delete-marker
                     return createAndStoreObject(bucketName, bucket, entry.key,
                         objMD, authInfo, canonicalID, null, request,
-                        deleteInfo.newDeleteMarker, null, log, (err, result) =>
+                        deleteInfo.newDeleteMarker, null, null, log, (err, result) =>
```
jonathan-gramain: So delete markers don't have overhead fields at all?

I am thinking of the particular case of versioning-suspended buckets: in this case, a delete marker can override an existing non-deleted null version, stored in the master key, where I think we should take this into account as well. Same for other locations where we create delete markers.

lib/api/objectCopy.js
```diff
@@ -314,6 +319,10 @@ function objectCopy(authInfo, request, sourceBucket,
                         dataStoreContext.metaHeaders =
                           storeMetadataParams.metaHeaders;
                     }
+
+                    // eslint-disable-next-line
```
ghost: Maybe `eslint-disable-next-line no-param-reassign` is better here?

lib/api/multiObjectDelete.js
ghost: I agree. But in case we need some kind of granularity between the methods where we want to pass these headers, how about an helper method, or dictionnary? Of course this only makes sense if at some point different methods sends different headers, but that's what I saw on the main RING ticket.

lib/services.js
```diff
@@ -746,7 +750,13 @@ const services = {
...
             'content-length': size,
         };
-        metadata.putObjectMD(mpuBucketName, partKey, omVal, {}, log, err => {
+
+        const params = {};
+        if (overheadField) {
+            params.overheadField = overheadField;
+        }
+
+        metadata.putObjectMD(mpuBucketName, partKey, omVal, params, log, err => {
```
ghost, line 759: it feels strange having a mix of calls where some recreate objects with the `overheadField` key, and some just pass the array, and here we create an object outside, how about
```suggestion
        metadata.putObjectMD(mpuBucketName, partKey, omVal, { overheadField }, log, err => {
```

Or change in https://github.com/scality/cloudserver/pull/5354/files#diff-9c3967d9affbf1081dd9c5593beebe425aafd67dfbbb96b2ebd900c5dc3019d8R318 to be consistent?

tests/unit/api/objectDelete.js
```diff
@@ -197,6 +201,7 @@ describe('objectDelete API', () => {
                             sinon.match.any,
                             sinon.match.any
                         );
+                        // sinon.assert.calledWith(
```
ghost: Maybe something to remove

lib/services.js
ghost: I also have a side question, I am trying to understand the context of these changes (I found nothing on Jira), why do we need to pass additional arguments as a new parameter while we have, here, for example, already the `omVal` variable that contains the content-length? Is it because from metadata, we do not know what are the expected headers for each API?

ghost (commented):
First review, I will refine it after I have a little bit more context :smile:

lib/services.js
tmacro: The high level reason is that some operations don't leave enough context in the oplog for some of the things we want to do with it. For example if a non-versioned object is deleted we can see what key, but the content-lenth or who owned the object is lost.
This param is used to ask Metadata to attach the "previous" value of the field to the oplog entry to provide this missing context.

Specifically on passing the fields we have 2 things going on:
1. Cloudserver backend ops are async and if it were to pass the data directly there is a chance that the data could change if there was a concurrent write to the same key that was processed first.
2. Metadata doesn't know mush about the structure of object metadata.

To work around 1 we need to push the work to Metadata where requests are ordered and we can make atomic guarantees about the "previous" value.
Since 2 is a nice property to maintain it we pass only the fields we want extracted and then have a generic  mechanism in Metadata do the retrieval. 

The full design is over here https://github.com/scality/citadel/blob/development/1.0/docs/design/raft-log-overhead.md

lib/api/multiObjectDelete.js
tmacro: I think your right wrt the suspended bucket case. I was expecting it to take the "nonversioned" code branch if versioning was suspended, but that doesn't seem to be the case.

lib/api/multiObjectDelete.js
tmacro: My main reason for declaring them separately was what @williamlardier mentioned, but they all ended up the same at the end. I'll move them to the constants file and we can add something more if we need it.

lib/services.js
ghost: Thank you very much for the detailed explanation, I read the design and now it makes sense :smile:

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -77,6 +77,7 @@ function _storeInMDandDeleteData(bucketName, dataGetInfo, cipherBundle,
  * @param {(object|null)} streamingV4Params - if v4 auth, object containing
  * accessKey, signatureFromRequest, region, scopeDate, timestamp, and
  * credentialScope (to be used for streaming v4 auth if applicable)
+ * @param {(object|null)} overheadField - fields to be included in metadata overhead
```
ghost, line 80: (Picking a random line)

Do we also need to support this new parameter in routeBackbeat, where we call the metadata putObjectMD function?
https://github.com/scality/cloudserver/blob/development/7.70/lib/routes/routeBackbeat.js#L554

I believe that we want to keep the same behavior during internal operation as well?

ghost (commented):
LGTM, after the common code for headers are put in the const file, and Jonathan's feedback is taken into account, thanks again for the explanation!
Please ping me once it's done for approval

lib/api/apiUtils/object/createAndStoreObject.js
tmacro: Good catch, Backbeat definitely needs to be supported

tmacro:
@williamlardier I believe it should be good to go now.
I'm holding off on fixing the conflicts until the final squash as there's quite a bit, and the code doesn't really change

ghost (approved):
LGTM

fmonjalet (commented):
About this approach: I am still worried relying on the client specifying the required fields will cause issues in the future. These fields **must** be added systematically for SCUBA to work properly.

- The current approach relies on the client to behave properly for the oplog to be correct (from SCUBA point of view). It means that if any kind of client is not compliant, SCUBA will record inaccurate metrics. Any client can be anything that communicates with repd, **ever** (including TS, scripts and other parts of the product I am not aware of). It also means that any book keeping done internally to metadata (PHD handling) will have to track these somehow.
- Enforcing this behavior in Metadata ensures it is always applied and does not open a "bug window" by relying on (potentially unknown) other components.

This is why I still don't understand the motivation or the added value of adding these fields as HTTP headers. See https://scality.atlassian.net/wiki/spaces/OS/pages/2185592886/Additional+fields+in+the+op+log?focusedCommentId=2188214365 for more details. What I suggest is just adding these fields in Metadata configuration (or code directly) for safety.

constants.js
```diff
@@ -185,6 +185,10 @@ const constants = {
         ORPHAN_DM_TYPE: 'orphan',
     },
     maxScannedLifecycleListingEntries: 10000,
+    overheadField: [
+        'content-length',
+        'owner-id',
+    ],
```
fmonjalet: Could we also add `isNull` and `versionId`? (https://scality.atlassian.net/browse/RING-43323?focusedCommentId=356769)

tmacro:
@fmonjalet I tried out what you suggested with a quick POC, and I think it could work. It is not a drop in replacement though, and will need logic changes on the Metadata side to handle some of the weird edge cases. Since we're under some time pressure I think we should go ahead with what we have, then revisit and make changes as needed.

jonathan-gramain (approved):

nicolas2bert (approved):

PR #5365, opened 2023-09-28
CLDSRV-446 Limiting entries scanned during the lifecycle listing

nicolas2bert:
Implement the "New Limit on Scanned Entries Approach" for metadata filtering to improve response time and resource management.

**Description:**
Metadata filtering process might experience delays, impacting response times and server resources. To address this issue, we propose implementing the "New Limit on Scanned Entries Approach". 
It was currently implemented as an "Internal Timeout Approach": cf https://scality.atlassian.net/browse/S3C-7928

**Approach:**
The "New Limit on Scanned Entries Approach" involves setting a maximum number of entries to scan and filter before responding with up to 1000 entries. This approach aims to provide more predictable resource usage while potentially delivering more comprehensive results.

**Benefits:**
- Predictable resource usage.
- Potential for more comprehensive results.

### DIFF
--- a/lib/api/apiUtils/object/lifecycle.js
+++ b/lib/api/apiUtils/object/lifecycle.js
@@ -21,6 +21,7 @@ function processCurrents(bucketName, listParams, isBucketVersioned, list) {
         Name: bucketName,
         Prefix: listParams.prefix,
         MaxKeys: listParams.maxKeys,
+        MaxScannedLifecycleListingEntries: listParams.maxScannedLifecycleListingEntries,
         IsTruncated: !!list.IsTruncated,
         Marker: listParams.marker,
         BeforeDate: listParams.beforeDate,
@@ -77,6 +78,7 @@ function processNonCurrents(bucketName, listParams, list) {
         Name: bucketName,
         Prefix: listParams.prefix,
         MaxKeys: listParams.maxKeys,
+        MaxScannedLifecycleListingEntries: listParams.maxScannedLifecycleListingEntries,
         IsTruncated: !!list.IsTruncated,
         KeyMarker: listParams.keyMarker,
         VersionIdMarker: versionIdMarker,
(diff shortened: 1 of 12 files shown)


benzekrimaha (approved):
LGTM

lib/api/backbeat/listLifecycleCurrents.js
```diff
@@ -43,6 +43,18 @@ function listLifecycleCurrents(authInfo, locationConstraints, request, log, call
...
+    if (params['max-scanned-lifecycle-listing-entries']) {
+        const maxEntriesParams = Number.parseInt(params['max-scanned-lifecycle-listing-entries'], 10);
+        if (Number.isNaN(maxEntriesParams) || maxEntriesParams <= 0 ||
+        maxEntriesParams > maxScannedLifecycleListingEntries) {
+            monitoring.promMetrics(
+                'GET', bucketName, 400, 'listLifecycleCurrents');
+            return callback(errors.InvalidArgument);
+        }
+        maxScannedLifecycleListingEntries = maxEntriesParams;
+    }
```
francoisferrand: this block is duplicated, should introduce a function

lib/api/apiUtils/object/lifecycle.js
```diff
@@ -151,9 +154,37 @@ function getLocationConstraintErrorMessage(locationName) {
         `- ${locationName} - is not listed in the locationConstraint config`;
 }
 
+/**
+ * validateMaxScannedEntries - Validates and returns the maximum scanned entries value.
+ *
+ * @param {object} params - Query parameters
+ * @param {object} config - CloudServer configuration
+ * @param {number} min - Minimum number of entries to be scanned
```
francoisferrand, line 162: we don't have arbitrary values here : actually only two values (1 or 3), with a significant explanation for the "3":
> 3 is required as a minimum because we must scan at least three entries to determine version eligibility.
> Two entries representing the master key and the following one representing the non-current version.

So it seems to me we actually want to embed this knowledge in the function, with a much simpler parameter (like `currentOnly` ?)

francoisferrand (approved):

nicolas2bert:
@bert-e approve

PR #5378, opened 2023-10-05
CLDSRV-455 orphan delete marker list interruption skips processed key

nicolas2bert:
The key marker in the orphan delete marker listing response should match the last key in the response's key array. 
This ensures that the next listing begins after the key that has already been returned.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3",
-  "version": "7.70.26",
+  "version": "7.70.27",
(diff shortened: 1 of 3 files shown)


nicolas2bert:
@bert-e reset

rachedbenmustapha (approved):

francoisferrand (approved):

nicolas2bert:
@bert-e approve

PR #5403, opened 2023-10-30
bf(CLDSRV-463): Strictly validate checksum algorithm headers

tmacro:


constants.js
```diff
@@ -179,6 +179,18 @@ const constants = {
     assumedRoleArnResourceType: 'assumed-role',
     // Session name of the backbeat lifecycle assumed role session.
     backbeatLifecycleSessionName: 'backbeat-lifecycle',
+    possibleSignatureChecksums: new Set([
```
jonathan-gramain: To avoid duplicates you could turn it into `unsupportedSignatureChecksums` and remove the supported ones, I think the behavior would stay the same.

lib/api/apiUtils/object/validateChecksumHeaders.js
```diff
@@ -0,0 +1,32 @@
...
+
+    if (supportedSignatureChecksums.has(signatureChecksum)) {
+        return null;
+    }
+
+    // If the value is not one of the possible checksum algorithms
+    // the only other valid value is the actual sha256 checksum of the payload.
+    // Do a simple sanity check of the length to guard against future algos.
+    // If the value is an unknown algo, then it will fail checksum validation.
+    if (!possibleSignatureChecksums.has(signatureChecksum) && signatureChecksum.length === 64) {
```
jonathan-gramain: I am surprised that the checksum should be 64 characters in length, because that would imply that it is always passed in hexadecimal encoding, but the documentation states that it should be in base64, this is what it says about the header `x-amz-checksum-sha256`:

> This header can be used as a data integrity check to verify that the data received is the same data that was originally sent. This header specifies the base64-encoded, 256-bit SHA-256 digest of the object. For more information, see Checking object integrity in the Amazon S3 User Guide.

From https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html

tmacro: That's a different header.
The API docs doesn't list `x-amz-content-sha256` on the individual operations page for some reason, but it's vaguely documented over here https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html
It defines it as "Actual payload checksum value"
From observation it seems to be hex encoded.

jonathan-gramain (approved):

nicolas2bert (approved):

PR #5406, opened 2023-11-02
improvement CLDSRV-466 add timestamp for exceptions

tmacro:
Adds a timestamp to uncaught exceptions to aid in debugging.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,10 @@
 'use strict'; // eslint-disable-line strict
 
+/**
+ * Catch uncaught exceptions and add timestamp to aid debugging
+ */
+process.on('uncaughtException', err => {
+    process.stderr.write(`${new Date().toISOString()}: Uncaught exception: \n${err.stack}`);
+});
+
 require('./lib/server.js')();


index.js
```diff
@@ -1,3 +1,10 @@
 'use strict'; // eslint-disable-line strict
 
+/**
+ * Catch uncaught exceptions and add timestamp to aid debugging
+ */
+process.on('uncaughtException', err => {
+    process.stderr.write(`${new Date().toISOString()} ${err.stack}`);
```
jonathan-gramain: Since the exception backtrace is multi-line, I'd recommend for readability to add short context info around the timestamp, followed by a newline, then the exception backtrace. This way the first backtrace line is aligned with the other lines.

Here's what I'm thinking about:
```
process.stderr.write(`${new Date().toISOString()}: Uncaught exception: \n${err.stack}`); 
```

index.js
```diff
@@ -1,3 +1,10 @@
 'use strict'; // eslint-disable-line strict
 
+/**
+ * Catch uncaught exceptions and add timestamp to aid debugging
+ */
+process.on('uncaughtException', err => {
```
ghost, line 6: Do we want to also catch the rejections?

jonathan-gramain (approved):

index.js
```diff
@@ -1,3 +1,10 @@
...
+    process.stderr.write(`${new Date().toISOString()}: Uncaught exception: \n${err.stack}`);
+});
```
ghost, line 8: This will change the service behavior in case of error right? Like, we won't crash, and we won't end the current request either, is that what we want? Maybe I'm missing something here?

tmacro: That's what I thought too, but from observation he worker still exits after the message is printed.

ghost (approved):

index.js
BourgoisMickael: It's been a long time but this does catch rejections as well

index.js
BourgoisMickael: old PR, but the worker exits because he adds its own listener (https://github.com/scality/cloudserver/blob/4f7aa548866068608fc534b3eca6f266583c599c/lib/server.js#L82-L91) to shutdown after this one.

However the primary  (that uses fork and has the monitoring server for prom metrics) don't have another listener, so this is keep the primary up in undefined state if an exception is caught

PR #5420, opened 2023-11-07
Improvement/cldsrv-426-ACL-Implicit-Deny

benzekrimaha:
PR opened after closing : https://github.com/scality/cloudserver/pull/5323

Bucket policies are not correctly interpreted, this is part of the following epic to fix that: https://github.com/scality/Arsenal/pull/2181

This PR is aiming to update ACL checks for APIs with multiple permission , ticket linked to this issue here : https://scality.atlassian.net/browse/CLDSRV-426

PRs providing implicit Deny logic to CS for processing in this PR
https://github.com/scality/Arsenal/pull/2181
https://github.com/scality/Vault/pull/2135
https://github.com/scality/cloudserver/pull/5322

Tests have also been added.

**I'm not bumping a new CLDSRV version since a new version has been created in this merged PR : https://github.com/scality/cloudserver/pull/5322 , Please let me know if it needs to be done anyways.**

PR #5322, opened 2023-09-11
CLDSRV-424: api call updated with implicit deny logic

KazToozs:
In this PR we are updating ApiCall method Auth checks to handle Implicit denies from Vault it's part of the following epic : https://github.com/scality/Arsenal/pull/2181

The ticket for this issue : https://scality.atlassian.net/browse/CLDSRV-424

Related issues
PRs providing implicit Deny logic to CS for processing in this PR
https://github.com/scality/Arsenal/pull/2181
https://github.com/scality/Vault/pull/2135

PR #5421, opened 2023-11-07
Improvement/cldsrv 427 permissions check

benzekrimaha:
PR opened after closing : https://github.com/scality/cloudserver/pull/5323

Bucket policies are not correctly interpreted, this is part of the following epic to fix that: https://github.com/scality/Arsenal/pull/2181

This PR is aiming enable authorization logics from IAM policies, bucket policies and ACLs to be interpreted in aggregate in CloudServer's 'isBucketAuthorized' and 'isObjectAuthorized' functions.  , ticket linked to this issue here : https://scality.atlassian.net/browse/CLDSRV-427

PRs providing implicit Deny logic to CS for processing in this PR
https://github.com/scality/Arsenal/pull/2181
https://github.com/scality/Vault/pull/2135
https://github.com/scality/cloudserver/pull/5322
https://github.com/scality/cloudserver/pull/5420

I'm not bumping a new CLDSRV version since a new version has been created in this merged PR : https://github.com/scality/cloudserver/pull/5322 , Please let me know if it needs to be done anyways.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -287,130 +287,90 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
     return permission;
 }
 
-function isBucketAuthorized(bucket, requestTypes, canonicalID, authInfo, log, request, actionImplicitDenies) {
+function isAuthorized(method, bucket, objectMD, requestTypes, canonicalID, authInfo, log, request,
+    actionImplicitDenies, withIam) {
+    if (!withIam) {
+        // eslint-disable-next-line no-param-reassign
+        withIam = false;
+    }
     if (!Array.isArray(requestTypes)) {
         // eslint-disable-next-line no-param-reassign
         requestTypes = [requestTypes];
     }
+    // By default, all missing actions are defined as allowed from IAM, to be
+    // backward compatible
     if (!actionImplicitDenies) {
         // eslint-disable-next-line no-param-reassign
         actionImplicitDenies = {};
     }
-    // By default, all missing actions are defined as allowed from IAM, to be
-    // backward compatible
     requestTypes.forEach(requestType => {
         if (actionImplicitDenies[requestType] === undefined) {
             // eslint-disable-next-line no-param-reassign
             actionImplicitDenies[requestType] = false;
         }
-    });
-    const mainApiCall = requestTypes[0];
-    const results = {};
-    requestTypes.forEach(_requestType => {
-        // Check to see if user is authorized to perform a
-        // particular action on bucket based on ACLs.
-        // TODO: Add IAM checks
-        let requesterIsNotUser = true;
-        let arn = null;
-        if (authInfo) {
-            requesterIsNotUser = !authInfo.isRequesterAnIAMUser();
-            arn = authInfo.getArn();
-        }
-        // if the bucket owner is an account, users should not have default access
-        if ((bucket.getOwner() === canonicalID) && requesterIsNotUser) {
-            results[_requestType] = actionImplicitDenies[_requestType] === false;
-            return;
-        }
-        const aclPermission = checkBucketAcls(bucket, _requestType, canonicalID, mainApiCall);
-        const bucketPolicy = bucket.getBucketPolicy();
-        if (!bucketPolicy) {
-            results[_requestType] = actionImplicitDenies[_requestType] === false && aclPermission;
-            return;
-        }
-        const bucketPolicyPermission = checkBucketPolicy(bucketPolicy, _requestType,
-            canonicalID, arn, bucket.getOwner(), log, request);
-        if (bucketPolicyPermission === 'explicitDeny') {
-            results[_requestType] = false;
-            return;
-        }
-        // If the bucket policy returns an allow, we accept the request, as the
-        // IAM response here is either Allow or implicit deny.
-        if (bucketPolicyPermission === 'allow') {
-            results[_requestType] = true;
-            return;
-        }
-        results[_requestType] = actionImplicitDenies[_requestType] === false && aclPermission;
     });
 
-    // final result is true if all the results are true
-    return Object.keys(results).every(key => results[key] === true);
-}
-function isObjAuthorized(bucket, objectMD, requestTypes, canonicalID, authInfo, log, request, actionImplicitDenies) {
-    if (!Array.isArray(requestTypes)) {
-        // eslint-disable-next-line no-param-reassign
-        requestTypes = [requestTypes];
-    }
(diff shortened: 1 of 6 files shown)


lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -268,75 +268,120 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
-    if (!bucketPolicy) {
-        return aclPermission;
-    }
-    const bucketPolicyPermission = checkBucketPolicy(bucketPolicy, requestType,
-        canonicalID, arn, bucket.getOwner(), log, request);
-    if (bucketPolicyPermission === 'explicitDeny') {
-        return false;
-    }
-    return (aclPermission || (bucketPolicyPermission === 'allow'));
+        if (method === 'isBucketAuthorized') {
```
KazToozs: I'm of the personal opinion that if such conditions are necessary for such large parts of code, it's better practice to keep this separate into smaller, different functions.

KazToozs (commented):
Code makes sense otherwise

lib/api/apiUtils/authorization/permissionChecks.js
benzekrimaha: Actually this check was only added for the bucket specific checks/ object ones in the else , as the functions are nearly identical besides for those checks I used the method param passed by the functions 'isBucketAUthorized', 'isObjectAuthorized' and 'evaluateBucketPolicyWithIAM' to separate them and keep all of the changes related to authorizations in only one function (can ease modifications in the future as well)

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -268,75 +268,120 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
+    if (!actionImplicitDenies) {
+        // eslint-disable-next-line no-param-reassign
+        actionImplicitDenies = {};
+    }
+    requestTypes.forEach(requestType => {
+        if (actionImplicitDenies[requestType] === undefined) {
+            // eslint-disable-next-line no-param-reassign
+            actionImplicitDenies[requestType] = false;
+        }
+    });
```
welansari: No need to initialize `undefined` values of `actionImplicitDenies`.
Inside the second `forEach` you can directly test if it's false with `!!actionImplicitDenies[requestType]`.

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -268,75 +268,120 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
-function isObjAuthorized(bucket, objectMD, requestType, canonicalID, authInfo, log, request) {
-    const bucketOwner = bucket.getOwner();
-    if (!objectMD) {
-        // User is already authorized on the bucket for FULL_CONTROL or WRITE or
-        // bucket has canned ACL public-read-write
-        if (requestType === 'objectPut' || requestType === 'objectDelete') {
-            return true;
+    const results = {};
+    const mainApiCall = requestTypes[0];
+    requestTypes.forEach(_requestType => {
```
welansari: You can directly return `requestTypes.every(_requestType => {...})` instead of storing and checking all the results.
This will also avoid looping over everything when you encounter a first error.

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -268,75 +268,120 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
+        }
+        // If the bucket policy returns an allow, we accept the request, as the
+        // IAM response here is either Allow or implicit deny.
+        if (bucketPolicyPermission === 'allow') {
+            results[_requestType] = true;
+            return;
+        }
+        // eslint-disable-next-line max-len
+        results[_requestType] = withIam === true ? actionImplicitDenies[_requestType] === false :
+            actionImplicitDenies[_requestType] === false && aclPermission;
```
Kerkesni, line 367: I find it hard to understand what's going on compared to the old code, i suggest using more helper functions that separate the logic and that don't depend on the caller method

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -268,75 +268,120 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
-        return false;
-    }
-    return (aclPermission || (bucketPolicyPermission === 'allow'));
-}
+function isAuthorized(method, bucket, objectMD, requestTypes, canonicalID, authInfo, log, request,
+    actionImplicitDenies, withIam) {
+    if (!withIam) {
+        // eslint-disable-next-line no-param-reassign
+        withIam = false;
+    }
```
welansari:
```suggestion
    actionImplicitDenies, withIam = false) {
```

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -268,75 +268,120 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
+    if (!Array.isArray(requestTypes)) {
+        // eslint-disable-next-line no-param-reassign
+        requestTypes = [requestTypes];
+    }
+    // By default, all missing actions are defined as allowed from IAM, to be
+    // backward compatible
+    if (!actionImplicitDenies) {
+        // eslint-disable-next-line no-param-reassign
+        actionImplicitDenies = {};
+    }
```
welansari, line 325: Same as for `withIam`, we can have a default value equal to `{}` in the function parameter declaration

lib/api/apiUtils/authorization/permissionChecks.js
welansari: I find it hard to understand what's going on compared to the old code, i suggest using more helper functions that separate the logic and that don't depend on the caller method

lib/api/apiUtils/authorization/permissionChecks.js
benzekrimaha: This checks also when we call the function with false as a parameter for actionImplicitDenies so putting the value by default will prevent assigning the empty object when false is passed to the function

lib/api/apiUtils/authorization/permissionChecks.js
benzekrimaha: I guess I'll separate the functions (as before)  and just do fixups instead of refactoring the 3 in one function as it will be easier to read/understand

lib/api/apiUtils/authorization/permissionChecks.js
benzekrimaha: As I'm keeping 3 separate functions , this param will no longer be needed

PR #5464, opened 2023-11-30
CLDSRV-473: fix cors issues in getVeeamFile

hervedombya:
# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -18,7 +18,7 @@ function getVeeamFile(request, response, bucketMd, log) {
         return responseXMLBody(errors.NoSuchBucket, null, response, log);
     }
     if ('tagging' in request.query) {
-        return respondWithData(request, response, log, request.bucketName,
+        return respondWithData(request, response, log, bucketMd,
             buildHeadXML('<Tagging><TagSet></TagSet></Tagging>'));
     }
     return metadata.getBucket(request.bucketName, log, (err, data) => {
@@ -38,7 +38,7 @@ function getVeeamFile(request, response, bucketMd, log) {
         const builder = new xml2js.Builder({
             headless: true,
         });
-        return respondWithData(request, response, log, request.bucketName,
+        return respondWithData(request, response, log, data,
             buildHeadXML(builder.buildObject(fileToBuild.value)), modified);
     });
 }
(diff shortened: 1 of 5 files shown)


anurag4DSB (dismissed):

lib/routes/veeam/get.js
```diff
@@ -18,7 +18,7 @@ function getVeeamFile(request, response, bucketMd, log) {
         return responseXMLBody(errors.NoSuchBucket, null, response, log);
     }
     if ('tagging' in request.query) {
-        return respondWithData(request, response, log, request.bucketName,
+        return respondWithData(request, response, log, bucketMd,
```
ghost, line 21: Context for reviewers: passing the whole bucket object instead of the name only will allow to support proxies by returning the right CORS config, API format is not impacted

lib/routes/veeam/head.js
```diff
@@ -35,7 +35,7 @@ function headVeeamFile(request, response, bucketMd, log) {
             headless: true,
         });
         const dataBuffer = Buffer.from(buildHeadXML(builder.buildObject(fileToBuild)));
-        return responseContentHeaders(null, {}, getResponseHeader(request, request.bucketName,
+        return responseContentHeaders(null, {}, getResponseHeader(request, data,
```
ghost, line 38: Worth adding a test if possible to ensure that the CORS headers are returned, usually we ensure that each bugfix ticket is merged with a test to ensure that we never recreate the issue again

lib/routes/veeam/get.js
hervedombya: Context: We had an error in local env when retrieving the capacity xml file, so to fix it we sent the bucket object as a parameter in response instead of just the bucket name.

lib/routes/veeam/head.js
hervedombya: Done in 363afcd17f77653640e39dbe7698df79d8601d90

ghost (approved):
LGTM
You can add a commit as the last one to bump the project before mergning, if you need to release. Note: make sure the targeted branch is indeed what you want to fix this bug.

JBWatenbergScality (approved):
LGTM

PR #5481, opened 2023-12-08
Improvement/cldsrv 436 bp conditions

KazToozs:
### Motivation and context

Adds Bucket Policy condition handling for limiting IPs and Object Lock retention time.

### Related issues

This PR is based on and dependant of code in CLDSRV-431.

Functionality has been hand tested and is worked as expected.

### DIFF
--- a/tests/functional/aws-node-sdk/test/bucket/putBucketPolicy.js
+++ b/tests/functional/aws-node-sdk/test/bucket/putBucketPolicy.js
@@ -156,5 +156,90 @@ describe('aws-sdk test put bucket policy', () => {
             s3.putBucketPolicy(params, err =>
                 assertError(err, 'MalformedPolicy', done));
         });
+
+        it('should allow bucket policy with valid SourceIp condition', done => {
+            const params = getPolicyParams({
+                key: 'Condition', value: {
+                    IpAddress: {
+                        'aws:SourceIp': '192.168.100.0/24',
+                    },
+                },
+            });
+            s3.putBucketPolicy(params, err => assertError(err, null, done));
+        });
+
+        it('should not allow bucket policy with invalid SourceIp format', done => {
+            const params = getPolicyParams({
+                key: 'Condition', value: {
+                    IpAddress: {
+                        'aws:SourceIp': '192.168.100', // Invalid IP format
+                    },
+                },
+            });
+            s3.putBucketPolicy(params, err => assertError(err, 'MalformedPolicy', done));
+        });
+
+        it('should allow bucket policy with valid s3:object-lock-remaining-retention-days condition', done => {
+            const params = getPolicyParams({
+                key: 'Condition', value: {
+                    NumericGreaterThanEquals: {
+                        's3:object-lock-remaining-retention-days': '30',
+                    },
+                },
+            });
+            s3.putBucketPolicy(params, err => assertError(err, null, done));
+        });
+
+        // yep, this is the expected behaviour
+        it('should not reject policy with invalid s3:object-lock-remaining-retention-days value', done => {
+            const params = getPolicyParams({
+                key: 'Condition', value: {
+                    NumericGreaterThanEquals: {
+                        's3:object-lock-remaining-retention-days': '-1', // Invalid value
+                    },
+                },
+            });
+            s3.putBucketPolicy(params, err => assertError(err, null, done));
+        });
+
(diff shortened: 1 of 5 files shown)


tests/unit/api/bucketPutPolicy.js
```diff
@@ -78,7 +78,7 @@ describe('putBucketPolicy API', () => {
         });
     });
 
-    it('should return error if policy contains conditions', done => {
+    it.skip('should return error if policy contains conditions', done => {
```
anurag4DSB: why are we skipping this test?

KazToozs: It's not needed: we now handle policy conditions, so we should not send back error any more. I want confirmation to delete it, all good?

anurag4DSB: We should delete if not needed or re-purpose :)

anurag4DSB: Is there a test for checking error is not returned when policy contains conditions?

benzekrimaha (approved):
LGTM

KazToozs:
ping

KazToozs:
ping

tests/functional/aws-node-sdk/test/bucket/putBucketPolicy.js
```diff
@@ -156,5 +156,101 @@ describe('aws-sdk test put bucket policy', () => {
...
+        });
+
+        it('should return error if a condition one of the condition values is invalid', done => {
+            const params = getPolicyParams({
+                key: 'Condition', value: {
+                    IpAddress: {
+                        'aws:SourceIp': '192.168.100',
+                    },
+                    NumericGreaterThanEquals: {
+                        's3:object-lock-remaining-retention-days': '30',
```
anurag4DSB, line 249: Question:
Where is the logic to test the 30 day?
From what I understand this condition allowed client with IP '192.168.100' to access resources which has remaining-retention-days as 30.

When a user tries to access such an object, where do we check this and deny the request?

anurag4DSB: We should add a test for the logic in integration, which puts object lock retention for a period and tests whether a user is able to access or not access the resource.
I will take a note on the tests.
But I am still not sure where the logic to check that is. Can you help guide me?

KazToozs: > Where is the logic to test the 30 day?

Here, we're not testing the logic of the condition values and whether it makes sense to set such conditions, here we're simply testing the ability to set them; functionality of such conditions is for integration testing. And yes, supposing we have a valid IP here, in a real scenario, this would limit the call on an object to the given IP to objects that have 30 or more days retention left.

> But I am still not sure where the logic to check that is. Can you help guide me?

Do you mean where in the code we test this behaciour? Or do you mean the coherency of such a test case? In Integration, we would have test cases that have more realistic use cases I'm sure.

KazToozs: > When a user tries to access such an object, where do we check this and deny the request?

Integration, as this is a bucket policy authorisation check question

anurag4DSB: I meant, 

If bucket policy is set with a condition for retention days of  less than 30
where in the code do we check if the retention day is less than 30 and if not deny the request.
For example, IAM policies had the logic already to verify the IP address.

I am not talking about a test with mocha, I am talking about authorization logic of the API request. Agreed the test will go in integration.

anurag4DSB: have you tried seeing if we are denied access if the object lock condition is false and allowed if it is true
I know it works correctly for IP as IAM uses the same logic, I am not sure right now about the object lock condition part.

anurag4DSB: Reference: Description of https://scality.atlassian.net/browse/S3C-3580
> Update bucket policy to support s3:object-lock-remaining-retention-days condition

]The above is done in this PR


> Update putBucketObjectLockConfig to check for the bucket policy and prevent setting retention period outside of the allowed limit of bucket policy

This we need to do in a PR

PR #5500, opened 2023-12-19
CLDSRV-477: change position of ACL check

KazToozs:
# Pull request template

### Motivation and context

ACL checks for PutObject and DeleteObject come before bucket policy checks. This makes actions that should go into bucket policy checks to get authorised to fail. The solution proposed here is to simply move the check after Bucket policy checks. 

### Related issues

Found during hand testing of [CLDSRV-436](https://scality.atlassian.net/jira/software/c/projects/OS/boards/214?assignee=61360619c425a20068e9dfcb&selectedIssue=CLDSRV-436&sprints=3162)

[CLDSRV-436]: https://scality.atlassian.net/browse/CLDSRV-436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -383,16 +383,16 @@ function isObjAuthorized(bucket, objectMD, requestTypesInput, canonicalID, authI
             ? _requestType.slice(0, -7) : _requestType;
         const bucketOwner = bucket.getOwner();
         if (!objectMD) {
-        // User is already authorized on the bucket for FULL_CONTROL or WRITE or
-        // bucket has canned ACL public-read-write
-            if (parsedMethodName === 'objectPut' || parsedMethodName === 'objectDelete') {
-                results[_requestType] = actionImplicitDenies[_requestType] === false;
-                return results[_requestType];
-            }
             // check bucket has read access
             // 'bucketGet' covers listObjects and listMultipartUploads, bucket read actions
             results[_requestType] = isBucketAuthorized(bucket, 'bucketGet', canonicalID, authInfo, log, request,
                 actionImplicitDenies);
+            // User is already authorized on the bucket for FULL_CONTROL or WRITE or
+            // bucket has canned ACL public-read-write
+            if ((parsedMethodName === 'objectPut' || parsedMethodName === 'objectDelete')
+                && results[_requestType] === false) {
+                results[_requestType] = actionImplicitDenies[_requestType] === false;
+            }
             return results[_requestType];
         }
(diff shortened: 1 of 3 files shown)


KazToozs:
ping

nicolas2bert (approved):

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -383,16 +383,16 @@ function isObjAuthorized(bucket, objectMD, requestTypesInput, canonicalID, authI
...
             // check bucket has read access
             // 'bucketGet' covers listObjects and listMultipartUploads, bucket read actions
             results[_requestType] = isBucketAuthorized(bucket, 'bucketGet', canonicalID, authInfo, log, request,
                 actionImplicitDenies);
+            // User is already authorized on the bucket for FULL_CONTROL or WRITE or
+            // bucket has canned ACL public-read-write
+            if ((parsedMethodName === 'objectPut' || parsedMethodName === 'objectDelete')
+                && results[_requestType] === false) {
+                results[_requestType] = actionImplicitDenies[_requestType] === false;
+            }
```
jonathan-gramain, line 395: As an optional suggestion, it might make the logic easier to follow written this way I think, so we understand that a positive "bucket authorized" result trumps any object ACL result:

```suggestion
            if (results[_requestType] !== false) {
                // or `return true`?
                return results[_requestType];
            }
            // User is already authorized on the bucket for FULL_CONTROL or WRITE or
            // bucket has canned ACL public-read-write
            if (parsedMethodName === 'objectPut' || parsedMethodName === 'objectDelete') {
                results[_requestType] = (actionImplicitDenies[_requestType] === false);
            }
            return results[_requestType];
```

Also I don't see where saving the results in the `results` object is needed, if not used it could simplify the code, but haven't looked very thoroughly and it's not related to the PR.

jonathan-gramain (approved):

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Thanks for the suggestion, I'll keep it as is as it makes the code less lengthy. The results object is needed in the different checks throughout the code.

KazToozs:
ping

benzekrimaha (changes requested):
Putting a blocker comment as CI needs to be fixed before PR can be merged

ghost (approved):

benzekrimaha (approved):

PR #5501, opened 2023-12-19
Improvement/cldsrv 436 bp conds

KazToozs:
### Motivation and context

Adds Bucket Policy condition handling for limiting IPs and Object Lock retention time.

### Related issues

This PR is based on and dependant of code in CLDSRV-431.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -480,7 +480,13 @@ function validatePolicyResource(bucketName, policy) {
 function checkIp(value) {
     const errString = 'Invalid IP address in Conditions';
 
-    // these preliminary checks are
+    // these preliminary checks are validating the provided
+    // ip address against ipaddr.js, the library we use when
+    // evaluating IP condition keys. It ensures compatibility,
+    // but additional checks are required to enforce the right
+    // notation (e.g., xxx.xxx.xxx.xxx/xx for IPv4). Otherwise,
+    // we would accept different ip formats, which is not
+    // standard in an AWS use case.
     try {
         try {
             parseCIDR(value);
@@ -488,28 +494,26 @@ function checkIp(value) {
             isValid(value);
         }
     } catch (err) {
-        if (err) { return errString; }
+        return errString;
     }
 
     // credit to Theodore John.S
     // Medium article: Validating IPv4 and IPv6 Addresses with Ease
     // — Unveiling the Power of Validation in JavaScript
-    function validateRegex(ip) {
-        const ipv4Regex = /^(\d{1,3}\.){3}\d{1,3}$/;
-        const ipv6Regex = /^([\da-f]{1,4}:){7}[\da-f]{1,4}$/i;
-        if (ipv4Regex.test(ip)) {
+    const validateIpRegex = (ip) => {
+        if (constants.ipv4Regex.test(ip)) {
             return ip.split('.').every(part => parseInt(part, 10) <= 255);
         }
-        if (ipv6Regex.test(ip)) {
+        if (constants.ipv6Regex.test(ip)) {
             return ip.split(':').every(part => part.length <= 4);
         }
         return false;
-    }
+    };
 
-    if (validateRegex(value) === true) {
+    if (validateIpRegex(value) === true) {
         return null;
     }
-    return 'Invalid IP address in Conditions';
+    return errString;
 }
 
 // This function checks all bucket policy conditions if the values provided
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -1,5 +1,8 @@
-const { evaluators, actionMaps, RequestContext } = require('arsenal').policies;
+const { evaluators, actionMaps, RequestContext, requestUtils } = require('arsenal').policies;
+const { errors } = require('arsenal');
+const { parseCIDR, isValid } = require('ipaddr.js');
 const constants = require('../../../../constants');
+const { config } = require('../../../Config');
 
 const {
     allAuthedUsersId, bucketOwnerActions, logId, publicId,
@@ -230,6 +233,20 @@ function _checkBucketPolicyResources(request, resource, log) {
     return evaluators.isResourceApplicable(requestContext, resource, log);
 }
 
+function _checkBucketPolicyConditions(request, conditions, log) {
+    const ip = request ? requestUtils.getClientIp(request, config) : undefined;
+    if (!conditions) {
+        return true;
+    }
+    // build request context from the request!
+    const requestContext = new RequestContext(request.headers, request.query,
+        request.bucketName, request.objectKey, ip,
+        request.connection.encrypted, request.resourceType, 's3', null, null,
+        null, null, null, null, null, null, null, null, null,
+        request.objectLockRetentionDays);
+    return evaluators.meetConditions(requestContext, conditions, log);
+}
+
 function _getAccountId(arn) {
     // account or user arn is of format 'arn:aws:iam::<12-digit-acct-id>:etc...
     return arn.substr(13, 12);
@@ -287,12 +304,13 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
         const principalMatch = _checkPrincipals(canonicalID, arn, s.Principal);
         const actionMatch = _checkBucketPolicyActions(requestType, s.Action, log);
         const resourceMatch = _checkBucketPolicyResources(request, s.Resource, log);
+        const conditionsMatch = _checkBucketPolicyConditions(request, s.Condition, log);
 
-        if (principalMatch && actionMatch && resourceMatch && s.Effect === 'Deny') {
+        if (principalMatch && actionMatch && resourceMatch && conditionsMatch && s.Effect === 'Deny') {
             // explicit deny trumps any allows, so return immediately
             return 'explicitDeny';
         }
-        if (principalMatch && actionMatch && resourceMatch && s.Effect === 'Allow') {
+        if (principalMatch && actionMatch && resourceMatch && conditionsMatch && s.Effect === 'Allow') {
             permission = 'allow';
         }
         copiedStatement = copiedStatement.splice(1);
@@ -459,6 +477,85 @@ function validatePolicyResource(bucketName, policy) {
     });
 }
 
+function checkIp(value) {
+    const errString = 'Invalid IP address in Conditions';
+
+    // these preliminary checks are
+    try {
+        try {
+            parseCIDR(value);
+        } catch (err) {
+            isValid(value);
+        }
+    } catch (err) {
+        if (err) { return errString; }
+    }
+
+    // credit to Theodore John.S
+    // Medium article: Validating IPv4 and IPv6 Addresses with Ease
+    // — Unveiling the Power of Validation in JavaScript
+    function validateRegex(ip) {
+        const ipv4Regex = /^(\d{1,3}\.){3}\d{1,3}$/;
+        const ipv6Regex = /^([\da-f]{1,4}:){7}[\da-f]{1,4}$/i;
+        if (ipv4Regex.test(ip)) {
+            return ip.split('.').every(part => parseInt(part, 10) <= 255);
+        }
+        if (ipv6Regex.test(ip)) {
+            return ip.split(':').every(part => part.length <= 4);
+        }
+        return false;
+    }
+
+    if (validateRegex(value) === true) {
+        return null;
+    }
+    return 'Invalid IP address in Conditions';
+}
+
+// This function checks all bucket policy conditions if the values provided
+// are valid for the condition type. If not it returns a relevant Malformed policy error string
+function validatePolicyConditions(policy) {
+    const validConditions = [
+        { conditionKey: 'aws:SourceIp', conditionValueTypeChecker: checkIp },
+        { conditionKey: 's3:object-lock-remaining-retention-days' },
+    ];
+    // keys where value type does not seem to be checked by AWS:
+    // - s3:object-lock-remaining-retention-days
+
+    if (!policy.Statement || !Array.isArray(policy.Statement) || policy.Statement.length === 0) {
+        return null;
+    }
+
+    // there can be multiple statements in the policy, each with a Condition enclosure
+    for (let i = 0; i < policy.Statement.length; i++) {
+        const s = policy.Statement[i];
+        if (s.Condition) {
+            const conditionOperators = Object.keys(s.Condition);
+            // there can be multiple condition operations in the Condition enclosure
+            for (const conditionOperator of conditionOperators) {
+                const conditionKey = Object.keys(s.Condition[conditionOperator])[0];
+                const conditionValue = s.Condition[conditionOperator][conditionKey];
+                const validCondition = validConditions.find(validCondition =>
+                    validCondition.conditionKey === conditionKey
+                );
+                // AWS returns does not return an error if the condition starts with 'aws:'
+                // so we reproduce this behaviour
+                if (!validCondition && !conditionKey.startsWith('aws:')) {
+                    return errors.MalformedPolicy.customizeDescription('Policy has an invalid condition key');
+                }
+                if (validCondition && validCondition.conditionValueTypeChecker) {
+                    const conditionValueTypeError = validCondition.conditionValueTypeChecker(conditionValue);
+                    if (conditionValueTypeError) {
+                        return errors.MalformedPolicy.customizeDescription(conditionValueTypeError);
+                    }
+                }
+            }
+        }
+    }
+    return null;
+}
+
+
 /** isLifecycleSession - check if it is the Lifecycle assumed role session arn.
  * @param {string} arn - Amazon resource name - example:
  * arn:aws:sts::257038443293:assumed-role/rolename/backbeat-lifecycle
@@ -488,6 +585,7 @@ module.exports = {
     checkBucketAcls,
     checkObjectAcls,
     validatePolicyResource,
+    validatePolicyConditions,
     isLifecycleSession,
     evaluateBucketPolicyWithIAM,
 };
--- a/lib/api/bucketPutPolicy.js
+++ b/lib/api/bucketPutPolicy.js
@@ -1,10 +1,9 @@
 const async = require('async');
 const { errors, models } = require('arsenal');
-
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const metadata = require('../metadata/wrapper');
 const { standardMetadataValidateBucket } = require('../metadata/metadataUtils');
-const { validatePolicyResource } =
+const { validatePolicyResource, validatePolicyConditions } =
     require('./apiUtils/authorization/permissionChecks');
 const { BucketPolicy } = models;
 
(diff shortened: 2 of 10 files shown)


lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,55 @@ function validatePolicyResource(bucketName, policy) {
...
+    // - s3:object-lock-remaining-retention-days
+
+    // there can be multiple statements in the policy, each with a Condition enclosure
+    for (let i = 0; i < policy.Statement.length; i++) {
+        const s = policy.Statement[i];
+        if (s.Condition) {
+            const conditionOperators = Object.keys(s.Condition);
+            // there can be multiple condition operations in the Condition enclosure
+            for (let j = 0; j < conditionOperators.length; j++) {
+                const conditionOperator = conditionOperators[j];
```
jonathan-gramain:
```suggestion
            for (const conditionOperator of conditionOperators) {
```

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,55 @@ function validatePolicyResource(bucketName, policy) {
...
+                const conditionKey = Object.keys(s.Condition[conditionOperator])[0];
+                const conditionValue = s.Condition[conditionOperator][conditionKey];
+                const validCondition = validConditions.find(validCondition =>
+                    validCondition.conditionKey === conditionKey
+                );
+                // this is the seen behaviour on AWS... don't ask me why
```
jonathan-gramain: The comment could be more instructive explaining what is the AWS behavior that we reproduce here, if it is not obvious, e.g.

```suggestion
                // AWS returns an error if the condition key doesn't start with `aws:`, so we reproduce this behavior here
```

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,55 @@ function validatePolicyResource(bucketName, policy) {
...
+                if (!validCondition && !conditionKey.startsWith('aws:')) {
+                    return 'Policy has an invalid condition key';
+                }
+                let conditionValueTypeError;
+                if (validCondition && validCondition.conditionValueTypeChecker) {
+                    conditionValueTypeError = validCondition.conditionValueTypeChecker(conditionValue);
+                }
+                if (conditionValueTypeError) {
+                    return conditionValueTypeError;
+                }
```
jonathan-gramain:
```suggestion
                if (validCondition && validCondition.conditionValueTypeChecker) {
                    const conditionValueTypeError = validCondition.conditionValueTypeChecker(conditionValue);
                    if (conditionValueTypeError) {
                        return conditionValueTypeError;
                    }
                }
```

lib/api/bucketPutPolicy.js
```diff
@@ -70,6 +69,16 @@ function bucketPutPolicy(authInfo, request, log, callback) {
                 return next(null, bucketPolicy);
             });
         },
+        (bucketPolicy, next) => {
```
jonathan-gramain: Do we need an extra step in the waterflow? I think we could just add the `validatePolicyConditions` block right after the `validatePolicyResource` one above.

KazToozs: More like [this](https://github.com/scality/cloudserver/pull/5501/files#diff-f9ed4547634b44b1019792bc2c7bdd3fd5c512d9e0f6ed3ba3c67b62b8745efaR69)?

jonathan-gramain (approved):

benzekrimaha (approved):
LGTM

benzekrimaha (changes requested):
blocking this PR until CI passing on 8.x

package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#7.10.51",
+    "arsenal": "git+https://github.com/scality/arsenal#62736abba4833f6aee67a66dfd320fe6317e798b",
```
ghost: Adding another blocker for this PR as this Arsenal version is not a valid tag, let's merge it first

ghost (changes requested):

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -230,6 +231,20 @@ function _checkBucketPolicyResources(request, resource, log) {
     return evaluators.isResourceApplicable(requestContext, resource, log);
 }
 
+function _checkBucketPolicyConditions(request, conditions, log) {
+    const ip = request ? requestUtils.getClientIp(request, config) : undefined;
```
ghost, line 237: Not for this PR, but maybe worth testing the IP bindings with a fully deployed product (Integration, Zenko), and ensure we are properly passing and accessing the IP from behind a reverse proxy (nginx).

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,54 @@ function validatePolicyResource(bucketName, policy) {
     });
 }
 
+function checkIp(value) {
+    const isValid = /^(\d{1,3}\.){3}\d{1,3}(\/\d{1,2})?$/.test(value);
+    if (isValid) {
+        return null;
+    }
+    return 'Invalid IP address in Conditions';
+}
```
ghost: Performance-wise this may not be the best way to check if the IP is valid. When I worked on the product performance, IP checks indeed had performance impacts, but this is something we do already in the ` IpCheck.ts`  file from Arsenal. 

You will have access to the `ipaddr.IPv4.isValid` function, as well as the `ipaddr.IPv6.isValid` one, as we should support both ipv6 and ipv4. ALso if would make more sense to return a boolean for such a function (see https://github.com/scality/Arsenal/blob/development/7.70/lib/ipCheck.ts#L36)

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,54 @@ function validatePolicyResource(bucketName, policy) {
...
+
+// This function checks all bucket policy conditions if the values provided
+// are valid for the condition type. If not it returns a relevant Malformed policy error string
+function validatePolicyConditions(policy) {
```
ghost, line 521: We will need some unit tests for this function to make sure we are not crashing under specific cases

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,54 @@ function validatePolicyResource(bucketName, policy) {
...
+    const validConditions = [
+        { conditionKey: 'aws:SourceIp', conditionValueTypeChecker: checkIp },
```
ghost, line 523: From what I can see here, the checker is the above function, that only checks ipv4 addresses without much logic (e.g., the CIDR can be 99 and be accepted, while it's an invalid value). So I am not sure this checker function will be enough, wile relying on the helpers we have in Arsenal will make it robust.

Noite: in Arsenal, in `conditions.ts`, we already support conditions checks for IP addresses, with NotIpAddress and IpAddress cases.-

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,54 @@ function validatePolicyResource(bucketName, policy) {
...
+                if (validCondition && validCondition.conditionValueTypeChecker) {
+                    const conditionValueTypeError = validCondition.conditionValueTypeChecker(conditionValue);
+                    if (conditionValueTypeError) {
+                        return conditionValueTypeError;
+                    }
+                }
+            }
+        }
+    }
+    return null;
```
ghost, line 559: Same here, as we return a string and end up returning the `InvalidPolicyDocument` error, we can just return booleans in this function.

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Example from AWS here: https://docs.aws.amazon.com/marketplace/latest/buyerguide/buyer-security-ipv6-upgrade.html

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +466,54 @@ function validatePolicyResource(bucketName, policy) {
...
+            const conditionOperators = Object.keys(s.Condition);
+            // there can be multiple condition operations in the Condition enclosure
+            for (const conditionOperator of conditionOperators) {
+                const conditionKey = Object.keys(s.Condition[conditionOperator])[0];
+                const conditionValue = s.Condition[conditionOperator][conditionKey];
+                const validCondition = validConditions.find(validCondition =>
+                    validCondition.conditionKey === conditionKey
+                );
+                // AWS returns does not return an error if the condition starts with 'aws:'
+                // so we reproduce this behaviour
```
ghost, line 546: While standard, is it something we want to support, as this will have no effect?
I am thinking about very long condition keys; as the BP is stored in the bucket MD, we want to avoid using all the available space for th BP (yes, this is to protect against some attacks, not the nominal use case :slightly_smiling_face: )

ghost (changes requested):

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Yes, I believe I cover this in the [integration PR](https://github.com/scality/Integration/pull/1218) you have seen. I'll resolve this conversation, feel free to reopen.

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: `MalformedPolicy` error is returned upon an error in the policy, with relevant information to what the malformation is, hence I chose to use the abstractions we saw in the comments above to return the valid error string per error. Do you suggest doing this in another way?

package.json
KazToozs: updated 418144caa4a30dae305e46c66802550efd29dc34

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Using the arsenal function here: ad0d3adaa8432feb3d73676cae53fabd088bd423

maintaining the structure to return error message per AWs standard; see [this](https://github.com/scality/cloudserver/pull/5501#discussion_r1444759471) comment and let me know what you think

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Same as above: 
Using the arsenal function here: ad0d3adaa8432feb3d73676cae53fabd088bd423

maintaining the structure to return error message per AWs standard; see [this](https://github.com/scality/cloudserver/pull/5501#discussion_r1444759471) comment and let me know what you think

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: There is a conversation to be had here indeed. I have simply been basing myself on AWS behaviour. I'll address it with the team somehow.

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: I'll wait for another review to confirm on what the function should be returning before addressing this comment

KazToozs:
ping

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Ticket created
https://scality.slack.com/archives/C01M77VU8RJ/p1704819662752699?thread_ts=1704792330.505129&cid=C01M77VU8RJ

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Confirmed we are sending back errors directly

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Added new verifications for IP using ipaddr from arsenal

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Addressed: we are sending back the errors directly

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: Updated to use checks from arsenal

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +468,85 @@ function validatePolicyResource(bucketName, policy) {
     });
 }
 
+function checkIp(value) {
+    const errString = 'Invalid IP address in Conditions';
+
+    // these preliminary checks are
```
ghost: I think the sentence is missing some words :slightly_smiling_face: 

```suggestion
    // These preliminary checks are validating the provided
    // ip address against ipaddr.js, the library we use when
    // evaluating IP condition keys. It ensures compatibility,
    // but additional checks are required to enforce the right
    // notation (e.g., xxx.xxx.xxx.xxx/xx for IPv4). Otherwise,
    // we would accept different ip formats, which is not
    // standard in an AWS use case.
```

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +468,85 @@ function validatePolicyResource(bucketName, policy) {
...
+    try {
+        try {
+            parseCIDR(value);
+        } catch (err) {
+            isValid(value);
+        }
+    } catch (err) {
+        if (err) { return errString; }
```
ghost:
```suggestion
        if (err) {
            return errString;
        }
```

Also I'm not sure the `if` is needed as we always execute this code when we had an error; unless we want to filter out some specific errors, just returning should work

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +468,85 @@ function validatePolicyResource(bucketName, policy) {
...
+    }
+
+    // credit to Theodore John.S
+    // Medium article: Validating IPv4 and IPv6 Addresses with Ease
+    // — Unveiling the Power of Validation in JavaScript
+    function validateRegex(ip) {
+        const ipv4Regex = /^(\d{1,3}\.){3}\d{1,3}(\/(3[0-2]|[12]?\d))?$/;
+        const ipv6Regex = /^([\da-f]{1,4}:){7}[\da-f]{1,4}$/i;
```
ghost: Maybe worth a constant in the const file, to avoid recreating it?

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +468,85 @@ function validatePolicyResource(bucketName, policy) {
...
+            isValid(value);
+        }
+    } catch (err) {
+        if (err) { return errString; }
+    }
+
+    // credit to Theodore John.S
+    // Medium article: Validating IPv4 and IPv6 Addresses with Ease
+    // — Unveiling the Power of Validation in JavaScript
+    function validateRegex(ip) {
```
ghost: Only personal preference but nested functions can be misleading, you can do that:

```suggestion
    const validateRegex = (ip) => {
```

Or create another function outside of `checkIp`

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -450,6 +468,85 @@ function validatePolicyResource(bucketName, policy) {
...
+        if (ipv6Regex.test(ip)) {
+            return ip.split(':').every(part => part.length <= 4);
+        }
+        return false;
+    }
+
+    if (validateRegex(value) === true) {
+        return null;
+    }
+    return 'Invalid IP address in Conditions';
```
ghost:
```suggestion
    return errString;
```

tests/functional/aws-node-sdk/test/bucket/putBucketPolicy.js
```diff
@@ -156,5 +156,101 @@ describe('aws-sdk test put bucket policy', () => {
...
+                    NumericGreaterThanEquals: {
+                        's3:object-lock-remaining-retention-days': '30',
+                    },
+                },
+            });
+            s3.putBucketPolicy(params, err => assertError(err, null, done));
+        });
+
+        // yep, this is the expected behaviour
+        it('should not reject policy with invalid s3:object-lock-remaining-retention-days value', done => {
```
ghost, line 194: AWS mentions a maximum of 100 years as the max retention period, can we test that as well?
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-managing.html#object-lock-managing-retention-limits

KazToozs: I think that's a test for putBucketRetention and putObjectRetention, and not directly related to conditions so is not in the scope of this PR.
To support this point, I was able to add the following bucket policy to an AWS bucket, 'accepting' objects with more than 50000 days.
```
{
            "Sid": "SetRetentionPeriod",
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::554628645944:user/WillTest"
            },
            "Action": "s3:PutObjectRetention",
            "Resource": "arn:aws:s3:::test-bucket-retention-condition2/*",
            "Condition": {
                "NumericGreaterThan": {
                    "s3:object-lock-remaining-retention-days": "50000"
                }
            }
        }
```

Of note, on testing the following command with retention until `2223`:
```
aws s3api put-object-retention --bucket test-bucket-retention-condition2 --key image1.png --retention Mode=GOVERNANCE,RetainUntilDate=2223-12-23T23:59:59 --profile playground-test-user --bypass-governance-retention
```

It passes, where the AWS interface plainly gives me the following:

> Retain until date
> December 24, 2223, 00:59:59 (UTC+01:00)

It seems this 100 year limit only applies to the default bucket retention configuration, as I indeed cannot set that to more than 100 years. 
I will create a ticket for that, as well as more tests generally for the PutObjectLockConfiguration API as from I see there are not many currently.

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: done 98d51fd670f3b01e0876692a64765bd53c8542b6

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: done 98d51fd670f3b01e0876692a64765bd53c8542b6

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: done 92f719e34fdb39708803e0f792aa4f896757fd96

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: done 98d51fd670f3b01e0876692a64765bd53c8542b6

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: done 98d51fd670f3b01e0876692a64765bd53c8542b6

lib/api/apiUtils/authorization/permissionChecks.js
KazToozs: choosing to maintain the nested function as it's only relevant to the scope of checkIp

tests/functional/aws-node-sdk/test/bucket/putBucketPolicy.js
ghost: The doc was making me think we were rejected if trying to put a limit higher than 100 years, but if it works, then all good

ghost (approved):

benzekrimaha (approved):

KazToozs:
ping

KazToozs:
ping

PR #5510, opened 2024-01-02
CLDSRV-482: fix head implicit index with routing rule

BourgoisMickael:
HEAD is not implemented like GET, it checks routing rules on the website after the index suffix is appended to the key.

This produces a wrong redirect if there is a rule on `index.html` path when a request is sent to the FQDN or on a directory.

Merge website GET and HEAD functions in a single function with a few conditions for HEAD as this is mostly duplicates


BourgoisMickael:
ping

BourgoisMickael:
ping

jonathan-gramain (commented):
The HEAD and GET code looks to have a lot of duplicate code, as I said in another PR. This bug seems to be a direct consequence of this, i.e. the code is very similar but bugs exist in one and not the other, where they should be the same.

For this reason I think if there is no big time crush it is worth to take some time to put the common code in a separate function and call it from both GET and HEAD implementations, and test that function separately, keeping the specific website GET and HEAD tests where they test specific behavior of either GET or HEAD, and for high-level functional testing.

BourgoisMickael:
@jonathan-gramain should I make another issue for that ?

BourgoisMickael:
ping

jonathan-gramain:
@BourgoisMickael I think it can be done in this PR.

lib/api/website.js
```diff
@@ -76,15 +79,32 @@ function _errorActions(err, errorDocument, routingRules,
...
+function capitalize(str) {
+    if (!str || typeof str !== 'string') return str;
+    return str.charAt(0).toUpperCase() + str.slice(1);
+}
+
+function callbackGetToHead(callback) {
+    return (err, userErrorPageFailure, dataGetInfo,
+        resMetaHeaders, redirectInfo, key) =>
+        callback(err, resMetaHeaders, redirectInfo, key);
+}
```
BourgoisMickael, line 100: Should I rather update the callback in Arsenal (https://github.com/scality/Arsenal/blob/development/7.10/lib/s3routes/routes/routeWebsite.ts#L57) to match HEAD and GET ?

jonathan-gramain: I don't know, do as you see fit :wink: I realized that the code was duplicated and mentioned it can be worth merging it, but I'll leave the implementation details to your discretion.

BourgoisMickael: Ok I'll leave it like that, I don't know if this arsenal part could be used somewhere else

nicolas2bert (approved):

lib/api/website.js
```diff
@@ -76,15 +79,39 @@ function _errorActions(err, errorDocument, routingRules,
     return callback(err, false, null, corsHeaders);
 }
 
+function capitalize(str) {
+    if (!str || typeof str !== 'string') return str;
+    return str.charAt(0).toUpperCase() + str.slice(1);
+}
```
ghost:
```suggestion
function capitalize(str) {
    if (!str || typeof str !== 'string') {
        return str;
    }
    return str.charAt(0).toUpperCase() + str.slice(1);
}
```

ghost (approved):

benzekrimaha (approved):

PR #5517, opened 2024-01-05
CLDSRV-488: fix website 404 with bucket policy

BourgoisMickael:
If object doesn’t exist but bucket policy allows access to the object, the response should be a 404 and not a 403.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -281,6 +281,12 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
     if (bucketOwner === canonicalID && actionImplicitDenies[requestType] === false) {
         permission = 'allow';
     }
+    // for website, requestType must be bucketGet to check bucket acl
+    // but objectGet to check bucket policy
+    if (request && request.websiteOverrideRequestTypeForBucketPolicy) {
+        // eslint-disable-next-line no-param-reassign
+        requestType = 'objectGet';
+    }
     let copiedStatement = JSON.parse(JSON.stringify(policy.Statement));
     while (copiedStatement.length > 0) {
         const s = copiedStatement[0];
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -353,7 +353,7 @@ function isBucketAuthorized(bucket, requestTypesInput, canonicalID, authInfo, lo
         if (isWebsite && _requestType === 'bucketGet') {
             // eslint-disable-next-line no-param-reassign
             _requestType = 'objectGet';
-            actionImplicitDenies[_requestType] = actionImplicitDenies[_requestType] || false;
+            actionImplicitDenies['objectGet'] = actionImplicitDenies['objectGet'] || false;
         }
         return processBucketPolicy(_requestType, bucket, canonicalID, arn, bucket.getOwner(), log,
             request, aclPermission, results, actionImplicitDenies);
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -353,7 +353,7 @@ function isBucketAuthorized(bucket, requestTypesInput, canonicalID, authInfo, lo
         if (isWebsite && _requestType === 'bucketGet') {
             // eslint-disable-next-line no-param-reassign
             _requestType = 'objectGet';
-            actionImplicitDenies['objectGet'] = actionImplicitDenies['objectGet'] || false;
+            actionImplicitDenies.objectGet = actionImplicitDenies.objectGet || false;
         }
         return processBucketPolicy(_requestType, bucket, canonicalID, arn, bucket.getOwner(), log,
             request, aclPermission, results, actionImplicitDenies);
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -281,12 +281,6 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
     if (bucketOwner === canonicalID && actionImplicitDenies[requestType] === false) {
         permission = 'allow';
     }
-    // for website, requestType must be bucketGet to check bucket acl
-    // but objectGet to check bucket policy
-    if (request && request.websiteOverrideRequestTypeForBucketPolicy) {
-        // eslint-disable-next-line no-param-reassign
-        requestType = 'objectGet';
-    }
     let copiedStatement = JSON.parse(JSON.stringify(policy.Statement));
     while (copiedStatement.length > 0) {
         const s = copiedStatement[0];
@@ -328,7 +322,7 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
 }
 
 function isBucketAuthorized(bucket, requestTypesInput, canonicalID, authInfo, log, request,
-    actionImplicitDeniesInput = {}) {
+    actionImplicitDeniesInput = {}, isWebsite = false) {
     const requestTypes = Array.isArray(requestTypesInput) ? requestTypesInput : [requestTypesInput];
     const actionImplicitDenies = !actionImplicitDeniesInput ? {} : actionImplicitDeniesInput;
     const mainApiCall = requestTypes[0];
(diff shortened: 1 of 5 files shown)


KillianG (approved):

jonathan-gramain (commented):
Maybe it technically works but I'm not fond of this approach, I see potentially two issues:
- changing `requestType` may cause unwanted side-effects down the line, during or after `checkBucketPolicy` is done. At least if it is a workaround, we should revert it to its original value when we don't need the change anymore, and add a clear comment why we change it in the first place
- I think it deserves additional comments explaining more in depth the reasoning, it's not clear to me and I don't know where to look to find the explanation of why the `objectGet` request type must be set for bucket website policy checks.

Also if there's a cleaner way to deal with this issue that doesn't involve a special flag in the request object nor to change the request type, I'd prefer that.

lib/api/websiteGet.js
```diff
@@ -138,6 +138,10 @@ function websiteGet(request, log, callback) {
                     log.trace('error retrieving object metadata',
                     { error: err });
                     let returnErr = err;
+                    // requestType must be bucketGet to check bucket acl
+                    // but objectGet to check bucket policy
+                    // eslint-disable-next-line no-param-reassign
+                    request.websiteOverrideRequestTypeForBucketPolicy = true;
```
ghost: As this is API-specific logic, would it work better if we handle the return code directly in the API, in case of an access denied?
Based on the authz results from IAM,overriding the request type may cause corner cases where some actions are allowed or not by the ACLs and bucket policy. I think, if we need to check both bucket ACLs and ObjectGet permission, that we should call this function twice and behave like AWS.

In websiteGet, we already check the bucket ACLs and policy with the `bucketGet` request type in case we have an error getting the object MD:
```js
// if index object does not exist and bucket is private AWS
                    // returns 403 - AccessDenied error.
                    if (err.is.NoSuchKey && !bucketAuthorized) {
                        returnErr = errors.AccessDenied;
                    }
                    return _errorActions(returnErr,
                      websiteConfig.getErrorDocument(), routingRules,
                      bucket, reqObjectKey, corsHeaders, request, log,
                      callback);
```
Here, we should be able to either evaluate the `objectGet` permission to return 404. We have the standardMetadataValidateBucketAndObj function that can be used, and from which you can override the request type, without altering the request object, eg:

```js
    const mdValParams = {
        authInfo,
        bucketName,
        objectKey,
        versionId,
        getDeleteMarker: true,
        requestType: 'objectHead',
        request,
    };

    return standardMetadataValidateBucketAndObj(mdValParams, request.actionImplicitDenies, log, [...]
```

BourgoisMickael: I can't run it twice, all these functions seems to encapsulate both `checkBucketAcls` and `processBucketPolicy` but I need to run them separately for `bucketGet` and `objectGet`.

`processBucketPolicy` with `bucketGet` is invalid because website only accepts `s3:GetObject` so if there is `s3:ListBucket` in a policy this will result in an invalid access.

`checkBucketAcls` returns always `true` for `objectGet` (to delegate to object acl, but I'm in the case where there is no object to check later so I should not do this check)

BourgoisMickael:
> Maybe it technically works but I'm not fond of this approach, I see potentially two issues:
> 
> * changing `requestType` may cause unwanted side-effects down the line, during or after `checkBucketPolicy` is done. At least if it is a workaround, we should revert it to its original value when we don't need the change anymore, and add a clear comment why we change it in the first place
> * I think it deserves additional comments explaining more in depth the reasoning, it's not clear to me and I don't know where to look to find the explanation of why the `objectGet` request type must be set for bucket website policy checks.
> 
> Also if there's a cleaner way to deal with this issue that doesn't involve a special flag in the request object nor to change the request type, I'd prefer that.

@jonathan-gramain 

The website doesn't work exactly like a `GetObject`.
`GetObject` needs `s3:ListBucket` action on bucket to determine if an error is 404 or 403.

But website only uses `s3:GetObject` in bucket policy: https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteAccessPermissionsReqd.html#bucket-policy-static-site

They don't mention `s3:ListBucket` and after doing some tests, it doesn't work at all.

So an `s3:ListBucket` in bucket policy is ignored by S3 and only `s3:GetObject` can be used to determine if an error is 403 or 404.

Now if I use this policy:

```json
{
   "Version": "2012-10-17",
   "Statement": [
      {
         "Sid": "PublicListObject",
         "Effect": "Allow",
         "Principal": "*",
         "Action": [
            "s3:ListBucket",
            "s3:PutObjectTagging"
         ],
         "Resource": [
            "arn:aws:s3:::mickael-bourgois-err",
            "arn:aws:s3:::mickael-bourgois-err/*"
         ]
      }
   ]
}
```
Cloudserver in case of error, with the `bucketGet` will apply the `s3:ListBucket` (that is a bucket action and not an object action) to any object because of the resource with `/*`. (I think it should not do that btw) and return 404s instead of 403.

The expected behavior on S3 would be:
```
curl --head http://mickael-bourgois-err.s3-website.eu-north-1.amazonaws.com/invalid
HTTP/1.1 403 Forbidden
x-amz-error-code: AccessDenied
x-amz-error-message: Access Denied
x-amz-request-id: 90MB6HKNG7VAKEWK
x-amz-id-2: 7FYMC9Iv8Idc0RWPEWtMXFpVV2aDETNzTzRzXNdSy2L/NmayXsBvQ+d0DNFfH4cZ1HGeElYekBI=
Date: Tue, 09 Jan 2024 13:41:24 GMT
Server: AmazonS3
```
___

We can't replace `bucketGet` with `objectGet` because `checkBucketAcls` will return `true`.

That's why we need
- `checkBucketAcls` to use `bucketGet`
- `checkBucketPolicy` to use `objectGet`
___

I moved the override to `isBucketAuthorized` and replaced the flag with an new optional argument

lib/api/websiteGet.js
ghost: The idea is to have an additional permission check, only if the object is not found but we are supposed to have access to it (given that it exists), in this case, we should check the policies against `s3:ObjectGet` instead of `s3:bucketGet`.
If the `isBucketAuthorized` or `isObjAuthorized` are checking the ACLs but you want to skip them, shouldn't we just update these functions with a new optional parameter so that we skip the ACL checks, to avoid changing the generic `checkBucketPolicy` function?

BourgoisMickael: Yes but I didn't wanted to add another parameter as theses functions already have a lot of parameters.
But I can do that so I don't put a flag on the request and I can override the request type earlier

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -346,6 +346,15 @@ function isBucketAuthorized(bucket, requestTypesInput, canonicalID, authInfo, lo
...
         }
         const aclPermission = checkBucketAcls(bucket, _requestType, canonicalID, mainApiCall);
+        // In case of error bucket access is checked with bucketGet
+        // For website, bucket policy only uses objectGet and ignores bucketGet
+        // https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteAccessPermissionsReqd.html
+        // bucketGet should be used to check acl but switched to objectGet for bucket policy
+        if (isWebsite && _requestType === 'bucketGet') {
+            // eslint-disable-next-line no-param-reassign
+            _requestType = 'objectGet';
+            actionImplicitDenies[_requestType] = actionImplicitDenies[_requestType] || false;
```
jonathan-gramain: minor clarity suggestion

```suggestion
            actionImplicitDenies['objectGet'] = actionImplicitDenies['objectGet'] || false;
```

jonathan-gramain (approved):
I think it is better this way, and clearer with the comment, thanks

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -322,7 +322,7 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
 }
 
 function isBucketAuthorized(bucket, requestTypesInput, canonicalID, authInfo, log, request,
-    actionImplicitDeniesInput = {}) {
+    actionImplicitDeniesInput = {}, isWebsite = false) {
```
ghost, line 325: Echoing your comment around the number of args here, another way could be to have a last parameter as an object, where we can set various properties depending on the needs, it would allow us to reuse if we need more specific logic, but not a blocker for this PR

ghost (approved):

PR #5521, opened 2024-01-10
CLDSRV-492: fix monitoring for website head for 7.70

BourgoisMickael:
Some monitoring is present in 7.70 but not 7.10, so I didn't see it in integration branches in https://github.com/scality/cloudserver/pull/5510 when merging GET and HEAD files.

I matched version of websiteHead before the merge: https://github.com/scality/cloudserver/blob/b52f2356baf8709fceecff3dff2e72757375513a/lib/api/websiteHead.js

### DIFF
--- a/lib/api/website.js
+++ b/lib/api/website.js
@@ -34,8 +34,10 @@ function _errorActions(err, errorDocument, routingRules,
         objectKey, err.code);
     if (errRoutingRule) {
         // route will redirect
-        monitoring.promMetrics(
-            'GET', bucketName, err.code, 'getObject');
+        if (request.method === 'GET') {
+            monitoring.promMetrics(
+                'GET', bucketName, err.code, 'getObject');
+        }
         return callback(err, false, null, corsHeaders, errRoutingRule,
             objectKey);
     }
@@ -136,6 +138,7 @@ function website(request, log, callback) {
         callback = callbackGetToHead(callback);
     }
     const methodCapitalized = capitalize(request.method);
+    const action = `${request.method.toLowerCase()}Object`;
     log.debug('processing request', { method: `website${methodCapitalized}` });
     const bucketName = request.bucketName;
     const reqObjectKey = request.objectKey ? request.objectKey : '';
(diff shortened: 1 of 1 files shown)


lib/api/website.js
```diff
@@ -34,8 +34,10 @@ function _errorActions(err, errorDocument, routingRules,
         objectKey, err.code);
     if (errRoutingRule) {
         // route will redirect
-        monitoring.promMetrics(
-            'GET', bucketName, err.code, 'getObject');
+        if (request.method === 'GET') {
```
BourgoisMickael: some monitoring is present in GET but not in HEAD for errors. I'm not sure if it's expected to be like that or if I should put the same monitoring in HEAD.

BourgoisMickael:
ping

lib/api/website.js
```diff
@@ -197,8 +200,10 @@ function website(request, log, callback) {
                 if (err) {
                     log.trace('error retrieving object metadata',
                     { error: err });
-                    monitoring.promMetrics(
-                        'GET', bucketName, err.code, 'getObject');
+                    if (request.method === 'GET') {
+                        monitoring.promMetrics(
```
BourgoisMickael: this seems to duplicate monitoring for GET as there is also monitoring called in `_errorActions`

lib/api/website.js
jonathan-gramain: I think the same monitoring should be present for HEAD and GET, it's likely an overlook

lib/api/website.js
```diff
@@ -136,6 +138,7 @@ function website(request, log, callback) {
         callback = callbackGetToHead(callback);
     }
     const methodCapitalized = capitalize(request.method);
+    const action = `${request.method.toLowerCase()}Object`;
```
jonathan-gramain: I suggest using a regular `if (request.method === 'GET') {...}` etc., it's more readable and maintainable.

lib/api/website.js
tmacro: Looks duplicated to me as well. Probably should be removed

nicolas2bert: Yes, there is monitoring duplication in the event of an error from metadata.getObjectMD. Could you please address this issue? Apart from that, everything looks good to me.

nicolas2bert (approved):

jonathan-gramain (approved):

PR #5524, opened 2024-01-18
Bugfix/cldsrv 494/fix generate v4 headers for put with body requests

fredmnl:
When attempting to send lifecycle configuration requests to AWS with this in-house client, I found that it did not include the necessary `content-md5` header. Unfortunately this header is also used when calling `backbeat`'s internal routes but the hash is expressed in hex form, not in base64.

Thus, I chose to go down the simplest path which is to add the header anyway but choose the encoding based on the route of the request. I am also adding the header all of the time, this does not have a negative impact on our tests in cloudserver (except for the one testing the behavior when the header is not present).

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#7.10.55",
+    "arsenal": "git+https://github.com/scality/arsenal#7.10.56",
     "async": "~2.5.0",
     "aws-sdk": "2.905.0",
     "azure-storage": "^2.1.0",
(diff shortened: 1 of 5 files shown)


jonathan-gramain (approved):

package.json
```diff
@@ -67,6 +67,7 @@
     "ft_management": "cd tests/functional/report && yarn test",
     "ft_node": "cd tests/functional/raw-node && yarn test",
     "ft_node_routes": "cd tests/functional/raw-node && yarn run test-routes",
+    "ft_node_route_backbeat": "cd tests/functional/raw-node && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 test/routes/routeBackbeat.js",
```
benzekrimaha, line 70: Do we need to keep this specific test ? 
Otherwise LGTM

benzekrimaha (approved):

package.json
fredmnl: Sorry I should have mentioned the commits that are meant to be dropped, I won't be adding these to the `package.json`

PR #5525, opened 2024-01-26
CLDSRV-497 Fix BackbeatClient.putMetadata with versionID

nicolas2bert:
**Issue**: When Cloudserver BackbeatClient.putMetadata() option fields are sent to Metadata through the query string, they are converted to strings. As a result, Metadata interprets the value undefined in the versionId field as an empty string ('').

**Background**: Previously, the 'crrExistingObject' script used this bug/behavior as a workaround to generate an internal version ID to replicate null version (= objects created before versioning was enabled). However, this approach has led to inconsistencies, occasionally resulting in the creation of multiple null internal versions.

**Resolution**: To address this issue, the 'crrExistingObject' workaround will be deprecated. Instead, Backbeat will be enhanced to support the replication of null versions directly, thereby ensuring more reliable and consistent behavior in handling versioning.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -522,7 +522,6 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         }
 
         const options = {
-            versionId,
             isNull,
         };
 
@@ -533,6 +532,14 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // Therefore, to avoid this confusion, we don't pass the versioning parameter at all if its value is false.
         if (versioning) {
             options.versioning = true;
+        }
+
+        // NOTE: When options fields are sent to Metadata through the query string,
+        // they are converted to strings. As a result, Metadata interprets the value undefined
+        // in the versionId field as an empty string ('').
+        // To prevent this, the versionId field is only included in options when it is defined.
+        if (versionId !== undefined) {
+            options.versionId = versionId;
         }
 
         log.trace('putting object version', {
--- a/tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
+++ b/tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
@@ -1,0 +1,181 @@
+const assert = require('assert');
+const async = require('async');
+const { models } = require('arsenal');
+const { ObjectMD } = models;
+
+const { makeBackbeatRequest } = require('../../utils/makeRequest');
+const BucketUtility = require('../../../aws-node-sdk/lib/utility/bucket-util');
+
+const describeSkipIfAWS = process.env.AWS_ON_AIR ? describe.skip : describe;
+
+const backbeatAuthCredentials = {
+    accessKey: 'accessKey1',
+    secretKey: 'verySecretKey1',
+};
+
+const testData = 'testkey data';
+
+describeSkipIfAWS('backbeat routes for replication', () => {
+    const bucketUtil = new BucketUtility(
+        'default', { signatureVersion: 'v4' });
+    const s3 = bucketUtil.s3;
+
+    const bucketSource = 'backbeatbucket-replication-source';
+    const bucketDestination = 'backbeatbucket-replication-destination';
+    const keyName = 'key0';
+    const storageClass = 'foo';
+
+    beforeEach(done =>
+        bucketUtil.emptyIfExists(bucketSource)
+            .then(() => s3.createBucket({ Bucket: bucketSource }).promise())
+            .then(() => bucketUtil.emptyIfExists(bucketDestination))
+            .then(() => s3.createBucket({ Bucket: bucketDestination }).promise())
+            .then(() => done(), err => done(err))
+    );
+
+    afterEach(done =>
+        bucketUtil.empty(bucketSource)
+            .then(() => s3.deleteBucket({ Bucket: bucketSource }).promise())
+            .then(() => bucketUtil.empty(bucketDestination))
+            .then(() => s3.deleteBucket({ Bucket: bucketDestination }).promise())
+            .then(() => done(), err => done(err))
+    );
+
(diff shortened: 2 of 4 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -535,6 +534,14 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             options.versioning = true;
         }
 
+        // NOTE: When options fields are sent to Metadata through the query string,
+        // they are converted to strings. As a result, Metadata interprets the value undefined
+        // in the versionId field as an empty string ('').
+        // To prevent this, the versionId field is only included in options when it is defined.
+        if (versionId !== undefined) {
+            options.versionId = versionId;
+        }
```
francoisferrand, line 543: can you confirm this will have no impact when backend is mongo?

nicolas2bert: Yes, I confirm

tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
```diff
@@ -0,0 +1,175 @@
...
+                authCredentials: backbeatAuthCredentials,
+                requestBody: objMD,
+            }, next),
+            next => s3.headObject({ Bucket: bucketDestination, Key: keyName, VersionId: 'null' }, next),
+            next => s3.listObjectVersions({ Bucket: bucketDestination }, next),
+        ], (err, data) => {
+            if (err) {
+                return done(err);
+            }
+            const headObjectRes = data[5];
```
jonathan-gramain: minor: could be slightly more intuitive to pass an object to `async.series` such as `{ putObject: next => s3.putObject(...), ... }` and finally access the results via `results.headObject` and `results.listObjectVersions`

tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
```diff
@@ -0,0 +1,175 @@
...
+                    requestBody: result.getSerialized(),
+                }, next);
+            },
+            next => s3.headObject({ Bucket: bucketDestination, Key: keyName, VersionId: 'null' }, next),
+            next => s3.listObjectVersions({ Bucket: bucketDestination }, next),
+        ], (err, data) => {
+            if (err) {
+                return done(err);
+            }
+            const headObjectRes = data[6];
```
jonathan-gramain: ditto for using an object

tests/functional/raw-node/utils/makeRequest.js
```diff
@@ -190,8 +190,42 @@ function makeGcpRequest(params, callback) {
...
+ * through backbeat route
+ * @param {object} params - params for making request
+ * @param {string} params.method - request method
+ * @param {string} params.bucket - bucket name
+ * @param {string} params.objectKey - object key
+ * @param {string} params.subCommand - subcommand to backbeat
+ * @param {object} [params.headers] - headers and their string values
+ * @param {object} [params.authCredentials] - authentication credentials
+ * @param {object} params.authCredentials.accessKey - access key
+ * @param {object} params.authCredentials.secretKey - secret key
```
jonathan-gramain, line 203:
```suggestion
 * @param {string} params.authCredentials.accessKey - access key
 * @param {string} params.authCredentials.secretKey - secret key
```

jonathan-gramain (approved):

tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
nicolas2bert: Yes, done.

tests/functional/raw-node/package.json
```diff
@@ -11,7 +11,7 @@
     "test-aws": "AWS_ON_AIR=true mocha -t 40000 test/",
     "test-gcp": "mocha -t 40000 test/GCP/",
     "test-routes": "mocha -t 40000 test/routes/",
-    "test": "mocha -t 40000 test/",
+    "test": "mocha -t 40000 test/routes/routeBackbeatForReplication.js",
```
benzekrimaha: shouldn't this test be renamed as it's targetting now specific tests (unless this was just for testing purpose)

benzekrimaha (approved):

tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
```diff
@@ -0,0 +1,181 @@
...
+};
+
+const testData = 'testkey data';
+
+describeSkipIfAWS('backbeat routes for replication', () => {
+    const bucketUtil = new BucketUtility(
+        'default', { signatureVersion: 'v4' });
+    const s3 = bucketUtil.s3;
+
+    const bucketSource = 'backbeatbucket-replication-source';
```
anurag4DSB, line 23: Should we use unique bucket names? Does it matter?

anurag4DSB: Same for destination

anurag4DSB (commented):
LGTM, but I want to know about Maha's comment as well, are we not planning to run the remaining tests?

tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
nicolas2bert: Unique bucket names might hide cleaning issues when done without a clear purpose. Why would we need it?

nicolas2bert:
@bert-e reset

tests/functional/raw-node/package.json
nicolas2bert: Good catch, I force-pushed it unwillingly :) https://github.com/scality/cloudserver/compare/44a1c1c4b492ec1a7c16617a3ef91da7430ec290..6a37bf33d836379203a0049ebae8e411040f59fc

nicolas2bert:
@bert-e approve

PR #5526, opened 2024-01-26
CLDSRV-498 Handling isNull master version with no versionId

nicolas2bert:
In certain cases, a master version may not have a versionId and be set as null (isNull:true). For instance, this occurs when a customer:

- Create a bucket.
- Put an object to it.
- Put bucket versioning.
- Put metadata (BackbeatClient.putMetadata), which results in the master version being set to null (isNull:true) with no versionId.

Currently, if an object is put after these steps, CloudServer fails to appropriately generate a null version. This is because CloudServer doesn't handle situations where the master version is set to isNull:true with no versionId.

The correct approach when an object is put should be to:
- Create the new version key.
- Create a new null version key, assigning it a “default non-version version id”.
- Update this “default non-version version id” to the `nullVersionId` field of the master key.

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -170,7 +170,7 @@ function processVersioningState(mst, vstat) {
         options.versioning = true;
         if (mst.exists) {
             // store master version in a new key
-            const versionId = mst.isNull ? mst.versionId : nonVersionedObjId;
+            const versionId = (mst.isNull && mst.versionId) ? mst.versionId : nonVersionedObjId;
             storeOptions.versionId = versionId;
             storeOptions.isNull = true;
             options.nullVersionId = versionId;
--- a/tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
+++ b/tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
@@ -178,4 +178,73 @@ describeSkipIfAWS('backbeat routes for replication', () => {
             return done();
         });
     });
+
+    it('should successfully put object after replicating a null version', done => {
+        let objMD;
+        let expectedVersionId;
+
+        async.series({
+            putObjectSource: next => s3.putObject(
+                { Bucket: bucketSource, Key: keyName, Body: Buffer.from(testData) }, next),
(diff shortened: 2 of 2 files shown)


lib/api/apiUtils/object/versioning.js
```diff
@@ -170,7 +170,7 @@ function processVersioningState(mst, vstat) {
         options.versioning = true;
         if (mst.exists) {
             // store master version in a new key
-            const versionId = mst.isNull ? mst.versionId : nonVersionedObjId;
+            const versionId = (mst.isNull && mst.versionId) ? mst.versionId : nonVersionedObjId;
```
jonathan-gramain, line 173: Is this meant to support the existing "wrong" master with `isNull` but no `versionId`, that we want to avoid creating from now on? I'm fine with it, just wondering if that's the purpose.

jonathan-gramain: > In certain cases, a master version may not have a versionId and be set as null (isNull:true)

Actually if I understand this, it is a "normal" situation so we need to take care of it properly.

(It's much likely a bug that we should fix by the way, we should not set the `isNull` flag in that case, but obviously we need to support it due to the existing data).

nicolas2bert: > Is this meant to support the existing "wrong" master with isNull but no versionId, that we want to avoid creating from now on? I'm fine with it, just wondering if that's the purpose.

Yes, it is to support this case created by putMetadata when updating a standalone null master.

> Actually if I understand this, it is a "normal" situation so we need to take care of it properly.

Correct.

Also, as you said, we might want to fix it. I will create a ticket for it.

jonathan-gramain (approved):

nicolas2bert:
ping

lib/api/apiUtils/object/versioning.js
nicolas2bert: [CLDSRV-509](https://scality.atlassian.net/browse/CLDSRV-509)

benzekrimaha (approved):
LGTM

nicolas2bert:
@bert-e approve

PR #5528, opened 2024-02-05
Fix gha triggers

francoisferrand:
- Add release name to release job
- Fix trigger for codeql jobs
- Fix alert job trigger to match other premerge job (8.x only)

Issue: CLDSRV-502


ghost (approved):
LGTM

> Fix alert job trigger to match other premerge job (8.x only)

Is it an expected change on the 8.x branches only? I see no integration branch yet

welansari (approved):

PR #5530, opened 2024-02-14
Bugfix/cldsrv 507 bp fixes

ghost:
This PR fixes permission checks in the context of supporting the standard IAM+Bucket Policies authorization flow.

Fixes 
https://scality.atlassian.net/browse/S3C-8459
https://scality.atlassian.net/browse/S3C-8460
https://scality.atlassian.net/browse/S3C-8461
https://scality.atlassian.net/browse/S3C-8462

- MPU APIs rely on the associated permission, but now that we check the ACLs even if IAM returned an implicit deny, we end up in the case where the object metadata is empty, hence, the logic is to check another permission that is unrelated to the MPU APIs, causing Allowed accesses even if no policy was provided.

-> Require changing the permission sent from these APIs and the logic when evaluation the BP and ACLs, to consider these APIs as separate ones.

- BucketShield must be also updated to account for the change.

- The APIs sent to BP evaluation logic is different from the ones sent to IAM, as we rely on static strings, so Versioned APIs are not properly checked.

### This was fully tested end2end using CTST in Zenko (which share the same code as S3C for authz)

All S3 APIs are checked in 20 different scenarios of authz for IAM Users, IAM Roles (assumed) and IAM Roles (assumed, cross account)

Note: in the 8.x branch, new APIs are also updated (e.g., objectRetsore.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -198,6 +198,11 @@ const constants = {
     ]),
     ipv4Regex: /^(\d{1,3}\.){3}\d{1,3}(\/(3[0-2]|[12]?\d))?$/,
     ipv6Regex: /^([\da-f]{1,4}:){7}[\da-f]{1,4}$/i,
+    actionsToConsiderAsObjectPut: [
+        'initiateMultipartUpload',
+        'objectPutPart',
+        'completeMultipartUpload',
+    ],
 };
 
 module.exports = constants;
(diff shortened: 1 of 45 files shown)


anurag4DSB (commented):
We also need bucket tagging and 7.70 specific APIs updated.
related integration PR: https://github.com/scality/Integration/pull/1242

ghost:
@anurag4DSB they are already updated (some of the APIs already had the `apiMethods` part), see 

https://github.com/scality/cloudserver/blob/development/7.70/lib/api/bucketGetTagging.js#L73
https://github.com/scality/cloudserver/blob/development/7.70/lib/api/bucketPutTagging.js#L41
https://github.com/scality/cloudserver/blob/development/7.70/lib/api/bucketDeleteTagging.js#L29


What's left is only metadataSearch, but it's not a standard S3 API so keeping the static variable is fine (same as CopyPart or CopyObject APIs because it would introduce breaking change as currently they do not check the right permission in the versioned case).

KazToozs (approved):

benzekrimaha (approved):
LGTM

PR #5531, opened 2024-02-14
CLDSRV-501 PutMetadata should write metadata on top of a null version

nicolas2bert:
NOTE: only the tests will be forwarded to `development/7.70` because the null key behavior has changed in `development/7.70`.

The `options.isNull` parameter indicates support for the new null keys logic (S3C-7352). 
However, in `development/7.10`,  the new null keys logic is not available.

This PR functionally tests: https://github.com/scality/Arsenal/pull/2215

### DIFF
--- a/tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
+++ b/tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
@@ -247,4 +247,861 @@ describeSkipIfAWS('backbeat routes for replication', () => {
             return done();
         });
     });
+
+    it('should replicate/put metadata to a destination that has a version', done => {
+        let objMD;
+        let firstVersionId;
+        let secondVersionId;
+
+        async.series({
+            enableVersioningDestination: next => s3.putBucketVersioning(
+                { Bucket: bucketDestination, VersioningConfiguration: { Status: 'Enabled' } }, next),
+            putObjectDestination: next => s3.putObject(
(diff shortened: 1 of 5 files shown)


benzekrimaha (approved):
LGTM

BourgoisMickael (approved):

tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
```diff
@@ -247,4 +247,861 @@ describeSkipIfAWS('backbeat routes for replication', () => {
...
+            assert.strictEqual(currentVersion.IsLatest, true);
+
+            assert.strictEqual(nonCurrentVersion.VersionId, 'null');
+            assert.strictEqual(nonCurrentVersion.IsLatest, false);
+
+            return done();
+        });
+    });
+
+    it('should replicate/put metadata to a destination that has a lifecycled null version', done => {
```
jonathan-gramain: What "lifecycled" refers to in this context? I think I understand that we're creating a nonversioned object on the destination, then a null version on top (which should erase it), but not sure about the link with lifecycle.

jonathan-gramain (approved):

tests/functional/raw-node/test/routes/routeBackbeatForReplication.js
nicolas2bert: This test makes sure we are able to PUT metadata to null version that had its metadata previously updated. 

I rephrased it.

nicolas2bert:
ping

nicolas2bert:
@bert-e approve

PR #5534, opened 2024-02-16
Bugfix/cldsrv 505 ip handling fix

KazToozs:
# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -480,40 +480,44 @@ function validatePolicyResource(bucketName, policy) {
 function checkIp(value) {
     const errString = 'Invalid IP address in Conditions';
 
-    // these preliminary checks are validating the provided
-    // ip address against ipaddr.js, the library we use when
-    // evaluating IP condition keys. It ensures compatibility,
-    // but additional checks are required to enforce the right
-    // notation (e.g., xxx.xxx.xxx.xxx/xx for IPv4). Otherwise,
-    // we would accept different ip formats, which is not
-    // standard in an AWS use case.
-    try {
+    const values = Array.isArray(value) ? value : [value];
+
+    for (let i = 0; i < values.length; i++) {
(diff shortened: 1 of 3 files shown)


tests/unit/auth/permissionChecks.js
```diff
@@ -382,17 +382,50 @@ describe('validatePolicyConditions', () => {
...
+            description: 'Should accept arrays of IPs',
             inputPolicy: {
                 Statement: [{
                     Condition: {
-                        IpAddress: { 'aws:SourceIp': [] }, // IP should be a string, not an array
+                        IpAddress: {
+                            'aws:SourceIp': [
+                                '10.0.11.0/24',
+                                '10.0.1.0/24',
+                            ],
```
ghost, line 393: Can we have a test that checks we properly authorize an ip given a range in the policy? (and on the other hand, deny if it doesn't match the range). As this condition specifically might be used a lot in production, I would feel more confident). It can be done in the Integration test suite.

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -480,40 +480,44 @@ function validatePolicyResource(bucketName, policy) {
...
+            if (constants.ipv6Regex.test(ip)) {
+                return ip.split(':').every(part => part.length <= 4);
+            }
+            return false;
+        };
 
-    if (validateIpRegex(value) === true) {
-        return null;
+        if (validateIpRegex(values[i]) !== true) {
+            return errString; // Return immediately if an invalid IP is found
```
ghost: I would say this specific comment is not mandatory as it says what the code does

ghost (approved):

tests/unit/auth/permissionChecks.js
KazToozs: Yes yes can add a test to Integration

benzekrimaha (approved):
LGTM

PR #5539, opened 2024-02-20
Bugfix/cldsrv 508 fix bucket tagging

ghost:
The Put and Delete Bucket Tagging APIs were missing the `request` field when evaluating the bucket policies.
The Delete API had a a previously-existing bug on 8.* branches when the bucket policy checks fail, causing a service crash.
Both, with also GetBucketPolicy, had 0 unit test, so adding some to improve the coverage.

Blocker for the next Artesca 1.7 and Ring 9.3 releases


KazToozs (approved):
LGTM

benzekrimaha (approved):

PR #5543, opened 2024-02-26
Use scaledMsPerDay when restore-adjust

KillianG:
Use scaledMsPerday when restoring an object that has already been restored to be able to make the time go faster for testing purpose

Issue: CLDSRV-512


KazToozs (approved):

francoisferrand (approved):

rachedbenmustapha (approved):

ghost (approved):

ghost (commented):
As per our discussion, maybe you also want to bump the project version before merging if we need that to test at the Zenko level?

PR #5547, opened 2024-03-06
Bugfix/cldsrv 514 handling of metadata storage errors

ghost:
A first set of fixes to reduce the occurence of orphans creation, when the fix is "easy", that is, we can delete the orphan in the same API.

Note: The code to set delete markers is safe, as only metadata is updated. However, when deleting the data (usually, after the metadata), it becomes possible to create orphans in the storage, in this case, we only log it, for rnow, before a more consistent approach.

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -197,6 +197,8 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
     /* eslint-disable camelcase */
     const dontSkipBackend = externalBackends;
     /* eslint-enable camelcase */
+    let dataGetInfoArr;
+    let needsToCleanStorage = false;
 
     const requestLogger =
         logger.newRequestLoggerFromSerializedUids(log.getSerializedUids());
@@ -250,7 +252,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             const prefixedDataStoreETag = dataStoreETag
                       ? `1:${dataStoreETag}`
                       : `1:${calculatedHash}`;
-            const dataGetInfoArr = [{ key, size, start: 0, dataStoreName,
+            dataGetInfoArr = [{ key, size, start: 0, dataStoreName,
                 dataStoreType, dataStoreETag: prefixedDataStoreETag,
                 dataStoreVersionId }];
             if (cipherBundle) {
@@ -294,11 +296,30 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             if (options.extraMD) {
                 Object.assign(metadataStoreParams, options.extraMD);
             }
-            return _storeInMDandDeleteData(bucketName, infoArr,
+            return _storeInMDandDeleteData(bucketName, dataGetInfoArr,
                 cipherBundle, metadataStoreParams,
-                options.dataToDelete, requestLogger, requestMethod, next);
+                options.dataToDelete, requestLogger, requestMethod, (err, data) => {
+                    if (err) {
+                        needsToCleanStorage = true;
+                    }
+                    return next(err, data);
+                });
         },
-    ], callback);
+    ], (err, result) => {
+        if (needsToCleanStorage && dataGetInfoArr) {
+            return data.batchDelete(dataGetInfoArr, requestMethod, null,
+                requestLogger, _err => {
+                    if (_err) {
+                        log.warn('potential orphan in storage', {
+                            error: _err,
+                            objects: dataGetInfoArr,
+                        });
+                    }
+                    return callback(err, result);
+                });
+        }
+        return callback(err, result);
+    });
 }
 
 module.exports = createAndStoreObject;
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -364,7 +364,7 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                         objMD, authInfo, canonicalID, null, request,
                         deleteInfo.newDeleteMarker, null, overheadField, log,
                         's3:ObjectRemoved:DeleteMarkerCreated', (err, result) =>
-                            callback(err, objMD, deleteInfo, result.versionId));
+                            callback(err, objMD, deleteInfo, result?.versionId));
                 },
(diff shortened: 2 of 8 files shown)


lib/services.js
```diff
@@ -363,12 +363,24 @@ const services = {
...
+                        return data.delete(objectMD.location, deleteLog, err => {
+                            if (err) {
+                                log.warn('potential orphan in storage', {
+                                    object: objectMD.location,
+                                    error: err,
+                                });
+                                return cb(err);
+                            }
+                            return cb(null, res);
+                        });
```
ghost, line 375: Note for reviewers: this code was not looking for any error when deleting data from storage. In case of error, that would create both orphans and invisible error codes for the user. Let me know if that should not be changed due to some non-documented reason(s).

lib/api/multiObjectDelete.js
```diff
@@ -364,7 +364,7 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                         objMD, authInfo, canonicalID, null, request,
                         deleteInfo.newDeleteMarker, null, overheadField, log,
                         's3:ObjectRemoved:DeleteMarkerCreated', (err, result) =>
-                            callback(err, objMD, deleteInfo, result.versionId));
+                            callback(err, objMD, deleteInfo, result?.versionId));
```
ghost, line 367: Note for reviewers: this code is not strictly related to this PR, but when the function returns an error, `result` is null so we crash here.

KazToozs (approved):
Supposing the await for the 'more consistent' approach is why we don't see tests for the cases where we can't delete? 
Do you know what this approach will be? 
LGTM otherwise

benzekrimaha (approved):
LGTM

ghost:
> Supposing the await for the 'more consistent' approach is why we don't see tests for the cases where we can't delete?
Do you know what this approach will be?

If I understand well your question @KazToozs , you are referring to the remaining oprhans we crate, or the cases where we only log it. My suggested approach, for Zenko, is to rely on transactions to perform atomic operations on the database. This way, we can easily avoid partial metadata updates that lead to either oprhans on the storage side, or in the metadata DB.
This however requires more design work beforehand.

Another solution, to complete it, because we can still have orphans with atomic updates (because we delete data from 2 different storage backends), it to persist the list of known keys that are (maybe) orphans, and have an internal job (or manual operation) taking care of them, if needed. This also requires some design.

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -197,6 +197,8 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
     /* eslint-disable camelcase */
     const dontSkipBackend = externalBackends;
     /* eslint-enable camelcase */
+    let dataGetInfoArr;
```
francoisferrand, line 200: if you make this variable global, you need to remove it from the variables which trickle through the waterfall... (e.g. infoArr).... and it should problably be renamed, since it is not really "dataGet" anymore

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -294,11 +296,30 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             if (options.extraMD) {
                 Object.assign(metadataStoreParams, options.extraMD);
             }
-            return _storeInMDandDeleteData(bucketName, infoArr,
+            return _storeInMDandDeleteData(bucketName, dataGetInfoArr,
                 cipherBundle, metadataStoreParams,
-                options.dataToDelete, requestLogger, requestMethod, next);
+                options.dataToDelete, requestLogger, requestMethod, (err, data) => {
+                    if (err) {
+                        needsToCleanStorage = true;
```
francoisferrand, line 303: instead of introducing 2 variables, it seems we simply need to batchDelete here : since this is the only case where cleanup needs to happen...

lib/api/objectPutCopyPart.js
```diff
@@ -331,6 +334,7 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
                     if (err) {
                         log.debug('error storing new metadata',
                         { error: err, method: 'storeNewPartMetadata' });
+                        needsToCleanStorage = true;
```
francoisferrand, line 337: this is the only case, might as well cleanup here: avoiding the global variables and keeping the existing waterfall logic...

lib/api/objectPutPart.js
```diff
@@ -333,6 +336,7 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                             error: err,
                             method: 'objectPutPart::metadata.putObjectMD',
                         });
+                        needsToCleanStorage = true;
```
francoisferrand, line 339: same, best to cleanup here

jonathan-gramain:
I haven't looked at the details of this PR but would like to mention that for S3C, it is a deliberate choice not to cleanup orphans. Indeed it's possible to have a dangling metadata entry because we are not sure if the metadata write actually failed for real, when we get an error. The dangling entry can cause serious issues to applications or suspicion of data loss because we cannot always know what is the history of this entry and if it has had an error.

Maybe a middle ground to tackle this issue better could be to defer the orphan cleanup after some time long enough to let the Metadata layer settle all its pending requests or timeout, then re-check what is the metadata state before doing the orphan deletion.

jonathan-gramain:
Also, when we have a good solution in mind, we should definitely consider applying it on 7.x branches (but we could do a later backport after more testing if we are concerned about the risk of regression on S3C).

ghost:
@jonathan-gramain , do you mean we can have errors returned by _metadata_ in the S3C case, and this approach for MongoDB is not safe for the 7.x branches? Or do you mean, even with MongoDB, we should not "trust" the errors returned by the driver, as it might report an error, while the metadata was actually written?
Note that here, we tackle orphans **in the storage, not in metadata.**

Having something running after a while seem unsafe: we can, between this timelapse, have other operations on this object's metadata and have it changed/deleted in a way that would not solve our issue here. E.g.: we really fail to write the metadata at first, but the data A is written in the storage. Then the client retries and succeed, metadata is written and data B is stored. Then the cleanup job detects that the metadata is here, and does nothing. At the end, we have an orphan.

Anyway, putting this work on hold as we will need a unified solution for both branches (IMHO, in our APIs, as we should rely on their return codes perfectly).

PR #5548, opened 2024-03-12
CLDSRV-518 Duplication of version ID in metadata

nicolas2bert:


nicolas2bert:
ping

lib/routes/routeBackbeat.js
```diff
@@ -497,6 +497,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     // If no new versioned objects are added for given object(s), they look like
                     // standalone master keys.
                     versioning = false;
+                    // Delete the version id from the version metadata payload to prevent issues
+                    // with creating a non-version object (versioning set to false) that includes a version id.
+                    // For example, this version ID might come from a null version of a suspended bucket being
+                    // replicated to this bucket.
+                    delete omVal.versionId;
```
jonathan-gramain: It would look clearer to me if we do this after `omVal.isNull = true;` above, if that doesn't change the core logic of the fix. But I am not sure I understand fully the logic.

nicolas2bert: It is indeed cleaner to have this check after `omVal.isNull = true;` and indeed does not change the core logic since putMetadata to **insert** a new metadata key is only used for replication by S3C (Artesca uses Cloud provider SDKs) that only supports **versioned bucket**.

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e force_reset

jonathan-gramain (approved):

BourgoisMickael (approved):

nicolas2bert:
@bert-e approve

PR #5554, opened 2024-04-03
Feature/cldsrv 515 quotas

ghost:
Adding quota evaluation logic in S3:
- Both accounts (from authz results) and buckets
- Use a Scubaclient instance to get the latest metrics with inflight metrics (i.e., ongoing bytes not yet acknowledged) 
- Periodic healthchecks to dynamically handle problems with the quota service (today, we disable quota checks)
- Add configuration for `scubaclient` (host & port)
- Support WRITE and DELETE APIs with inflights. Copy APIs are handled slightly differently as we need to get the original object size.
- Unit tests
- Functional tests (no account coverage, to be done during integration testing)

> Note: won't pass the CI because we cannot install scubaclient without setup in the dockerfile, or having scubaclient public.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -365,6 +365,50 @@ jobs:
         source: /tmp/artifacts
       if: always()
 
+  quota-tests:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        inflights:
+          - name: "With Inflights"
+            value: "true"
+          - name: "Without Inflights"
+            value: "false"
+    env:
+      S3METADATA: mongodb
+      S3BACKEND: mem
+      S3QUOTA: scuba
+      QUOTA_ENABLE_INFLIGHTS: ${{ matrix.inflights.value }}
+      SCUBA_HOST: localhost
+      SCUBA_PORT: 8100
+      SCUBA_HEALTHCHECK_FREQUENCY: 100
+      CLOUDSERVER_IMAGE: ghcr.io/${{ github.repository }}:${{ github.sha }}
+      MONGODB_IMAGE: ghcr.io/${{ github.repository }}/ci-mongodb:${{ github.sha }}
+      JOB_NAME: ${{ github.job }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Setup CI environment
+      uses: ./.github/actions/setup-ci
+    - name: Setup CI services
+      run: docker compose --profile mongo up -d
+      working-directory: .github/docker
+    - name: Run quota tests
+      run: |-
+        set -ex -o pipefail;
+        bash wait_for_local_port.bash 8000 40
+        yarn run test_quota | tee /tmp/artifacts/${{ github.job }}/tests.log
+    - name: Upload logs to artifacts
+      uses: scality/action-artifacts@v4
+      with:
+        method: upload
+        url: https://artifacts.scality.net
+        user: ${{ secrets.ARTIFACTS_USER }}
+        password: ${{ secrets.ARTIFACTS_PASSWORD }}
+        source: /tmp/artifacts
+      if: always()
+
   kmip-ft-tests:
     runs-on: ubuntu-latest
     needs: build
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get update \
 
 ENV PYTHON=python3
 COPY package.json yarn.lock /usr/src/app/
+RUN npm install typescript -g
 RUN yarn install --production --ignore-optional --frozen-lockfile --ignore-engines --network-concurrency 1
 
 ################################################################################
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1117,6 +1117,44 @@ class Config extends EventEmitter {
                 this.redis.password = config.redis.password;
             }
         }
+        if (config.scuba) {
+            this.scuba = {};
+            if (config.scuba.host) {
+                assert(typeof config.scuba.host === 'string',
+                    'bad config: scuba host must be a string');
+                this.scuba.host = config.scuba.host;
+            }
+            if (config.scuba.port) {
+                assert(Number.isInteger(config.scuba.port)
+                    && config.scuba.port > 0,
+                    'bad config: scuba port must be a positive integer');
+                this.scuba.port = config.scuba.port;
+            }
+        }
+        if (process.env.SCUBA_HOST && process.env.SCUBA_PORT) {
+            assert(typeof process.env.SCUBA_HOST === 'string',
+                'bad config: scuba host must be a string');
+            assert(Number.isInteger(Number(process.env.SCUBA_PORT))
+                && Number(process.env.SCUBA_PORT) > 0,
+                'bad config: scuba port must be a positive integer');
+            this.scuba = {
+                host: process.env.SCUBA_HOST,
+                port: Number(process.env.SCUBA_PORT),
+            };
+        }
+        if (this.scuba) {
+            this.quotaEnabled = true;
+        }
+        const maxStaleness = Number(process.env.QUOTA_MAX_STALENESS_MS) ||
+            config.quota?.maxStatenessMS ||
+            24 * 60 * 60 * 1000;
+        assert(Number.isInteger(maxStaleness), 'bad config: maxStalenessMS must be an integer');
+        const enableInflights = process.env.QUOTA_ENABLE_INFLIGHTS === 'true' ||
+            config.quota?.enableInflights || false;
+        this.quota = {
+            maxStaleness,
+            enableInflights,
+        };
         if (config.utapi) {
             this.utapi = { component: 's3' };
             if (config.utapi.host) {
@@ -1706,6 +1744,7 @@ class Config extends EventEmitter {
         let data = 'multiple';
         let metadata = 'file';
         let kms = 'file';
+        let quota = 'none';
         if (process.env.S3BACKEND) {
             const validBackends = ['mem', 'file', 'scality', 'cdmi'];
             assert(validBackends.indexOf(process.env.S3BACKEND) > -1,
@@ -1766,11 +1805,15 @@ class Config extends EventEmitter {
         if (process.env.S3KMS) {
             kms = process.env.S3KMS;
         }
+        if (process.env.S3QUOTA) {
+            quota = process.env.S3QUOTA;
+        }
         this.backends = {
             auth,
             data,
             metadata,
             kms,
+            quota,
         };
     }
 
@@ -1899,6 +1942,14 @@ class Config extends EventEmitter {
         this.publicInstanceId = crypto.createHash('sha256')
                                 .update(instanceId)
                                 .digest('hex');
+    }
+
+    isQuotaEnabled() {
+        return !!this.quotaEnabled;
+    }
+
+    isQuotaInflightEnabled() {
+        return this.quota.enableInflights;
     }
 }
 
--- a/lib/api/apiUtils/object/objectRestore.js
+++ b/lib/api/apiUtils/object/objectRestore.js
@@ -8,7 +8,7 @@ const { pushMetric } = require('../../../utapi/utilities');
 const { decodeVersionId } = require('./versioning');
 const collectCorsHeaders = require('../../../utilities/collectCorsHeaders');
 const { parseRestoreRequestXml } = s3middleware.objectRestore;
-
+const { processBytesToWrite, validateQuotas } = require('../quotas/quotaUtils');
 
 /**
  * Check if tier is supported
@@ -59,6 +59,14 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
         objectKey,
         versionId: decodedVidResult,
         requestType: request.apiMethods || 'restoreObject',
+        /**
+         * Restoring an object might not cause any impact on
+         * the storage, if the object is already restored: in
+         * this case, the duration is extended. We disable the
+         * quota evaluation and trigger it manually.
+         */
+        checkQuota: false,
+        request,
     };
 
     return async.waterfall([
@@ -115,6 +123,16 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
                         isObjectRestored = _isObjectRestored;
                         return next(err, bucketMD, objectMD);
                     });
+            },
+            function evaluateQuotas(bucketMD, objectMD, next) {
+                if (isObjectRestored) {
+                    return next(null, bucketMD, objectMD);
+                }
+                const actions = Array.isArray(mdValueParams.requestType) ?
+                    mdValueParams.requestType : [mdValueParams.requestType];
+                const bytes = processBytesToWrite(request.apiMethod, bucketMD, mdValueParams.versionId, 0, objectMD);
+                return validateQuotas(request, bucketMD, request.accountQuotas, actions, request.apiMethod, bytes, log,
+                    err => next(err, bucketMD, objectMD));
             },
             function updateObjectMD(bucketMD, objectMD, next) {
                 const params = objectMD.versionId ? { versionId: objectMD.versionId } : {};
--- a/lib/api/apiUtils/quotas/quotaUtils.js
+++ b/lib/api/apiUtils/quotas/quotaUtils.js
@@ -15,18 +15,28 @@ const QuotaService = require('../../../quotas/quotas');
  * @param {string} versionId - version id of the object
  * @param {number} contentLength - content length of the object
  * @param {object} objMD - object metadata
+ * @param {object} destObjMD - destination object metadata
  * @return {number} processed content length
  */
-function processBytesToWrite(apiMethod, bucket, versionId, contentLength, objMD) {
+function processBytesToWrite(apiMethod, bucket, versionId, contentLength, objMD, destObjMD = null) {
     let bytes = contentLength;
-    if (!bytes && objMD?.['content-length']) {
-        if (!bucket.isVersioningEnabled() || bucket.isVersioningEnabled() && versionId) {
+    if (apiMethod === 'objectRestore') {
+        // object is being restored
+        bytes = Number.parseInt(objMD['content-length'], 10);
+    } else if (!bytes && objMD?.['content-length']) {
+        if (apiMethod === 'objectCopy' || apiMethod === 'objectPutCopyPart') {
+            if (!destObjMD || bucket.isVersioningEnabled()) {
+                // object is being copied
+                bytes = Number.parseInt(objMD['content-length'], 10);
+            } else if (!bucket.isVersioningEnabled()) {
+                // object is being copied and replaces the target
+                bytes = Number.parseInt(objMD['content-length'], 10) -
+                    Number.parseInt(destObjMD['content-length'], 10);
+            }
+        } else if (!bucket.isVersioningEnabled() || bucket.isVersioningEnabled() && versionId) {
             // object is being deleted
             bytes = -Number.parseInt(objMD['content-length'], 10);
         }
-    } else if (apiMethod === 'objectRestore') {
-        // object is being restored
-        bytes = Number.parseInt(objMD['content-length'], 10);
     } else if (bytes && objMD?.['content-length'] && !bucket.isVersioningEnabled()) {
         // object is being replaced: store the diff, if the bucket is not versioned
         bytes = bytes - Number.parseInt(objMD['content-length'], 10);
@@ -57,26 +67,6 @@ function isMetricStale(metric, resourceType, resourceName, action, inflight, log
         return true;
     }
     return false;
-}
-
-/**
- * Finalizes a request by clearing quota updates for API methods and
- * invoking the callback function.
- *
- * @param {Object} request - The request object.
- * @param {Function} callback - The callback function to be invoked.
- * @returns {undefined} - Returns nothing.
- */
-function finalizeRequestQuotas(request, callback) {
-    if (request.clearQuotaUpdatesForAPI) {
-        return async.forEachLimit(request.apiMethods, 5, (apiMethod, next) => {
-            if (request.clearQuotaUpdatesForAPI[apiMethod]) {
-                return request.clearQuotaUpdatesForAPI[apiMethod](next);
-            }
-            return next();
-        }, callback);
-    }
-    return callback();
 }
 
 /**
@@ -187,41 +177,41 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
             if (err) {
                 return done(err);
             }
-            const removeInflights = _done => async.parallel({
-                deleteBucketInflights: parallelDone => {
-                    if (bucketQuota > 0) {
-                        return QuotaService.getUtilizationMetrics('bucket',
-                            `${bucket.getName()}_${creationDate}`, null, {
-                            action: apiName,
-                            ..._inflights,
-                        }, parallelDone);
-                    }
-                    return parallelDone();
-                },
-                deleteAccountInflights: parallelDone => {
-                    if (accountQuota > 0 && account?.account) {
-                        return QuotaService.getUtilizationMetrics('account',
-                            account.account, null, {
-                            action: apiName,
-                            ..._inflights,
-                        }, parallelDone);
-                    }
-                    return parallelDone();
-                },
-            }, _done);
+            // Inflights are inverted: in case of cleanup, we just re-issue
+            // the same API call.
             if (_inflights) {
                 _inflights.inflight = -_inflights.inflight;
             }
-            // Remove any inflight if the quotas are exceeded
-            if (!actionWithDataDeletion[apiMethod] && (bucketQuotaExceeded || accountQuotaExceeded)) {
-                return removeInflights(done);
-            }
+            const removeInflights = (errorFromAPI, _done) => {
+                // Quotas are cleaned only in case of error in the API
+                if (!errorFromAPI) {
+                    return _done();
+                }
+                return async.parallel({
+                    deleteBucketInflights: parallelDone => {
+                        if (bucketQuota > 0) {
+                            return QuotaService.getUtilizationMetrics('bucket',
+                                `${bucket.getName()}_${creationDate}`, null, {
+                                action: apiName,
+                                ..._inflights,
+                            }, parallelDone);
+                        }
+                        return parallelDone();
+                    },
+                    deleteAccountInflights: parallelDone => {
+                        if (accountQuota > 0 && account?.account) {
+                            return QuotaService.getUtilizationMetrics('account',
+                                account.account, null, {
+                                action: apiName,
+                                ..._inflights,
+                            }, parallelDone);
+                        }
+                        return parallelDone();
+                    },
+                }, _done);
+            };
             if (shouldSendInflights) {
-                // eslint-disable-next-line no-param-reassign
-                request.clearQuotaUpdatesForAPI = {
-                    ...request.clearQuotaUpdatesForAPI,
-                    [apiName]: removeInflights,
-                };
+                request.finalizerHooks.push(removeInflights);
             }
             return done();
         });
@@ -244,6 +234,5 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
 module.exports = {
     processBytesToWrite,
     isMetricStale,
-    finalizeRequestQuotas,
     validateQuotas,
 };
--- a/lib/api/apiUtils/quotas/quotaUtils.js
+++ b/lib/api/apiUtils/quotas/quotaUtils.js
@@ -1,0 +1,249 @@
+const async = require('async');
+const { errors } = require('arsenal');
+const {
+    actionNeedQuotaCheckCopy,
+    actionNeedQuotaCheck,
+    actionWithDataDeletion,
+} = require('arsenal').policies;
+const { config } = require('../../../Config');
+const QuotaService = require('../../../quotas/quotas');
+
+/**
+ * Process the bytes to write based on the request and object metadata
+ * @param {string} apiMethod - api method
+ * @param {BucketInfo} bucket - bucket info
+ * @param {string} versionId - version id of the object
+ * @param {number} contentLength - content length of the object
+ * @param {object} objMD - object metadata
+ * @return {number} processed content length
+ */
+function processBytesToWrite(apiMethod, bucket, versionId, contentLength, objMD) {
+    let bytes = contentLength;
+    if (!bytes && objMD?.['content-length']) {
+        if (!bucket.isVersioningEnabled() || bucket.isVersioningEnabled() && versionId) {
+            // object is being deleted
+            bytes = -Number.parseInt(objMD['content-length'], 10);
+        }
+    } else if (apiMethod === 'objectRestore') {
+        // object is being restored
+        bytes = Number.parseInt(objMD['content-length'], 10);
+    } else if (bytes && objMD?.['content-length'] && !bucket.isVersioningEnabled()) {
+        // object is being replaced: store the diff, if the bucket is not versioned
+        bytes = bytes - Number.parseInt(objMD['content-length'], 10);
+    }
+    return bytes || 0;
+}
+
+/**
+ * Checks if a metric is stale based on the provided parameters.
+ *
+ * @param {Object} metric - The metric object to check.
+ * @param {string} resourceType - The type of the resource.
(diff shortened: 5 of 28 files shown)


ghost:
I have a question for you @tmacro : is scubaclient meant to become public at some point? If we use it here, it might break the service for the open source community. It also breaks the dockerfile builds, unless we pass the github token in it...

tests/utilities/mock/Scuba.js
```diff
@@ -0,0 +1,95 @@
...
+        }
+        if (timeout) {
+            setTimeout(() => {
+                if (this._data.bucket[bucket]) {
+                    this._data.bucket[bucket].inflight += inflight;
+                }
+            }, timeout);
+        } else {
+            if (this._data.bucket[bucket]) {
+                this._data.bucket[bucket].inflight += inflight;
```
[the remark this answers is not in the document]
ghost: This is fixed.

tests/utilities/mock/Scuba.js
```diff
@@ -0,0 +1,95 @@
...
+        if (action === 'objectRestore') {
+            timeout = 0;
+        }
+        if (!this._data.bucket[bucket]) {
+            this._data.bucket[bucket] = { current: 0, nonCurrent: 0, inflight: 0 };
+        }
+        if (timeout) {
+            setTimeout(() => {
+                if (this._data.bucket[bucket]) {
+                    this._data.bucket[bucket].inflight += inflight;
```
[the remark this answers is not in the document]
ghost: This is fixed.

.github/workflows/tests.yaml
```diff
@@ -363,6 +363,40 @@ jobs:
...
+      S3BACKEND: mem
+      SCUBA_HOST: localhost
+      SCUBA_PORT: 8100
+      SCUBA_HEALTHCHECK_FREQUENCY: 100
+      CLOUDSERVER_IMAGE: ghcr.io/${{ github.repository }}/cloudserver:${{ github.sha }}
+      MONGODB_IMAGE: ghcr.io/${{ github.repository }}/ci-mongodb:${{ github.sha }}
+      JOB_NAME: ${{ github.job }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
```
francoisferrand:
```suggestion
      uses: actions/checkout@v4
```

ghost: https://github.com/scality/cloudserver/pull/5554/commits/0d4cc38000c41e2f53946a1b9ec225eab16d1e2a

.github/workflows/tests.yaml
```diff
@@ -363,6 +363,40 @@ jobs:
...
+    - name: Setup CI services
+      run: docker compose up -d
+      working-directory: .github/docker
+    - name: Run scuba tests
+      run: |-
+        set -ex -o pipefail;
+        bash wait_for_local_port.bash 8000 40
+        yarn run test_scuba | tee /tmp/artifacts/${{ github.job }}/tests.log
+    - name: Upload logs to artifacts
+      uses: scality/action-artifacts@v3
```
francoisferrand: @v4

.github/workflows/tests.yaml
```diff
@@ -363,6 +363,40 @@ jobs:
...
 
+  scuba-tests:
+    runs-on: ubuntu-latest
+    needs: build
+    env:
+      S3BACKEND: mem
+      SCUBA_HOST: localhost
+      SCUBA_PORT: 8100
+      SCUBA_HEALTHCHECK_FREQUENCY: 100
+      CLOUDSERVER_IMAGE: ghcr.io/${{ github.repository }}/cloudserver:${{ github.sha }}
```
francoisferrand: ghcr.io/${{ github.repository }}:${{ github.sha }}

ghost: I updated the actions.

ghost: Actually, no, I cannot make this change currently, the PR is based on another one, that is not yet up to date with ghcr migration, so the image in the CI still is the one you commented, with /cloudserver

This will be done before merging / while rebasing

Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN apt-get update \
 
 ENV PYTHON=python3
 COPY package.json yarn.lock /usr/src/app/
+RUN npm install typescript -g
```
francoisferrand, line 26: why do we need to install typescript ?
should it not be already installed by `yarn install` if needed ?

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -372,6 +381,141 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
...
+                bucketName: bucket.getName(),
+            });
+        }
+        return callback(null);
+    }
+    const creationDate = new Date(bucket.getCreationDate()).getTime();
+
+    return async.forEach(apiNames, (apiName, done) => {
+        let shouldEvaluateCopyObject = false;
+        if (apiName === 'objectGet' && (apiMethod === 'objectCopy' || apiMethod === 'objectPutCopyPart')) {
```
francoisferrand: should we not have maps similar to `actionWithDataDeletion` for detecting this case?

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -372,6 +381,141 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
...
+                if (bucketQuota > 0) {
+                    return ScubaClientInstance.getUtilizationMetrics('bucket',
+                        `${bucket.getName()}_${creationDate}`, null, {
+                        action: apiName,
+                        inflight,
+                    }, (err, bucketMetrics) => {
+                        if (err) {
+                            return parallelDone(err);
+                        }
+                        if (bucketMetrics.bytesTotal > bucketQuota) {
```
francoisferrand, line 432: nit:
```suggestion
                        if (!err && bucketMetrics.bytesTotal > bucketQuota) {
```

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -372,6 +381,141 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
...
+                    });
+                }
+                return parallelDone();
+            },
+        }, err => {
+            if (err) {
+                return done(err);
+            }
+            if (!actionWithDataDeletion[apiMethod] && (bucketQuotaExceeded || accountQuotaExceeded)) {
+                if (apiMethod?.endsWith('Delete')) {
```
francoisferrand: seems hard-coded, can we do better: something like

```suggestion
                if (actionWithDataDeletion[apiMethod]) {
```

ghost: Yes, missed this one, thx

lib/api/apiUtils/authorization/permissionChecks.js
ghost: yes, but not in arsenal, in the cloudserver's consts I think, this is not cross service.

Dockerfile
ghost: When building the sub dependencies, especially scubaclient, we have an error where tsc is not found (for scubaclient), because we need to install tsc globally first. I did not try to dig into it much, but it's pretty common to install tsc globally in dockerfiles of ts projects (or projects using it).

tests/scuba/awsNodeSdk.js
francoisferrand, line 1: should this not be a _quota_ folder instead of _scuba_ ?

tests/unit/auth/permissionChecks.js
```diff
@@ -534,3 +554,307 @@ describe('validatePolicyConditions', () => {
...
+    beforeEach(() => {
+        ScubaClientInstance.enabled = true;
+        ScubaClientInstance._getLatestmetricsCallback = sinon.stub().resolves({});
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    it('should return null if quota is <= 0 or scuba is disabled', done => {
```
francoisferrand:
```suggestion
    it('should return null if quota is <= 0', done => {
```

tests/unit/auth/permissionChecks.js
```diff
@@ -534,3 +554,307 @@ describe('validatePolicyConditions', () => {
...
+    it('should return null if scuba is disabled', done => {
+        ScubaClientInstance.enabled = false;
+        validateQuotas(mockBucket, {}, [], '', false, mockLog, err => {
+            assert.ifError(err);
+        assert.strictEqual(ScubaClientInstance._getLatestmetricsCallback.called, false);
+            done();
+        });
+    });
+
+    it('should return null if metrics retrieval fails', done => {
```
francoisferrand: side note (not for this PR in any case) : bypassing quota in that case may allow for easily bypassing the quotas with DOS attack... e.g. if we try to hammer the ser-ver with many putObject requests, scuba will probably fail before cloudserver.... and the request may bypass quota :/

.github/workflows/tests.yaml
```diff
@@ -363,6 +363,40 @@ jobs:
...
+    - name: Setup CI environment
+      uses: ./.github/actions/setup-ci
+    - name: Setup CI services
+      run: docker compose up -d
+      working-directory: .github/docker
+    - name: Run scuba tests
+      run: |-
+        set -ex -o pipefail;
+        bash wait_for_local_port.bash 8000 40
+        yarn run test_scuba | tee /tmp/artifacts/${{ github.job }}/tests.log
```
francoisferrand: even with the function tests, we actually don't test with the "real" scuba : so we may easily introduce issue if the mock does not match the actual scuba API... → we should find a way to run these tests with either/both mock scuba or "real" scuba
(i.e. get feedback earlier than when it reaches zenko...)

ghost: The functional tests are only testing cloudserver contract here, but yes, to me this is a second step (as we'll have the integration test in Zenko), in the same way we test Vault + MongoDB today, we can really test with scuba, what do you think?

lib/api/apiUtils/authorization/permissionChecks.js
francoisferrand: theorically maybe, but practically it could make sense in arsenal as well:
* arsenal already includes lots of cloudserver (or backbeat, or vault) specific code ;
* it may make sense to keep this together with the other 2 variables, since this whole algorithm is based on these variables. The question is really "_would it make sense to modify just this new variable without considering if an update is required on the other 2_" ?

tests/unit/auth/permissionChecks.js
ghost: It can be worth discussing this PR. We have two things:
- The periodic healthchecks, they are the one that determine the state of the quota checks.
- The error handling, when evaluating quotas. When there is no quota and scuba is enabled, it means count items never ran for the current bucket/account, and thus we need to ignore the quotas.

Today we ignore the quotas if the healthchecks failed, or if there is an error when getting the metrics (without discriminating between really missing metrics and other issues).

If the bucket/quota exists, we could try to still return an error in this case, but I'm not sure how.

But anyway, the scuba service must be sized properly, so that it is not a bottleneck. Mongo can be, but scuba shouldn't...

tests/scuba/awsNodeSdk.js
ghost: Yes, if we have more use cases later, we can rename it. And maybe it'll be removed before, if we run the real scuba.

lib/api/apiUtils/authorization/permissionChecks.js
ghost: After reviewing this, I'm not sure anymore it makes sense as a map: we need to pass two strings, the current API and the current action we evaluate, so better having a small function; a map won't make a lot of sense... The function will be added as a new function in arsenal.

.github/workflows/tests.yaml
ghost: Done after rebasing.

.github/workflows/tests.yaml
ghost: Created https://scality.atlassian.net/browse/CLDSRV-528, in the improvement epic. It will be about not only scuba, but also vault (we don't test it today in the CI...)

lib/metadata/metadataUtils.js
```diff
@@ -238,6 +261,15 @@ function standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log,
...
             return next(null, bucket, objMD);
         },
+        (bucket, objMD, next) => {
+            let contentLength = request?.parsedContentLength || 0;
+            if (!contentLength && !objMD) {
+                return next(null, bucket, objMD);
+            }
+            contentLength = processBytesToWrite(request, bucket, versionId, contentLength, objMD);
+            return validateQuotas(bucket, request.accountQuotas, requestType, request.apiMethod,
+                contentLength, log, err => next(err, bucket, objMD));
```
francoisferrand, line 251: Putting the quota check here will impact many APIs, introducing a very slight risk (like quota unexpectedly applied on read...), as well as negigible cost (extra test), but also possibly leading to unexpected error : if an invalid request is made (like restore on a hot object), it may thus get a quota error instead of the invalid state error...

Wondering what is the expected behavior, and/or if we should change it now or in a followup ticket.

lib/Config.js
```diff
@@ -1117,6 +1117,47 @@ class Config extends EventEmitter {
...
+                assert(Number.isInteger(config.quota.maxStalenessMS),
+                    'bad config: quota.maxStalenessMS must be an integer');
+                this.quota = { maxStalenessMS: config.quota.maxStalenessMS };
+            }
+        } else {
+            this.quota = { maxStalenessMS: 24 * 60 * 60 * 1000 };
+        }
+        if (process.env.QUOTA_MAX_STALENESS_MS) {
+            assert(Number.isInteger(Number(process.env.QUOTA_MAX_STALENESS_MS)),
+                'bad config: quota.maxStalenessMS must be an integer');
```
```suggestion
                'bad config: QUOTA_MAX_STALENESS_MS must be an integer');
```

lib/Config.js
```diff
@@ -1117,6 +1117,47 @@ class Config extends EventEmitter {
...
+        }
+        if (this.scuba) {
+            this.quotaEnabled = true;
+        }
+        if (config.quota) {
+            if (config.quota.maxStalenessMS) {
+                assert(Number.isInteger(config.quota.maxStalenessMS),
+                    'bad config: quota.maxStalenessMS must be an integer');
+                this.quota = { maxStalenessMS: config.quota.maxStalenessMS };
+            }
```
francoisferrand: in the `else` case, this.quota is not initialized...

lib/Config.js
francoisferrand: or better yet, merge this with the previous case, so we do this only once. something like:

```
const maxStaleness = process.env.QUOTA_MAX_STALENESS_MS || config.quota?.maxStatenessMS || 24 * 60 * 60 * 1000;
assert(Number.isInteger(maxStaleness), 'bad config: maxStalenessMS must be an integer');
this.quota = { maxStaleness };
```

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -372,6 +382,141 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
...
+                if (bucketQuota > 0) {
+                    return ScubaClientInstance.getUtilizationMetrics('bucket',
+                        `${bucket.getName()}_${creationDate}`, null, {
+                        action: apiName,
+                        inflight,
+                    }, (err, bucketMetrics) => {
+                        if (err) {
+                            return parallelDone(err);
+                        }
+                        if (bucketMetrics.bytesTotal > bucketQuota) {
```
francoisferrand: we should check here as well that the data returned by Scuba is not stale (as the quota request will return the date as well)

lib/metadata/metadataUtils.js
ghost: As discussed, we will, to better centralize the logic and minimize the risks, keep the quota evaluation logic as part of the authz checks. But:

- We will handle the restore object separately, as extending the restoratin duration means no quota change.
- We will revert the inflights set in case of error at the end of an API call, if needed.

ghost: Changes are done:

- Generalized the quota logic implementation, it can be another implementation, not always "scuba" (which is one of the possible solutions from S3 point of view). The backend is to be specified with `S3QUOTA`. To be also more future-proof, by default, the inflights are disabled, as this will be our nominal use case later. A configuration is thus added under `quota` to support it (ZKOP is also updated, and can be later bound to a scuba version filtering).

- To handle errors in APIs, the quota evaluation logic modifies the request object and adds some functions in a field, evaluated for each action of the API at the end, in case of error. For example, it ensures we do not change the inflights (when enabled) if we try to delete a locked object. **Note**: I found that we always pass the `objectReplicate` action in the route backbeat, so quotas are in fact **not** evaluated in there. This will be updated later based on the decision we make.

- The quota logic is better filtered and will only execute when matching the list of known APIs, or if we pass specific flags (e.g., during copy operation).

- Moved the quota logic in a single "quotaUtils" file.

- Tests are adapted.

lib/Config.js
ghost: Updated

lib/api/apiUtils/authorization/permissionChecks.js
ghost: The logic was updated with the `isMetricStale` method

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -142,6 +143,12 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
             if (isObjectRestored) {
                 monitoring.promMetrics(
                     'POST', bucketName, '200', 'restoreObject');
+                // If quotas are evaluated, increasing the quota duration should
+                // not affect the utilization metrics.
+                if (request.clearQuotaUpdatesForAPI?.[mdValueParams]) {
```
francoisferrand: * the name of this attribute should be improved, not very clear at the moment
* I don't see where this variable is set... 🧐
* since this is set in each request, I guess we should be able to directly set the appropriate callback (which would be cleaner IMO)... but then I did not see how it is initalized or added to the request

.github/workflows/tests.yaml
```diff
@@ -365,6 +365,42 @@ jobs:
         source: /tmp/artifacts
       if: always()
 
+  quota-tests:
+    runs-on: ubuntu-latest
+    needs: build
+    env:
+      S3BACKEND: mem
+      S3QUOTA: scuba
+      QUOTA_ENABLE_INFLIGHTS: "true"
```
francoisferrand: in-flights are always enabled in tests,  but disabled by default in Cloudserver: so we are introducing risk (aka debt, somehow...), which may hit us once we start using cloudserver with no inflight...

If we do not enable inflight processing by default, this case should be tested as well to avoid future problems. (or we could stick with the inflight enabled for now, even if ZKOP/... force enable the flag, to allow us to later disable it...)

lib/api/apiUtils/object/objectRestore.js
ghost: The attribute is set only after the quota evaluation logic, in https://github.com/scality/cloudserver/pull/5554/files#diff-8ed436ddf1cb95712b422e07d470067ab7d749d99faf6f35e26c2d614cf41096R220, and if inflights are enabled

If there is no error, or if inflights are disabled, or quotas are not evaluated for whatever reason (wrong API, no scuba available...), the default callback is called.

.github/workflows/tests.yaml
ghost: We have (unit) test covering when there is no inflight enabled. Not for the functional tests, because the current scuba + mock both use the inflights, I'm not sure this introduces any debt...

If we enable inflights by default, and we someday deploy a real scuba in the CI, we will still have debt when inflights are dropped. If we disable inflights by default, this is the same situation.

What I would suggest is to rename the quota test `quota-tests-with-inflights`, and later we can have another test suite without, if needed... What do you think?

ghost: (In ZKOP, the flag is enforced anyway)

lib/api/apiUtils/object/objectRestore.js
ghost: > since this is set in each request, I guess we should be able to directly set the appropriate callback

I am not sure we can, the variable is set during the quota evaluation, and each API has custom ways to call their final callbacks, depending on the HTTP verb, etc.

.github/workflows/tests.yaml
francoisferrand: > If we enable inflights by default, and we someday deploy a real scuba in the CI, we will still have debt when inflights are dropped.

In that case, we will actually be making a new change: which must be tested accordingly. In particular, having to disable in-flights when supporting a real scuba is a way to force adding these tests, as otherwise it would not work.

The debt/risk you are avoiding is not about cloudserver itself, but really about having it _silently_ enabled in cloudserver, and thus implicitely enabled... which would indeed break silently if that default value is changed later.

Since the flag is set anyway in ZKOP, we may keep the default in cloudserver to match the test: and be future-proof as Zenko/ZKOP will gracefully handle changes to this default (not very satifying as this is technically a breaking change, but risk should be managed already in this specific case...)

> If we disable inflights by default, this is the same situation.

If we disable inflights by default, but run tests only for another default: we introduce a false sense that the feature is present and tested... And when people start using it, there is a risk that no extra test is added: they will simply build on it, and if the code is good enough not notice the lack of (functional) test...

The default case is the default, so it must be most thoroughly tested. Hence either adding tests for that default case or switching the default.

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -0,0 +1,248 @@
...
+                },
+            }, _done);
+            if (_inflights) {
+                _inflights.inflight = -_inflights.inflight;
+            }
+            if (shouldSendInflights) {
+                // eslint-disable-next-line no-param-reassign
+                request.removeQuotaEvaluationInflights = {
+                    ...request.removeQuotaEvaluationInflights,
+                    [apiName]: removeInflights,
```
francoisferrand: * it seems to be it may be simpler and cleaner to just introduce a callback called in any case at the end of the request : and let this callback decide if it needs to clear the inflights or not.
* what is the point of indexing by `apiName` ? we should have only one API name for each request...

francoisferrand: can this really happen?
this is not a generic finalizer/hooking facility, but specific to quota: and no other place where we set it AFAICS...

lib/quotas/scuba/wrapper.js
```diff
@@ -0,0 +1,69 @@
...
+            this.periodicHealthCheck();
+        }
+    }
+
+    _healthCheck() {
+        return this.healthCheck().then(data => {
+            if (data?.date) {
+                const date = new Date(data.date);
+                if (Date.now() - date.getTime() > this.maxStaleness) {
+                    throw new Error('Data is stale');
```
francoisferrand: nit: `Data us stale, disabling quotas`

lib/api/objectPutCopyPart.js
```diff
@@ -45,6 +46,7 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
         versionId: reqVersionId,
         getDeleteMarker: true,
         requestType: 'objectGet',
+        enableQuota: true,
```
francoisferrand: should probably add a comment here, to explain why the quota check is forcefully enabled on `Get` and disabled on `Put`, contrary to expectations...

francoisferrand: (same for `ObjectCopy`)

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -142,6 +143,12 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
             if (isObjectRestored) {
                 monitoring.promMetrics(
                     'POST', bucketName, '200', 'restoreObject');
+                // If quotas are evaluated, increasing the quota duration should
+                // not affect the utilization metrics.
+                if (request.removeQuotaEvaluationInflights?.[mdValueParams]) {
+                    return request.removeQuotaEvaluationInflights?.[mdValueParams.requestType]?.(
+                        () => callback(null, 200, responseHeaders));
```
francoisferrand: instead of calling the cleanup callback, it would be cleaner to forcefully disable the quota check from `standardMetadataValidateBucketAndObj` (for this case only, e.g. add `disableQuota: true` at line 63), and thus explicitely call `validateQuota()` from here only when needed...

lib/metadata/metadataUtils.js
```diff
@@ -238,6 +238,16 @@ function standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log,
             }
             return next(null, bucket, objMD);
         },
+        (bucket, objMD, next) => {
+            if (params.disableQuota || (!params.enableQuota && !requestType.some(type => actionNeedQuotaCheck[type] ||
+                actionWithDataDeletion[type]))) {
```
francoisferrand: nit: API may be simpler with a single parameter (`checkQuota` ?) instead of two parameters (`disableQuota` and `enableQuota`)
* `checkQuota === undefined` : automatically depending on the action
* `checkQuota === true` : always check quota
* `checkQuota === false` : never check quota

```suggestion
            const needQuotaCheck = requestType -> requestType.some(type => actionNeedQuotaCheck[type] || actionWithDataDeletion[type]);
            const checkQuota = params.checkQuota === undefined ? needQuotaCheck(requestType) : params.checkQuota;
            if (!checkQuota) {
```

tests/unit/quotas/quotas.js
francoisferrand, line 1: empty file...

.github/workflows/tests.yaml
ghost: I will add new tests - the mock will be adapted and disable inflights. Enabling the inflights by default is not what we want (e.g., for an open source use case).

lib/api/objectPutCopyPart.js
ghost: Yes, I will add it, but we already have some info in the commit
https://github.com/scality/cloudserver/pull/5554/commits/6ac3b31006d161891713dfa19e2023dbb3479ced

ghost: Done in both files

lib/api/apiUtils/quotas/quotaUtils.js
ghost: Closing as it was removed in favor of the table

.github/workflows/tests.yaml
ghost: The mock was updated to support the no-inflight mode (S3 API calls will, in this case, directly update the mock, to simulate a system that would work without passing any inflights). At the end, the same tests are running for both scenarios.

lib/api/apiUtils/quotas/quotaUtils.js
ghost: This was changed to a table as discussed: it also has the benefit of handling in a clean way multiple actions having the same name (e.g., during multi object deletes).

lib/api/apiUtils/object/objectRestore.js
ghost: The API was updated: quota check is disabled during authz, then we evaluate it manually just before updating the MD, and if the object state matches. THis also ensures we return `InvalidObjectState` before `QuotaExceeded`. I had to rework the tests, so that we edit the object MD using a metadata client (here, mongodb) to really test a restore.

lib/metadata/metadataUtils.js
ghost: Thanks, I cannot use your suggestion directly as I need to update the logic, but this works well indeed - PR was updated

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -0,0 +1,231 @@
...
+ */
+function processBytesToWrite(apiMethod, bucket, versionId, contentLength, objMD) {
+    let bytes = contentLength;
+    if (apiMethod === 'objectRestore') {
+        // object is being restored
+        bytes = Number.parseInt(objMD['content-length'], 10);
+    } else if (!bytes && objMD?.['content-length']) {
+        if (apiMethod === 'objectCopy') {
+            // object is being copied
+            bytes = Number.parseInt(objMD['content-length'], 10);
```
francoisferrand: the target of the copy may already exist, and should be accounted for in non-versioned buckets.

and we actually have this case in the product, when updating metadata (from the UI) in case bucket is not versioned...

lib/api/objectPutCopyPart.js
```diff
@@ -45,6 +46,14 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
         versionId: reqVersionId,
         getDeleteMarker: true,
         requestType: 'objectGet',
+        /**
+         * Authorization will first check the target object, with an objectPut
+         * action. But in this context, the source object metadata is still
+         * unknown. In the context of quotas, to know the number of bytes that
+         * are being written, we explicitly enable the quota evaluation logic
+         * during the objectGet action.
```
francoisferrand: nit:
```suggestion
         * during the objectGet action instead.
```

lib/api/objectPutCopyPart.js
```diff
@@ -88,6 +98,7 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
         objectKey: destObjectKey,
         partNumber: paddedPartNumber,
         uploadId,
+        enableQuota: true,
```
francoisferrand, line 101: why do we need a second quota check in the `uploadPartCopy` ?

ghost: Upload part copy acts as an upload part, but the part is an existing object: we need to evaluate the quotas in the same way as copyObject during the `get` part, because we write a new object...
Or maybe I misunderstood your question?

lib/api/apiUtils/quotas/quotaUtils.js
ghost: You are right, I updated the logic by passing an optional parameter: `destObjMD`, known at the quota evaluation time for Copy APIs, so that we can, in the `processBytesToWrite` function, know if the copy is a replacement. This is unit tested.
Also added a filter on the copy part because it was missing

lib/api/objectPutCopyPart.js
ghost: I added a missing filter on the copy part API

lib/api/apiUtils/quotas/quotaUtils.js
ghost: It's in https://github.com/scality/cloudserver/pull/5554/commits/3ca848057cff1ddf55f357c24cb8759aad1c2f94

tests/unit/Config.js
```diff
@@ -350,6 +350,92 @@ describe('Config', () => {
...
+            assert.deepStrictEqual(
+                config.quota,
+                {
+                    maxStaleness: 1234,
+                    enableInflights: true,
+                },
+            );
+        });
+    });
+
```
benzekrimaha, line 454: could be good to have a test where staleness is not an integer to make sure the error is thrown

tests/quota/awsNodeSdk.js
```diff
@@ -252,7 +257,7 @@ function multiObjectDelete(bucket, keys, callback) {
     }, callback);
 }
 
-describe('quota evaluation with scuba metrics', function t() {
+runIfMongo('quota evaluation with scuba metrics', function t() {
```
benzekrimaha: as runIfMongo not used anywhere else : 
`(process.env.S3METADATA === 'mongodb' ? describe : describe.skip)`

tests/unit/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -500,12 +500,40 @@ describe('processBytesToWrite', () => {
...
         bucket.isVersioningEnabled.returns(true);
         objMD = { 'content-length': 100 };
-        contentLength = 150;
 
         const bytes = processBytesToWrite('objectCopy', bucket, versionId, 0, objMD);
 
         assert.strictEqual(bytes, 100);
     });
+
+    it('should return content length of the bject metadata is the action is a copy part (getObject authz)', () => {
```
benzekrimaha: `should return content length of the bject metadata **if** the action is a copy part (getObject authz)`

benzekrimaha (approved):
LGTM (very minor none blocking comments as others have been adressed and the logic tested on the plateform as well), fixups commits should be squashed as well before merging

tests/unit/Config.js
ghost: added a test that confirms that in this case, we fallback to the default value

francoisferrand (approved):

PR #5556, opened 2024-04-11
Improvement/cldsrv 520 quota monitoring

ghost:
Adding monitoring for Quotas evaluation logic

image

- Health of the scuba service as seen by cloudserver, displayed as a gauge and a % value.
  - Chose to use a counter with two labels in order to ignore when scuba is not deployed
- Number of quota success or failure, displayed as op/s
- The latency, with granularity to know the impact of the quota evaluation logic given different cases: bucket or account quota exceeded, passing, or data deletion.
- The number of bucket with enabled
- The op/s of the quota evaluation by API

An alert (warning) is added if the utilization metrics are stale or not accessible.

### DIFF
--- a/lib/api/apiUtils/quotas/quotaUtils.js
+++ b/lib/api/apiUtils/quotas/quotaUtils.js
@@ -90,6 +90,7 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
     let type;
     let bucketQuotaExceeded = false;
     let accountQuotaExceeded = false;
+    let quotaEvaluationDuration;
     const requestStartTime = process.hrtime.bigint();
     const bucketQuota = bucket.getQuota();
     const accountQuota = account?.quota || 0;
@@ -199,54 +200,70 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
             if (_inflights) {
                 _inflights.inflight = -_inflights.inflight;
             }
-            const removeInflights = (errorFromAPI, _done) => {
+            const finalizeQuotas = (errorFromAPI, _done) => {
                 // Quotas are cleaned only in case of error in the API
-                if (!errorFromAPI) {
-                    return _done();
+                const quotaCleanUpStartTime = process.hrtime.bigint();
+                const code = (bucketQuotaExceeded || accountQuotaExceeded) ? 429 : 200;
+                if (errorFromAPI) {
+                    return async.parallel({
+                        deleteBucketInflights: parallelDone => {
+                            if (bucketQuota > 0) {
+                                return QuotaService.getUtilizationMetrics('bucket',
+                                    `${bucket.getName()}_${creationDate}`, null, {
+                                    action: apiName,
+                                    ..._inflights,
+                                }, parallelDone);
+                            }
+                            return parallelDone();
+                        },
+                        deleteAccountInflights: parallelDone => {
+                            if (accountQuota > 0 && account?.account) {
+                                return QuotaService.getUtilizationMetrics('account',
+                                    account.account, null, {
+                                    action: apiName,
+                                    ..._inflights,
+                                }, parallelDone);
+                            }
+                            return parallelDone();
+                        },
+                    }, err => {
+                        if (err) {
+                            log.warn('Error deleting inflights from the quota service', {
+                                error: err.name,
+                                description: err.message,
+                            });
+                        }
+                        const quotaCleanUpDuration = Number(process.hrtime.bigint() - quotaCleanUpStartTime);
+                        monitoring.quotaEvaluationDuration.labels({
+                            action: apiMethod,
+                            type,
+                            code,
+                        }).observe((quotaEvaluationDuration + quotaCleanUpDuration) / 1e9);
+                        return _done();
+                    });
                 }
-                return async.parallel({
-                    deleteBucketInflights: parallelDone => {
-                        if (bucketQuota > 0) {
-                            return QuotaService.getUtilizationMetrics('bucket',
-                                `${bucket.getName()}_${creationDate}`, null, {
-                                action: apiName,
-                                ..._inflights,
-                            }, parallelDone);
-                        }
-                        return parallelDone();
-                    },
-                    deleteAccountInflights: parallelDone => {
-                        if (accountQuota > 0 && account?.account) {
-                            return QuotaService.getUtilizationMetrics('account',
-                                account.account, null, {
-                                action: apiName,
-                                ..._inflights,
-                            }, parallelDone);
-                        }
-                        return parallelDone();
-                    },
-                }, _done);
+                monitoring.quotaEvaluationDuration.labels({
+                    action: apiMethod,
+                    type,
+                    code,
+                }).observe(quotaEvaluationDuration / 1e9);
+                return _done();
             };
-            if (shouldSendInflights) {
-                request.finalizerHooks.push(removeInflights);
-            }
+            request.finalizerHooks.push(finalizeQuotas);
             return done();
         });
     }, err => {
-        const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
+        quotaEvaluationDuration = Number(process.hrtime.bigint() - requestStartTime);
         if (err) {
             log.warn('Error getting metrics from the quota service, allowing the request', {
                 error: err.name,
                 description: err.message,
             });
         }
-        if (actionWithDataDeletion[apiMethod]) {
-            monitorQuotaEvaluation(apiMethod, type, 200, responseTimeInNs / 1e9);
-        } else if (bucketQuotaExceeded || accountQuotaExceeded) {
-            monitorQuotaEvaluation(apiMethod, type, 429, responseTimeInNs / 1e9);
+        if (!actionWithDataDeletion[apiMethod] &&
+            (bucketQuotaExceeded || accountQuotaExceeded)) {
             return callback(errors.QuotaExceeded);
         }
-        monitorQuotaEvaluation(apiMethod, type, 200, responseTimeInNs / 1e9);
         return callback();
     });
 }
--- a/lib/api/apiUtils/quotas/quotaUtils.js
+++ b/lib/api/apiUtils/quotas/quotaUtils.js
@@ -1,5 +1,6 @@
 const async = require('async');
 const { errors } = require('arsenal');
+const monitoring = require('../../../utilities/monitoringHandler');
 const {
     actionNeedQuotaCheckCopy,
     actionNeedQuotaCheck,
@@ -64,6 +65,7 @@ function isMetricStale(metric, resourceType, resourceName, action, inflight, log
             action,
             inflight,
         });
+        monitoring.requestWithQuotaMetricsUnavailable.inc();
         return true;
     }
     return false;
@@ -85,17 +87,32 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
     if (!config.isQuotaEnabled() || !inflight) {
         return callback(null);
     }
+    let type;
     let bucketQuotaExceeded = false;
     let accountQuotaExceeded = false;
+    const requestStartTime = process.hrtime.bigint();
     const bucketQuota = bucket.getQuota();
     const accountQuota = account?.quota || 0;
     const shouldSendInflights = config.isQuotaInflightEnabled();
+
+    if (bucketQuota && accountQuota) {
+        type = 'bucket+account';
+    } else if (bucketQuota) {
+        type = 'bucket';
+    } else {
+        type = 'account';
+    }
+
+    if (actionWithDataDeletion[apiMethod]) {
+        type = 'delete';
+    }
 
     if ((bucketQuota <= 0 && accountQuota <= 0) || !QuotaService?.enabled) {
         if (bucketQuota > 0 || accountQuota > 0) {
             log.warn('quota is set for a bucket, but the quota service is disabled', {
                 bucketName: bucket.getName(),
             });
+            monitoring.requestWithQuotaMetricsUnavailable.inc();
         }
         return callback(null);
     }
@@ -216,6 +233,7 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
             return done();
         });
     }, err => {
+        const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
         if (err) {
             log.warn('Error getting metrics from the quota service, allowing the request', {
                 error: err.name,
@@ -223,10 +241,12 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
             });
         }
         if (actionWithDataDeletion[apiMethod]) {
-            return callback();
+            monitorQuotaEvaluation(apiMethod, type, 200, responseTimeInNs / 1e9);
         } else if (bucketQuotaExceeded || accountQuotaExceeded) {
+            monitorQuotaEvaluation(apiMethod, type, 429, responseTimeInNs / 1e9);
             return callback(errors.QuotaExceeded);
         }
+        monitorQuotaEvaluation(apiMethod, type, 200, responseTimeInNs / 1e9);
         return callback();
     });
 }
--- a/lib/api/apiUtils/quotas/quotaUtils.js
+++ b/lib/api/apiUtils/quotas/quotaUtils.js
@@ -72,6 +72,116 @@ function isMetricStale(metric, resourceType, resourceName, action, inflight, log
 }
 
 /**
+ * Evaluates quotas for a bucket and an account and update inflight count.
+ *
+ * @param {number} bucketQuota - The quota limit for the bucket.
+ * @param {number} accountQuota - The quota limit for the account.
+ * @param {object} bucket - The bucket object.
+ * @param {object} account - The account object.
+ * @param {number} inflight - The number of inflight requests.
+ * @param {number} inflightForCheck - The number of inflight requests for checking quotas.
+ * @param {string} action - The action being performed.
+ * @param {object} log - The logger object.
+ * @param {function} callback - The callback function to be called when evaluation is complete.
+ * @returns {object} - The result of the evaluation.
+ */
+function _evaluateQuotas(
+    bucketQuota,
+    accountQuota,
+    bucket,
+    account,
+    inflight,
+    inflightForCheck,
+    action,
+    log,
+    callback,
+) {
+    let bucketQuotaExceeded = false;
+    let accountQuotaExceeded = false;
+    const creationDate = new Date(bucket.getCreationDate()).getTime();
+    return async.parallel({
+        bucketQuota: parallelDone => {
+            if (bucketQuota > 0) {
+                return QuotaService.getUtilizationMetrics('bucket',
+                    `${bucket.getName()}_${creationDate}`, null, {
+                    action,
+                    inflight,
+                }, (err, bucketMetrics) => {
+                    if (err || inflight < 0) {
+                        return parallelDone(err);
+                    }
+                    if (!isMetricStale(bucketMetrics, 'bucket', bucket.getName(), action, inflight, log) &&
+                        bucketMetrics.bytesTotal + inflightForCheck > bucketQuota) {
+                        log.debug('Bucket quota exceeded', {
+                            bucket: bucket.getName(),
+                            action,
+                            inflight,
+                            quota: bucketQuota,
+                            bytesTotal: bucketMetrics.bytesTotal,
+                        });
+                        bucketQuotaExceeded = true;
+                    }
+                    return parallelDone();
+                });
+            }
+            return parallelDone();
+        },
+        accountQuota: parallelDone => {
+            if (accountQuota > 0 && account?.account) {
+                return QuotaService.getUtilizationMetrics('account',
+                    account.account, null, {
+                    action,
+                    inflight,
+                }, (err, accountMetrics) => {
+                    if (err || inflight < 0) {
+                        return parallelDone(err);
+                    }
+                    if (!isMetricStale(accountMetrics, 'account', account.account, action, inflight, log) &&
+                        accountMetrics.bytesTotal + inflightForCheck > accountQuota) {
+                        log.debug('Account quota exceeded', {
+                            accountId: account.account,
+                            action,
+                            inflight,
+                            quota: accountQuota,
+                            bytesTotal: accountMetrics.bytesTotal,
+                        });
+                        accountQuotaExceeded = true;
+                    }
+                    return parallelDone();
+                });
+            }
+            return parallelDone();
+        },
+    }, err => {
+        if (err) {
+            log.warn('Error evaluating quotas', {
+                error: err.name,
+                description: err.message,
+                isInflightDeletion: inflight < 0,
+            });
+        }
+        return callback(err, bucketQuotaExceeded, accountQuotaExceeded);
+    });
+}
+
+/**
+ * Monitors the duration of quota evaluation for a specific API method.
+ *
+ * @param {string} apiMethod - The name of the API method being monitored.
+ * @param {string} type - The type of quota being evaluated.
+ * @param {string} code - The code associated with the quota being evaluated.
+ * @param {number} duration - The duration of the quota evaluation in nanoseconds.
+ * @returns {undefined} - Returns nothing.
+ */
+function monitorQuotaEvaluationDuration(apiMethod, type, code, duration) {
+    monitoring.quotaEvaluationDuration.labels({
+        action: apiMethod,
+        type,
+        code,
+    }).observe(duration / 1e9);
+}
+
+/**
  *
  * @param {Request} request - request object
  * @param {BucketInfo} bucket - bucket object
@@ -117,7 +227,6 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
         }
         return callback(null);
     }
-    const creationDate = new Date(bucket.getCreationDate()).getTime();
 
     return async.forEach(apiNames, (apiName, done) => {
         // Object copy operations first check the target object,
@@ -136,122 +245,44 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
         // are not sent to the utilization service. When inflights are enabled,
         // the sum of the current utilization metrics only are compared with the
         // quota. They include the current inflight bytes sent in the request.
-        const _inflights = shouldSendInflights ? { inflight } : undefined;
-        const _inflightForCheck = shouldSendInflights ? 0 : inflight;
-        return async.parallel({
-            bucketQuota: parallelDone => {
-                if (bucketQuota > 0) {
-                    return QuotaService.getUtilizationMetrics('bucket',
-                        `${bucket.getName()}_${creationDate}`, null, {
-                        action: apiName,
-                        ..._inflights,
-                    }, (err, bucketMetrics) => {
-                        if (err) {
-                            return parallelDone(err);
-                        }
-                        if (!isMetricStale(bucketMetrics, 'bucket', bucket.getName(), apiName, inflight, log) &&
-                            bucketMetrics.bytesTotal + _inflightForCheck > bucketQuota) {
-                            log.debug('Bucket quota exceeded', {
-                                bucket: bucket.getName(),
-                                action: apiName,
-                                inflight,
-                                quota: bucketQuota,
-                                bytesTotal: bucketMetrics.bytesTotal,
-                            });
-                            bucketQuotaExceeded = true;
-                        }
-                        return parallelDone();
-                    });
+        let _inflights = shouldSendInflights ? inflight : undefined;
+        const inflightForCheck = shouldSendInflights ? 0 : inflight;
+        return _evaluateQuotas(bucketQuota, accountQuota, bucket, account, _inflights,
+            inflightForCheck, apiName, log, (err, _bucketQuotaExceeded, _accountQuotaExceeded) => {
+                if (err) {
+                    return done(err);
                 }
-                return parallelDone();
-            },
-            accountQuota: parallelDone => {
-                if (accountQuota > 0 && account?.account) {
-                    return QuotaService.getUtilizationMetrics('account',
-                        account.account, null, {
-                        action: apiName,
-                        ..._inflights,
-                    }, (err, accountMetrics) => {
-                        if (err) {
-                            return parallelDone(err);
-                        }
-                        if (!isMetricStale(accountMetrics, 'account', account.account, apiName, inflight, log) &&
-                            accountMetrics.bytesTotal + _inflightForCheck > accountQuota) {
-                            log.debug('Account quota exceeded', {
-                                accountId: account.account,
-                                action: apiName,
-                                inflight,
-                                quota: accountQuota,
-                                bytesTotal: accountMetrics.bytesTotal,
-                            });
-                            accountQuotaExceeded = true;
-                        }
-                        return parallelDone();
-                    });
+
+                bucketQuotaExceeded = _bucketQuotaExceeded;
+                accountQuotaExceeded = _accountQuotaExceeded;
+
+                // Inflights are inverted: in case of cleanup, we just re-issue
+                // the same API call.
+                if (_inflights) {
+                    _inflights = -_inflights;
                 }
-                return parallelDone();
-            },
-        }, err => {
-            if (err) {
-                return done(err);
-            }
-            // Inflights are inverted: in case of cleanup, we just re-issue
-            // the same API call.
-            if (_inflights) {
-                _inflights.inflight = -_inflights.inflight;
-            }
-            const finalizeQuotas = (errorFromAPI, _done) => {
-                // Quotas are cleaned only in case of error in the API
-                const quotaCleanUpStartTime = process.hrtime.bigint();
-                const code = (bucketQuotaExceeded || accountQuotaExceeded) ? 429 : 200;
-                if (errorFromAPI) {
-                    return async.parallel({
-                        deleteBucketInflights: parallelDone => {
-                            if (bucketQuota > 0) {
-                                return QuotaService.getUtilizationMetrics('bucket',
-                                    `${bucket.getName()}_${creationDate}`, null, {
-                                    action: apiName,
-                                    ..._inflights,
-                                }, parallelDone);
+
+                request.finalizerHooks.push((errorFromAPI, _done) => {
+                    const code = (bucketQuotaExceeded || accountQuotaExceeded) ? 429 : 200;
+                    const quotaCleanUpStartTime = process.hrtime.bigint();
+                    // Quotas are cleaned only in case of error in the API
+                    async.waterfall([
+                        cb => {
+                            if (errorFromAPI) {
+                                return _evaluateQuotas(bucketQuota, accountQuota, bucket, account, _inflights,
+                                    null, apiName, log, cb);
                             }
-                            return parallelDone();
+                            return cb();
                         },
-                        deleteAccountInflights: parallelDone => {
-                            if (accountQuota > 0 && account?.account) {
-                                return QuotaService.getUtilizationMetrics('account',
-                                    account.account, null, {
-                                    action: apiName,
-                                    ..._inflights,
-                                }, parallelDone);
-                            }
-                            return parallelDone();
-                        },
-                    }, err => {
-                        if (err) {
-                            log.warn('Error deleting inflights from the quota service', {
-                                error: err.name,
-                                description: err.message,
-                            });
-                        }
-                        const quotaCleanUpDuration = Number(process.hrtime.bigint() - quotaCleanUpStartTime);
-                        monitoring.quotaEvaluationDuration.labels({
-                            action: apiMethod,
-                            type,
-                            code,
-                        }).observe((quotaEvaluationDuration + quotaCleanUpDuration) / 1e9);
+                    ], () => {
+                        monitorQuotaEvaluationDuration(apiMethod, type, code, quotaEvaluationDuration +
+                            Number(process.hrtime.bigint() - quotaCleanUpStartTime));
                         return _done();
                     });
-                }
-                monitoring.quotaEvaluationDuration.labels({
-                    action: apiMethod,
-                    type,
-                    code,
-                }).observe(quotaEvaluationDuration / 1e9);
-                return _done();
-            };
-            request.finalizerHooks.push(finalizeQuotas);
-            return done();
-        });
+                });
+
+                return done();
+            });
     }, err => {
         quotaEvaluationDuration = Number(process.hrtime.bigint() - requestStartTime);
         if (err) {
--- a/lib/quotas/scuba/wrapper.js
+++ b/lib/quotas/scuba/wrapper.js
@@ -1,6 +1,7 @@
 const util = require('util');
 const { default: ScubaClient } = require('scubaclient');
 const { externalBackendHealthCheckInterval } = require('../../../constants');
+const monitoring = require('../../utilities/monitoringHandler');
 
 class ScubaClientImpl extends ScubaClient {
     constructor(config) {
@@ -36,6 +37,7 @@ class ScubaClientImpl extends ScubaClient {
             if (!this.enabled) {
                 this._log.info('Scuba health check passed, enabling quotas');
             }
+            monitoring.utilizationServiceAvailable.set(1);
             this.enabled = true;
         }).catch(err => {
             if (this.enabled) {
@@ -44,6 +46,7 @@ class ScubaClientImpl extends ScubaClient {
                     description: err.message,
                 });
             }
+            monitoring.utilizationServiceAvailable.set(0);
             this.enabled = false;
         });
     }
@@ -60,7 +63,22 @@ class ScubaClientImpl extends ScubaClient {
     }
 
     getUtilizationMetrics(metricsClass, resourceName, options, body, callback) {
-        return this._getLatestMetricsCallback(metricsClass, resourceName, options, body, callback);
+        const requestStartTime = process.hrtime.bigint();
+        return this._getLatestMetricsCallback(metricsClass, resourceName, options, body, (err, data) => {
+            const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
+            if (err) {
+                monitoring.utilizationMetricsRetrievalDuration.labels({
+                    code: err.statusCode || 500,
+                    class: metricsClass,
+                }).observe(responseTimeInNs / 1e9);
+                return callback(err);
+            }
+            monitoring.utilizationMetricsRetrievalDuration.labels({
+                code: 200,
+                class: metricsClass,
+            }).observe(responseTimeInNs / 1e9);
+            return callback(null, data);
+        });
     }
 }
 
--- a/lib/quotas/scuba/wrapper.js
+++ b/lib/quotas/scuba/wrapper.js
@@ -66,18 +66,11 @@ class ScubaClientImpl extends ScubaClient {
         const requestStartTime = process.hrtime.bigint();
         return this._getLatestMetricsCallback(metricsClass, resourceName, options, body, (err, data) => {
             const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
-            if (err) {
-                monitoring.utilizationMetricsRetrievalDuration.labels({
-                    code: err.statusCode || 500,
-                    class: metricsClass,
-                }).observe(responseTimeInNs / 1e9);
-                return callback(err);
-            }
             monitoring.utilizationMetricsRetrievalDuration.labels({
-                code: 200,
+                code: err ? (err.statusCode || 500) : 200,
                 class: metricsClass,
             }).observe(responseTimeInNs / 1e9);
-            return callback(null, data);
+            return callback(err, data);
         });
     }
 }
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -68,8 +68,8 @@ const httpResponseSizeBytes = new client.Summary({
 let quotaEvaluationDuration;
 let utilizationMetricsRetrievalDuration;
 let utilizationServiceAvailable;
-let numberOfBucketQuota;
-let numberOfAccountQuota;
+let bucketsWithQuota;
+let accountsWithQuota;
 let requestWithQuotaMetricsUnavailable;
 
 if (config.isQuotaEnabled) {
@@ -92,18 +92,18 @@ if (config.isQuotaEnabled) {
         help: 'Availability of the utilization service',
     });
 
-    numberOfBucketQuota = new client.Gauge({
+    bucketsWithQuota = new client.Gauge({
         name: 's3_cloudserver_quota_buckets_count',
         help: 'Total number of buckets quota',
     });
 
-    numberOfAccountQuota = new client.Gauge({
+    accountsWithQuota = new client.Gauge({
         name: 's3_cloudserver_quota_accounts_count',
         help: 'Total number of account quota',
     });
 
     requestWithQuotaMetricsUnavailable = new client.Counter({
-        name: 's3_cloudserver_quota_request_with_metrics_unavailable',
+        name: 's3_cloudserver_quota_unavailable_count',
         help: 'Total number of requests with quota metrics unavailable',
     });
 }
@@ -188,8 +188,8 @@ function crrCacheToProm(crrResults) {
             numberOfObjects.set(crrResults.getObjectCount.objects || 0);
         }
         if (config.isQuotaEnabled) {
-            numberOfBucketQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
-            numberOfAccountQuota.set(crrResults?.getVaultReport?.accountWithQuota || 0);
+            bucketsWithQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
+            accountsWithQuota.set(crrResults?.getVaultReport?.accountWithQuota || 0);
         }
         if (crrResults.getDataDiskUsage) {
             dataDiskAvailable.set(crrResults.getDataDiskUsage.available || 0);
@@ -271,6 +271,6 @@ module.exports = {
     quotaEvaluationDuration,
     utilizationMetricsRetrievalDuration,
     utilizationServiceAvailable,
-    numberOfBucketQuota,
+    bucketsWithQuota,
     requestWithQuotaMetricsUnavailable,
 };
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -187,6 +187,10 @@ function crrCacheToProm(crrResults) {
             numberOfBuckets.set(crrResults.getObjectCount.buckets || 0);
             numberOfObjects.set(crrResults.getObjectCount.objects || 0);
         }
+        if (config.isQuotaEnabled) {
+            numberOfBucketQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
+            numberOfAccountQuota.set(crrResults?.getVaultReport?.accountWithQuota || 0);
+        }
         if (crrResults.getDataDiskUsage) {
             dataDiskAvailable.set(crrResults.getDataDiskUsage.available || 0);
             dataDiskFree.set(crrResults.getDataDiskUsage.free || 0);
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -1,5 +1,6 @@
 const { errors } = require('arsenal');
 const client = require('prom-client');
+const { config } = require('../Config');
 
 const collectDefaultMetrics = client.collectDefaultMetrics;
 const numberOfBuckets = new client.Gauge({
@@ -63,6 +64,49 @@ const httpResponseSizeBytes = new client.Summary({
     name: 's3_cloudserver_http_response_size_bytes',
     help: 'Cloudserver HTTP response sizes in bytes',
 });
+
+let quotaEvaluationDuration;
+let utilizationMetricsRetrievalDuration;
+let utilizationServiceAvailable;
+let numberOfBucketQuota;
+let numberOfAccountQuota;
+let requestWithQuotaMetricsUnavailable;
+
+if (config.isQuotaEnabled) {
+    quotaEvaluationDuration = new client.Histogram({
+        name: 's3_cloudserver_quota_evaluation_duration_seconds',
+        help: 'Duration of the quota evaluation operation',
+        labelNames: ['action', 'code', 'type'],
+        buckets: [0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1],
+    });
+
+    utilizationMetricsRetrievalDuration = new client.Histogram({
+        name: 's3_cloudserver_quota_metrics_retrieval_duration_seconds',
+        help: 'Duration of the utilization metrics retrieval operation',
+        labelNames: ['code', 'class'],
+        buckets: [0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5],
+    });
+
+    utilizationServiceAvailable = new client.Gauge({
+        name: 's3_cloudserver_quota_utilization_service_available',
+        help: 'Availability of the utilization service',
+    });
+
+    numberOfBucketQuota = new client.Gauge({
+        name: 's3_cloudserver_quota_buckets_count',
+        help: 'Total number of buckets quota',
+    });
+
+    numberOfAccountQuota = new client.Gauge({
+        name: 's3_cloudserver_quota_accounts_count',
+        help: 'Total number of account quota',
+    });
+
+    requestWithQuotaMetricsUnavailable = new client.Counter({
+        name: 's3_cloudserver_quota_request_with_metrics_unavailable',
+        help: 'Total number of requests with quota metrics unavailable',
+    });
+}
 
 // Lifecycle duration metric, to track the completion of restore.
 // This metric is used to track the time it takes to complete the lifecycle operation (restore).
@@ -220,4 +264,9 @@ module.exports = {
     httpRequestsTotal,
     httpActiveRequests,
     lifecycleDuration,
+    quotaEvaluationDuration,
+    utilizationMetricsRetrievalDuration,
+    utilizationServiceAvailable,
+    numberOfBucketQuota,
+    requestWithQuotaMetricsUnavailable,
 };
--- a/monitoring/alerts.yaml
+++ b/monitoring/alerts.yaml
@@ -166,7 +166,7 @@ groups:
   # being unavailable
   - alert: QuotaUnavailable
     expr: |
-      sum(rate(s3_cloudserver_quota_request_with_metrics_unavailable{namespace="${namespace}",service="${service}"}[2m]))
+      sum(increase(s3_cloudserver_quota_unavailable_count{namespace="${namespace}",service="${service}"}[2m]))
         > 0
     for: 5m
     labels:
--- a/monitoring/alerts.yaml
+++ b/monitoring/alerts.yaml
@@ -26,6 +26,9 @@ x-inputs:
   - name: deleteLatencyCriticalThreshold
     type: config
     value: 1.000
+  - name: quotaUnavailabilityThreshold
+    type: config
+    value: 0.500
 
 groups:
 - name: CloudServer
@@ -132,3 +135,42 @@ groups:
     annotations:
       description: "Latency of delete object operations is more than 1s"
       summary: "Very high delete latency"
+
+  # As a platform admin I want to be alerted (warning) when the utilization metrics service is enabled
+  # but not available for at least half of the S3 services during the last minute
+  - alert: QuotaMetricsNotAvailable
+    expr: |
+      avg(avg_over_time(s3_cloudserver_quota_utilization_service_available{namespace="${namespace}",service="${service}"}[1m]))
+        < ${quotaUnavailabilityThreshold}
+    for: 1m
+    labels:
+      severity: warning
+    annotations:
+      description: "The storage metrics required for Account or S3 Bucket Quota checks are not available, the quotas are disabled."
+      summary: "Utilization metrics service not available"
+
+  # As a platform admin I want to be alerted (critical) when the utilization metrics service is enabled
+  # but not available during the last 10 minutes
+  - alert: QuotaMetricsNotAvailable
+    expr: |
+      avg(avg_over_time(s3_cloudserver_quota_utilization_service_available{namespace="${namespace}",service="${service}"}[1m]))
+        < ${quotaUnavailabilityThreshold}
+    for: 10m
+    labels:
+      severity: critical
+    annotations:
+      description: "The storage metrics required for Account or S3 Bucket Quota checks are not available, the quotas are disabled."
+      summary: "Utilization metrics service not available"
+
+  # As a platform admin I want to be alerted (critical) when quotas were not honored due to metrics
+  # being unavailable
+  - alert: QuotaUnavailable
+    expr: |
+      sum(rate(s3_cloudserver_quota_request_with_metrics_unavailable{namespace="${namespace}",service="${service}"}[2m]))
+        > 0
+    for: 5m
+    labels:
+      severity: critical
+    annotations:
+      description: "Quotas were not honored due to metrics being unavailable. If the S3 Bucket or Account was created recently, the metrics may not be available yet."
+      summary: "High number of quota requests with metrics unavailable"
--- a/monitoring/dashboard.py
+++ b/monitoring/dashboard.py
@@ -366,36 +366,26 @@ def average_latency_target(title, action="", by=""):
     )
 
 
-def average_quota_latency_target(title, code="", type="", by=""):
-    # type: (str, str, str, str) -> Target
+def average_quota_latency_target(code="", by=""):
+    # type: (str, str) -> Target
     extra = ', code=' + code if code else ""
-    extra += ', type=' + type if type else ""
     by = " by (" + by + ")" if by else ""
-    return Target(
-        expr="\n".join([
-            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_sum{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
-            "   /",
-            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
-        ]),
-        legendFormat=title,
-    )
+    return "\n".join([
+        'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_sum{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
+        "   /",
+        'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
+    ])
 
 
-def average_quota_retrieval_latency(
-        title, _class="", code="", codeExclude="", by=""):
-    # type: (str, str, str, str, str) -> Target
-    extra = ', class=' + _class if _class else ""
-    extra += ', code=' + code if code else ""
-    extra += ', code!=' + codeExclude if codeExclude else ""
+def average_quota_retrieval_latency(code="", by=""):
+    # type: (str, str) -> Target
+    extra = ', code=' + code if code else ""
     by = " by (" + by + ")" if by else ""
-    return Target(
-        expr="\n".join([
-            'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_sum{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
-            "   /",
-            'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_count{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
-        ]),
-        legendFormat=title,
-    )
+    return "\n".join([
+        'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_sum{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
+        "   /",
+        'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_count{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
+    ])
 
 
 averageLatencies = TimeSeries(
@@ -441,7 +431,7 @@ requestTime = Heatmap(
     yAxis=YAxis(format=UNITS.SECONDS),
     color=HeatmapColor(mode="opacity"),
     targets=[Target(
-        expr='sum by(le) (increase(s3_cloudserver_http_request_duration_seconds_bucket{namespace="${namespace}", job=~"$job"}[$__rate_interval]))',  # noqa: E501
+        expr='sum by(le) (increase(s3_cloudserver_http_request_duration_seconds_bucket{namespace="${namespace}", job="${job}"}[$__rate_interval]))',  # noqa: E501
         format="heatmap",
         legendFormat="{{ le }}",
     )],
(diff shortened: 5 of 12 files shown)


lib/api/bucketUpdateQuota.js
```diff
@@ -67,6 +67,7 @@ function bucketUpdateQuota(authInfo, request, log, callback) {
                 'updateBucketQuota');
             return callback(err, err.code, corsHeaders);
         } else {
+            monitoring.numberOfBucketQuota.inc();
```
BourgoisMickael: But cloudserver is stateless, so once it restarts you lose the count of buckets with quota and the panel in the dashboard becomes wrong right ?

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,31 @@ const httpResponseSizeBytes = new client.Summary({
     help: 'Cloudserver HTTP response sizes in bytes',
 });
 
+const quotaEvaluationDuration = new client.Histogram({
```
BourgoisMickael: could be nice to have a condition to disable those metrics for S3C if it doesn't need this metrics

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,31 @@ const httpResponseSizeBytes = new client.Summary({
...
+    name: 's3_quota_evaluation_duration_seconds',
```
BourgoisMickael: maybe all the names should start with `s3_cloudserver_`

monitoring/dashboard.py
```diff
@@ -525,6 +542,111 @@ def top10_errors_by_bucket(title, code):
...
+    dataSource="${DS_PROMETHEUS}",
+    fillOpacity=30,
+    lineInterpolation="smooth",
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum by (code) (rate(s3_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"2.."}[$__rate_interval]))',  # noqa: E501
+        legendFormat="Success",
+    ), Target(
+        expr='sum by (code) (rate(s3_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"4.."}[$__rate_interval]))',  # noqa: E501
+        legendFormat="Quota Exceeded",
```
BourgoisMickael: There is no specific code for quota exceeded, all the errors are quota exceeded ?

What if you receives 400 or 403 or 404 ?

And you don't want to monitor 500 ?

monitoring/dashboard.py
```diff
@@ -525,6 +542,111 @@ def top10_errors_by_bucket(title, code):
...
+    calc="mean",
+    format=UNITS.PERCENT_FORMAT,
+    min=0,
+    max=100,
+    noValue="-",
+    targets=[Target(
+        expr="\n".join([
+            'sum(rate(s3_utilization_service_state{namespace="${namespace}", state="healthy"}[$__rate_interval])) * 100',  # noqa: E501
+            "   /",
+            'sum(rate(s3_utilization_service_state{namespace="${namespace}", state=~"^(unhealthy|healthy)$"}[$__rate_interval]))',  # noqa: E501
```
BourgoisMickael: For promql I think you don't need `^$` in regex:

> Regex matches are fully anchored. A match of env=\~"foo" is treated as env=~"^foo$".

https://prometheus.io/docs/prometheus/latest/querying/basics/

```suggestion
            'sum(rate(s3_utilization_service_state{namespace="${namespace}", state=~"unhealthy|healthy"}[$__rate_interval]))',  # noqa: E501

```

lib/api/bucketUpdateQuota.js
ghost: The count will stay forever, even if you restart the service (we already follow this logic to count the number of objects & buckets)
image

To illustrate, I restarted the service after creating the bucket

lib/utilities/monitoringHandler.js
ghost: I think we will want to have this feature in S3C. For now I am only targeting Zenko branches. We have quotas (account only, and limited) in S3C today though.
But this is a valid point, we can easily disable the metrics if there is no quota evaluation in place...

lib/utilities/monitoringHandler.js
ghost: Yes, I was unsure at that time if we wanted to create a single dashboard for quotas; now that all the logic is in S3, we can keep the `s3_cloudserver_` prefix

monitoring/dashboard.py
```diff
@@ -525,6 +542,111 @@ def top10_errors_by_bucket(title, code):
...
+        average_quota_latency_target(
+            title="Account quota exceeded", type='"account"', code='~"4.."'),
+        average_quota_latency_target(
+            title="Account and Bucket quota (success)", type='"bucketAccount"',
+            code='~"2.."'),
+        average_quota_latency_target(
+            title="Account quota exceeded", type='"bucketAccount"',
+            code='~"4.."'),
+        average_quota_latency_target(
+            title="Quota exceeded (delete data)", type='"delete"'),
```
BourgoisMickael: `data` is a little vague, is it all delete action ?
Also is the delete included in the above quota on account and bucket ?

monitoring/dashboard.py
```diff
@@ -525,6 +542,111 @@ def top10_errors_by_bucket(title, code):
...
+    description=(
+        "Number of S3 buckets with quota enabled in the cluster."
+    ),
+    dataSource="${DS_PROMETHEUS}",
+    colorMode="value",
+    format=UNITS.SHORT,
+    noValue="0",
+    reduceCalc="lastNotNull",
+    targets=[Target(
+        expr='sum(s3_cloudserver_buckets_quota_count{namespace="${namespace}"})',  # noqa: E501
```
BourgoisMickael:
```suggestion
        expr='sum(s3_quota_evaluation_duration_seconds_count{namespace="${namespace}"})',  # noqa: E501
```

monitoring/dashboard.py
ghost: QuotaExceeded is 429

We can create two other filters: for the 4.. (not 429) as bad requests, and 5.. as internal service errors. What do you think?

monitoring/dashboard.py
ghost: This is indeed for the delete APIs. I can rename it "Data deletion".
This metric is not part of the above values, it's a special case when you delete data, the other ones track the writes.

monitoring/dashboard.py
BourgoisMickael: No nevermind, I wrote the wrong thing I think

lib/api/bucketUpdateQuota.js
BourgoisMickael: Yes because it got scraped by prometheus.
But now let's stay you got 100 buckets created in a few seconds and before prometheus can scrape (every 30s on artesca) the server restarts, then the 100 increment is lost.

Also the metric should restart from 0 so the counter will go down if prometheus scrape the metric with a value lower than the one before the restart I think

monitoring/dashboard.py
BourgoisMickael: Yes I think it could be nice to see different errors

monitoring/dashboard.py
BourgoisMickael: maybe you should describe that as well for other metrics

lib/api/bucketUpdateQuota.js
ghost: In this case the information is lost. What I have in mind, and this is also already done today, is to extend the `reportHandler` (called once every hour) to recompute the up to date metrics, along with a new method in arsenal to count the number of bucket (and accounts) with quota enabled. Then we keep this code as-is, and I add a description that explains the metrics are periodically re-computed.

Note: reportHandler is a special (unique) instance of cloudserver deployed in zenko (and also S3C IIRC)

BourgoisMickael: You mean the `reportHandler` will list all buckets to know the up to date number of buckets with quota, number of objects, etc ?

It seems to me this kind of (stateful) counters should maybe not be coming from prometheus but another source like logs form loki

monitoring/dashboard.py
```diff
@@ -525,6 +542,111 @@ def top10_errors_by_bucket(title, code):
     title="5xx : Top10 by Bucket", code='~"5.."'
 )
 
+
+quotaHealth = GaugePanel(
+    title="Quota service uptime",
```
BourgoisMickael: It's a little weird to show a gauge for the `mean` during `__rate_interval`

You could see more info if you graph in a time series to see over time when it was up or down and also know at the current time if it's up or down without a `mean`

lib/api/bucketUpdateQuota.js
ghost: Yes, the report handler lists all buckets every hour and compute various statistics.
The loki source cannot be used as loki is currently disabled in Artesca, so this is not reliable.

BourgoisMickael: Ok so your metric should eventualy be up to date again every hours if there are some restarts in between.

But the list doesn't contain all the information, you'll still have to send another request for each bucket to get their config ?
And this seems quit expensive if you have millions of buckets

ghost: Done in https://github.com/scality/cloudserver/pull/5556/commits/24c9a3ca6df345acb3aaa509f8ee69eae2266805
Arsenal was also updated to increment a counter when the processed bucket has a quota set. See https://github.com/scality/Arsenal/pull/2233/commits/63cc54d3fa3c859773cabeef840fe8aad9e3ccae

lib/utilities/monitoringHandler.js
ghost: Done in https://github.com/scality/cloudserver/pull/5556/commits/488f594a560879625cca9562daa15ef4df4854e0

monitoring/dashboard.py
ghost: Done here: https://github.com/scality/cloudserver/pull/5556/commits/20d2f8bacce5ca1e4b18bbee9747fd8b6a8e88f0#diff-9534a9dab1150e97d07f2e6a3541659b7d37c0aa2fb7db06e95f25fc3864078fR593
We have the success, quota exceeded, bad requests & service errors

monitoring/dashboard.py
ghost: I reworked the timeserie, now we say if we write or delete, with some details:

image

monitoring/dashboard.py
ghost: Updated with on/off display over time:

image
See https://github.com/scality/cloudserver/pull/5556/commits/a435c5357a694235ae315e02497fc025ff615646#diff-9534a9dab1150e97d07f2e6a3541659b7d37c0aa2fb7db06e95f25fc3864078fR563

Not related, but FYI, I also added another graph, with the utilization metrics retrieval latencies, I was not using the metric before, while it's useful to see the impact of the internal calls in during the quota evaluation logic:

image
(the one in the middle, bottom)
See https://github.com/scality/cloudserver/pull/5556/commits/20d2f8bacce5ca1e4b18bbee9747fd8b6a8e88f0

lib/api/bucketUpdateQuota.js
ghost: > But the list doesn't contain all the information, you'll still have to send another request for each bucket to get their config ?

It's an existing code already in prod, we iterate over all the buckets. In a zenko context the number is definitely not millions, but thousands maximum. Anyway millions is also supported, just slower. The change only add a counter, no complexity added.

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -396,6 +399,12 @@ async function validateQuotas(bucket, account, apiNames, apiMethod, inflight, lo
                 bucketName: bucket.getName(),
             });
         }
+        const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
+        monitoring.quotaEvaluationDuration.labels({
+            action: apiMethod,
+            type: 'noCheck',
+            code: 200,
+        }).observe(responseTimeInNs / 1e9);
```
BourgoisMickael: I don't see the `noCheck` in the dashboard, and I'm not sure this one provides any value, if there is no quota evaluation it could be ok to not monitor anything to not have unused metric, except if you plan to use it later ?

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -453,30 +462,62 @@ async function validateQuotas(bucket, account, apiNames, apiMethod, inflight, lo
             }
         }
 
+        if (actionWithDataDeletion[apiMethod]) {
+            const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
+            monitoring.quotaEvaluationDuration.labels({
+                action: apiMethod,
```
BourgoisMickael: as pointed here https://github.com/scality/Arsenal/pull/2233/files#r1568992349 I think if `apiMethod` comes from request, you might have the delete object version being called `objectDelete`

monitoring/dashboard.py
BourgoisMickael: I wonder if it's better to call them 4xx and 5xx errors, to not mistake a potential 403 for a 400 bad request

monitoring/dashboard.py
BourgoisMickael: for `Data write (bucket)` you could add that it's for success so we don't think it encapsulate the exceeded also

monitoring/dashboard.py
BourgoisMickael: About On / Off, I think you have a percentage with your promql, so you could keep displaying a percentage to see if you have some instances that can't connect or scuba down over time rather than a boolean

BourgoisMickael: And for the metric retrieval latencies, should you really have a big diff between this and the quota evaluation ? Also you only record success for metric retriaval

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -453,30 +462,62 @@ async function validateQuotas(bucket, account, apiNames, apiMethod, inflight, lo
...
-                await ScubaClientInstance.getLatestMetrics('bucket',
+                await ScubaClientInstance.getUtilizationMetrics('bucket',
                     `${bucket.getName()}_${creationDate}`, null, {
                     action: apiMethod,
                     inflight: -inflight,
                 });
                 if (account?.quota) {
                     // eslint-disable-next-line no-await-in-loop
-                    await ScubaClientInstance.getLatestMetrics('account',
+                    await ScubaClientInstance.getUtilizationMetrics('account',
```
BourgoisMickael: would it be more interesting to have another endpoint in scuba to retrieve both bucket and account metrics so you send only one request ?

BourgoisMickael: and also where you could pass both `inflight` and `-inflight` for delete to reduce the number of api calls

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -453,30 +462,62 @@ async function validateQuotas(bucket, account, apiNames, apiMethod, inflight, lo
...
+            type = 'account';
+        }
+        monitoring.quotaEvaluationDuration.labels({
+            action: apiMethod,
+            type,
+            code: 200,
+        }).observe(responseTimeInNs / 1e9);
+
         return callback();
     } catch (err) {
```
BourgoisMickael: Should you observe the quota evaluation duration in the catch for errors from Scuba ? I don't see how you got some Bad Requests in your graph otherwise

lib/scuba/wrapper.js
```diff
@@ -36,6 +40,9 @@ class ScubaClientImpl extends ScubaClient {
                     description: err.message,
                 });
             }
+            monitoring.utilizationServiceHealth.labels({
+                state: 'unhealthy',
+            }).inc();
```
BourgoisMickael: Incrementing counter is not ideal here if you want to have an accurate healthy percentage in your panel.
If scuba goes up and down multiple times, you'll have counters that keep incrementing.

You might want to use a gauge to set a boolean like this:
```js
// success
utilizationServiceHealth.labels(healthy).set(1)
utilizationServiceHealth.labels(unhealthy).set(0)
// error
utilizationServiceHealth.labels(healthy).set(0)
utilizationServiceHealth.labels(unhealthy).set(1)
```

monitoring/dashboard.py
ghost: Yes, metric retrieval can be done twice in an API call, in case we need to cleanup when quota is exceeded. We still want to be able the latency with scuba, independently from the quota evaluation logic

(please feel free to challenge that)

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Yes, this is fine, we do not need to differentiate between versioned and non-versioned deletes

lib/api/apiUtils/authorization/permissionChecks.js
ghost: > would it be more interesting to have another endpoint in scuba to retrieve both bucket and account metrics so you send only one request ?

Yes, this is something we can do; for now the use case will be to have only bucket quotas, but in case they have both, we can add another API

But this breaks compatibility with the existing scuba APIs, so better doing that in a future iteration to improve the performance, what do you think?

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Yes, I agree that there is no point in monitoring "nothing", I will remove it

monitoring/dashboard.py
BourgoisMickael: I think the utilization metric retrieval metric is the same as `scuba_http_request_duration_seconds{action="GetMetrics"}`
(https://github.com/scality/scuba/pull/104/files#diff-7c16f2a12075d20e3640e31bcb11cc1b87aad07a0fc048ace705020732a5a2cfR21-R28) but with network time over it.

But then maybe scuba metrics can record request that come from other places than cloudserver, so maybe you should actually keep this

___
Otherwise I posted in another comment, it could be nice to have another scuba endpoint to aggregate all the requests into one, especially if it's a code path that's called often.

lib/scuba/wrapper.js
```diff
@@ -50,6 +57,20 @@ class ScubaClientImpl extends ScubaClient {
...
+
+    async getUtilizationMetrics(metricsClass, resourceName, options = null, body = null) {
+        const requestStartTime = process.hrtime.bigint();
+        return this.getLatestMetrics(metricsClass, resourceName, options, body).then(data => {
+            const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
+            monitoring.utilizationMetricsRetrievalDuration.labels({
+                method: 'POST',
+                action: 'getUtilizationMetrics',
+                code: 200,
+                class: metricsClass,
```
BourgoisMickael: Should you also monitor errors ? Otherwise the `code` could be removed, but your dashboard is ready to accept errors on this metric.

Constant `method` and `action` labels here seems unnecessary and could be removed.

`class` is useful ? Should you display it on your dashboard ?

monitoring/alerts.yaml
```diff
@@ -132,3 +135,17 @@ groups:
     annotations:
       description: "Latency of delete object operations is more than 1s"
       summary: "Very high delete latency"
+
+  # As a platform admin I want to be alerted (warning) when the utilization metrics service is enabled
+  # but not available during the last 5 minutes
+  - alert: UtilizationMetricsServiceNotAvailable
+    expr: |
+      sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}",state="healthy"}[1m])) * 100
+          / sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}",state=~"^(unhealthy|healthy)$"}[1m]))
```
BourgoisMickael:
```suggestion
          / sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}",state=~"unhealthy|healthy"}[1m]))
```

same as the dashboard one

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,31 @@ const httpResponseSizeBytes = new client.Summary({
...
+
+const utilizationMetricsRetrievalDuration = new client.Histogram({
+    name: 's3_cloudserver_utilization_metrics_retrieval_duration_seconds',
+    help: 'Duration of the utilization metrics retrieval operation',
+    labelNames: [...labelNames, 'class'],
+    buckets: [0.0001, 0.00025, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5],
+});
+
+const utilizationServiceHealth = new client.Counter({
+    name: 's3_cloudserver_utilization_service_state',
```
BourgoisMickael: why not call it scuba ?

lib/api/apiUtils/authorization/permissionChecks.js
BourgoisMickael: yep, maybe make a ticket so we don't forget

ghost: I created an epic and the ticket is inside: https://scality.atlassian.net/browse/ARTESCA-12124
We already identified a bunch of things we need to improve, so this epic will be useful for that.

monitoring/dashboard.py
ghost: > But then maybe scuba metrics can record request that come from other places than cloudserver, so maybe you should actually keep this

Yes it's indeed the same, but this is useful to see the impact of internal network communications, as seen by S3, as the scuba metrics may be, in the future, including public ones with authz, expected to be slower...

For the other endpoint, yes, see https://scality.atlassian.net/browse/ZENKO-4801 (the current epic supports account, but the use case is bucket)

lib/scuba/wrapper.js
ghost: I will remove method and action indeed.

The `class` is useful because I display the latency for the following cases:
- bucket, success
- bucket, error
- account, success
- account, error

Later we will also have both, if we mutualize in a single API

lib/utilities/monitoringHandler.js
ghost: I don't know if we want to use Scuba publicly, because it was supposed to be dropped in favor of SUR, but anyway the reason is that I wanted to dissociate the naming of the metric with the backend, as the backend can change independently of cloudserver...

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Done in https://github.com/scality/cloudserver/pull/5556/commits/0006a0f07d83cb818352e0f6caec6662323aee19

I chose the error code 202, because in case of error we still process the request. Note that the error from scubaclient is also monitored, so information is not lost.

lib/scuba/wrapper.js
ghost: Done in https://github.com/scality/cloudserver/pull/5556/commits/8ef63ab5cd687bbe34c0a5eb42f8d56821a641e2 (dashboard JSON is updated in a subsequent commit)

lib/scuba/wrapper.js
ghost: Done in https://github.com/scality/cloudserver/pull/5556/commits/460404bd7391f06347009eb2e5bd4063db5cccca

monitoring/alerts.yaml
ghost: Done in https://github.com/scality/cloudserver/pull/5556/commits/95ef7fcd9db20f1ae3b11e5590c27e7a76fc45cf

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Removed in https://github.com/scality/cloudserver/pull/5556/commits/9facef770aed6b5a13ad72993095f74d20f7eb09

monitoring/alerts.yaml
```diff
@@ -132,3 +135,17 @@ groups:
...
       description: "Latency of delete object operations is more than 1s"
       summary: "Very high delete latency"
+
+  # As a platform admin I want to be alerted (warning) when the utilization metrics service is enabled
+  # but not available during the last 5 minutes
+  - alert: UtilizationMetricsServiceNotAvailable
+    expr: |
+      sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}",state="healthy"}[1m])) * 100
+          / sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}",state=~"unhealthy|healthy"}[1m]))
+        < ${utilizationServiceUnavailabilityThreshold}
```
francoisferrand: should filter on `${service}`

monitoring/dashboard.py
```diff
@@ -366,6 +366,38 @@ def average_latency_target(title, action="", by=""):
...
+def average_quota_latency_target(title, code="", type="", by=""):
+    # type: (str, str, str, str) -> Target
+    extra = ', code=' + code if code else ""
+    extra += ', type=' + type if type else ""
+    by = " by (" + by + ")" if by else ""
+    return Target(
+        expr="\n".join([
+            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_sum{namespace="${namespace}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
+            "   /",
+            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
```
francoisferrand: need to filter on service/job

monitoring/dashboard.py
```diff
@@ -366,6 +366,38 @@ def average_latency_target(title, action="", by=""):
...
+    # type: (str, str, str, str, str) -> Target
+    extra = ', class=' + _class if _class else ""
+    extra += ', code=' + code if code else ""
+    extra += ', code!=' + codeExclude if codeExclude else ""
+    by = " by (" + by + ")" if by else ""
+    return Target(
+        expr="\n".join([
+            'sum(rate(s3_cloudserver_utilization_metrics_retrieval_duration_seconds_sum{namespace="${namespace}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
+            "   /",
+            'sum(rate(s3_cloudserver_utilization_metrics_retrieval_duration_seconds_count{namespace="${namespace}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
```
francoisferrand: need to filter on service/job

lib/utilities/monitoringHandler.js
francoisferrand: what is the point of disabling the metrics?
if the metric is not "used" in the code (e.g. never `observed()`), no time series will be generated...

lib/utilities/monitoringHandler.js
francoisferrand: `utilization_service_state` seems out of place (the state of the utilization service should be monitored in that service... here we monitor the availability of the feature in cloudserver which depends on it, i.e. the impact of unavailability of that service?), and it feels like we loose the "relationship" with the quota.

maybe something like `s3_cloudserver_quota_availaibility` or `s3_cloudserver_quota_state` ?

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,36 @@ const httpResponseSizeBytes = new client.Summary({
...
+
+const numberOfBucketQuota = new client.Gauge({
+    name: 's3_cloudserver_buckets_quota_count',
+    help: 'Total number of buckets quota',
+});
+
+const numberOfAccountQuota = new client.Gauge({
+    name: 's3_cloudserver_account_quota_count',
+    help: 'Total number of account quota',
+});
```
francoisferrand: `s3_cloudserver_quota_buckets_count` and `s3_cloudserver_quota_accounts_count`

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,36 @@ const httpResponseSizeBytes = new client.Summary({
...
 
+const quotaEvaluationDuration = new client.Histogram({
+    name: 's3_cloudserver_quota_evaluation_duration_seconds',
+    help: 'Duration of the quota evaluation operation',
+    labelNames: ['action', 'code', 'type'],
+    buckets: [0.0001, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5],
+});
+
+const utilizationMetricsRetrievalDuration = new client.Histogram({
+    name: 's3_cloudserver_utilization_metrics_retrieval_duration_seconds',
```
francoisferrand: do we need to differentiate both durations ?
i.e. once the utilization metrics have been read, quota evaluation is just a few ops, with no more network ops or complex computation?

francoisferrand: i think we should keep the same "prefix" for all these metrics, to make it easier to manage the (long) list of metrics...

not sure how best to do it in this case (unless it makes sens to remove the metrics, c.f. earlier comment 😁), the only thing I could come up with is `s3_cloudserver_quota_retrieval_duration_seconds` (which is not _correct_, but probably good enough...)

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,36 @@ const httpResponseSizeBytes = new client.Summary({
     help: 'Cloudserver HTTP response sizes in bytes',
 });
 
+const quotaEvaluationDuration = new client.Histogram({
+    name: 's3_cloudserver_quota_evaluation_duration_seconds',
+    help: 'Duration of the quota evaluation operation',
+    labelNames: ['action', 'code', 'type'],
```
francoisferrand: do we need to filter by `code` or `action` ? i got mixed feeling here:
* we don't really care about the action with regard to the _duration_, since all these actions should be affected the same way?
* the code (= result of the evaluation) is definitely not relevant for the _duration_
* however, it could indeed be interesting to have this labels to allow checking which requests cause the most quota checks or the quota check success/rate
* but at the same time, we can probably get this information by just looking at the request count (which has action & code).... the only thing missing from the request count may be the `type` of quota check (i.e. none, bucket, account, bucket+account)

francoisferrand: we may have either account or bucket quotas, but could also have both : how is this managed in the `type` label? should we maybe have 2 labels instead (`bucket` and `account`) ?

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,36 @@ const httpResponseSizeBytes = new client.Summary({
...
+    buckets: [0.0001, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5],
```
francoisferrand: `0.1ms` does not seem realistic...
Also, you mentioned the impact (in your test) was around 1ms, so i guess we should have more buckets in this range: maybe `[1ms, 2,5ms, 5ms, 10ms, 25ms, 50ms, 100ms, 200ms, 400ms, 800ms]` ?

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,36 @@ const httpResponseSizeBytes = new client.Summary({
...
+    name: 's3_cloudserver_utilization_metrics_retrieval_duration_seconds',
+    help: 'Duration of the utilization metrics retrieval operation',
+    labelNames: ['code', 'class'],
+    buckets: [0.0001, 0.00025, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5],
+});
+
+const utilizationServiceHealth = new client.Gauge({
+    name: 's3_cloudserver_utilization_service_state',
+    help: 'Availability of the utilization service',
+    labelNames: ['state'],
```
francoisferrand: what are the possible states? and values for the metric?

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -385,10 +387,19 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
...
  */
 async function validateQuotas(bucket, account, apiNames, apiMethod, inflight, log, callback) {
+    const requestStartTime = process.hrtime.bigint();
     const bucketQuota = bucket.getQuota();
     const accountQuota = account?.quota || 0;
     let bucketQuotaExceeded = false;
     let accountQuotaExceeded = false;
+    let type = 'bucket';
+    if (bucketQuota && accountQuota) {
+        type = 'bucketAccount';
```
francoisferrand: probably should make it parsable: `bucket+account` or `bucket,account` ?
(or use separate labels, c.f. other comment)

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -385,10 +387,19 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
  * @returns {boolean} - true if the quota is valid, false otherwise
  */
 async function validateQuotas(bucket, account, apiNames, apiMethod, inflight, log, callback) {
```
francoisferrand: probably not in this PR, but this function is very weird: being both async and with a callback...

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -453,39 +462,70 @@ async function validateQuotas(bucket, account, apiNames, apiMethod, inflight, lo
             }
         }
 
+        if (actionWithDataDeletion[apiMethod]) {
+            const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
+            monitoring.quotaEvaluationDuration.labels({
```
francoisferrand: can we move this up (where the function is called), to avoid duplication?

lib/scuba/wrapper.js
```diff
@@ -28,6 +29,8 @@ class ScubaClientImpl extends ScubaClient {
             if (!this.enabled) {
                 this._log.info('Scuba health check passed, enabling quotas');
             }
+            monitoring.utilizationServiceHealth.labels({ state: 'healthy' }).set(1);
+            monitoring.utilizationServiceHealth.labels({ state: 'unhealthy' }).set(0);
```
francoisferrand: avoid multiple metrics like this does not seem very "user friendly": a single `available` metric with a pseuod-boolean value (0 = no, 1 = yes), like the `up` metrics, seems a more standard approach...

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +64,36 @@ const httpResponseSizeBytes = new client.Summary({
...
+});
+
+const numberOfBucketQuota = new client.Gauge({
```
francoisferrand, line 87: numberOfXXX → XXXXCount

monitoring/dashboard.py
```diff
@@ -386,6 +418,7 @@ def average_latency_target(title, action="", by=""):
     ],
 )
 
+
```
francoisferrand: unrealted

monitoring/dashboard.py
```diff
@@ -400,6 +433,7 @@ def average_latency_target(title, action="", by=""):
     ]
 )
 
+
```
francoisferrand: unrelated

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+    title="Quota service uptime",
+    dataSource="${DS_PROMETHEUS}",
+    lineInterpolation="stepAfter",
+    fillOpacity=30,
+    unit=UNITS.PERCENT_FORMAT,
+    targets=[Target(
+        expr="\n".join([
+            'sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}", state="healthy"}[$__rate_interval])) * 100',  # noqa: E501
+            "   /",
+            'sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}", state=~"unhealthy|healthy"}[$__rate_interval]))',  # noqa: E501
```
francoisferrand: for the `uptime`, i think it is enough to graph `s3_cloudserver_utilization_service_state{state=healthy}` : it is already 0 when down and 1 when up...

or may you are trying to compute the % of time the service is available _over time_ ?
in that case, it should be more something like `avg_over_time(s3_cloudserver_utilization_service_state{state=healthy})`...

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+
+quotaHealth = TimeSeries(
+    title="Quota service uptime",
+    dataSource="${DS_PROMETHEUS}",
+    lineInterpolation="stepAfter",
+    fillOpacity=30,
+    unit=UNITS.PERCENT_FORMAT,
+    targets=[Target(
+        expr="\n".join([
+            'sum(rate(s3_cloudserver_utilization_service_state{namespace="${namespace}", state="healthy"}[$__rate_interval])) * 100',  # noqa: E501
```
francoisferrand: must not use `rate` on gauge (only on counters)

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+
+quotaStatusCode = TimeSeries(
+    title="Quota evaluation status code over time",
+    dataSource="${DS_PROMETHEUS}",
+    fillOpacity=30,
+    lineInterpolation="smooth",
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum by (code) (rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"2.."}[$__rate_interval]))',  # noqa: E501
+        legendFormat="Success",
```
francoisferrand: * either you `sum by(code)`, to get one time serie per code → the legendFormat must be based on the code (typically `{{ code }}`)
* or you want a single serie "success" → use a simpler sum, without grouping
* 
```suggestion
        expr='sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"2.."}[$__rate_interval]))',  # noqa: E501
        legendFormat="Success",
```

francoisferrand: must also filter by job/service

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+    ), Target(
+        expr='sum by (code) (rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code="429"}[$__rate_interval]))',  # noqa: E501
+        legendFormat="Quota Exceeded",
+    ), Target(
+        expr='sum by (code) (rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"4..", code!="429"}[$__rate_interval]))',  # noqa: E501
+        legendFormat="Status 4xx",
+    ), Target(
+        expr='sum by (code) (rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"5.."}[$__rate_interval]))',  # noqa: E501
+        legendFormat="Status 5xx",
```
francoisferrand: same

francoisferrand: are these 4xx and 5xx status relevant for this metric? at the end of the evaluation, we can only have either 'proceed' or '429'... Even if the query eventually ends up with another error code, the metric will have already recorded either 200 or 429...

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+        "This value is computed asynchronously, and update "
+        "may be delayed up to 1h."
+    ),
+    dataSource="${DS_PROMETHEUS}",
+    colorMode="value",
+    format=UNITS.SHORT,
+    noValue="0",
+    reduceCalc="lastNotNull",
+    targets=[Target(
+        expr='sum(s3_cloudserver_account_quota_count{namespace="${namespace}"})',  # noqa: E501
```
francoisferrand: `max()` may be more appropriate than `sum`, it case we have multiple series at the same time (e.g. pod update maybe?)

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+    title="Accounts with quota",
+    description=(
+        "Number of accounts with quota enabled in the cluster.\n"
+        "This value is computed asynchronously, and update "
+        "may be delayed up to 1h."
+    ),
+    dataSource="${DS_PROMETHEUS}",
+    colorMode="value",
+    format=UNITS.SHORT,
+    noValue="0",
```
francoisferrand: if we don't have a value, we may want to make it visibile we don't actually know : instead of reporting a possibly false `0`...

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+    title="Buckets with quota",
+    description=(
+        "Number of S3 buckets with quota enabled in the cluster.\n"
+        "This value is computed asynchronously, and update "
+        "may be delayed up to 1h."
+    ),
+    dataSource="${DS_PROMETHEUS}",
+    colorMode="value",
+    format=UNITS.SHORT,
+    noValue="0",
```
francoisferrand: if we don't have a value, we may want to make it visibile we don't actually know : instead of reporting a possibly false 0...

monitoring/dashboard.py
```diff
@@ -525,6 +559,167 @@ def top10_errors_by_bucket(title, code):
...
+    reduceCalc="lastNotNull",
+    targets=[Target(
+        expr='sum(s3_cloudserver_buckets_quota_count{namespace="${namespace}"})',  # noqa: E501
```
francoisferrand: max() may be more appropriate than sum, it case we have multiple series at the same time (e.g. pod update maybe?)

lib/utilities/monitoringHandler.js
ghost: Yes, because we might do more than one call to scuba in a single evaluation loop, for example to remove the inflights after detecting a quota exceeded, this help shaving visibility on the internal call latencies, versus the quota evaluation logic durations (with different cases)

ghost: I will use `s3_cloudserver_quota_metrics_retrieval_duration_seconds`

lib/utilities/monitoringHandler.js
ghost: Depending on the API (write vs delete) and the response (quota exceeded vs passing), both action and code are relevant to discriminate between the different flows of the quota evaluation logic
For example quota exceeded will have more impact as we need to clean the inflights in case of quota exceeded

ghost: Right now we have 3 different strings (bucket, account, bucketAccount), if we want to add more types later, I thought it made sense to have a single label for it...

lib/utilities/monitoringHandler.js
ghost: Closing as this will be removed, we will use a pseudo boolean.

lib/api/apiUtils/authorization/permissionChecks.js
ghost: I will make it parsable indeed, will choose the first option

lib/api/apiUtils/authorization/permissionChecks.js
ghost: As discussed I will callbackify the call to scubaclient in the PR this one is based on

lib/api/apiUtils/authorization/permissionChecks.js
ghost: No because this is called in a centralized place, and in specific APIs sometimes (copy object and multi object deletes need that, as we do not have the obj metadata at the authorization time)...
But we could make another layer of function, do you see any benefit of doing so?

lib/utilities/monitoringHandler.js
BourgoisMickael: Yes, for metrics that have labels like this one, no time series is generated, but for other metrics without label (like `s3_cloudserver_quota_buckets_count`) they are initialized at 0 and will return 0 at every scrape even if unused, so prometheus will keep storing 0 (every 15s on the ring) and increase storage size.

As the prometheus/thanos DB can take a lot of space, it would be nice to not scrape unused metrics for the feature that's not enabled.

lib/utilities/monitoringHandler.js
BourgoisMickael: Then you can also change the buckets for the metric_retrieval_duration metric as the quota depends on this one and could be double the metric retrieval time ?

And then maybe change as well the buckets in scuba to match with the retrieval buckets ?

___

Also I wonder if it could be possible that  the time will differ between a mongodb backend for artesca and a leveldb/metadata backend for the ring and we should have different buckets configured for each product ?

monitoring/dashboard.py
```diff
@@ -630,6 +819,18 @@ def top10_errors_by_bucket(title, code):
...
             ], height=8),
+            RowPanel(title="Quotas"),
+            layout.row([
+                layout.resize([quotaHealth], width=6),
+                layout.resize([quotaStatusCode], width=6),
+                quotaByAction,
+            ], height=8),
+            layout.row([
+                layout.resize([bucketQuotaCounter], width=3),
+                layout.resize([accountQuotaCounter], width=3),
```
francoisferrand: we probably want to stack these on top of each other, will look better (they look quite stretched, due to the small width & large height)

monitoring/dashboard.py
```diff
@@ -630,6 +819,18 @@ def top10_errors_by_bucket(title, code):
...
+                layout.resize([averageMetricsRetrievalLatencies], width=6),
+                averageQuotaLatencies,
```
francoisferrand: we store histogram metrics, but only display the average → we should try to use the information to provide more info, like the distribution (heatmap) or 95th percentile...

lib/utilities/monitoringHandler.js
ghost: Is this approach answering your concerns? https://github.com/scality/cloudserver/pull/5556/files#diff-3daa33d35f5375a7070d74a0b174c01ab8b982392db4025ebc49cf3690aa16bcR75

lib/utilities/monitoringHandler.js
ghost: Done in https://github.com/scality/cloudserver/pull/5556/files#diff-3daa33d35f5375a7070d74a0b174c01ab8b982392db4025ebc49cf3690aa16bcR91

I still have the utilization service inside as this is what defines the availability of the feature

lib/utilities/monitoringHandler.js
ghost: See https://github.com/scality/cloudserver/pull/5556/files#diff-3daa33d35f5375a7070d74a0b174c01ab8b982392db4025ebc49cf3690aa16bcR84

lib/utilities/monitoringHandler.js
ghost: Updated in https://github.com/scality/cloudserver/pull/5556/files#diff-3daa33d35f5375a7070d74a0b174c01ab8b982392db4025ebc49cf3690aa16bcR80

I will assess if changes are needed in Scuba, as the latency in scuba is lower (no network latency added, as compared with cloudserver)

lib/api/apiUtils/authorization/permissionChecks.js
ghost: See https://github.com/scality/cloudserver/pull/5556/files#diff-773344fcfc0f64145ff95eb6eccd556588c19449dc2c263a2532d4030bf00dffR407

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Callbackify was done in the base PR.

lib/scuba/wrapper.js
ghost: Now we have no label https://github.com/scality/cloudserver/pull/5556/files#diff-7a879f848f4d512bac1250da6f152d89b8feab003de8d9fc6fa202cb6b17dcf5R34

monitoring/dashboard.py
ghost: Yes, I wanted to have the %, but also the average for all cloudserver instances, is this correct (also similar to the alert, so it's logical) https://github.com/scality/cloudserver/pull/5556/files#diff-9534a9dab1150e97d07f2e6a3541659b7d37c0aa2fb7db06e95f25fc3864078fR568

monitoring/dashboard.py
ghost: Job added.

ghost: Removed.

monitoring/dashboard.py
ghost: Indeed, see https://github.com/scality/cloudserver/pull/5556/files#diff-9534a9dab1150e97d07f2e6a3541659b7d37c0aa2fb7db06e95f25fc3864078fR696

monitoring/dashboard.py
ghost: Removed.

monitoring/dashboard.py
ghost: Done: https://github.com/scality/cloudserver/pull/5556/files#diff-9534a9dab1150e97d07f2e6a3541659b7d37c0aa2fb7db06e95f25fc3864078fR718

monitoring/dashboard.py
ghost: Reworked the layout: https://github.com/scality/cloudserver/pull/5556/files#diff-9534a9dab1150e97d07f2e6a3541659b7d37c0aa2fb7db06e95f25fc3864078fR832

image

monitoring/dashboard.py
ghost: What do you think of the heatmap I added? (see comment above)

lib/utilities/monitoringHandler.js
BourgoisMickael: Yep exactly. And we could do the same in scuba repo for mongo specific metrics

monitoring/dashboard.py
```diff
@@ -525,6 +557,173 @@ def top10_errors_by_bucket(title, code):
     title="5xx : Top10 by Bucket", code='~"5.."'
 )
 
+
+quotaHealth = TimeSeries(
```
BourgoisMickael, line 551: you might want to put a max at `100%` as it shows 200% in your screenshot.

In the legend you could put the latest value so we don't have to hover to see the tooltip

monitoring/dashboard.py
```diff
@@ -525,6 +557,173 @@ def top10_errors_by_bucket(title, code):
...
+)
+
+
+averageQuotaDuration = Heatmap(
+    title="Quota evaluation duration",
+    dataSource="${DS_PROMETHEUS}",
+    dataFormat="tsbuckets",
+    maxDataPoints=25,
+    tooltip=Tooltip(show=True, showHistogram=True),
+    yAxis=YAxis(format=UNITS.DURATION_SECONDS),
```
BourgoisMickael: I don't know why grafana displays that duration seconds with `milliseconds` instead of `ms` but if you pick only seconds instead of duration seconds, it seems to abbreviate correctly and display `ms`.
But I'm not sure what's the difference between the 2 units

monitoring/dashboard.py
ghost: Unfortunately it's not yet available in the python lib we use, I'll see if I can do something about it.

BourgoisMickael (approved):

monitoring/dashboard.py
```diff
@@ -525,6 +557,173 @@ def top10_errors_by_bucket(title, code):
...
+
+quotaHealth = TimeSeries(
+    title="Quota service uptime",
+    dataSource="${DS_PROMETHEUS}",
+    lineInterpolation="stepAfter",
+    fillOpacity=30,
+    unit=UNITS.PERCENT_FORMAT,
+    targets=[Target(
+        expr='avg(avg_over_time(s3_cloudserver_quota_utilization_service_available{namespace="${namespace}",job=~"$job"}[$__rate_interval])) * 100',  # noqa: E501
+        legendFormat="State",
```
francoisferrand: legend not needed, since there is only one series

ghost: Updated

monitoring/dashboard.py
```diff
@@ -525,6 +557,172 @@ def top10_errors_by_bucket(title, code):
...
+averageQuotaDuration = Heatmap(
+    title="Quota evaluation duration",
+    dataSource="${DS_PROMETHEUS}",
+    dataFormat="tsbuckets",
+    maxDataPoints=25,
+    tooltip=Tooltip(show=True, showHistogram=True),
+    yAxis=YAxis(format=UNITS.SECONDS),
+    color=HeatmapColor(mode="opacity"),
+    targets=[Target(
+        expr='sum by(le) (increase(s3_cloudserver_quota_evacdluation_duration_seconds_bucket{namespace="${namespace}", job=~"$job"}[$__rate_interval]))',  # noqa: E501
```
francoisferrand: by convention, prefer to use braces around _inputs_ variables (everywhere in this file): makes it easier to see with syntax highlighting (esp. in github) and differentiate from _template_ variables (managed by grafana at runtime).

Also, you have a single job, so you can use the simpler (and likely faster) exact match operator.

```suggestion
        expr='sum by(le) (increase(s3_cloudserver_quota_evacdluation_duration_seconds_bucket{namespace="${namespace}", job="${job}"}[$__rate_interval]))',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +557,172 @@ def top10_errors_by_bucket(title, code):
...
+    targets=[
+        average_quota_retrieval_latency(title="Bucket (success)",
+                                        _class='"bucket"', code='~"2.."'),
+        average_quota_retrieval_latency(title="Account (success)",
+                                        _class='"account"', code='~"2.."'),
+        average_quota_retrieval_latency(title="Bucket (error)",
+                                        _class='"bucket"', code='~"4..|5.."'),
+        average_quota_retrieval_latency(title="Account (error)",
+                                        _class='"account"', code='~"4..|5.."'),
+    ],
```
francoisferrand: what about `bucket+account` ?

To avoid duplication, you can simply define 2 targets where the prometheus query has a `by(class)` clause, something like (using functions may be fine, but probably cleaner to use them only to build the expression):

```
targets:[
    Target(
        expr: 'sum(rate(quota_metrics_retrieval{code=~"2.."}[$__rate_interval])) by(class)',
        legendFormat: '{{ class }} (success)',
    ),
    Target(
        expr: 'sum(rate(quota_metrics_retrieval{code=~"4..|5.."}[$__rate_interval])) by(class)',
        legendFormat: '{{ class }} (error)',
    )
]
```

monitoring/dashboard.py
```diff
@@ -525,6 +557,172 @@ def top10_errors_by_bucket(title, code):
...
+        "This value is computed asynchronously, and update "
+        "may be delayed up to 1h."
+    ),
+    dataSource="${DS_PROMETHEUS}",
+    colorMode="value",
+    format=UNITS.SHORT,
+    noValue="-",
+    reduceCalc="lastNotNull",
+    targets=[Target(
+        expr='max(s3_cloudserver_quota_buckets_count{namespace="${namespace}", job=~"${reportJob}"})',  # noqa: E501
```
francoisferrand, line 687:
```suggestion
        expr='max(s3_cloudserver_quota_buckets_count{namespace="${namespace}", job="${reportJob}"})',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +557,172 @@ def top10_errors_by_bucket(title, code):
...
+        "This value is computed asynchronously, and update "
+        "may be delayed up to 1h."
+    ),
+    dataSource="${DS_PROMETHEUS}",
+    colorMode="value",
+    format=UNITS.SHORT,
+    noValue="-",
+    reduceCalc="lastNotNull",
+    targets=[Target(
+        expr='max(s3_cloudserver_quota_accounts_count{namespace="${namespace}", job=~"${reportJob}"})',  # noqa: E501
```
francoisferrand, line 709:
```suggestion
        expr='max(s3_cloudserver_quota_accounts_count{namespace="${namespace}", job="${reportJob}"})',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +557,172 @@ def top10_errors_by_bucket(title, code):
...
+    )],
+    thresholds=[
+        Threshold("#808080", 0, 0.0),
+        Threshold("blue", 1, 0.0),
+    ],
+)
+
```
francoisferrand, line 716: `s3_cloudserver_quota_request_with_metrics_unavailable` is maybe the most important metric, should be in dashboard (generally: if you have an alert, you should be able to monitor the metric...)

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -225,17 +246,20 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
...
+        const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
         if (err) {
             log.warn('Error getting metrics from the quota service, allowing the request', {
                 error: err.name,
                 description: err.message,
             });
         }
         if (actionWithDataDeletion[apiMethod]) {
-            return callback();
+            monitorQuotaEvaluation(apiMethod, type, 200, responseTimeInNs / 1e9);
```
francoisferrand: duplicated with call on line 262

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -225,17 +246,20 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
...
         } else if (bucketQuotaExceeded || accountQuotaExceeded) {
+            monitorQuotaEvaluation(apiMethod, type, 429, responseTimeInNs / 1e9);
             return callback(errors.QuotaExceeded);
         }
+        monitorQuotaEvaluation(apiMethod, type, 200, responseTimeInNs / 1e9);
```
francoisferrand: this computes the metrics for the call to `validateQuotas` ; however, it does not account for the eventual "cleanup" which may happen at the end of failed request...
→ to be "precise", we should store the value here (in the request), always call the cleanup callback, and there also add the cleanup time (if any) before observing the value...
→ otherwise, maybe we should do a second measure if the cleanup is executed.... but then maybe the metric is not correctly named? or maybe the cleanup can be observed via another metric?
→ or maybe this is correct, as we are really only interested in the actual _evaluation_ part, and not the "total cost of quota" ? That may be fine, but should be made clear in the comments (or description?) of the metric

ghost: I agree that monitoring should be adapted with the new hooks, and I chose the first approach, where we consistently finalize the quota logic in a final function, allowing us to monitor in a single place, and in a consistent way, the quota evaluation, see https://github.com/scality/cloudserver/pull/5556/commits/890b53515241d0d525bfeac43ee48ccad13ba314

The main point is that we split the durations; if we clean the quotas, we will add both the initial and the cleanup...

monitoring/dashboard.py
ghost: Were you thinking of something like this https://github.com/scality/cloudserver/pull/5556/commits/5b7dd3b8413927327636c3fad2c021369bb946be ?
I'll deploy it tomorrow and see if it needs some updates / improvements, and I will also try to display the `s3_cloudserver_quota_request_with_metrics_unavailable` metric.

ghost: Tested in my platform with success
image

Also fixed the type for deletes, as it was never set.

monitoring/dashboard.py
ghost: Added as a timeserie (middle-top)
image

This directly maps with the alert, as we have one if we have >0 values for 5min (maybe it should be less?)

The description of the critical alert was also updated, to be understandable by someone looking at the platform without context

lib/utilities/monitoringHandler.js
```diff
@@ -64,6 +65,49 @@ const httpResponseSizeBytes = new client.Summary({
...
+        help: 'Total number of buckets quota',
+    });
+
+    numberOfAccountQuota = new client.Gauge({
+        name: 's3_cloudserver_quota_accounts_count',
+        help: 'Total number of account quota',
+    });
+
+    requestWithQuotaMetricsUnavailable = new client.Counter({
+        name: 's3_cloudserver_quota_request_with_metrics_unavailable',
```
francoisferrand: i guess just`s3_cloudserver_quota_unavailable_count` is enough ?
i.e. the number of "events" where quota were not available...

lib/utilities/monitoringHandler.js
```diff
@@ -143,6 +187,10 @@ function crrCacheToProm(crrResults) {
             numberOfBuckets.set(crrResults.getObjectCount.buckets || 0);
             numberOfObjects.set(crrResults.getObjectCount.objects || 0);
         }
+        if (config.isQuotaEnabled) {
+            numberOfBucketQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
+            numberOfAccountQuota.set(crrResults?.getVaultReport?.accountWithQuota || 0);
```
francoisferrand: nit:
```suggestion
            bucketsWithQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
            accountsWithQuota.set(crrResults?.getVaultReport?.accountWithQuota || 0);
```

monitoring/alerts.yaml
```diff
@@ -132,3 +135,42 @@ groups:
...
+      severity: critical
+    annotations:
+      description: "The storage metrics required for Account or S3 Bucket Quota checks are not available, the quotas are disabled."
+      summary: "Utilization metrics service not available"
+
+  # As a platform admin I want to be alerted (critical) when quotas were not honored due to metrics
+  # being unavailable
+  - alert: QuotaUnavailable
+    expr: |
+      sum(rate(s3_cloudserver_quota_request_with_metrics_unavailable{namespace="${namespace}",service="${service}"}[2m]))
```
francoisferrand: may be more relevant with `increase()`, to limit rounding of values close to 0...

monitoring/dashboard.py
```diff
@@ -366,6 +366,32 @@ def average_latency_target(title, action="", by=""):
...
 
 
+def average_quota_latency_target(title, code="", type="", by=""):
+    # type: (str, str, str, str) -> Target
+    extra = ', code=' + code if code else ""
+    extra += ', type=' + type if type else ""
+    by = " by (" + by + ")" if by else ""
+    return Target(
+        expr="\n".join([
+            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_sum{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
```
francoisferrand:
```suggestion
            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_sum{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -366,6 +366,32 @@ def average_latency_target(title, action="", by=""):
...
+            "   /",
+            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
```
francoisferrand:
```suggestion
            'sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
```

monitoring/dashboard.py
```diff
@@ -366,6 +366,32 @@ def average_latency_target(title, action="", by=""):
...
+        legendFormat=title,
+    )
+
+
+def average_quota_retrieval_latency(code="", by=""):
+    # type: (str, str) -> Target
+    extra = ', code=' + code if code else ""
+    by = " by (" + by + ")" if by else ""
+    return "\n".join([
+        'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_sum{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
```
francoisferrand:
```suggestion
        'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_sum{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -366,6 +366,32 @@ def average_latency_target(title, action="", by=""):
...
+        "   /",
+        'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_count{namespace="${namespace}", job=~"$job"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
```
francoisferrand:
```suggestion
        'sum(rate(s3_cloudserver_quota_metrics_retrieval_duration_seconds_count{namespace="${namespace}", job="${job}"' + extra + "}[$__rate_interval]))" + by,  # noqa: E501,
```

monitoring/dashboard.py
```diff
@@ -525,6 +551,189 @@ def top10_errors_by_bucket(title, code):
...
+
+quotaHealth = TimeSeries(
+    title="Quota service uptime",
+    legendDisplayMode="hidden",
+    dataSource="${DS_PROMETHEUS}",
+    lineInterpolation="stepAfter",
+    fillOpacity=30,
+    unit=UNITS.PERCENT_FORMAT,
+    targets=[Target(
+        expr='avg(avg_over_time(s3_cloudserver_quota_utilization_service_available{namespace="${namespace}",job=~"$job"}[$__rate_interval])) * 100',  # noqa: E501
```
francoisferrand:
```suggestion
        expr='avg(avg_over_time(s3_cloudserver_quota_utilization_service_available{namespace="${namespace}",job="${job}"}[$__rate_interval])) * 100',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +551,189 @@ def top10_errors_by_bucket(title, code):
...
+
+
+quotaStatusCode = TimeSeries(
+    title="Quota evaluation status code over time",
+    dataSource="${DS_PROMETHEUS}",
+    fillOpacity=30,
+    lineInterpolation="smooth",
+    unit=UNITS.OPS_PER_SEC,
+    targets=[Target(
+        expr='sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"2..", job=~"$job"}[$__rate_interval]))',  # noqa: E501
```
francoisferrand:
```suggestion
        expr='sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code=~"2..", job="${job}"}[$__rate_interval]))',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +551,189 @@ def top10_errors_by_bucket(title, code):
...
+        legendFormat="Success",
+    ), Target(
+        expr='sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code="429", job=~"$job"}[$__rate_interval]))',  # noqa: E501
```
francoisferrand:
```suggestion
        expr='sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", code="429", job="${job}"}[$__rate_interval]))',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +551,189 @@ def top10_errors_by_bucket(title, code):
...
+    title="Quota evaluaton rate per S3 action",
+    dataSource="${DS_PROMETHEUS}",
+    legendDisplayMode="table",
+    legendPlacement="right",
+    legendValues=["min", "mean", "max"],
+    lineInterpolation="smooth",
+    unit=UNITS.OPS_PER_SEC,
+    targets=[
+        Target(
+            expr='sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", job=~"$job"}[$__rate_interval])) by(action)',  # noqa: E501
```
francoisferrand:
```suggestion
            expr='sum(rate(s3_cloudserver_quota_evaluation_duration_seconds_count{namespace="${namespace}", job="${job}"}[$__rate_interval])) by(action)',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +551,189 @@ def top10_errors_by_bucket(title, code):
...
+averageQuotaDuration = Heatmap(
+    title="Quota evaluation duration",
+    dataSource="${DS_PROMETHEUS}",
+    dataFormat="tsbuckets",
+    maxDataPoints=25,
+    tooltip=Tooltip(show=True, showHistogram=True),
+    yAxis=YAxis(format=UNITS.SECONDS),
+    color=HeatmapColor(mode="opacity"),
+    targets=[Target(
+        expr='sum by(le) (increase(s3_cloudserver_quota_evaluation_duration_seconds_bucket{namespace="${namespace}", job=~"$job"}[$__rate_interval]))',  # noqa: E501
```
francoisferrand:
```suggestion
        expr='sum by(le) (increase(s3_cloudserver_quota_evaluation_duration_seconds_bucket{namespace="${namespace}", job="${job}"}[$__rate_interval]))',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +551,189 @@ def top10_errors_by_bucket(title, code):
...
+
+operationsWithUnavailableMetrics = TimeSeries(
+    title="Operations with unavailable metrics",
+    dataSource="${DS_PROMETHEUS}",
+    fillOpacity=30,
+    lineInterpolation="smooth",
+    unit=UNITS.OPS_PER_SEC,
+    legendDisplayMode="hidden",
+    targets=[Target(
+        expr='sum(rate(s3_cloudserver_quota_request_with_metrics_unavailable{namespace="${namespace}", job=~"$job"}[$__rate_interval]))',  # noqa: E501
```
francoisferrand:
```suggestion
        expr='sum(rate(s3_cloudserver_quota_request_with_metrics_unavailable{namespace="${namespace}", job="${job}"}[$__rate_interval]))',  # noqa: E501
```

monitoring/dashboard.py
```diff
@@ -525,6 +551,189 @@ def top10_errors_by_bucket(title, code):
...
+            title="Account, exceeded",
+            type='"account"', code='"429"'),
+        average_quota_latency_target(
+            title="Account & bucket, success",
+            type='"bucket+account"',
+            code='~"2.."'),
+        average_quota_latency_target(
+            title="Account & bucket, exceeded",
+            type='"bucket+account"',
+            code='"429"'),
```
francoisferrand: can be deduplicated into just 2 target definitions (`code~"2.."` and `code="429"`), by using queries aggregated `by(type)` with a dynamic legend (`{{ type }}`)

ghost: Updated in https://github.com/scality/cloudserver/pull/5556/commits/293d3afeaaf88d8f489fe70ad4b0bf118c07c14c
I test it on the platform, here is the visual:

image

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -64,11 +65,124 @@ function isMetricStale(metric, resourceType, resourceName, action, inflight, log
...
+ * Evaluates quotas for a bucket and an account.
+ *
+ * @param {number} bucketQuota - The quota limit for the bucket.
+ * @param {number} accountQuota - The quota limit for the account.
+ * @param {object} bucket - The bucket object.
+ * @param {object} account - The account object.
+ * @param {number} inflight - The number of inflight requests.
+ * @param {number} inflightForCheck - The number of inflight requests for checking quotas.
+ * @param {string} action - The action being performed.
+ * @param {boolean} cleanUp - Flag indicating whether to ignore quotas.
```
francoisferrand: i don't think this flag is even necessary: we could just rely on the fact that inflight will be `< 0` in that case, and just ignore the quota check result: it will be the same number of request anyway...

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -64,11 +65,124 @@ function isMetricStale(metric, resourceType, resourceName, action, inflight, log
...
             inflight,
         });
+        monitoring.requestWithQuotaMetricsUnavailable.inc();
         return true;
     }
     return false;
 }
 
+/**
+ * Evaluates quotas for a bucket and an account.
```
francoisferrand: nit:
```suggestion
 * Evaluates quotas for a bucket and an account and update inflight count.
```

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -118,113 +247,51 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
...
+                    if (errorFromAPI) {
+                        return _evaluateQuotas(bucketQuota, accountQuota, bucket, account, _inflights,
+                            null, apiName, true, log, () => {
+                                monitorQuotaEvaluationDuration(apiMethod, type, code, quotaEvaluationDuration +
+                                    Number(process.hrtime.bigint() - quotaCleanUpStartTime));
+                                return _done();
+                            });
+                    }
+                    monitorQuotaEvaluationDuration(apiMethod, type, code, quotaEvaluationDuration);
                     return _done();
```
francoisferrand: use waterflow to simplify code:

```suggestion
                    async.waterflow(
                        cb => {
                            if (errorFromAPI) {
                                return _evaluateQuotas(bucketQuota,      accountQuota, bucket, account, _inflights,
                            null, apiName, true, log, cb);
                            }
                            return cb();
                        },
                        () => {
                            monitorQuotaEvaluationDuration(apiMethod, type, code, quotaEvaluationDuration +
                                Number(process.hrtime.bigint() - quotaCleanUpStartTime));
                            return _done();
                        }
                    )
```

lib/quotas/scuba/wrapper.js
```diff
@@ -60,7 +63,22 @@ class ScubaClientImpl extends ScubaClient {
...
+                    code: err.statusCode || 500,
+                    class: metricsClass,
+                }).observe(responseTimeInNs / 1e9);
+                return callback(err);
+            }
+            monitoring.utilizationMetricsRetrievalDuration.labels({
+                code: 200,
+                class: metricsClass,
+            }).observe(responseTimeInNs / 1e9);
+            return callback(null, data);
```
francoisferrand:
```suggestion
            monitoring.utilizationMetricsRetrievalDuration.labels({
                code: err ? (err.statusCode || 500) : 200,
                class: metricsClass,
            }).observe(responseTimeInNs / 1e9);
            return callback(err, data);
```

francoisferrand (approved):
LGTM, just minor cleanups

PR #5557, opened 2024-04-16
Migrate to ghcr

francoisferrand:
(already reviewed as https://github.com/scality/cloudserver/pull/5555)

- Migrate to ghcr
- Upgrade actions
- Build pykmip image
- Use official docker build steps

Issue: CLDSRV-524

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,24 +11,24 @@ on:
 
 jobs:
   build-federation-image:
-    uses: scality/workflows/.github/workflows/docker-build.yaml@v1
+    uses: scality/workflows/.github/workflows/docker-build.yaml@v2
     secrets: inherit
     with:
       push: true
-      registry: registry.scality.com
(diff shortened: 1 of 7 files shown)


nicolas2bert (approved):

.github/workflows/release.yaml
```diff
@@ -11,36 +11,59 @@ on:
...
+          password: ${{ github.token }}
+      - name: Build and push image
+        uses: docker/build-push-action@v5
+        with:
+          push: true
+          context: .
+          tags: |
+            ghcr.io/${{ github.repository }}:${{ github.event.inputs.tag }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
ghost, line 58: Do we want to have a scope here?

francoisferrand: not really: this is the "main" image, so we can use the default scope.
as long as no scope is shared between different images, there is no issue.

BourgoisMickael (approved):

PR #5566, opened 2024-05-10
Improvement/cldsrv 520 fix feredation image

ghost:
Fixed release for federation

### DIFF
--- a/images/svc-base/Dockerfile
+++ b/images/svc-base/Dockerfile
@@ -14,8 +14,10 @@ RUN rm -f ~/.gitconfig && \
     git config --global --add safe.directory . && \
     git lfs install && \
     GIT_LFS_SKIP_SMUDGE=1 && \
+    yarn global add typescript && \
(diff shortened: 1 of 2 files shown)


images/svc-base/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN rm -f ~/.gitconfig && \
     git config --global --add safe.directory . && \
     git lfs install && \
     GIT_LFS_SKIP_SMUDGE=1 && \
+    sudo npm install typescript -g && \
     yarn install --frozen-lockfile --production --network-concurrency 1 && \
     yarn cache clean --all
```
francoisferrand: should we not _remove_ typescript eventually?
(since the image is not multi-stage with a builder, need to cleanup to keep things small...)

ghost: Yes, added a step to clean the globally-installed typescript package

francoisferrand (approved):

benzekrimaha (approved):
LGTM

PR #5569, opened 2024-05-16
Bugfix/cldsrv 553

ghost:
Fixes a limitation during a corner case, when restoring an object within account/bucket with enabled exceeded quota:

* First, we reserve space when processing `RestoreObject`
* When the data is being actually restored, the cold backend will PUT the object tot he HOT location (with a specific header, `x-scal-s3-version-id`). But then tries to increase the inflights: in this case, if the quota is exceeded already, or the bytes are near to the limit, we fail, even if the space was reserved.

This PR introduces a way to detect DMF calls (the specific header) and, in this case, only check the quotas, without increasing the in-flights.

We compare the quota with the current bytes usage minus the reserved space. This ensure that any already-exceeded quota won't block the operation, as the space was reserved.

This solution is not optimal as it is expected that the quota will never exceed the limit by a lot: at the end, we might allow all objects to be restored. But this case can happen only if some in-flights were lost due to the limitations of the metrics update frequency.

A better solution to account for our limitation would use the total number of reserved space bytes, and perform the following operation:

```
utilizationMetrics - restoringBytes + currentObjectBytes <= quota
```

In this case, we can better account for the real reserved space: we keep track of the reserved bytes, and thus restores will not exceed space already used by other operations before.

This will be done in a second iteration: for now, we consider that the reserved space is reserved, and lost in-flights are part of the known limitations of the feature.

---

Alternatively, we can decide to not subtract the current bytes with the current object to restore' size: but in this case, any quota exceeded will deny the request.

### DIFF
--- a/lib/api/apiUtils/object/objectRestore.js
+++ b/lib/api/apiUtils/object/objectRestore.js
@@ -131,8 +131,8 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
                 const actions = Array.isArray(mdValueParams.requestType) ?
                     mdValueParams.requestType : [mdValueParams.requestType];
                 const bytes = processBytesToWrite(request.apiMethod, bucketMD, mdValueParams.versionId, 0, objectMD);
-                return validateQuotas(request, bucketMD, request.accountQuotas, actions, request.apiMethod, bytes, log,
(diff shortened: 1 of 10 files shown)


ghost:
See https://github.com/scality/cloudserver/pull/5570

PR #5570, opened 2024-05-16
Bugfix/cldsrv 533

ghost:
Fixes a limitation during a corner case, when restoring an object within account/bucket with enabled exceeded quota:

* First, we reserve space when processing `RestoreObject`
* When the data is being actually restored, the cold backend will PUT the object tot he HOT location (with a specific header, `x-scal-s3-version-id`). But then tries to increase the inflights: in this case, if the quota is exceeded already, or the bytes are near to the limit, we fail, even if the space was reserved.

This PR introduces a way to detect DMF calls (the specific header) and, in this case, only check the quotas, without increasing the in-flights.

### DIFF
--- a/lib/metadata/metadataUtils.js
+++ b/lib/metadata/metadataUtils.js
@@ -184,7 +184,7 @@ function validateBucket(bucket, params, log, actionImplicitDenies = {}) {
  * @return {undefined} - and call callback with params err, bucket md
  */
 function standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log, callback) {
-    const { authInfo, bucketName, objectKey, versionId, getDeleteMarker, request } = params;
+    const { authInfo, bucketName, objectKey, versionId, getDeleteMarker, request, withVersionId } = params;
     let requestType = params.requestType;
     if (!Array.isArray(requestType)) {
         requestType = [requestType];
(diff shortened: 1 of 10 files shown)


benzekrimaha (approved):
LGTM

lib/metadata/metadataUtils.js
```diff
@@ -242,13 +242,16 @@ function standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log,
             const needQuotaCheck = requestType => requestType.some(type => actionNeedQuotaCheck[type] ||
                 actionWithDataDeletion[type]);
             const checkQuota = params.checkQuota === undefined ? needQuotaCheck(requestType) : params.checkQuota;
-            if (!checkQuota) {
+            // withVersionId cover cases when an object is being restored with a specific version ID.
+            // In this case, the storage space was already accounted for when the RestoreObject API call
+            // was made, so we don't need to add any inflight, but quota must be evaluated.
+            if (!checkQuota && !withVersionId) {
```
francoisferrand, line 248: is this really needed?

francoisferrand: is this really needed?


```suggestion
            if (!checkQuota) {
```

ghost: Indeed, it had been removed (and squashed)

francoisferrand (approved):

PR #5574, opened 2024-05-22
[HF 7.70.21-8] CLDSRV-466: timestamps in stderr

tmacro:


index.js
```diff
@@ -1,3 +1,10 @@
 'use strict'; // eslint-disable-line strict
 
+/**
+ * Catch uncaught exceptions and add timestamp to aid debugging
+ */
+process.on('uncaughtException', err => {
+    process.stderr.write(`${new Date().toISOString()}: Uncaught exception: \n${err.stack}`);
```
BourgoisMickael, line 7: We don't use `werelogs` to format as JSON on stderr right ?

BourgoisMickael (approved):

jonathan-gramain (approved):

index.js
tmacro: yeah its just plain text.

PR #5580, opened 2024-06-13
Bugfix/cldsrv 541

ghost:
Currently:

* DeleteObjects API do not send any request context when authorizing the request, as we authorize per-object during the API evaluation.
* When authorizing objects, we use an internal API that only returns the authorization result (**not** any other information like the user profile or the quota).
* Eventually we end up with no quota returned at the beginning of the API, so the utilization metrics is not decreased as we detect no account quota.

What we want is to know the account quota, two solutions:
* We get the quota for each object we authorize in the API, but `CheckPolicies` only return an array, so any change will be breaking.
* We send the request contexts at the beginning of the API, get an implicit deny as we pass no resource, and get the quota information. This is the solution implemented here. Note that the authorization result in this case is ignored by the API, on purpose, so any decision will not impact the API, only the objects authz will.

Note: cannot be tested here, will be tested at the integration level.

### DIFF
--- a/lib/api/apiUtils/authorization/prepareRequestContexts.js
+++ b/lib/api/apiUtils/authorization/prepareRequestContexts.js
@@ -52,7 +52,7 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
             apiMethod, 's3');
     }
 
-    if (apiMethod === 'multiObjectDelete' || apiMethod === 'bucketPut') {
+    if (apiMethod === 'bucketPut') {
         return null;
     }
 
@@ -65,7 +65,17 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
 
     const requestContexts = [];
 
-    if (apiMethodAfterVersionCheck === 'objectCopy'
+    if (apiMethod === 'multiObjectDelete') {
+        // MultiObjectDelete does not require any authorization when evaluating
+        // the API. Instead, we authorize each object passed.
+        // But in order to get any relevant information from the authorization service
+        // for example, the account quota, we must send a request context object
+        // with no `specificResource`. We expect the result to be an implicit deny.
+        // In the API, we then ignore these authorization results, and we can use
+        // any information returned, e.g., the quota.
+        const requestContextMultiObjectDelete = generateRequestContext('objectDelete');
+          requestContexts.push(requestContextMultiObjectDelete);
+    } else if (apiMethodAfterVersionCheck === 'objectCopy'
         || apiMethodAfterVersionCheck === 'objectPutCopyPart') {
         const objectGetAction = sourceVersionId ? 'objectGetVersion' :
           'objectGet';
--- a/tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
+++ b/tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
@@ -15,13 +15,15 @@ const sourceObject = 'objectsource';
 const sourceVersionId = 'vid1';
 
 describe('prepareRequestContexts', () => {
-    it('should return null if multiObjectDelete method', () => {
+    it('should return s3:DeleteObject if multiObjectDelete method', () => {
         const apiMethod = 'multiObjectDelete';
         const request = makeRequest();
         const results = prepareRequestContexts(apiMethod, request, sourceBucket,
         sourceObject, sourceVersionId);
 
-        assert.strictEqual(results, null);
+        assert.strictEqual(results.length, 1);
+        const expectedAction = 's3:DeleteObject';
+        assert.strictEqual(results[0].getAction(), expectedAction);
     });
 
     it('should return s3:PutObjectVersion request context action for objectPut method with x-scal-s3-version-id' +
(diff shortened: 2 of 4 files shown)


lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -65,7 +65,17 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
...
-    if (apiMethodAfterVersionCheck === 'objectCopy'
+    if (apiMethod === 'multiObjectDelete') {
+        // MultiObjectDelete does not require any authorization when evaluating
+        // the API. Instead, we authorize each object passed.
+        // But in order to get any relevant information from the authorization service
+        // for example, the account quota, we must send a request context object
+        // with no `specificResource`. We expect the result to be an implicit deny.
+        // In the API, we then ignore these authorization results, and we can use
+        // any information returned, e.g., the quota.
+        const requestContextMultiObjectDelete = generateRequestContext('objectDelete');
```
francoisferrand, line 76: can we instead pass `multiObjectDelete` in the request context?
it will be easier to follow in lgos, and maybe eventually allow policy-based denial of multiObjectDelete...

lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -65,7 +65,17 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
...
     const requestContexts = [];
 
-    if (apiMethodAfterVersionCheck === 'objectCopy'
+    if (apiMethod === 'multiObjectDelete') {
+        // MultiObjectDelete does not require any authorization when evaluating
+        // the API. Instead, we authorize each object passed.
+        // But in order to get any relevant information from the authorization service
+        // for example, the account quota, we must send a request context object
+        // with no `specificResource`. We expect the result to be an implicit deny.
+        // In the API, we then ignore these authorization results, and we can use
```
francoisferrand, line 74: > In the API, we then ignore these authorization results

we should add a test (unless we already have it ?) to verify that the multiObjectDelete will indeed ignore the result from this "initial" request

tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -15,13 +15,15 @@ const sourceObject = 'objectsource';
 const sourceVersionId = 'vid1';
 
 describe('prepareRequestContexts', () => {
-    it('should return null if multiObjectDelete method', () => {
+    it('should s3:DeleteObject null if multiObjectDelete method', () => {
```
francoisferrand: typo?

```suggestion
    it('should return s3:DeleteObject if multiObjectDelete method', () => {
```

lib/api/apiUtils/authorization/prepareRequestContexts.js
ghost: I think we can pass it indeed, but there is no such permission, MultiObjectDelete has no dedicated s3 permission, it's s3:DeleteObject, so I fear that in this case, vault will always detect that we have not applicable policy...

ghost: So, update, we cannot use this action today, or we need to add a permission, otherwise, the machinery in Arsenal will fail to convert the apiName to an action, causing issues. Also, we do not want to introduce any non-standard permission anymore.

lib/api/apiUtils/authorization/prepareRequestContexts.js
ghost: We have tests but they are not very explicit about permission, so I added one: https://github.com/scality/cloudserver/pull/5580/commits/0117a5b0b4f6df19887b5957fff67b84943ad1ca

tmacro (approved):

lib/api/apiUtils/authorization/prepareRequestContexts.js
francoisferrand: just to be sure, you mention:
> We expect the result to be an implicit deny

do we "expect" based on the spec (i.e. can we "garantee" we will get this), or just because this is how our implementation currently behaves (and it may change with the code) ?

i.e. I get that we don't want to introduce extra custom permissions (and agree), but I wonder if it may not be more future-proof to introduce an explicit bypass path: as it may more future-proof to introduce a dedicated and explicit case instead of relying on some corner of the implementation which may change...

ghost: Sorry, it's not true that we always get an implicit deny, we can also have allow or explicit denies, but whatever the result is, we do not use it. We only use the authv4 internal API to get the information we need.

That is why we do not need to support a dedicated permission for the DeleteObjects API: in AWS, this API never returns AccessDenied, but a 200 response with an object list, and possibly AccessDenied errors, or the deletion result.

francoisferrand: actually, even if we ignore the result, I still have the same question: are we "garanteed" by the spec that calling the authv4 api will get us the information we need, or could it happen (with some implementation change, which may happen over time) that the information is not returned when we detect such kind of "malformed" or incomplete requests?

i.e. are we relying on a detail of the current implementation, or on a specified behavior of the contract/api ?

francoisferrand (approved):

PR #5598, opened 2024-06-24
Support reading archive details

francoisferrand:
- Expose transition-in-progress flag in user metadata: x-amz-meta-scal-s3-transition-in-progress
- HeadObject: expose archive info and other "internal" metadata, which may be used for testing or troubleshooting. This is protected by an extra permission `scality:GetObjectArchiveInfo`.

Issue: CLDSRV-545

### DIFF
--- a/lib/api/apiUtils/object/coldStorage.js
+++ b/lib/api/apiUtils/object/coldStorage.js
@@ -35,6 +35,39 @@ function getAmzRestoreResHeader(objMD) {
 }
 
 /**
+ * set archiveInfo headers to target header object
+ * @param {object} objMD - object metadata
+ * @returns {object} headers - target header object
+ */
+function setArchiveInfoHeaders(objMD) {
+    const headers = {};
+
+    if (objMD['x-amz-scal-transition-in-progress']) {
+        headers['x-amz-scal-transition-in-progress'] = true;
+        headers['x-amz-scal-transition-time'] = new Date(objMD['x-amz-scal-transition-time']).toUTCString();
+    }
+
+    if (objMD.archive) {
+        headers['x-amz-scal-archive-info'] = JSON.stringify(objMD.archive.archiveInfo);
+
+        if (objMD.archive.restoreRequestedAt) {
+            headers['x-amz-scal-restore-requested-at'] = new Date(objMD.archive.restoreRequestedAt).toUTCString();
+            headers['x-amz-scal-restore-requested-days'] = objMD.archive.restoreRequestedDays;
+        }
+
+        if (objMD.archive.restoreCompletedAt) {
+            headers['x-amz-scal-restore-completed-at'] = new Date(objMD.archive.restoreCompletedAt).toUTCString();
+            headers['x-amz-scal-restore-will-expire-at'] = new Date(objMD.archive.restoreWillExpireAt).toUTCString();
+        }
+    }
+
+    // Always get the "real" storage class (even when STANDARD) in this case
+    headers['x-amz-storage-class'] = objMD['x-amz-storage-class'] || objMD.dataStoreName;
+
+    return headers;
+}
+
+/**
  * Check if restore can be done.
  *
  * @param {ObjectMD} objectMD - object metadata
@@ -244,4 +277,5 @@ module.exports = {
     getAmzRestoreResHeader,
     validatePutVersionId,
     verifyColdObjectAvailable,
+    setArchiveInfoHeaders,
 };
--- a/lib/api/apiUtils/object/coldStorage.js
+++ b/lib/api/apiUtils/object/coldStorage.js
@@ -64,6 +64,9 @@ function setArchiveInfoHeaders(objMD) {
     // Always get the "real" storage class (even when STANDARD) in this case
     headers['x-amz-storage-class'] = objMD['x-amz-storage-class'] || objMD.dataStoreName;
 
+    // Get the owner-id
+    headers['x-amz-scal-owner-id'] = objMD['owner-id'];
+
     return headers;
 }
 
--- a/lib/api/objectHead.js
+++ b/lib/api/objectHead.js
@@ -16,6 +16,7 @@ const { locationConstraints } = config;
 const { standardMetadataValidateBucketAndObj } = require('../metadata/metadataUtils');
 const { maximumAllowedPartCount } = require('../../constants');
 const { setExpirationHeaders } = require('./apiUtils/object/expirationHeaders');
+const { setArchiveInfoHeaders } = require('./apiUtils/object/coldStorage');
 
 /**
  * HEAD Object - Same as Get Object but only respond with headers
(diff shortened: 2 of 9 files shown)


tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -96,6 +96,60 @@ describe('prepareRequestContexts', () => {
         assert.strictEqual(results[1].getAction(), expectedAction2);
     });
 
+    it('should return s3:HeadObject for headObject', () => {
```
ghost:
```suggestion
    it('should return s3:GetObject for headObject', () => {
```

tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -96,6 +96,60 @@ describe('prepareRequestContexts', () => {
...
+        const request = makeRequest({
+        });
+        const results = prepareRequestContexts(apiMethod, request, sourceBucket,
+            sourceObject, sourceVersionId);
+
+        assert.strictEqual(results.length, 1);
+        assert.strictEqual(results[0].getAction(), 's3:GetObject');
+    });
+
+    it('should return s3:HeadObject and s3:GetObjectVersion for headObject', () => {
```
ghost:
```suggestion
    it('should return s3:GetObject and s3:GetObjectVersion for headObject', () => {
```

tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -96,6 +96,60 @@ describe('prepareRequestContexts', () => {
...
+        });
+        const results = prepareRequestContexts(apiMethod, request, sourceBucket,
+            sourceObject, sourceVersionId);
+
+        assert.strictEqual(results.length, 2);
+        assert.strictEqual(results[0].getAction(), 's3:GetObject');
+        assert.strictEqual(results[1].getAction(), 's3:GetObjectVersion');
+    });
+
+    it('should return s3:HeadObject and scality:GetObjectArchiveInfo for headObject ' +
```
ghost:
```suggestion
    it('should return s3:GetObject and scality:GetObjectArchiveInfo for headObject ' +
```

tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -96,6 +96,60 @@ describe('prepareRequestContexts', () => {
...
+        });
+        const results = prepareRequestContexts(apiMethod, request, sourceBucket,
+            sourceObject, sourceVersionId);
+
+        assert.strictEqual(results.length, 2);
+        assert.strictEqual(results[0].getAction(), 's3:GetObject');
+        assert.strictEqual(results[1].getAction(), 'scality:GetObjectArchiveInfo');
+    });
+
+    it('should return s3:HeadObject, s3:GetObjectVersion and scality:GetObjectArchiveInfo ' +
```
ghost:
```suggestion
    it('should return s3:GetObject, s3:GetObjectVersion and scality:GetObjectArchiveInfo ' +
```

tests/unit/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -96,6 +96,60 @@ describe('prepareRequestContexts', () => {
...
+            'x-amz-version-id': '0987654323456789',
+            'x-amz-scal-archive-info': 'true',
+        });
+        const results = prepareRequestContexts(apiMethod, request, sourceBucket,
+            sourceObject, sourceVersionId);
+
+        assert.strictEqual(results.length, 3);
+        assert.strictEqual(results[0].getAction(), 's3:GetObject');
+        assert.strictEqual(results[1].getAction(), 's3:GetObjectVersion');
+        assert.strictEqual(results[2].getAction(), 'scality:GetObjectArchiveInfo');
```
ghost, line 150: Beware that in an Artesca context, the Storage Account Owner and Data Consumer won't have this permission. The first could create a policy allowing it though.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -34,6 +34,39 @@ function getAmzRestoreResHeader(objMD) {
     return undefined;
 }
 
+/**
+ * set archiveInfo headers to target header object
+ * @param {object} objMD - object metadata
+ * @returns {object} headers - target header object
+ */
+function setArchiveInfoHeaders(objMD) {
```
ghost, line 42: I see no unit testing of this function, can we test it?

lib/api/objectHead.js
ghost, line 1: Shall we also support this behavior for the GetObject API? Today we will rely on HeadObject, but in case we also want to access the data, we would need two API calls...

package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.12.0",
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.130",
+    "arsenal": "git+https://github.com/scality/arsenal#14a1f13ace63346c6e1c94ff3dee191d38df9b76",
```
ghost: Just a comment not to forget the bump before merging

lib/api/objectHead.js
francoisferrand: this is really a corner case, so probably not needed (can be added later anyway if needed)...

lib/api/apiUtils/object/coldStorage.js
francoisferrand: it is tested in `objectHead` tests, which ensures we retrieve the right fields if/only when requested

ghost (approved):
Will need a rebase for the package.json changes, i will do it

KillianG (approved):

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -34,6 +34,42 @@ function getAmzRestoreResHeader(objMD) {
...
+        }
+
+        if (objMD.archive.restoreCompletedAt) {
+            headers['x-amz-scal-restore-completed-at'] = new Date(objMD.archive.restoreCompletedAt).toUTCString();
+            headers['x-amz-scal-restore-will-expire-at'] = new Date(objMD.archive.restoreWillExpireAt).toUTCString();
+        }
+    }
+
+    // Always get the "real" storage class (even when STANDARD) in this case
+    headers['x-amz-storage-class'] = objMD['x-amz-storage-class'] || objMD.dataStoreName;
```
welansari, line 65: I think this won't work as expected as `x-amz-storage-class` is never empty. It's set to `STANDARD` when the object is hot. But in any case, using `x-amz-storage-class` all the time should work too as it's set to the cold location name when the object is transitioned and is kept that way even on restores.

ghost: I think - it won't hurt keeping a fallback in case of any issue (or different API path leading to an empty x-amz-storage-class somehow?)

welansari (approved):

lib/api/apiUtils/object/coldStorage.js
welansari: yes i agree

PR #5602, opened 2024-06-25
bf: CLDSRV-547 update redis config for utapi reindex

jonathan-gramain:
Update the redis configuration of utapi reindex to include a list of sentinels, rather than a single sentinel (previously set to "localhost" in Federation).

I took this opportunity to cleanup tech debt related to parsing redis configuration, using "joi" for validation instead and making it common across the three different places where redis config is parsed. Not doing so would have required yet another copy-paste of dumb and error-prone validation code. Added unit tests for the new validation.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -87,6 +87,47 @@ function parseSproxydConfig(configSproxyd) {
     return joi.attempt(configSproxyd, joiSchema, 'bad config');
 }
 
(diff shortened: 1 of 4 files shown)


jonathan-gramain:
ping

BourgoisMickael (approved):

nicolas2bert (approved):

PR #5603, opened 2024-06-26
CLDSRV-549 restore 'git.commit-sha' and 'git.repository' labels

jonathan-gramain:
Add back the 'git.commit-sha' and 'git.repository' labels to pushed images, which were not attached anymore after the change of registry.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -145,6 +145,9 @@ jobs:
           provenance: false
           tags: |
             ghcr.io/${{ github.repository }}:${{ github.sha }}
+          labels: |
+            git.repository=${{ github.repository }}
+            git.commit-sha=${{ github.sha }}
           cache-from: type=gha,scope=cloudserver
           cache-to: type=gha,mode=max,scope=cloudserver
       - name: Build and push pykmip image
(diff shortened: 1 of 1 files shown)


jonathan-gramain:
ping

.github/workflows/tests.yaml
```diff
@@ -145,6 +145,9 @@ jobs:
           provenance: false
           tags: |
             ghcr.io/${{ github.repository }}:${{ github.sha }}
+          labels: |
+            git.repository=${{ github.repository }}
+            git.commit-sha=${{ github.sha }}
```
francoisferrand, line 150: Do we need _these_ `git.repository` and `git.commit` labels? [i get that we want to have the values, the point is really about the labels and their name]

The "standard" for this kind of information is defined by OCI [1], would be `org.opencontainers.image.source`, and `org.opencontainers.image.revision` (plus `org.opencontainers.image.version` for releases).

Thes can be automatically computed with the [metadata action](https://github.com/docker/metadata-action), providing both `labels` and `annotations` (and tags, but this would require defining what kind of tagging schemes we want). I wanted to introduce this when doing the migration, but did not get the time to try and test...

[1] https://github.com/opencontainers/image-spec/blob/main/annotations.md

francoisferrand (approved):
No formal objection, but I would rather we go to the next level (e.g. use standard annotations/labels generated with the standard metadata-action)...

BourgoisMickael (approved):

.github/workflows/tests.yaml
jonathan-gramain: Good to know, I think at the time I chatted with @tcarmet and we just went for a pragmatic solution, I don't think we considered following a standard here.

I don't see a reason not to follow the standard eventually, although I may want to merge this one as is if only to minimize changes required and ease integration testing for the quite urgent work on S3C-6474.

In order to implement the standard properly, would it make sense to go back to using a custom workflow in https://github.com/scality/workflows, to avoid repeating the changes in every repository where we want it?

PR #5613, opened 2024-07-02
Improvement/cldsrv 548 flag update oplog

benzekrimaha:
This PR aims to introduce a flag that will be used by Arsenal to introduce a hack to write the MD twice in the oplog: one "deleted: true" copy of the previous MD, followed by the expected update with the new metadata. 

Issue: CLDSRV-548

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -193,6 +193,11 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         metadataStoreParams.oldReplayId = objMD.uploadId;
     }
 
+    if (objMD && !bucketMD.isVersioningEnabled() && objMD?.archive?.archiveInfo) {
+        metadataStoreParams.needOplogUpdate = true;
+        metadataStoreParams.originOp = 's3:ReplaceArchivedObject';
+    }
+
     /* eslint-disable camelcase */
     const dontSkipBackend = externalBackends;
     /* eslint-enable camelcase */
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zenko/cloudserver",
-  "version": "8.8.27",
+  "version": "8.8.28",
   "description": "Zenko CloudServer, an open-source Node.js implementation of a server handling the Amazon S3 protocol",
   "main": "index.js",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.12.0",
     "@hapi/joi": "^17.1.0",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.12.0",
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.130",
+    "arsenal": "git+https://github.com/scality/arsenal#1f20fe0fdbc0aa55ef71a4c870cd9230f75d494d",
```
benzekrimaha: The project version will be bumped once arsn released

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -193,6 +193,10 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         metadataStoreParams.oldReplayId = objMD.uploadId;
     }
 
+    if (objMD && !bucketMD.isVersioningEnabled() && objMD.archive && Object.keys(objMD.archive).length > 0) {
```
francoisferrand: * This handles the case of bucket without versioning ; we may however have the same issue in version suspended case ? @Kerkesni, is there maybe already a function to check if we are in these kind of condition?
* no reason to check Object.keys(objMD.archive) : we should simply check if the object has been written to "cold", i.e. `objMD?.archive?.archiveInfo` (and you must not check the content of `archiveInfo`, it is supposed to stay opaque: hence the type `any` in arsenal's `ObjectMDArchive`)

AFAIR, race conditions (e.g. object deleted/replaced/...) during the archival process should be handled already when finishing the transition : if the object does not exist (or etag does not match), backbeat will post a Delete/GC request back to tape I think...

francoisferrand (approved):

lib/api/apiUtils/object/createAndStoreObject.js
welansari: Looking at the function  `isVersioningEnabled`
```
    isVersioningEnabled() {
        const versioningConfig = this.getVersioningConfiguration();
        return versioningConfig ? versioningConfig.Status === 'Enabled' : false;
    }
```
This should handle the versioning suspended case also. (returns false when the bucket is non versioned and when it's in versioning suspended mode)

package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.12.0",
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.130",
+    "arsenal": "git+https://github.com/scality/arsenal#817bb836ec0e04fbf083b4c57e42e3986dcdc9bc",
```
ghost: Just a reminder to bump before merging

ghost (approved):

PR #5626, opened 2024-07-15
Fixup on deleteObject 

benzekrimaha:
Introduced by https://github.com/scality/cloudserver/pull/5580 we now do send a requestContext with no specific resource instead of "null", which results in a policy evaluation error. As we get an implicit deny for the requestType "objectDelete", causing the processed result to be false , thus sending an empty array of objects to vault , resulting in a deny even when the policy allows the action on specific objects.

Linked Issue : https://scality.atlassian.net/browse/CLDSRV-555

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -397,8 +397,9 @@ function multiObjectDelete(authInfo, request, log, callback) {
                 if (bucketShield(bucketMD, 'objectDelete')) {
                     return next(errors.NoSuchBucket);
                 }
-                if (!isBucketAuthorized(bucketMD, 'objectDelete', canonicalID, authInfo, log, request,
-                    request.actionImplicitDenies)) {
+                // The implicit deny flag is ignored in the DeleteObjects API, as authorization only
+                // affects the objects.
+                if (!isBucketAuthorized(bucketMD, 'objectDelete', canonicalID, authInfo, log, request)) {
                     log.trace("access denied due to bucket acl's");
                     // if access denied at the bucket level, no access for
                     // any of the objects so all results will be error results


anurag4DSB (commented):
We should make this change in development branches starting from 7.x (assuming this is present in 7.x as well)

benzekrimaha:
@anurag4DSB 7.x branches are not concerned by this fixup as the https://github.com/scality/cloudserver/pull/5580 was merged in 8.8 only

ghost:
@benzekrimaha we can indeed backport this fix and have it in 7.x branches, if we also pass the implicit flag in there, as we might end up with the same problem if we implement quotas with scuba at some point, even if the bug is not (yet) in 7.x

anurag4DSB (approved):
As this is not present in 7.x approving.
I will let you decide if you want this in 7.x or not (following @williamlardier's comment)

benzekrimaha:
PR will be merged starting 7.10

benzekrimaha:
@bert-e create_pull_requests

lib/api/multiObjectDelete.js
```diff
@@ -398,7 +398,7 @@ function multiObjectDelete(authInfo, request, log, callback) {
                     return next(errors.NoSuchBucket);
                 }
                 if (!isBucketAuthorized(bucketMD, 'objectDelete', canonicalID, authInfo, log, request,
-                    request.actionImplicitDenies)) {
+                    false)) {
```
ghost: IIRC we can also just remove the argument, and have a small comment to explain why we don't pass it:

```suggestion
                // The implicit deny flag is ignored in the DeleteObjects API, as authorization only
                // affects the objects.
                if (!isBucketAuthorized(bucketMD, 'objectDelete', canonicalID, authInfo, log, request) {
```

ghost (approved):

benzekrimaha:
@bert-e reset

PR #5666, opened 2024-08-23
POST AuthV4

ghost:
Bump vaultclient, this will use POST by default when issuing AuthV4 internal requests (verifySignatureV4) with a security token (unbounded) or request contexts (same, but here it authorizes as well).

Fixes issues when any of those is too long for the backend to handle (as nodejs imposes max lengths for the http GET requests).

-> No new branch for now, as there is no breaking change in this service, but only vaultclient...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zenko/cloudserver",
-  "version": "8.8.30",
+  "version": "8.8.31",
   "description": "Zenko CloudServer, an open-source Node.js implementation of a server handling the Amazon S3 protocol",
   "main": "index.js",
   "engines": {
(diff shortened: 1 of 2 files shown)


welansari (approved):

package.json
```diff
@@ -47,7 +47,7 @@
     "utf-8-validate": "^5.0.8",
     "utf8": "~2.1.1",
     "uuid": "^8.3.2",
-    "vaultclient": "scality/vaultclient#8.3.20",
+    "vaultclient": "scality/vaultclient#f53bc147c433212f2c2a5ea1d4b8d01668b7f4ef",
```
KillianG: Dont forget to change that

KillianG (approved):

PR #5678, opened 2024-09-06
Fix handling of archived object replacement

francoisferrand:
- The `delete` oplog entry must not be generated when putObject is used
to restore the object: otherwise a GC request will be sent to cold
backend, and the object will never be expired.
- The `originOp` of the (new) metadata should not be affected:
"s3:ReplaceArchivedObject" should only be set on the extra `delete` op
in the oplog.

Issue: CLDSRV-560

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -193,11 +193,6 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         metadataStoreParams.oldReplayId = objMD.uploadId;
     }
 
-    if (objMD && !bucketMD.isVersioningEnabled() && objMD?.archive?.archiveInfo) {
-        metadataStoreParams.needOplogUpdate = true;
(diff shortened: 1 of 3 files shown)


lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -273,6 +268,9 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             if (isPutVersion) {
                 const options = overwritingVersioning(objMD, metadataStoreParams);
                 return process.nextTick(() => next(null, options, infoArr));
+            } else if (!bucketMD.isVersioningEnabled() && objMD?.archive?.archiveInfo) {
```
ghost, line 271: Maybe a good candidate for a unit test... It's a very sensitive function

ghost (approved):

benzekrimaha (approved):

lib/api/apiUtils/object/createAndStoreObject.js
francoisferrand: not sure unit test is the right place, but definitely a gap here → I created CLDSRV-561

PR #5681, opened 2024-09-19
Bump arsenal 8.1.137

francoisferrand:
This fixes metrics report, which fails while trying to process
`system.views` as a regular "bucket" collection.

Issue: CLDSRV-562


ghost (approved):

ghost:
For the CI error, I think we should pin the scubaclient version we want to use (or fix the introduced issue...)

benzekrimaha (approved):

PR #5683, opened 2024-10-01
Keep original MPU object etag after restore

welansari:
When restoring an MPU, the object data might be PUT with a different number of parts, which in turn changes the etag of the object as it is calculated based on the etag of each uploaded part.

We now keep the same etag in case of a restore to avoid breaking the expiration of the restored object as it checks the etag of the object before expiting.

The new etag is kept in the metadata of the object and can be viewed with a head-object operation when passing the `x-amz-scal-archive-info` header.

The new etag field is added to the `x-amz-restore` section of the metadata, so it will be removed when the object expires.

Issue: CLDSRV-564

### DIFF
--- a/lib/api/apiUtils/object/coldStorage.js
+++ b/lib/api/apiUtils/object/coldStorage.js
@@ -58,6 +58,7 @@ function setArchiveInfoHeaders(objMD) {
         if (objMD.archive.restoreCompletedAt) {
             headers['x-amz-scal-restore-completed-at'] = new Date(objMD.archive.restoreCompletedAt).toUTCString();
             headers['x-amz-scal-restore-will-expire-at'] = new Date(objMD.archive.restoreWillExpireAt).toUTCString();
+            headers['x-amz-scal-restore-etag'] = objMD['x-amz-restore']?.['content-md5'];
         }
     }
 
@@ -190,6 +191,7 @@ function _updateObjectExpirationDate(objectMD, log) {
         objectMD['x-amz-restore'] = {
             'ongoing-request': false,
             'expiry-date': expiryDate,
+            'content-md5': objectMD['x-amz-restore']?.['content-md5'],
         };
         /* eslint-enable no-param-reassign */
     }
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -518,6 +518,14 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     // set correct originOp
     metadataStoreParams.originOp = 's3:ObjectRestore:Completed';
 
+    // When restoring an MPU with different number of parts the etag changes
+    // as it's calculated based on the etags of the parts. We keep the original
+    // etag to allow proper expiration of the restored object.
+    if (metadataStoreParams.contentMD5 !== objMD['content-md5']) {
+        metadataStoreParams.restoredEtag = metadataStoreParams.contentMD5;
+        metadataStoreParams.contentMD5 = objMD['content-md5'];
+    }
+
     // update restore
     const days = objMD.archive?.restoreRequestedDays;
     const now = Date.now();
--- a/lib/services.js
+++ b/lib/services.js
@@ -109,7 +109,8 @@ const services = {
             tagging, taggingCopy, replicationInfo, defaultRetention,
             dataStoreName, creationTime, retentionMode, retentionDate,
             legalHold, originOp, updateMicroVersionId, archive, oldReplayId,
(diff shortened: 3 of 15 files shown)


lib/api/apiUtils/object/versioning.js
```diff
@@ -518,6 +518,14 @@ function overwritingVersioning(objMD, metadataStoreParams) {
     // set correct originOp
     metadataStoreParams.originOp = 's3:ObjectRestore:Completed';
 
+    // When restoring an MPU with different number of parts the etag changes
+    // as it's calculated based on the etags of the parts. We keep the original
+    // etag to allow proper exxpiration of the restored object.
```
KillianG: nit 
```suggestion
    // etag to allow proper expiration of the restored object.
```

package.json
```diff
@@ -41,7 +41,7 @@
     "npm-run-all": "~4.1.5",
     "prom-client": "14.2.0",
     "request": "^2.81.0",
-    "scubaclient": "git+https://github.com/scality/scubaclient.git",
+    "scubaclient": "git+https://github.com/scality/scubaclient.git#fb7375a9298bda7df0e9f9ed81d7fc5b363590a9",
```
KillianG, line 44: To be changed to proper version tag before merge I guess ?

KillianG (approved):

package.json
ghost: We do not have any released tag yet, for scubaclient

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -58,6 +58,7 @@ function setArchiveInfoHeaders(objMD) {
         if (objMD.archive.restoreCompletedAt) {
             headers['x-amz-scal-restore-completed-at'] = new Date(objMD.archive.restoreCompletedAt).toUTCString();
             headers['x-amz-scal-restore-will-expire-at'] = new Date(objMD.archive.restoreWillExpireAt).toUTCString();
+            headers['x-amz-scal-restore-etag'] = objMD['x-amz-restore'].etag;
```
ghost: I think in this code path that the header x-amz-restore might not always be set, can we ensure it exists before access this property (same applies to other relevant places)?

For example in the build I see this, from the quota tests:
```
{"name":"S3","time":1727859015536,"error":"Cannot read properties of undefined (reading 'etag')","stack":"TypeError: Cannot read properties of undefined (reading 'etag')\n    at _updateObjectExpirationDate (/usr/src/app/lib/api/apiUtils/object/coldStorage.js:194:47)\n    at Object.startRestore (/usr/src/app/lib/api/apiUtils/object/coldStorage.js:257:37)\n    at startRestore (/usr/src/app/lib/api/apiUtils/object/objectRestore.js:121:36)\n    at nextTask (/usr/src/app/node_modules/async/dist/async.js:5297:14)\n    at next (/usr/src/app/node_modules/async/dist/async.js:5304:9)\n    at /usr/src/app/node_modules/async/dist/async.js:906:16\n    at /usr/src/app/lib/api/apiUtils/object/objectRestore.js:116:28\n    at /usr/src/app/node_modules/arsenal/build/lib/s3middleware/objectRestore.js:97:16\n    at Generator.next (<anonymous>)\n    at fulfilled (/usr/src/app/node_modules/arsenal/build/lib/s3middleware/objectRestore.js:5:58)","level":"fatal","message":"caught error","hostname":"fv-az661-183","pid":68}
```

ghost (approved):

lib/services.js
```diff
@@ -200,6 +201,7 @@ const services = {
             md.setAmzRestore({
                 'ongoing-request': false,
                 'expiry-date': archive.restoreWillExpireAt,
+                'etag': restoredEtag,
```
francoisferrand: Should this be named 'content-md5', for consistency?

(i.e. always use `content-md5` in ObjectMD, and `etag` in API)

lib/services.js
```diff
@@ -200,6 +201,7 @@ const services = {
             md.setAmzRestore({
                 'ongoing-request': false,
                 'expiry-date': archive.restoreWillExpireAt,
+                'content-md5': restoredEtag,
```
francoisferrand, line 204: will this field be "discarded" automatically when `restoredEtag` is not provided (e.g. when it matches)?
or should we always set `restoredEtag` ?

francoisferrand (approved):

lib/services.js
welansari: It gets discarded from the metadata when `restoredEtag`  is not provided.

PR #5690, opened 2024-10-22
CLDSRV-569 isAccountEncryptionEnabled removed after deleteBucketEncry…

nicolas2bert:
Context:
When a bucket is deleted, we don't remove the encryption information entirely; instead, we hide it using the mandatory property. This allows the default encryption master key to be reused.

Issue:
After calling S3.deleteBucketEncryption, the isAccountEncryptionEnabled flag is being removed. This flag is crucial, as it prevents the master encryption key from being deleted during a bucket deletion when the key is set at the account level.

### DIFF
--- a/lib/api/bucketDeleteEncryption.js
+++ b/lib/api/bucketDeleteEncryption.js
@@ -42,6 +42,11 @@ function bucketDeleteEncryption(authInfo, request, log, callback) {
                 configuredMasterKeyId: sseConfig.configuredMasterKeyId,
             };
 
+            const { isAccountEncryptionEnabled } = sseConfig;
+            if (isAccountEncryptionEnabled) {
+                updatedConfig.isAccountEncryptionEnabled = isAccountEncryptionEnabled;
+            }
+
             bucket.setServerSideEncryption(updatedConfig);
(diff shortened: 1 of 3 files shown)


nicolas2bert:
ping

lib/api/bucketDeleteEncryption.js
```diff
@@ -42,6 +42,11 @@ function bucketDeleteEncryption(authInfo, request, log, callback) {
                 configuredMasterKeyId: sseConfig.configuredMasterKeyId,
```
anurag4DSB, line 42: can we directly add it here?
```
isAccountEncryptionEnabled: sseConfig.isAccountEncryptionEnabled,
```

dvasilas (approved):

lib/api/bucketDeleteEncryption.js
nicolas2bert: It is just to avoid storing the property when it is not needed.

anurag4DSB: Okay, I am good if its an intentional choice.

anurag4DSB (approved):

PR #5691, opened 2024-10-29
Fix complete MPU with missing parts

KillianG:
In case of a complete MPU error after the storage of the object MD, the abort must delete object MDs and object Data, some part may have been deleted already in the MPU metadata, hence we need to verify in the object MD for the parts it's linked to, to avoid creating orphans

Issue: CLDSRV-570

### DIFF
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -18,6 +18,8 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
         preciseRequestType: request.apiMethods || 'multipartDelete',
         request,
     };
+
+    log.debug('processing request', { method: 'abortMultipartUpload' });
     // For validating the request at the destinationBucket level
     // params are the same as validating at the MPU level
     // but the requestType is the more general 'objectDelete'
@@ -30,6 +32,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             standardMetadataValidateBucketAndObj(metadataValParams, authzIdentityResult, log,
                 (err, destinationBucket, objectMD) => {
                     if (err) {
+                        log.error('error validating request', { error: err });
                         return next(err, destinationBucket);
                     }
                     if (destinationBucket.policies) {
@@ -50,6 +53,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             services.metadataValidateMultipart(metadataValParams,
                 (err, mpuBucket, mpuOverviewObj) => {
                     if (err) {
+                        log.error('error validating multipart', { error: err });
                         return next(err, destBucket);
                     }
                     return next(err, mpuBucket, mpuOverviewObj, destBucket, objectMD);
@@ -67,6 +71,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 // eslint-disable-next-line no-param-reassign
                 request.actionImplicitDenies = originalIdentityAuthzResults;
                 if (err) {
+                    log.error('error aborting MPU', { error: err });
                     return next(err, destBucket);
                 }
                 // for Azure and GCP we do not need to delete data
@@ -79,6 +84,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             services.getMPUparts(mpuBucket.getName(), uploadId, log,
                 (err, result) => {
                     if (err) {
+                        log.error('error getting parts', { error: err });
                         return next(err, destBucket);
                     }
                     const storedParts = result.Contents;
@@ -115,7 +121,11 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             // In case there has been an error during cleanup after a complete MPU
             // (e.g. failure to delete MPU MD in shadow bucket),
             // we need to ensure that the MPU metadata is deleted.
-
+            log.info('Object has existing metadata, deleting them', {
+                method: 'abortMultipartUpload',
+                bucketName,
+                objectKey,
+            });
             return metadata.deleteObjectMD(bucketName, objectKey, undefined, log, err => {
                 if (err) {
                     log.error('error deleting object metadata', { error: err });
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -104,7 +104,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 bucketName,
                 objectKey,
                 uploadId,
-                versionId: objectMD.versionId
+                versionId: objectMD.versionId,
             });
             return metadata.deleteObjectMD(bucketName, objectKey, { versionId: objectMD.versionId }, log, err => {
                 if (err) {
@@ -120,16 +120,15 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             }
             // The locations were sent to metadata as an array
             // under partLocations.  Pull the partLocations.
-            let locations = storedParts.flatMap(item => item.value.partLocations);
+            const locations = storedParts.flatMap(item => item.value.partLocations);
             if (locations.length === 0) {
                 return next(null, mpuBucket, storedParts, destBucket);
             }
 
-            const existingLocations = new Set(locations.map(loc => loc.key));
-
             if (objectMD?.location) {
+                const existingLocations = new Set(locations.map(loc => loc.key));
                 const remainingObjectLocations = objectMD.location.filter(loc => !existingLocations.has(loc.key));
-                locations = [...locations, ...remainingObjectLocations];
+                locations.push(...remainingObjectLocations);
             }
 
             return async.eachLimit(locations, 5, (loc, cb) => {
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -128,7 +128,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 uploadId,
                 versionId: objectMD.versionId
             });
-            return metadata.deleteObjectMD(bucketName, objectKey, {versionId: objectMD.versionId}, log, err => {
+            return metadata.deleteObjectMD(bucketName, objectKey, { versionId: objectMD.versionId }, log, err => {
                 if (err) {
                     log.error('error deleting object metadata', { error: err });
                 }
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -120,17 +120,16 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             }
             // The locations were sent to metadata as an array
             // under partLocations.  Pull the partLocations.
-            let locations = storedParts.map(item => item.value.partLocations);
+            let locations = storedParts.flatMap(item => item.value.partLocations);
             if (locations.length === 0) {
                 return next(null, mpuBucket, storedParts, destBucket);
             }
-            // flatten the array
-            locations = [].concat(...locations);
+
+            const existingLocations = new Set(locations.map(loc => loc.key));
 
             if (objectMD?.location) {
-                const objectLocationLeft = objectMD.location.filter(loc =>
-                    !locations.some(existingLoc => existingLoc.key === loc.key));
-                locations = locations.concat(objectLocationLeft);
+                const remainingObjectLocations = objectMD.location.filter(loc => !existingLocations.has(loc.key));
+                locations = [...locations, ...remainingObjectLocations];
             }
 
             return async.eachLimit(locations, 5, (loc, cb) => {
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -125,8 +125,10 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 method: 'abortMultipartUpload',
                 bucketName,
                 objectKey,
+                uploadId,
+                versionId: objectMD.versionId
             });
-            return metadata.deleteObjectMD(bucketName, objectKey, undefined, log, err => {
+            return metadata.deleteObjectMD(bucketName, objectKey, {versionId: objectMD.versionId}, log, err => {
                 if (err) {
                     log.error('error deleting object metadata', { error: err });
                 }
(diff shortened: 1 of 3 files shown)


lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -105,9 +106,39 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+            return async.eachLimit(objectLocationLeft, 5, (loc, cb) => {
+                data.delete(loc, log, err => {
+                    if (err) {
+                        log.fatal('delete object data failed', { err });
+                    }
+                    cb();
+                });
+            }, () => next(null, mpuBucket, storedParts, destBucket, objectMD));
+        },
+        function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objectMD, next) {
```
jonathan-gramain: If I'm not mistaken I think it's better to delete the metadata first, to avoid leaving the object in an inconsistent state (i.e. unreadable), especially if there are failures.

KillianG: correct me if I'm wrong, but if we crash for some reason after deleting the metadata first, we will end up with orphans never recoverable nor deletable.. ?

ghost: From what I understand, usually we prefer having orphans in the storage layer rather than having an orphan in the metadata (orphan in the sense that we cannot do anything with it, and the data is not accessible). Both options are not perfect, but from the client point of view it allows to have a clean state. Nowadays we run tools to cleanup such orphans, but later we may be smarter and queue these operations internally...

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -105,9 +112,47 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                     }
                     cb();
                 });
+            }, () => next(null, mpuBucket, storedParts, destBucket, objectMD));
+        },
+        function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objectMD, next) {
+            if (!objectMD || metadataValMPUparams.uploadId !== objectMD.uploadId) {
```
jonathan-gramain, line 96: I don't think it's a huge issue and not sure we can do much better, but we have to keep in mind that in case the bucket is non-versioned, we may end up deleting a newer "version" of the object that overwrote an older "version", which ends with neither the old nor the new state (no object at all, where there were always an object before). However, the application is supposed to want to overwrite the old object in that case so shouldn't care about it, and the application aborts the new object which means it doesn't care about it either. But it could become an issue if e.g. the application needs to read the old object in order to write the new object on top, since with the deletion it will not be able to anymore.

I am thinking out loud here, it may be okay for the time being, but maybe longer term we need to have something more clever to get rid of orphaned MPU keys that doesn't delete parts that are already part of a valid object.

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -105,9 +112,47 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+                return next(null, mpuBucket, storedParts, destBucket, objectMD);
+            }
+            // In case there has been an error during cleanup after a complete MPU
+            // (e.g. failure to delete MPU MD in shadow bucket),
+            // we need to ensure that the MPU metadata is deleted.
+            log.info('Object has existing metadata, deleting them', {
+                method: 'abortMultipartUpload',
+                bucketName,
+                objectKey,
```
jonathan-gramain, line 105: It could be good to log the `versionId` and `uploadId` as well (but maybe it's already in the log context?)

jonathan-gramain (approved):

tests/unit/api/multipartUpload.js
```diff
@@ -1917,6 +1917,160 @@ describe('Multipart Upload API', () => {
...
+                    headers: { host: `${bucketName}.s3.amazonaws.com` },
+                    query: { uploadId: testUploadId },
+                    post: completeBody,
+                    actionImplicitDenies: false,
+                };
+                completeMultipartUpload(authInfo,
+                    completeRequest, log, err => {
+                        assert(err.is.InternalError);
+                        const MD = metadata.keyMaps.get(bucketName)
+                                                    .get(objectKey);
```
ghost: The indentation looks a bit strange, can't we have a single line for lines like this one?

```suggestion
                        const MD = metadata.keyMaps.get(bucketName).get(objectKey);
```

lib/api/apiUtils/object/abortMultipartUpload.js
ghost: We discussed this within the team and the problem wasn't clear to us, I think I got it, so i'm trying to rephrase- it, please let me know if I'm wrong:

For example, given a non-versioned bucket.

- We create an MPU for `objectA`, complete it: `objectA` exists with upload id `1`.
- Then we do another MPU for the same object (upload id `2`).
- Then the CompleteMPU fails after cloudserver replaced `objectA` with upload id `2`.
- Before the abort, we replace the object again (with a PUT or whatever).
- And then we call the abort, and we end up deleting the wrong object.

But in this very last case, that's what we are actually trying to avoid: if there is no object md, or if the upload id does not match, then we do nothing, so nothing is deleted.
Am I missing something?

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -105,9 +112,47 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+            });
+        },
+        function deleteObjectData(mpuBucket, storedParts, destBucket, objectMD, next) {
+            if (!objectMD?.location) {
+                return next(null, mpuBucket, storedParts, destBucket);
+            }
+            // Filtering parts that has already been delete by the previous step
+            let partLocations = storedParts.map(item => item.value.partLocations);
+            partLocations = [].concat(...partLocations);
+            partLocations = new Set(partLocations.map(loc => loc.key));
```
ghost: This looks under optimised: we map an array, then we deep copy it and recreate a set with another map. We should be able to merge it and create a set from a reduce of the partLocations keys

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -105,9 +112,47 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+            const objectLocationLeft = objectMD.location.filter(loc => !partLocations.has(loc.key));
+
+            return async.eachLimit(objectLocationLeft, 5, (loc, cb) => {
+                data.delete(loc, log, err => {
+                    if (err) {
+                        log.fatal('delete object data failed', { err });
+                    }
+                    cb();
```
ghost: We do not return any error in the callback yet log a fatal error: fatal is reserved to application crashes. Shall we log an error instead, and return the callback with the error?

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -105,9 +112,47 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+            }
+            // In case there has been an error during cleanup after a complete MPU
+            // (e.g. failure to delete MPU MD in shadow bucket),
+            // we need to ensure that the MPU metadata is deleted.
+            log.info('Object has existing metadata, deleting them', {
+                method: 'abortMultipartUpload',
+                bucketName,
+                objectKey,
+            });
+            return metadata.deleteObjectMD(bucketName, objectKey, undefined, log, err => {
```
francoisferrand: this should be made version-aware : i.e. in versioned bucket, need to specify the versionID of the object MD to delete...

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -105,9 +112,47 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+                if (err) {
+                    log.error('error deleting object metadata', { error: err });
+                }
+                return next(err, mpuBucket, storedParts, destBucket, objectMD);
+            });
+        },
+        function deleteObjectData(mpuBucket, storedParts, destBucket, objectMD, next) {
```
francoisferrand: similar to the "parts", shouldn't we first delete data, then metadata? otherwise we can get in a bad state, where we don't know of the parts anymore...

this has the added benefit of allowing to remove all parts in a single stage, just combine the MPU parts with the object parts (if any)...

(note: if we decide to first delete the ObjectMD, to avoid having an object pointing to delete data, we need to do it before removing the MPU parts data, as they may be refereced ; and this allows deleting in a single step...)

lib/api/apiUtils/object/abortMultipartUpload.js
francoisferrand: I think @jonathan-gramain is mentioning the case where the user would want to "update the data" (think download/edit/upload) : the first upload fails, but the Abort causes the data to be removed (neither initial not new version are available anymore).... so the application cannot do the update anymore...

Probably not a big issue for now (it is already like that), but ideally we find a solution in the future...

lib/api/apiUtils/object/abortMultipartUpload.js
KillianG: If I understand correctly, I think we prefer having "data orphans" instead of "metadata orphans" because in case of a "metadata orphans" user can think of a data loss because they can list and head the object but not get it..

This has been discussed here with jonathan
https://github.com/scality/cloudserver/pull/5691#discussion_r1841460336

lib/api/apiUtils/object/abortMultipartUpload.js
KillianG: Not sure we want to do that, return an error would stop the async.eachLimit, not sure we want that (to create more data orphans..)
This is done this way just above in `deleteData` function

lib/api/apiUtils/object/abortMultipartUpload.js
KillianG: Created a ticket to keep track of that issue but won't address it in this PR
https://scality.atlassian.net/browse/CLDSRV-579

lib/api/apiUtils/object/abortMultipartUpload.js
francoisferrand: if we prefer having data orphans, then we need to delete the (object) metadata first : i.e. delete object metadata before the part's data (i.e. before `deleteData()` function), since the parts are likely used in the object as well...

...and doing that we don't need to add a separate step to remove the remaining object's data: we simply need to remove both the parts' data and object data in the `deleteDate()` step.

lib/api/apiUtils/object/abortMultipartUpload.js
francoisferrand: unfortunately, I am not sure it is so simple, to cover all situations (versioned, non-versioned and version-suspended) : it probably works for both versioned and non-versioned cases, but i think it may break for the version-suspended case...

Also not sure if the field needs to be decoded/encoded here...

@Kerkesni Can you confirm? There helper functions to cover all cases, right?

tests/unit/api/multipartUpload.js
```diff
@@ -30,6 +30,7 @@ const metadataswitch = require('../metadataswitch');
 const { metadata } = storage.metadata.inMemory.metadata;
 const metadataBackend = storage.metadata.inMemory.metastore;
 const { ds } = storage.data.inMemory.datastore;
+const ms = storage.metadata.inMemory.metastore;
```
francoisferrand: already defined at line 31: `const metadataBackend = storage.metadata.inMemory.metastore;`

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -98,6 +126,13 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             }
             // flatten the array
             locations = [].concat(...locations);
+
+            if (objectMD?.location) {
+                const objectLocationLeft = objectMD.location.filter(loc =>
```
francoisferrand: nit: "left" looks like the directoin, maybe best to rename to something more intuitive, maybe `remainingObjectLocations`

more importantly, this can be quite CPU intensive for large objects (`n^2` complexity, with `n` up to 10000): need to make this more efficient...

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -66,26 +71,49 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+            }
+            // In case there has been an error during cleanup after a complete MPU
+            // (e.g. failure to delete MPU MD in shadow bucket),
+            // we need to ensure that the MPU metadata is deleted.
+            log.info('Object has existing metadata, deleting them', {
+                method: 'abortMultipartUpload',
+                bucketName,
+                objectKey,
+                uploadId,
+                versionId: objectMD.versionId
```
ghost:
```suggestion
                versionId: objectMD.versionId,
```

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -66,26 +71,49 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
         },
-        function deleteData(mpuBucket, storedParts, destBucket,
+        function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objectMD, skipDataDelete, next) {
+            if (!objectMD || metadataValMPUparams.uploadId !== objectMD.uploadId) {
+                return next(null, mpuBucket, storedParts, destBucket, objectMD, skipDataDelete);
+            }
+            // In case there has been an error during cleanup after a complete MPU
+            // (e.g. failure to delete MPU MD in shadow bucket),
+            // we need to ensure that the MPU metadata is deleted.
+            log.info('Object has existing metadata, deleting them', {
```
ghost, line 102: I would put the upload id (of the original object we want to delete, not the one from the request) and maybe more info about its MD in the log, if we can have the info, so we can track the problem more easily in case this is not exactly what we see at the customer side

ghost (approved):

lib/api/apiUtils/object/abortMultipartUpload.js
welansari: Maybe a more optimized approach for this would be to use a map ?

```
if (objectMD?.location) {
    const locationMap = new Map(locations.map(loc => [loc.key, true]));
    const objectLocationLeft = objectMD.location.filter(loc => !locationMap.has(loc.key));
    locations = locations.concat(objectLocationLeft);
}
```

francoisferrand: since we are building a list of locations, maybe we could directly use a `map` instead (and pass it to async.eachLimit directly, to process each "pair"?)

lib/api/apiUtils/object/abortMultipartUpload.js
welansari: @francoisferrand If i'm not mistaken this should work as is. internal metadata APIs rely on undecoded versionID.

For null versions:
- Objects created in a versioning suspended bucket do not have a versionID.
- In a versioning suspended bucket, putting a version on top of an already created versioned object creates a null version that has an internal versionID.

Both these cases should work with the current implementation.

lib/api/apiUtils/object/abortMultipartUpload.js
KillianG: Pushed a more optimized version here
https://github.com/scality/cloudserver/pull/5691/commits/e9ae9c22ec15e3ae2965eef3c3d1115d15a69c2a

FOr locations being a Set directly, I can do it, but as we'll need to map over it anyway for getting keys, we'll end up needing an array (as `map` function does not exist for Sets)

lib/api/apiUtils/object/abortMultipartUpload.js
KillianG: When we are in this log, both uploadId are the same
https://github.com/scality/cloudserver/pull/5691/files#diff-75b103d02d5ce4b1be1ff37a8a3058d683af6f676c0acef32e4fa1929d23ae7dR96

francoisferrand (approved):

PR #5694, opened 2024-11-06
CLDSRV-573: Fix crash because of prom-client timeout

BourgoisMickael:
Instead of crashing it will now stay alive, return a 500 with body `{"message":"Error: Operation timed out."}` and log

```json
{"name":"S3","clientIP":"::1","clientPort":53518,"httpMethod":"GET","httpURL":"/metrics","err":{"message":"Operation timed out."},"time":1730901149499,"req_id":"31b0c58cb14cad5c9583","elapsed_ms":5002.625237,"level":"warn","message":"monitoring error","hostname":"MDM-RING-46789-store-1","pid":115}
```

For other arsenal error  we will have the message field: `"err":{"MethodNotAllowed":true,"message":"The specified method is not allowed against this resource."}`

This changes will not go into ZENKO as they don't use the cluster module with prom-client

> [!IMPORTANT]
>
> This problem happens often in low resource platform (like CI).
> This will fix many flaky CI on Federation that fails because the step `Check if s3 Prometheus exporter is active` retry 3 times ith small delay, crashing s3 multiple time

### DIFF
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -24,7 +24,15 @@ async function routeHandler(req, res, cb) {
     if (req.method !== 'GET') {
         return cb(errors.BadRequest, []);
     }
-    const metrics = await aggregatorRegistry.clusterMetrics();
+    let metrics;
+    try {
+        // Catch timeout on IPC between worker and primary
+        // prom-client has a 5s hardcoded timeout
+        metrics = await aggregatorRegistry.clusterMetrics();
+    } catch (err) {
+        return cb(err, { message: err.toString() });
+    }
+
     const contentLen = Buffer.byteLength(metrics, 'utf8');
     res.writeHead(200, {
         'Content-Length': contentLen,
@@ -48,7 +56,11 @@ function monitoringHandler(clientIP, req, res, log) {
     function monitoringEndHandler(err, results) {
         writeResponse(res, err, results, error => {
             if (error) {
-                return log.end().warn('monitoring error', { err: error });
+                return log.end().warn('monitoring error', { err: {
+                    ...error,
+                    // For ArsenalError message is in description
+                    message: error.description || error.message,
+                } });
             }
             return log.end();
         });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3",
-  "version": "7.70.55",
+  "version": "7.70.56",
   "description": "S3 connector",
   "main": "index.js",
   "engines": {


BourgoisMickael:
ping

dvasilas (approved):
minor (very minor if this only happens in the CI):
`"err":{},"results":{"message":"Error: Operation timed out."}`

The `results` field is a bit unintuitive I think (I wouldn't grep for results in the logs).

Could we maybe have something like `"error":{"message":"Error: Operation timed out."}` ?

anurag4DSB:
> minor (very minor if this only happens in the CI): `"err":{},"results":{"message":"Error: Operation timed out."}`
> 
> The `results` field is a bit unintuitive I think (I wouldn't grep for results in the logs).
> 
> Could we maybe have something like `"error":{"message":"Error: Operation timed out."}` ?

I haven't seen the issue in production labs but I have in pre-production labs, one of them was FreePro.
Edit: It happens during CS startup time.

lib/utilities/monitoringHandler.js
```diff
@@ -48,7 +56,7 @@ function monitoringHandler(clientIP, req, res, log) {
     function monitoringEndHandler(err, results) {
         writeResponse(res, err, results, error => {
             if (error) {
-                return log.end().warn('monitoring error', { err: error });
+                return log.end().warn('monitoring error', { err: error, error: results });
             }
```
anurag4DSB: Right now, we're logging both err and error as the same object. It would be clearer to separate them if they represent different errors.

lib/utilities/monitoringHandler.js
```diff
@@ -48,7 +56,7 @@ function monitoringHandler(clientIP, req, res, log) {
     function monitoringEndHandler(err, results) {
```
anurag4DSB, line 56: Worth renaming this method to `handleMonitoringResponse`
Probably out of scope of this PR.

anurag4DSB (commented):
We need better errors in logs
I want to approve, but as the /approve command is there, it will be merged automatically without control.

lib/utilities/monitoringHandler.js
BourgoisMickael: This is to prevent changing the previous behavior, and it's the same error

- `err` is the error, so for arsenal error it will be like: `{"MethodNotAllowed": true}`, but for regular `Error` like the one catched here it will be `{}`
- `error: result` is the body passed to the response, so for arsenal error, any error managed by us it's `[]`, but specifically for the caugh prom error, I pass the message `err.toString()` as result in body so it will be `{"message":"Error: Operation timed out."}` because nodejs `Error` and `ArsenalError` logging don't mix well.

I added this so we can see if the metric error is something like `Operation timed out` or `write EPIPE` or any other cluster internal problem

anurag4DSB: okay, I understand that.
Can we make the code more readable, for example right now error is a key and a variable in the same scope?

I also prefer now having an `err:` and in the same line `error` unless I am missing something, there is no specific definition for err and error, so worth renaming them so CS or an admin can debug and understand.

BourgoisMickael: I could make something like `{message: err.toString(), ...err}`

For arsenal error it will look like `"err":{"MethodNotAllowed": true, "message": "This method is not allowed on this resource"}`

And for `Error` it will look like `"err":{"message":"Error: Operation timed out."}`

anurag4DSB: Thanks, that sounds good to me.
Considering this approving the PR.

anurag4DSB (approved):

PR #5697, opened 2024-11-12
CLDSRV-574 implement KMS health check

nicolas2bert:
**Context:** 
Cloudserver deep healthcheck is used in S3C by

S3 Frontend (Nginx) to manage the distribution of incoming S3 API requests based on backend server health. It sends a deep health check call every 2 seconds. Based on the deep health check results, Nginx dynamically adjusts the distribution of incoming S3 API requests. Healthy servers receive traffic, while unhealthy ones are temporarily removed from the pool to prevent failed requests.

Prometheus to monitor the health of bucketd. It sends an cloudserver deep health check call every 30 seconds. Prometheus requests blackbox that retrieves the cloudserver health, which has the bucketd health.

**Current behavior:**

Currently, KMS (KMIP/AWS KMS) is not part of the cloudserver deep health check. This prevents the cloudserver health check from failing if KMS is down, which is expected because only requests to encrypted buckets should fail. KMS being down should not prevent the server from starting up or run. (bugfix: https://scality.atlassian.net/browse/S3C-4833 )

**Expected behavior:**

We want KMS to be part of the cloudserver deep health check but still prevent the cloudserver health check from failing if KMS is down. Since only requests to encrypted buckets should fail when KMS is down, the server should continue to start up and run without issues.

**Why do we need to check for the KMS health?**

we want to be alerted if KMS is down to take appropriate action.

as a side effect, including KMS in the health check will help keep the connection alive. Some KMS providers require maintaining an active connection because, for example, when cloudserver connection to KMS remains idle for more than 24 hours, the internal Thales token expires. (ref: https://scality.atlassian.net/browse/S3C-8464 )

**NOTEs:**

Based on how frequently the cloudserver deep health check is performed, successful KMS requests should be cached with a cache expiration time set to less than 24 hours to keep the KMS connection alive.Caching will lower the number of requests made to the KMS service → less system resources used. Also, if KMS requests have costs, it will save some.

This approach is not a proper design for handling connection maintenance but rather a good side effect of integrating KMS into the health check. To handle connection “upkeep” properly, we should clearly separate both concerns: health checks vs connection maintenance.

### DIFF
--- a/lib/kms/Cache.js
+++ b/lib/kms/Cache.js
@@ -1,0 +1,65 @@
+class Cache {
+    constructor() {
+        this.lastChecked = null;
+        this.result = null;
+    }
+
+    /**
+     * Retrieves the cached result with the last checked timestamp.
+     * @returns {object|null} An object containing the result and lastChecked, or null if not set.
+     */
+    getResult() {
+        if (!this.result) {
+            return null;
+        }
+
+        return Object.assign({}, this.result, {
+            lastChecked: this.lastChecked ? new Date(this.lastChecked).toISOString() : null,
+        });
+    }
+
+    /**
+     * Retrieves the last checked timestamp.
+     * @returns {number|null} The timestamp of the last check or null if never checked.
+     */
+    getLastChecked() {
+        return this.lastChecked;
+    }
+
+    /**
+     * Updates the cache with a new result and timestamp.
+     * @param {object} result - The result to cache.
+     * @returns {undefined}
+     */
+    setResult(result) {
+        this.lastChecked = Date.now();
+        this.result = result;
+    }
+
+    /**
+     * Determines if the cache should be refreshed based on the last checked time.
+     * @param {number} duration - Duration in milliseconds for cache validity.
+     * @returns {boolean} true if the cache should be refreshed, else false.
+     */
+    shouldRefresh(duration = 1 * 60 * 60 * 1000) { // Default: 1 hour
+        if (!this.lastChecked) {
+            return true;
+        }
+
+        const now = Date.now();
+        const elapsed = now - this.lastChecked;
+        const jitter = Math.floor(Math.random() * 15 * 60 * 1000); // Up to 15 minutes
+        return elapsed > (duration - jitter);
+    }
+
+    /**
+     * Clears the cache.
+     * @returns {undefined}
+     */
+    clear() {
+        this.lastChecked = null;
+        this.result = null;
+    }
+}
+
+module.exports = Cache;
--- a/lib/kms/wrapper.js
+++ b/lib/kms/wrapper.js
@@ -8,6 +8,8 @@ const inMemory = require('./in_memory/backend').backend;
 const file = require('./file/backend');
 const KMIPClient = require('arsenal').network.kmipClient;
 const Common = require('./common');
+const Cache = require('./Cache');
+const cache = new Cache();
 let scalityKMS;
 let scalityKMSImpl;
 try {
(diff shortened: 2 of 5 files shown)


nicolas2bert:
ping

anurag4DSB (approved):
I like that its not gonna bring CS down :)

tests/unit/encryption/healthCheckCache.js
```diff
@@ -0,0 +1,125 @@
...
+
+describe('Cache Class', () => {
+    let cache;
+    let originalDateNow;
+    let originalMathRandom;
+
+    beforeEach(() => {
+        cache = new Cache();
+        originalDateNow = Date.now;
+        originalMathRandom = Math.random;
```
BourgoisMickael: Could use sinon stub

lib/kms/wrapper.js
```diff
@@ -355,26 +357,49 @@ class KMS {
...
+                    // still allow requests to this CloudServer instance. This way, an unavailable KMIP
+                    // host or a connection issue with KMIP will not prevent the normal operation of CloudServer.
+                    respBody[implName] = {
+                        code: 200,
+                        message: 'KMS health check failed',
+                        description: err.description,
+                    };
+                    logger.warn('KMS health check failed', { errorCode: err.code, error: err.description });
+                } else {
+                    respBody[implName] = {
```
BourgoisMickael: I wonder if it could be interesting to put the date of the last cached health check if cached for a long time, so you can see it directly in the response without needing debug logs

lib/kms/wrapper.js
```diff
@@ -355,26 +357,49 @@ class KMS {
...
+                        code: 200,
+                        message: 'OK',
+                    };
+                    logger.info('KMS health check succeeded', { respBody });
+                }
+                // Only update the cache on success
+                cache.setResult(respBody);
```
BourgoisMickael: It means if there is a failure, kms down or other problem, it's going to retry for every deepHealthcheck (every 2 seconds or 30 seconds) indefinitely.

And if the endpoint is slow there could potentially be multiple health check concurrently.

Maybe we could prevent this from happening, maybe it could be desynchronized from the incoming request, by having kms health check on interval to manage failure retry and the request retrieved only the cached value ?

BourgoisMickael (approved):

lib/kms/wrapper.js
nicolas2bert: It's a leftover comment from the initial implementation. After discussing with the product team, we decided to cache both successful and failed responses.

I just removed it. Thanks.

lib/kms/wrapper.js
nicolas2bert: Yes, agree. It's good to be transparent.

lib/kms/Cache.js
```diff
@@ -0,0 +1,56 @@
...
+        this.result = result;
+    }
+
+    /**
+     * Determines if the cache should be refreshed based on the last checked time.
+     * @param {number} duration - Duration in milliseconds for cache validity.
+     * @returns {boolean} true if the cache should be refreshed, else false.
+     */
+    shouldRefresh(duration = 1 * 60 * 60 * 1000) { // Default: 1 hours
+        if (!this.lastChecked) return true;
```
jonathan-gramain:
```suggestion
        if (!this.lastChecked) {
            return true;
        }
```

tests/unit/encryption/checkHealth.js
```diff
@@ -0,0 +1,91 @@
...
+        delete memBackend.healthcheck;
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    it('should return OK when kms backend does not have healthcheck method', (done) => {
+        kms.checkHealth(log, (err, result) => {
+            assert.strictEqual(err, null);
```
jonathan-gramain: I tend to prefer the more idiomatic `assert.ifError(err)` (but up to you)

jonathan-gramain (approved):

nicolas2bert:
@bert-e approve

nicolas2bert:
ping

nicolas2bert:
@bert-e reset

nicolas2bert:
ping

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e reset

PR #5699, opened 2024-11-15
Fix handling of archived object replacement

francoisferrand:
When an objects overwrites an object in a non-versioned bucket, need
to add special handling to ensure an oplog entry is generated about the
"removal" of object: so that Backbeat can detect it, and notify cold
storage backend.

This was fixed already for putObject in CLDSRV-560, but the MPU and
copyObject cases were not handled.

Issue: CLDSRV-577

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -268,7 +268,8 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             if (isPutVersion) {
                 const options = overwritingVersioning(objMD, metadataStoreParams);
                 return process.nextTick(() => next(null, options, infoArr));
-            } else if (!bucketMD.isVersioningEnabled() && objMD?.archive?.archiveInfo) {
+            }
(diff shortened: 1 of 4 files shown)


welansari (approved):

ghost (commented):
I understand the urgency, but we really should cover these changes with tests, unless we already have a follow up ticket

benzekrimaha (approved):

francoisferrand:
> I understand the urgency, but we really should cover these changes with tests, unless we already have a follow up ticket

no easy way to test this in Cloudserver (since the behavior is not 
"observable" at cloudserver level), so it is tested Zenko : https://scality.atlassian.net/browse/ZENKO-4928

ghost:
@francoisferrand I disagree that it's not easily testable: we can always add unit tests for that, no? We are already testing all these functions with mocks when needed... (e.g. completeMPU)

KillianG (approved):

francoisferrand:
> @francoisferrand I disagree that it's not easily testable: we can always add unit tests for that, no? We are already testing all these functions with mocks when needed... (e.g. completeMPU)

https://scality.atlassian.net/browse/CLDSRV-581 added to next sprint

ghost (approved):

PR #5700, opened 2024-11-15
CLDSRV-580 Custom KMS key id persisted after deleting bucket encrypti…

nicolas2bert:
After deleting a bucket's encryption configuration that was initially set with a custom AWS KMS master key ID, uploading an object with the aws:kms encryption algorithm header incorrectly uses the previously specified custom master key ID instead of the default one.

**Steps to reproduce**
- Create a bucket
- Set bucket encryption with custom KMS master key id
- Delete bucket encryption configuration
- Upload an object with aws:kms encryption header
- Head the object

**Current behavior**
The uploaded object is encrypted using the custom KMS master key id that was previously specified in the now-deleted bucket encryption configuration.

**Expected behavior**
After deleting the bucket encryption configuration, uploading an object with the aws:kms encryption algorithm header should use the default KMS master key id, not the previously specified custom one.

### DIFF
--- a/lib/api/bucketDeleteEncryption.js
+++ b/lib/api/bucketDeleteEncryption.js
@@ -30,24 +30,33 @@ function bucketDeleteEncryption(authInfo, request, log, callback) {
         (bucket, next) => checkExpectedBucketOwner(request.headers, bucket, log, err => next(err, bucket)),
         (bucket, next) => {
             const sseConfig = bucket.getServerSideEncryption();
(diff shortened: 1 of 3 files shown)


anurag4DSB (approved):
If relevant, we should merge this in development as well.

BourgoisMickael (approved):

nicolas2bert:
@bert-e approve

PR #5702, opened 2024-11-19
PutObjectRetention and DeleteObject authorization optimization

ghost:
Previously, we were dynamically calling the internal authorization API to check if the identity, when not an account, has the permission to bypass the governance lock configuration. This call sent request contexts with information we already have when authorizing the API, except the type of identity.

This commit introduces an optimization by including the Bypass action, when the header is set, directly when the API is authorized. If the identity is an account, the overhead is lower than doing another API call: no policy are evaluated.

Both `PutObjectRetention` and `DeleteObject` API calls are updated accordingly. Multi object delete can be improved but the complexity is higher, and is already subject for improvement in a dedicated ticket.

Unit tests updated for the list of actions to authorize, and logic is already covered by existing tests - tested on a real platform too with success.

### DIFF
--- a/lib/api/apiUtils/authorization/prepareRequestContexts.js
+++ b/lib/api/apiUtils/authorization/prepareRequestContexts.js
@@ -1,5 +1,6 @@
 const { policies } = require('arsenal');
 const { config } = require('../../../Config');
+const { hasGovernanceBypassHeader } = require('../object/objectLockHelpers');
 
 const { RequestContext, requestUtils } = policies;
 let apiMethodAfterVersionCheck;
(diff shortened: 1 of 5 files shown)


lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -193,13 +194,28 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
                 const putObjectLockRequestContext =
                   generateRequestContext('objectPutRetention');
                 requestContexts.push(putObjectLockRequestContext);
+                if (hasGovernanceBypassHeader(request.headers)) {
```
francoisferrand, line 197: in the previous code, there was an extra condition `isRequesterNonAccountUser(authInfo)` : which skips the user-governance-bypass right for "non-account user"...

is this not needed anymore? Will this skipping of the `bypassGovernanceRetention` check be done automatically by Vault when not an account?

ghost: Yes, as explained in the commit message, when we authorize against vault, and the identity is an account, the result is an Allow, and no policy is evaluated. So in all cases, we send this action to be authorized:
- If account: allowed, the overhead is that we added one more action to check, but we do that in a single api call instead of 2 sequential calls, so the point here is that the overhead is much lower in the general case, as none of our customers use accounts access keys directly.
- if not account: same as before, but in a single api call.

francoisferrand (approved):

welansari (approved):

PR #5707, opened 2024-11-26
CLDSRV-584 Limit backbeat API versioning check to replication operations

nicolas2bert:
**Context:**

In Artesca, when creating a location (whether used for replication or not), you specify both the endpoint and the bucket name where data will be stored. At this stage, Zenko checks that the bucket has versioning enabled.

In S3C, it is a bit different, a “replication location” consists of a list of S3C endpoints (replicationEndpoints), and the destination bucket is specified later in the API replication rule (with put-bucket-replication API). Currently, S3C does not check if the destination bucket has versioning enabled. This means data could be replicated from a versioned source bucket to a non-versioned or suspended destination bucket, which can cause issues. e.g. https://scality.atlassian.net/browse/S3C-821 

**Current solution:**

A validation step was added to the Backbeat API in cloudserver. This check rejects any requests made to buckets that either do not have versioning enabled or have versioning suspended. This check makes sure the Backbeat API only works with buckets that have versioning enabled. This prevents operations on destination buckets where versioning is disabled or suspended, hense maintaining proper version control.

On the source, we should be fine. Cloudserver prevents setting up replication on a bucket if its versioning is disabled or suspended, and it prevents changing a bucket’s versioning to suspended while replication is ongoing.

**Issue:**

The Backbeat API is also used for lifecycle, which works on both versioned and non-versioned buckets. The current versioning check affects the lifecycle operations, which should be allowed on non-versioned buckets.

**Proposed solution:**

Modify the Backbeat API to apply the bucket versioning check only to actions related to the replication destination buckets (such as PUT, POST, and DELETE requests).

Implement a new client header x-scal-versioning-required in the Backbeat client to make sure that replication targets only versioned buckets. When this header is set, Cloudserver will make sure that the destination bucket has versioning enabled, preventing replication to non-versioned buckets.

This change will ensure that replication only targets versioned buckets and allows lifecycle operations on non-versioned buckets.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -1289,7 +1289,10 @@ function routeBackbeat(clientIP, request, response, log) {
                     [request.query.operation](request, response, log, next);
             }
             const versioningConfig = bucketInfo.getVersioningConfiguration();
-            if (!versioningConfig || versioningConfig.Status !== 'Enabled') {
+            // The following makes sure that only replication destination-related operations
+            // target buckets with versioning enabled.
+            const isVersioningRequired = request.headers['x-scal-versioning-required'] === 'true';
+            if (isVersioningRequired && (!versioningConfig || versioningConfig.Status !== 'Enabled')) {
                 log.debug('bucket versioning is not enabled', {
                     method: request.method,
                     bucketName: request.bucketName,
--- a/tests/functional/raw-node/test/routes/routeBackbeat.js
+++ b/tests/functional/raw-node/test/routes/routeBackbeat.js
@@ -19,6 +19,7 @@ const TEST_BUCKET = 'backbeatbucket';
 const TEST_ENCRYPTED_BUCKET = 'backbeatbucket-encrypted';
 const TEST_KEY = 'fookey';
 const NONVERSIONED_BUCKET = 'backbeatbucket-non-versioned';
+const VERSION_SUSPENDED_BUCKET = 'backbeatbucket-version-suspended';
 const BUCKET_FOR_NULL_VERSION = 'backbeatbucket-null-version';
 
 const testArn = 'aws::iam:123456789012:user/bart';
@@ -120,7 +121,8 @@ describeSkipIfAWS('backbeat routes', () => {
         bucketUtil = new BucketUtility(
             'default', { signatureVersion: 'v4' });
         s3 = bucketUtil.s3;
-        bucketUtil.emptyManyIfExists([TEST_BUCKET, TEST_ENCRYPTED_BUCKET, NONVERSIONED_BUCKET])
+        bucketUtil.emptyManyIfExists([TEST_BUCKET, TEST_ENCRYPTED_BUCKET, NONVERSIONED_BUCKET,
+        VERSION_SUSPENDED_BUCKET])
             .then(() => s3.createBucket({ Bucket: TEST_BUCKET }).promise())
             .then(() => s3.putBucketVersioning(
                 {
@@ -130,6 +132,12 @@ describeSkipIfAWS('backbeat routes', () => {
             .then(() => s3.createBucket({
                 Bucket: NONVERSIONED_BUCKET,
             }).promise())
+            .then(() => s3.createBucket({ Bucket: VERSION_SUSPENDED_BUCKET }).promise())
+            .then(() => s3.putBucketVersioning(
+                {
+                    Bucket: VERSION_SUSPENDED_BUCKET,
+                    VersioningConfiguration: { Status: 'Suspended' },
+                }).promise())
             .then(() => s3.createBucket({ Bucket: TEST_ENCRYPTED_BUCKET }).promise())
             .then(() => s3.putBucketVersioning(
                 {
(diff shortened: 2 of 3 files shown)


nicolas2bert:
@bert-e reset

nicolas2bert:
ping

anurag4DSB (approved):
LGTM

lib/routes/routeBackbeat.js
```diff
@@ -1289,7 +1289,10 @@ function routeBackbeat(clientIP, request, response, log) {
                     [request.query.operation](request, response, log, next);
             }
             const versioningConfig = bucketInfo.getVersioningConfiguration();
-            if (!versioningConfig || versioningConfig.Status !== 'Enabled') {
+            // The following makes sure that only replication destination-related operations (non-GET requests)
+            // target buckets with versioning enabled.
+            // This allows lifecycle operations (which use GET requests) to work on non-versioned buckets.
+            if (request.method !== 'GET' && (!versioningConfig || versioningConfig.Status !== 'Enabled')) {
```
francoisferrand: lifecycle does not only perform `GET` operations : it also deletes objects (expiration), deletes data associated with objects (transition and restored-object-expiration), and puts metadata of objects...

so i fear such broad filtering in the router may introduce other issues.

→ Is there a way instead to identify the few "actions" performed by CRR replication (e.g. putObjectVersion, putObjectMetadata), maybe with some additional condition on the parameters/fields, to confirm? Then the check cna be performed direclty in each of these action handlers.
→ Otherwise, may be better to "isolate" the calls with specific CRR/lifecycle routes, or maybe simply add an extra parameter to enforce the check (`IfVersioned` maybe ?)
→ Alternatively, instead of making it a "core" feature in the code, could it be achieve through policies? e.g. the replication user will not have permissions to write to non-versioned bucket? (need to check if this is a kind of policy we support or even can define, but would be an elegant and simple solution I think)

francoisferrand: Replication is supposed to work on any versioned bucket : so it _must_ work as well on version-suspended buckets
```suggestion
            if (request.method !== 'GET' && (!versioningConfig || versioningConfig.Status === 'Disabled')) {
```

nicolas2bert: - S3C Lifecycle expiration (transition is not supported yet) only uses backbeat client GET operation: `getMetadata`. Delete object/version is done by the S3 client itself that uses the s3 route.

> Is there a way instead to identify the few "actions" performed by CRR replication (e.g. putObjectVersion, putObjectMetadata)
- Replication uses `putData`, `putMetadata` and `batchDelete` 's backebatAPI routes. Limiting the versioning check to those actions makes sense and is more explicit. However `putMetadata` is also used by lifecycle transition so it will not solve the lifecycle transition future implementation.

- This PR is a fix, so I am not comfortable introducing new CRR/lifeycle routes for it. But it will be needed once we introduce lifecycle transition in S3C.

> the replication user will not have permissions to write to non-versioned bucket?
- I thought about it, but it does not exist.

nicolas2bert: In CRR replication both source and replication have to be version enabled. I will keep enforcing it.

francoisferrand: `batchDelete` route is used for lifecycle as well.

> This PR is a fix, so I am not comfortable introducing new CRR/lifeycle routes for it. But it will be needed once we introduce lifecycle transition in S3C.

Fine with me ; however the code will waterflow and in the not-so-long future we should converge Cloudserver: so whatever you change will break lifecycle in Zenko.

So I'd rather make the check conditional anyway, and let the "caller" decide if the bucket must be versionned: e..g. simply add an extra parameter to enforce the check (`IfVersioned` maybe ?) : then backbeat can be updated to required versioning if and if required, without introducing backbeat logic/constraints in cloudserver.

lib/routes/routeBackbeat.js
```diff
@@ -1289,7 +1289,12 @@ function routeBackbeat(clientIP, request, response, log) {
                     [request.query.operation](request, response, log, next);
             }
             const versioningConfig = bucketInfo.getVersioningConfiguration();
-            if (!versioningConfig || versioningConfig.Status !== 'Enabled') {
+            // The following makes sure that only replication destination-related operations
+            // target buckets with versioning enabled.
+            // This allows lifecycle expiration operations (getMetadata) to work on non-versioned buckets.
+            const isCRRDestinationRequest = request.method === 'PUT' &&
+                (request.resourceType === 'data' || request.resourceType === 'metadata');
+            if (isCRRDestinationRequest && (!versioningConfig || versioningConfig.Status !== 'Enabled')) {
```
francoisferrand: To reduce coupling, I'd rather make this generic and let the caller (i.e. backbeat, who has full context!) specify if versioning is required, without introducing replication-related logic (which could actually change!) here.

something like:

```suggestion
            if (request.headers['x-scal-if-versionning-enabled'] && (!versioningConfig || versioningConfig.Status !== 'Enabled')) {
```

lib/routes/routeBackbeat.js
nicolas2bert: Yes, I am aligned with it.

lib/routes/routeBackbeat.js
nicolas2bert: updated.

lib/routes/routeBackbeat.js
```diff
@@ -1289,7 +1289,10 @@ function routeBackbeat(clientIP, request, response, log) {
                     [request.query.operation](request, response, log, next);
             }
             const versioningConfig = bucketInfo.getVersioningConfiguration();
-            if (!versioningConfig || versioningConfig.Status !== 'Enabled') {
+            // The following makes sure that only replication destination-related operations
+            // target buckets with versioning enabled.
+            const isVersioningEnabled = request.headers['x-scal-versioning-enabled'] === 'true';
```
francoisferrand: nit: `isVersioningEnabled` should be something like `checkVersioningEnabled`

similarly, i wonder if the header should be renamed to something more explicit, like `x-scal-versioning-required` or `x-scal-if-versioning-enabled` ?

tests/functional/raw-node/test/routes/routeBackbeat.js
```diff
@@ -1513,13 +1514,16 @@ describeSkipIfAWS('backbeat routes', () => {
             done();
         }));
```
francoisferrand: should add tests to verify versioning is not checked when `x-scal-versioning-enabled` is not defined or false.
("should not refuse PUT data when bucket is not versioned if x-scal-versioning-enabled is not set")

francoisferrand (approved):
LGTM, just some small cleanup.

In addition, I wonder if we should be more strict and fail the request if this header is passed to other routes (which do not support it) ?

nicolas2bert:
@bert-e approve

nicolas2bert:
@bert-e reset

PR #5708, opened 2024-11-28
Quota & negative inflights

ghost:
- Improve tests to ensure S3 will behave correctly if we use negative inflights
- Add support for quotas in backbeat routes (we considered only S3 router till now).

### DIFF
--- a/tests/quota/awsNodeSdk.js
+++ b/tests/quota/awsNodeSdk.js
@@ -607,6 +607,60 @@ function multiObjectDelete(bucket, keys, size, callback) {
             ], done);
         });
 
+        it('should allow writes after deleting data with quotas below the current number of inflights', done => {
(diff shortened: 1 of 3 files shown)


lib/routes/routeBackbeat.js
ghost, line 1: Backbeat routes will be tested in Zenko CI - existing tests are not up to date, plus they are skipped.

francoisferrand: this was a separate ticket (cldsrv-586) :/

ghost: Please feel free to do it separately, I will drop the commit

francoisferrand (approved):

KillianG (approved):

PR #5711, opened 2024-12-06
Improvement/cldsrv 553 bump deps

benzekrimaha:
Issue: CLDSRV-553


Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG NODE_VERSION=16.20.2-bookworm-slim
+ARG NODE_VERSION=22-bookworm-slim
```
francoisferrand: should stick to a specific node version, to ensure reproductible & repeatable builds

Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN apt-get update \
     && ssh-keyscan -H github.com > /root/ssh/known_hosts
 
 ENV PYTHON=python3
+RUN npm install -g node-gyp
 COPY package.json yarn.lock /usr/src/app/
 RUN npm install typescript@4.9.5 -g
```
francoisferrand: best to install all "global" packages in a single step,
before package.json copy (to avoid invalidatoin),
and specifying the node-gyp version (just like we do for typescript)

```suggestion
RUN npm install -g \
    node-gyp@11.0.0 \
    typescript@4.9.5
COPY package.json yarn.lock /usr/src/app/
```

lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -154,13 +154,6 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
                 generateRequestContext('objectPutTaggingVersion');
             requestContexts.push(putObjectVersionRequestContext);
         }
```
francoisferrand, line 157: why are we changing permissions?
if it was (already) a bug, it shoudl be fixed on 8.8 as well instead here...

lib/api/apiUtils/authorization/tagConditionKeys.js
```diff
@@ -61,7 +61,6 @@ function updateRequestContextsWithTags(request, requestContexts, apiMethod, log,
             return cb(err);
         }
         // FIXME introduced by CLDSRV-256, this syntax should be allowed by the linter
-        // eslint-disable-next-line no-restricted-syntax
         for (const rc of requestContexts) {
```
francoisferrand, line 64: instead of disabling the linter, we should either update the linter configuration (to allow this), or retry this in a conforming maner : `requestContext.each(rc -> { ... })`

lib/Config.js
```diff
@@ -1220,145 +1175,6 @@ class Config extends EventEmitter {
             maxStaleness,
             enableInflights,
         };
-        if (config.utapi) {
```
francoisferrand: we should keep the code, only need to replace the condition to make sure it never happens:


```suggestion
        if (false && config.utapi) {
```

(and we can keep the `parseUtapiReindex` function as well)

lib/utapi/utilities.js
francoisferrand, line 1: i think it may be better to keep Utapi in the code:
* only need to comment these few lines  the `require(utapi)` add an early retrun in `pushMetrics`
* the change will be much smaller (no need to change all places where this module is used), and thus easier to review & identify node22-related changes
* any change to utapi-code (in 7.x) will still be propagated with waterflow
* we can fix lint issues already, saving time when actually bumping utapi
* having it in a commit still makes it easy to revert, once utapi is upgraded

.github/workflows/tests.yaml
```diff
@@ -342,39 +342,6 @@ jobs:
         source: /tmp/artifacts
       if: always()
```
francoisferrand, line 363: should be skipped instead of removed:

```suggestion
  utapi-v2-tests:
    if: false
```

francoisferrand:
please rebase, there are conflicts.
(sorry, add some pending commensts, which may not be relevant anymore : please with this in mind...)

bin/list_metrics.js
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env node
```
francoisferrand, line 1: should keep this file, and just add early return (if needed) in `listMetrics()`

bin/list_bucket_metrics.js
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env node
-'use strict'; // eslint-disable-line strict
-
-require('../lib/utapi/utilities.js').listMetrics('buckets');
```
francoisferrand, line 4: should keep this file, and just add early return (if needed) in `listMetrics()`

.github/workflows/tests.yaml
```diff
@@ -361,38 +361,38 @@ jobs:
...
-    - name: Upload logs to artifacts
-      uses: scality/action-artifacts@v4
-      with:
-        method: upload
-        url: https://artifacts.scality.net
-        user: ${{ secrets.ARTIFACTS_USER }}
-        password: ${{ secrets.ARTIFACTS_PASSWORD }}
-        source: /tmp/artifacts
-      if: always()
+  # utapi-v2-tests:
```
francoisferrand, line 364: better to skip the job than to comment it:

```suggestion
  utapi-v2-tests:
    if: false
    [...]

File: .github/workflows/tests.yaml [hunk 10]
```diff
@@ -149,7 +149,7 @@ jobs:
       if: always()
 
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
francoisferrand, line 152: best to specify the version explicitely (24.04)

lib/api/apiUtils/authorization/prepareRequestContexts.js
francoisferrand: @benzekrimaha did you miss this one?

lib/api/bucketPutACL.js
```diff
@@ -261,6 +263,9 @@ function bucketPutACL(authInfo, request, log, callback) {
                         return next(null, bucket, revisedAddACLParams);
                     });
             }
+            if (hasError) {
```
francoisferrand: should be earlier in the code (eright after the `jsonGrants.forEach` loop) : otherwise we may continue while we had an error, and risk calling the callback again (like on line 263)

lib/api/objectPutACL.js
```diff
@@ -272,6 +274,9 @@ function objectPutACL(authInfo, request, log, cb) {
                             revisedAddACLParams);
                     });
             }
+            if (hasError) {
```
francoisferrand: same, should be called earlier : to avoid having a duplicate call via line 273...

lib/utapi/utapiReindex.js
```diff
@@ -1,11 +1,11 @@
...
-    reindexConfig.password = config.utapi && config.utapi.redis
-    && config.utapi.redis.password;
-}
-const reindex = new UtapiReindex(reindexConfig);
-reindex.start();
+// const reindexConfig = config.utapi && config.utapi.reindex;
+// if (reindexConfig && reindexConfig.password === undefined) {
+//     reindexConfig.password = config.utapi && config.utapi.redis
+//     && config.utapi.redis.password;
+// }
```
francoisferrand: no need to comment these

lib/utapi/utapiReplay.js
```diff
@@ -1,8 +1,8 @@
...
+// require('werelogs').stderrUtils.catchAndTimestampStderr();
+// const UtapiReplay = require('utapi').UtapiReplay;
+// const _config = require('../Config').config;
 
-const utapiConfig = _config.utapi &&
-    Object.assign({}, _config.utapi, { redis: _config.redis });
-const replay = new UtapiReplay(utapiConfig); // start utapi server
-replay.start();
+// const utapiConfig = _config.utapi &&
+//     Object.assign({}, _config.utapi, { redis: _config.redis });
```
francoisferrand: no need to comment

lib/Config.js
```diff
@@ -13,14 +13,11 @@ const { isValidBucketName } = s3routes.routesUtils;
...
-const {
-    azureAccountNameRegex,
-    base64Regex,
-    allowedUtapiEventFilterFields,
-    allowedUtapiEventFilterStates,
-    supportedLifecycleRules,
+const { azureAccountNameRegex, base64Regex,
+    allowedUtapiEventFilterFields, allowedUtapiEventFilterStates,
+    supportedLifecycleRules
 } = require('../constants');
```
francoisferrand: change not needed?

lib/Config.js
```diff
@@ -1236,7 +1233,7 @@ class Config extends EventEmitter {
             maxStaleness,
             enableInflights,
         };
-        if (config.utapi) {
+        if (config.utapi && false) { // eslint-disable-line no-constant-condition
```
francoisferrand: no need to change, will work in case we don't have utapi?

lib/Config.js
```diff
@@ -1368,8 +1365,9 @@ class Config extends EventEmitter {
                 this.utapi.filter = utapiResourceFilters;
             }
         }
+        // eslint-disable-next-line no-constant-condition
         if (Object.keys(this.locationConstraints).some(
-        loc => this.locationConstraints[loc].sizeLimitGB)) {
+        loc => this.locationConstraints[loc].sizeLimitGB) && false) {
```
francoisferrand: no need to 'fase' this, we should never _run_ into this code anyway

lib/Config.js
```diff
@@ -3,7 +3,7 @@ const { EventEmitter } = require('events');
 const fs = require('fs');
 const path = require('path');
 const url = require('url');
-const crypto = require('crypto');
+const mycrypto = require('crypto');
```
francoisferrand: why rename?
can't we use node22's new built-in crypto module?

tests/functional/aws-node-sdk/lib/utility/createEncryptedBucket.js
```diff
@@ -1,5 +1,5 @@
 const childProcess = require('child_process');
-const Promise = require('bluebird');
+const myPromise = require('bluebird');
```
francoisferrand: rename `bluebirdPromise` (or `bbPromise`).... or can we switch to using "standard" node.js Promise instead ?

tests/functional/aws-node-sdk/test/bucket/get.js
```diff
@@ -1,6 +1,6 @@
 const assert = require('assert');
 const tv4 = require('tv4');
-const Promise = require('bluebird');
+const myPromise = require('bluebird');
```
francoisferrand: same

tests/functional/aws-node-sdk/test/legacy/authV2QueryTests.js
```diff
@@ -1,5 +1,5 @@
 const assert = require('assert');
-const process = require('process');
+const myProcess = require('process');
```
francoisferrand: can't we switch to using the new standard process module?

```
const process = require('node:process');
```

tests/functional/aws-node-sdk/test/legacy/authV4QueryTests.js
```diff
@@ -1,5 +1,5 @@
 const assert = require('assert');
-const process = require('process');
+const myProcess = require('process');
```
francoisferrand: same

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/azureObjectCopy.js
```diff
@@ -31,7 +31,7 @@ const bigMD5 = '5f363e0e58a95f06cbe9bbc662c5dfb6';
 const emptyMD5 = 'd41d8cd98f00b204e9800998ecf8427e';
 const locMetaHeader = constants.objectLocationConstraintHeader.substring(11);
 
-const Promise = require('bluebird');
+const myPromise = require('bluebird');
```
francoisferrand: use default Promise & promisify?

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/objectCopy.js
```diff
@@ -19,7 +19,7 @@ const body = Buffer.from('I am a body', 'utf8');
 const correctMD5 = 'be747eb4b75517bf6b3cf7c5fbb62f3a';
 const emptyMD5 = 'd41d8cd98f00b204e9800998ecf8427e';
 const locMetaHeader = constants.objectLocationConstraintHeader.substring(11);
-const Promise = require('bluebird');
+const myPromise = require('bluebird');
```
francoisferrand: use default Promise & promisify?

tests/functional/aws-node-sdk/test/legacy/tests.js
```diff
@@ -1,5 +1,5 @@
 const assert = require('assert');
-const crypto = require('crypto');
+const myCrypto = require('crypto');
```
francoisferrand: use default node:crypto ?

tests/functional/aws-node-sdk/test/multipleBackend/put/put.js
```diff
@@ -11,7 +11,7 @@ const { versioningEnabled } = require('../../../lib/utility/versioning-util');
 const { describeSkipIfNotMultiple, getAwsRetry, awsLocation,
     awsLocationEncryption, memLocation, fileLocation, genUniqID, isCEPH }
     = require('../utils');
-const Promise = require('bluebird');
+const myPromise = require('bluebird');
```
francoisferrand: same

tests/functional/aws-node-sdk/test/multipleBackend/utils.js
```diff
@@ -1,5 +1,5 @@
 const assert = require('assert');
-const crypto = require('crypto');
+const myCrypto = require('crypto');
```
francoisferrand: same

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -1,7 +1,7 @@
 const assert = require('assert');
-const crypto = require('crypto');
+const myCrypto = require('crypto');
 
-const Promise = require('bluebird');
+const myPromise = require('bluebird');
```
francoisferrand: same

tests/functional/aws-node-sdk/test/quota/tooling.js
```diff
@@ -33,18 +33,20 @@ const sendRequest = async (method, host, path, body = '', config = null) =>
...
 
-        try {
-            const response = await fetch(url, options);
-            const text = await response.text();
-            const result = await xml2js.parseStringPromise(text);
-            if (result && result.Error) {
-                reject(result);
-            } else {
-                resolve(result);
+        (async () => {
```
francoisferrand: why do we need this?

since the code is already async/wait, we can simply remove the outer Promise, and keep the function async:

```
const sendRequest = async (method, host, path, body = '', config = null) =>
    const service = 's3';
    const endpoint = new AWS.Endpoint(host);
   
    [...]
   
    const response = await nodeFetch(url, options);
    const text = await response.text();
    const result = await xml2js.parseStringPromise(text);
    if (result && result.Error) {
        throw result;
    }

    return result;
}
```

tests/functional/metadata/MixedVersionFormat.js
```diff
@@ -97,7 +97,9 @@ describe('Mongo backend mixed bucket format versions', () => {
             });
         });
 
-        after(done => mongoClient.close(true).then(() => done()).catch(() => done()));
+        after(() => {
+            mongoClient.close(true);
+        });
```
francoisferrand: mongodClient.close() is async (ie returns a Promise), so need to make the function passed to after async as well: too ensure it waits until mongo client is closed.

```suggestion
        after(() => mongoClient.close(true));
```

or

```suggestion
        after(async () => {
            await mongoClient.close(true);
        });
```

tests/functional/metadata/MongoClientInterface.js
```diff
@@ -44,7 +44,7 @@ const updatedObjVal = { updated: true };
 const runIfMongo =
     process.env.S3METADATA === 'mongodb' ? describe : describe.skip;
 
-function unescape(obj) {
+function unescapeFunc(obj) {
```
francoisferrand:
```suggestion
function unescapeJSON(obj) {
```

tests/functional/raw-node/utils/HttpRequestAuthV4.js
```diff
@@ -1,11 +1,11 @@
-const crypto = require('crypto');
+const myCrypto = require('crypto');
```
francoisferrand: use node:crypto?

tests/functional/s3cmd/tests.js
```diff
@@ -1,7 +1,7 @@
-'use strict'; // eslint-disable-line strict
+'use strict';
 
 const proc = require('child_process');
-const process = require('process');
+const myProcess = require('process');
```
francoisferrand: use node:process ?

tests/functional/s3curl/tests.js
```diff
@@ -1,8 +1,8 @@
-'use strict'; // eslint-disable-line strict
+'use strict';
 
 const assert = require('assert');
 const proc = require('child_process');
-const process = require('process');
+const myProcess = require('process');
 const parseString = require('xml2js').parseString;
```
francoisferrand: same

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -1,7 +1,7 @@
 const assert = require('assert');
 const AWS = require('aws-sdk');
 const async = require('async');
-const crypto = require('crypto');
+const myCrypto = require('crypto');
```
francoisferrand: use node:crypto?

tests/multipleBackend/routes/routeVeeam.js
```diff
@@ -1,5 +1,5 @@
 const assert = require('assert');
-const crypto = require('crypto');
+const myCrypto = require('crypto');
```
francoisferrand: same

tests/quota/awsNodeSdk.js
```diff
@@ -338,11 +338,18 @@ function multiObjectDelete(bucket, keys, size, callback) {
...
-            return metadata.setup(err => wait(2000, () => done(err)));
+            await new Promise((resolve, reject) => {
+                metadata.setup(err => {
+                    if (err) {
+                        reject(err);
+                    }
+                    wait(2000, resolve);
+                });
+            });
         });
```
francoisferrand:
```suggestion
        before(() => {
            const config = getConfig('default', { signatureVersion: 'v4', maxRetries: 0 });
            s3Client = new S3(config);
            scuba.start();
            return new Promise((resolve, reject) => {
                metadata.setup(err => {
                    if (err) {
                        reject(err);
                    }
                    wait(2000, resolve);
                });
            });
        });
```

tests/unit/api/objectReplicationMD.js
```diff
@@ -77,7 +77,17 @@ const emptyReplicationMD = {
     role: '',
     storageType: '',
     dataStoreVersionId: '',
-    isNFS: null,
+    isNFS: undefined,
+};
+const emptyReplicationMDUndef = {
```
francoisferrand, line 82: did you mean `emptyReplicationInfo` ?

francoisferrand: I don't think we put replicationInfo with `isNFS: undefined` : same (empty) fields are indeed not returned anymore (change in arsenal), but we did not change the way we write the metadata, so the "input" parameter of the tests should not change


```suggestion
    isNFS: null,
};
const expectedEmptyReplicationMD = {
```

tests/unit/Config.js
```diff
@@ -470,7 +470,7 @@ describe('Config', () => {
             process.env.S3_CONFIG_FILE = oldConfig;
         });
 
-        it('should set up utapi local cache', () => {
+        it.skip('should set up utapi local cache', () => {
```
francoisferrand, line 473: this should make linter fail : as it does not, it seems something is missing from the config :-(
so we'll need to think about adding it later...

lib/api/apiUtils/authorization/prepareRequestContexts.js
benzekrimaha: it's not changing the permission ,this code is actually duplicated see here : https://github.com/scality/cloudserver/pull/5711/files#diff-465d8f1849b67c1a40847391c0a7ea074b6357a8b7cdeb2db2be89ad9dfce50bL79-L80

we are never going to cross this condition

tests/unit/api/objectReplicationMD.js
benzekrimaha: The test is failing otherwise waiting for undefined instead of nul

Dockerfile
francoisferrand: ping

lib/Config.js
francoisferrand: ?

lib/Config.js
```diff
@@ -13,14 +13,11 @@ const { isValidBucketName } = s3routes.routesUtils;
...
-    base64Regex,
-    allowedUtapiEventFilterFields,
-    allowedUtapiEventFilterStates,
-    supportedLifecycleRules,
+const { azureAccountNameRegex, base64Regex,
+    allowedUtapiEventFilterFields, allowedUtapiEventFilterStates,
+    supportedLifecycleRules
 } = require('../constants');
-const { utapiVersion } = require('utapi');
+// const { utapiVersion } = require('utapi');
```
francoisferrand: "mock" the variable definition, so you don't need to add `eslint-disable-line no-undef`

```suggestion
const utapiVersion = 0; // require('utapi');
```

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/azureObjectCopy.js
```diff
@@ -31,7 +31,7 @@ const bigMD5 = '5f363e0e58a95f06cbe9bbc662c5dfb6';
 const emptyMD5 = 'd41d8cd98f00b204e9800998ecf8427e';
 const locMetaHeader = constants.objectLocationConstraintHeader.substring(11);
 
-const Promise = require('bluebird');
+const bbPromise = require('bluebird');
```
francoisferrand, line 35: maybe not for this PR, but from your experience here, do you think we could/should remove this dependency? would it be resonnable?

if so, please create a followup ticket

francoisferrand: this is recomended by bluebird: https://github.com/petkaantonov/bluebird/#%EF%B8%8Fnote%EF%B8%8F

> Please use native promises instead if at all possible. Native Promises have been stable in Node.js and browsers for around 10 years now and they have been fast for around 7. Any utility bluebird has like .map has native equivalents (like Node streams' .map).
>
> This is a good thing, the people working on Bluebird and promises have been able to help incorporate most of the useful things from Bluebird into JavaScript itself and platforms/engines.
>
> If there is a feature that keeps you using bluebird. Please let us know so we can try and upstream it :)
>
> Currently - it is only recommended to use Bluebird if you need to support really old browsers or EoL Node.js or as an intermediate step to use warnings/monitoring to find bugs.

tests/functional/aws-node-sdk/test/quota/tooling.js
```diff
@@ -1,9 +1,9 @@
-const fetch = require('node-fetch');
+const nodeFetch = require('node-fetch');
 const AWS = require('aws-sdk');
 const xml2js = require('xml2js');
 
-const sendRequest = async (method, host, path, body = '', config = null) =>
-    new Promise(async (resolve, reject) => {
+const sendRequest = async (method, host, path, body = '', config = null) => {
+
         const service = 's3';
```
francoisferrand: * extra empty line
* bad indent: should be just 4 spaces

tests/quota/awsNodeSdk.js
```diff
@@ -338,11 +338,18 @@ function multiObjectDelete(bucket, keys, size, callback) {
...
-        before(done => {
+        before(() => {
             const config = getConfig('default', { signatureVersion: 'v4', maxRetries: 0 });
             s3Client = new S3(config);
             scuba.start();
-            return metadata.setup(err => wait(2000, () => done(err)));
+            return new Promise((resolve, reject) => {
+                metadata.setup(err => {
+                    if (err) {
+                        reject(err);
```
francoisferrand: missing `return`

tests/quota/awsNodeSdk.js
```diff
@@ -338,11 +338,18 @@ function multiObjectDelete(bucket, keys, size, callback) {
...
+                    }
+                    wait(2000, resolve);
+                });
+            });
```
francoisferrand: do we really _need_ to change the callback-based hook here?
as we are still using only callback-based functions internally, the updated code is more complex than the old one: if no constraint, may be better to keep it until those functions have been updated

tests/unit/api/objectReplicationMD.js
francoisferrand: > we did not change the way we write the metadata, so the "input" parameter of the tests should not change

if the test fail, it is probably because we at some point compare to `emptyReplicationMD` instead of `emptyReplicationMDUndef` (which would be better renamed `expectedEmptyReplicationMD` I think)

francoisferrand (commented):
LGTM overall, but a few comments have not been adressed yet (esp. in gha)

tests/functional/aws-node-sdk/test/multipleBackend/objectCopy/azureObjectCopy.js
benzekrimaha: indeed this needs to be reworked later on, here's the ticket created : https://scality.atlassian.net/browse/CLDSRV-605

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -292,16 +291,14 @@ describe('backbeat routes', () => {
             assert.strictEqual(StorageClass, 'STANDARD');
         }
 
-        beforeEach(done =>
+        beforeEach(() =>
             bucketUtil.emptyIfExists(BUCKET_FOR_NULL_VERSION)
                 .then(() => s3.createBucket({ Bucket: BUCKET_FOR_NULL_VERSION }).promise())
-                .then(() => done(), err => done(err))
         );
```
francoisferrand, line 297: i don't understand how this code "waits" for the bucket to be created :(
(it would be the case if promise was returned, but i don't see how it is here...)

benzekrimaha:
closing pull request as https://github.com/scality/cloudserver/pull/5734 was merged

PR #5712, opened 2024-12-06
Ensure backbeat retrieve the account quotas

ghost:
Fix for Quota during expiration

Issue: CLDSRV-590

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -1372,6 +1372,14 @@ function routeBackbeat(clientIP, request, response, log) {
     _normalizeBackbeatRequest(request);
     const requestContexts = prepareRequestContexts('objectReplicate', request);
 
+    if (request.resourceType === 'expiration') {
+        // Request account quotas, as it will not be added for the 'objectReplicate' action which
+        // is used by default for all backbeat operations
+        requestContexts.forEach(context => {
+            context._needQuota = true; // eslint-disable-line no-param-reassign
+        });
+    }
+
     // Ensure backbeat operations like expiration can properly use quotas
     // eslint-disable-next-line no-param-reassign
     request.finalizerHooks = [];
@@ -1388,7 +1396,7 @@ function routeBackbeat(clientIP, request, response, log) {
         const path = request.url.replace('/_/backbeat/api', '/_/');
         const { host, port } = config.backbeat;
         const target = `http://${host}:${port}${path}`;
-        return auth.server.doAuth(request, log, (err, userInfo) => {
+        return auth.server.doAuth(request, log, (err, userInfo, streamingV4Params, infos) => {
             if (err) {
                 log.debug('authentication error', {
                     error: err,
@@ -1398,6 +1406,8 @@ function routeBackbeat(clientIP, request, response, log) {
                 });
                 return responseJSONBody(err, null, response, log);
             }
+            // eslint-disable-next-line no-param-reassign
+            request.accountQuotas = infos?.accountQuota;
             // FIXME for now, any authenticated user can access API
             // routes. We should introduce admin accounts or accounts
             // with admin privileges, and restrict access to those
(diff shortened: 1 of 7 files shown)


benzekrimaha (approved):

francoisferrand (approved):

lib/api/objectDelete.js
```diff
@@ -51,6 +51,13 @@ function objectDeleteInternal(authInfo, request, log, isExpiration, cb) {
     const reqVersionId = decodedVidResult;
     const hasGovernanceBypass = hasGovernanceBypassHeader(request.headers);
 
+    // Reassign a specific apiMethod, as monitoring can rely on the
+    // type of cloudserver instance to determine it comes from Backbeat routes.
+    // This is also needed for quota evaluation, where "routeBackbeat" cannot
+    // be used for different operations, such as putObject
+    // eslint-disable-next-line no-param-reassign
+    request.apiMethod = reqVersionId ? 'objectDeleteVersion' : 'deleteObject';
```
francoisferrand: would it not be better or safer to keep the existing `apiMethod`, and just pass a different `requestType` in case of `isExpiration` ?
→ this ensures only `valParams` is changed, i.e. only the behavior of `standardMetadataValidateBucketAndObj()` is affected
→ it explicitely tweaks the behavior for the backbeat route only, ensuring the "user" api/path is not affected

lib/api/objectDelete.js
```diff
@@ -51,6 +51,13 @@ function objectDeleteInternal(authInfo, request, log, isExpiration, cb) {
     const reqVersionId = decodedVidResult;
     const hasGovernanceBypass = hasGovernanceBypassHeader(request.headers);
 
+    // Reassign a specific apiMethod, as monitoring can rely on the
```
francoisferrand: as far as I can tell (looking at the code), `apiMethod` is not used for monitoring: but mostly to
* override the default `requestType` passed to `standardMetadataValidateBucketAndObj()` in every api....
* in prepareRequestContexts().... but not for backbeat routes, which do the evaluation by assuming an api method of 'objectReplicate' 😱
* in the "standard" routes, to handle special handling of specific APIs (which may be needed for backbeat routes as welll...fortunately we don't seem to use these affected API -mostly `GetObject`/`PutObject`- through backbeat routes)

lib/api/objectDelete.js
francoisferrand: instead of the "generic" objectDelete, we do this only in `deleteObjectFromExpiration` :

```suggestion
function deleteObjectFromExpiration(request, response, userInfo, log, callback) {
    // Reassign a specific apiMethod, as it is needed for quota evaluation (at least), where
    // "routeBackbeat" cannot be used as it is used for all backbeat API operations...
    // eslint-disable-next-line no-param-reassign
    request.apiMethod = 'objectDelete';
```

lib/routes/routeBackbeat.js
```diff
@@ -1388,7 +1388,7 @@ function routeBackbeat(clientIP, request, response, log) {
         const path = request.url.replace('/_/backbeat/api', '/_/');
         const { host, port } = config.backbeat;
         const target = `http://${host}:${port}${path}`;
-        return auth.server.doAuth(request, log, (err, userInfo) => {
+        return auth.server.doAuth(request, log, (err, userInfo, streamingV4Params, infos) => {
```
francoisferrand: not the right one to change: this is used when forwarding to backbeat api pod...

the auth call for "cloudserver" backbeat routes is below... (line ~1465)

lib/routes/routeBackbeat.js
```diff
@@ -1176,7 +1178,24 @@ function batchDelete(request, response, userInfo, log, callback) {
                 return callback(err);
             }
             log.debug('batch delete successful', { locations });
-            return _respond(response, null, log, callback);
+
+            // Update inflight metrics for the data which has just been freed
+            const bucket = parsedPayload.Bucket;
```
ghost: this part LGTM

PR #5713, opened 2024-12-10
Add tests for object replacement in suspended buckets

francoisferrand:
- Add tests for version-suspended
- gha: add codecov for unit tests
- gha: split lint & unit tests stages

Issue: CLDSRV-594

### DIFF
--- a/tests/unit/api/apiUtils/quotas/quotaUtils.js
+++ b/tests/unit/api/apiUtils/quotas/quotaUtils.js
@@ -7,6 +7,7 @@ const {
     isMetricStale,
 } = require('../../../../../lib/api/apiUtils/quotas/quotaUtils');
 const QuotaService = require('../../../../../lib/quotas/quotas');
+const BucketInfo = require('arsenal').models.BucketInfo;
 
 const mockLog = {
     warn: sinon.stub(),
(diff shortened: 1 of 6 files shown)


tests/unit/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -637,28 +638,43 @@ describe('processBytesToWrite', () => {
...
+        ['the length when deleting expired object version', expiredObject, 'Suspended', 'versionId', -100],
+
+        // delete marker in version-suspended case
+        ['the length when adding delete marker over hot object', hotObject, 'Suspended', undefined, -100],
+        ['0 when adding delete marker over cold object', coldObject, 'Suspended', undefined, 0],
+        ['the length when adding delete marker over restoring object', restoringObject, 'Suspended', undefined, -100],
+        ['the length when adding delete marker over restored object', restoredObject, 'Suspended', undefined, -100],
+        ['the length when adding delete marker over expired object', expiredObject, 'Suspended', undefined, -100],
+    ].forEach(([scenario, object, versioningStatus, reqVersionId, expected]) => {
+        it(`should return ${scenario} and versioning is ${versioningStatus}`, () => {
```
ghost, line 675:
```suggestion
        it(`should return ${scenario} when versioning is ${versioningStatus}`, () => {
```

ghost (approved):

welansari (approved):

tests/unit/api/apiUtils/quotas/quotaUtils.js
francoisferrand: `and` is correct, as each scenario actually actually has a `when` : for example _"the length when adding delete marker over expired object"_... which gives the following when executed:

> it should return the length when adding delete marker over expired object and versioning is enabled

ghost: Yes that is why I commented, the correct word should be "when" or at least "and when", "and" alone creates ambiguity

"if" would also be ok

PR #5714, opened 2024-12-10
Standard authorization for backbeat routes

ghost:
**Better review commit per commit**

The goal here is to handle the authorization from the IAM module in the backbeat routes, and not rely on implicit deny === access denied, as this is not true anymore.
By doing so, we centralize the logic with all S3 APIs, and introduce some unit test to complete the functional tests from `tests/multipleBackend/routes/routeBackbeat.js`

Current state: functional tests for the ceph test suite and the ones using a real azure/aws/gcp are still not running because they were disabled long time ago, and the resources are not existing anymore, some are still there, but it will be done in a separate ticket.

### DIFF
--- a/docs/BACKBEAT_ROUTES.md
+++ b/docs/BACKBEAT_ROUTES.md
@@ -6,11 +6,11 @@ This special router is responsible for handling all the requests that are
 related to the Backbeat service. Backbeat may call any of the below APIs to
 perform operations on either data or s3 objects (metadata).
 
-These route follow the same authorization and validation as the S3 routes:
+These routes follow the same authorization and validation as the S3 routes:
 
 - Authorize the request with support for Implicit Denies from the IAM service.
 - Retrieve the bucket and object metadata if applicable.
-- Evaluate the S3 Bucket Policies and ACLs before authozing the request.
+- Evaluate the S3 Bucket Policies and ACLs before authorizing the request.
   - Backbeat routes are only authorized given the right permission, currently,
     `objectReplicate` as a unique permission for all these special routes.
   - In order to be authorized without S3 Bucket Policy, the caller must be
@@ -25,6 +25,7 @@ PUT /_/backbeat/metadata/<bucket name>/<object key>
 ```
 
 To edit one existing S3 Object's metadata.
+In the CRR case, this is used to put metadata for new objects.
 
 ```plaintext
 GET /_/backbeat/metadata/<bucket name>/<object key>?versionId=<version id>
--- a/docs/BACKBEAT_ROUTES.md
+++ b/docs/BACKBEAT_ROUTES.md
@@ -45,56 +45,56 @@ PUT /_/backbeat/multiplebackenddata/<bucket name>/<object key>?operation=putobje
 ```
 
 To put directly to the storage layer the data for an existing S3 Object.
-Use case: Cross Region Replication (CRR).
+Use case: Zenko Replication.
 
 ```plaintext
 PUT /_/backbeat/multiplebackenddata/<bucket name>/<object key>?operation=putpart
 ```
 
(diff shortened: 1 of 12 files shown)


docs/BACKBEAT_ROUTES.md
```diff
@@ -0,0 +1,152 @@
+# Backbeat routes
+
+Backbeat routes are implemented in `lib/routes/routeBackbeat.js`.
+
+This special router is responsible for handling all the requests that are
+related to the Backbeat service. Backbeat may call any of the below APIs to
+perform operations on either data or s3 objects (metadata).
+
+These route follow the same authorization and validation as the S3 routes:
```
KillianG:
```suggestion
These routes follows the same authorization and validation as the S3 routes:
```

docs/BACKBEAT_ROUTES.md
```diff
@@ -0,0 +1,152 @@
...
+
+- Authorize the request with support for Implicit Denies from the IAM service.
+- Retrieve the bucket and object metadata if applicable.
+- Evaluate the S3 Bucket Policies and ACLs before authozing the request.
```
KillianG:
```suggestion
- Evaluate the S3 Bucket Policies and ACLs before authorizing the request.
```

KillianG (approved):

lib/routes/routeBackbeat.js
```diff
@@ -300,51 +301,21 @@ function handleTaggingOperation(request, response, type, dataStoreVersionId,
...
-GET /_/backbeat/multiplebackendmetadata/<bucket name>/<object key>
-POST /_/backbeat/batchdelete
-GET /_/backbeat/lifecycle/<bucket name>?list-type=current
-GET /_/backbeat/lifecycle/<bucket name>?list-type=noncurrent
-GET /_/backbeat/lifecycle/<bucket name>?list-type=orphan
-POST /_/backbeat/index/<bucket name>?operation=add
-POST /_/backbeat/index/<bucket name>?operation=delete
-DELETE /_/backbeat/index/<bucket name>
-*/
-
```
welansari, line 347: Should we leave a comment somewhere in this file pointing to where the new API doc is at ? I feel like people working on this file will probably not go looking for the documentation and will just assume it doesn't exist, hence it not getting updated in the future.

docs/BACKBEAT_ROUTES.md
```diff
@@ -0,0 +1,152 @@
...
+    are allowed.
+- Finally, evaluate the quotas before allowing the request to proceed.
+
+## List of supported APIs
+
+```plaintext
+PUT /_/backbeat/metadata/<bucket name>/<object key>
+```
+
+To edit one existing S3 Object's metadata.
```
welansari, line 27: I believe this also puts metadata for new objects, this API is used by CRR.

lib/routes/routeBackbeat.js
ghost: I added something at the top of the file

docs/BACKBEAT_ROUTES.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+To put directly to the storage layer the data for an existing S3 Object.
+
+```plaintext
+PUT /_/backbeat/multiplebackenddata/<bucket name>/<object key>?operation=putobject
+```
+
+To put directly to the storage layer the data for an existing S3 Object.
+Use case: Cross Region Replication (CRR).
```
welansari: All `multiplebackenddata` APIs are used by Zenko Replication and not CRR.
CRR uses the following APIs:
```
GET /_/backbeat/metadata/<bucket name>/<object key>?versionId=<version id>
```
as well as 
```
POST /_/backbeat/batchdelete
```
which is used by other features as well.

welansari (approved):

PR #5716, opened 2024-12-16
Converge cloudserver images

francoisferrand:
Build federation image as additional layers on top of "standard" cloudserver image, to avoid duplication and ensure the image we tested is used in production.
There are still some differences at the moment (e.g. name of user, path) which should be further converged to minimize differences, may be done over time after considering all things.

This allows us to fully control (and minimize) what gets into the image, relying on any official node image. Also, now switch from python-based supervisord to a golang reimplementation [1], to minimize dependencies and CVEs.

Finally, took the change to fix startup command of the image, to ensure it properly propagates OS signals.

[1] ochinchina/supervisord

Issue: CLDSRV-597


images/federation/Dockerfile
```diff
@@ -0,0 +1,42 @@
+ARG BACKBEAT_VERSION=latest
```
benzekrimaha, line 1: ARG CLOUDSERVER_VERSION=latest

benzekrimaha: `ARG CLOUDSERVER_VERSION=latest`

images/federation/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN mkdir ${LOG_DIR} && chown ${USER} ${LOG_DIR} && \
+    mkdir ${CONF_DIR} && chown ${USER} ${CONF_DIR} && \
+    mkdir ${DATA_DIR} && chown ${USER} ${DATA_DIR} && \
+    mkdir -m 777 ${SUP_RUN_DIR} && chown ${USER} ${SUP_RUN_DIR}
+
+USER ${USER}
+WORKDIR ${HOME_DIR}/s3
+
+# Keep same output as chown command without group (use group 0)
+COPY --chown=${USER}:0 --from=builder /usr/src/app ${HOME_DIR}/backbeat
```
benzekrimaha: COPY --chown=${USER}:0 --from=builder /usr/src/app ${HOME_DIR}/cloudserver

.github/workflows/tests.yaml
```diff
@@ -177,6 +180,23 @@ jobs:
...
+      - name: Build and push federation image
+        uses: docker/build-push-action@v5
+        with:
+          push: true
+          context: images/federation
+          provenance: false
+          build-args:
+            CLOUDSERVER_VERSION=${{ github.sha }}
+          tags: |
+            ghcr.io/${{ github.repository }}:${{ github.sha }}-federation
```
ghost, line 193: shall we seize this opportunity to use a version before the sha?

francoisferrand: no completely trivial (need to compute the version) : will be done separately.
Also, this is needed on 8.8 as well: while this target 9.0 only.

ghost (approved):

benzekrimaha (approved):

rachedbenmustapha (commented):
Would be good to post to the #eng-federation chan for review.

PR #5718, opened 2024-12-18
CLDSRV-598: Add overhreadField to metadata-only operations

dvasilas:
Metadata-only operations (objectPutRetention, objectPutLegalHold, objectPutTagging, objectPutACL) overwrite the metadata of an object.

These operations currently do not attach the overheadField to metadata put operations.

As a result, the entries in the metadata op log for those operations do not include overhead information, and scuba interprets them as object puts, resulting in wrong metric updates.

Issue: CLDSRV-598

### DIFF
--- a/lib/api/objectPutACL.js
+++ b/lib/api/objectPutACL.js
@@ -12,6 +12,7 @@ const { decodeVersionId, getVersionIdResHeader, getVersionSpecificMetadataOption
 const { standardMetadataValidateBucketAndObj } = require('../metadata/metadataUtils');
 const monitoring = require('../utilities/metrics');
 const { config } = require('../Config');
+const { overheadField } = require('../../constants');
 
 /*
    Format of xml request:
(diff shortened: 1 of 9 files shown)


dvasilas:
ping

lib/api/objectPutACL.js
ghost, line 1: > As a result, the entries in the metadata op log for those operations do not include overhead information, and scuba interprets them as object puts, resulting in wrong metric updates.

We may want to also review the metadata operations done in backbeat or route backbeat here, just to make sure they doesn't lead to wrong behavior in scuba

ghost: Also, don't we have the same issue with metadata-only deletes, that also alter the metadata?

dvasilas: > We may want to also review the metadata operations done in backbeat or route backbeat here, just to make sure they doesn't lead to wrong behavior in scuba.

I created https://scality.atlassian.net/browse/S3C-9618 for that.

> Also, don't we have the same issue with metadata-only deletes, that also alter the metadata?

Good point ! 
I see `objectDeleteTagging` that should be fixed. Is there another one that I am missing ?

ghost: I don't see anything else, but maybe @Kerkesni or @francoisferrand will have some

dvasilas:
I am closing this because this is not the correct fix for the issue. 
The fix will be handled by CLDSRV-552

PR #5719, opened 2024-12-20
Support erasing inflights in CompleteMPU and AbortMPU

ghost:
The added test ensure that the logic within the API is coherent, as unit testing is not possible here.

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -146,6 +146,16 @@ const api = {
 
         const requestContexts = prepareRequestContexts(apiMethod, request,
             sourceBucket, sourceObject, sourceVersionId);
+
+        if (apiMethod === 'completeMultipartUpload' || apiMethod === 'abortMultipartUpload') {
+            // Request account quotas explicitly for MPU requests, to consider parts cleanup
+            // NOTE: we need quota for these, but it will be evaluated at the end of the API,
+            // once the parts have actually been deleted (not via standardMetadataValidateBucketAndObj)
+            requestContexts.forEach(context => {
+                context._needQuota = true; // eslint-disable-line no-param-reassign
+            });
+        }
+
         // Extract all the _apiMethods and store them in an array
         const apiMethods = requestContexts ? requestContexts.map(context => context._apiMethod) : [];
         // Attach the names to the current request
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -147,7 +147,7 @@ const api = {
         const requestContexts = prepareRequestContexts(apiMethod, request,
             sourceBucket, sourceObject, sourceVersionId);
 
-        if (apiMethod === 'completeMultipartUpload' || apiMethod === 'abortMultipartUpload') {
+        if (apiMethod === 'completeMultipartUpload' || apiMethod === 'multipartDelete') {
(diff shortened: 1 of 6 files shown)


lib/api/api.js
```diff
@@ -146,6 +146,16 @@ const api = {
...
             sourceBucket, sourceObject, sourceVersionId);
+
+        if (apiMethod === 'completeMultipartUpload' || apiMethod === 'multipartDelete') {
+            // Request account quotas explicitly for MPU requests, to consider parts cleanup
+            // NOTE: we need quota for these, but it will be evaluated at the end of the API,
+            // once the parts have actually been deleted (not via standardMetadataValidateBucketAndObj)
+            requestContexts.forEach(context => {
+                context._needQuota = true; // eslint-disable-line no-param-reassign
+            });
+        }
```
francoisferrand: should this not be added inside `prepareRequestContexts`, since it has all the different cases for each api method?

lib/api/completeMultipartUpload.js
```diff
@@ -499,31 +504,44 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
                     }
                     return next(null, mpuBucket, keysToDelete, aggregateETag,
                         extraPartLocations, destinationBucket,
-                        generatedVersionId);
+                        generatedVersionId, droppedMPUSize);
                 });
         },
         function deletePartsMetadata(mpuBucket, keysToDelete, aggregateETag,
-            extraPartLocations, destinationBucket, generatedVersionId, next) {
+            extraPartLocations, destinationBucket, generatedVersionId, droppedMPUSize, next) {
```
francoisferrand, line 517: "droppedMPUSize" can be computed around here (sum of every `extraPartLocation`'s size), as it is the place where we drop the extra parts : code will be must simpler & easier to follow, with no risk of mistake between all the steps...

this also makes it easier to handle partial delete object metadata errors : i.e. update quota with only the parts which were succesfully deleted...

lib/api/completeMultipartUpload.js
```diff
@@ -499,31 +504,44 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                        if (err) {
-                            return next(err);
-                        }
-                        return next(null, destinationBucket, aggregateETag,
-                            generatedVersionId);
+                return data.batchDelete(extraPartLocations, request.method, null, log, err => {
+                    if (err) {
+                        return next(err);
+                    }
+
```
francoisferrand, line 529: "droppedMPUSize" can be computed around here (sum of every extraPartLocation's size), as it is the place where we drop the extra parts : code will be must simpler & easier to follow, with no risk of mistake between all the steps...

this also makes it easier to handle partial delete object metadata errors : i.e. update quota with only the parts which were succesfully deleted...

ghost: I see you resolved the comment but just to confirm: the extraPartLocations do not have the information about the size of each part. We need to compute it before, so it is easier to process all parts at once to avoid processing the dropped parts only.

KillianG (approved):

benzekrimaha (approved):

PR #5724, opened 2025-01-06
Support SSL condition

ghost:
Currently, behind a proxy, we cannot detect the right http scheme. Also, we use the request.connection.encrypted variable, that is (1) deprecated and (2) already removed since Node 16, so the whole thing end up being undefined.

Here we call the new function in Arsenal to detect the right scheme, and add it to the request contexts constructor.

We also improve how bucket policies are evaluated:
- Previously we were computing the request context for each policy statement, and twice per statement
- There was a case where, if the request is undefined, then one of the internal evaluation would crash

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -255,28 +255,17 @@ function _checkBucketPolicyActions(requestType, actions, log) {
     return evaluators.isActionApplicable(mappedAction, actions, log);
 }
 
-function _checkBucketPolicyResources(request, resource, log) {
-    if (!request || (Array.isArray(resource) && resource.length === 0)) {
+function _checkBucketPolicyResources(requestContext, resource, log) {
+    if (!requestContext || (Array.isArray(resource) && resource.length === 0)) {
         return true;
     }
-    // build request context from the request!
-    const requestContext = new RequestContext(request.headers, request.query,
-        request.bucketName, request.objectKey, null,
-        request.connection.encrypted, request.resourceType, 's3');
     return evaluators.isResourceApplicable(requestContext, resource, log);
 }
 
-function _checkBucketPolicyConditions(request, conditions, log) {
-    const ip = request ? requestUtils.getClientIp(request, config) : undefined;
-    if (!conditions) {
+function _checkBucketPolicyConditions(requestContext, conditions, log) {
+    if (!requestContext || !conditions) {
         return true;
     }
-    // build request context from the request!
-    const requestContext = new RequestContext(request.headers, request.query,
-        request.bucketName, request.objectKey, ip,
-        request.connection.encrypted, request.resourceType, 's3', null, null,
-        null, null, null, null, null, null, null, null, null,
-        request.objectLockRetentionDays);
     return evaluators.meetConditions(requestContext, conditions, log);
 }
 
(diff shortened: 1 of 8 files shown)


lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -332,12 +321,29 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
-        const conditionsMatch = _checkBucketPolicyConditions(request, s.Condition, log);
+        let resourceMatch = false;
+        let conditionsMatch = false;
+        if (request) {
+            resourceMatch = _checkBucketPolicyResources(requestContext, s.Resource, log);
+            conditionsMatch = _checkBucketPolicyConditions(requestContext, s.Condition, log);
+        } else {
+            resourceMatch = true;
+            conditionsMatch = true;
+        }
```
francoisferrand:
```suggestion
        let resourceMatch = true;
        let conditionsMatch = true;
        if (requestContext) {
            resourceMatch = _checkBucketPolicyResources(requestContext, s.Resource, log);
            conditionsMatch = _checkBucketPolicyConditions(requestContext, s.Condition, log);
        }
```
or
```suggestion
        const resourceMatch = !requestContext || _checkBucketPolicyResources(requestContext, s.Resource, log);
        const conditionsMatch = !requestContext || _checkBucketPolicyConditions(requestContext, s.Condition, log);
```
or even maybe keep the check on `requestContext` inside `_checkBucketPolicyResources` and `_checkBucketPolicyResources` (in place of the check on `request`)

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -332,12 +321,29 @@ function checkBucketPolicy(policy, requestType, canonicalID, arn, bucketOwner, l
...
     }
     let copiedStatement = JSON.parse(JSON.stringify(policy.Statement));
+
+    const ip = request ? requestUtils.getClientIp(request, config) : undefined;
+    const isSecure = request ? requestUtils.getHttpProtocolSecurity(request, config) : undefined;
+    const requestContext = request ? new RequestContext(request.headers, request.query,
+        request.bucketName, request.objectKey, ip,
+        isSecure, request.resourceType, 's3', null, null,
+        null, null, null, null, null, null, null, null, null,
+        request.objectLockRetentionDays) : undefined;
```
francoisferrand, line 331: this checks cannot (and should not) be done in _internal_ cloudserver instances : which are used by internal services, and should not be affected by "user" policies.
→ could we "force" `ip` and `isSecure` to undefined in that context, to ensure such policies do not apply in that case?

lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -45,12 +45,12 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
...
 
     const ip = requestUtils.getClientIp(request, config);
+    const isSecure = requestUtils.getHttpProtocolSecurity(request, config);
 
     function generateRequestContext(apiMethod) {
         return new RequestContext(request.headers,
             request.query, request.bucketName, request.objectKey,
-            ip, request.connection.encrypted,
-            apiMethod, 's3');
+            ip, isSecure, apiMethod, 's3');
```
francoisferrand, line 53: building `requestContext` and/or `requestContextParams` seems very redundant, with lots of things repeatedly initiated from the `request`... could/should we refactor this (in a separate debt ticket) ?

ghost: Yes, there is something to improve here...

lib/api/apiUtils/authorization/permissionChecks.js
ghost: I am putting this on hold as we need to discuss a proper solution, even for the normal cloudserver, when the backbeat routes are called, to not block operations due to client-defined bucket policies (or even attached IAM policies?)

ghost: A good start could be to properly use `isLifecycleSession` (already existing, not used for bucket policy evaluation, so it's currently a dead code), to whitelist the bucket policy evaluation part for such use cases, what do you think? It would allow to keep the standard approach, but be able to consistently, for all s3 instances, "allow" from the bucket policies, for lifecycle internal identities.

lib/api/apiUtils/authorization/prepareRequestContexts.js
ghost: Created CLDSRV-603.

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Actually this is not the right PR to be on hold: the evaluation of bucket policies is currently not done in the backbeat routes context.
The PR on hold i https://github.com/scality/cloudserver/pull/5714

The problem is not the ssl/ip conditions, but everything: the bucket policies might still deny lifecycle operations if we, for example, deny all identities that are not a specific IAM user. SO no point setting them to undefined here: to be done in #5714, possible as I shared in my previous comment.

francoisferrand (approved):

welansari (approved):

PR #5725, opened 2025-01-07
CLDSRV-552: Set originOp for putObjectRetention and putObjectLegalHold

dvasilas:
`putObjectRetention` and `putObjectLegalHold` operations do not set the `originOp` metadata field and as a result copy the `originOp` of the previous operation on the object (e.g. `s3:ObjectCreated`).

Issue: CLDSRV-552

### DIFF
--- a/lib/api/objectPutLegalHold.js
+++ b/lib/api/objectPutLegalHold.js
@@ -94,6 +94,8 @@ function objectPutLegalHold(authInfo, request, log, callback) {
                 objectMD.replicationInfo = Object.assign({},
                     objectMD.replicationInfo, replicationInfo);
             }
+            // eslint-disable-next-line no-param-reassign
+            objectMD.originOp = 's3:ObjectLegalHold:Put';
             metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params,
                 log, err => next(err, bucket, objectMD));
         },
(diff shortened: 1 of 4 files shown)


fredmnl (approved):

anurag4DSB (approved):
LGTM

lib/api/objectPutLegalHold.js
```diff
@@ -94,6 +94,8 @@ function objectPutLegalHold(authInfo, request, log, callback) {
                 objectMD.replicationInfo = Object.assign({},
                     objectMD.replicationInfo, replicationInfo);
             }
+            // eslint-disable-next-line no-param-reassign
+            objectMD.originOp = 's3:ObjectLegalHold:Put';
```
ghost, line 98: I don't know if we want to reflect this event back to the client through bucket notifications but in Arsenal we have the `supportedNotificationEvents` variable in the `lib/constants.ts` file, that might require updates with these new events (if we want to return them).
If we edit arsenal, it should ideally be done before this PR.

dvasilas: We discussed about this with @francoisferrand, and we don't want to add event types in that list that are not in AWSs' list of supported event types: https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html .

Basically, if AWS does not send notifications for those operations (I checked that), we should not either.

tmacro (approved):

BourgoisMickael (approved):

PR #5728, opened 2025-01-10
Support transitions in OOB

francoisferrand:
Transition in OOB really only requires that an extra x-amz-meta-scal-version-id metadata is added when restoring the object, to allow backbeat to map the new objet being ingested with the existing metadata object.
This also requires this field to be added to overhead field, so it is provided in the oplog if the object is removed (from "backend" side) while restored.

In addition, as it requires using both endpoint, this opens new risk: to ensure OOB is "in-sync", writes through the OOB will not have the same VersionID as the backend. This is acutally not really a new behavior, as it allows this case to be consistent with the state of metadata when the object is simply ingested.

Issue: CLDSRV-563

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -195,6 +195,14 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         metadataStoreParams.oldReplayId = objMD.uploadId;
     }
 
+    if (isPutVersion && location === bucketMD.getLocationConstraint() && bucketMD.isIngestionBucket()) {
+        // When restoring to OOB bucket, we cannot force the versionId of the object written to the
+        // backend, and it is thus not match the versionId of the ingested object. Thus we add extra
+        // user metadata to allow OOB to allow ingestion processor to "match" the (new) restored
+        // object with the existing ingested object.
+        objectKeyContext.metaHeaders['x-amz-meta-scal-version-id'] = putVersionId;
+    }
+
     /* eslint-disable camelcase */
     const dontSkipBackend = externalBackends;
     /* eslint-enable camelcase */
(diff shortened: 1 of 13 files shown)


lib/api/objectCopy.js
```diff
@@ -492,6 +492,19 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                        // If the object is being written to the "ingested" storage location, keep the same
+                        // versionId for consistency and to avoid creating an extra version when it gets
+                        // ingested
+                        // eslint-disable-next-line
+                        const backendVersionId = decodeVID(destDataGetInfoArr[0].dataStoreVersionId);
+                        if (!(backendVersionId instanceof Error)) {
+                            options.versionId = backendVersionId; // eslint-disable-line no-param-reassign
+                        }
+                    }
+
```
Kerkesni, line 507: Copying an object into the same bucket the original object is at is a supported use case. Current behaviour is that it creates a new version of the object. This will introduce a kind of inconsistency depending on wether or not OOB is configured, as now it will rewrite the previous object (or just fail ?). Is this something we know and want ?

welansari (approved):

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -195,6 +195,17 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
...
 
+    if (isPutVersion && location === bucketMD.getLocationConstraint() && bucketMD.isIngestionBucket()) {
+        // When restoring to OOB bucket, we cannot force the versionId of the object written to the
+        // backend, and it is thus not match the versionId of the ingested object. Thus we add extra
+        // user metadata to allow OOB to allow ingestion processor to "match" the (new) restored
+        // object with the existing ingested object.
+        objectKeyContext.metaHeaders = {
+            ...objectKeyContext.metaHeaders,
+            'x-amz-meta-scal-version-id': putVersionId,
+        };
```
jonathan-gramain: Would this work?

```suggestion
        objectKeyContext.metaHeaders['x-amz-meta-scal-version-id'] = putVersionId;
```

lib/api/apiUtils/object/deleteObject.js
```diff
@@ -1,18 +1,22 @@
...
     // Default behavior is to require an oplog update
-    if (!bucket || !bucket.getLifecycleConfiguration || !bucket.getNotificationConfiguration) {
+    if (!bucket || !bucket.getNotificationConfiguration) {
         return true;
     }
-    // If the bucket has lifecycle configuration or notification configuration
-    // set, we also require an oplog update
-    return bucket.getLifecycleConfiguration() || bucket.getNotificationConfiguration();
+    // If the bucket has notification configuration set, we also require an oplog update
+    return bucket.getNotificationConfiguration();
```
jonathan-gramain: Just to satisfy the function's contract of returned type

```suggestion
    return Boolean(bucket.getNotificationConfiguration());
```

We may also use `!= null` but converting to Boolean avoids changing the behavior in case it's `undefined`.

jonathan-gramain (approved):

PR #5729, opened 2025-01-15
Support bigint during quota evaluation

ghost:
Support comparing quotas using bigint, to reduce the impact of handling large numbers.

Issue: CLDSRV-606

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -272,7 +272,7 @@ const api = {
             if (err) {
                 return callback(err);
             }
-            request.accountQuotas = infos?.accountQuota;
+            request.accountQuotas = BigInt(infos?.accountQuota || 0);
             if (authorizationResults) {
                 const checkedResults = checkAuthResults(authorizationResults);
                 if (checkedResults instanceof Error) {
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -272,7 +272,7 @@ const api = {
             if (err) {
                 return callback(err);
             }
-            request.accountQuotas = BigInt(infos?.accountQuota || 0);
+            request.accountQuotas = infos?.accountQuota;
             if (authorizationResults) {
                 const checkedResults = checkAuthResults(authorizationResults);
                 if (checkedResults instanceof Error) {
--- a/lib/api/apiUtils/quotas/quotaUtils.js
+++ b/lib/api/apiUtils/quotas/quotaUtils.js
@@ -122,7 +122,7 @@ function _evaluateQuotas(
                         return parallelDone(err);
                     }
                     if (!isMetricStale(bucketMetrics, 'bucket', bucket.getName(), action, inflight, log) &&
-                        bucketMetrics.bytesTotal + inflightForCheck > bucketQuota) {
+                        BigInt(bucketMetrics.bytesTotal || 0) + BigInt(inflightForCheck || 0) > bucketQuota) {
                         log.debug('Bucket quota exceeded', {
                             bucket: bucket.getName(),
                             action,
@@ -147,8 +147,9 @@ function _evaluateQuotas(
                     if (err || inflight < 0) {
                         return parallelDone(err);
                     }
+                    // Metrics are served as BigInt strings
                     if (!isMetricStale(accountMetrics, 'account', account.account, action, inflight, log) &&
-                        accountMetrics.bytesTotal + inflightForCheck > accountQuota) {
+                        BigInt(accountMetrics.bytesTotal || 0) + BigInt(inflightForCheck || 0) > accountQuota) {
                         log.debug('Account quota exceeded', {
                             accountId: account.account,
                             action,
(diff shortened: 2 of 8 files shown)


welansari (approved):

benzekrimaha (approved):
LGTM

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -122,7 +122,7 @@ function _evaluateQuotas(
                         return parallelDone(err);
                     }
                     if (!isMetricStale(bucketMetrics, 'bucket', bucket.getName(), action, inflight, log) &&
-                        bucketMetrics.bytesTotal + inflightForCheck > bucketQuota) {
+                        BigInt(bucketMetrics.bytesTotal || 0) + BigInt(inflightForCheck || 0) > bucketQuota) {
```
francoisferrand, line 125: should these conversions not be done by `ScubaClient` (when parsing bucketMetrics from result) ?

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -147,8 +147,9 @@ function _evaluateQuotas(
                     if (err || inflight < 0) {
                         return parallelDone(err);
                     }
+                    // Metrics are served as BigInt strings
                     if (!isMetricStale(accountMetrics, 'account', account.account, action, inflight, log) &&
-                        accountMetrics.bytesTotal + inflightForCheck > accountQuota) {
+                        BigInt(accountMetrics.bytesTotal || 0) + BigInt(inflightForCheck || 0) > accountQuota) {
```
francoisferrand, line 152: shoud these conversions not be done by `VaultClient` (when parsing accountMetrics from vault request)?

lib/api/api.js
```diff
@@ -272,7 +272,7 @@ const api = {
             if (err) {
                 return callback(err);
             }
-            request.accountQuotas = infos?.accountQuota;
+            request.accountQuotas = BigInt(infos?.accountQuota);
```
francoisferrand: infos is coming from vault, through vaultclient --> the BigInt conversion should be handled there

tests/utilities/mock/Scuba.js
```diff
@@ -31,21 +36,21 @@ class Scuba {
...
+            const immediateInflights = req.body?.action === 'objectRestore' ? 0n : inflight;
             return res.json({
-                bytesTotal: (this._data.bucket.get(bucketName)?.current || 0) +
-                    (this._data.bucket.get(bucketName)?.nonCurrent || 0) +
-                    (this._data.bucket.get(bucketName)?.inflight || 0) +
-                    immediateInflights,
+                bytesTotal: ((this._data.bucket.get(bucketName)?.current || 0n) +
+                    (this._data.bucket.get(bucketName)?.nonCurrent || 0n) +
+                    (this._data.bucket.get(bucketName)?.inflight || 0n) +
+                    immediateInflights).toString(),
```
francoisferrand: do we need toString() here? Is it not the purpose of `BigInt.prototype.toJSON` above? or should we remove that, if it does not work or is not needed?

lib/api/apiUtils/quotas/quotaUtils.js
ghost: I need to double check this one: vaultclient uses a unique function to handle response, hence we cannot easily do it. Plus it is also used as a cli tool so we need to ensure this breaks nothing.

benzekrimaha (approved):
LGTM

lib/api/apiUtils/quotas/quotaUtils.js
```diff
@@ -215,8 +216,9 @@ function validateQuotas(request, bucket, account, apiNames, apiMethod, inflight,
     let accountQuotaExceeded = false;
     let quotaEvaluationDuration;
     const requestStartTime = process.hrtime.bigint();
-    const bucketQuota = bucket.getQuota();
-    const accountQuota = account?.quota || 0;
+    const bucketQuota = bucket.getQuota() || 0n;
```
francoisferrand, line 219: do we have some sort of "functional" test, where the bucketInfo is retrieved from Mongo (and 'parsed' to BigInt) ?
should we wrap this in BigInt() constructor for extra safety? (`BigInt(bucket.getQuota())`) ?

francoisferrand (approved):

lib/api/apiUtils/quotas/quotaUtils.js
ghost: We have this test at the Zenko level, currently. I can update to be safe indeed, but the new BucketInfo class already do it in constructor or `setQuota`?

ghost: Actually yes we have the `tests/quota/awsNodeSdk.js` file, it uses a mock for scuba but it really create the buckets and uses the quota APIs to change the quotas, that are then evaluated.

PR #5734, opened 2025-02-13
Upgrade to NodeJS 22

ghost:
Several bumps and edits to make cloudserver compatible with node 22.

From https://github.com/scality/cloudserver/pull/5711

Two next tickets to work on after this PR: 
- ARSN-449 to move a function unit-tested here while it should be in arsenal.
- CLDSRV-596 to re-enable utapi-related code (also causing temporarily a drop in the codecov score).

New PR because all commits were reworked.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG NODE_VERSION=16.20.2-bookworm-slim
+ARG NODE_VERSION=22.4.0-bookworm-slim
```
francoisferrand: there is already 22.14.0-bullseye-slim, we should use it.

or should we switch to the newer 22.14.0-bookworm-slim ?

.github/workflows/tests.yaml
```diff
@@ -413,10 +415,10 @@ jobs:
         user: ${{ secrets.ARTIFACTS_USER }}
         password: ${{ secrets.ARTIFACTS_PASSWORD }}
```
francoisferrand: this simply disables the `Upload logs to artifacts` step : should be on the job (`utapi-v2-tests`) itself

Dockerfile
francoisferrand:
```suggestion
ARG NODE_VERSION=22.14.0-bookworm-slim
```

tests/functional/raw-node/package.json
```diff
@@ -8,11 +8,11 @@
...
-    "test-aws": "AWS_ON_AIR=true mocha -t 40000 test/",
-    "test-gcp": "mocha -t 40000 test/GCP/",
-    "test-routes": "mocha -t 40000 test/routes/",
-    "test": "mocha -t 40000 test/",
-    "test-debug": "_mocha -t 40000 test/"
+    "test-aws": "AWS_ON_AIR=true mocha -t 40000 test/ --exit",
+    "test-gcp": "mocha -t 40000 test/GCP/ --exit",
+    "test-routes": "mocha -t 40000 test/routes/ --exit",
+    "test": "mocha -t 40000 test/ --exit",
+    "test-debug": "_mocha -t 40000 test/ --exit"
```
francoisferrand, line 15: this is likely an issue in the tests, which should be fixed so we can remove the flag --> created CLDSRV-611 to track

tests/functional/aws-node-sdk/test/object/rangeTest.js
```diff
@@ -64,20 +64,28 @@ function checkRanges(range, bytes) {
...
+                Bucket: bucket,
+                Key: key,
+                PartNumber: part,
+                UploadId: uploadId,
+                Body: createReadStream(`${name}.mpuPart${part}`),
+            }).promise();
+        } catch (error) {
+            throw new Error(`Error uploading part ${part}: ${error.message}`);
+        }
+    }
```
francoisferrand: this is actually not the same, as the 2 parts which used to be uploaded in parallel are now send sequentially.

the equivalient behavior can be achieved with `Promise.all([1, 2].map(...))` (or `Promise.allSettled()`):
```suggestion
    return Promise.all([1, 2].map(async part => {
        try {
            await execFileAsync('dd', [
                `if=${name}`,
                `of=${name}.mpuPart${part}`,
                'bs=5242880', 
                `skip=${part - 1}`,
                'count=1',
            ]);
            await s3.uploadPart({
                Bucket: bucket,
                Key: key,
                PartNumber: part,
                UploadId: uploadId,
                Body: createReadStream(`${name}.mpuPart${part}`),
            }).promise();
        } catch (error) {
            throw new Error(`Error uploading part ${part}: ${error.message}`);
        }
    }));
```

not sure if this is relevant or really affects the tests...

tests/functional/aws-node-sdk/test/object/putObjAcl.js
```diff
@@ -65,23 +64,18 @@ describe('PUT Object ACL', () => {
...
-                .then(data => {
-                    assert(data);
-                    done();
-                })
-                .catch(done);
-        });
+            for (const param of objects) {
+                await s3.putObject(param).promise();
+            }
+            const data = await s3.putObjectAcl({ Bucket, Key, ACL: 'public-read' }).promise();
```
francoisferrand, line 76: used to be done in parallel (could still be done with `Promise.all()` or `Promise.allSettled()`) : is it important? may affect the perf/duration of the test?

tests/functional/aws-node-sdk/test/bucket/get.js
```diff
@@ -342,27 +341,18 @@ describe('GET Bucket - AWS.S3.listObjects', () => {
...
-                        }
-                        return data;
-                    }).then(data => {
-                        test.assertions(data, Bucket);
-                        done();
-                    })
-                    .catch(done);
+                for (const param of test.objectPutParams(Bucket)) {
+                    await s3.putObject(param).promise();
+                }
```
francoisferrand, line 349: used to be done in parallel, which can be achieved still with `Promise.all()` or `Promise.allSettled()`

package.json
```diff
@@ -19,79 +19,82 @@
...
-    "google-auto-auth": "^0.9.1",
-    "http-proxy": "^1.17.0",
-    "http-proxy-agent": "^4.0.1",
-    "https-proxy-agent": "^2.2.0",
+    "@azure/storage-blob": "^12.25.0",
+    "@hapi/joi": "^17.1.1",
+    "arsenal": "git+https://github.com/scality/arsenal#8.2.1",
+    "async": "2.6.4",
+    "aws-sdk": "^2.1692.0",
+    "bucketclient": "scality/bucketclient#878134d36e6b1bd65f1573292ed18f91a61335d0",
```
francoisferrand: just released:
```suggestion
    "bucketclient": "scality/bucketclient#8.2.0",
```

tests/functional/raw-node/package.json
ghost: It can also be as simple as http servers we open for the test, but don't properly close (not always the tests themselves)

tests/functional/aws-node-sdk/test/object/putObjAcl.js
ghost: It was done sequentially, as bluebird's mapSeries function was used, and it takes an iterator as input, so that won't change anything

tests/functional/aws-node-sdk/test/bucket/get.js
ghost: Same here, about the bluebird's `mapSeries` function. The equivalent is `Promise.each`.
cf http://bluebirdjs.com/docs/api/promise.mapseries.html

tests/functional/aws-node-sdk/test/object/rangeTest.js
ghost: I don't think this helps with test coverage, but I applied the suggestion.

ghost:
Note for reviewers: codecov is failing, but only due to deactivation of utapi. This is temporary and will be re-enabled in CLDSRV-610.

francoisferrand (approved):

benzekrimaha (approved):

PR #5743, opened 2025-02-20
extractProtocolFromHeader configuration handling

ghost:
- Bump Arsenal version (hdclient bump) -> Might be updated before merge to include https://github.com/scality/Arsenal/pull/2288
- Enforce the `extractProtocolFromHeader` configuration when a proxy is used, but set it to an empty string otherwise

Issue: CLDSRV-614


francoisferrand (approved):
LGTM

However I wonder if we also need to install `openssl` in the Docker image, as we need to do in backbeat (and it may be needed by entrypoint when `SSL=true`) : may not be related to this PR, but I'd rather we validate the need (or not) before making the 9.0.1 release.

benzekrimaha (approved):

PR #5748, opened 2025-02-25
CLDSRV-616: Fix bucket policy check for anonymous requests

dvasilas:
When checking bucket policies and the following conditions are true:
- The request is anonymous (`--no-sign-request`)
- There is a bucket policy with AWS principal

Then `_getAccountId` is called in `arn === undefined` and causes an exception to be thrown.

The reason is that vault return the following `authInfo` with anonymous requests:
`{
  arn: undefined,
  canonicalID: 'http://acs.amazonaws.com/groups/global/AllUsers',
  shortid: undefined,
  email: undefined,
  accountDisplayName: undefined,
  IAMdisplayName: undefined
}`

The fix is to check is to check is `arn === undefined` and fail the check if the policy principal is not `*`

### DIFF
--- a/tests/unit/api/bucketPolicyAuth.js
+++ b/tests/unit/api/bucketPolicyAuth.js
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const { BucketInfo, BucketPolicy } = require('arsenal').models;
+const AuthInfo = require('arsenal').auth.AuthInfo;
 const constants = require('../../../constants');
 const { isBucketAuthorized, isObjAuthorized, validatePolicyResource }
     = require('../../../lib/api/apiUtils/authorization/permissionChecks');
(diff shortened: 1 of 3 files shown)


anurag4DSB (approved):

tests/unit/api/bucketPolicyAuth.js
```diff
@@ -35,6 +36,9 @@ const basePolicyObj = {
 };
 const bucketName = 'matchme';
 const log = new DummyRequestLogger();
+const publicUserAuthInfo = new AuthInfo({
+    canonicalID: 'http://acs.amazonaws.com/groups/global/AllUsers',
```
BourgoisMickael: Can you pick this from constants: https://github.com/scality/cloudserver/blob/77d41dffcf6d59a9623586bde51f5adf2ac8bcc1/constants.js#L47

BourgoisMickael (approved):

ghost (approved):

dvasilas:
ping

PR #5749, opened 2025-02-26
Logging improvements

ghost:
Improve logging to use better levels, or add logs that were missing in `routeBackbeat`.
The delete operation also will log a warning when the expiration tries to delete a master version (due to any race condition).

Also bump Arsenal from https://github.com/scality/Arsenal/pull/2300

### DIFF
--- a/lib/api/apiUtils/object/coldStorage.js
+++ b/lib/api/apiUtils/object/coldStorage.js
@@ -253,7 +253,6 @@ function startRestore(objectMD, restoreParam, log, cb) {
         });
         return cb(checkResultError);
     }
-    log.debug('Updating restore information.');
     const updateResultError = _updateRestoreInfo(objectMD, restoreParam, log);
     if (updateResultError) {
         log.debug('Failed to update restore information.', {
@@ -262,6 +261,9 @@ function startRestore(objectMD, restoreParam, log, cb) {
         });
         return cb(updateResultError);
     }
+    log.debug('Validated and updated restore information', {
+         method: 'startRestore'
+    });
     const isObjectAlreadyRestored = _updateObjectExpirationDate(objectMD, log);
     return cb(null, isObjectAlreadyRestored);
 }
--- a/lib/api/apiUtils/object/coldStorage.js
+++ b/lib/api/apiUtils/object/coldStorage.js
@@ -244,14 +244,22 @@ function _updateRestoreInfo(objectMD, restoreParam, log) {
  *
  */
 function startRestore(objectMD, restoreParam, log, cb) {
-    log.info('Validating if restore can be done or not.');
+    log.debug('Validating if restore can be done or not.');
     const checkResultError = _validateStartRestore(objectMD, log);
     if (checkResultError) {
+        log.debug('Restore cannot be done.', {
+            error: checkResultError,
+            method: 'startRestore'
+        });
         return cb(checkResultError);
     }
-    log.info('Updating restore information.');
+    log.debug('Updating restore information.');
     const updateResultError = _updateRestoreInfo(objectMD, restoreParam, log);
     if (updateResultError) {
+        log.debug('Failed to update restore information.', {
+            error: updateResultError,
+            method: 'startRestore'
+        });
         return cb(updateResultError);
     }
     const isObjectAlreadyRestored = _updateObjectExpirationDate(objectMD, log);
--- a/lib/api/apiUtils/object/objectRestore.js
+++ b/lib/api/apiUtils/object/objectRestore.js
@@ -94,7 +94,7 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
                         log.trace('version is a delete marker', { method: METHOD, error: err });
                         return next(err, bucketMD, objectMD);
                     }
-                    log.info('it acquired the object metadata.', {
+                    log.debug('it acquired the object metadata.', {
                         'method': METHOD,
                     });
                     return next(null, bucketMD, objectMD);
@@ -108,7 +108,7 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
                     if (err) {
                         return next(err, bucketMD, objectMD, restoreInfo);
                     }
-                    log.info('it parsed xml of the request body.', { method: METHOD, value: restoreInfo });
+                    log.debug('it parsed xml of the request body.', { method: METHOD, value: restoreInfo });
                     const checkTierResult = checkTierSupported(restoreInfo);
                     if (checkTierResult instanceof Error) {
                         return next(checkTierResult);
--- a/lib/api/apiUtils/object/objectRestore.js
+++ b/lib/api/apiUtils/object/objectRestore.js
@@ -94,7 +94,7 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
                         log.trace('version is a delete marker', { method: METHOD, error: err });
                         return next(err, bucketMD, objectMD);
                     }
-                    log.debug('it acquired the object metadata.', {
+                    log.debug('acquired the object metadata.', {
                         'method': METHOD,
                     });
                     return next(null, bucketMD, objectMD);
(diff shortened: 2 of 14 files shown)


francoisferrand:
as a side note / complementary effort : we are changing some log levels here, but we should remember the logs (esp. debug/trace) are used in 2 contexts:
* Either when the user activate more verbose logging → display all the time, and indeed we may need to streamline these
* In case of error, every log related to the request is dumped again → in that specific context, having more (and separate lines) logs may/could be acceptable; but there may be another improvement in werelogs, to ensure these logs (some of which may already have been logged!) are clearly isolated: maybe by including them in the "final" json line or adding some sort of prefix/...,  instead of just printing them like regular logs

That is probably a separate improvement, but sharing as I think it may bring a different point of view on the updating of logs.

lib/api/apiUtils/object/coldStorage.js
```diff
@@ -244,14 +244,22 @@ function _updateRestoreInfo(objectMD, restoreParam, log) {
...
-    log.info('Updating restore information.');
+    log.debug('Updating restore information.');
     const updateResultError = _updateRestoreInfo(objectMD, restoreParam, log);
     if (updateResultError) {
+        log.debug('Failed to update restore information.', {
+            error: updateResultError,
+            method: 'startRestore'
+        });
         return cb(updateResultError);
     }
```
francoisferrand: should we dedup logs?
these are synchronous calls, not sure what it brings...
(+ add `method` attributes for consistency & allow filtering)

```suggestion
    const checkResultError = _validateStartRestore(objectMD, log);
    if (checkResultError) {
        log.debug('Restore cannot be done.', {
            error: checkResultError,
            method: 'startRestore'
        });
        return cb(checkResultError);
    }
    const updateResultError = _updateRestoreInfo(objectMD, restoreParam, log);
    if (updateResultError) {
        log.debug('Failed to update restore information.', {
            error: updateResultError,
            method: 'startRestore'
        });
        return cb(updateResultError);
    }
    log.debug('Validated and updated restore information', {
         method: 'startRestore'
    });
```

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -94,7 +94,7 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
                         log.trace('version is a delete marker', { method: METHOD, error: err });
                         return next(err, bucketMD, objectMD);
                     }
-                    log.info('it acquired the object metadata.', {
+                    log.debug('it acquired the object metadata.', {
```
francoisferrand:
```suggestion
                    log.debug('acquired the object metadata.', {
```

lib/api/apiUtils/object/objectRestore.js
```diff
@@ -108,7 +108,7 @@ function objectRestore(metadata, mdUtils, userInfo, request, log, callback) {
                     if (err) {
                         return next(err, bucketMD, objectMD, restoreInfo);
                     }
-                    log.info('it parsed xml of the request body.', { method: METHOD, value: restoreInfo });
+                    log.debug('it parsed xml of the request body.', { method: METHOD, value: restoreInfo });
```
francoisferrand:
```suggestion
                    log.debug('parsed xml of the request body.', { method: METHOD, value: restoreInfo });
```

lib/api/completeMultipartUpload.js
```diff
@@ -457,7 +457,7 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                 //   the previous MPU completion attempt
                 //
                 if (!locationKeysHaveChanged(objMD.location, dataLocations)) {
-                    log.info('MPU complete request replay detected', {
+                    log.debug('MPU complete request replay detected', {
```
francoisferrand: should not happen often, may be best to keep in INFO logs? or is this really so common (and safe) that this should only be in `debug`?

lib/routes/routeBackbeat.js
```diff
@@ -605,7 +605,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     // transition processor.
                     omVal.location && Array.isArray(omVal.location) &&
                     locationKeysHaveChanged(objMd.location, omVal.location)) {
-                    log.info('removing old data locations', {
+                    log.debug('removing old data locations', {
```
francoisferrand: i think this specific case is important to keep with a higher log level : this is a case where we actually remove _data_ on a `putMetadata` call...
(though it should be noted that in Zenko it goes through `internal-cloudserver`, which does has lower verbosity by default and does not log `info` level...)

lib/routes/routeBackbeat.js
```diff
@@ -1160,7 +1160,7 @@ function batchDelete(request, response, userInfo, log, callback) {
             }
             dataWrapper.data.delete(_loc, log, err => {
                 if (err?.is?.ObjNotFound) {
-                    log.info('batch delete: data location do not exist', {
+                    log.warn('batch delete: data location do not exist', {
```
francoisferrand: not sure this is a warning : it actually happens in some cases when deleting an archived object (cold).

lib/api/completeMultipartUpload.js
ghost: Yes, why not, I'm reverting this

package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.12.0",
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#8.1.146",
+    "arsenal": "git+https://github.com/scality/arsenal#ca9170ff0be40451934f3a4dc5687061f7cb35f2",
```
francoisferrand: reminder to update when arsenal is released

lib/routes/routeBackbeat.js
ghost: How about using a `warn` here? I think that would make sense, as we delete data.

ghost: I updated to warn, please let me know if you disagree.

francoisferrand: I don't think it is a warning, since we are expecting it to remove data: so in a sense WARN is too alarmist...

Increasing the log level allows seeing these in `internal-cloudserver`, but I am not sure we really _need_ to see everyone of these in general: it will happen for every transition or restore expiration, nothing significant from a system point of view.

Basically i think that it can be confusing that data gets deleted on a `putMetadata`, and esp. counter-intuitive: so should not require too much verbosity to surface, when people start looking for it.

ghost: Makes sense, I am reverting as info. I would say that having more log in internal cloudserver would be great anyway, at the very least, to know what requests are being processed (not only when there is an error), but at the same time, it leads to way less logs, so lower chances to keep all relevant error in the reports...

williamlardier: Done, closing this comment.

francoisferrand (approved):

lib/api/objectDelete.js
```diff
@@ -87,6 +87,20 @@ function objectDeleteInternal(authInfo, request, log, isExpiration, cb) {
                     // versioning has been configured
                     return next(null, bucketMD, objMD);
                 }
+
+                if (versioningCfg && versioningCfg.Status === 'Enabled' &&
+                    objMD.versionId === reqVersionId && isExpiration) {
+                        log.warn('expiration is trying to delete a master version ' +
```
welansari: should we filter out delete markers since it's a normal use case of lifecycle ?

ghost: but if we are creating a delete marker, there is no `reqVersionId`, right? in this case, this is already handled

welansari: Delete markers do have a version ID, and when deleting them, we do need to specify the version id, otherwise it is considered as if we are soft deleting the object that was already soft deleted. I think nothing is done in this case.

ghost: Ah ok so you meant not logging when we delete an object that was already marked for deletion (i.e. objMD.isDeleteMarker is `true`). I added this check.

welansari (approved):

benzekrimaha (approved):

PR #5751, opened 2025-02-26
Improvement/s3 c 9769/ignore trailing checksum

fredmnl:
# Motivation

Since aws-cli 2.23 (and associated SDK version), put-object now defaults to using unsigned payload with trailing checksums for integrity.

This is particularly nice from a computational point of view because the client only needs to read their file once to perform the upload. With AuthV4 signed payload, you would need to first compute the checksum of the object, to eventually include it in the signature contained in the header. With trailing checksums (and unsigned payload), the checksum is computed while the upload is happening and the checksum is sent at the end of the body. 

# Implementation

In this first iteration we are only interested in accepting the incoming requests using trailing checksums and unsigned payload. We are not yet verifying the checksum although the current proposal is extensible to perform this checksum verification as well.


lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,92 @@
...
+                    this.outputBuffer = Buffer.concat([this.outputBuffer, chunk]);
+                    break;
+                }
+                // chunk is bigger than the advertised size
+                this.outputBuffer = Buffer.concat([this.outputBuffer, chunk.subarray(0, this.bytesToRead)]);
+                chunk = chunk.subarray(this.bytesToRead);
+                this.bytesToRead = 0;
+                continue;
+            }
+            const lineBreakIndex = chunk.indexOf('\r\n');
```
jonathan-gramain: I think it needs to be:

```suggestion
const lineBreakIndex = this.chunkSizeBuffer.indexOf('\r\n');
```

Otherwise, it wouldn't be able to detect the newline with two chunks such as:

```
abcd\r  # chunk 1
\nefgh  # chunk 2
```

Then, you'll need to adapt the computed indices below with this change.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,92 @@
...
+     * This function will remove the trailing checksum from the stream
+     *
+     * @param {Buffer} chunkInput - chunk from request body
+     * @param {string} encoding - Data encoding
+     * @param {function} callback - Callback(err, justDataChunk, encoding)
+     * @return {function} executes callback with err if applicable
+     */
+    _transform(chunkInput, encoding, callback) {
+        let chunk = chunkInput;
+        for (;;) {
```
jonathan-gramain: go-ism? :wink: 

Maybe more idiomatic in JS: `while (true) {...}`

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,92 @@
...
+            }
+            if (this.streamClosed) {
+                return callback(null, '', encoding);
+            }
+            if (this.bytesToRead > 0) {
+                if (chunk.byteLength <= this.bytesToRead) {
+                    // chunk is smaller than the advertised size
+                    // forward the whole chunk
+                    this.bytesToRead -= chunk.byteLength;
+                    this.outputBuffer = Buffer.concat([this.outputBuffer, chunk]);
```
jonathan-gramain: Performance-wise, it may be better to store an array of buffers, to store individual chunks as they are received, instead of concatenating buffers repeatedly, especially if chunking on application side is much larger than nodejs chunking (normally 64KB by default IIRC).

So initializing with:

```
this.outputBuffer = [];
```

Then pushing to it each chunk with

```
this.outputBuffer.push(chunk);
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,92 @@
...
+            this.chunkSizeBuffer = Buffer.alloc(0);
+            if (dataSize === 0) {
+                // last chunk, no more data to read, the stream is closed
+                this.streamClosed = true;
+                break;
+            }
+            this.bytesToRead = dataSize;
+        }
+
+        this.toFlush = this.outputBuffer;
```
jonathan-gramain: Then to followup with `this.outputBuffer` being an array, instead you could repeatedly call `this.push()` on each item, so you don't have to concatenate buffers.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,92 @@
...
+     * @param {object} log - logger object
+     * @param {function} errCb - callback called if an error occurs
+     */
+    constructor(log, errCb) {
+        super({});
+        this.log = log;
+        this.errCb = errCb;
+        this.chunkSizeBuffer = Buffer.alloc(0);
+        this.outputBuffer = Buffer.alloc(0);
+        this.toFlush = Buffer.alloc(0);
```
jonathan-gramain: I don't think you need to store this field, as it just seems to be computed before being sent out in `_transform`.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,92 @@
...
+            this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk.subarray(0, lineBreakIndex)]);
+            chunk = chunk.subarray(lineBreakIndex + 2);
+            // chunk-size is sent in hex
+            const dataSize = Number.parseInt(this.chunkSizeBuffer.toString(), 16);
+            if (Number.isNaN(dataSize)) {
+                return callback(errors.InvalidArgument);
+            }
+            this.chunkSizeBuffer = Buffer.alloc(0);
+            if (dataSize === 0) {
+                // last chunk, no more data to read, the stream is closed
```
jonathan-gramain, line 89: I would add a TODO to implement parsing and checking the trailing checksum here.

jonathan-gramain (changes requested):

lib/auth/streamingV4/trailingChecksumTransform.js
jonathan-gramain: Actually you wouldn't even have to store the chunk references in `this.outputBuffer`, just `this.push()` them (or return them in callback) as they are received by `_transform`, until you reach the intended size.

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: Actually the linter does not like `while (true) {`, telling me that it does not like constant condition in a `while`. It does not pass the CI because it creates a warning. This `for (;;)` passes so :shrug:. What do you think?

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: Good point!

lib/auth/streamingV4/trailingChecksumTransform.js
jonathan-gramain: you could then replace it with `while (chunk.byteLength !== 0) {...}`.

Otherwise, could use an `eslint-disable` rule if the `while (true)` is justified.

fredmnl:
@jonathan-gramain Thanks again for the review, I reworked the whole class to address your concerns. 

I thought that my tests were covering all of the tricky splits that could happen but actually the http request module used was squashing the chunks together before sending. I resorted to coding a unit tests for the class which is much neater. I might have overdone it but they run pretty quickly so we could keep them all. `2487 passing (389ms)`

I reworked the logic in a way which I think is much neater. I'm only detecting `\r` instead of \r\n, which can't be split across chunks. I'm not actually checking for the '\n', but I think that's acceptable as the number of bytes cannot contain `\r` anyway. (We have never checked the `\r\n` after data for example). I'm effectively looking for regex `[^\r]+\r..{number_of_bytes}..` until `0\r` is found. 

I think that eventually we will want to refactor V4Transform to integrate this as well as checksum checking.

S3C-9776-repro.txt
jonathan-gramain, line 1: I suggest moving this file under `docs/` (possibly in a subdir of `docs`, but not sure how to name it), I feel the file is potentially useful but the top-level directory may not be the right place.

Or maybe somewhere in `tests/`?

conf/lab-ca-csr.cnf
jonathan-gramain, line 1: I am not sure you intended to commit those files in `conf`?

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+     * @param {object} log - logger object
+     * @param {function} errCb - callback called if an error occurs
+     */
+    constructor(log, errCb) {
+        super({});
+        this.log = log;
+        this.errCb = errCb;
+        this.chunkSizeBuffer = Buffer.alloc(0);
+        this.outputBuffer = [];
+        this.bytesToDiscard = 0; // when trailing /r/n are present, we discard them but they can be in different chunks
```
jonathan-gramain:
```suggestion
        this.bytesToDiscard = 0; // when trailing \r\n are present, we discard them but they can be in different chunks
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+                // no delimiter, we'll keep the chunk for later
+                this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk]);
+                break;
+            }
+
+            this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk.subarray(0, lineBreakIndex2)]);
+            chunk = chunk.subarray(lineBreakIndex2);
+
+            // chunk-size is sent in hex
+            const dataSize = Number.parseInt(this.chunkSizeBuffer.toString(), 16);
```
jonathan-gramain, line 79: Maybe not so important but as a refinement: we could validate that the string is indeed a well-formed hexadecimal integer until the end before parsing it into a number (e.g. with a regexp such as `/^[0-9a-fA-F]+$/.test(this.chunkSizeBuffer.toString())`), as `Number.parseInt` doesn't do full validation, in particular, it stops at the first invalid character encountered and returns the number parsed so far.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+                // TODO: check if the checksum is correct
+                // last chunk, no more data to read, the stream is closed
+                this.streamClosed = true;
+            }
+            this.bytesToRead = dataSize;
+            this.bytesToDiscard = 2;
+        }
+
+        // TODO: push the stream into a checksum accumulator
+        const toFlush = Buffer.concat(this.outputBuffer);
```
jonathan-gramain: To avoid concatenating the buffers, which I think requires copying memory around, I'd suggest to repeatedly call `this.push()` on each item of `this.outputBuffer` instead. We would then not return `toFlush` in the callback anymore.

This being said, we can even directly call `this.push()` in place of pushing to `this.outputBuffer` wherever it's done in this function, so we don't even have to keep that list.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+                this.outputBuffer.push(chunk.subarray(0, toRead));
+                chunk = chunk.subarray(toRead);
+                this.bytesToRead -= toRead;
+                if (this.bytesToRead === 0) {
+                    this.bytesToDiscard = 2;
+                }
+                continue;
+            }
+
+            const lineBreakIndex2 = chunk.indexOf('\r');
```
jonathan-gramain: It looks functional to me, I still have a slight preference for finding the full sequence `\r\n` in the `chunkSizeBuffer` because it also validates the stream contents more strongly. But it's not blocking for me, I'm fine with this approach too.

tests/unit/auth/TrailingChecksumTransform.js
```diff
@@ -0,0 +1,109 @@
...
+                assert.strictEqual(data, objDataWithoutTrailingChecksum);
+                done();
+            });
+            trailingChecksumTransform.on('error', err => {
+                assert.ifError(err);
+            });
+        });
+    });
+
+    // test all trisection of the input string
```
jonathan-gramain, line 82: While we're at it :smile: why not adding a single test where each chunk is a single byte?

S3C-9776-repro.txt
fredmnl: I'm sorry, I should have mentioned the DROP commits, I won't merge this file into the branch

conf/lab-ca-csr.cnf
fredmnl: Same here, won't merge these files

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: I think I'd like to refactor both Transforms (the AuthV4 too) into a single processor that would actually work on a configurable FSM to both validate and process the stream. I think it would also make the V4Transform much more readable. With that in mind, we could deliver this iteration as is (we're aiming for a timely release on this fix). There will at some point be a further iteration to validate the checksums (prioritization aside)

tests/unit/auth/TrailingChecksumTransform.js
fredmnl: I don't know if you're kidding or not, but I love the idea, I'm using it :+1: 

Should we keep the trisection test? It's about 4000 cases but they run really fast, I don't know what the good practice is here.

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: Ohhhh, I didn't understand when you wrote it above, I thought we needed to push the stream in the callback but `this.push` does the same thing and can be called multiple times. That's much neater, thanks!

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: Good point :+1:

tests/unit/auth/TrailingChecksumTransform.js
jonathan-gramain: I was not kidding :smile: 

Yeah I don't think the trisection tests hurt, especially if they all run in less than a second and they could catch some potential corner cases.

fredmnl:
@jonathan-gramain I think we're ready for another review here

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+                if (this.bytesToRead === 0) {
+                    this.bytesToDiscard = 2;
+                }
+                continue;
+            }
+
+            const lineBreakIndex = chunk.indexOf('\r');
+            if (lineBreakIndex === -1) {
+                if (this.chunkSizeBuffer.byteLength + chunk.byteLength > 10) {
+                    this.log.debug('chunk size field too big', {
```
jonathan-gramain: Two suggestions:

Turn into `info` level, as it can be useful for debugging on production platforms in case their app sends chunks that are too big for us to support

You could do first:

```
this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk]);
```

then check that the size is below what we support.

Then you can log directly `this.chunkSizeBuffer.subarray(0, 16).toString()`:

- You can remove the hex encoding since it's already JSON-encoded in logs

- I think 16 chars allow to see the full size, while 8 chars may mask the least significant digits of the hex-encoded length

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+            // chunk-size is sent in hex
+            if (!/^[0-9a-fA-F]+$/.test(this.chunkSizeBuffer.toString())) {
+                this.log.debug('chunk size is not a valid hex number', {
+                    chunkSizeBuffer: this.chunkSizeBuffer.toString('hex'),
+                });
+                return callback(errors.InvalidArgument);
+            }
+            const dataSize = Number.parseInt(this.chunkSizeBuffer.toString(), 16);
+            if (Number.isNaN(dataSize)) {
+                this.log.debug('unable to parse chunk size', {
```
jonathan-gramain: This case now looks unlikely due to the sanity check, so I think you can log it with error level

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+
+            const lineBreakIndex = chunk.indexOf('\r');
+            if (lineBreakIndex === -1) {
+                if (this.chunkSizeBuffer.byteLength + chunk.byteLength > 10) {
+                    this.log.debug('chunk size field too big', {
+                        chunkSizeBuffer: this.chunkSizeBuffer.toString('hex'),
+                        truncatedChunk: chunk.subarray(0, 8).toString('hex'),
+                    });
+                    // if bigger, the chunk would be over 5 GB
+                    // returning early to avoid a DoS by memory exhaustion
```
jonathan-gramain, line 61: I believe this comment is outdated now that we don't store the entire chunk in memory, maybe simply state that it would be suspicious to have such a large chunk.

tcarmet (approved):

tests/unit/auth/TrailingChecksumTransform.js
```diff
@@ -0,0 +1,131 @@
...
+            });
+        });
+    });
+
+    it('should correctly remove checksums, cut at each individual byte', done => {
+        const trailingChecksumTransform = new TrailingChecksumTransform(log, err => {
+            assert.strictEqual(err, null);
+        });
+        const chunks = [];
+        for (let i = 0; i < objDataWithTrailingChecksum.length - 1; i++) {
```
jonathan-gramain:
```suggestion
        for (let i = 0; i < objDataWithTrailingChecksum.length; i++) {
```

jonathan-gramain (approved):
LGTM with some minor suggestions

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: That's not what's happening. If somehow a client is not sending a payload with trailing checksums (because it messed up the headers for some reason), we would be storing the payload in the `chunkSizeBuffer`. This could be done maliciously as well. 

This safeguard is what enables us to keep the `chunkSizeBuffer` in memory and `concat` to it, since we know it will be at most a few bytes.

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: Sounds good, will do :+1:

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: (Also see comment below, which I wrote first)

`5GB` is an AWS limit, it's forbidden to perform a single upload of more than `5GB` hence the hard specification limit here, and `5GB` needs 9 hex characters to be written, we give it an extra character for `\r`. 

```
this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk]);
```

could potentially make us perform a huge buffer concatenation (although I think that Node would rechunk things before reaching this part of the code)

:+1: on hex

:+1: Yes 16 is better, in case the chunkSizeBuffer is empty, we would not see the whole 10 characters that we could have been interested in.

lib/api/apiUtils/object/prepareStream.js
```diff
@@ -24,11 +25,26 @@ function prepareStream(stream, streamingV4Params, log, errCb) {
...
         return v4Transform;
     }
     return stream;
 }
 
+function stripTrailingChecksumStream(stream, log, errCb) {
+    // don't do anything if x-amz-trailer has not been set
+    if (stream.headers['x-amz-trailer'] === undefined ||
+        stream.headers['x-amz-trailer'] === '') {
+        return stream;
```
ghost, line 38: We can just check if the value is set (both cases would return `true` here, and I believe we want to handle `null` the same way)

```suggestion
    if (!stream.headers['x-amz-trailer']) {
        return stream;
```

lib/api/apiUtils/object/validateChecksumHeaders.js
```diff
@@ -5,8 +5,10 @@ const { unsupportedSignatureChecksums, supportedSignatureChecksums } = require('
 function validateChecksumHeaders(headers) {
     // If the x-amz-trailer header is present the request is using one of the
     // trailing checksum algorithms, which are not supported.
-    if (headers['x-amz-trailer'] !== undefined) {
-        return errors.BadRequest.customizeDescription('trailing checksum is not supported');
+
+    if (headers['x-amz-trailer'] !== undefined &&
+        headers['x-amz-content-sha256'] !== 'STREAMING-UNSIGNED-PAYLOAD-TRAILER') {
+        return errors.BadRequest.customizeDescription('signed trailing checksum is not supported');
```
ghost, line 11: usually BadRequest (error 400) is for errors from client side, here, I would instead return a `NotImplemented` (error code 501)

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+            const dataSize = Number.parseInt(this.chunkSizeBuffer.toString(), 16);
+            if (Number.isNaN(dataSize)) {
+                this.log.error('unable to parse chunk size', {
+                    chunkSizeBuffer: this.chunkSizeBuffer.toString(),
+                });
+                return callback(errors.InvalidArgument);
+            }
+            this.chunkSizeBuffer = Buffer.alloc(0);
+            if (dataSize === 0) {
+                // TODO: check if the checksum is correct
```
ghost, line 88: Please create a ticket and mention the ticket number with the todo, it helps not forgetting it :smile:

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+class TrailingChecksumTransform extends Transform {
+    /**
+     * @constructor
+     * @param {object} log - logger object
+     * @param {function} errCb - callback called if an error occurs
+     */
+    constructor(log, errCb) {
+        super({});
+        this.log = log;
+        this.errCb = errCb;
```
ghost, line 17: We store this error callback in the class (which is a bit strange as passing such error function as a member of a class makes it complex to know when this will be called, and the impacts on the call path for the API).

But the main point is that this callback is not used anywhere: either we expect no error and we should remove this callback, or we can have error (handling of the `Transform` events?) and we need to use it somehow...

lib/api/apiUtils/object/prepareStream.js
```diff
@@ -24,11 +25,26 @@ function prepareStream(stream, streamingV4Params, log, errCb) {
...
+    }
+
+    const trailingChecksumTransform = new TrailingChecksumTransform(log, errCb);
```
ghost, line 41: We should not pass a callback for error handling here: we just pipe two streams, and then we can handle the errors in the same place as before.

Note: piping streams, in the past, led to memory leaks. We must ensure all errors /events are properly handled not to miss something here. See [here](https://github.com/scality/Arsenal/pull/2177/files) for an example.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk.subarray(0, lineBreakIndex)]);
+            chunk = chunk.subarray(lineBreakIndex);
+
+            // chunk-size is sent in hex
+            if (!/^[0-9a-fA-F]+$/.test(this.chunkSizeBuffer.toString())) {
+                this.log.info('chunk size is not a valid hex number', {
+                    chunkSizeBuffer: this.chunkSizeBuffer.toString(),
+                });
```
ghost, line 76:
```suggestion
                this.log.error('chunk size is not a valid hex number', {
                    chunkSizeBuffer: this.chunkSizeBuffer.toString(),
                });
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+                }
+                continue;
+            }
+
+            const lineBreakIndex = chunk.indexOf('\r');
+            if (lineBreakIndex === -1) {
+                if (this.chunkSizeBuffer.byteLength + chunk.byteLength > 10) {
+                    this.log.info('chunk size field too big', {
+                        chunkSizeBuffer: this.chunkSizeBuffer.toString(),
+                        truncatedChunk: chunk.subarray(0, 16).toString(),
```
ghost, line 58:
```suggestion
                    this.log.error('chunk size field too big', {
                        chunkSizeBuffer: this.chunkSizeBuffer.toString(),
                        truncatedChunk: chunk.subarray(0, 16).toString(),
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,100 @@
...
+                    // returning early to avoid a DoS by memory exhaustion
+                    return callback(errors.InvalidArgument);
+                }
+                // no delimiter, we'll keep the chunk for later
+                this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk]);
+                return callback();
+            }
+
+            this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk.subarray(0, lineBreakIndex)]);
+            chunk = chunk.subarray(lineBreakIndex);
```
ghost, line 70: I'm not sure this is correct (I may be wrong)
The delimiter is \r, but don't we have \n\r at the end of each chunk?

In such case, should we do something like

```
if (lineBreakIndex === -1 || lineBreakIndex + 1 >= chunk.length || chunk[lineBreakIndex + 1] !== '\n'.charCodeAt(0)) {
    // Handle error or wait for more data
}
this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk.subarray(0, lineBreakIndex)]);
chunk = chunk.subarray(lineBreakIndex + 2);
```

ghost (commented):
First review iteration, I will need to complete my review of the TrailingChecksumTransform class

fredmnl:
Closing this PR in favor of [https://github.com/scality/cloudserver/pull/5757](https://github.com/scality/cloudserver/pull/5757)

PR #5757, opened 2025-03-07
CLDSRV-620: Ignore trailing checksums in upload requests

fredmnl:
# Motivation

Since aws-cli 2.23 (and associated SDK version), put-object now defaults to using unsigned payload with trailing checksums for integrity.

This is particularly nice from a computational point of view because the client only needs to read their file once to perform the upload. With AuthV4 signed payload, you would need to first compute the checksum of the object, to eventually include it in the signature contained in the header. With trailing checksums (and unsigned payload), the checksum is computed while the upload is happening and the checksum is sent at the end of the body. 

# Implementation

In this first iteration we are only interested in accepting the incoming requests using trailing checksums and unsigned payload. We are not yet verifying the checksum although the current proposal is extensible to perform this checksum verification as well.

### DIFF
--- a/lib/api/apiUtils/object/prepareStream.js
+++ b/lib/api/apiUtils/object/prepareStream.js
@@ -1,4 +1,5 @@
 const V4Transform = require('../../../auth/streamingV4/V4Transform');
+const TrailingChecksumTransform = require('../../../auth/streamingV4/trailingChecksumTransform');
 
 /**
  * Prepares the stream if the chunks are sent in a v4 Auth request
@@ -24,11 +25,25 @@ function prepareStream(stream, streamingV4Params, log, errCb) {
         }
         const v4Transform = new V4Transform(streamingV4Params, log, errCb);
         stream.pipe(v4Transform);
+        v4Transform.headers = stream.headers;
         return v4Transform;
     }
     return stream;
 }
 
+function stripTrailingChecksumStream(stream, log) {
+    // don't do anything if we are not in the correct integrity check mode
+    if (stream.headers['x-amz-content-sha256'] !== 'STREAMING-UNSIGNED-PAYLOAD-TRAILER') {
+        return stream;
+    }
+
+    const trailingChecksumTransform = new TrailingChecksumTransform(log);
+    stream.pipe(trailingChecksumTransform);
+    trailingChecksumTransform.headers = stream.headers;
+    return trailingChecksumTransform;
+}
+
 module.exports = {
     prepareStream,
+    stripTrailingChecksumStream,
 };
--- a/lib/auth/streamingV4/trailingChecksumTransform.js
+++ b/lib/auth/streamingV4/trailingChecksumTransform.js
@@ -1,0 +1,119 @@
+const { Transform } = require('stream');
+const { errors } = require('arsenal');
+const { maximumAllowedPartSize } = require('../../../constants');
+
+/**
+ * This class is designed to handle the chunks sent in a streaming
+ * unsigned playload trailer request. In this iteration, we are not checking
+ * the checksums, but we are removing them from the stream.
+ * S3C-9732 will deal with checksum verification.
+ */
+class TrailingChecksumTransform extends Transform {
+    /**
+     * @constructor
+     * @param {object} log - logger object
+     */
+    constructor(log) {
+        super({});
+        this.log = log;
+        this.chunkSizeBuffer = Buffer.alloc(0);
+        this.bytesToDiscard = 0; // when trailing \r\n are present, we discard them but they can be in different chunks
+        this.bytesToRead = 0; // when a chunk is advertised, the size is put here and we forward all bytes
+        this.streamClosed = false;
+    }
+
+    /**
+     * This function is executed when there is no more data to be read but before the stream is closed
+     * We will verify that the trailing checksum structure was upheld
+     *
+     * @param {function} callback - Callback(err, data)
+     * @return {function} executes callback with err if applicable
+     */
+    _flush(callback) {
+        if (!this.streamClosed) {
+            this.log.error('stream ended without closing chunked encoding');
+            return callback(errors.InvalidArgument);
+        }
+        return callback();
+    }
+
+    /**
+     * This function will remove the trailing checksum from the stream
+     *
+     * @param {Buffer} chunkInput - chunk from request body
+     * @param {string} encoding - Data encoding
+     * @param {function} callback - Callback(err, justDataChunk, encoding)
+     * @return {function} executes callback with err if applicable
+     */
+    _transform(chunkInput, encoding, callback) {
+        let chunk = chunkInput;
+        while (chunk.byteLength > 0 && !this.streamClosed) {
+            if (this.bytesToDiscard > 0) {
+                const toDiscard = Math.min(this.bytesToDiscard, chunk.byteLength);
+                chunk = chunk.subarray(toDiscard);
+                this.bytesToDiscard -= toDiscard;
+                continue;
+            }
+            // forward up to bytesToRead bytes from the chunk, restart processing on leftover
+            if (this.bytesToRead > 0) {
+                const toRead = Math.min(this.bytesToRead, chunk.byteLength);
+                this.push(chunk.subarray(0, toRead));
+                chunk = chunk.subarray(toRead);
+                this.bytesToRead -= toRead;
+                if (this.bytesToRead === 0) {
+                    this.bytesToDiscard = 2;
+                }
+                continue;
+            }
+
+            // we are now looking for the chunk size field
+            // no need to look further than 10 bytes since the field cannot be bigger: the max
+            // chunk size is 5GB (see constants.maximumAllowedPartSize)
+            const lineBreakIndex = chunk.subarray(0, 10).indexOf('\r');
+            const bytesToKeep = lineBreakIndex === -1 ? chunk.byteLength : lineBreakIndex;
+            if (this.chunkSizeBuffer.byteLength + bytesToKeep > 10) {
+                this.log.error('chunk size field too big', {
+                    chunkSizeBuffer: this.chunkSizeBuffer.subarray(0, 11).toString('hex'),
+                    chunkSizeBufferLength: this.chunkSizeBuffer.length,
+                    truncatedChunk: chunk.subarray(0, 10).toString('hex'),
+                });
+                // if bigger, the chunk would be over 5 GB
+                // returning early to avoid a DoS by memory exhaustion
+                return callback(errors.InvalidArgument);
+            }
+            if (lineBreakIndex === -1) {
+                // no delimiter, we'll keep the chunk for later
+                this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk]);
+                return callback();
+            }
+
+            this.chunkSizeBuffer = Buffer.concat([this.chunkSizeBuffer, chunk.subarray(0, lineBreakIndex)]);
+            chunk = chunk.subarray(lineBreakIndex);
+
+            // chunk-size is sent in hex
+            const chunkSizeStr = this.chunkSizeBuffer.toString();
+            const dataSize = parseInt(chunkSizeStr, 16);
+            // we check that the parsing is correct (parseInt returns a partial parse when it fails)
+            if (isNaN(dataSize) || dataSize.toString(16) !== chunkSizeStr.toLowerCase()) {
+                this.log.error('invalid chunk size', { chunkSizeBuffer: chunkSizeStr });
+                return callback(errors.InvalidArgument);
+            }
+            this.chunkSizeBuffer = Buffer.alloc(0);
+            if (dataSize === 0) {
+                // TODO: check if the checksum is correct (S3C-9732)
+                // last chunk, no more data to read, the stream is closed
+                this.streamClosed = true;
+            }
+            if (dataSize > maximumAllowedPartSize) {
+                this.log.error('chunk size too big', { dataSize });
+                return callback(errors.EntityTooLarge);
+            }
+            this.bytesToRead = dataSize;
+            this.bytesToDiscard = 2;
+        }
+
+        return callback();
+    }
+}
+
+module.exports = TrailingChecksumTransform;
--- a/tests/functional/raw-node/test/trailingChecksums.js
+++ b/tests/functional/raw-node/test/trailingChecksums.js
@@ -1,0 +1,133 @@
+const assert = require('assert');
+const async = require('async');
+const { makeS3Request } = require('../utils/makeRequest');
+const HttpRequestAuthV4 = require('../utils/HttpRequestAuthV4');
+
+const bucket = 'testunsupportedchecksumsbucket';
+const objectKey = 'key';
+const objData = Buffer.alloc(1024, 'a');
+// note this is not the correct checksum in objDataWithTrailingChecksum
+const objDataWithTrailingChecksum = '10\r\n0123456789abcdef\r\n' +
+                                    '10\r\n0123456789abcdef\r\n' +
+                                    '0\r\nx-amz-checksum-crc64nvme:YeIDuLa7tU0=\r\n';
+const objDataWithoutTrailingChecksum = '0123456789abcdef0123456789abcdef';
+
+const config = require('../../config.json');
+const authCredentials = {
+    accessKey: config.accessKey,
+    secretKey: config.secretKey,
+};
+
+const itSkipIfAWS = process.env.AWS_ON_AIR ? it.skip : it;
+
+describe('trailing checksum requests:', () => {
+    before(done => {
+        makeS3Request({
+            method: 'PUT',
+            authCredentials,
+            bucket,
+        }, err => {
+            assert.ifError(err);
+            done();
+        });
+    });
+
+    after(done => {
+        async.series([
+            next => makeS3Request({
+                method: 'DELETE',
+                authCredentials,
+                bucket,
+                objectKey,
+            }, next),
+            next => makeS3Request({
+                method: 'DELETE',
+                authCredentials,
+                bucket,
+            }, next),
+        ], err => {
+            assert.ifError(err);
+            done();
+        });
+    });
+
+    it('should accept unsigned trailing checksum', done => {
+        const req = new HttpRequestAuthV4(
+            `http://localhost:8000/${bucket}/${objectKey}`,
+            Object.assign(
+                {
+                    method: 'PUT',
+                    headers: {
+                        'content-length': objDataWithTrailingChecksum.length,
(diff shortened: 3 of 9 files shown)


lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+                this.push(chunk.subarray(0, toRead));
+                chunk = chunk.subarray(toRead);
+                this.bytesToRead -= toRead;
+                if (this.bytesToRead === 0) {
+                    this.bytesToDiscard = 2;
+                }
+                continue;
+            }
+
+            const lineBreakIndex = chunk.indexOf('\r');
```
ghost: `\r` alone is not the documented CRLF (`\r\n`), and might be present in the data alone (as the buffer might have any data, unless we confirm this is not possible?). We must not split based on this character only. For instance the following unit test should pass, but it returns an `InvalidArgument` error:

```js
    it.only('should handle CR in data properly with fixed implementation', done => {
        const trailingChecksumTransform = new TrailingChecksumTransform(log);
        const chunks = [
            // First chunk with size and CRLF delimiter
            Buffer.from('5\r\n'),
            // Second chunk with data containing a CR character
            Buffer.from('a\rb\nc'),
            // Third chunk with CRLF after data, then next chunk
            Buffer.from('\r\n5\r\nhello\r\n0\r\nchecksum:xyz=\r\n')
        ];
        
        const chunkedReader = new ChunkedReader(chunks);
        chunkedReader.pipe(trailingChecksumTransform);
        const outputChunks = [];
        
        trailingChecksumTransform.on('data', chunk => {
            console.log('Data:', chunk.toString());
            outputChunks.push(Buffer.from(chunk));
        });
        
        trailingChecksumTransform.on('finish', () => {
            const data = Buffer.concat(outputChunks).toString();
            console.log('Final Output:', data);
            assert.strictEqual(data, 'a\rb\nc');
            done();
        });
        
        trailingChecksumTransform.on('error', err => {
            console.log('Error:', err);
            done(err);
        });
    });
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
+const { Transform } = require('stream');
+const { errors } = require('arsenal');
+
+/**
+ * This class is designed to handle the chunks sent in a streaming
+ * v4 Auth request
+ */
+class TrailingChecksumTransform extends Transform {
```
ghost, line 11: If you need to handle errors in this class, you may implement the `_flush` method, something like

```js
_flush(callback) {
    if (!this.streamClosed) {
        this.log.error('stream ended without closing chunked encoding');
        return callback(errors.InvalidArgument);
    }
    callback();
}
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
+const { Transform } = require('stream');
+const { errors } = require('arsenal');
+
+/**
+ * This class is designed to handle the chunks sent in a streaming
+ * v4 Auth request
+ */
```
ghost: Worth adding a comment to explain that we are not validating the checksums, and why

lib/api/apiUtils/object/prepareStream.js
```diff
@@ -24,11 +25,26 @@ function prepareStream(stream, streamingV4Params, log, errCb) {
...
+    // don't do anything if we are not in the correct integrity check mode
+    if (stream.headers['x-amz-content-sha256'] !==
+        'STREAMING-UNSIGNED-PAYLOAD-TRAILER') {
+        return stream;
+    }
+
+    const trailingChecksumTransform = new TrailingChecksumTransform(log);
+    stream.pipe(trailingChecksumTransform);
+    trailingChecksumTransform.headers = stream.headers;
+    return trailingChecksumTransform;
```
ghost, line 43: Good to confirm as well if we are still enforcing the common md5 check in this mode (both stored in the object md and that we properly fail in case of mismatch). Could be through a new unit/ft test.

lib/auth/streamingV4/trailingChecksumTransform.js
ghost: Closing as discussed as it works as expected

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+                });
+                return callback(errors.InvalidArgument);
+            }
+            const dataSize = Number.parseInt(this.chunkSizeBuffer.toString(), 16);
+            if (Number.isNaN(dataSize)) {
+                this.log.error('unable to parse chunk size', {
+                    chunkSizeBuffer: this.chunkSizeBuffer.toString(),
+                });
+                return callback(errors.InvalidArgument);
+            }
```
ghost:
```suggestion
            const chunkSizeStr = this.chunkSizeBuffer.toString();
            const dataSize = parseInt(chunkSizeStr, 16);
            if (isNaN(dataSize) || dataSize.toString(16) !== chunkSizeStr.toLowerCase()) {
                  this.log.error('invalid chunk size', { chunkSizeBuffer: chunkSizeStr });
                  return callback(errors.InvalidArgument);
            }
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+                continue;
+            }
+
+            const lineBreakIndex = chunk.indexOf('\r');
+            if (lineBreakIndex === -1) {
+                if (this.chunkSizeBuffer.byteLength + chunk.byteLength > 10) {
+                    this.log.error('chunk size field too big', {
+                        chunkSizeBuffer: this.chunkSizeBuffer.toString(),
+                        truncatedChunk: chunk.subarray(0, 16).toString(),
+                    });
```
ghost: Is there a risk of DoS here if we log the actual data sent? If we send a very long number, like, very big, shall we just log the length and not the actual buffer?
Same if the buffer has invalid characters, we might break the json-valid logs.

Same comment in other places when applicable.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,98 @@
...
+                chunk = chunk.subarray(toRead);
+                this.bytesToRead -= toRead;
+                if (this.bytesToRead === 0) {
+                    this.bytesToDiscard = 2;
+                }
+                continue;
+            }
+
+            const lineBreakIndex = chunk.indexOf('\r');
+            if (lineBreakIndex === -1) {
```
ghost, line 84: The length should be 10 bytes max, according to this function path, but if it's part of the same chunk (so NOT entering this `if`), I am able to send a very large number, like `500000000000000000000\r\n`; is that expected?

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: You're right, we were not enforcing the length restriction when the big field is found within the chunk. Fixing it.

lib/api/apiUtils/object/prepareStream.js
fredmnl: So I checked and when the trailing checksum mode is activated, the client does not send `md5` checksums. Therefore in this mode we do not check integrity. This is not a first, there are already ways to upload where `cloudserver` does not check payload integrity. In fact, it only checks payload integrity in two scenarios: `content-md5` header is provided, or streaming AuthV4 is used (in this case, the signature verification also indirectly verifies payload integrity).

In the absence of integrity check in cloudserver, integrity is still verified at the TCP level (although it's quite a weak verification) and at the HTTPS level. 

I have circled back with David to triple check that this is what we want from a product standpoint.

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: Correct, I fixed it, thanks :+1:

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: That's great! :+1:

tcarmet (approved):

tests/functional/raw-node/test/unsupportedChecksums.js
```diff
@@ -1,70 +0,0 @@
-const assert = require('assert');
```
tmacro: Unless an upcoming PR will be landing support for `STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER` I think we should keep this test case

tmacro (approved):

tests/functional/raw-node/test/unsupportedChecksums.js
fredmnl: Absolutely, not sure why I did that, probably got confused with the different headers. Good catch :+1:

lib/api/apiUtils/object/prepareStream.js
ghost: I see, if we all align, good to mention it in the release notes: integrity is not guaranteed by the S3 service with trailing checksums PUTs.

lib/api/apiUtils/object/prepareStream.js
```diff
@@ -24,11 +25,26 @@ function prepareStream(stream, streamingV4Params, log, errCb) {
...
+        v4Transform.headers = stream.headers;
         return v4Transform;
     }
     return stream;
 }
 
+function stripTrailingChecksumStream(stream, log) {
+    // don't do anything if we are not in the correct integrity check mode
+    if (stream.headers['x-amz-content-sha256'] !==
+        'STREAMING-UNSIGNED-PAYLOAD-TRAILER') {
```
ghost:
```suggestion
    if (stream.headers['x-amz-content-sha256'] !== `STREAMING-UNSIGNED-PAYLOAD-TRAILER') {
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,112 @@
...
+                chunk = chunk.subarray(toRead);
+                this.bytesToRead -= toRead;
+                if (this.bytesToRead === 0) {
+                    this.bytesToDiscard = 2;
+                }
+                continue;
+            }
+
+            // we are now looking for the chunk size field
+            // no need to look further than 10 bytes since the field cannot be bigger
```
ghost:
```suggestion
            // no need to look further than 10 bytes since the field cannot be bigger: the max
            // chunk size is 5GB (see constants.maximumAllowedPartSize)
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,112 @@
...
+            const lineBreakIndex = chunk.subarray(0, 10).indexOf('\r');
+            const bytesToKeep = lineBreakIndex === -1 ? chunk.byteLength : lineBreakIndex;
+            if (this.chunkSizeBuffer.byteLength + bytesToKeep > 10) {
+                this.log.error('chunk size field too big', {
+                    chunkSizeBuffer: this.chunkSizeBuffer.toString('hex'),
+                    truncatedChunk: chunk.subarray(0, 10).toString('hex'),
+                });
+                // if bigger, the chunk would be over 5 GB
+                // returning early to avoid a DoS by memory exhaustion
+                return callback(errors.InvalidArgument);
```
ghost, line 82: To return a standard error (see `lib/api/objectPutPart.js`) :

```suggestion
                return callback(errors.EntityTooLarge);
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            // we are now looking for the chunk size field
+            // no need to look further than 10 bytes since the field cannot be bigger
+            const lineBreakIndex = chunk.subarray(0, 10).indexOf('\r');
+            const bytesToKeep = lineBreakIndex === -1 ? chunk.byteLength : lineBreakIndex;
+            if (this.chunkSizeBuffer.byteLength + bytesToKeep > 10) {
+                this.log.error('chunk size field too big', {
+                    chunkSizeBuffer: this.chunkSizeBuffer.toString('hex'),
+                    truncatedChunk: chunk.subarray(0, 10).toString('hex'),
```
ghost: Better not to log fully the chunk if its size exceeds 10, as we then don't know how big it is. We may log one more byte and the total length to have some info.

```suggestion
                    chunkSizeBuffer: this.chunkSizeBuffer.subarray(0, 11).toString('hex'),
                    chunkSizeBufferLength: this.chunkSizeBuffer.length,
                    truncatedChunk: chunk.subarray(0, 10).toString('hex'),
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,112 @@
...
+            chunk = chunk.subarray(lineBreakIndex);
+
+            // chunk-size is sent in hex
+            const chunkSizeStr = this.chunkSizeBuffer.toString();
+            const dataSize = parseInt(chunkSizeStr, 16);
+            // we check that the parsing is correct (parseInt returns a partial parse when it fails)
+            if (isNaN(dataSize) || dataSize.toString(16) !== chunkSizeStr.toLowerCase()) {
+                this.log.error('invalid chunk size', { chunkSizeBuffer: chunkSizeStr });
+                return callback(errors.InvalidArgument);
+            }
```
ghost, line 100: I re-thought about it and I found that with this approach we cannot support leading zeros (like in `000019`). I am not sure this is a valid way to send the values, but leading zeros are common practice with hexa...

Here is a regex-free approach to support it (if needed!):

```suggestion
            const dataSize = parseInt(chunkSizeStr, 16);
            if (isNaN(dataSize)) {
                this.log.error('invalid chunk size', { chunkSizeBuffer: chunkSizeStr });
                return callback(errors.InvalidArgument);
            }
            const hexSize = dataSize.toString(16).toLowerCase();
            let leadingZeros = 0;
            while (leadingZeros < chunkSizeStr.length && chunkSizeStr[leadingZeros] === '0') {
                leadingZeros++;
            }
            if (dataSize === 0) {
                if (leadingZeros !== chunkSizeStr.length) {
                    this.log.error('invalid chunk size', { chunkSizeBuffer: chunkSizeStr });
                    return callback(errors.InvalidArgument);
                }
            } else if (hexSize.length + leadingZeros !== chunkSizeStr.length) {
                this.log.error('invalid chunk size', { chunkSizeBuffer: chunkSizeStr });
                return callback(errors.InvalidArgument);
            }
```

Here is a test that can trigger the problem:

```js
    it('should handle CR in data properly with leading zeros', done => {
        const trailingChecksumTransform = new TrailingChecksumTransform(log);
        const chunks = [
            Buffer.from('00000005\r\n'),         // Chunk size 5 with leading zeros
            Buffer.from('a\nb\nc'),             // 5 bytes of data: 'a', '\n', 'b', '\n', 'c'
            Buffer.from('\r\n0\r\nchecksum:xyz=\r\n') // CRLF, zero chunk, trailer
        ];
    
        const chunkedReader = new ChunkedReader(chunks);
        chunkedReader.pipe(trailingChecksumTransform);
        const outputChunks = [];
    
        trailingChecksumTransform.on('data', chunk => {
            console.log('Data:', chunk.toString());
            outputChunks.push(Buffer.from(chunk));
        });
    
        trailingChecksumTransform.on('finish', () => {
            const data = Buffer.concat(outputChunks).toString();
            console.log('Final Output:', data);
            assert.strictEqual(data, 'a\nb\nc');
            done();
        });
    
        trailingChecksumTransform.on('error', err => {
            console.log('Error:', err);
            done(err);
        });
    });
```

tests/functional/raw-node/test/trailingChecksums.js
```diff
@@ -0,0 +1,132 @@
...
+// note this is not the correct checksum in objDataWithTrailingChecksum
+const objDataWithTrailingChecksum = '10\r\n0123456789abcdef\r\n' +
+                                    '10\r\n0123456789abcdef\r\n' +
+                                    '0\r\nx-amz-checksum-crc64nvme:YeIDuLa7tU0=\r\n';
+const objDataWithoutTrailingChecksum = '0123456789abcdef0123456789abcdef';
+
+const authCredentials = {
+    accessKey: 'accessKey1',
+    secretKey: 'verySecretKey1',
+};
```
ghost: This should be available through `tests/functional/aws-node-sdk/test/support/awsConfig.js`

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -0,0 +1,112 @@
...
+                    this.bytesToDiscard = 2;
+                }
+                continue;
+            }
+
+            // we are now looking for the chunk size field
+            // no need to look further than 10 bytes since the field cannot be bigger
+            const lineBreakIndex = chunk.subarray(0, 10).indexOf('\r');
+            const bytesToKeep = lineBreakIndex === -1 ? chunk.byteLength : lineBreakIndex;
+            if (this.chunkSizeBuffer.byteLength + bytesToKeep > 10) {
```
ghost, line 74: Should we not only check the number of bytes in the buffer, but also the actual value, and ensure we don't exceed the max supported (5GB)? See `constants.maximumAllowedPartSize` for the constant you can use.

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: I thought about that as well but decided against it as the AWS doc, the `aws-cli` implementation and the Wikipedia article on chunked transfer ([https://en.wikipedia.org/wiki/Chunked_transfer_encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding)) all do not use leading zeros. I think we can forgo the additional bloat.

ghost: Sure, at least we now about it and we can improve if a client were to use it.

lib/auth/streamingV4/trailingChecksumTransform.js
fredmnl: I'd rather leave it as `Invalid` because we do not actually check what we read at this point. If the client sends garbage, he should receive `Invalid` rather than `EntityTooLarge`. Furthermore I don't think that `Invalid` is incorrect, it's just not necessarily that helpful which I don't think is so important here.

ghost: Yes ok, maybe we can customize the error description if needed. I'm closing this comment

ghost (approved):

ghost (commented):
Just a comment so we don't forget to bump cloudserver's version before merging, as we need it for the next release(s) :slightly_smiling_face:

PR #5761, opened 2025-03-14
CLDSRV-623: Bump arsenal for perf around error

BourgoisMickael:
Check by commit

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -1,5 +1,5 @@
 const { evaluators, actionMaps, RequestContext, requestUtils } = require('arsenal').policies;
-const { errors } = require('arsenal');
+const { errorInstances } = require('arsenal');
 const { parseCIDR, isValid } = require('ipaddr.js');
 const constants = require('../../../../constants');
 const { config } = require('../../../Config');
@@ -566,12 +566,12 @@ function validatePolicyConditions(policy) {
                 // AWS returns does not return an error if the condition starts with 'aws:'
                 // so we reproduce this behaviour
                 if (!validCondition && !conditionKey.startsWith('aws:')) {
-                    return errors.MalformedPolicy.customizeDescription('Policy has an invalid condition key');
+                    return errorInstances.MalformedPolicy.customizeDescription('Policy has an invalid condition key');
                 }
                 if (validCondition && validCondition.conditionValueTypeChecker) {
                     const conditionValueTypeError = validCondition.conditionValueTypeChecker(conditionValue);
                     if (conditionValueTypeError) {
-                        return errors.MalformedPolicy.customizeDescription(conditionValueTypeError);
+                        return errorInstances.MalformedPolicy.customizeDescription(conditionValueTypeError);
                     }
                 }
             }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "s3",
-  "version": "7.70.63",
+  "version": "7.70.64",
   "description": "S3 connector",
   "main": "index.js",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#7.70.35",
+    "arsenal": "git+https://github.com/scality/arsenal#7.70.39",
     "async": "~2.5.0",
     "aws-sdk": "2.905.0",
     "azure-storage": "^2.1.0",
--- a/lib/api/apiUtils/bucket/bucketCors.js
+++ b/lib/api/apiUtils/bucket/bucketCors.js
@@ -1,5 +1,5 @@
 const { parseString } = require('xml2js');
-const { errors, s3middleware } = require('arsenal');
+const { errors, errorInstances, s3middleware } = require('arsenal');
 
 const escapeForXml = s3middleware.escapeForXml;
 
@@ -82,7 +82,7 @@ const _validator = {
         const parsedValue = parseInt(seconds[0], 10);
         const errMsg = `MaxAgeSeconds "${seconds[0]}" is not a valid value.`;
         if (Number.isNaN(parsedValue) || parsedValue < 0) {
-            return errors.MalformedXML.customizeDescription(errMsg);
+            return errorInstances.MalformedXML.customizeDescription(errMsg);
         }
         return parsedValue;
     },
@@ -92,7 +92,7 @@ const _validator = {
     */
     validateNumberRules(length) {
         if (length > 100) {
-            return errors.InvalidRequest
+            return errorInstances.InvalidRequest
             .customizeDescription(customizedErrs.numberRules);
         }
         return true;
@@ -110,7 +110,7 @@ const _validator = {
         allowedMethods.length > 0) {
             return true;
         }
-        return errors.MalformedXML
+        return errorInstances.MalformedXML
             .customizeDescription(customizedErrs.originAndMethodExist);
     },
     /** _validator.validateMethods - check values of AllowedMethod's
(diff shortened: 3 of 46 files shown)


lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -508,26 +519,11 @@ function checkIp(value) {
         // we would accept different ip formats, which is not
         // standard in an AWS use case.
         try {
-            try {
-                parseCIDR(values[i]);
-            } catch (err) {
-                isValid(values[i]);
-            }
+            parseCIDR(values[i]);
```
BourgoisMickael, line 522: Check the comment on the commit.

This was running but any error was completely discarded.
Based on the comment, shouldn't we drop the parsing from ipaddr.js as we have stricted regex ?

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -507,6 +518,28 @@ function checkIp(value) {
...
+
+        Fixing this breaks the tests.
+
+        I'm not even sure of the purpose of using ipaddr.js
+        and using our regex after as our regex seems to be
+        more restrictive.
+
+        ipaddr.js accepts valid ip like:
+            - 0010.0xa5.1.1
+            - ::ffff:222.1.41.900
```
BourgoisMickael: An if instead of a catch for `isValid` should be used, but do we really need to use ipaddr.js ?

package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#7.70.35",
+    "arsenal": "git+https://github.com/scality/arsenal#user/mickael-improve",
```
ghost: Just a note so we don't forget to bump before merging

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -494,6 +494,17 @@ function validatePolicyResource(bucketName, policy) {
...
+// Apply the existing IP validation logic to each element
+function validateIpRegex(ip) {
+    if (constants.ipv4Regex.test(ip)) {
+        return ip.split('.').every(part => parseInt(part, 10) <= 255);
+    }
+    if (constants.ipv6Regex.test(ip)) {
+        return ip.split(':').every(part => part.length <= 4);
+    }
+    return false;
+}
```
ghost: We should unit test it when it's done

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -507,6 +518,28 @@ function checkIp(value) {
...
+        discards completely any error from ipaddr.parseCIDR.
+
+        Meaning in case of invalid ip there will be 2 ipaddr.js parsing,
+        reducing perf for an ignored error.
+
+        Fixing this breaks the tests.
+
+        I'm not even sure of the purpose of using ipaddr.js
+        and using our regex after as our regex seems to be
+        more restrictive.
```
ghost: From what I saw in https://github.com/scality/Arsenal/issues/133, the rationale was to support the ipv6 format used by node, convert it to ipv4 for further usage. The thing is, do we really want to restrict to ipv4 usage? E.g., in node 17, they started favor the use of ipv6 over ipv4. If we get IPv6 addresses as input, this may not work as expected...

lib/api/apiUtils/authorization/permissionChecks.js
ghost: I think we can try to drop it but we need to be careful that it doesn't break the IP conditions in IAM/Bucket policies. These policies can test ranges and mix of ipv4 & ipv6 addresses.

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -516,17 +549,16 @@ function checkIp(value) {
...
-            }
-            if (constants.ipv6Regex.test(ip)) {
-                return ip.split(':').every(part => part.length <= 4);
+        */
+        try {
+            parseCIDR(values[i]);
+        } catch (err) {
+            if (!isValid(values[i])) {
+                // eslint-disable-next-line no-console
+                console.log('=== INVALID IP ===', values[i], err);
```
ghost: Leftover from testing?

BourgoisMickael: yep just to print, I should drop the 2 last commits.
I just reverted the 2 last commits

lib/api/apiUtils/authorization/permissionChecks.js
BourgoisMickael: This was already there I just moved the arrow function out of the loop to avoid recreating it multiple time

lib/api/apiUtils/authorization/permissionChecks.js
BourgoisMickael: Note: this was implemented 1 year ago here: https://github.com/scality/cloudserver/pull/5501/files#diff-773344fcfc0f64145ff95eb6eccd556588c19449dc2c263a2532d4030bf00dffR490-R515

Since then the ipaddr.js parsing error has always been ignored due to the double try.

So removing it won't change the behavior but improve perf in case of invalid IP.

Now I don't know if we should handle special variations of ipv4 or ipv6 like it's done by ipaddr.js: https://www.npmjs.com/package/ipaddr.js/v/2.2.0?activeTab=code

And maybe our ipv6 regex is missing the cidr suffix: https://github.com/scality/cloudserver/blob/development/9.0/constants.js#L236-L237 ?

lib/api/apiUtils/authorization/permissionChecks.js
BourgoisMickael: This convertion is done in another place, here the `checkIp` function validates the format of the IP in the Bucket policy condition

lib/api/apiUtils/authorization/permissionChecks.js
ghost: > Now I don't know if we should handle special variations of ipv4 or ipv6 like it's done by ipaddr.js

We control the ip we get in the request, from the code we have in arsenal, so as long as we guarantee the right IP format is returned, we are fine. In AWS, the condition either supports ipv4 or ipv6: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-sourceip.
So I would say it's fine not to use ipaddr, as long as the regex is enough to ensure the lib will accept it in the actual logic (in Arsenal).

> And maybe our ipv6 regex is missing the cidr suffix

Indeed.

lib/api/apiUtils/authorization/permissionChecks.js
ghost: Yes, I meant, now that it's not inline but a function, we can test it independently :slightly_smiling_face:

lib/api/apiUtils/authorization/permissionChecks.js
BourgoisMickael: I'll drop the commit about ipCheck and have another ticket to compare with aws if the current regex is good and if we need to handle special ip

BourgoisMickael: Dropped and ticket is https://scality.atlassian.net/browse/CLDSRV-627

lib/api/apiUtils/authorization/permissionChecks.js
BourgoisMickael: Dropped and will be done in ticket https://scality.atlassian.net/browse/CLDSRV-627

ghost (approved):
LGTM
There is a merge commit that should be properly removed by rebasing the PR (it's hard to review with it)

BourgoisMickael:
> LGTM There is a merge commit that should be properly removed by rebasing the PR (it's hard to review with it)

@williamlardier I only see my changes on this merge commit, I don't see the other changes it brings in the review
image

ghost:
Maybe a github display issue, all good if we see only your changes :slightly_smiling_face:

dvasilas (approved):

PR #5762, opened 2025-03-18
Support updating Account info in PutMetadata API

welansari:
- Update Account info when the account id is provided. In S3C, Backbeat is the one that calling this API and replacing the info in the MD before sending them, but in Zenko we have to delegate this task to the replication destination's Cloudserver as Vault Admin APIs are not accessible externally.
- Passed an additional flag to the MongoClient's PutMetadata to update the master to the latest version each time we put an object.
- Re-enabled BackbeatRoutes CRR tests and added additional tests to cover the changes made.
- Generate coverage in functional tests

Issue: CLDSRV-618

### DIFF
--- a/.github/docker/docker-compose.yaml
+++ b/.github/docker/docker-compose.yaml
@@ -1,13 +1,28 @@
 services:
   cloudserver:
     image: ${CLOUDSERVER_IMAGE}
-    command: sh -c "yarn start > /artifacts/s3.log"
+    command: |
+      bash -c "
+        # Using tini to handle signals properly
+        tini -s -g -- npx nyc --clean --silent yarn start > /artifacts/s3.log &
+        PID=$$!
+        generate_coverage() {
+          echo 'Stopping NodeJS processes...'
+          kill -TERM $$PID 2>/dev/null || true
+          wait $$PID
+          echo 'Generating coverage report...'
+          npx nyc report --report-dir /coverage/test --reporter=lcov --reporter=text-summary
+        }
+        trap generate_coverage SIGTERM
+        wait $$PID
+      "
     network_mode: "host"
     volumes:
       - /tmp/ssl:/ssl
       - /tmp/ssl-kmip:/ssl-kmip
       - ${HOME}/.aws/credentials:/root/.aws/credentials
       - /tmp/artifacts/${JOB_NAME}:/artifacts
+      - /tmp/coverage/${JOB_NAME}:/coverage/test
     environment:
       - CI=true
       - ENABLE_LOCAL_CACHE=true
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -253,6 +253,14 @@ jobs:
         yarn run multiple_backend_test | tee /tmp/artifacts/${{ github.job }}/tests.log
       env:
         S3_LOCATION_FILE: tests/locationConfig/locationConfigTests.json
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: >
+          sproxyd
+          mongo
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
@@ -291,6 +299,12 @@ jobs:
         yarn run ft_test | tee /tmp/artifacts/${{ github.job }}/tests.log
       env:
         S3_LOCATION_FILE: tests/locationConfig/locationConfigTests.json
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: mongo
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
@@ -331,6 +345,12 @@ jobs:
         yarn run ft_mixed_bucket_format_version | tee /tmp/artifacts/${{ github.job }}/mixed-tests.log
       env:
         S3_LOCATION_FILE: tests/locationConfig/locationConfigTests.json
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: mongo
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
@@ -374,6 +394,12 @@ jobs:
         set -o pipefail;
         bash wait_for_local_port.bash 8000 40
         yarn run ft_test | tee /tmp/artifacts/${{ matrix.job-name }}/tests.log
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+        flags: ${{ matrix.job-name }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
@@ -408,6 +434,11 @@ jobs:
         set -ex -o pipefail;
         bash wait_for_local_port.bash 8000 40
         yarn run test_utapi_v2 | tee /tmp/artifacts/${{ github.job }}/tests.log
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
@@ -452,6 +483,13 @@ jobs:
         set -ex -o pipefail;
         bash wait_for_local_port.bash 8000 40
         yarn run test_quota | tee /tmp/artifacts/${{ github.job }}/tests.log
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: mongo
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+        flags: ${{ github.job }}${{ matrix.inflights.value == 'true' && '-inflights' || '' }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
@@ -490,6 +528,12 @@ jobs:
         bash wait_for_local_port.bash 8000 40
         bash wait_for_local_port.bash 5696 40
         yarn run ft_kmip | tee /tmp/artifacts/${{ github.job }}/tests.log
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: pykmip
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
@@ -568,6 +612,12 @@ jobs:
         S3BACKEND: file
         S3VAULT: mem
         S3METADATA: mongodb
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: ceph
+        codecov-token: ${{ secrets.CODECOV_TOKEN }}
+      if: always()
     - name: Upload logs to artifacts
       uses: scality/action-artifacts@v4
       with:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -229,6 +229,7 @@ jobs:
       S3BACKEND: mem
       S3_LOCATION_FILE: /usr/src/app/tests/locationConfig/locationConfigTests.json
       S3DATA: multiple
+      S3METADATA: mongodb
       JOB_NAME: ${{ github.job }}
     steps:
     - name: Checkout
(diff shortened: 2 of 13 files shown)


francoisferrand:
> Update Account info when the account id is provided. In S3C, Backbeat is the one that calling this API and replacing the info in the MD before sending them, but in Zenko we have to delegate this task to the replication destination's Cloudserver as Vault Admin APIs are not accessible externally.

is there a reason to keep the existing S3C behavior (i.e. mapping done in backbeat, on the source), or should we move S3C to this new approach as well?

lib/routes/routeBackbeat.js
```diff
@@ -578,62 +619,86 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // To prevent this, the versionId field is only included in options when it is defined.
         if (versionId !== undefined) {
             options.versionId = versionId;
+            // In the MongoDB metadata backend, setting the versionId option leads to the creation
+            // or update of the version object, the master object is only updated if its versionId
+            // is the same as the version. This can lead to inconsistencies when replicating objects
+            // in the wrong order (can happen when retrying after failures), where the master object
+            // might point to a non current version. To prevent this, we need to set repairMaster to
+            // true which will update the master to the latest version available at the time of the put.
+            options.repairMaster = true;
```
francoisferrand: `putMetadata` is used for many things, and often (in lifecycle) to only _"update"_ metadata : ideally we should avoing triggering repairMaster in that case...
(note: we will probably also need to be aware of this difference to fix the metadata-update-race-condition)

lib/routes/routeBackbeat.js
```diff
@@ -578,62 +619,86 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
 
-        log.trace('putting object version', {
-            objectKey: request.objectKey, omVal, options });
-        return metadata.putObjectMD(bucketName, objectKey, omVal, options, log,
-            (err, md) => {
+        return async.series([
+            // Zenko's Backbeat delegates replacing the account
+            // information to the destination's Cloudserver, as
+            // Vault admin APIs are not exposed externally.
+            next => getCanonicalIdsByAccountId(request.query?.accountId, log, (err, res) => {
```
francoisferrand: similarly: updating the account Id is not needed in case the request is coming from a remote CRR, and useless when used internally → should not degrade performance in that case.

if there is no better way, we could add an extra option on the query, defaulting to false and set to true only by CRR, to request this mapping?

lib/routes/routeBackbeat.js
francoisferrand: or maybe (by some luck) in case we really just _replace_ the metadata of an object, repairing is not triggered (in the implementation of MongoClientInterface) ?

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -751,7 +752,7 @@ describe('backbeat routes', () => {
             });
         });
 
-        it('should create a new null version if versioning suspended and delete marker null version', done => {
+        it.skip('should create a new null version if versioning suspended and delete marker null version', done => {
```
francoisferrand, line 755: remember not to commit this

francoisferrand:
should this target 9.0, or a (to be created before merging) 9.1 ?

lib/routes/routeBackbeat.js
welansari: We have access to the previous md here so we can check if we are only updating or creating a new version.

lib/routes/routeBackbeat.js
```diff
@@ -464,6 +464,47 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
+ * 
+ * @param {string} accountId - account ID
+ * @param {Log} log - logger instance
+ * @param {CanonicalIdCallback} cb - callback function
+ * @returns {undefined}
+ */
+function getCanonicalIdsByAccountId(accountId, log, cb) {
+    if (!accountId) {
+        return cb();
+    }
```
ghost: Should we add a debug log here?

lib/routes/routeBackbeat.js
```diff
@@ -464,6 +464,47 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
+            return cb(err);
+        }
+        if (res.length === 0) {
+            log.error('account ID not found', {
+                accountId,
+                method: 'getCanonicalIdsByAccountIds',
+            });
+            return cb(errors.AccountNotFound);
+        }
+        return cb(null, res[0]);
```
ghost, line 501: To avoid having to deal with arrays, Vault supports the `getAccount` API. This API accepts any info to uniquely identity an account, including the name, email, ID, and canonical id.
See https://github.com/scality/vault2/blob/development/8.10/lib/Handlers/Account.js#L1041

If we only use one account at a time, preefer using this API. This will also simplify handling in Arsenal.

.github/workflows/tests.yaml
```diff
@@ -451,6 +507,17 @@ jobs:
...
+    - name: Stop CI services
+      run: docker compose --profile mongo down
+      working-directory: .github/docker
+    - name: Upload coverage reports to Codecov
+      uses: codecov/codecov-action@v5
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        directory: /tmp/coverage/${{ github.job }}
+        flags: ${{ github.job }}${{ matrix.inflights.value == 'true' && '-inflights' || '' }}
+      if: ${{ !cancelled() }}
```
ghost: Worth creation a local action to avoid duplication this

lib/routes/routeBackbeat.js
welansari: > updating the account Id is not needed in case the request is coming from a remote CRR, and useless when used internally → should not degrade performance in that case.

The `accountId` is a new parameter only passed by CRR, so this shouldn't affect performance of this API in other features.

I don't think there's a need for an extra parameter here, `accountId`'s only purpose is this, so CRR should just not pass it in S3C (or pass it if we decide to update both Cloudserver and Backbeat to the 9.0 and abandon the part where it updates the account info).

tests/multipleBackend/routes/routeBackbeat.js
welansari: This was skipped on purpose and will be investigate along with the other skipped tests in CLDSRV-626.
The failures in these tests was triggered after switching the metadata backend from memory to mongodb.

lib/routes/routeBackbeat.js
francoisferrand: Is it safe enough? i.e. could this introduce some subtle issues in case there is race condition between an update and a creation?

In other words: in the code, in many -if not all- cases we know our intent and are sure if we want to "update an existing one" or "create a new revision" (which may mean replacing the document, in non-versioned case). Trying to recover that simple information, like what is done in MongClientInterface, may work, but creates a risk to introduce more complexity and corner cases...

lib/routes/routeBackbeat.js
francoisferrand: so we need to condition the call to `getCanonicalIdsByAccountId`, and do it only if `accountId` is specified?

lib/routes/routeBackbeat.js
francoisferrand: should be moved out of this function: for readability, best to keep the logic of "skipping" the call at the caller's.... where it is also easier to answer our question (_is a log needed?_)
→ by moving this condition to `putMetadata()`, it will be clear we only do it if the account is provided (possibly with a comment explaining in which case it happens)

(if we really need to keep the test here, then we should try and make it clear the function does not always get the canonicalId : e.g. something like `maybeGetCanonicalIdsByAccountId`. At leat it gives an indication to the reader that it may not be done, but it still forces to look through the abstraction, and hence IMHO not as readable...)

lib/routes/routeBackbeat.js
welansari: The condition was already there inside `getCanonicalIdsByAccountId`, but looking at your other comment, sure i'll move it a level above.

lib/routes/routeBackbeat.js
welansari: `getAccount` is an authenticated API, The Vault instance we have here don't have access to vault admin credentials.

lib/routes/routeBackbeat.js
welansari: Should be safe enough. After getting the latest version, the master is only updated if it has a greater or equal versionId => points to the same or older version.

welansari:
> is there a reason to keep the existing S3C behavior (i.e. mapping done in backbeat, on the source), or should we move S3C to this new approach as well?

There's nothing keeping us from doing it aside from updating both Cloudserver and Backbeat in S3C.

> should this target 9.0, or a (to be created before merging) 9.1 ?

No need to do as no breaking change was introduced. `PutMetadata` is only used to update existing objects in Zenko so the `repairMaster` flag will never be used.

lib/routes/routeBackbeat.js
```diff
@@ -578,62 +616,99 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
-                        bucketName,
-                        objectKey,
-                    });
-                    async.eachLimit(objMd.location, 5,
-                    (loc, next) => dataWrapper.data.delete(loc, log, err => {
+                return getCanonicalIdsByAccountId(request.query.accountId, log, (err, res) => {
+                    if (err) {
+                        return next(err);
+                    }
+                    if (res) {
```
ghost: Isn't `res` always set as per the checks in `getCanonicalIdsByAccountId`? Then this `if` is not needed.

By the way, shouldn't `getCanonicalIdsByAccountId` be a method in arsenal, more than a function in the backbeat routes? In case we need this primitive from other places later...

lib/routes/routeBackbeat.js
```diff
@@ -578,62 +616,99 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                            return next(err);
+                        }
+                        pushReplicationMetric(objMd, omVal, bucketName, objectKey, log);
+                        if (objMd &&
+                            headers['x-scal-replication-content'] !== 'METADATA' &&
+                            versionId &&
+                            // The new data location is set to null when archiving to a Cold site.
+                            // In that case "removing old data location key" is handled by the lifecycle
+                            // transition processor.
+                            omVal.location && Array.isArray(omVal.location) &&
```
ghost, line 677: Can be simplified?

```suggestion
                            Array.isArray(omVal.location) &&
```

lib/routes/routeBackbeat.js
```diff
@@ -578,62 +616,99 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                            locationKeysHaveChanged(objMd.location, omVal.location)) {
+                            log.info('removing old data locations', {
+                                method: 'putMetadata',
                                 bucketName,
                                 objectKey,
-                                locationKey: loc,
-                                error: err.message,
+                            });
+                            async.eachLimit(objMd.location, 5,
+                            (loc, nextEach) => dataWrapper.data.delete(loc, log, err => {
```
ghost, line 685: Not related to your changes but IIRC there is the `data.batchDelete` method from the data wrapper, that might help simplifying your function (not blocking, just to share)

ghost (approved):

BourgoisMickael (approved):

.github/docker/docker-compose.yaml
```diff
@@ -1,13 +1,28 @@
...
+      PID=$$!
+      generate_coverage() {
+        echo 'Stopping NodeJS processes...'
+        kill -TERM $$PID 2>/dev/null || true
+        wait $$PID
+        echo 'Generating coverage report...'
+        npx nyc report --report-dir /coverage/test --reporter=lcov --reporter=text-summary
+      }
+      trap generate_coverage SIGTERM
+      wait $$PID
```
francoisferrand: nit indent to make is visible this is inside the bash command string:
```suggestion
        # Using tini to handle signals properly
        tini -s -g -- npx nyc --clean --silent yarn start > /artifacts/s3.log &
        PID=$$!
        generate_coverage() {
          echo 'Stopping NodeJS processes...'
          kill -TERM $$PID 2>/dev/null || true
          wait $$PID
          echo 'Generating coverage report...'
          npx nyc report --report-dir /coverage/test --reporter=lcov --reporter=text-summary
        }
        trap generate_coverage SIGTERM
        wait $$PID
```

.github/workflows/tests.yaml
```diff
@@ -252,6 +253,12 @@ jobs:
         yarn run multiple_backend_test | tee /tmp/artifacts/${{ github.job }}/tests.log
       env:
         S3_LOCATION_FILE: tests/locationConfig/locationConfigTests.json
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: 'sproxyd mongo'
```
francoisferrand: nit: smartish yaml way to do the same and look like it is regular yaml list :)
```suggestion
        profiles: >
            sproxyd
            mongo
```

.github/workflows/tests.yaml
```diff
@@ -290,6 +297,12 @@ jobs:
         yarn run ft_test | tee /tmp/artifacts/${{ github.job }}/tests.log
       env:
         S3_LOCATION_FILE: tests/locationConfig/locationConfigTests.json
+    - name: Cleanup and upload coverage
+      uses: ./.github/actions/cleanup-and-coverage
+      with:
+        profiles: 'mongo'
```
francoisferrand: quotes not needed

lib/routes/routeBackbeat.js
francoisferrand: it is already reusable: the only thing this wrapper adds is logs, which is not reusable IMO and specific to this use case (or could be inlined)

+1 check on `res` should not be needed

francoisferrand (approved):

PR #5766, opened 2025-03-19
Support miria location

francoisferrand:
- Add tlp location type
- Remove dmf and azure_archive location types

Issue: CLDSRV-624

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -300,103 +300,9 @@ function hdClientLocationConstraintAssert(configHd) {
     return hdclientFields;
 }
 
-function azureArchiveLocationConstraintAssert(locationObj) {
-    const checkedFields = [
-        'azureContainerName',
-        'azureStorageEndpoint',
-    ];
-    if (Object.keys(locationObj.details).length === 0 ||
-        !checkedFields.every(field => field in locationObj.details)) {
-        return;
-    }
-    const {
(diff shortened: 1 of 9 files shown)


welansari (approved):

lib/Config.js
```diff
@@ -467,6 +373,15 @@ function locationConstraintAssert(locationConstraints) {
...
                 'bad config: credentials must include secretKey as string');
         }
+
+        if (locationConstraints[l].type === 'tlp') {
+            // should have no detail
+            assert(Object.keys(locationConstraints[l].details).length === 0);
+
+            // should be a cold location
+            assert(locationConstraints[l].isCold === true);
+        }
```
ghost, line 383: Why do we stop validating the locations here? is it because we rely on the validation at the location creation time? If yes, can we put a comment?

ghost (approved):

lib/Config.js
francoisferrand: what do you mean "stop" validating locations?
* we keep the whole validation, there is no early exit
* we do not store any details anymore (as they are useless for cloudserver, and contain secrets), so the validation of these details is now removed, and we now check that `details` is indeed empty
* the block was just moved early to simplify testing no details are provided (since default fields are added later in the process)

PR #5773, opened 2025-03-24
Improvement/cldsrv 610 re enable utapi

benzekrimaha:
_Please note that the coverage is reported because we've enabled already existing code parts that were disabled when bumping dependencies after which codecov was enabled_

Issue: CLDSRV-610

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -20,11 +20,7 @@ const {
     allowedUtapiEventFilterStates,
     supportedLifecycleRules,
 } = require('../constants');
-
-// TODO CLDSRV-610: re-enable utapi
-// const { utapiVersion } = require('utapi');
-const utapiVersion = null;
-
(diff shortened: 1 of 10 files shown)


lib/Config.js
francoisferrand, line 1: (random location)

the commit does not look like a `git revert` : is there any difference, or anything preventing from using `git revert`?

Using git revert is the better way to rollback changes, as it gives visibility on the intent and links to the original commit, while helping to make sure we do not forget anything (which may not have been "flagged" in the code)

benzekrimaha: Indeed , done here : https://github.com/scality/cloudserver/pull/5773/commits/73423e7e3cf0d97f96eabe8478f935f05f23a172

ghost (approved):

anurag4DSB (approved):

PR #5777, opened 2025-04-01
Improvement/cldsrv 631

ghost:
Use SUR backend instead of bucket MD to get the utilization metrics. This makes the Veeam SOSAPI cronjob obsolete: it will update the metrics, but this won't affect the API anymore. It means we can safely remove it later.


lib/routes/veeam/get.js
```diff
@@ -25,21 +26,44 @@ function getVeeamFile(request, response, bucketMd, log) {
...
-        // Extract the last modified date, but do not include it when computing
-        // the file's ETag (md5)
-        modified = fileToBuild.value.LastModified;
-        delete fileToBuild.value.LastModified;
+        const finalizeRequest = bucketMetrics => {
+            const fileToBuild = getFileToBuild(request, data._capabilities?.VeeamSOSApi);
+            if (fileToBuild.error) {
+                return responseXMLBody(fileToBuild.error, null, response, log);
+            }
+            let modified = new Date().toISOString();
```
welansari: Just for my understanding, what's the point of having this line here since we overwrite it immediatly after ?

lib/routes/veeam/get.js
```diff
@@ -25,21 +26,44 @@ function getVeeamFile(request, response, bucketMd, log) {
...
+            delete fileToBuild.value.LastModified;
+
+            // The SOSAPI metrics are dynamically computed using the SUR backend.
+            if (bucketMetrics && !fileToBuild.value.CapacityInfo?.Used) {
+                fileToBuild.value.CapacityInfo.Used = Number(bucketMetrics.bytesTotal);
+                fileToBuild.value.CapacityInfo.Available =
+                    Number(fileToBuild.value.CapacityInfo.Capacity) - Number(bucketMetrics.bytesTotal);
+                // TODO when SUR backend supports realtime metrics: enable this, it will report the
+                // real last modified date of the bucket, instead of the current date, ensuring
+                // no issue in a SOSAPI context.
```
welansari: should have a ticket linked to this ?

tests/sur/routeVeeam.js
```diff
@@ -0,0 +1,582 @@
...
+                });
+        });
+        afterEach(done => {
+            bucketUtil.empty(TEST_BUCKET)
+                .then(() => s3.deleteBucket({ Bucket: TEST_BUCKET }).promise())
+                .then(() => done());
+        });
+
+        [
+            // ['.system-d26a9498-cb7c-4a87-a44a-8ae204f5ba6c/system.xml', testSystem, testSystemMd5],
```
welansari: was this intentional ?

tests/sur/routeVeeam.js
```diff
@@ -0,0 +1,582 @@
...
+        path: `/_/veeam/${bucket}/${objectKey}`,
+        urlForSignature: `/${bucket}/${objectKey}`,
+        requestBody,
+        jsonResponse: false,
+        queryObj,
+    };
+    makeRequest(options, callback);
+}
+
+(process.env.S3METADATA === 'mongodb' ? describe.only : describe.skip)('Veeam routes:', () => {
```
welansari:
```suggestion
(process.env.S3METADATA === 'mongodb' ? describe : describe.skip)('Veeam routes:', () => {
```

lib/routes/veeam/get.js
ghost: Not yet, we are not done with the design...

tests/sur/routeVeeam.js
ghost: no :slightly_smiling_face:

lib/routes/veeam/get.js
ghost: It was useful at first because we could have no date, but a commit in between changed the logic but didn't simplify this line.

lib/routes/veeam/get.js
ghost: I added one, but it might change or we might forget it for some time, but at least we have a ticket number with context...

welansari (approved):

lib/routes/veeam/get.js
```diff
@@ -25,21 +26,44 @@ function getVeeamFile(request, response, bucketMd, log) {
...
+                // modified = new Date(bucketMetrics.date).toISOString();
+            }
+
+            const builder = new xml2js.Builder({
+                headless: true,
+            });
+            return respondWithData(request, response, log, data,
+                buildHeadXML(builder.buildObject(fileToBuild.value)), modified);
+        };
+        if (!isSystemXML(request.objectKey)) {
```
welansari, line 56: Should we also check that the scuba backend is enabled ?

ghost: Practically this won't change much: in this branch and the branch before, scuba is always deployed. Inflight quota service is enabled based on the scuba presence in the config. here, having no scuba would just cause an error we already handle. We do not want to default to zero metrics when there is no Scuba, so we can only return a more specific error.

lib/routes/veeam/get.js
```diff
@@ -25,21 +26,44 @@ function getVeeamFile(request, response, bucketMd, log) {
...
-        });
-        return respondWithData(request, response, log, data,
-            buildHeadXML(builder.buildObject(fileToBuild.value)), modified);
+            // The SOSAPI metrics are dynamically computed using the SUR backend.
+            if (bucketMetrics && !fileToBuild.value.CapacityInfo?.Used) {
+                fileToBuild.value.CapacityInfo.Used = Number(bucketMetrics.bytesTotal);
+                fileToBuild.value.CapacityInfo.Available =
+                    Number(fileToBuild.value.CapacityInfo.Capacity) - Number(bucketMetrics.bytesTotal);
+                // TODO CLDSRV-633 when SUR backend supports realtime metrics: enable this, it will
+                // report the real last modified date of the bucket, instead of the current date,
```
francoisferrand: we don't want to report the `last modified date of the bucket`, as it would indicate stale data if the bucket has not been modified in a long time: we probably want to report the `date of the last "scan" of the bucket`, i.e. (when trailing oplog) something like the date of the last metadata operation which has been computed in the utilization metrics.

tests/sur/routeVeeam.js
```diff
@@ -0,0 +1,582 @@
...
+    afterEach(() => {
+        scuba.stop();
+    });
+
+    describe('veeam PUT routes:', () => {
+        before(done => {
+            bucketUtil = new BucketUtility(
+                'default', { signatureVersion: 'v4' });
+            s3 = bucketUtil.s3;
+            s3.createBucket({ Bucket: TEST_BUCKET }).promise()
```
francoisferrand, line 151: shouldn't we use await and stop using callbacks?

ghost: Yes, changing the bucket util is likely a long work that I would prefer doing separately as we use it in many test files...

lib/routes/veeam/get.js
ghost: Yes I meant the current cseq (not for a specific bucket: if the bucket is not updated, the "date of the last scan of the bucket" will never happen, unless we repair). I'll update the ticket, the comment remains fine to me.

tests/sur/routeVeeam.js
francoisferrand: sorry, i missed the rename of the file... 
(but i was not speaking so moch of changing bucket util, just changing the tests to drop the callback and use async functions in there, and await instead of `.promise().then()`)

lib/routes/veeam/get.js
francoisferrand: i think the comment may be misleading for someone discovering the issue here: if we update the ticket, we might as well update the comment to avoid any confusion...

francoisferrand (approved):
should we target 9.0 or 9.1 ?

esp. if there is some "dependency" (direct/explicit or not) on other changes, may be best to merge in 9.1. to avoid any issue and better control when we bump.

PR #5779, opened 2025-04-01
OS-934 create dependabot PRs on default branch only

tcarmet:
Part of our retro action at Hopo, is to stop targetting old branches for dependabot and focus only on the latest or default branch.


jonathan-gramain (approved):

BourgoisMickael (approved):

BourgoisMickael:
If we can have once per week instead of day

PR #5788, opened 2025-04-11
CLDSRV-636: SSE with both internal & external KMS (HF 9.2.0.12)

BourgoisMickael:
This is generic but simplified for file to aws use case and will be extended in latest version to handle any backend.

- Store all KMS keys with an arn describing backend provider (see https://github.com/scality/Arsenal/pull/2349)
  - option (`hideScalityArn`) to keep backward compatibility
 - Migrate KMS Keys with `sseMigration` config
   - For bucket, replace `file` key with `aws` key
   - For object reformat key into an arn
- Allow usage of keys from previous KMS provider (`file`) to decript and read files (or for ongoing MPU)

### Overall example of a new Key Format and migration

```mermaid
flowchart TD
    GetObject:::red --> bktArn
    objArn --NO GetObject--> objUpdate[Update Object: Prefix KMS key with arn]:::red
    %% Get 0-1 Links

    %% Common route
    bktArn{Bucket SSE is arn?}:::purple
        --NO--> bktKey["
            Create new external KMS Key with arn prefix
            Update Bucket attribute
        "]:::purple
    bktArn
        --Then--> objArn{Object SSE is arn?}:::purple
        --Then--> ...:::purple
        --> KMS
    subgraph KMS[KMS#nbsp;Object#nbsp;Decrypt/Encrypt]
        extract[Extract KMS provider and KeyId from arn]:::purple
        --> client[Pick or instantiate KMS client based on provider from Key arn]:::purple
        --> send["
            Backend Decrypts/Encrypts with KeyId
            But if Arn is sent, client still extract KeyId from arn
        "]:::purple
    end
    %% 2-7 Links common

    PutObject:::blue --> bktArn
    objArn --NO PutObject--> PutPrefix[Prefix KMS Key with arn if missing]:::blue
    %% PUT 8-9 Links

    classDef red stroke:red
    classDef blue stroke:blue
    classDef purple stroke:purple

    linkStyle 0,1 stroke:red
    linkStyle 8,9 stroke:blue
    linkStyle 2,3,4,5,6,7 stroke:purple
```

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -17,6 +17,7 @@ const { azureAccountNameRegex, base64Regex,
 } = require('../constants');
 const { utapiVersion } = require('utapi');
 const { versioning } = require('arsenal');
+const { KmsType, KmsProtocol, isValidProvider } = require('arsenal/build/lib/network/KMSInterface');
 
 const versionIdUtils = versioning.VersionID;
 
@@ -411,13 +412,19 @@ class Config extends EventEmitter {
         }
         let kmsAWS = {};
 
-        const { region, endpoint, ak, sk, tls } = config.kmsAWS;
+        const { providerName, region, endpoint, ak, sk, tls, noAwsArn } = config.kmsAWS;
 
+        assert(providerName, 'Configuration Error: providerName must be defined in kmsAWS');
+        assert(isValidProvider(providerName),
+            'Configuration Error: kmsAWS.providerNamer must be lowercase alphanumeric only');
         assert(endpoint, 'Configuration Error: endpoint must be defined in kmsAWS');
         assert(ak, 'Configuration Error: ak must be defined in kmsAWS');
         assert(sk, 'Configuration Error: sk must be defined in kmsAWS');
+        assert(['undefined', 'boolean'].some(type => type === typeof noAwsArn),
+            'Configuration Error:: kmsAWS.noAwsArn must be a boolean or not set');
 
         kmsAWS = {
+            providerName,
             endpoint,
             ak,
             sk,
@@ -425,6 +432,10 @@ class Config extends EventEmitter {
 
         if (region) {
             kmsAWS.region = region;
+        }
+
+        if (noAwsArn) {
+            kmsAWS.noAwsArn = noAwsArn;
         }
 
         if (tls) {
@@ -1095,8 +1106,12 @@ class Config extends EventEmitter {
 
         this.kms = {};
         if (config.kms) {
+            assert(config.kms.providerName, 'config.kms.providerName must be provided');
+            assert(isValidProvider(config.kms.providerName),
+                'config.kms.providerName must be lowercase alphanumeric only');
             assert(typeof config.kms.userName === 'string');
             assert(typeof config.kms.password === 'string');
+            this.kms.providerName = config.kms.providerName;
             this.kms.userName = config.kms.userName;
             this.kms.password = config.kms.password;
             if (config.kms.helperProgram !== undefined) {
@@ -1163,6 +1178,10 @@ class Config extends EventEmitter {
             },
         };
         if (config.kmip) {
+            assert(config.kmip.providerName, 'config.kmip.providerName must be defined');
+            assert(isValidProvider(config.kmip.providerName),
+                'config.kmip.providerName must be lowercase alphanumeric only');
+            this.kmip.providerName = config.kmip.providerName;
             if (config.kmip.client) {
                 if (config.kmip.client.compoundCreateActivate) {
                     assert(typeof config.kmip.client.compoundCreateActivate ===
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -17,7 +17,13 @@ const { azureAccountNameRegex, base64Regex,
 } = require('../constants');
 const { utapiVersion } = require('utapi');
 const { versioning } = require('arsenal');
-const { KmsType, KmsProtocol, isValidProvider } = require('arsenal/build/lib/network/KMSInterface');
+const {
+    KmsType,
+    KmsProtocol,
+    isValidProvider,
+    isValidType,
+    isValidProtocol,
+} = require('arsenal/build/lib/network/KMSInterface');
 
 const versionIdUtils = versioning.VersionID;
 
@@ -402,8 +408,9 @@ class Config extends EventEmitter {
 
         // Read config automatically
         this._getLocationConfig();
-        this._getConfig();
+        const config = this._getConfig();
         this._configureBackends();
+        this._sseMigration(config);
     }
 
     _parseKmsAWS(config) {
@@ -1434,6 +1441,7 @@ class Config extends EventEmitter {
         }
 
         this.lifecycleRoleName = config.lifecycleRoleName || null;
+        return config;
     }
 
     _configureBackends() {
@@ -1507,6 +1515,61 @@ class Config extends EventEmitter {
             metadata,
             kms,
         };
+    }
+
+    _sseMigration(config) {
+        if (config.sseMigration) {
+            /**
+             * For data that was encrypted internally by default and a new external provider is setup.
+             * This config helps detect the existing encryption key to decrypt with the good provider.
+             * The key format will be migrated automatically on GET/HEADs to include provider details.
+             */
+            this.sseMigration = {};
+            const { previousKeyType, previousKeyProtocol, previousKeyProvider } = config.sseMigration;
+            if (!previousKeyType) {
+                assert.fail(
+                    'NotImplemented: No dynamic KMS key migration. Set sseMigration.previousKeyType');
+            }
+
+            // If previousKeyType is provided it's used as static value to migrate the format of the key
+            // without additional dynamic evaluation if the key provider is unknown.
+            assert(isValidType(previousKeyType),
+                'ssenMigration.previousKeyType must be "internal" or "external"');
+            this.sseMigration.previousKeyType = previousKeyType;
+
+            let expectedProtocol;
+            if (previousKeyType === KmsType.internal) {
+                // For internal key type default protocol is file and provider is scality
+                this.sseMigration.previousKeyProtocol = previousKeyProtocol || KmsProtocol.file;
+                this.sseMigration.previousKeyProvider = previousKeyProvider || 'scality';
+                expectedProtocol = [KmsProtocol.scality, KmsProtocol.mem, KmsProtocol.file];
+            } else if (previousKeyType === KmsType.external) {
+                // No defaults allowed for external provider
+                assert(previousKeyProtocol,
+                    'sseMigration.previousKeyProtocol must be defined for external provider');
+                this.sseMigration.previousKeyProtocol = previousKeyProtocol;
+                assert(previousKeyProvider,
+                    'sseMigration.previousKeyProvider must be defined for external provider');
+                this.sseMigration.previousKeyProvider = previousKeyProvider;
+                expectedProtocol = [KmsProtocol.kmip, KmsProtocol.aws_kms];
+            }
+
+            assert(isValidProtocol(previousKeyType, this.sseMigration.previousKeyProtocol),
+                `sseMigration.previousKeyProtocol must be one of ${expectedProtocol}`);
+            assert(isValidProvider(previousKeyProvider),
+                'sseMigration.previousKeyProvider must be lowercase alphanumeric only');
+
+            if (this.sseMigration.previousKeyType === KmsType.external) {
+                if ([KmsProtocol.file, KmsProtocol.mem].includes(this.backends.kms)) {
+                    assert.fail(
+                        `sseMigration.previousKeyType "external" can't migrate to "internal" KMS provider ${
+                            this.backends.kms}`
+                    );
+                }
+                // We'd have to compare protocol & providerName
+                assert.fail('sseMigration.previousKeyType "external" is not yet available');
+            }
+        }
     }
 
     _verifyRedisPassword(password) {
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1248,6 +1248,11 @@ class Config extends EventEmitter {
         assert(typeof this.defaultEncryptionKeyPerAccount === 'boolean',
             'config.defaultEncryptionKeyPerAccount must be a boolean');
 
+        this.kmsHideScalityArn = Object.hasOwnProperty.call(config, 'kmsHideScalityArn')
+            ? config.kmsHideScalityArn
+            : true; // By default hide scality arn to keep backward compatibility and simplicity
+        assert.strictEqual(typeof this.kmsHideScalityArn, 'boolean');
+
         this.healthChecks = defaultHealthChecks;
         if (config.healthChecks && config.healthChecks.allowFrom) {
             assert(config.healthChecks.allowFrom instanceof Array,
--- a/lib/api/apiUtils/bucket/updateEncryption.js
+++ b/lib/api/apiUtils/bucket/updateEncryption.js
@@ -1,0 +1,141 @@
+const { getVersionSpecificMetadataOptions } = require('../object/versioning');
+// const getReplicationInfo = require('../object/getReplicationInfo');
+const { config } = require('../../../Config');
+const kms = require('../../../kms/wrapper');
+const metadata = require('../../../metadata/wrapper');
+const { isScalityKmsArn, makeScalityArnPrefix } = require('arsenal/build/lib/network/KMSInterface');
+
+// Bucket need a key from the new KMS, not a simple reformating
+function updateBucketEncryption(bucket, log, cb) {
+    const sse = bucket.getServerSideEncryption();
+
+    if (!sse) {
+        return cb(null, bucket);
+    }
+
+    const masterKey = sse.masterKeyId;
+    const configuredKey = sse.configuredMasterKeyId;
+
+    // Note: if migration is from an external to an external, absence of arn is not enough
+    // a comparison of arn will be necessary but config validation blocks this for now
+    const updateMaster = masterKey && !isScalityKmsArn(masterKey);
+    const updateConfigured = configuredKey && !isScalityKmsArn(configuredKey);
+
+    if (!updateMaster && !updateConfigured) {
+        return cb(null, bucket);
+    }
+    log.debug('trying to update bucket encryption', { oldKey: masterKey || configuredKey });
+    // this should trigger vault account key update as well
+    return kms.createBucketKey(bucket, log, (err, newSse) => {
+        if (err) {
+            return cb(err, bucket);
+        }
+        // if both keys needs migration, it is ok the use the same KMS key
+        // as the configured one should be used and the only way to use the
+        // masterKeyId is to PutBucketEncryption to AES256 but then nothing
+        // will break and the same KMS key will continue to be used.
+        // And the key is managed (created) by Scality, not passed from input.
+        if (updateMaster) {
+            sse.masterKeyId = newSse.masterKeyArn;
+        }
+        if (updateConfigured) {
+            sse.configuredMasterKeyId = newSse.masterKeyArn;
+        }
+        // KMS account key will not be deleted when bucket is deleted
+        if (newSse.isAccountEncryptionEnabled) {
+            sse.isAccountEncryptionEnabled = newSse.isAccountEncryptionEnabled;
+        }
+
+        log.info('updating bucket encryption', {
+            oldKey: masterKey || configuredKey,
+            newKey: newSse.masterKeyArn,
+            isAccount: newSse.isAccountEncryptionEnabled,
+        });
+        return metadata.updateBucket(bucket.getName(), bucket, log, err => cb(err, bucket));
+    });
+}
+
+// Only reformat the key, don't generate a new one.
+// Use opts.skipObjectUpdate to only prepare objMD without sending the update to metadata
+// if a metadata.putObjectMD is expected later in call flow. (Downside: update skipped if error)
+function updateObjectEncryption(bucket, objMD, objectKey, log, keyArnPrefix, opts, cb) {
+    if (!objMD) {
+        return cb(null, bucket, objMD);
+    }
+
+    const key = objMD['x-amz-server-side-encryption-aws-kms-key-id'];
+
+    if (!key || isScalityKmsArn(key)) {
+        return cb(null, bucket, objMD);
+    }
+    const newKey = `${keyArnPrefix}${key}`;
+    // eslint-disable-next-line no-param-reassign
+    objMD['x-amz-server-side-encryption-aws-kms-key-id'] = newKey;
+    // Doesn't seem to be used but update as well
+    for (const dataLocator of objMD.location || []) {
+        if (dataLocator.masterKeyId) {
+            dataLocator.masterKeyId = `${keyArnPrefix}${dataLocator.masterKeyId}`;
+        }
+    }
+    // eslint-disable-next-line no-param-reassign
+    objMD.originOp = 's3:ObjectCreated:Copy';
+    // Copy should be tested for 9.5 in INTGR-1038
+    // to make sure it does not impact backbeat CRR / bucket notif
+    const params = getVersionSpecificMetadataOptions(objMD, config.nullVersionCompatMode);
+
+    log.info('reformating object encryption key', { oldKey: key, newKey, skipUpdate: opts.skipObjectUpdate });
+    if (opts.skipObjectUpdate) {
+        return cb(null, bucket, objMD);
+    }
+    return metadata.putObjectMD(bucket.getName(), objectKey, objMD, params,
+    log, err => cb(err, bucket, objMD));
+}
+
+/**
+ * Update encryption of bucket and object if kms provider changed
+ *
+ * @param {Error} err - error coming from metadata validate before the action handling
+ * @param {BucketInfo} bucket - bucket
+ * @param {Object} [objMD] - object metadata
+ * @param {string} objectKey  - objectKey from request.
+ * @param {Logger} log - request logger
+ * @param {Object} opts - options for sseMigration
+ * @param {boolean} [opts.skipObject] - ignore object update
+ * @param {boolean} [opts.skipObjectUpdate] - don't update metadata but prepare objMD for later update
+ * @param {Function} cb - callback (err, bucket, objMD)
+ * @returns {undefined}
+ */
+function updateEncryption(err, bucket, objMD, objectKey, log, opts, cb) {
+    // Error passed here to call the function inbetween the metadataValidate and its callback
+    if (err) {
+        return cb(err);
+    }
+    // if objMD missing, still try updateBucketEncryption
+    if (!config.sseMigration) {
+        return cb(null, bucket, objMD);
+    }
+
+    const { previousKeyType, previousKeyProtocol, previousKeyProvider } = config.sseMigration;
+    // previousKeyType is required and validated in Config.js
+    // for now it is the only implementation we need.
+    // See TAD Seamless decryption with internal and external KMS: https://scality.atlassian.net/wiki/x/EgADu
+    // for other method of migration without a previousKeyType
+
+    const keyArnPrefix = makeScalityArnPrefix(previousKeyType, previousKeyProtocol, previousKeyProvider);
+
+    return updateBucketEncryption(bucket, log, (err, bucket) => {
+        // Any error in updating encryption at bucket or object level is returned to client.
+        // Other possibilities: ignore error, include sse migration notice in error message.
+        if (err) {
+            return cb(err, bucket, objMD);
+        }
+        if (opts.skipObject) {
+            return cb(err, bucket, objMD);
+        }
+        return updateObjectEncryption(bucket, objMD, objectKey, log, keyArnPrefix, opts, cb);
+    });
+}
+
+module.exports = {
+    updateEncryption,
+};
--- a/lib/api/apiUtils/object/sseHeaders.js
+++ b/lib/api/apiUtils/object/sseHeaders.js
@@ -1,0 +1,18 @@
+const { config } = require('../../../Config');
+const { getKeyIdFromArn } = require('arsenal/build/lib/network/KMSInterface');
+
+function setSSEHeaders(headers, algo, kmsKey) {
+    if (algo) {
+        // eslint-disable-next-line no-param-reassign
+        headers['x-amz-server-side-encryption'] = algo;
+        if (kmsKey && algo === 'aws:kms') {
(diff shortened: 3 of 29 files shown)


lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+    if (!sse) {
+        return cb(null, bucket);
+    }
+
+    const masterKey = sse.masterKeyId;
+    const configuredKey = sse.configuredMasterKeyId;
+
+    // Note: if migration is from an ext to an ext, absence or arn is not enough
+    // a comparison of arn will be necessary
+    const updateMaster = masterKey && !masterKey.startsWith('arn:scality:kms:');
```
nicolas2bert: isScalityKmsArn(...) from arsenal could be used here?

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+    const sse = bucket.getServerSideEncryption();
+
+    if (!sse) {
+        return cb(null, bucket);
+    }
+
+    const masterKey = sse.masterKeyId;
+    const configuredKey = sse.configuredMasterKeyId;
+
+    // Note: if migration is from an ext to an ext, absence or arn is not enough
```
nicolas2bert: absence **of** arn prefix?

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+    const updateConfigured = configuredKey && !configuredKey.startsWith('arn:scality:kms:');
+
+    if (!updateMaster && !updateConfigured) {
+        return cb(null, bucket);
+    }
+    log.debug('trying to update bucket encryption', { oldKey: masterKey || configuredKey });
+    // this should trigger vault account key update as well
+    return kms.createBucketKey(bucket, log, (err, { masterKeyArn }) => {
+        if (err) {
+            return cb(err, bucket);
```
nicolas2bert, line 31: should we log something?

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+        }
+        if (updateMaster) {
+            sse.masterKeyId = masterKeyArn;
+        }
+        if (updateConfigured) {
+            sse.configuredMasterKeyId = masterKeyArn;
```
nicolas2bert: Are `sse.masterKeyId` and `sse.configuredMasterKeyId` not suppose to be different?
Does it break something if we have them both be the same value?

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
+const { getVersionSpecificMetadataOptions } = require('../object/versioning');
+// const getReplicationInfo = require('../object/getReplicationInfo');
+const { config } = require('../../../Config');
+const kms = require('../../../kms/wrapper');
+const metadata = require('../../../metadata/wrapper');
+
+// Bucket need a key from the new KMS, not a simple reformating
```
nicolas2bert, line 8: If we migrate from an external KMS to the same external KMS, we don't need to recreate master keys. Right? 
We don't handle the case here because it is for Outscale only that does an internal to external migration?

At the end the issue is not breaking since we will just generate duplicated KMS master key, right?

tcarmet (approved):

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        log.info('updating bucket encryption', { oldKey: masterKey || configuredKey, newKey: masterKeyArn });
+        return metadata.updateBucket(bucket.getName(), bucket, log, err => cb(err, bucket));
+    });
+}
+
+// Only reformat the key, don't generate a new one.
+// Use opts.skipObjectUpdate to only prepare objMD without sending the update to metadata
+// if an putObject is expected later in call flow. (Downside: update skipped if error)
```
nicolas2bert: Do you mean: "if a put object metadata (i.e. `metadata.putObjectMD`)"

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+    if (!key || key.startsWith('arn:scality:kms:')) {
+        return cb(null, bucket, objMD);
+    }
+    const newKey = `${keyArnPrefix}${key}`;
+    // eslint-disable-next-line no-param-reassign
+    objMD['x-amz-server-side-encryption-aws-kms-key-id'] = newKey;
+    // Doesn't seem to be used but update as well
+    for (const dataLocator of objMD.location || []) {
+        if (dataLocator.masterKeyId) {
+            dataLocator.masterKeyId = `${keyArnPrefix}${dataLocator.masterKeyId}`;
```
nicolas2bert, line 77: Could it be that the key and dataLocator.masterKeyId were, or could have been, updated separately?

- What if key has the old format, and dataLocator.masterKeyId has the new one?
In that case, dataLocator.masterKeyId might be updated with the same prefix multiple times.

- What if key has the new format, and dataLocator.masterKeyId still uses the old one?
In that case, dataLocator.masterKeyId might never get updated to the new format.

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+    // eslint-disable-next-line no-param-reassign
+    objMD.originOp = 's3:ObjectCreated:Copy';
+    const params = getVersionSpecificMetadataOptions(objMD, config.nullVersionCompatMode);
+
+    // TODO: what about CRR ?
```
nicolas2bert: What case should be handled for CRR?

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,140 @@
...
+        // TODO future dynamic evaluation of sse key
+        process.emitWarning('Only explicit KMS encryption key type migration is supported for now', {
+            name: 'S3C-9949',
+            detail: 'You need to set config.sseMigration.previousKeyType to int or ext',
+        });
+        return cb(null, bucket, objMD);
+    }
+
+    const keyArnPrefix =
+        `arn:scality:kms:${previousKeyType}:${previousKeyProtocol}:${previousKeyProvider}:key/`;
```
nicolas2bert: Should we centralize the creation of the prefix in one place? I saw it in different classes/repos.

lib/api/objectGet.js
```diff
@@ -51,6 +52,7 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
     };
 
     return metadataValidateBucketAndObj(mdValParams, log,
+    (err, bucket, objMD) => updateEncryption(err, bucket, objMD, objectKey, log, {},
```
nicolas2bert, line 55: Should we fail the GET/HEAD object call if update encryption fails?

lib/kms/wrapper.js
```diff
@@ -309,17 +377,21 @@ class KMS {
             cryptoScheme: serverSideEncryptionInfo.cryptoScheme,
             decipher: null,
         };
+
+        // shadowing global client for key - implName already used can't be shadowed here
+        const { client, implName: _impl, key } = getClientForKey(serverSideEncryptionInfo.masterKeyId);
```
nicolas2bert: We always need to check the client because there is no way to know that a migration never happened before? If we had a way to know that no migration happened before, we would be safe to use the default KMS, right? How is the impact of getClientForKey on the GET object latency?

lib/kms/wrapper.js
```diff
@@ -196,10 +261,13 @@ class KMS {
             log.debug('using user configured kms master key id');
             masterKeyId = configuredMasterKeyId;
         }
+        // shadowing global client for key
+        // but should not happen to cipher for another client as Puts should use current KMS
+        const { client, implName, key } = getClientForKey(masterKeyId);
```
nicolas2bert: Is getClientForKey needed for PUT object? We know that the client will always be the default one. What is the impact on our PUT object latency?

lib/kms/wrapper.js
```diff
@@ -10,49 +10,106 @@ const KMIPClient = require('arsenal').network.kmipClient;
...
+clientInstances[`${type}:${protocol}:${provider}`] = { client, implName };
+
+const mapKmsProtocolToClient = {
+    [KmsProtocol.aws_kms]: 'aws',
+    // others already match
+};
+
+// TODO store KMS identifier differently to avoid concatenations + split of arn
+// multiple times for a request
+function getClientForKey(key) {
```
nicolas2bert: Can we be certain that key always follows the new format?

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Yes

lib/kms/wrapper.js
BourgoisMickael: We should be, but I need to validate the format of the key passed to  PutObject if a scality arn is provided. The other places it's generated internally so we are good

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Yes they should be different, I need to fix this

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: For outscale it's internal to external, but there should be validation in place to prevent configuring a migration to the same provider.

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: I don't think so, they are not used to get the object as the GetObject rewrite those to decrypt the data: https://github.com/scality/cloudserver/blob/0f788abd327750d2a58d630babf1df3cfb423bca/lib/api/objectGet.js#L247-L251

I'll check why they are set on objects sometimes

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: make sure this internal migration is ignored for CRR and bucket notification.

But outscale don't use those, so it could be done later

lib/api/objectGet.js
BourgoisMickael: Right now it will fail, the error is passed to the next callback.

But we could provide a flag through config to ignore migration failure and still return the object.

The only reason where the update could fail but not the get is if you can't write, because disk is full but still wants to read from it.

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: errors should already be logged by kms implementation. But we could append a message to the error to describe the error is due to the SSE migration

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Actually nothing should break, right now they are different because they are set at different times, but having the same key is fine

lib/kms/wrapper.js
BourgoisMickael: I've added a validation function if the key matches the scality arn format.

Now we are sure the format is valid before using it with the KMS

lib/kms/wrapper.js
BourgoisMickael: It's needed if kms key is passed in scality arn format on a PUT request. Then the function will extract the KeyId from the arn and validate the arn scality format.

It also returns the client identifier so we can block usage of other client like if an arn for the `file` KMS is used when there is a configured external KMS, it should be forbidden: https://github.com/scality/cloudserver/pull/5788/files#diff-c506ce0254987c99cd5ff5a10cd1ecabc6e1bc59e643fe2401103b9feb7b27a6R332-R339

___

And for any future case where multiple KMS providers can be used it will become useful

lib/kms/wrapper.js
BourgoisMickael: If we know the migration never happened before, we'd rather know we should use the `previous` KMS instead of the new default external KMS.

But the way to know for now for outscale is if the scality kms arnPrefix is present or not.

I've also pre instantiated the previous KMS as `file` is the only allowed for now in migration, to avoid latency of first GET after startup.

The impact on GET object latency is:
- extract key detail from arn: `split('/')` and `split(':')` on the key: https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R122-R137
- if key is scality arnPrefix
  - validate the key parts: https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R145-R167
  - concatenate the `clientIdentifier` string: https://github.com/scality/cloudserver/pull/5788/files#diff-c506ce0254987c99cd5ff5a10cd1ecabc6e1bc59e643fe2401103b9feb7b27a6R131
  - return the client and KeyId
- if key is not scality arnPrefix
  - return the client (default or previous based on `sseMigration` configuration presence) and the KeyId

But as the GetObject migrate to new scality arn format, it should be expected the first Get on a bucket and object is slower for migration, and then we fall in the case where it's a scality arnPrefix

lib/kms/wrapper.js
BourgoisMickael: As for PUT latency, it will be same as GET:

The impact on PUT object latency is:

- extract key detail from arn: split('/') and split(':') on the key: https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R122-R137
- if key is scality arnPrefix
  - validate the key parts: https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R145-R167
  - concatenate the clientIdentifier string: https://github.com/scality/cloudserver/pull/5788/files#diff-c506ce0254987c99cd5ff5a10cd1ecabc6e1bc59e643fe2401103b9feb7b27a6R131
  - return the client and KeyId
- if key is not scality arnPrefix (should not happen as the PUT will prefix the key with a scality arn)
  - return the client (default or previous based on sseMigration configuration presence) and the KeyId

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: For any GET | Website | Copy the dataLocator sse is always rewritten with the metadata sse.

I suspect it should not be saved into DB but it's set on PUT for the dataWrapper to encrypt the data but in some cases end up being written into metadata

lib/Config.js
```diff
@@ -1485,6 +1517,61 @@ class Config extends EventEmitter {
...
+            }
+
+            // If previousKeyType is provided it's used as static value to migrate the format of the key
+            // without additional dynamic evaluation if the key provider is unknown.
+            assert(isValidType(previousKeyType),
+                'ssenMigration.previousKeyType must be "int" or "ext" (internal / external)');
+            this.sseMigration.previousKeyType = previousKeyType;
+
+            let expectedProtocol;
+            if (previousKeyType === KmsType.int) {
```
anurag4DSB: We should use `internal` here as `int` can give the impression of integers.
Other places in the config as well.

BourgoisMickael: It's been shipped like that, but each values will be documented as they don't represent the regular meaning in an aws arn. 
And it should not be customer facing, by default the arn is stripped when the key is returned to the client

lib/kms/wrapper.js
```diff
@@ -10,49 +10,157 @@ const KMIPClient = require('arsenal').network.kmipClient;
...
+} = require('arsenal/build/lib/network/KMSInterface');
+
+function getScalityKms() {
+    let scalityKMS;
+    let scalityKMSImpl;
+
+    try {
+        // eslint-disable-next-line import/no-unresolved
+        const ScalityKMS = require('scality-kms');
+        // TODO add scality arn in scality-kms
```
anurag4DSB: Do we still need this todo?

lib/Config.js
```diff
@@ -1509,6 +1517,61 @@ class Config extends EventEmitter {
...
+            /**
+             * For data that was encrypted internally by default and a new external provider is setup.
+             * This config helps detect the existing encryption key to decrypt with the good provider.
+             * The key format will be migrated automatically on GET/HEADs to include provider details.
+             */
+            this.sseMigration = {};
+            const { previousKeyType, previousKeyProtocol, previousKeyProvider } = config.sseMigration;
+            if (!previousKeyType) {
+                assert.fail(
+                    'NotImplemented: No dynamic KMS key migration. Set sseMigration.previousKeyType');
```
anurag4DSB, line 1531: NIT: Will it be helpful to use AWS (arsenal) Not implemented error with a customized description for this?
I know we do not have to comply with AWS specs here, that's why its a nit.

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,144 @@
...
+        // will break and the same KMS key will continue to be used.
+        // And the key is managed (created) by Scality, not passed from input.
+        if (updateMaster) {
+            sse.masterKeyId = masterKeyArn;
+        }
+        if (updateConfigured) {
+            sse.configuredMasterKeyId = masterKeyArn;
+        }
+
+        log.info('updating bucket encryption', { oldKey: masterKey || configuredKey, newKey: masterKeyArn });
```
anurag4DSB: This might be worth a debug log and not info log.
For info log we try to make sure we only have entry and exit logs.
I am curious why info will be helpful here for the extra log it might add.

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,144 @@
...
+    const params = getVersionSpecificMetadataOptions(objMD, config.nullVersionCompatMode);
+
+    // TODO: what about CRR ?
+    // const replicationInfo = getReplicationInfo(objectKey, bucket, true,
+    //     0, undefined, objMD);
+    // if (replicationInfo) {
+    //     // eslint-disable-next-line no-param-reassign
+    //     objMD.replicationInfo = Object.assign({},
+    //         objMD.replicationInfo, replicationInfo);
+    // }
```
anurag4DSB: Do we still need this code?
We should not merge commented code.

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,144 @@
...
+
+    log.info('reformating object encryption key', { oldKey: key, newKey, skipUpdate: opts.skipObjectUpdate });
```
anurag4DSB, line 86: Same for this one, why do we have info here

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,144 @@
...
+        return cb(err);
+    }
+    // if objMD missing, still try updateBucketEncryption
+    if (!config.sseMigration) {
+        return cb(null, bucket, objMD);
+    }
+
+    const { previousKeyType, previousKeyProtocol, previousKeyProvider } = config.sseMigration;
+    if (!previousKeyType) {
+        // TODO future dynamic evaluation of sse key
```
anurag4DSB: If not already, please create a ticket and add it here in the comment for reference.

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -0,0 +1,144 @@
...
+        process.emitWarning('Only explicit KMS encryption key type migration is supported for now', {
+            name: 'S3C-9949',
+            detail: 'You need to set config.sseMigration.previousKeyType to int or ext',
+        });
```
anurag4DSB: Do we really want to put the ticket number here?

lib/kms/wrapper.js
```diff
@@ -31,35 +38,127 @@ function getScalityKms() {
...
+        ({ scalityKMS: client, scalityKMSImpl: implName } = getScalityKms());
+    } else if (kms === 'kmip') {
+        const kmipConfig = { kmip: config.kmip };
+        if (!kmipConfig.kmip) {
+            throw new Error('KMIP KMS driver configuration is missing.');
+        }
+        client = new KMIPClient(kmipConfig);
+        implName = 'kmip';
+    } else if (kms === 'aws') {
+        const awsConfig = { kmsAWS: config.kmsAWS };
```
anurag4DSB: Like others do we need a validation here as well

lib/kms/wrapper.js
```diff
@@ -31,35 +38,127 @@ function getScalityKms() {
...
-} else if (config.backends.kms === 'file' || config.backends.kms === 'cdmi') {
-    client = file;
-    implName = 'fileKms';
-} else if (config.backends.kms === 'scality') {
-    ({ scalityKMS: client, scalityKMSImpl: implName } = getScalityKms());
-} else if (config.backends.kms === 'kmip') {
-    const kmipConfig = { kmip: config.kmip };
-    if (!kmipConfig.kmip) {
-        throw new Error('KMIP KMS driver configuration is missing.');
+function getClient(kms) {
```
anurag4DSB: Not picking.
This function is getting too complex, too many if/else
We could use a strategy here
Something like, may be lowercase part is not needed.
This keeps each backend’s logic self-contained, makes adding new ones as simple as extending the map, and consolidates your error handling and ensures error consistency. I would say standardize on either always throwing for missing config or letting the client constructors fail. At a minimum, document the behavior clearly.
```
const implementations = {
  mem:  { factory: () => inMemory,          name: 'memoryKms' },
  file: { factory: () => file,              name: 'fileKms' },
  cdmi: { factory: () => file,              name: 'fileKms' },
  scality: { factory: () => {
      const { scalityKMS, scalityKMSImpl } = getScalityKms();
      return { client: scalityKMS, implName: scalityKMSImpl };
    }
  },
  kmip: { factory: () => {
      if (!config.kmip) throw new Error('KMIP config missing');
      return { client: new KMIPClient({ kmip: config.kmip }), implName: 'kmip' };
    }
  },
  aws: { factory: () => {
      if (!config.kmsAWS) throw new Error('AWS config missing');
      return { client: new KmsAWSClient({ kmsAWS: config.kmsAWS }), implName: 'aws' };
    }
  },
};

function getClient(kms) {
  kms = (kms || '').toLowerCase();
  const impl = implementations[kms];
  if (!impl) throw new Error(`KMS backend is not configured: ${kms}`);
  return impl.factory();
}
```

lib/api/objectPutPart.js
```diff
@@ -218,6 +234,7 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
                     return next(err, destinationBucket);
                 }
                 // if data backend handles mpu, skip to end of waterfall
+                // TODO (artesca) data backend should return SSE to include in response headers
```
anurag4DSB: Can we back this up by a ticket as well.

lib/api/apiUtils/object/sseHeaders.js
```diff
@@ -0,0 +1,18 @@
...
+    if (algo) {
+        // eslint-disable-next-line no-param-reassign
+        headers['x-amz-server-side-encryption'] = algo;
+        if (kmsKey && algo === 'aws:kms') {
+            // eslint-disable-next-line no-param-reassign
+            headers['x-amz-server-side-encryption-aws-kms-key-id'] =
+                config.kmsHideScalityArn ? getKeyIdFromArn(kmsKey) : kmsKey;
+        }
+    }
+}
```
anurag4DSB, line 14: General question, instead of mutating, why don't we return a fresh headers?

package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/arsenal#7.70.4-2",
+    "arsenal": "git+https://github.com/scality/arsenal#improvement/ARSN-485-kms-provider",
```
anurag4DSB: Adding comment here to change this before merge

lib/kms/wrapper.js
BourgoisMickael: This is not needed for outscale but this is a reminder to be done in 9.5 when cherry-picking.
Here ticket: https://scality.atlassian.net/browse/KMS-7

lib/Config.js
BourgoisMickael: No need here, I want to fail the startup of cloudserver, like any other assert on config.

But for 9.5 it will be validated in Federation beforehand

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Usually logger is in info level and I wanted to be sure the print the log as this is something that should be happen once per bucket and once per object (below), so we can clearly see the migration is happening in logs without having to enable debug logs.

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: I did not check yet if the migration (using CopyObject as an operation) impacts backbeat CRR or bucket notifications as outscale don't use it.

Something I planned to check for 9.5

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: We are waiting for customer's need for other use cases of migration (and it might never happen, if only a few use SSE and none have mixed intern and extern KMS).

For that I think ticket would come from product if it ever happen.
This comment is rather a placeholder, maybe I should rather not write `TODO` inthis

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: I wrote this warning when I did not completely validated the config.

Now the config is validated (https://github.com/scality/cloudserver/pull/5788/files#diff-b06d72fe5b9eba1ab27b65f49f4fc461894ee5bf7507753734fdb5fff1eee387R1529)

This branch is not reachable anymore I can remove the emitWarning

lib/api/apiUtils/object/sseHeaders.js
BourgoisMickael: To follow how adding other headers have been implemented so far. As this is the end of request handling, it is more efficient to use the same object

- https://github.com/scality/cloudserver/blob/0f788abd327750d2a58d630babf1df3cfb423bca/lib/api/apiUtils/object/expirationHeaders.js#L130-L133 This function doesn't return anything it modify input headers
- https://github.com/scality/cloudserver/blob/0f788abd327750d2a58d630babf1df3cfb423bca/lib/api/objectGetACL.js#L194

- https://github.com/scality/cloudserver/blob/0f788abd327750d2a58d630babf1df3cfb423bca/lib/api/objectCopy.js#L604-L624

lib/api/objectPutPart.js
BourgoisMickael: https://scality.atlassian.net/browse/CLDSRV-640

lib/kms/wrapper.js
BourgoisMickael: Dropped the comment: https://github.com/scality/cloudserver/pull/5788/commits/fd1ca6bd2ef29c59a6d4c478c822ffd818aa7e0e

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: This is actually useful to torubleshoot the migration specially on special hotfix versions.

We should rather keep them as info for a while with our first clients using this. Maybe later it could me moved to debug

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Removed the comment: https://github.com/scality/cloudserver/pull/5788/commits/53a76a671a00ff266374dffbf3f60783382fc089

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Replaced the todo: https://github.com/scality/cloudserver/pull/5788/commits/53a76a671a00ff266374dffbf3f60783382fc089

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Dropped the lines: https://github.com/scality/cloudserver/pull/5788/commits/53a76a671a00ff266374dffbf3f60783382fc089

lib/kms/wrapper.js
BourgoisMickael: Done: https://github.com/scality/cloudserver/pull/5788/commits/fd1ca6bd2ef29c59a6d4c478c822ffd818aa7e0e

lib/kms/wrapper.js
BourgoisMickael: Yep this is a very nice way, I've done it here: https://github.com/scality/cloudserver/pull/5788/commits/fd1ca6bd2ef29c59a6d4c478c822ffd818aa7e0e

lib/api/objectPutPart.js
BourgoisMickael: Added a ref https://github.com/scality/cloudserver/pull/5788/commits/05d6766bd03952b5b3f3ac2c7fc7acef0ab2bceb

package.json
BourgoisMickael: There: https://github.com/scality/cloudserver/pull/5788/commits/8ff7b58ccc01ec3a6360e22730dbafface11eede

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: see previous comment

lib/api/apiUtils/bucket/updateEncryption.js
anurag4DSB: Okay, in general we should enable debug logs in such cases.
if this is the case in development versions as well, we need a ticket to resolve it

anurag4DSB (approved):
Thanks for the changes

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: Here a ticket to keep track: [CLDSRV-644](https://scality.atlassian.net/browse/CLDSRV-644)

For example with OKMS we did not ask to use debug logs and they ran without it, so having it in info for a while for a few NFI is helpful

PR #5792, opened 2025-05-02
CLDSRV-636: Hardcode outscale config for container

BourgoisMickael:
We will provide container as is to patch outscale
without generating a whole S3C

This will be merged in a new branch hotfix for outscale
___

Ignore failing tests because if hardcoded config value

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -419,6 +419,20 @@ class Config extends EventEmitter {
         }
         let kmsAWS = {};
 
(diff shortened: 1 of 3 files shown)


anurag4DSB (approved):
Same for this one like vault, hard coded should be avoided in development release branches,I understand this is just for outscale and on a special branch.

tcarmet (approved):

PR #5796, opened 2025-05-09
bf: CLDSRV-647 crash in getReplicationInfo with no replication endpoint

jonathan-gramain:
- Fix an exception in getReplicationInfo when:

  - no StorageClass is provided
  - no replication endpoint is configured

  Instead, return `undefined` so the object is not set to PENDING state for replication, even if the bucket already has a replication configuration enabled.

- [refactor] pass S3 config to getReplicationInfo
    
  Instead of pulling the location constraints and replication endpoints from the Cloudserver config within the getReplicationInfo helper itself, pass the config object explicitly. This makes the function pure, using only its input parameters, hence more test cases can be easily added with a different configuration for location constraints and/or replication endpoints.


tcarmet (approved):

anurag4DSB (approved):

anurag4DSB (approved):
At some places we are using `config` and others `s3config`, worth having that consistent or is it intentional?

jonathan-gramain:
@anurag4DSB

> At some places we are using config and others s3config, worth having that consistent or is it intentional?

In the caller's context, it's usually named `config` because it's how it's imported globally and doesn't cause an ambiguity at this level

In the callee's context (`getReplicationInfo`), it may be ambiguous as there's the replication config as well, and since it's a helper I though it best to emphasize that the config is (or has the same format than) the global S3 config.

PR #5797, opened 2025-05-12
Fixes for running cloudserver on s3c

francoisferrand:
- fix project path
- fix utapi config
- support null compatibility mode
- Enable multiObjectDelete optim' for mongo only
- do not fail on head with part number if object is empty
- All headObject on part1 for non mpu objects
- Allow disabling versioning if replication rules are disabled
- Remove useless storage class update
- Allow configuring valid storage classes
- Support batchDelete route when bucketName is not present
- Delete data when replacing object metadata with an empty object
- Trigger replication on putACL only for CRR

Issue: CLDSRV-638

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -18,7 +18,7 @@ export default [...compat.extends('@scality/scality'), {
     },
 
     languageOptions: {
-        ecmaVersion: 2020,
+        ecmaVersion: 2021,
         sourceType: "script",
     },
 
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1284,21 +1284,19 @@ class Config extends EventEmitter {
                 assert(config.localCache, 'missing required property of utapi ' +
                     'configuration: localCache');
                 this.utapi.localCache = this.localCache;
-                assert(config.redis, 'missing required property of utapi ' +
+                assert(config.utapi.redis, 'missing required property of utapi ' +
                     'configuration: redis');
-                if (config.utapi.redis) {
-                    this.utapi.redis = parseRedisConfig(config.utapi.redis);
-                    if (this.utapi.redis.retry === undefined) {
-                        this.utapi.redis.retry = {
-                            connectBackoff: {
-                                min: 10,
-                                max: 1000,
-                                jitter: 0.1,
-                                factor: 1.5,
-                                deadline: 10000,
-                            },
-                        };
-                    }
+                this.utapi.redis = parseRedisConfig(config.utapi.redis);
+                if (this.utapi.redis.retry === undefined) {
+                    this.utapi.redis.retry = {
+                        connectBackoff: {
+                            min: 10,
+                            max: 1000,
+                            jitter: 0.1,
+                            factor: 1.5,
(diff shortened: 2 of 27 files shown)


tests/unit/Config.js
```diff
@@ -1,8 +1,10 @@
 const assert = require('assert');
 const {
+    azureGetStorageAccountName,
+    azureGetLocationCredentials,
     locationConstraintAssert,
     parseSupportedLifecycleRules,
-    ConfigObject: ConfigObjectForTest,
+    ConfigObject: ConfigObject,
```
jonathan-gramain:
```suggestion
    ConfigObject,
```

lib/Config.js
```diff
@@ -1770,6 +1766,12 @@ class Config extends EventEmitter {
             kms,
             quota,
         };
+
+        // Mongodb backend does not support null keys, so we must enforce null version compatibility
+        // mode. With other backends (esp. metadata), this is used during migration for v0 to v1
```
jonathan-gramain:
```suggestion
        // mode. With other backends (esp. metadata), this is used during migration from v0 to v1
```

lib/api/objectHead.js
```diff
@@ -151,11 +151,14 @@ function objectHead(authInfo, request, log, callback) {
                     return callback(errors.BadRequest, corsHeaders);
                 }
                 const partSize = getPartSize(objMD, partNumber);
-                const isEmptyObject = objLength === 0;
-                if (!partSize && !isEmptyObject) {
+                if (partSize) {
```
jonathan-gramain, line 148: You may want to squash this change with the previous commit's

lib/metadata/acl.js
```diff
@@ -41,8 +42,21 @@ const acl = {
...
             objectMD.originOp = 's3:ObjectAcl:Put';
-            return metadata.putObjectMD(bucket.getName(), objectKey, objectMD, params, log,
-                cb);
+
+            // Use storageType to determine if replication update is needed, as it is set only for
+            // "cloud" locations. This  ensures that we reset replication when CRR is used, but not
+            // when multi-backend replication (i.e. Zenko) is used.
+            // TODO: this should be refactored to properly update the replication info, accounting
+            // for multiple rules and resetting the status only if needed CLDSRV-646
+            const replicationInfo = getReplicationInfo(objectKey, bucket, true);
```
jonathan-gramain, line 51: Note for later: CLDSRV-647 is slated to add an extra parameter to this function call (to provide the S3 config explicitly), but it hasn't been reviewed yet so it may change.

jonathan-gramain (approved):

eslint.config.mjs
```diff
@@ -18,7 +18,7 @@ export default [...compat.extends('@scality/scality'), {
     },
 
     languageOptions: {
-        ecmaVersion: 2020,
+        ecmaVersion: 2021,
```
ghost, line 21: Can we use the latest or some parts are not yet compatible (and should probably be identified)?

ghost (approved):
lgtm

PR #5803, opened 2025-05-20
CLDSRV-648: KMIP retries on cluster

BourgoisMickael:


lib/Config.js
```diff
@@ -629,6 +630,14 @@ class Config extends EventEmitter {
                 if (Array.isArray(config.kmip.transport)) {
                     this.kmip.transport = config.kmip.transport.map(t =>
                         this._parseKmipTransport(t));
+                    if (config.kmip.retries) {
+                        assert(typeof config.kmip.retries === 'number',
+                            'bad config: KMIP Cluster retries must be a number');
+                        assert(config.kmip.retries <= this.kmip.transport.length - 1,
```
dvasilas, line 636: Is it a good idea to assert in this case?
Maybe we could overwrite with `this.kmip.transport.length - 1` and `log.warn` or `log.error`?

package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
     "@hapi/joi": "^17.1.0",
-    "arsenal": "git+https://github.com/scality/Arsenal#7.70.44",
+    "arsenal": "git+https://github.com/scality/Arsenal#improvement/ARSN-495-kmip-retry",
```
dvasilas: Just a reminder to update to 7.70.45 before merging.

dvasilas (approved):

lib/Config.js
anurag4DSB: Also why `-1`, probably a dumb question.

anurag4DSB: Also why `-1` here, probably a dumb question

anurag4DSB (approved):

lib/Config.js
dvasilas: I guess it's because 1 original try +  (this.kmip.transport.length - 1) retries = 1 try for each host (that's my understanding / interpretation, but I may be wrong).

BourgoisMickael: Well I'd say we don't want to do something else than what the user configured and expected to happen.

I prefer a strict validation rather than implicitly ignore misconfiguration.

But I could include a check in Federation as well to have a faster fail for misconfig

dvasilas: Ok, makes sense.

BourgoisMickael:
> Also why -1 here, probably a dumb question

@anurag4DSB 

Because the request was already sent to 1 host at first.
So on error it becomes unhealthy and the remaining hosts in the cluster that can receive requests (for retry) is equal to `kmip.transport.length - 1`

lib/Config.js
BourgoisMickael: There the check [d7b2328](https://github.com/scality/Federation/pull/6197/commits/d7b2328e68b7ca9ef5d59a03bcd570df8f008b33)

PR #5806, opened 2025-05-22
do not allow using crr location as a locationConstraint

welansari:
- Support the 'CRR' type for locations
- Do not allow setting the CRR location as a locationConstraint

Issue: CLDSRV-653

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -305,7 +305,7 @@ function hdClientLocationConstraintAssert(configHd) {
 
 function locationConstraintAssert(locationConstraints) {
     const supportedBackends = [
-        'mem', 'file', 'scality', 'mongodb', 'tlp',
+        'mem', 'file', 'scality', 'mongodb', 'tlp', 'crr'
     ].concat(Object.keys(validExternalBackends));
     assert(typeof locationConstraints === 'object',
         'bad config: locationConstraints must be an object');
--- a/lib/api/bucketPut.js
+++ b/lib/api/bucketPut.js
@@ -59,7 +59,8 @@ function checkLocationConstraint(request, locationConstraint, log) {
         return { error: errorInstances.InvalidLocationConstraint.
             customizeDescription(errMsg) };
     }
-    if (locationConstraints[locationConstraintChecked].isCold) {
+    if (locationConstraints[locationConstraintChecked].isCold ||
+        locationConstraints[locationConstraintChecked].isCRR) {
         return { error: errors.InvalidLocationConstraint };
     }
     if (locationConstraintAddon) {
--- a/tests/multipleBackend/backendHealthcheckResponse.js
+++ b/tests/multipleBackend/backendHealthcheckResponse.js
@@ -11,6 +11,11 @@ const {
     getAzureContainerName,
 } = require('../functional/aws-node-sdk/test/multipleBackend/utils');
 
+const {
+    LOCATION_NAME_DMF,
+    LOCATION_NAME_CRR,
+} = require('../constants');
+
 const log = new DummyRequestLogger();
 const locConstraints = Object.keys(config.locationConstraints);
 const azureClient = getAzureClient();
@@ -22,7 +27,7 @@ describe('Healthcheck response', () => {
         clientCheck(true, log, (err, results) => {
             const resultKeys = Object.keys(results);
             locConstraints.forEach(constraint => {
-                if (constraint === 'location-dmf-v1' || constraint === 'location-crr-v1') {
+                if (constraint === LOCATION_NAME_DMF || constraint === LOCATION_NAME_CRR) {
                     // FIXME: location-dmf-v1 and location-crr-v1 are not in results, see CLDSRV-440
                     return;
                 }
@@ -45,7 +50,7 @@ describe('Healthcheck response', () => {
         clientCheck(false, log, (err, results) => {
             assert.notStrictEqual(results.length, locConstraints.length);
             locConstraints.forEach(constraint => {
-                if (constraint === 'location-dmf-v1' || constraint === 'location-crr-v1') {
+                if (constraint === LOCATION_NAME_DMF || constraint === LOCATION_NAME_CRR) {
                     // FIXME: location-dmf-v1 and location-crr-v1 are not in results, see CLDSRV-440
                     return;
                 }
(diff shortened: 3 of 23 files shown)


lib/api/bucketPut.js
```diff
@@ -59,7 +59,8 @@ function checkLocationConstraint(request, locationConstraint, log) {
         return { error: errorInstances.InvalidLocationConstraint.
             customizeDescription(errMsg) };
     }
-    if (locationConstraints[locationConstraintChecked].isCold) {
+    if (locationConstraints[locationConstraintChecked].isCold ||
+        locationConstraints[locationConstraintChecked].type === 'crr') {
```
francoisferrand: instead of doing the check based on the `type`, maybe we could to introduce a `feature` parameter, like isCold → `isCrr` ?

lib/api/bucketPut.js
```diff
@@ -59,7 +59,8 @@ function checkLocationConstraint(request, locationConstraint, log) {
         return { error: errorInstances.InvalidLocationConstraint.
             customizeDescription(errMsg) };
     }
-    if (locationConstraints[locationConstraintChecked].isCold) {
+    if (locationConstraints[locationConstraintChecked].isCold ||
```
francoisferrand, line 62: location constraint can be set on bucket (and should thus be rejected), but can also be specified directly in putObject / copyObject / putMpu : this needs to be handled as well

tests/multipleBackend/backendHealthcheckResponse.js
```diff
@@ -22,8 +22,8 @@ describe('Healthcheck response', () => {
         clientCheck(true, log, (err, results) => {
             const resultKeys = Object.keys(results);
             locConstraints.forEach(constraint => {
-                if (constraint === 'location-dmf-v1') {
-                    // FIXME: location-dmf-v1 is not in results, see CLDSRV-440
+                if (constraint === 'location-dmf-v1' || constraint === 'location-crr-v1') {
+                    // FIXME: location-dmf-v1 and location-crr-v1 are not in results, see CLDSRV-440
```
ghost, line 31: Maybe we shoudl update the ticket to include CRR now?

lib/Config.js
```diff
@@ -385,6 +385,10 @@ function locationConstraintAssert(locationConstraints) {
             assert(locationConstraints[l].isCold === true);
         }
 
+        if (locationConstraints[l].type === 'crr') {
+            assert(locationConstraints[l].isCrr === true);
+        }
```
ghost: If we always have the "type" crr and expect isCRR to be true, why not just relying on the `type` to detect if this is a CRR location instead?

lib/Config.js
```diff
@@ -385,6 +385,10 @@ function locationConstraintAssert(locationConstraints) {
             assert(locationConstraints[l].isCold === true);
         }
 
+        if (locationConstraints[l].type === 'crr') {
```
francoisferrand: why do we want to check this?
The code should use `isCrr` to check if this is a car-only location, so we should not check for the type? (type could actually be scality, in that case)

tests/unit/api/objectCopy.js
```diff
@@ -338,6 +339,22 @@ describe('non-versioned objectCopy', () => {
...
+            'x-amz-metadata-directive': 'REPLACE', // needed to take the locationConstraint into account
+            [objectLocationConstraintHeader]: 'location-crr-v1',
+        });
+
+        async.series([
+            next => objectPut(authInfo, testPutDestObjectRequest, undefined, log, next),
+            next => objectCopy(authInfo, testObjectCopyRequest, sourceBucketName, objectKey,
+                undefined, log, next),
+        ], err => {
+            assert(err.is.InvalidArgument);
```
francoisferrand, line 358: is this the right error?
should we not sent `InvalidLocationConstraint` in that case (i.e. trying to write on a valid location, but which does not support writing) instead of `InvalidArgument` (simply because we have no backend class for the type `crr` I guess?)

lib/api/bucketPut.js
francoisferrand: it seems weird that we don't have a similar check on `isCrr` in the other route (objectPut, `initiateMultiPartIUpload`, `copyObject` - or a common function called by these), to explicitly detect (and reject) the write to a crr location.
how come the test succeed? Is there some filtering somewhere?

lib/Config.js
francoisferrand: the point is actually the opposite: there may be multiple "types" of locations which support CRR (i.e. must be set in backbeat replication endpoint) but cannot be used for writing directly. So the flag `isCrr` was introduced to give this flexibility; and type may be used if we have different kinds of crr locations.

tests/unit/api/objectCopy.js
welansari: Looking at this test, it seems like there was a reason behind returning this error, not sure why however.
https://github.com/scality/cloudserver/blob/96a0315a8b7d436e4945630b4c0821bf5e7480aa/tests/unit/multipleBackend/locationConstraintCheck.js#L37-L46

lib/api/bucketPut.js
welansari: We don't do any filtering in those, we directly use the locationConstraint check function defined in Arsenal. We call
https://github.com/scality/cloudserver/blob/development/9.0/lib/api/apiUtils/object/locationConstraintCheck.js at some point in the code to get the controlling locationConstraint or fail when it's CRR. PutObject runs the check quite late (we do it in `createAndStoreObject`)

francoisferrand (approved):

tests/unit/api/multipartUpload.js
```diff
@@ -2174,6 +2174,32 @@ describe('Multipart Upload API', () => {
...
+        const initiateRequest = {
+            socket: {
+                remoteAddress: '1.1.1.1',
+            },
+            bucketName,
+            namespace,
+            objectKey,
+            headers: {
+                host: `${bucketName}.s3.amazonaws.com`,
+                [constants.objectLocationConstraintHeader]: 'location-crr-v1',
```
ghost: This is used multiple times, maybe better to add a global constant instead? (same for DMF?)

ghost (approved):

PR #5810, opened 2025-06-02
CLDSRV-657: KMS tests internal / external migration & arn prefix

BourgoisMickael:
Test is run in 3 steps:
- seed before migration
- run migration
- run kms kms tests on fresh buckets and objects (arnPrefix)

Those tests are based on the kmip test scenario but with more tests around CopyObject, MPU, and asserting on headers as well.


anurag4DSB (approved):

.github/docker/local.sh
```diff
@@ -0,0 +1,131 @@
...
+# run kms migration tests locally
+# in .github/docker
+
+export S3BACKEND=file
+export S3METADATA=file
+export S3VAULT=scality
+export MPU_TESTING="yes"
+
+export CLOUDSERVER_IMAGE_BEFORE_SSE_MIGRATION=ghcr.io/scality/cloudserver:7.70.66
+export CLOUDSERVER_IMAGE_ORIGINAL=ghcr.io/scality/cloudserver:7.70.70
```
dvasilas, line 13: Are we going to be updating the after migration images to the latest ones?

dvasilas (approved):

.github/docker/local.sh
BourgoisMickael: We can but don't need to for those tests.

PR #4926, opened 2022-12-20
Feature/cldsrv 244/migrate to GitHub actions

alexanderchan-scality:
Migrate CI from Eve to Github Actions:
* multiple-backend-tests
* file-ft-tests
* kmip-ft-tests
* utapi-v2-tests

PR #5824, opened 2025-06-11
CLDSRV-657: Fix kmip tests to use pykmip

BourgoisMickael:
kmip tests were not using pykmip but file kms...

Seems to be since migration from eve to GHA (Issue [CLDSRV-244](https://scality.atlassian.net/browse/CLDSRV-244))
where the kmip config via env was forgotten.

Lost here: https://github.com/scality/cloudserver/pull/4926/files#diff-b0a705e0cbace75695b5e1e57b39c3a73ba96df02fd969ce4c16fd0a9a05df0cL96-L113

I change the ssl-kmip mount in container to match the same
path on host and container for certs as we will reuse the
same config on host to run tests (sse migration)

[CLDSRV-244]: https://scality.atlassian.net/browse/CLDSRV-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/docker/docker-compose.yaml
+++ b/.github/docker/docker-compose.yaml
@@ -5,7 +5,7 @@ services:
     network_mode: "host"
     volumes:
       - /tmp/ssl:/ssl
-      - /tmp/ssl-kmip:/ssl-kmip
+      - /tmp/ssl-kmip:/tmp/ssl-kmip
       - ${HOME}/.aws/credentials:/root/.aws/credentials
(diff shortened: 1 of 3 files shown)


tcarmet (approved):
nice catch

dvasilas (approved):

PR #5826, opened 2025-06-12
Pass instanceId to mongodb metadata backend

welansari:
`instanceId` will be used when generating versionIds.
This will ensure uniqueness of the versionID across
all Cloudserver pods in the cluster.

The `instanceId` is a combination of the short hash
contained in the name of the pods, which is unique
across all pods within the same deployment, and a
prefix for differenciating the different types of
deployments (internal vs external).

see. https://scality.atlassian.net/wiki/spaces/OS/pages/3225583692/VersionID+Collisions+in+Zenko

Related PR: https://github.com/scality/Arsenal/pull/2445

Issue: CLDSRV-665

### DIFF
--- a/.github/docker/docker-compose.yaml
+++ b/.github/docker/docker-compose.yaml
@@ -45,6 +45,7 @@ services:
       - SCUBA_HEALTHCHECK_FREQUENCY
       - S3QUOTA
       - QUOTA_ENABLE_INFLIGHTS
+      - S3_VERSION_ID_ENCODING_TYPE
     env_file:
       - creds.env
     depends_on:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -293,6 +293,7 @@ jobs:
       S3METADATA: mongodb
       S3KMS: file
       S3_LOCATION_FILE: /usr/src/app/tests/locationConfig/locationConfigTests.json
+      S3_VERSION_ID_ENCODING_TYPE: base62
       DEFAULT_BUCKET_KEY_FORMAT: v1
       METADATA_MAX_CACHED_BUCKETS: 1
       MONGODB_IMAGE: ghcr.io/${{ github.repository }}/ci-mongodb:${{ github.sha }}
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -968,6 +968,38 @@ class Config extends EventEmitter {
             this.replicationGroupId = 'RG001';
         }
 
+        const instanceId = process.env.CLOUDSERVER_INSTANCE_ID || config.instanceId;
+        if (instanceId) {
+            assert(typeof instanceId === 'string',
+                'bad config: instanceId must be a string');
+            // versionID generation code will truncate instanceId to 6 characters
+            // so we enforce this limit here to make the behavior predictable
+            assert(instanceId.length <= 6,
+                'bad config: instanceId must be at most 6 characters long');
+            this.instanceId = instanceId;
+        } else if (process.env.HOSTNAME) {
+            // If running in Kubernetes, the pod name is set in the HOSTNAME env var
+            // and is of the form: cloudserver-7869f99955-nr984
+            // where:
+            // - cloudserver is the name of the pod
+            // - 7869f99955 is a long hash that guarantees uniqueness within the namespace
+            // - nr984 is a short hash that is unique only within the current ReplicaSet/Deployment
+            // We use the last part of the pod name as the instanceId, prefixed with 'i' for internal
+            // or 'e' for external cloudserver.
+            // This allows having a unique identifier that will be used when generating versionIds
+            const instanceType = process.env.HOSTNAME.includes('internal') ? 'i' : 'e';
+            const parts = process.env.HOSTNAME.split('-');
+            if (parts.length >= 2) {
+                const shortId = parts[parts.length - 1];
+                this.instanceId = instanceType + shortId;
+            } else {
+                // fallback: generate a random string if format is unexpected
+                this.instanceId = uuidv4().replace(/-/g, '').slice(0, 6);
+            }
+        } else {
+            this.instanceId = uuidv4().replace(/-/g, '').slice(0, 6);
+        }
+
         this.replicationEndpoints = [];
         if (config.replicationEndpoints) {
             const { replicationEndpoints } = config;
--- a/tests/unit/Config.js
+++ b/tests/unit/Config.js
@@ -1,7 +1,10 @@
 const assert = require('assert');
+const sinon = require('sinon');
+const fs = require('fs');
+const path = require('path');
 const {
     azureArchiveLocationConstraintAssert,
-    ConfigObject: ConfigObjectForTest,
+    ConfigObject,
 } = require('../../lib/Config');
 
 describe('Config', () => {
@@ -25,7 +28,6 @@ describe('Config', () => {
     });
 
     it('should emit an event when auth data is updated', done => {
-        const { ConfigObject } = require('../../lib/Config');
         const config = new ConfigObject();
         let emitted = false;
         config.on('authdata-update', () => {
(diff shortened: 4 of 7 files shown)


ghost (commented):
LGTM, as discussed, let's add the base62 as the default during tests as well, so we are actually testing what is used in production

lib/Config.js
```diff
@@ -906,6 +906,15 @@ class Config extends EventEmitter {
...
             this.replicationGroupId = config.replicationGroupId;
+        } else if (process.env.HOSTNAME) {
+            // Extract the hash part from the pod name (e.g., cloudserver-7869f99955-nr984)
+            const parts = process.env.HOSTNAME.split('-');
+            // Kubernetes pods typically have two UID parts:
+            // - A long hash (10 characters) that guarantees uniqueness within the namespace.
+            // - A short hash (5 characters) that is unique only within the current ReplicaSet/Deployment.
+            // To avoid the risk of collision between internal and external Cloudservers we use the longer
+            // hash as a replicationGroupId.
+            this.replicationGroupId = parts.length > 2 ? parts[parts.length - 2] : parts.pop();
```
ghost: Actually we cannot use only the long: as this will be the same for all the pods in the same deployments.

My suggestion, given the 7 characters constraint, would be to use the short one plus the two first characters of the long one.
It guarantees the uniqueness for a deployment, and the odds of facing this with the internal cloudserver is infinitesimal...

welansari: agreed

ghost (approved):

tests/unit/Config.js
```diff
@@ -810,4 +810,56 @@ describe('Config', () => {
...
+            } else {
+                delete process.env.HOSTNAME;
+            }
+        });
+
+        it('should return the replicationGroupId from config', () => {
+            const config = new ConfigObject();
+            assert.strictEqual(config.replicationGroupId, 'RG001');
+        });
+        it('should genrate a new replicationGroupId if not set in config', () => {
```
jonathan-gramain:
```suggestion
        it('should generate a new replicationGroupId if not set in config', () => {
```

lib/Config.js
```diff
@@ -906,6 +906,20 @@ class Config extends EventEmitter {
...
+            if (parts.length > 2) {
+                const longId = parts[parts.length - 2];
+                const shortId = parts[parts.length - 1];
+                this.replicationGroupId = longId.slice(0, 2) + shortId;
+            } else {
+                // fallback: just use the last part if format is unexpected
+                this.replicationGroupId = parts.pop();
+            }
         } else {
             this.replicationGroupId = 'RG001';
```
jonathan-gramain: We may one day hit the same issue if we have a "CRR" configuration on Artesca that uses a different replication group ID, right? Could be worth ensuring that if we're in an Artesca setting (how?), we don't have a replication group ID configured, and maybe refuse to start if that's the case.

jonathan-gramain (approved):

lib/Config.js
jonathan-gramain: Checking if the metadata backend is of type "mongo" could be the appropriate way I think, since the issue is related to our generation of client-side versions in the mongo wrapper.

welansari:
Note for reviewers, i updated this PR following the changes in the TAD.
We no longer use the `replicationGroupId` as a local unique id but a new field called `instanceId`.

ghost:
@Kerkesni do we want to target 8.8? Some clients might not be ready to upgrade to the latest product version right away

welansari:
@williamlardier yes forgot about that part, thanks

lib/Config.js
```diff
@@ -968,6 +968,29 @@ class Config extends EventEmitter {
...
+            // If running in Kubernetes, the pod name is set in the HOSTNAME env var
+            // and is of the form: cloudserver-7869f99955-nr984
+            // where:
+            // - cloudserver is the name of the pod
+            // - 7869f99955 is a long hash that guarantees uniqueness within the namespace
+            // - nr984 is a short hash that is unique only within the current ReplicaSet/Deployment
+            // We use the last part of the pod name as the instanceId, prefixed with 'i' for internal
+            // or 'e' for external cloudserver.
+            // This allows having a unique identifier that will be used when generating versionIds
+            const instanceType = process.env.HOSTNAME?.includes('internal') ? 'i' : 'e';
```
ghost: The `?` is not needed as we already check above

```suggestion
            const instanceType = process.env.HOSTNAME.includes('internal') ? 'i' : 'e';
```

ghost (commented):
lgtm otherwise, but let's wait for the finalized Arsenal part

lib/Config.js
```diff
@@ -968,6 +968,29 @@ class Config extends EventEmitter {
...
+                const shortId = parts[parts.length - 1];
+                this.instanceId = instanceType + shortId;
+            } else {
+                // fallback: generate a random string if format is unexpected
+                this.instanceId = uuidv4().replace(/-/g, '').slice(0, 6);
+            }
+        } else {
+            this.instanceId = uuidv4().replace(/-/g, '').slice(0, 6);
+        }
+
```
francoisferrand, line 1002: it feels wrong to include this logic (tied very strongly to k8s implementation details) in here : this really breaks separation of concerns...

However, not sure if/how we can do so much better. It may be possible to [inject k8s pod fields](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) into the pod, and possible use both `name` (cloudserver-7869f99955-nr984) and `computedName` (cloudserver-7869f99955) to reduce coupling, possibly using a ZKOP-defined initContainer....

....but would need much more investigation and dependency with ZKOP, and not sure we can really wait here.

→ Let's create a ticket to improve this, and (eventually) move this code out of cloudserver and into ZKOP
→ In the mean time, we should keep Cloudserver futureproof by allowing to explicitly set the instanceId:

```
   const instanceId = env.CLOUDSERVER_INSTANCE_ID || config.InstanceID;
   if (instanceId) {
      this.instanceId = instanceId;
   } else if (process.env.HOSTNAME) {
      ...
   } else {
      this.instanceId = uuidv4().replace(/-/g, '').slice(0, 6);
   }
```

.github/workflows/tests.yaml
```diff
@@ -255,6 +255,7 @@ jobs:
       S3METADATA: mongodb
       S3KMS: file
       S3_LOCATION_FILE: /usr/src/app/tests/locationConfig/locationConfigTests.json
+      S3_VERSION_ID_ENCODING_TYPE: base62
```
francoisferrand: should we keep this "unset" in mongo-v0, to exercise both behaviors?
(or actually just remove mongo-v0, as we always use v1 with mongo)

.github/docker/docker-compose.yaml
```diff
@@ -45,6 +45,7 @@ services:
       - SCUBA_HEALTHCHECK_FREQUENCY
       - S3QUOTA
       - QUOTA_ENABLE_INFLIGHTS
+      - S3_VERSION_ID_ENCODING_TYPE
```
francoisferrand, line 48: instead of setting this here, should we (can we?) set it automatically in config.js when backend is `mongo` ?

francoisferrand (approved):

.github/workflows/tests.yaml
welansari: The code base still supports both v0 and v1 so we should keep the tests. I made v0 tests run with hex versionIDs

.github/docker/docker-compose.yaml
welansari: This will need quite a rework as `S3_VERSION_ID_ENCODING_TYPE` is used by the versionID generation code inside of Arsenal, it's not a simple change in Cloudserver

lib/Config.js
welansari: created CLDSRV-722

PR #5836, opened 2025-06-23
CLDSRV-670: KMS error message for Copy actions

BourgoisMickael:
Return a details KMS error for copy actions

And add tests cases for that

Needs https://github.com/scality/Arsenal/pull/2432

### DIFF
--- a/tests/functional/sse-kms-migration/arnPrefix.js
+++ b/tests/functional/sse-kms-migration/arnPrefix.js
@@ -505,3 +505,163 @@ describe('ensure MPU use good SSE', () => {
         );
     });
 });
+describe('KMS error', () => {
+    const sseConfig = { algo: 'aws:kms', masterKeyId: true };
+    const Bucket = 'bkt-kms-err';
+    const Key = 'obj';
+    const body = 'content';
+
+    let mpuEncrypted;
+    let mpuPlaintext;
+
+    let masterKeyId;
+    let masterKeyArn;
+
+    let expected;
+
+    const expectedKMIP = {
+        code: 'KMS.NotFoundException',
+        msg: (action, keyId) => new RegExp(`^KMS \\(KMIP\\) error for ${action} on ${keyId}\\..*`),
+    };
+    const expectedAWS = {
+        code: 'KMS.KMSInvalidStateException',
+        msg: (_, keyId) => new RegExp(`${keyId} is pending deletion\\.`),
+    };
+    /**
+     * localkms container returns a different error message when the key is pending deletion
+     * as we decrypt without passing the keyId, so we need to handle it separately
+     */
+    const expectedLocalKms = {
+        code: 'KMS.AccessDeniedException',
+        msg: () => new RegExp(
+            'The ciphertext refers to a customer master key that does not exist, ' +
+            'does not exist in this region, or you are not allowed to access\\.'
+        ),
+    };
+    if (helpers.config.backends.kms === 'kmip') {
+        expected = expectedKMIP;
+    } else if (helpers.config.backends.kms === 'aws') {
+        expected = expectedAWS;
+    } else {
+        throw new Error(`Unsupported KMS backend: ${helpers.config.backends.kms}`);
+    }
+
+    function assertKmsError(action, keyId) {
+        return err => {
+            if (helpers.config.backends.kms === 'aws' && action === 'Decrypt') {
+                assert.strictEqual(err.name, expectedLocalKms.code);
+                assert.match(err.message, expectedLocalKms.msg(action, keyId));
+                return true;
+            }
+            assert.strictEqual(err.name, expected.code);
+            assert.match(err.message, expected.msg(action, keyId));
+            return true;
+        };
+    }
+
+    before(async () => {
+        void await helpers.s3.createBucket({ Bucket }).promise();
+
+        await helpers.s3.putObject({
+            ...helpers.putObjParams(Bucket, 'plaintext', {}, null),
+            Body: body,
+        }).promise();
+
+        mpuPlaintext = await helpers.s3.createMultipartUpload(
+            helpers.putObjParams(Bucket, 'mpuPlaintext', {}, null)).promise();
+
+        ({ masterKeyId, masterKeyArn } = await helpers.createKmsKey(log));
+
+        await helpers.putEncryptedObject(Bucket, Key, sseConfig, masterKeyArn, body);
+        // ensure we can decrypt and read the object
+        const obj = await helpers.s3.getObject({ Bucket, Key }).promise();
+        assert.strictEqual(obj.Body.toString(), body);
+
+        mpuEncrypted = await helpers.s3.createMultipartUpload(
+            helpers.putObjParams(Bucket, 'mpuEncrypted', sseConfig, masterKeyArn)).promise();
+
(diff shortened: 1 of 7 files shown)


tests/functional/sse-kms-migration/arnPrefix.js
```diff
@@ -505,3 +506,163 @@ describe('ensure MPU use good SSE', () => {
         );
     });
 });
+
+describe('KMS error', () => {
+    const destroyKmsKey = promisify(kms.destroyBucketKey);
```
anurag4DSB: Promisifying at module-load time freezes the current reference; could we move this  inside `before()` so stubs/mocks added later in the test suite are picked up.

tests/functional/sse-kms-migration/arnPrefix.js
```diff
@@ -505,3 +506,163 @@ describe('ensure MPU use good SSE', () => {
...
+    const sseConfig = { algo: 'aws:kms', masterKeyId: true }
```
anurag4DSB: [NIT] A boolean `true` doesn’t convey which key is used and can be mis-read as “auto-pick”; drop the field or pass an explicit key ID so intent is unambiguous.

tests/functional/sse-kms-migration/arnPrefix.js
```diff
@@ -505,3 +506,163 @@ describe('ensure MPU use good SSE', () => {
...
+    const expectedKMIP = {
+        code: /KMS\.NotFoundException/,
+        msg: (action, keyId) => new RegExp(`^KMS \\(KMIP\\) error for ${action} on ${keyId}\\..*`),
+    };
+    // localkms container returns a different error message when the key is pending deletion
+    // as we decrypt without passing the keyId, so we need to handle it separately
+    const localKms = 'The ciphertext refers to a customer master key that does not exist, ' +
+        'does not exist in this region, or you are not allowed to access\\.';
+    const expectedAWS = {
+        code: /KMS\.KMSInvalidStateException|KMS\.AccessDeniedException/,
```
anurag4DSB: {nit} Two unrelated AWS failure modes in one regex can hide drift; split them into separate, focused assertions.

tests/functional/sse-kms-migration/arnPrefix.js
```diff
@@ -505,3 +506,163 @@ describe('ensure MPU use good SSE', () => {
...
+        const obj = await helpers.s3.getObject({ Bucket, Key }).promise();
+        assert.strictEqual(obj.Body.toString(), body);
+        void await destroyKmsKey(masterKeyArn, log);
+    });
+
+    after(async () => {
+        void await helpers.cleanup(Bucket);
+        if (masterKeyArn) {
+            try {
+                void await destroyKmsKey(masterKeyArn, log);
```
anurag4DSB: Dropping the awaited value with `void` swallows any rejection; await directly (or wrap in try/catch + log) so cleanup failures surface in CI.

tests/functional/sse-kms-migration/arnPrefix.js
```diff
@@ -505,3 +506,163 @@ describe('ensure MPU use good SSE', () => {
...
+        ({ masterKeyId, masterKeyArn } = await helpers.createKmsKey(log));
+        await helpers.putEncryptedObject(Bucket, Key, sseConfig, masterKeyArn, body);
+        // ensure we can decrypt and read the object
+        const obj = await helpers.s3.getObject({ Bucket, Key }).promise();
+        assert.strictEqual(obj.Body.toString(), body);
+        void await destroyKmsKey(masterKeyArn, log);
+    });
+
+    after(async () => {
+        void await helpers.cleanup(Bucket);
```
anurag4DSB, line 588: Same comment here as below

anurag4DSB: and other voids

tests/functional/sse-kms-migration/arnPrefix.js
```diff
@@ -505,3 +506,163 @@ describe('ensure MPU use good SSE', () => {
...
+    afterEach(async () => {
+        if (anotherKeyInfo) {
+            try {
+                void await destroyKmsKey(anotherKeyInfo.masterKeyArn, log);
+            } catch (e) { void e; }
+            anotherKeyInfo = null;
+        }
+    });
+
+    it('putObject should fail with kms error', async () => {
```
anurag4DSB: Am I understanding this right, six nearly identical tests below differ only by operation name, parameterise them in a loop to cut duplication and make adding new KMS paths a one-liner.

tests/functional/sse-kms-migration/arnPrefix.js
BourgoisMickael: The key is passed as next parameter. this is for this function https://github.com/scality/cloudserver/blob/f2075a3b3d9ed5d8386f5614f46ec5affec92a4e/tests/functional/sse-kms-migration/helpers.js#L33-L44

The signature is already like that and I reuse the same config but with different keys

anurag4DSB (approved):

tests/functional/sse-kms-migration/arnPrefix.js
BourgoisMickael: void only replaces the expression returned value with `undefined`. It does not swallow any `await`ed rejection throws. Here the `try catch` does actually catch and ignore an error.

I added a void for eslint rule https://eslint.org/docs/latest/rules/no-unused-expressions

I don't see in the doc where void could swallow rejections or thrown errors (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void)

tests/functional/sse-kms-migration/arnPrefix.js
BourgoisMickael: This is functional test there should and will be no stubs or mock on this as we need this client to send real messages to the KMS.

But I moved this into a helper module where there are already other promisified clients to be reused [b4878d1](https://github.com/scality/cloudserver/pull/5836/commits/b4878d140a092213710450f202f1d0d2d9160453)

tests/functional/sse-kms-migration/arnPrefix.js
BourgoisMickael: I've splitted to use the localkms only on a specific condition [94a95d2](https://github.com/scality/cloudserver/pull/5836/commits/94a95d27e48d2b04a78b7c49344d3cb73faf2856)

tests/functional/sse-kms-migration/arnPrefix.js
BourgoisMickael: I've factorized [94a95d2](https://github.com/scality/cloudserver/pull/5836/commits/94a95d27e48d2b04a78b7c49344d3cb73faf2856)

The test array takes almost more lines

tests/functional/sse-kms-migration/arnPrefix.js
BourgoisMickael: I'll make another PR to replace this with ignoring specific linting rule

BourgoisMickael: There https://github.com/scality/cloudserver/pull/5847

tmacro (approved):

jonathan-gramain (approved):

PR #5846, opened 2025-06-23
CLDSRV-671: deactivate dependabot

tcarmet:
As decided, deactivating dependabot until we make the necessary changes to be able to merge the PR made by the bot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,0 @@
----
-version: 2
-updates:
-  - package-ecosystem: npm
-    directory: "/"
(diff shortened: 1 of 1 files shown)


jonathan-gramain (approved):

BourgoisMickael (approved):
Can't we turn it off via the repo settings without touching the config file ?

tcarmet:
> Can't we turn it off via the repo settings without touching the config file ?

we can't. We can disable some things at the repo settings level, but as long as there's a file it will be read and taken into account.

PR #5852, opened 2025-06-29
CLDSRV-673: KMIP Cluster test multi IP

BourgoisMickael:


.github/workflows/tests.yaml
```diff
@@ -372,6 +372,67 @@ jobs:
...
+    - name: Upload logs to artifacts
+      uses: scality/action-artifacts@v4
+      with:
+        method: upload
+        url: https://artifacts.scality.net
+        user: ${{ secrets.ARTIFACTS_USER }}
+        password: ${{ secrets.ARTIFACTS_PASSWORD }}
+        source: /tmp/artifacts
+      if: always()
+
```
tcarmet, line 435: So it's only done on `development/9.0` and onwards so we can do it in a separate PR or add an extra commit to the `w/9.0` branch, but let's upload the test results to codecov:

```yaml
    - name: Upload test results to Codecov
      uses: codecov/test-results-action@v1
      with:
        token: ${{ secrets.CODECOV_TOKEN }}
        files: '**/junit/*junit*.xml'
        flags: kmip-cluster-ft-test
```

I can set it up after the PR is merged, just let me know :)

tcarmet (approved):

.github/workflows/tests.yaml
BourgoisMickael: Oh same will need to be done for `sse-kms-migration-tests` then as well, so it might be another PR for that branch.

I see we also have this block
```
    - name: Cleanup and upload coverage
      uses: ./.github/actions/cleanup-and-coverage
      with:
        profiles: pykmip
        codecov-token: ${{ secrets.CODECOV_TOKEN }}
```
That uploads the whole directory `        directory: /tmp/coverage/${{ github.job }}` can't we push everything in the same directory to upload only once ?

nicolas2bert (approved):

.github/workflows/tests.yaml
tcarmet: They are different kind of files and require different type of push, so I would say no. code coverage is one thing, test result another.

PR #5853, opened 2025-07-01
Improvement/cldsrv 646

benzekrimaha:
Issue : [CLDSRV-646](https://github.com/scality/cloudserver/tree/improvement/CLDSRV-646)

[CLDSRV-646]: https://scality.atlassian.net/browse/CLDSRV-646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/metadata/acl.js
+++ b/lib/metadata/acl.js
@@ -44,17 +44,43 @@ const acl = {
             objectMD.acl = addACLParams;
             objectMD.originOp = 's3:ObjectAcl:Put';
 
-            // Use storageType to determine if replication update is needed, as it is set only for
-            // "cloud" locations. This  ensures that we reset replication when CRR is used, but not
-            // when multi-backend replication (i.e. Zenko) is used.
-            // TODO: this should be refactored to properly update the replication info, accounting
-            // for multiple rules and resetting the status only if needed CLDSRV-646
             const replicationInfo = getReplicationInfo(config, objectKey, bucket, true);
-            if (replicationInfo && !replicationInfo.storageType) {
-                objectMD.replicationInfo = {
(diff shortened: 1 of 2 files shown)


lib/metadata/acl.js
```diff
@@ -44,17 +44,43 @@ const acl = {
...
-            // for multiple rules and resetting the status only if needed CLDSRV-646
             const replicationInfo = getReplicationInfo(config, objectKey, bucket, true);
-            if (replicationInfo && !replicationInfo.storageType) {
-                objectMD.replicationInfo = {
-                    ...objectMD.replicationInfo,
-                    ...replicationInfo,
-                };
+            
+            // Split the storageClass and iterate over each one
+            if (replicationInfo) {
```
francoisferrand, line 50: this does not seem the right place to do this:
- it is the goal of `_getReplicationInfo` to build the replication info
- that function is already doing the parsing/splitting of storageClasses, and associated locationConstrainsts
- the last param to `getReplicationInfo` indicates this is a MD-only operation (which currently should be replicated only to CRR), so this can be used deeper (currently this gets obfuscated in `getReplicationInfo` by converting it to `const content = isMD || objSize === 0 ? ['METADATA'] : ['DATA', 'METADATA']`, but this may be moved to `_getReplicationInfo`)

francoisferrand (commented):
I understand this is a draft PR, but this is still a first stage of publicating the changes : so there should already be some PR comment and commit message to document what changing you are introducing.

PR #5854, opened 2025-07-01
AbortMPU cleanup - revisited

ghost:
Relates to https://scality.atlassian.net/wiki/spaces/OS/pages/3258220958/Retriable+CompleteMultipartUpload

See ticket & commits for context

Issue: CLDSRV-669

Following this PR, we will fix the behavior of the CompleteMPU API to avoid promoting wrongly s3 objects if not all parts are valid. This might not be feasible because we want to be unified with S3C, that is not 100% standard with MPU management, as we can abort after a complete officially.

### DIFF
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -78,10 +78,10 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                     }
                     // for Azure and GCP we do not need to delete data
                     // for all other backends, skipDataDelete will be set to false
-                    return next(null, mpuBucket, mpuOverviewObj, destBucket, objectMD, skipDataDelete);
+                    return next(null, mpuBucket, destBucket, objectMD, skipDataDelete);
                 });
         },
-        function getPartLocations(mpuBucket, mpuOverviewObj, destBucket, objectMD, skipDataDelete,
+        function getPartLocations(mpuBucket, destBucket, objectMD, skipDataDelete,
             next) {
             services.getMPUparts(mpuBucket.getName(), uploadId, log,
                 (err, result) => {
@@ -90,7 +90,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                         return next(err, destBucket);
                     }
                     const storedParts = result.Contents;
-                    return next(null, mpuBucket, mpuOverviewObj, storedParts, destBucket, objectMD,
+                    return next(null, mpuBucket, storedParts, destBucket, objectMD,
                         skipDataDelete);
                 });
         },
@@ -103,13 +103,8 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
         // to honor the uploadId filter in standardMetadataValidateBucketAndObj
         // ensuring the objMD returned has the right uploadId. But this is not
         // supported by Metadata.
-        function ensureCleanupIsRequired(mpuBucket, mpuOverviewObj, storedParts, destBucket,
+        function ensureCleanupIsRequired(mpuBucket, storedParts, destBucket,
             objectMD, skipDataDelete, next) {
-            // If no overview object, skip - this wasn't a failed completeMPU
-            if (!mpuOverviewObj) {
-                return next(null, mpuBucket, storedParts, destBucket, null, skipDataDelete);
-            }
-
             // If objectMD exists and has matching uploadId, use it directly
             // This handles all non-versioned cases, and some versioned cases.
             if (objectMD && objectMD.uploadId === uploadId) {
@@ -235,7 +230,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             return async.eachLimit(locations, 5, (loc, cb) => {
                 data.delete(loc, log, err => {
                     if (err) {
-                        log.fatal('delete ObjectPart failed', { err });
+                        log.warn('delete ObjectPart failed', { err });
                     }
                     cb();
                 });
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -3,12 +3,10 @@ const async = require('async');
 const constants = require('../../../../constants');
 const { data } = require('../../../data/wrapper');
 const locationConstraintCheck = require('../object/locationConstraintCheck');
-const { standardMetadataValidateBucketAndObj } =
-    require('../../../metadata/metadataUtils');
+const metadataUtils = require('../../../metadata/metadataUtils');
 const { validateQuotas } = require('../quotas/quotaUtils');
 const services = require('../../../services');
 const metadata = require('../../../metadata/wrapper');
-const { versioning } = require('arsenal');
 
 function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
     callback, request) {
@@ -31,7 +29,8 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
 
     async.waterfall([
         function checkDestBucketVal(next) {
-            standardMetadataValidateBucketAndObj(metadataValParams, authzIdentityResult, log,
+            // FIX: Call the function from the imported module object.
+            metadataUtils.standardMetadataValidateBucketAndObj(metadataValParams, authzIdentityResult, log,
                 (err, destinationBucket, objectMD) => {
                     if (err) {
                         log.error('error validating request', { error: err });
@@ -105,9 +104,13 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
         // supported by Metadata.
         function ensureCleanupIsRequired(mpuBucket, storedParts, destBucket,
             objectMD, skipDataDelete, next) {
+            if (!objectMD) {
+                return next(null, mpuBucket, storedParts, destBucket, null, skipDataDelete);
+            }
+
             // If objectMD exists and has matching uploadId, use it directly
             // This handles all non-versioned cases, and some versioned cases.
-            if (objectMD && objectMD.uploadId === uploadId) {
+            if (objectMD.uploadId === uploadId) {
                 return next(null, mpuBucket, storedParts, destBucket, objectMD, skipDataDelete);
             }
 
@@ -117,64 +120,21 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 return next(null, mpuBucket, storedParts, destBucket, null, skipDataDelete);
             }
 
-            // Otherwise, we list all versions of the object. We try to stop as early
-            // as possible, by using pagination.
-            let keyMarker = null;
-            let versionIdMarker = null;
-            let foundVersion = null;
-            let shouldContinue = true;
-
-            return async.whilst(
-                () => shouldContinue && !foundVersion,
-                callback => {
-                    const listParams = {
-                        listingType: 'DelimiterVersions',
-                        // To only list the specific key, we need to add the versionId separator
-                        prefix: `${objectKey}${versioning.VersioningConstants.VersionId.Separator}`,
-                        maxKeys: 1000,
-                    };
-
-                    if (keyMarker) {
-                        listParams.keyMarker = keyMarker;
-                    }
-                    if (versionIdMarker) {
-                        listParams.versionIdMarker = versionIdMarker;
-                    }
-
-                    return services.getObjectListing(bucketName, listParams, log, (err, listResponse) => {
-                        if (err) {
-                            log.error('error listing object versions', { error: err });
-                            return callback(err);
-                        }
-
-                        // Check each version in current batch for matching uploadId
-                        const matchedVersion = (listResponse.Versions || []).find(version =>
-                            version.key === objectKey &&
-                            version.value &&
-                            version.value.uploadId === uploadId
-                        );
-
-                        if (matchedVersion) {
-                            foundVersion = matchedVersion.value;
-                        }
-
-                        // Set up for next iteration if needed
-                        if (listResponse.IsTruncated && !foundVersion) {
-                            keyMarker = listResponse.NextKeyMarker;
-                            versionIdMarker = listResponse.NextVersionIdMarker;
-                        } else {
-                            shouldContinue = false;
-                        }
-
-                        return callback();
+            // Otherwise, list all versions to find one with a matching uploadId.
+            return services.findObjectVersionByUploadId(bucketName, objectKey, uploadId, log, (err, foundVersion) => {
+                if (err) {
+                    log.error('error finding object version by uploadId, proceeding without cleanup', {
+                        error: err,
+                        method: 'abortMultipartUpload.ensureCleanupIsRequired',
                     });
-                },
-                err => next(null, mpuBucket, storedParts, destBucket, err ? null : foundVersion, skipDataDelete)
-            );
+                    // On error, continue the abort without an objectMD to clean up.
+                    return next(null, mpuBucket, storedParts, destBucket, null, skipDataDelete);
+                }
+                return next(null, mpuBucket, storedParts, destBucket, foundVersion, skipDataDelete);
+            });
         },
         function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objMDWithMatchingUploadID,
             skipDataDelete, next) {
-            // If no objMDWithMatchingUploadID, nothing to delete
             if (!objMDWithMatchingUploadID) {
                 return next(null, mpuBucket, storedParts, destBucket, objMDWithMatchingUploadID, skipDataDelete);
             }
@@ -220,7 +180,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
             }
 
             // Add object data locations if they exist and we can trust uploadId matches
-            if (objMDWithMatchingUploadID && objMDWithMatchingUploadID.location) {
+            if (objMDWithMatchingUploadID?.location) {
                 const existingLocations = new Set(locations.map(loc => loc.key));
                 const remainingObjectLocations = objMDWithMatchingUploadID.
                     location.filter(loc => !existingLocations.has(loc.key));
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -101,7 +101,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
         // to honor the uploadId filter in standardMetadataValidateBucketAndObj
         // ensuring the objMD returned has the right uploadId. But this is not
         // supported by Metadata.
-        function ensureCleanupIsRequired(mpuBucket, storedParts, destBucket,
+        function findObjectToCleanup(mpuBucket, storedParts, destBucket,
             objectMD, skipDataDelete, next) {
             if (!objectMD) {
                 return next(null, mpuBucket, storedParts, destBucket, null, skipDataDelete);
@@ -124,7 +124,7 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 if (err) {
                     log.warn('error finding object version by uploadId, proceeding without cleanup', {
                         error: err,
-                        method: 'abortMultipartUpload.ensureCleanupIsRequired',
+                        method: 'abortMultipartUpload.findObjectToCleanup',
                     });
                     // On error, continue the abort without an objectMD to clean up.
                     return next(null, mpuBucket, storedParts, destBucket, null, skipDataDelete);
@@ -132,21 +132,21 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 return next(null, mpuBucket, storedParts, destBucket, foundVersion, skipDataDelete);
             });
         },
-        function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objMDWithMatchingUploadID,
+        function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objectMD,
             skipDataDelete, next) {
-            if (!objMDWithMatchingUploadID) {
-                return next(null, mpuBucket, storedParts, destBucket, objMDWithMatchingUploadID, skipDataDelete);
+            if (!objectMD) {
+                return next(null, mpuBucket, storedParts, destBucket, objectMD, skipDataDelete);
             }
 
-            log.debug('Object has existing metadata with matching uploadId, deleting them', {
+            log.debug('Object has existing metadata, deleting them', {
                 method: 'abortMultipartUpload',
                 bucketName,
                 objectKey,
                 uploadId,
-                versionId: objMDWithMatchingUploadID.versionId,
+                versionId: objectMD.versionId,
             });
             return metadata.deleteObjectMD(bucketName, objectKey, {
-                versionId: objMDWithMatchingUploadID.versionId,
+                versionId: objectMD.versionId,
             }, log, err => {
                 if (err) {
                     // Handle concurrent deletion of this object metadata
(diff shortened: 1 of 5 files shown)


lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -61,61 +62,158 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+            let foundVersion = null;
+            let shouldContinue = true;
+
+            return async.whilst(
+                () => shouldContinue && !foundVersion,
+                callback => {
+                    const listParams = {
+                        listingType: 'DelimiterVersions',
+                        // To only list the specific key, we need to add the versionId separator
+                        prefix: `${objectKey}${versioning.VersioningConstants.VersionId.Separator}`,
```
ghost: Will need further tests to ensure this works file - we do not officially have a way to list all versions of just one object

ghost: Test done, it's working well. Example:

```json
{
    "CommonPrefixes": [],
    "Versions": [
        {
            "key": "bug-test-bda30555",
            "value": {
                "Size": 1048576,
                "ETag": "3eaf2c9512f76a55da265ad51ac21393-1",
                "VersionId": "98248542529027999999RG00001",
                "IsDeleteMarker": false,
                "LastModified": "2025-07-02T11:57:50.972Z",
                "Owner": {
                    "DisplayName": "acc",
                    "ID": "5874c85ff66d0ede3bd6e8fe9610f0640d1700a4607f8b7e414503fbd2cb03de"
                },
                "StorageClass": "STANDARD"
            }
        },
        {
            "key": "bug-test-bda30555",
            "value": {
                "Size": 1048576,
                "ETag": "3eaf2c9512f76a55da265ad51ac21393-1",
                "VersionId": "98248542540139999999RG00001",
                "IsDeleteMarker": false,
                "LastModified": "2025-07-02T11:57:39.857Z",
                "Owner": {
                    "DisplayName": "acc",
                    "ID": "5874c85ff66d0ede3bd6e8fe9610f0640d1700a4607f8b7e414503fbd2cb03de"
                },
                "StorageClass": "STANDARD"
            }
        }
    ],
    "IsTruncated": false
}
```

(while the same key with suffix exists)

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -61,61 +62,158 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+        // an object version when some of the parts are missing.
+        // By passing a null objectMD, we tell the subsequent steps
+        // to skip the cleanup.
+        // Another approach is possible, but not supported by all backends:
+        // to honor the uploadId filter in standardMetadataValidateBucketAndObj
+        // ensuring the objMD returned has the right uploadId. But this is not
+        // supported by Metadata.
+        function ensureCleanupIsRequired(mpuBucket, mpuOverviewObj, storedParts, destBucket,
+            objectMD, skipDataDelete, next) {
+            // If no overview object, skip - this wasn't a failed completeMPU
```
DarkIsDude: Comment describe the code it self, is it useful ? Also how can we be in that case ?

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -61,61 +62,158 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+                            keyMarker = listResponse.NextKeyMarker;
+                            versionIdMarker = listResponse.NextVersionIdMarker;
+                        } else {
+                            shouldContinue = false;
+                        }
+
+                        return callback();
+                    });
+                },
+                err => next(null, mpuBucket, storedParts, destBucket, err ? null : foundVersion, skipDataDelete)
```
DarkIsDude: Why the err is always null here? If we have an err `services.getObjectListing` we should propagate ?

DarkIsDude: Ok, I have my answer directly in the JIRA description 🙏

tests/functional/aws-node-sdk/test/object/abortMPU.js
```diff
@@ -25,18 +25,20 @@ describe('Abort MPU', () => {
...
-                uploadId = res.UploadId;
-                return s3.uploadPart({ Bucket: bucket, Key: key,
-                    PartNumber: 1, UploadId: uploadId, Body: bodyFirstPart,
-                }).promise();
-            })
-            .catch(err => {
-                process.stdout.write(`Error in beforeEach: ${err}\n`);
-                throw err;
-            });
+                .then(() => s3.createMultipartUpload({
```
DarkIsDude, line 28: We should ensure a format in the CI I guess ?

tests/functional/aws-node-sdk/test/object/abortMPU.js
```diff
@@ -273,16 +276,221 @@ describe('Abort MPU - No Such Upload', () => {
...
+                    Bucket: bucketName,
+                    Key: upload.Key,
+                    UploadId: upload.UploadId,
+                }).promise().catch(err => {
+                    if (err.code !== 'NoSuchUpload') throw err;
+                })
+            ));
+
+            // Clean up all object versions
+            const listVersionsResponse = await s3.listObjectVersions({ Bucket: bucketName }).promise();
```
DarkIsDude, line 333: We can DRY :

```
const allObjects = [
    ...listVersionsResponse.Versions,
    ...listVersionsResponse.DeleteMarkers,
];
```

tests/unit/api/apiUtils/object/abortMultipartUpload.spec.js
```diff
@@ -0,0 +1,677 @@
...
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    // Helper to create bucket and MPU, returns uploadId
+    function createBucketAndMPU(versioned, callback) {
+        if (versioned) {
+            bucketPut(authInfo, bucketRequest, log, err => {
```
DarkIsDude: Seems like a hell callback here ? Why not async.waterfall then ? You can also DRY and add condition in the waterfall for bucketVersioning ?

tests/unit/api/apiUtils/object/abortMultipartUpload.spec.js
```diff
@@ -0,0 +1,677 @@
...
+                    calculatedHash,
+                }, postBody);
+
+                objectPutPart(authInfo, partRequest, undefined, log, err => {
+                    assert.ifError(err);
+
+                    abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log, err => {
+                        assert.strictEqual(err, null);
+                        sinon.assert.calledOnce(dataAbortMPUStub);
+                        sinon.assert.called(dataDeleteStub); // Should delete part data
```
DarkIsDude: Does it make sense to check with which parameter we call it (just to be sure that we deleted the right part?)

tests/unit/api/apiUtils/object/abortMultipartUpload.spec.js
```diff
@@ -0,0 +1,677 @@
...
+
+                abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log, err => {
+                    assert(err);
+                    assert.strictEqual(err.is.InternalError, true);
+                    done();
+                }, abortRequest);
+            });
+        });
+
+        it('should continue despite data deletion errors', done => {
```
DarkIsDude: I don't know if it make sense here but the log is very important here (as we have a silent error). We want 1. to be sure that we continue even if we have an error 2. but also to have a log to be able to debug and understand. A "good" practice can be to check that we called the log function. We can avoid a refactor later that will mute this error. It's not useful to test all logs but I think here it's a good one ?

tests/functional/aws-node-sdk/test/object/abortMPU.js
ghost: I'm not sure to understand your comment?

DarkIsDude: My point is that you are only "reindenting" the code here, we should add something in the CI to make sure we are consistent and everyone apply the same rule. I guess it's out of scope of the PR, so you can ignore I guess

tests/unit/api/apiUtils/object/abortMultipartUpload.spec.js
ghost: Simplfied: 

https://github.com/scality/cloudserver/pull/5854/commits/0a1b5f0c2c58101e94c2dc0de859cce9afdb7d07

lib/api/apiUtils/object/abortMultipartUpload.js
ghost: I removed this, indeed, leftover from my tests, this variable is always set, unless we have an error.

tests/unit/api/apiUtils/object/abortMultipartUpload.spec.js
ghost: We can, yes, just a side note: logging below the "logLevel", unless the **same** request hit a `dumpLevel` log **after** (defined as `error` mostly, used to dump all accumulated logs in that request) will not output anything. If we return no error and complete the API properly, then we will not have that error in the production log, unless we change logging levels (which we try to avoid).

Here it seems we had a fatal log, which is also incorrect, the fatal logging level should come with a service crash, it should be a warning (which will be logged directly).

ghost: https://github.com/scality/cloudserver/pull/5854/commits/9880217c0defd0094079976e1fb7b0c2556a2249

tests/functional/aws-node-sdk/test/object/abortMPU.js
ghost: I think it's just because we have this in the repo eslint config:

```
        "indent": "off",
```

Actually we were supposed to bump eslint in this repo to v9, like in others, so that should be done at the same time. I saw we have no ticket so I'll create one, maybe do it right away... We should use a eslint.config.mjs file now.

ghost: Ah no, it's just not in dev / 8.8 branch, but it is in the 9.0+. I will just see to refine the rule for these branches.

ghost: See https://github.com/scality/cloudserver/pull/5877

DarkIsDude: 🙏 thanks

.github/workflows/tests.yaml
```diff
@@ -519,10 +519,10 @@ jobs:
       uses: ./.github/actions/setup-ci
     - uses: ruby/setup-ruby@v1
       with:
-        ruby-version: '2.5.9'
```
DarkIsDude: I'm not sure to understand why it's in the diff here and now ?

DarkIsDude (approved):

.github/workflows/tests.yaml
ghost: Ruby version is not working anymore - we did a fix but on 9.0+ only, I'm just cherry-picking it to pass the CI in 8.8 (the branch I'm targeting here)

francoisferrand: did we not merge this fix already (i.e. this commit will be dropped by rebasing)?

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -90,32 +91,124 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+            }
+
+            // Otherwise, we list all versions of the object. We try to stop as early
+            // as possible, by using pagination.
+            let keyMarker = null;
+            let versionIdMarker = null;
+            let foundVersion = null;
+            let shouldContinue = true;
+
+            return async.whilst(
```
francoisferrand: this function is already quite big, would make code more readable to introduce a helper function like `findVersionWithUploadId()`

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -90,32 +91,124 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
-            if (!objectMD || metadataValMPUparams.uploadId !== objectMD.uploadId) {
+        // During Abort, we dynamically detect if the previous CompleteMPU call
+        // created potential object metadata wrongly, e.g. by creating
+        // an object version when some of the parts are missing.
+        // By passing a null objectMD, we tell the subsequent steps
+        // to skip the cleanup.
+        // Another approach is possible, but not supported by all backends:
+        // to honor the uploadId filter in standardMetadataValidateBucketAndObj
+        // ensuring the objMD returned has the right uploadId. But this is not
+        // supported by Metadata.
```
francoisferrand, line 103: this can be combined : we can (in specific backend) pass the `uploadId` filter in `standardMetadataValidateBucketAndObj`, so that a backend _could_ does the search more efficiently.... though indeed we also need to implement the fallback code like you did, for backend(s) which do not implement the filter.

This also makes me wonder if this is the right place to implement : i.e. if we pass `uploadId` filter, then we could "abstract" this listing in a lower layer (maybe in `MetadataWrapper` ?) so that this feature is available for all, and we focus only on "MPU" logic in here (delete MD, delete data, ...)

Can you check if it'd make sense and make code cleaner/more readable? What do you think?

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -90,32 +91,124 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+        function ensureCleanupIsRequired(mpuBucket, storedParts, destBucket,
+            objectMD, skipDataDelete, next) {
+            // If objectMD exists and has matching uploadId, use it directly
+            // This handles all non-versioned cases, and some versioned cases.
+            if (objectMD && objectMD.uploadId === uploadId) {
```
francoisferrand: if we did not find objectMD, (`if !objectMD`), do we really need to do the listing? should we not just directly go to the next stage?

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -126,16 +219,18 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
                 return next(null, mpuBucket, storedParts, destBucket);
             }
 
-            if (objectMD && objectMD.location && objectMD.uploadId === metadataValMPUparams.uploadId) {
+            // Add object data locations if they exist and we can trust uploadId matches
+            if (objMDWithMatchingUploadID && objMDWithMatchingUploadID.location) {
```
francoisferrand: nit:
```suggestion
            if (objMDWithMatchingUploadID?.location) {
```

.github/workflows/tests.yaml
ghost: Not for this branch, as I'm targeting 8.8 here.

lib/api/apiUtils/object/abortMultipartUpload.js
ghost: I agree for the code location, the right place here should be the `services.js` file, like for other MD-calls from s3 APIs. It definitely simplifies the code in the abortMPU logic. It's also good to keep that in a place that can be reused in other APIs if needed. I don't think that should be moved to the MetadataWrapper, because we add the pagination logic that is not always wanted, and this has some s3-specific logic. But there will need to be some cleanup in the `services.js` vs `MetadataWrapper` abstractions for sure...

However, I kind of disagree with the first part: it will make the code hard to follow/debug and maintain. We would have to support two different behaviors depending on the backend, and given "multiplebackend" cases, it might even require more work to ensure it works for all. Plus, I believe it goes against the effort of unifying the logic. Instead, I think we should first create the support for uploadId in Metadata and then we can "optimize" (though I don't think this is really needed here). What do you think? if you align, I will create a ticket.

lib/api/apiUtils/object/abortMultipartUpload.js
ghost: I moved it to `services.js`, added tests for it, and refactored as a result the existing ones.

lib/api/apiUtils/object/abortMultipartUpload.js
francoisferrand: > However, I kind of disagree with the first part: it will make the code hard to follow/debug and maintain. We would have to support two different behaviors depending on the backend, and given "multiplebackend" cases, it might even require more work to ensure it works for all. Plus, I believe it goes against the effort of unifying the logic. Instead, I think we should first create the support for uploadId in Metadata and then we can "optimize" (though I don't think this is really needed here). What do you think? if you align, I will create a ticket.

we can create a ticket in Metadata (and it may not be right place to handle in MetadataWrapper, I did not look into enough details), but my point is more generic here:
* "Unifying" does not mean there is a single code path. That would be ideal obviously, but practically there will always be differences.... And thus there will always be slight difference in behaviors to handle. So it is really not a fundamental issue.
* Just like we have different implementations of Metadata backend (behind MetadataWrapper), and like MetadataWrapper knows how to adapt to features of different backend (like calling a backend's copyObject when available, and falling back to 
* Generally, we want to avoid coupling between different backends to allow development to flow freely. Supporting different behaviors (even temporarily) is a cheap solution to do this, and allow improving one backend before the others without breaking the code.

That said, we are not there yet : so we can proceed without "preparing" for a more optimal implementation ; and we will update cloudserver and reconsider if we can/need to sync changes in all backends or need to support both behaviors when we work on this change.

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -93,29 +92,82 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+        // During Abort, we dynamically detect if the previous CompleteMPU call
+        // created potential object metadata wrongly, e.g. by creating
+        // an object version when some of the parts are missing.
+        // By passing a null objectMD, we tell the subsequent steps
+        // to skip the cleanup.
+        // Another approach is possible, but not supported by all backends:
+        // to honor the uploadId filter in standardMetadataValidateBucketAndObj
+        // ensuring the objMD returned has the right uploadId. But this is not
+        // supported by Metadata.
+        function ensureCleanupIsRequired(mpuBucket, storedParts, destBucket,
```
francoisferrand: nit: `findObjectToCleanup` ?

lib/api/apiUtils/object/abortMultipartUpload.js
```diff
@@ -93,29 +92,82 @@ function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
...
+                        error: err,
+                        method: 'abortMultipartUpload.ensureCleanupIsRequired',
+                    });
+                    // On error, continue the abort without an objectMD to clean up.
+                    return next(null, mpuBucket, storedParts, destBucket, null, skipDataDelete);
+                }
+                return next(null, mpuBucket, storedParts, destBucket, foundVersion, skipDataDelete);
+            });
+        },
+        function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objMDWithMatchingUploadID,
```
francoisferrand: nit: should we keep the old name of the parameter, to make the diff behave better (and see we are really just introducing an extra step):

```suggestion
        function deleteObjectMetadata(mpuBucket, storedParts, destBucket, objectMD,
```

(also the mention of `uploadID` later making things confusing, since we don't check it there : IMHO the step you are adding makes the code correct by "fixing" the `objectMD` where needed, so that throughout we can now trust the `objectMD`.... so the problem is abstracted a way from that point, and we don't need to make the code look more complicated)

lib/services.js
```diff
@@ -418,6 +419,71 @@ const services = {
...
+     * @returns {undefined}
+     */
+    findObjectVersionByUploadId(bucketName, objectKey, uploadId, log, cb) {
+        let keyMarker = null;
+        let versionIdMarker = null;
+        let foundVersion = null;
+        let shouldContinue = true;
+
+        return async.whilst(
+            () => shouldContinue && !foundVersion,
```
francoisferrand, line 439: nit: I wonder if we may optimize this a bit, to avoid the worst case scenario when there are **many** versions, i.e.
- start listing from the `versionID` of the objectMD we already found (since it should have been created "before" the one we found? or can we find a more recent version in some weird racy scenario?)
- stop when the version we list is older than the MPU document?

not sure it is worth it or possible, not blocking, but I was wondering

tests/functional/aws-node-sdk/test/object/abortMPU.js
```diff
@@ -25,18 +25,21 @@ describe('Abort MPU', () => {
...
-                }).promise();
-            })
-            .catch(err => {
-                process.stdout.write(`Error in beforeEach: ${err}\n`);
-                throw err;
-            });
+                .then(() => s3.createMultipartUpload({
+                    Bucket: bucket,
+                    Key: key,
+                }).promise())
```
francoisferrand, line 31: should this not be indented? (it is the parameters of the "inner" function, clang-format would indent; but we don't have a defined formatter to answer the question)

tests/functional/aws-node-sdk/test/object/abortMPU.js
```diff
@@ -45,23 +48,24 @@ describe('Abort MPU', () => {
...
+            done => {
+                s3.abortMultipartUpload({
+                    Bucket: bucket,
+                    Key: '',
+                    UploadId: uploadId
+                },
+                    err => {
+                        checkError(err, 'InvalidRequest', 'A key must be specified');
+                        done();
+                    });
```
francoisferrand, line 67: better like this?
```suggestion
                }, err => {
                    checkError(err, 'InvalidRequest', 'A key must be specified');
                    done();
                });
```

(same at line 277, if you agree)

tests/functional/aws-node-sdk/test/object/abortMPU.js
```diff
@@ -273,16 +277,218 @@ describe('Abort MPU - No Such Upload', () => {
...
+                        assert.notEqual(err, null, 'Expected failure but got success');
+                        assert.strictEqual(err.code, 'NoSuchUpload');
+                        done();
+                    });
             });
+    });
+});
+
+describe('Abort MPU - Versioned Bucket Cleanup', function testSuite() {
+    this.timeout(120000);
```
francoisferrand: 2 minutes is a long time, do we really need this much?

tests/unit/lib/services.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+                const listParams = getObjectListingStub.getCall(0).args[1];
+                const expectedPrefix = `${objectKey}${VersionId.Separator}`;
+                assert.strictEqual(listParams.prefix, expectedPrefix);
+                assert.strictEqual(listParams.listingType, 'DelimiterVersions');
+                done();
+            });
+        });
+
+        it('should correctly handle pagination', done => {
```
francoisferrand: this test is more "should read all pages", but it does not fully check we correctly handle pagination...
e.g. we have have other tests to check that we don't call later pages if we find the entry we are looking for, or that we are able indeed to find an object on a later page...

(I can imagine 5 cases : not found, fond on only page, found on first page of many, found on last page of many, found on "middle" page of many... which could be reduced based on the conditions in the code, but there are a bit too many nested variables/lambdas for me to be completely sure)

francoisferrand (approved):

lib/services.js
ghost: I created CLDSRV-726, the "racy" analysis will be a whole work in itself, better to split it. I would say it is likely possible, but that needs more testing.

tests/functional/aws-node-sdk/test/object/abortMPU.js
ghost: https://github.com/scality/cloudserver/pull/5877 will solve that.
For now I trust my IDE prettier, and it indents like this (which is logical to me).

tests/functional/aws-node-sdk/test/object/abortMPU.js
ghost: Switched to 30s. It's more a safe config to ensure test passes if the CI is momentarily slower.

tests/unit/lib/services.spec.js
ghost: I added these tests.

PR #5856, opened 2025-07-03
Improve handling of errors in CompleteMPU to make the API really idempotent

ghost:
Relates to https://scality.atlassian.net/wiki/spaces/OS/pages/3258220958/Retriable+CompleteMultipartUpload

The original problem, reported by one customer, was that they got non retryable

errors during CompleteMPU calls. In this case, their client would follow up
with a call to the AbortMPU API, logically.

We made a mistake during the last bugfix:

- We incorrectly thought that the reason of the error was a failure during
  the extra part deletion. And fixed it by cleaning everything during AbortMPU.
  But this is wrong: when an error happens in this step of the CompleteMPU API,
  it means the S3 object was properly created AND the overview keys and parts
  metadata were already cleaned up. Aborting in such case would just return
  a 404 error, but won't do anything. Instead, the state that the customer
  encountered was different: they failed at the `deletePartsMetadata` step.
  This is something we support with the `CompleteMultiPartUpload` retry logic.

The clean up in the AbortMPU API is actually great: when we end up in this
situation, the client has two available (and valid) choices:

- They retry the CompleteMultiPartUpload.
- They call the AbortMPU API.

The latter is now possible, since the cleanup introduction. In this case, we
detect that the object is in an inconsistent state and proceed with the
removal of everything (both in the S3 bucket and the mpuShadowBucket), so
the result is consistent: the MPU is aborted and the customer must re-upload
the whole object.

The first option is preferred, as it keeps the object in the bucket and avoids
the need to re-upload.

But the problem here is that we do not guarantee that a retryable error is
returned to the client: in case of MD update failure, the returned error
might not be retryable: we can have DeleteConflict or NoSuchKey errors.
Other usually are retryable (InternalError, etc).

So we want to make sure of two behaviors:
- In the batchDeleteObjectMetadata error callback case, we should always
  return a retryable error. It means that in case of a DeleteConflict, we
  should always retry, and in case of a NoSuchKey, we should not retry.
- During the batchDeleteExtraParts, we already cleaned up the mpu bucket,
  so returning an error to the client if the operation fails will wrongly
  lead to a completeMPU retry or ABort, while the operation succeeded, and
  we only created ghosts. In any case, the ghosts are permanent: without
  the mpu bucket, we cannot clean them in subsequent calls.

Issue: CLDSRV-669

### DIFF
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -524,15 +524,64 @@ function completeMultipartUpload(authInfo, request, log, callback) {
         function deletePartsMetadata(mpuBucket, keysToDelete, aggregateETag,
             extraPartLocations, destinationBucket, generatedVersionId, droppedMPUSize, next) {
             services.batchDeleteObjectMetadata(mpuBucket.getName(),
-                keysToDelete, log, err => next(err, extraPartLocations,
-                    destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize));
+                keysToDelete, log, err => {
+                    if (err) {
+                        log.error('error deleting MPU metadata during completeMPU', {
+                            method: 'completeMultipartUpload',
+                            mpuBucket: mpuBucket.getName(),
+                            keysToDeleteCount: keysToDelete.length,
+                            error: err,
+                        });
+
+                        // Handle specific error cases according to retry strategy
+                        if (err.is?.DeleteConflict) {
+                            // DeleteConflict should trigger automatic retry
+                            // Convert to InternalError to make it retryable
+                            return next(errors.InternalError, extraPartLocations,
+                                destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize);
+                        }
+
+                        // For NoSuchKey and other errors, return them as-is
+                        // NoSuchKey is non-retryable, InternalError and others are retryable
+                        return next(err, extraPartLocations,
+                            destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize);
+                    }
+
+                    log.debug('successfully deleted MPU metadata during completeMPU', {
+                        method: 'completeMultipartUpload',
+                        mpuBucket: mpuBucket.getName(),
+                        keysToDeleteCount: keysToDelete.length,
+                    });
+                    return next(null, extraPartLocations,
+                        destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize);
+                });
         },
         function batchDeleteExtraParts(extraPartLocations, destinationBucket,
             aggregateETag, generatedVersionId, droppedMPUSize, next) {
             if (extraPartLocations && extraPartLocations.length > 0) {
                 return data.batchDelete(extraPartLocations, request.method, null, log, err => {
                     if (err) {
-                        return next(err);
+                        // Extra part deletion failure should not fail the operation
+                        // The S3 object was created successfully and MPU metadata was cleaned up
+                        // Orphaned extra parts are acceptable since the main operation succeeded
+                        log.warn('failed to delete extra parts, keeping orphan but returning success', {
+                            method: 'completeMultipartUpload',
+                            extraPartLocationsCount: extraPartLocations.length,
+                            error: err,
+                        });
+
+                        // Continue with quota validation but don't fail if it errors either
+                        return validateQuotas(request, destinationBucket, request.accountQuotas,
+                            ['objectDelete'], 'objectDelete', -droppedMPUSize, false, log, quotaErr => {
+                                if (quotaErr) {
+                                    log.warn('failed to update inflights after extra part deletion failure', {
+                                        method: 'completeMultipartUpload',
+                                        error: quotaErr,
+                                    });
+                                }
+                                // Always succeed - the core operation was successful
+                                return next(null, destinationBucket, aggregateETag, generatedVersionId);
+                            });
                     }
 
                     return validateQuotas(request, destinationBucket, request.accountQuotas,
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -569,39 +569,28 @@ function completeMultipartUpload(authInfo, request, log, callback) {
                             extraPartLocationsCount: extraPartLocations.length,
                             error: err,
                         });
-
-                        // Continue with quota validation but don't fail if it errors either
-                        return validateQuotas(request, destinationBucket, request.accountQuotas,
-                            ['objectDelete'], 'objectDelete', -droppedMPUSize, false, log, quotaErr => {
-                                if (quotaErr) {
-                                    log.warn('failed to update inflights after extra part deletion failure', {
-                                        method: 'completeMultipartUpload',
-                                        error: quotaErr,
-                                    });
-                                }
-                                // Always succeed - the core operation was successful
-                                return next(null, destinationBucket, aggregateETag, generatedVersionId);
-                            });
                     }
-
-                    return validateQuotas(request, destinationBucket, request.accountQuotas,
-                        ['objectDelete'], 'objectDelete', -droppedMPUSize, false, log, err => {
-                            if (err) {
-                                // Ignore error, as the data has been deleted already: only inflight count
-                                // has not been updated, and will be eventually consistent anyway
-                                log.warn('failed to update inflights', {
-                                    method: 'completeMultipartUpload',
-                                    extraPartLocations,
-                                    error: err,
-                                });
-                            }
-                            return next(null, destinationBucket, aggregateETag,
-                                generatedVersionId);
-                    });
+                    return next(null, destinationBucket, aggregateETag,
+                        generatedVersionId, droppedMPUSize);
                 });
             }
             return next(null, destinationBucket, aggregateETag,
-                generatedVersionId);
+                generatedVersionId, droppedMPUSize);
+        },
+        function updateQuotas(destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize, next) {
+            return validateQuotas(request, destinationBucket, request.accountQuotas,
+                ['objectDelete'], 'objectDelete', -droppedMPUSize, false, log, err => {
+                    if (err) {
+                        // Ignore error, as the data has been deleted already: only inflight count
+                        // has not been updated, and will be eventually consistent anyway
+                        log.warn('failed to update inflights', {
+                            method: 'completeMultipartUpload',
+                            error: err,
+                        });
+                    }
+                    return next(null, destinationBucket, aggregateETag,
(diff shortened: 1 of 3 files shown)


lib/api/completeMultipartUpload.js
```diff
@@ -524,15 +524,64 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                        // The S3 object was created successfully and MPU metadata was cleaned up
+                        // Orphaned extra parts are acceptable since the main operation succeeded
+                        log.warn('failed to delete extra parts, keeping orphan but returning success', {
+                            method: 'completeMultipartUpload',
+                            extraPartLocationsCount: extraPartLocations.length,
+                            error: err,
+                        });
+
+                        // Continue with quota validation but don't fail if it errors either
+                        return validateQuotas(request, destinationBucket, request.accountQuotas,
```
jonathan-gramain: Is it doable to make the call to `validateQuotas` an extra step in the waterflow, to avoid duplicating the code? I don't think we really care that the log doesn't specify `after extra part deletion failure` when failing to update quotas.

tests/unit/api/multipartUpload.js
```diff
@@ -2070,6 +2070,176 @@ describe('Multipart Upload API', () => {
...
+            next => bucketPut(authInfo, bucketPutRequest, log, next),
+            (corsHeaders, next) => initiateMultipartUpload(authInfo,
+                initiateRequest, log, next),
+            (result, corsHeaders, next) => parseString(result, next),
+        ],
+        (err, json) => {
+            assert.ifError(err);
+            const testUploadId = json.InitiateMultipartUploadResult.UploadId[0];
+            const md5Hash = crypto.createHash('md5').update(partBody);
+            const calculatedHash = md5Hash.digest('hex');
```
jonathan-gramain:
```suggestion
            const partHash = crypto.createHash('md5').update(partBody).digest('hex');
            // then use `partHash` directly
```

jonathan-gramain (approved):

lib/api/completeMultipartUpload.js
ghost: Yes it is, now that we do not return any error, I applied the suggestion :slightly_smiling_face:

ghost:
Rebased the branch following base branch's edits.

lib/api/completeMultipartUpload.js
```diff
@@ -524,35 +524,73 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                            // DeleteConflict should trigger automatic retry
+                            // Convert to InternalError to make it retryable
+                            return next(errors.InternalError, extraPartLocations,
+                                destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize);
+                        }
+
+                        // For NoSuchKey and other errors, return them as-is
+                        // NoSuchKey is non-retryable, InternalError and others are retryable
+                        return next(err, extraPartLocations,
+                            destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize);
```
DarkIsDude, line 540:
```suggestion
                        if (err.is?.DeleteConflict) {
                            // DeleteConflict should trigger automatic retry. Convert to InternalError to make it retryable
                            err = errors.InternalError
                        }

                        return next(err, extraPartLocations,
                            destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize);
```

An another way is to name variable : `const retryableErrors = errors.InternalError; err = retryableErrors`.

tests/unit/api/multipartUpload.js
```diff
@@ -2070,6 +2066,175 @@ describe('Multipart Upload API', () => {
...
+                partHash,
+            }, partBody);
+
+            objectPutPart(authInfo, partRequest, undefined, log, () => {
+                // Mock batchDeleteObjectMetadata to fail with non-retryable error
+                const services = require('../../../lib/services');
+                batchDeleteStub = sinon.stub(services, 'batchDeleteObjectMetadata')
+                    .callsFake((mpuBucketName, keysToDelete, log, cb) =>
+                        // Simulate a non-retryable error that should be converted to retryable
+                         cb(errors.NoSuchKey)
```
DarkIsDude, line 2103:
```suggestion
                const services = require('../../../lib/services');
                batchDeleteStub = sinon.stub(services, 'batchDeleteObjectMetadata')
                    .callsFake((mpuBucketName, keysToDelete, log, cb) =>
                         cb(errors.NoSuchKey)
```

Your code is clear enough

tests/unit/api/multipartUpload.js
```diff
@@ -2070,6 +2066,175 @@ describe('Multipart Upload API', () => {
...
+                };
+
+                completeMultipartUpload(authInfo, completeRequest, log, err => {
+                    // Restore original function
+                    batchDeleteStub.restore();
+
+                    // Should get an error (retryable behavior)
+                    assert(err, 'Expected an error when metadata deletion fails');
+
+                    // Verify S3 object was created successfully despite the error
```
DarkIsDude, line 2131:
```suggestion
                    batchDeleteStub.restore();
                    assert(err, 'Expected an error when metadata deletion fails');
```

Same here IMO. A nice clean code comment mindset https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#comments-rules if you want more. I'll report it only here to avoid noise on other comments in the tests 🙏

DarkIsDude (approved):

tests/unit/api/multipartUpload.js
ghost: Yes, TBH most of the unit tests are being generated with the help of ai - quite convenient given we have a large code base serving as examples, I'm mostly ensuring we are not forgetting any scenario, and that they test what they are supposed to. I'll refine my rules here, because indeed, no need to tell what the code is doing.

lib/api/completeMultipartUpload.js
```diff
@@ -524,35 +524,73 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                            method: 'completeMultipartUpload',
+                            mpuBucket: mpuBucket.getName(),
+                            keysToDeleteCount: keysToDelete.length,
+                            error: err,
+                        });
+
+                        // Handle specific error cases according to retry strategy
+                        if (err.is?.DeleteConflict) {
+                            // DeleteConflict should trigger automatic retry
+                            // Convert to InternalError to make it retryable
```
francoisferrand, line 532: could we not instead "make" the DeleteConflict error retryable?

i.e. it seems to me it would overall be better to still return the error (deleteConflict), but update it's definition so that is returnable (but of course this requires that
1. we always want to be retryable on DeleteConflict;
2. "retryable" can be fully specified on our side -i.e. arsenal- and does not depend on client/sdk side logic which hard-codes part of the retryable decision

lib/api/completeMultipartUpload.js
```diff
@@ -524,35 +524,73 @@ function completeMultipartUpload(authInfo, request, log, callback) {
         function deletePartsMetadata(mpuBucket, keysToDelete, aggregateETag,
             extraPartLocations, destinationBucket, generatedVersionId, droppedMPUSize, next) {
             services.batchDeleteObjectMetadata(mpuBucket.getName(),
-                keysToDelete, log, err => next(err, extraPartLocations,
-                    destinationBucket, aggregateETag, generatedVersionId, droppedMPUSize));
+                keysToDelete, log, err => {
+                    if (err) {
+                        log.error('error deleting MPU metadata during completeMPU', {
```
francoisferrand: do we need all these extra logs?
we already have layers of logs, is this not redundant?

lib/api/completeMultipartUpload.js
ghost: > "retryable" can be fully specified on our side -i.e. arsenal- and does not depend on client/sdk side logic which hard-codes part of the retryable decision

I am not sure about that? Can you elaborate on why you think we can define it?

To me we cannot make it retryable, as this behavior is implemented in the AWS client itself:

> Transient errors (HTTP status codes 400, 408, 500, 502, 503, and 504) and throttling errors (HTTP status codes 400, 403, 429, 502, 503, and 509) can all potentially be retried. SDK retry behavior is determined in combination with error codes or other data from the service.

While:
> DeleteConflict
The request was rejected because it attempted to delete a resource that has attached subordinate entities. The error message describes these entities.
HTTP Status Code: 409

I cannot find anything standard to define it on our side...

lib/api/completeMultipartUpload.js
ghost: Removed, I just kept the warn ones.

lib/api/completeMultipartUpload.js
ghost: We could instead specify a description for this error, and keep the error 500. The thing is, the DeleteConflict error already isn't quite right in this context, as the complete does not delete anything, it's internal, and the error message is misleading in this context.

Second note, the retryable abort/complete is not really standard either, this is something we support historically in the RING and it's quite convenient for clients actually, this approach wants to keep that aspect, but ensure the clients will not end up in a wrong state due to conflicting error codes returned, that does not match the behavior of our API.

Also, I don't think we should make DeleteConflict retryable at all, as this would cause problems in most of the APIs, in Vault too.

francoisferrand: indeed, should probably not be a `DeleteConflict` in the first place, and the "standard" `DeleteConflict` should not be retriable.

But then should we instead create another 500 error in Arsenal (with a different type & custom description) and use it instead of DeleteConflict, so that t will be retried automatically (without re-mapping here), but we can still precisely detect the error in our code, where needed (since we have a dedicated arsenal type)

ghost: What would be the benefit of introducing a non standard error code? As long as we can control the description, then InternalError is enough, and we can add a bit more context?

If we really want to detect it, then we can introduce a debug log in this case, so that it's dumped when the api logs with error log?

francoisferrand: I did not mean a non-standard error _code_, I meant a new error _type_ (in arsenal errors), so that we can (in _our code_!) identify this specific case if/when we need to react to this specific case elsewhere (like in cloudserver) : i.e. to allow the upper layers of the code to precisely identify this specific error generated deep down in mongoClientInterface.
But this may not be needed if we don't do have any such custom handling of this error.

My point is really that if we don't generate the right error, then it is best to fix the "root" of the issue (i.e. the place where we incorrectly return a `DeleteConflict` error) instead of remapping in upper layers.

ghost: I see. I don't this this is necessary as, as you said, we are not making anything of this error, beyond returning something retryable. The error is valid, as we use a generic function, it is just illogical in the context of that particular API to return it, given how the API behaves. The current code will be self-sufficient to me :slightly_smiling_face:

PR #5859, opened 2025-07-17
 CLDSRV-716 Disable default timeout

benzekrimaha:
Issue: CLDSRV-716

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -248,6 +248,11 @@ class S3Server {
             });
         }
 
+        // Starting NodeJS v18, the default timeout, when `undefined`, is
+        // 5 minutes. We must set the value to zero to allow for long
+        // upload durations.
+        server.requestTimeout = 0; // disabling request timeout
+
         server.on('connection', socket => {
             socket.on('error', err => logger.info('request rejected',
                 { error: err }));
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zenko/cloudserver",
-  "version": "9.0.19",
+  "version": "9.0.20",
   "description": "Zenko CloudServer, an open-source Node.js implementation of a server handling the Amazon S3 protocol",
   "main": "index.js",
   "engines": {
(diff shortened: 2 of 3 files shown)


lib/server.js
```diff
@@ -248,6 +248,8 @@ class S3Server {
             });
         }
 
+        server.requestTimeout = 0; // disabling request timeout
```
ghost, line 254:
```suggestion
        // Starting NodeJS v18, the default timeout, when `undefined`, is
        // 5 minutes. We must set the value to zero to allow for long
        // upload durations.
        server.requestTimeout = 0; // disabling request timeout
```

welansari (approved):

lib/server.js
```diff
@@ -248,6 +248,8 @@ class S3Server {
...
+
```
ghost, line 255: Then I suggest adding a unit test:

```js
const sinon = require('sinon');
const http = require('http');
const promClient = require('prom-client');

describe('S3Server request timeout', () => {
    let sandbox;
    let mockServer;

    beforeEach(() => {
        sandbox = sinon.createSandbox();
        
        // Clear prometheus registry to avoid conflicts
        promClient.register.clear();
        
        // Create a mock server to capture the requestTimeout setting
        mockServer = {
            requestTimeout: null,
            on: sandbox.stub(),
            listen: sandbox.stub(),
            address: sandbox.stub().returns({ address: '127.0.0.1', port: 8000 }),
        };
        
        // Mock server creation to return our mock
        sandbox.stub(http, 'createServer').returns(mockServer);
        sandbox.stub(https, 'createServer').returns(mockServer);
    });

    afterEach(() => {
        sandbox.restore();
    });

    it('should set server.requestTimeout to 0 when starting server', () => {
        const { S3Server } = require('../../lib/server');
        const server = new S3Server();
        
        // Call _startServer which should set requestTimeout = 0
        server._startServer(() => {}, 8000, '127.0.0.1');
        
        // Verify that requestTimeout was set to 0
        assert.strictEqual(mockServer.requestTimeout, 0);
    });
});
```

You'll need to also export the s3Server class from server.js

```js
module.exports.S3Server = S3Server;
```

I think this is important to avoid missing such issue again in the future.

ghost (approved):

benzekrimaha:
@bert-e create_pull_requests

PR #5865, opened 2025-07-17
CLDSRV-718: fix debian apt-get urls

leif-scality:
fix `apt-get` broken urls observed here https://github.com/scality/cloudserver/actions/runs/16341465257?pr=5855


leif-scality:
moved to https://github.com/scality/cloudserver/pull/5867

PR #5866, opened 2025-07-17
set 20kb payload limit on putBucketPolicy

SylvainSenechal:
Issue: [CLDSRV-700](https://scality.atlassian.net/browse/CLDSRV-700)

[CLDSRV-700]: https://scality.atlassian.net/browse/CLDSRV-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config.json
+++ b/config.json
@@ -142,5 +142,9 @@
     },
     "kmip": {
         "providerName": "thales"
+    },
+    "apiBodySizeLimits": {
+        "multiObjectDelete": 2097152,
+        "bucketPutPolicy": 20480
     }
 }
--- a/constants.js
+++ b/constants.js
@@ -96,11 +96,15 @@ const constants = {
     oneMegaBytes: 1024 * 1024,
     halfMegaBytes: 512 * 1024,
 
-    // Some apis may need a custom body length limit :
-    apisLengthLimits: {
+    // Some apis may need a custom body length limit
+    // Caution : Users will be able to override these values in config.json
+    defaultApiBodySizeLimits: {
         // Multi Objects Delete request can be large : up to 1000 keys of 1024 bytes is
         // already 1mb, with the other fields it could reach 2mb
         'multiObjectDelete': 2 * 1024 * 1024,
+        // AWS sets the maximum size for bucket policies to 20 KB
+        // https://docs.aws.amazon.com/AmazonS3/latest/userguide/add-bucket-policy.html
+        'bucketPutPolicy': 20 * 1024,
     },
 
     // hex digest of sha256 hash of empty string:
@@ -265,6 +269,5 @@ const constants = {
     // MethodNotAllowed error
     onlyOwnerAllowed: ['bucketDeletePolicy', 'bucketGetPolicy', 'bucketPutPolicy'],
 };
-
 
 module.exports = constants;
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1737,6 +1737,25 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+
+        this.apiBodySizeLimits = { ...constants.defaultApiBodySizeLimits };
+        if (config.apiBodySizeLimits) {
+            assert(typeof config.apiBodySizeLimits === 'object' &&
+                !Array.isArray(config.apiBodySizeLimits),
+                'bad config: apiBodySizeLimits must be an object');
+
+            for (const [apiKey, limit] of Object.entries(config.apiBodySizeLimits)) {
+                // Only allow modifications of predefined APIs from constants
+                assert(Object.hasOwn(constants.defaultApiBodySizeLimits, apiKey),
+                    `bad config: apiBodySizeLimits for "${apiKey}" cannot be configured. ` +
+                    `Valid APIs are: ${Object.keys(constants.defaultApiBodySizeLimits).join(', ')}`);
+                
+                assert(Number.isInteger(limit) && limit > 0,
+                    `bad config: apiBodySizeLimits for "${apiKey}" must be a positive integer`);
+                this.apiBodySizeLimits[apiKey] = limit;
+            }
+        }
+
         return config;
     }
 
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -75,6 +75,7 @@ const { tagConditionKeyAuth } = require('./apiUtils/authorization/tagConditionKe
 const { isRequesterASessionUser } = require('./apiUtils/authorization/permissionChecks');
 const checkHttpHeadersSize = require('./apiUtils/object/checkHttpHeadersSize');
 const constants = require('../../constants');
+const { config } = require('../Config.js');
 
 const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
 
@@ -219,15 +220,15 @@ const api = {
                 // issue 100 Continue to the client
                 writeContinue(request, response);
                
-                const defaultMaxPostLength = request.method === 'POST' ?
+                const defaultMaxBodyLength = request.method === 'POST' ?
                       constants.oneMegaBytes : constants.halfMegaBytes;
-                const MAX_POST_LENGTH = constants.apisLengthLimits[apiMethod] || defaultMaxPostLength;
+                const MAX_BODY_LENGTH = config.apiBodySizeLimits[apiMethod] || defaultMaxBodyLength;
                 const post = [];
-                let postLength = 0;
+                let bodyLength = 0;
                 request.on('data', chunk => {
-                    postLength += chunk.length;
+                    bodyLength += chunk.length;
                     // Sanity check on post length
-                    if (postLength <= MAX_POST_LENGTH) {
+                    if (bodyLength <= MAX_BODY_LENGTH) {
                         post.push(chunk);
                     }
                 });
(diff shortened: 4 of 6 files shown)


constants.js
```diff
@@ -89,6 +89,11 @@ const constants = {
     // Maximum HTTP headers size allowed
     maxHttpHeadersSize: 14122,
 
+    // AWS sets the maximum size for bucket policies to 20 KB
+    // https://docs.aws.amazon.com/AmazonS3/latest/userguide/add-bucket-policy.html
+    bucketPolicyMaxBytesSize: 20480,
```
SylvainSenechal, line 94: @francoisferrand Not too sure how we could (easily) make this configurable. I can store it in constants.js like I do here, or in config.js. But either way, the customers won't be able to edit these files when they have a deployed artesca

constants.js
```diff
@@ -89,6 +89,10 @@ const constants = {
     // Maximum HTTP headers size allowed
     maxHttpHeadersSize: 14122,
 
+    // AWS sets the maximum size for bucket policies to 20 KB
+    // https://docs.aws.amazon.com/AmazonS3/latest/userguide/add-bucket-policy.html
+    bucketPolicyMaxBytesSize: 20 * 1024,
```
SylvainSenechal: @francoisferrand Not too sure how we could (easily) make this configurable. I can store it in constants.js like I do here, or in config.js. But either way, the customers won't be able to edit these files when they have a deployed artesca

francoisferrand: quite easily, just add it to the Config (config.js) :
- in the constructor of Config object, use this constant by default (something like `this.bucketPolicyMaxBytesSize ||= constants.bucketPolicyMaxBytesSize;`)
- use the value from `config` instead of `constants` : `if (request.post && Buffer.byteLength(request.post, 'utf8') > constants.bucketPolicyMaxBytesSize) {`

lib/api/bucketPutPolicy.js
```diff
@@ -31,6 +32,14 @@ function _checkNotImplementedPolicy(policyString) {
 function bucketPutPolicy(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'bucketPutPolicy' });
 
+    if (request.post && Buffer.byteLength(request.post, 'utf8') > constants.bucketPolicyMaxBytesSize) {
+        return callback(errorInstances.MalformedPolicy.customizeDescription(
+            `The provided Bucket Policy is too large : ${request.post.length} bytes. ` +
```
francoisferrand: is this AWS error message?

SylvainSenechal: No, so I'll remove it like the other ones

lib/api/bucketPutPolicy.js
```diff
@@ -31,6 +32,10 @@ function _checkNotImplementedPolicy(policyString) {
 function bucketPutPolicy(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'bucketPutPolicy' });
 
+    if (request.post && Buffer.byteLength(request.post, 'utf8') > config.bucketPolicyMaxBytesSize) {
```
francoisferrand: I wonder if this is the right place:
- we already have a "network level" check on POST size : we should use it to customize the length for bucket policy, instead of adding a separate check
- if we want/need to add custom code, we should do this in BucketPolicy class (in artsenal), which has other bucket policy checks (https://github.com/scality/Arsenal/blob/80bb045e77ed9a5205af105820dc23b8351d3258/lib/models/BucketPolicy.ts#L92-L93)

either approach seem fine, what do you think?

SylvainSenechal: The bucket policy could use some more checking on the size of the poilcy, that's what I wanted to do first. 
But regardless, we also need an api check, and changing the bucket policy class to do this check early wasn't so straightforward when i looked at the code (no checks when the class is instanciated, only when calling getPolicy)

For now let's add this check in cloudserver

lib/Config.js
```diff
@@ -1730,6 +1730,23 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+
+        this.apisLengthLimits = { ...constants.defaultApisLengthLimits };
+        // If a value is provided in the config, it overwrites the default constant
```
DarkIsDude:
```suggestion
```

Comment are not really useful here 🙏 (same for the next one) ?

config.json
```diff
@@ -142,5 +142,9 @@
     },
     "kmip": {
         "providerName": "thales"
+    },
+    "apisLengthLimits": {
```
DarkIsDude:
```suggestion
    "apisPostSizeLimits": {
```

I would suggest also to use api instead of apis as Limits is already plural ?

constants.js
```diff
@@ -96,11 +96,14 @@ const constants = {
     oneMegaBytes: 1024 * 1024,
     halfMegaBytes: 512 * 1024,
 
-    // Some apis may need a custom body length limit :
-    apisLengthLimits: {
+    // Some apis may need a custom body length limit
+    defaultApisLengthLimits: {
```
DarkIsDude: It's not really a length but more a size ?

config.json
```diff
@@ -142,5 +142,9 @@
...
+        "multiObjectDelete": 2097152,
```
DarkIsDude, line 147: We should use `2024 * oneMegaBytes` to avoid magic number ?

config.json
DarkIsDude: Also why we define them here if we have already default value (that are exactly the same) ?

lib/Config.js
```diff
@@ -1730,6 +1730,23 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+
+        this.apisLengthLimits = { ...constants.defaultApisLengthLimits };
```
DarkIsDude: Should we add a test here ?

config.json
SylvainSenechal: yeah I wanted to do this but cannot in json file

lib/Config.js
SylvainSenechal: true I'll remove it

config.json
SylvainSenechal: I'm gonna update the name. This limit isnt even used for POST only, but also put. Actually this config.json here is more of an example config. In prod, users have a mechanism to update it. Here it's used as an example. When I did my test I used very low number to make sure it was working, and I kept it in the commit to demonstrate how it could be used

lib/Config.js
SylvainSenechal: I added a test

lib/Config.js
```diff
@@ -1730,6 +1730,21 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+
+        this.apiBodySizeLimits = { ...constants.defaultApiBodySizeLimits };
```
francoisferrand, line 1741: I don't think we want/need to allow users to override *every request size* : seems like an open door to creating inconsistent states...

while accessing `apiBodySizeLimits` through the config is fine, should we not keep things simple the config, or at the very least validate that overrides are only specified on specific api keys?

i.e. maybe something like this may be enough:

```js
this.apiBodySizeLimits = { ...constants.defaultApiBodySizeLimits };
if (config.bucketPolicyMaxBodySize) {
   assert(Number.isInteger(config.bucketPolicyMaxBodySize) && config.bucketPolicyMaxBodySize > 0,
                    `bad config: apiBodySizeLimits for "${apiKey}" must be a positive integer`);
   this.apiBodySizeLimits[....] = config.bucketPolicyMaxBodySize;
}
```

or otherwise introduce a white-list of allowed keys:
```js
// constant.js
configurableApiBodySizeLimits = [ 'bucketPutPolicy' ];

// config.js
for (const [apiKey, limit] of Object.entries(config.apiBodySizeLimits)) {
   assert(constant.configurableApiBodySizeLimits.contains(apiKey), "bad config: apiBodySizeLimits cannot be configured for ${apiKey}")
   ...
```

what do you think?

SylvainSenechal: I added a check line 1743 to only allow modifications of apis defined in constants.js

lib/Config.js
```diff
@@ -1730,6 +1730,26 @@ class Config extends EventEmitter {
...
+        if (config.apiBodySizeLimits) {
+            assert(typeof config.apiBodySizeLimits === 'object' &&
+                config.apiBodySizeLimits !== null &&
```
francoisferrand: nit: cannot happen inside `if (config.apiBodySizeLimits)`

lib/Config.js
```diff
@@ -1730,6 +1730,26 @@ class Config extends EventEmitter {
...
+                !Array.isArray(config.apiBodySizeLimits),
+                'bad config: apiBodySizeLimits must be an object');
+
+            for (const [apiKey, limit] of Object.entries(config.apiBodySizeLimits)) {
+                // Only allow modifications of predefined APIs from constants
+                assert(Object.hasOwn(constants.defaultApiBodySizeLimits, apiKey),
+                    `bad config: apiBodySizeLimits for "${apiKey}" is not a valid API. ` +
```
```suggestion
                    `bad config: apiBodySizeLimits for "${apiKey}" cannot be configured. ` +
```

francoisferrand (approved):

francoisferrand:
please also rebase and bump cloudserver version (9.1.0-preview.1) for release

lib/api/api.js
```diff
@@ -221,7 +222,7 @@ const api = {
                
                 const defaultMaxPostLength = request.method === 'POST' ?
                       constants.oneMegaBytes : constants.halfMegaBytes;
-                const MAX_POST_LENGTH = constants.apisLengthLimits[apiMethod] || defaultMaxPostLength;
+                const MAX_POST_LENGTH = config.apiBodySizeLimits[apiMethod] || defaultMaxPostLength;
```
BourgoisMickael: `MAX_BODY_LENGTH`

BourgoisMickael (approved):
We can include a functional test for PutBucketPolicy & MultiObjectDelete

PR #5867, opened 2025-07-17
CLDSRV-718: fix debian apt-get urls

leif-scality:
fix apt-get broken urls observed here https://github.com/scality/cloudserver/actions/runs/16341465257?pr=5855

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,9 @@ COPY ./package.json yarn.lock ./
 ENV PYTHON=python3.9
 ENV PY_VERSION=3.9.7
 
-RUN apt-get update \
+RUN sed -i 's|deb.debian.org/debian|archive.debian.org/debian|g' /etc/apt/sources.list \
+    && sed -i 's|security.debian.org/debian-security|archive.debian.org/debian-security|g' /etc/apt/sources.list \
+    && apt-get update \
     && apt-get install -y --no-install-recommends \
     jq \
     python \


leif-scality:
ping

leif-scality:
ping

Dockerfile
```diff
@@ -9,6 +9,9 @@ COPY ./package.json yarn.lock ./
 ENV PYTHON=python3.9
 ENV PY_VERSION=3.9.7
 
+RUN sed -i 's|deb.debian.org/debian|archive.debian.org/debian|g' /etc/apt/sources.list && \
```
BourgoisMickael: You can regroup this RUN with the RUN below

BourgoisMickael: To avoid having an additional layer

BourgoisMickael (approved):

Dockerfile
```diff
@@ -9,7 +9,9 @@ COPY ./package.json yarn.lock ./
 ENV PYTHON=python3.9
 ENV PY_VERSION=3.9.7
 
-RUN apt-get update \
+RUN sed -i 's|deb.debian.org/debian|archive.debian.org/debian|g' /etc/apt/sources.list \
+    && sed -i 's|security.debian.org/debian-security|archive.debian.org/debian-security|g' /etc/apt/sources.list \
+    && apt-get update \
```
nicolas2bert, line 14: Although it seems to pass the ci, the llm recommended me somethinglike: 

echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/99no-check-valid;

to ignore the stale timestamp. 

Can you double check it? 
Thanks

Dockerfile
leif-scality: done

Dockerfile
leif-scality: We have 404 errors not Release file expired, I also don't think accepting expired signatures is a good idea
https://unix.stackexchange.com/questions/2544/how-to-work-around-release-file-expired-problem-on-a-local-mirror

leif-scality: But I can try out of curiosity

leif-scality: The LLM suggestion did not work

https://github.com/scality/cloudserver/actions/runs/16346341543/job/46180929841?pr=5867

nicolas2bert (approved):

leif-scality:
@bert-e create_pull_requests

leif-scality:
ping

leif-scality:
ping

PR #5868, opened 2025-07-17
fix request size limit for multi delete

SylvainSenechal:
Issue: [CLDSRV-699](https://scality.atlassian.net/browse/CLDSRV-699)

[CLDSRV-699]: https://scality.atlassian.net/browse/CLDSRV-699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -89,6 +89,20 @@ const constants = {
     // Maximum HTTP headers size allowed
     maxHttpHeadersSize: 14122,
 
+    // AWS sets the maximum number of objects deleted to 1000 for multiDeleteObjects
+    // https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html
+    maxMultiObjectDeleteLen: 1000,
+
+    oneMegaBytes: 1024 * 1024,
+    halfMegaBytes: 512 * 1024,
+
+    // Some apis may need a custom body length limit :
+    apisLengthLimits: {
+        // Multi Objects Delete request can be large : up to 1000 keys of 1024 bytes is
+        // already 1mb, with the other fields it could reach 2mb
+        'multiObjectDelete': 2 * 1024 * 1024,
+    },
+
     // hex digest of sha256 hash of empty string:
     emptyStringHash: crypto.createHash('sha256')
         .update('', 'binary').digest('hex'),
@@ -251,5 +265,6 @@ const constants = {
     // MethodNotAllowed error
     onlyOwnerAllowed: ['bucketDeletePolicy', 'bucketGetPolicy', 'bucketPutPolicy'],
 };
+
 
 module.exports = constants;
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -74,6 +74,7 @@ const parseCopySource = require('./apiUtils/object/parseCopySource');
 const { tagConditionKeyAuth } = require('./apiUtils/authorization/tagConditionKeys');
 const { isRequesterASessionUser } = require('./apiUtils/authorization/permissionChecks');
 const checkHttpHeadersSize = require('./apiUtils/object/checkHttpHeadersSize');
+const constants = require('../../constants');
 
 const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
 
@@ -219,8 +220,10 @@ const api = {
                 }
                 // issue 100 Continue to the client
                 writeContinue(request, response);
-                const MAX_POST_LENGTH = request.method === 'POST' ?
-                      1024 * 1024 : 1024 * 1024 / 2; // 1 MB or 512 KB
+               
+                const defaultMaxPostLength = request.method === 'POST' ?
+                      constants.oneMegaBytes : constants.halfMegaBytes;
+                const MAX_POST_LENGTH = constants.apisLengthLimits[apiMethod] || defaultMaxPostLength;
                 const post = [];
                 let postLength = 0;
                 request.on('data', chunk => {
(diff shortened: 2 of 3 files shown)


lib/api/api.js
```diff
@@ -219,8 +219,14 @@ const api = {
...
                 writeContinue(request, response);
-                const MAX_POST_LENGTH = request.method === 'POST' ?
-                      1024 * 1024 : 1024 * 1024 / 2; // 1 MB or 512 KB
+                const apisLengthLimits = {
+                    // Multi Objects Delete request can be large : up to 1000 keys of 1024 bytes is
+                    // already 1mb, with the other fields it could reach 2mb
+                    'multiObjectDelete': 2 * 1024 * 1024,
+                };
+                const defaultMaxPostLength = request.method === 'POST' ?
+                      1024 * 1024 : 512 * 1024; // 1 MB or 512 KB
```
ghost: as an optim I think these values should be computed outside of the function, as it would be evaluated in every request otherwise, let's compute it on top of the file?

lib/api/multiObjectDelete.js
```diff
@@ -492,8 +493,11 @@ function multiObjectDelete(authInfo, request, log, callback) {
         function parseXML(next) {
             return _parseXml(request.post,
                 (err, quietSetting, objects) => {
-                    if (err || objects.length < 1 || objects.length > 1000) {
-                        return next(errors.MalformedXML);
+                    if (err || objects.length < 1 || objects.length > constants.maximumBatchDeleteObjects) {
+                        return next(errors.MalformedXML.customizeDescription(
```
ghost: To use customizeDescription we must not use `errors` but `errorInstances`. I think the error should stay the same, unless we are able to confirm what AWS exactly sends, we should be 100% consistent, so we can just keep the base error IMHO

lib/api/api.js
```diff
@@ -219,8 +219,14 @@ const api = {
                 }
                 // issue 100 Continue to the client
                 writeContinue(request, response);
-                const MAX_POST_LENGTH = request.method === 'POST' ?
-                      1024 * 1024 : 1024 * 1024 / 2; // 1 MB or 512 KB
+                const apisLengthLimits = {
+                    // Multi Objects Delete request can be large : up to 1000 keys of 1024 bytes is
```
ghost: Were you able to confirm we reach the limit with the 1000 keys?

francoisferrand: probably depends on the key length : which can be up to 1024 bytes per key, according to AWS...

lib/api/api.js
francoisferrand: it depends on `request.method`, so can't really be computed earlier... And `apisLengthLimits` is really a constant, so not sure it will change much...

lib/api/multiObjectDelete.js
```diff
@@ -492,8 +493,11 @@ function multiObjectDelete(authInfo, request, log, callback) {
...
+                            `The provided request tries to delete ${objects.length} objects, ` +
```
francoisferrand: is this the message that AWS returns? (the `AWS standard` part looks weird)

as a general rule of thumb, we want to stick to same behavior (esp errors) as AWS

lib/api/api.js
SylvainSenechal: Yeah I called the api with smaller requests, and change manually the max_post_length to a significantly lower value to trigger my errors. I logged the body length too and yeah with 1000 long keys it's gonna go above 1MB

lib/api/api.js
ghost: I mean the 1024\*1024 and 512\*1024, this can be the real value directly.

lib/api/multiObjectDelete.js
SylvainSenechal: I thought it would be nice to have a more descriptive errors but yeah if we want to stick the same aws errors I'll remove it

lib/api/multiObjectDelete.js
SylvainSenechal: No it's my own error, I realized we want to have the exact same errors as aws so I'll remove it

lib/api/api.js
SylvainSenechal: I can set it in the constants.js file

constants.js
```diff
@@ -252,4 +259,11 @@ const constants = {
     onlyOwnerAllowed: ['bucketDeletePolicy', 'bucketGetPolicy', 'bucketPutPolicy'],
 };
 
+// Some apis may need a custom body length limit :
+constants.apisLengthLimits = {
```
francoisferrand: should be set in `constants` definition directly?
(not sure it is worth setting variables for "oneMegaBytes" and "halfMegaBytes"), which cause this need to move it out...

lib/api/multiObjectDelete.js
```diff
@@ -492,7 +493,7 @@ function multiObjectDelete(authInfo, request, log, callback) {
         function parseXML(next) {
             return _parseXml(request.post,
                 (err, quietSetting, objects) => {
-                    if (err || objects.length < 1 || objects.length > 1000) {
+                    if (err || objects.length < 1 || objects.length > constants.maximumMultiObjectDelete) {
```
francoisferrand: `maximumMultiObjectDelete` name looks kind of weird in this context, maybe something `maxMultiObjectDeleteLen` ?

francoisferrand (approved):

constants.js
SylvainSenechal: yes actually its better in constants directly

ghost (approved):

PR #5877, opened 2025-07-22
Remove eslint custom rules and introduce prettier

ghost:
We should follow the eslint default, completed by the imported scality rules.

Any additional rule should add some constraints, like disabling .only tests to make the CI fail.

The reason is that we end up being too leniant on the style, and that translates into useless indent (or other style) changes in branches, due to local tooling auto-formatting the code, which make the reviews harder.

The main problem reported is the indentation: it is varying from file to file, making the code harder to read.

Almost all rules were already followed, so enabling them was not impacting the code at all.

Then we introduce prettier to ensure code style is consistent.

Issue: CLDSRV-723


eslint.config.mjs
BourgoisMickael, line 1: Ideally we should use `.editorconfig` for the IDE to format correctly.

We should use prettier for formatting stuff like indent.
And use eslint prettier plugin / config to turn off eslint stylistic rules like indent or quotes (they'll be moved and and not enabled by default in v10)

ghost: We can add a prettier yes. But we still want to enforce indentation before v10. How about introducing a prettier with an action to ensure it ran, then when upgrading to v10, we can remove the eslint rules? (actually, nothing to do, if they are removed).

I'll check for the .editorconfig at the same time.

BourgoisMickael: indentation is the job of prettier as it rewrite the file

ghost: I introduced prettier, now binded to eslint for consistent rules, with an `.editorconfig` and CI rules to verify the format.

As discussed, because this is a lot of changes, let's wait for unification to complete first.

Overall we really had a lot of formatting inconsistencies, so this will a bit slow down git blames as most of the lines are impacted, but it will be easier to work with integration branches from now on.

constants.js
```diff
@@ -46,8 +46,7 @@ const constants = {
     // only public resources
     publicId: 'http://acs.amazonaws.com/groups/global/AllUsers',
     // All Authenticated Users is an ACL group.
-    allAuthedUsersId: 'http://acs.amazonaws.com/groups/' +
-        'global/AuthenticatedUsers',
+    allAuthedUsersId: 'http://acs.amazonaws.com/groups/' + 'global/AuthenticatedUsers',
```
BourgoisMickael, line 49: This is because it was before limited to 80 chars, now it's 120, so those strings can be regrouped, we might have a lot like that in the code

package.json
```diff
@@ -1,132 +1,135 @@
...
-    "test_legacy_location": "CI=true S3QUOTA=scuba S3_LOCATION_FILE=tests/locationConfig/locationConfigLegacy.json S3BACKEND=mem mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json --recursive tests/unit --exit",
-    "test_utapi_v2": "mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json --recursive tests/utapi --exit",
-    "test_quota": "mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json --recursive tests/quota --exit",
-    "multiple_backend_test": "CI=true S3BACKEND=mem S3METADATA=mem S3DATA=multiple mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 20000 --recursive tests/multipleBackend --exit",
-    "cover": "nyc --clean --silent yarn run",
-    "postcover": "nyc report --report-dir ./coverage/test --reporter=lcov"
-  }
+    "name": "@zenko/cloudserver",
```
BourgoisMickael, line 2: this is 4 indents, not following prettierc

eslint.config.mjs
BourgoisMickael: editorconfig needs an IDE extension for some IDE: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

we could include a `.vscode/extensions.json` (like https://github.com/microsoft/TypeScript/blob/main/.vscode/extensions.json) so all devs could see a popup to install the extension in their vscode.

Same for prettier and eslint could have extensions to format on save.

PR #5881, opened 2025-07-22
CLDSRV-717: Unification prometheus monitoring cluster mode

BourgoisMickael:


anurag4DSB (approved):
Are there any tests for this?

BourgoisMickael:
> Are there any tests for this?

Unit tests here: https://github.com/scality/cloudserver/blob/development/9.0/tests/unit/utils/monitoring.js

They don't test based on config though so it only use the non cluster mode

lib/utilities/monitoringHandler.js
```diff
@@ -215,16 +215,29 @@ function writeResponse(res, error, results, cb) {
     });
 }
 
+const workers = process.env.S3BACKEND === 'mem' ? 1 : config.clusters || 1;
```
ghost: Should it be global & exported consts to avoid repeating?

lib/server.js
```diff
@@ -430,12 +422,16 @@ function main() {
...
         });
+
+        const metricServer = new S3Server(_config);
+        metricServer.startServer(_config.metricsListenOn,
+            _config.metricsPort, metricServer.routeAdminRequest);
     }
     if (this.cluster && cluster.isWorker) {
         const server = new S3Server(_config, cluster.worker);
         server.initiateStartup(logger.newRequestLogger());
+        monitoringClient.collectDefaultMetrics({ timeout: 5000 });
```
ghost, line 434: Could we unify this with a:

```js
    if (!isCluster || cluster.isWorker) {
        monitoringClient.collectDefaultMetrics({ timeout: 5000 });
    }
```

to avoid defining the timeout twice?

ghost (approved):

lib/utilities/monitoringHandler.js
BourgoisMickael: Yep, moved this into config to be reused: [cb91b7b](https://github.com/scality/cloudserver/pull/5881/commits/cb91b7bb306900caf19937e848ba7b57893a1ef8) & [368d4a0](https://github.com/scality/cloudserver/pull/5881/commits/368d4a0a661bae7ac000688e3ec53fdec4d4642b)

lib/server.js
BourgoisMickael: done [8830249](https://github.com/scality/cloudserver/pull/5881/commits/88302497a7c7b7f104fba9747ce8152f475c539e)

ghost (approved):

PR #5891, opened 2025-07-28
CLDSRV-717: Fix & drop config root redis for utapi

BourgoisMickael:
Currently, root redis is not defined neither by zenko or s3c. This breaks utapi replay with overrided redis config.

There has been multiple tickets S3C-1399 ZENKO-345 ZENKO-1051 ZENKO-3294

In the end root redis is not used and utapi should use utapi.redis and cloudserver uses localCache.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1298,7 +1298,10 @@ class Config extends EventEmitter {
         }
 
         if (config.redis) {
-            this.redis = parseRedisConfig(config.redis);
+            // Fail fast to make sure we detect any bad config
+            throw new Error(
+                'config.redis is not supported anymore: it should be config.utapi.redis or config.localCache'
+            );
         }
         if (config.scuba) {
             this.scuba = {};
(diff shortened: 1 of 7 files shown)


lib/Config.js
```diff
@@ -1298,7 +1298,10 @@ class Config extends EventEmitter {
         }
 
         if (config.redis) {
-            this.redis = parseRedisConfig(config.redis);
+            // Fail fast to make sure we detect any bad config
+            throw new Error(
+                '[S3C-10107 unification] config.redis is not used. It should be config.utapi.redis or config.localCache'
```
francoisferrand: would appear in logs, should not mention ticket:
```suggestion
                'config.redis is not supported anymore: it should be config.utapi.redis or config.localCache'
```

tcarmet (approved):

tmacro (approved):

francoisferrand (approved):

PR #5892, opened 2025-07-29
CLDSRV-717: Fix metadata route hooks promise & cb

BourgoisMickael:
Fix Integration error:
```
Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.
```

### DIFF
--- a/tests/functional/raw-node/test/routes/routeMetadata.js
+++ b/tests/functional/raw-node/test/routes/routeMetadata.js
@@ -44,18 +44,22 @@ describe('metadata routes with metadata', () => {
 
     // E2E tests use S3C metadata, whereas functional tests use mocked metadata.
     if (process.env.S3_END_TO_END) {
-        before(done => s3.createBucket({ Bucket: bucket1 }).promise()
+        before(done => {
+            s3.createBucket({ Bucket: bucket1 }).promise()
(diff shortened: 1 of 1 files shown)


dvasilas (approved):

tests/functional/raw-node/test/routes/routeMetadata.js
```diff
@@ -44,18 +44,22 @@ describe('metadata routes with metadata', () => {
...
-        after(done => bucketUtil.empty(bucket1)
+        after(done => {
+            bucketUtil.empty(bucket1)
             .then(() => s3.deleteBucket({ Bucket: bucket1 }).promise())
             .then(() => bucketUtil.empty(bucket2))
             .then(() => s3.deleteBucket({ Bucket: bucket2 }).promise())
-            .then(() => done(), err => done(err))
-        );
+            .then(() => done())
+            .catch(err => done(err));
```
ghost, line 61: Maybe the .finally can be used here as well :slightly_smiling_face:

ghost (approved):

tests/functional/raw-node/test/routes/routeMetadata.js
BourgoisMickael: I'd rather have a more global refacto from thenable callbacks to async await, because we have a mix of both throughout the tests

PR #5893, opened 2025-07-29
CLDSRV-717: Test lifecycle Transition condition

BourgoisMickael:
For S3C Integration that don't support Transition

### DIFF
--- a/tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
+++ b/tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
@@ -4,6 +4,7 @@ const { S3 } = require('aws-sdk');
 
 const getConfig = require('../support/config');
 const BucketUtility = require('../../lib/utility/bucket-util');
+const { config } = require('../../../../../lib/Config');
 
 const MAX_DAYS = 2147483647; // Max 32-bit signed binary integer.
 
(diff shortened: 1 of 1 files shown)


tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
```diff
@@ -505,7 +508,11 @@ describe('aws-sdk test put bucket lifecycle', () => {
                 }];
                 const params = getParams(transitions);
                 s3.putBucketLifecycleConfiguration(params, err => {
-                    assert.ifError(err);
+                    if (isSupported) {
+                        assert.ifError(err);
+                    } else {
+                        assert.strictEqual(err && err.code, 'NotImplemented');
```
francoisferrand: Would be better to have separate tests, expecting an error when the test is called "is supported" can lead to strange things...

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
```diff
@@ -484,6 +484,9 @@ describe('aws-sdk test put bucket lifecycle', () => {
         });
 
         describe('with Transitions', () => {
+            // Integration for S3C that doesn't support Transitions expects NotImplemented
+            const isSupported = !!process.env.S3_END_TO_END;
```
francoisferrand: The feature (in prod code) is configured through the config (and env I think) by setting a list of supported lifecycle workflows : IMHO it would be better to use the same variable in the test.

This way we reduce coupling, as the tests are not parametrized by an unspecified "s3c_e2e" variable but by an easer to understand "list of supported lifecycle workflows". The condition in test will be quite natural ("transition should succeed" if transitions are supported, and fail otherwise). And it would be easy to test new features if they are integrated in s3c.

Should be a general pattern IMO: if we introduce some flag controlling behavior in the code, we should use the same flag to adapt the tests.

BourgoisMickael: This flag is already in many different places in cloud server tests to work in Federation.

I can pick the supported list from the config and pass the config path env but some places need to be adapted to work with a different config than the default one

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
BourgoisMickael: I've changed the name of the test dynamically

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
BourgoisMickael: Changed.

The config file is passed to `S3_CONFIG_FILE` in Integration for the tests now

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
francoisferrand: not sure this is the most readable, esp. since a single test (or suite!) checking `putBucketLifecycleConfiguration` returns NotImplemented is probably enough when transitions are not supported, without complexifying each test and label...

francoisferrand (approved):

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
BourgoisMickael: I've reworked to have conditional describes and bring back the tests from branch 7.70

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
```diff
@@ -483,7 +484,63 @@ describe('aws-sdk test put bucket lifecycle', () => {
             });
         });
 
-        describe('with Transitions', () => {
+        // use env S3_CONFIG_FILE with tests (needed in S3C Integration tests)
+        const isTransitionSupported = config.supportedLifecycleRules.includes('Transition');
+
+        (isTransitionSupported ? describe.skip : describe)('with Transitions NOT supported', () => {
+            it('should return NotImplemented if Transitions rule', done => {
```
ghost, line 491:
```suggestion
            it('should return NotImplemented if rules does not include Transitions', done => {
```

?

ghost (approved):

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
BourgoisMickael: It's the reverse, if rules include Transitions

PR #5895, opened 2025-08-05
CLDSRV-717: Move utils out of multi backend

BourgoisMickael:
The awslocation in multiBackend breaks S3C Integration


anurag4DSB (approved):
Interesting way, could we have some developer comments for someone looking at the code to understand why it was done?

BourgoisMickael:
> Interesting way, could we have some developer comments for someone looking at the code to understand why it was done?

In Integration, the test `tests/functional/aws-node-sdk/test/object/rangeTest.js` would import that multiBackend utils that looks for `awslocation` in location constrainsts but we don't have this kind of location and it would crash

leif-scality (approved):

PR #5907, opened 2025-08-14
CLDSRV-717: Unify tests backbeat route

BourgoisMickael:
To be able to run backbeat routes test in Integration as it was before (in branch 7.70).


tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -270,7 +277,7 @@ describe('backbeat routes', () => {
             .then(() => done())
             .catch(err => {
                 process.stdout.write(`Error creating bucket: ${err}\n`);
-                throw err;
+                done(err);
```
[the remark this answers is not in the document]
BourgoisMickael: No, it times out the test because the promise is not returned nor awaited, so the `done` callback needs to be executed

tcarmet (approved):

lib/routes/routeBackbeat.js
```diff
@@ -1463,6 +1463,11 @@ function routeBackbeat(clientIP, request, response, log) {
     // Attach the apiMethod method to the request, so it can used by monitoring in the server
     // eslint-disable-next-line no-param-reassign
     request.apiMethod = 'routeBackbeat';
+    const contentLength = request.headers['x-amz-decoded-content-length'] ?
+        request.headers['x-amz-decoded-content-length'] :
+        request.headers['content-length'];
```
ghost:
```suggestion
    const contentLength = request.headers['x-amz-decoded-content-length'] || request.headers['content-length'];
```

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -633,7 +640,8 @@ describe('backbeat routes', () => {
             });
         });
 
-        it('should update metadata of a non-version object', done => {
+        // S3C Skipping is necessary because non-versioned buckets are not supported by S3C backbeat routes.
+        itSkipS3C('should update metadata of a non-version object', done => {
```
ghost: I don't have the full context in mind for backbeat, but don't we want to support it in the unified branches as well? If yes, do we have a ticket number we can reference? Otherwise, it is not really unified?

BourgoisMickael: I don't know, didn't heard anything about supporting non-versioned replication in S3C

BourgoisMickael: I can't find anything in citadel related to non-verisoned bucket CRR. We follow the aws spec, why is non-version buckets allowed to use CRR in artesca ? @williamlardier

BourgoisMickael: I found the only non version replication allowed is for NFS buckets, and we don't support that neither in S3C:

https://github.com/scality/cloudserver/blob/15ac4c656a4312d7a31bc22bd8d91895e47df812/lib/api/bucketPutReplication.js#L46-L50

ghost: Yes but the question is, if we end up with unified branches but don't support the same thing, is that really what we want, and is there a specific reason to support less (or more) in s3c than in zenko? Ideally we should have a unique test suite, no?

BourgoisMickael: not for now, we have no plan to support NFS the same way it's done in zenko, we don't support cold storage as well for now in RING9 but this is planned for RING10, so in the meantime we have to skip some tests because those features are not in S3C. 

We don't support external backends neither, nor remote management, we don't support mongodb

BourgoisMickael: Btw in S3C there's an `nfsd` service to deal with nfs, I don't know how it works but that might still differ from the way zenko does it

ghost: In this case I think we should be careful documenting all these differences somewhere, because the more differences for the same code path, the more likely bugs will be unexpectedly introduced. At least a link to the ticket where this feature debt will be resolved, so we don't forget.

From what I understand we have `S3_END_TO_END` set in Integration, where we rerun this repository's set of tests with a different environment setup. This is already a bit strange as not documented here, but also hiding a deeper problem where there is no test in this repository guarding against code edits that would affect S3C.

For S3C, a test for replicating to a non-versioned bucket should assert that the API returns an InvalidBucketState error, no? I didn't go through all the paths, but to me this is not sufficient to skip tests based on the backend, we should enrich them to fully test all supported behaviors instead.

BourgoisMickael: In S3C we can't setup replication without versioning, an error is thrown we can't replicate non-versioned bucket already. This is only a new feature with new tests from zenko branch.

I've a confluence page https://scality.atlassian.net/wiki/spaces/OS/pages/3281682450/Cloudserver+Arsenal+diff+7.70+-+9.0 to note all differences and produce release notes for that at the end of unification.

ghost: > In S3C we can't setup replication without versioning

I know, but my comment really is because I think the approach is strange (not just yours, but the whole test suite and how we use it in Integration), not well documented, and prone to future issues now that it is unified and we skip tests. Let me try to be more specific:

Both Zenko and S3C share that same code:

```
                // Replication requires that versioning is 'Enabled' unless it
                // is an NFS bucket.
                if (!bucket.isNFS() && !bucket.isVersioningEnabled(bucket)) {
                    return next(versioningNotEnabledError);
                }
```

So it is already not possible to have non-versioned buckets (without NFS) that reach this logic, yet we test it. So maybe there is something more to it?

When the backend is not `multiplebackend` then i see we reach:

```
            if (!useMultipleBackend) {
                const versioningConfig = bucketInfo.getVersioningConfiguration();
                // The following makes sure that only replication destination-related operations
                // target buckets with versioning enabled.
                const isVersioningRequired = request.headers['x-scal-versioning-required'] === 'true';
                if (isVersioningRequired && (!versioningConfig || versioningConfig.Status !== 'Enabled')) {
                    log.debug('bucket versioning is not enabled');
                    return next(errors.InvalidBucketState);
                }
                return backbeatRoutes[request.method][request.resourceType](
                    request, response, bucketInfo, objMd, log, next);
            }
```

In route backbeat. So in the non-multiplebackend case, with `x-scal-versioning-required` set, we should reject the request, I found that it is already tested. However:
- `x-scal-versioning-required` is only defined in PutData or PutMetadata, not GetMetadata for the backbeat route, so it should not be skipped?
- We can setup an expiration job on non versioned buckets, and it seems to me that (1) it uses the getMetadata logic we test here and (2) there is no specific difference in S3C (we can create the same rule). In both cases we accept non-versioned buckets and reach this logic.

francoisferrand: It seems to me this test is checking the behavior of PUT metadata route, which is not specific to replication : and should now be able to update metadata in non-versioned bucket as well... (this is not a "replicateObject" route)

→ IIMHO should be tested in S3C as well, the API should work and be tested accordingly... But there may be additional tests (purely in Federation/Integration) which verify that we cannot CRR to non-versioned bucket.
→ the only case where we may skip tests should be when the behavior of _cloudserver_ is actually different (and that should be linked to some config option). Like putLifecycleConfiguration which rejects transitions on S3C (since they are disabled).

> not well documented

yes, it seems these backbeat routes seem to have grown organically and with a non-specified contract, starting a "replication routes" and becoming much more generic... So there is a blurry line between bugs/implementation limits and quirky behavior inherited (and required) for the specific use case the API was initially used for...

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -223,7 +230,7 @@ function getMetadataToPut(putDataResponse) {
     return mdToPut;
 }
 
-describe('backbeat routes', () => {
+describeSkipIfAWS('backbeat routes', () => {
```
francoisferrand: I don't understand why we skip here, and esp. based on some AWS condition.
Backbeat routes are present on both Zenko and S3C, so should be tested... and the test here seem to apply to both...

tests/functional/aws-node-sdk/lib/utility/test-utils.js
```diff
@@ -1,8 +1,16 @@
...
+const isS3C = process.env.S3_END_TO_END !== undefined;
 const itSkipCeph = isCEPH ? it.skip : it;
+const itSkipS3C = isS3C ? it.skip : it;
+const itSkipNotS3C = isS3C ? it : it.skip;
+const itSkipCephS3C = isCEPH || isS3C ? it.skip : it;
+
 const describeSkipIfCeph = isCEPH ? describe.skip : describe.skip; // always skip
+const describeSkipIfAWS = process.env.AWS_ON_AIR ? describe.skip : describe;
+const describeSkipIfS3C = isS3C ? describe.skip : describe;
+
```
francoisferrand: we have a single shared code base : so we should not skip based on the "product", but only based on the features which are enabled in the target instance of cloudserver (i.e. based on the config.yaml used by that cloudserver and the locations that are defined)

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -1529,6 +1537,9 @@ describe('backbeat routes', () => {
                                 TEST_ENCRYPTED_BUCKET : TEST_BUCKET,
                             objectKey: testCase.encodedKey,
                             resourceType: 'metadata',
+                            queryObj: {
+                                versionId: versionIdUtils.encode(testMd.versionId),
+                            },
                             authCredentials: backbeatAuthCredentials,
```
francoisferrand: what is this extra parameter?
this was not required before (at least in Zenko), why do we need to pass it?

* Here, it seems the tests are about creating a new version automatically, so the caller (practically, that would be backbeat) should not provide a versionID; if we have another feature where we can create an object with a specific VersionID, that should be a separate test (if not already implemented)
* If this parameter is actually required (even for Zenko!) -even though maybe the tests did not see the issue- then we are missing some validation in the API

francoisferrand: (same applies to a few other tests below)

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -2540,12 +2586,13 @@ describe('backbeat routes', () => {
...
                     // should error out as location shall no longer exist
                     assert(err);
+                    assert.strictEqual(err.statusCode, 503);
                     done();
                 }),
             ], done);
         });
 
-        itSkipCeph('should batch delete a versioned AWS location', done => {
+        itSkipCephS3C('should batch delete a versioned AWS location', done => {
```
francoisferrand: condition should be on S3C, but on `AWS` availability

tests/multipleBackend/routes/routeBackbeat.js
BourgoisMickael: The `process.env.AWS_ON_AIR` (from branch 7.70) was used if we run tests against AWS directly and not our product

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -2642,7 +2690,7 @@ describe('backbeat routes', () => {
             ], done);
         });
 
-        it('should not put delete tags if the source is not Azure and ' +
+        itSkipS3C('should not put delete tags if the source is not Azure and ' +
```
francoisferrand: condition should be on S3C, but on AWS availability (it is called _not_ Azure, and actually uses AWS location)

francoisferrand: same for a few others below

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -2800,7 +2848,7 @@ describe('backbeat routes', () => {
             ], done);
         });
 
-        it('should not delete the object if the source is Azure and ' +
+        itSkipS3C('should not delete the object if the source is Azure and ' +
```
francoisferrand: condition should be on S3C, but on Azure availability (same below)

tests/multipleBackend/routes/routeBackbeat.js
francoisferrand: does not seem to rely on AWS backend, so should run as well (it's just API) in S3C

tests/multipleBackend/routes/routeBackbeatForReplication.js
```diff
@@ -35,12 +62,30 @@ function objectMDFromRequestBody(data) {
...
+    return objMD.getSerialized();
+}
+
+const scenarios = [
+    { name: 'same account', src, dst: src },
+    { name: 'cross account', src, dst },
+];
+
+scenarios.forEach(({ name, src, dst }) => {
+describeSkipIfAWS(`backbeat routes for replication (${name})`, () => {
```
francoisferrand: should not be skipped : backbeat routes for replication are present either way, and should be tested. And in most case they will work the same (except of course when there is some specific tweak depending on the backend location type) : so most, if not all, should always be tested.

tests/multipleBackend/routes/routeBackbeatForReplication.js
```diff
@@ -35,12 +62,30 @@ function objectMDFromRequestBody(data) {
...
+    }
+
+    return objMD.getSerialized();
+}
+
+const scenarios = [
+    { name: 'same account', src, dst: src },
+    { name: 'cross account', src, dst },
+];
+
```
francoisferrand: this does not seem to be related to the same PR, is this not adding new cross-account replication tests?

(it makes it kind of difficult to review, as it is hard to pinpoint if specific part of the change is related to these new tests or to unification)

tests/multipleBackend/routes/routeBackbeatForReplication.js
```diff
@@ -955,7 +945,7 @@ describe('backbeat routes for replication', () => {
         });
     });
 
-    it.skip('should replicate/put metadata to a destination that has a null version', done => {
+    itSkipNotS3C('should replicate/put metadata to a destination that has a null version', done => {
```
francoisferrand: these few are not "not S3C" : they are actually a bug in Zenko (in mongodb backend, with a slight divergence from metadata behavior?), which must be fixed in Zenko as well

--> let's skip these with a link to CLDSRV-632 ticket which aims to fix them, instead of marking them as an "expected" divergence in behavior

package.json
```diff
@@ -93,7 +93,7 @@
     "ft_backbeat": "cd tests/functional/backbeat && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
     "ft_node": "cd tests/functional/raw-node && yarn test",
     "ft_node_routes": "cd tests/functional/raw-node && yarn run test-routes",
-    "ft_node_route_backbeat": "cd tests/functional/raw-node && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 test/routes/routeBackbeat.js --exit",
+    "ft_route_backbeat": "cd tests/multipleBackend/routes && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 routeBackbeat.js routeBackbeatForReplication.js --exit",
```
francoisferrand, line 96: kind of duplicates `multiple_backend_test` script?

tests/multipleBackend/routes/routeBackbeatForReplication.js
BourgoisMickael: `AWS_ON_AIR` is used to run test suite against AWS to compare with real implementation instead of our cloudserver.

 They implement S3 API they don't have our internal backbeat route implementation, this was used to skip our custom features when running our tests scenarios against AWS S3 Implementation

package.json
BourgoisMickael: this only picks backbeat route tests without veeam route and other multiple backend tests, as it was executing in `ft_node_routes` before in 7.70

tests/multipleBackend/routes/routeBackbeatForReplication.js
BourgoisMickael: In S3C branch we had no `replication` account, all the tests were done on the same account, we had no cross account tests.

In Zenko branch the same account test was replaced by cross account.

So I've added 2 scenarios to test in both ways: same account (the way it as in S3C branch) and cross account (like zenko branch).

The only difference is on the test where the object has it's account info updated

tests/multipleBackend/routes/routeBackbeatForReplication.js
BourgoisMickael: added there [da2cb20](https://github.com/scality/cloudserver/pull/5907/commits/da2cb2056d472b8fac06172681de56ea3518de90)

tests/multipleBackend/routes/routeBackbeatForReplication.js
BourgoisMickael: Extracted this file into a separate PR: https://github.com/scality/cloudserver/pull/5919

tests/multipleBackend/routes/routeBackbeat.js
BourgoisMickael: Thanks, I fixed this, in the end there is only 1 test that doesn't pass on S3C: [0bd3d47](https://github.com/scality/cloudserver/pull/5907/commits/0bd3d473b32fe9b8f6d9c94bd1ab815b9d6bdbe6)

tests/multipleBackend/routes/routeBackbeat.js
BourgoisMickael: Fixed: [288606a](https://github.com/scality/cloudserver/pull/5907/commits/288606a7de0773121c7518cabfe4b124c7897aae)

tests/multipleBackend/routes/routeBackbeat.js
BourgoisMickael: This is not required, I removed it.

It came from branch 7.70 where the test did not create new version but updated the same versionId

tmacro (approved):

PR #5911, opened 2025-08-20
Evaluate policies in Backbeat routes

welansari:
Evaluate policies in Backbeat routes when running in the external Cloudserver.

_This PR is based on the implementation in https://github.com/scality/cloudserver/pull/5714_

Issue: CLDSRV-728

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -81,7 +81,68 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
 
 auth.setHandler(vault);
 
+function checkAuthResults(authResults, apiMethod, log) {
+    let returnTagCount = true;
+    const isImplicitDeny = {};
+    let isOnlyImplicitDeny = true;
+    if (apiMethod === 'objectGet') {
+        // first item checks s3:GetObject(Version) action
+        if (!authResults[0].isAllowed && !authResults[0].isImplicit) {
+            log.trace('get object authorization denial from Vault');
+            return errors.AccessDenied;
+        }
+        isImplicitDeny[authResults[0].action] = authResults[0].isImplicit;
+        // second item checks s3:GetObject(Version)Tagging action
+        if (!authResults[1].isAllowed) {
+            log.trace('get tagging authorization denial ' +
+            'from Vault');
+            returnTagCount = false;
+        }
+    } else {
+        for (let i = 0; i < authResults.length; i++) {
+            isImplicitDeny[authResults[i].action] = true;
+            if (!authResults[i].isAllowed && !authResults[i].isImplicit) {
+                // Any explicit deny rejects the current API call
+                log.trace('authorization denial from Vault');
+                return errors.AccessDenied;
+            }
+            if (authResults[i].isAllowed) {
+                // If the action is allowed, the result is not implicit
+                // Deny.
+                isImplicitDeny[authResults[i].action] = false;
+                isOnlyImplicitDeny = false;
+            }
+        }
+    }
+    // These two APIs cannot use ACLs or Bucket Policies, hence, any
+    // implicit deny from vault must be treated as an explicit deny.
+    if ((apiMethod === 'bucketPut' || apiMethod === 'serviceGet') && isOnlyImplicitDeny) {
+        return errors.AccessDenied;
+    }
+    return { returnTagCount, isImplicitDeny };
+}
+
 /* eslint-disable no-param-reassign */
+function handleAuthorizationResults(request, authorizationResults, apiMethod, returnTagCount, log, callback) {
+    if (authorizationResults) {
+        const checkedResults = checkAuthResults(authorizationResults, apiMethod, log);
+        if (checkedResults instanceof Error) {
+            return callback(checkedResults);
+        }
+        returnTagCount = checkedResults.returnTagCount;
+        request.actionImplicitDenies = checkedResults.isImplicitDeny;
+    } else {
+        // create an object of keys apiMethods with all values to false:
+        // for backward compatibility, all apiMethods are allowed by default
+        // thus it is explicitly allowed, so implicit deny is false
+        request.actionImplicitDenies = request.apiMethods.reduce((acc, curr) => {
+            acc[curr] = false;
+            return acc;
+        }, {});
+    }
+    return callback(null, { returnTagCount });
+}
+
 const api = {
     callApiMethod(apiMethod, request, response, log, callback) {
         // Attach the apiMethod method to the request, so it can used by monitoring in the server
@@ -151,47 +212,6 @@ const api = {
         const apiMethods = requestContexts ? requestContexts.map(context => context._apiMethod) : [];
         // Attach the names to the current request
         request.apiMethods = apiMethods;
-
-        function checkAuthResults(authResults) {
-            let returnTagCount = true;
-            const isImplicitDeny = {};
-            let isOnlyImplicitDeny = true;
-            if (apiMethod === 'objectGet') {
-                // first item checks s3:GetObject(Version) action
-                if (!authResults[0].isAllowed && !authResults[0].isImplicit) {
-                    log.trace('get object authorization denial from Vault');
-                    return errors.AccessDenied;
-                }
-                isImplicitDeny[authResults[0].action] = authResults[0].isImplicit;
-                // second item checks s3:GetObject(Version)Tagging action
-                if (!authResults[1].isAllowed) {
-                    log.trace('get tagging authorization denial ' +
-                    'from Vault');
-                    returnTagCount = false;
-                }
-            } else {
-                for (let i = 0; i < authResults.length; i++) {
-                    isImplicitDeny[authResults[i].action] = true;
-                    if (!authResults[i].isAllowed && !authResults[i].isImplicit) {
-                        // Any explicit deny rejects the current API call
-                        log.trace('authorization denial from Vault');
-                        return errors.AccessDenied;
-                    }
-                    if (authResults[i].isAllowed) {
-                        // If the action is allowed, the result is not implicit
-                        // Deny.
-                        isImplicitDeny[authResults[i].action] = false;
-                        isOnlyImplicitDeny = false;
-                    }
-                }
-            }
-            // These two APIs cannot use ACLs or Bucket Policies, hence, any
-            // implicit deny from vault must be treated as an explicit deny.
-            if ((apiMethod === 'bucketPut' || apiMethod === 'serviceGet') && isOnlyImplicitDeny) {
-                return errors.AccessDenied;
-            }
-            return { returnTagCount, isImplicitDeny };
-        }
 
         return async.waterfall([
             next => auth.server.doAuth(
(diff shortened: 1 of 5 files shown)


tests/unit/DummyRequest.js
```diff
@@ -25,6 +25,14 @@ class DummyRequest extends http.IncomingMessage {
...
         }
         this.push(null);
+        if (!this.socket) {
+            this.socket = {
+                remoteAddress: '127.0.0.1',
+                destroy: () => {},
+                on: () => {},
+                removeListener: () => {},
+            };
+        }
```
ghost:
```suggestion
        this.socket ??= {
            remoteAddress: '127.0.0.1',
            destroy: () => {},
            on: () => {},
            removeListener: () => {},
        };
```

lib/routes/routeBackbeat.js
```diff
@@ -1446,20 +1454,56 @@ function routeIndexingAPIs(request, response, userInfo, log) {
...
+        // We don't use the authorization results for now
+        // as the UI uses the external Cloudserver instance
+        // as a proxy to access the Backbeat API service.
+
+        // eslint-disable-next-line no-param-reassign
+        request.accountQuotas = infos?.accountQuota;
+        // FIXME for now, any authenticated user can access API
+        // routes. We should introduce admin accounts or accounts
+        // with admin privileges, and restrict access to those
+        // only.
```
ghost, line 1476: Not related but I think the comment is a bit misleading as we started to work on https://scality.atlassian.net/wiki/spaces/OS/pages/2895347722/Authorizing+Internal+Services+S3+Operations
where the routes are actually open on purpose, but protected with the routing rules (internal s3 is not accessible from outside)

tests/unit/routes/routeBackbeat.js
```diff
@@ -694,3 +700,312 @@ describe('routeBackbeat', () => {
...
+                assert.strictEqual(request.apiMethods.length, 1);
+                assert.strictEqual(request.accountQuotas, 1000);
+                return done();
+            });
+        });
+    });
+
+    // Although the authz result is by default an implicit deny, the
+    // ACL should prevent any further processing for non-service or
+    // non-account identities.
```
ghost: could we test all apis like before? They should properly be denied in this case as well, right?
Also as you mention "non service or non account", maybe we should also test both cases (allow & deny cases)? I believe it will be hard to test at the integration level, or even functionally...

ghost (approved):
Changes lgtm, just some more tests I believe we can add for completeness

lib/api/api.js
```diff
@@ -267,27 +287,19 @@ const api = {
...
+                        return next(err);
+                    }
+                    returnTagCount = res.returnTagCount;
+                    return next(null, userInfo, authorizationResults, streamingV4Params, infos);
+                }),
         ], (err, userInfo, authorizationResults, streamingV4Params, infos) => {
             if (err) {
                 return callback(err);
             }
             request.accountQuotas = infos?.accountQuota;
```
francoisferrand: this used to be set before authorization : is it ok to do it after, esp. in case of error?
(i.e. could there be some case where we need to "rollback" the accountQuota after the error)

[safe fix is to move this line up to line 292 ?]

lib/routes/routeBackbeat.js
```diff
@@ -1401,6 +1402,17 @@ const backbeatRoutes = {
     },
 };
 
+function respondToRequest(err, response, log, callback) {
+    responseJSONBody(err, null, response, log);
+    // The callback is optional, as it is only used for testing purposes
+    // but value may be set to non-undefined or null due to the arsenal
+    // routes implementation
```
francoisferrand: seems quite dangerous, aren't we missing some anonymous function in there to call the function with the expected parameters?

lib/routes/routeBackbeat.js
francoisferrand: this comment was already there (this whole function was moved from next chunk0, and is still valid: for these "api" routes (which are exposed on the public-facing port as well), we still don't have or enforce policies.

lib/routes/routeBackbeat.js
```diff
@@ -1446,20 +1454,56 @@ function routeIndexingAPIs(request, response, userInfo, log) {
...
+        return backbeatProxy.web(request, response, { target }, err => {
+            log.error('error proxying request to api server',
+                { error: err.message });
+            return respondToRequest(errors.ServiceUnavailable, response, log, callback);
+        });
+    }, 's3', requestContexts);
+}
 
-function routeBackbeat(clientIP, request, response, log) {
+function routeBackbeat(clientIP, request, response, log, callback) {
```
francoisferrand: do we actually need the callback? I think we can test by using a fake/mock/... response object to be notified when the response by fired (and the "route..." function has completed)

lib/routes/routeBackbeat.js
```diff
@@ -1554,41 +1572,56 @@ function routeBackbeat(clientIP, request, response, log) {
...
+                return next(err, userInfo, authorizationResults);
+            }, 's3', requestContexts),
+        (userInfo, authorizationResults, next) => {
+            // Using the same flag used to bypass user bucket policies
+            // for internal mode, so that we can bypass policy evaluation
+            // on backbeat routes. This ensures we don't break operations
+            // coming from internal services like the backbeat
+            if (request.bypassUserBucketPolicies) {
+                return next(null, userInfo);
             }
```
francoisferrand, line 1606: this "skip" on internal routes is already implemented deep inside bucket policies, inside `isBucketAuthorized` and `isObjectAuthorized` : is this redundant (i.e. we indeed call these later), or do we really need the extra skip here? In that case, should that check be added in `handleAuthorizationResults` ?

lib/routes/routeBackbeat.js
```diff
@@ -1554,41 +1572,56 @@ function routeBackbeat(clientIP, request, response, log) {
...
+                }
+                // eslint-disable-next-line no-param-reassign
+                request.accountQuotas = infos?.accountQuota;
+                return next(err, userInfo, authorizationResults);
+            }, 's3', requestContexts),
+        (userInfo, authorizationResults, next) => {
+            // Using the same flag used to bypass user bucket policies
+            // for internal mode, so that we can bypass policy evaluation
+            // on backbeat routes. This ensures we don't break operations
+            // coming from internal services like the backbeat
```
francoisferrand:
```suggestion
            // coming from internal services like backbeat
```

lib/routes/routeBackbeat.js
```diff
@@ -1634,25 +1671,37 @@ function routeBackbeat(clientIP, request, response, log) {
...
-                    });
-                    return responseJSONBody(err, null, response, log);
-                }
-                log.debug('backbeat route response sent successfully');
-                return undefined;
-            },
-        ));
+        err => {
+            if (err) {
+                return async.forEachLimit(
```
francoisferrand: as far as I understand hooks should _always_ be called (not just one error!), even though currently they are only useful to handle the error case.

lib/routes/routeBackbeat.js
```diff
@@ -1615,6 +1648,10 @@ function routeBackbeat(clientIP, request, response, log) {
             return standardMetadataValidateBucketAndObj(mdValParams, request.actionImplicitDenies, log, next);
         },
         (bucketInfo, objMd, next) => {
+            // function already called
```
francoisferrand: * comment not clear
* this also hints that this function should not be in the same waterfall : this function should only be called to finish the `standardMetadataValidateBucketAndObj` branch, and not in the earlier branch from line 1607

lib/api/api.js
welansari: I moved it to line 292, i think it should work similar to how it was before.
@williamlardier could you please confirm this is good ? I'm not very familiar with quotas.

lib/routes/routeBackbeat.js
welansari: The check is not redundant, as far as i understood we are only supposed to skip bucket policies when using an internal Cloudserver not all the policies, so handling the flag at a higher level like in `handleAuthorizationResults` would disable all policy evaluation including for standard APIs which is not something we want.

lib/routes/routeBackbeat.js
welansari: removed the callback

lib/api/api.js
ghost: Yes, lgtm

lib/api/api.js
```diff
@@ -81,7 +81,68 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+    let isOnlyImplicitDeny = true;
+    if (apiMethod === 'objectGet') {
+        // first item checks s3:GetObject(Version) action
+        if (!authResults[0].isAllowed && !authResults[0].isImplicit) {
+            log.trace('get object authorization denial from Vault');
+            return errors.AccessDenied;
+        }
+        isImplicitDeny[authResults[0].action] = authResults[0].isImplicit;
+        // second item checks s3:GetObject(Version)Tagging action
+        if (!authResults[1].isAllowed) {
```
SylvainSenechal, line 96: Will it always be safe to access indexes [0],[1] in authResults ?

lib/api/api.js
```diff
@@ -81,7 +81,68 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
 
 auth.setHandler(vault);
 
+function checkAuthResults(authResults, apiMethod, log) {
+    let returnTagCount = true;
+    const isImplicitDeny = {};
+    let isOnlyImplicitDeny = true;
```
SylvainSenechal, line 87: Nit: I think this should have a plural name (like you used line 133).
I would also define the const before or after all the let, instead of doing it in the middle

lib/routes/routeBackbeat.js
francoisferrand: > handling the flag at a higher level like in handleAuthorizationResults would disable all policy evaluation including for standard APIs which is not something we want.

That explains why we want to keep the checks deep in `isBucketAuthorized` and `isObjectAuthorized` instead of moving it `handleAuthorizationResults`.

But should we not evaluate these "standard" policies on backbeat routes as well (i.e. remove this extra bypass, and just rely on the deeper bypass) ? Or is it something we can only do later, when we have added the required (custom) permissions in arsenal/vault and given them to the internal users?

lib/api/api.js
francoisferrand: this code was moved

lib/api/api.js
francoisferrand: this code was moved

lib/routes/routeBackbeat.js
welansari: We could evaluate policies on the internal Cloudserver, however i'm not certain our internal services have the required permissions (policies) hence why i skipped the evaluation here.

I created a follow-up ticket to verify the policies we create for our services and eventually remove this check (ZENKO-5058)

francoisferrand (approved):

PR #5912, opened 2025-08-25
CLDSRV-717: Tests backbeat route isolate buckets

BourgoisMickael:
Add a suffix to isolate buckets
With versions if one test fails and bucket can't be cleaned, it will make other tests fails

### DIFF
--- a/tests/multipleBackend/routes/routeBackbeat.js
+++ b/tests/multipleBackend/routes/routeBackbeat.js
@@ -37,12 +37,12 @@ const backbeatAuthCredentials = {
     accessKey: accessKeyId,
     secretKey: secretAccessKey,
 };
-const TEST_BUCKET = 'backbeatbucket';
-const TEST_ENCRYPTED_BUCKET = 'backbeatbucket-encrypted';
+const TEST_BUCKET_PREFIX = 'backbeatbucket';
+const TEST_ENCRYPTED_BUCKET_PREFIX = 'backbeatbucket-encrypted';
 const TEST_KEY = 'fookey';
-const NONVERSIONED_BUCKET = 'backbeatbucket-non-versioned';
-const VERSION_SUSPENDED_BUCKET = 'backbeatbucket-version-suspended';
-const BUCKET_FOR_NULL_VERSION = 'backbeatbucket-null-version';
+const NONVERSIONED_BUCKET_PREFIX = 'backbeatbucket-non-versioned';
+const VERSION_SUSPENDED_BUCKET_PREFIX = 'backbeatbucket-version-suspended';
+const BUCKET_FOR_NULL_VERSION_PREFIX = 'backbeatbucket-null-version';
 
 const testArn = 'aws::iam:123456789012:user/bart';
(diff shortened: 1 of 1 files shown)


tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -306,14 +312,15 @@ describe('backbeat routes', () => {
             assert.strictEqual(StorageClass, 'STANDARD');
         }
 
-        beforeEach(() =>
-            bucketUtil.emptyIfExists(BUCKET_FOR_NULL_VERSION)
-                .then(() => s3.createBucket({ Bucket: BUCKET_FOR_NULL_VERSION }).promise())
-        );
+        beforeEach(() => {
+            bucket = `${BUCKET_FOR_NULL_VERSION_PREFIX}-${Date.now()}`;
```
anurag4DSB: Date.now() usually causes flaky tests, unless you are confident I would suggest to use UUID

anurag4DSB (approved):

tests/multipleBackend/routes/routeBackbeat.js
BourgoisMickael: right, fixed thanks :+1:

tcarmet (approved):

tests/multipleBackend/routes/routeBackbeat.js
```diff
@@ -37,12 +37,12 @@ const backbeatAuthCredentials = {
...
-const TEST_BUCKET = 'backbeatbucket';
-const TEST_ENCRYPTED_BUCKET = 'backbeatbucket-encrypted';
+const TEST_BUCKET_PREFIX = 'backbeatbucket';
+const TEST_ENCRYPTED_BUCKET_PREFIX = 'backbeatbucket-encrypted';
 const TEST_KEY = 'fookey';
-const NONVERSIONED_BUCKET = 'backbeatbucket-non-versioned';
-const VERSION_SUSPENDED_BUCKET = 'backbeatbucket-version-suspended';
-const BUCKET_FOR_NULL_VERSION = 'backbeatbucket-null-version';
+const NONVERSIONED_BUCKET_PREFIX = 'backbeatbucket-non-versioned';
+const VERSION_SUSPENDED_BUCKET_PREFIX = 'backbeatbucket-version-suspended';
```
ghost, line 44: With uuid v4 this is already 70 bytes long, e.g. `backbeatbucket-version-suspended-ae718b1c-a677-48e5-be1b-065706c06e8f`, i think the bucket name limit is 63 characters, so I'm not sure to understand how it can work? At least we should make is smaller

BourgoisMickael: Oh right maybe that's why I have multiple-backend + cepth tests failing. I guess they are the only ones testing backbeat route

```js
[AWS s3 400 0.025s 0 retries] headBucket({
  Bucket: 'backbeatbucket-non-versioned-96c3e932-f74c-4219-b7fe-db7b73af7e51'
})
Error creating bucket: BadRequest: null
```

PR #5913, opened 2025-08-25
CLDSRV-632 ✨ put metadata edge cases

DarkIsDude:
## Description

Reactivate some tests that were disable because some edge case were not covered.

### Motivation and context

Manage 2 use case under the put metadata route:
- Manage null version when an object is created before bucket versioning and put metadata called after bucket versioning enabled

```
/* eslint-disable no-console */
const util = require('util');

const { makeBackbeatRequest: makeBackbeatRequestBase } = require('./tests/functional/raw-node/utils/makeRequest');
const { models } = require('arsenal');
const { ObjectMD } = models;
const BucketUtility = require('./tests/functional/aws-node-sdk/lib/utility/bucket-util');

const makeBackbeatRequest = util.promisify(makeBackbeatRequestBase);

function objectMDFromRequestBody(data) {
    const bodyStr = JSON.parse(data.body).Body;
    return new ObjectMD(JSON.parse(bodyStr));
}

async function main() {
    const OBJECT_KEY = 'test-object';
    const BUCKET_NAME = 'test-bucket';
    const TEST_DATA = 'This is a test object for replication';

    const authCredentials = {
        accessKey: 'BJ8Q0L35PRJ92ABA2K0B',
        secretKey: 'kTgcfEaLjxvrLN5EKVcTnb4Ac046FU1m=33/baf1',
    };


    const bucketUtil = new BucketUtility('local-test-vault-s3', { signatureVersion: 'v4' });
    const s3 = bucketUtil.s3;

    console.info('Starting test for object metadata replication...');
    console.info('Checking if bucket exists...');
    await bucketUtil.emptyIfExists(BUCKET_NAME);

    if (await bucketUtil.bucketExists(BUCKET_NAME)) {
        console.info('Deleting bucket...');
        await s3.deleteBucket({ Bucket: BUCKET_NAME }).promise();
    }

    console.info('Creating bucket...');
    await s3.createBucket({ Bucket: BUCKET_NAME }).promise();
 
    console.info('Putting object with versioning disabled...');
    await s3.putObject({ Bucket: BUCKET_NAME, Key: OBJECT_KEY, Body: Buffer.from(TEST_DATA) }).promise();

    console.info('Enabling versioning on bucket...');
    await s3.putBucketVersioning({ Bucket: BUCKET_NAME, VersioningConfiguration: { Status: 'Enabled' } }).promise();
    const versionId = null;
    console.info('Retrieve metadata for the object with versioning enabled...');
    const data = await makeBackbeatRequest({
        method: 'GET',
        resourceType: 'metadata',
        bucket: BUCKET_NAME,
        objectKey: OBJECT_KEY,
        queryObj: {
            versionId,
        },
        authCredentials,
    });

    const objMD = objectMDFromRequestBody(data)
        .setContentLanguage('fr-FR')
        .getSerialized();

    console.info('Object metadata retrieved successfully:', objMD);
    console.info('Updating object metadata...');

    objMD.tags = {
        'fuck': 'test-value',
    };

    await makeBackbeatRequest({
        method: 'PUT',
        resourceType: 'metadata',
        bucket: BUCKET_NAME,
        objectKey: OBJECT_KEY,
        queryObj: {
            versionId,
        },
        authCredentials,
        requestBody: objMD,
    });

    console.info('Object metadata updated successfully.');

    const versions = await s3.listObjectVersions({
        Bucket: BUCKET_NAME,
        Prefix: OBJECT_KEY,
    }).promise();

    console.info({ versions });
}

main();
```

- Add missing metadata

```
/* eslint-disable no-console */
const util = require('util');

const { makeBackbeatRequest: makeBackbeatRequestBase } = require('./tests/functional/raw-node/utils/makeRequest');
const { models } = require('arsenal');
const { ObjectMD } = models;
const BucketUtility = require('./tests/functional/aws-node-sdk/lib/utility/bucket-util');

const makeBackbeatRequest = util.promisify(makeBackbeatRequestBase);

/**
 * The final result should be to have two versions of the object:
 * 1. The original version created when versioning was disabled (null versionId)
 * 2. A new version created when versioning was re-enabled (with a VersionId)
 * The new version should have the updated metadata (ContentLanguage set to 'fr-FR')
 * while the original version should remain unchanged.
 **/

const OBJECT_KEY_ROOT = 'test-object-root';
const OBJECT_KEY = 'test-object';
const BUCKET_NAME = 'test-bucket';
const TEST_DATA = 'This is a test object for replication';

const authCredentials = {
    accessKey: 'BJ8Q0L35PRJ92ABA2K0B',
    secretKey: 'kTgcfEaLjxvrLN5EKVcTnb4Ac046FU1m=33/baf1',
};

function objectMDFromRequestBody(data) {
    const bodyStr = JSON.parse(data.body).Body;
    return new ObjectMD(JSON.parse(bodyStr));
}

async function main() {
    const bucketUtil = new BucketUtility('local-test-vault-s3', { signatureVersion: 'v4' });
    const s3 = bucketUtil.s3;

    // await initTest(s3, bucketUtil);
    await updateObjectMetadata(s3);
}

async function initTest(s3, bucketUtil) {
    if (await bucketUtil.bucketExists(BUCKET_NAME)) {
        await bucketUtil.emptyIfExists(BUCKET_NAME);
        await s3.deleteBucket({ Bucket: BUCKET_NAME }).promise();
    }

    await s3.createBucket({ Bucket: BUCKET_NAME }).promise();
    await s3.putBucketVersioning({ Bucket: BUCKET_NAME, VersioningConfiguration: { Status: 'Enabled' } }).promise();
    await s3.putObject({ Bucket: BUCKET_NAME, Key: OBJECT_KEY_ROOT, Body: Buffer.from(TEST_DATA) }).promise();
    await s3.putBucketVersioning({ Bucket: BUCKET_NAME, VersioningConfiguration: { Status: 'Suspended' } }).promise();
    await s3.putObject({ Bucket: BUCKET_NAME, Key: OBJECT_KEY, Body: Buffer.from(TEST_DATA) }).promise();
    await s3.putBucketVersioning({ Bucket: BUCKET_NAME, VersioningConfiguration: { Status: 'Enabled' } }).promise();
}

async function updateObjectMetadata(s3) {
        const data = await makeBackbeatRequest({
        method: 'GET',
        resourceType: 'metadata',
        bucket: BUCKET_NAME,
        objectKey: OBJECT_KEY,
        queryObj: {
            versionId: null,
        },
        authCredentials,
    });

    const objMD = objectMDFromRequestBody(data)
        .setContentLanguage('fr-FR')
        .getSerialized();

    await makeBackbeatRequest({
        method: 'PUT',
        resourceType: 'metadata',
        bucket: BUCKET_NAME,
        objectKey: OBJECT_KEY,
        queryObj: {
            versionId: '393832343336313437373734363139393939393952473030312020333061396264',
        },
        authCredentials,
        requestBody: objMD,
    });

    const versions = await s3.listObjectVersions({
        Bucket: BUCKET_NAME,
        Prefix: OBJECT_KEY,
    }).promise();

    console.info({ Versions: versions.Versions });
 }

async function run() {
    try {
        await main();
    } catch (e) {
        console.error('Error running test:', e);
        process.exit(1);
    }
}

run();
```

### Related issues

https://scality.atlassian.net/browse/CLDSRV-632
https://github.com/scality/Arsenal/pull/2490

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -218,6 +218,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
                     metadataStoreParams.contentMD5 = constants.emptyFileMd5;
                     return next(null, null, null);
                 }
+
                 // Handle mdOnlyHeader as a metadata only operation. If
                 // the object in question is actually 0 byte or has a body size
                 // then handle normally.
@@ -244,6 +245,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
                     return next(null, dataGetInfo, _md5);
                 }
             }
+
             return dataStore(objectKeyContext, cipherBundle, request, size,
                     streamingV4Params, backendInfo, log, next);
         },
@@ -280,10 +282,12 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
                 const options = overwritingVersioning(objMD, metadataStoreParams);
                 return process.nextTick(() => next(null, options, infoArr));
             }
+
             if (!bucketMD.isVersioningEnabled() && objMD?.archive?.archiveInfo) {
                 // Ensure we trigger a "delete" event in the oplog for the previously archived object
                 metadataStoreParams.needOplogUpdate = 's3:ReplaceArchivedObject';
             }
+
             return versioningPreprocessing(bucketName, bucketMD,
                 metadataStoreParams.objectKey, objMD, log, (err, options) => {
                     if (err) {
@@ -316,9 +320,11 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
             metadataStoreParams.versioning = options.versioning;
             metadataStoreParams.isNull = options.isNull;
             metadataStoreParams.deleteNullKey = options.deleteNullKey;
+
             if (options.extraMD) {
                 Object.assign(metadataStoreParams, options.extraMD);
             }
+
             return _storeInMDandDeleteData(bucketName, infoArr,
                 cipherBundle, metadataStoreParams,
                 options.dataToDelete, log, requestMethod, next);
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -694,6 +694,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
 
                     if (versioningPreprocessingResult?.nullVersionId) {
                         omVal.nullVersionId = versioningPreprocessingResult.nullVersionId;
+                        options.deleteNullKey = versioningPreprocessingResult.deleteNullKey;
+
+                        if (versioningPreprocessingResult.extraMD) {
+                            Object.assign(omVal, options.extraMD);
+                        }
                     }
                 }
             },
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -665,7 +665,21 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             },
             async () => {
                 if (versioning && !objMd) {
-                    const masterMD = await metadata.getObjectMDPromised(bucketName, objectKey, {}, log);
+                    let masterMD;
+
+                    try {
+                        masterMD = await metadata.getObjectMDPromised(bucketName, objectKey, {}, log);
+                    } catch (err) {
+                        if (err.is?.NoSuchKey) {
+                            log.debug('no master found for versioned object', {
+                                method: 'putMetadata',
+                                bucketName,
+                                objectKey,
+                            });
+                        } else {
+                            throw err;
+                        }
+                    }
 
                     if (!masterMD) {
                         return;
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -616,6 +616,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // To prevent this, the versionId field is only included in options when it is defined.
         if (versionId !== undefined) {
             options.versionId = versionId;
+            omVal.versionId = versionId;
+
+            if (isNull) {
+                omVal.isNull = isNull;
+            }
 
             // In the MongoDB metadata backend, setting the versionId option leads to the creation
             // or update of the version object, the master object is only updated if its versionId
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -620,6 +620,10 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             omVal.versionId = versionId;
 
             if (isNull) {
+                if (!nullVersionCompatMode) {
+                  omVal.isNull2 = true;
+                }
+
                 omVal.isNull = isNull;
             }
 
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -661,6 +661,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 if (versioning && !objMd) {
                     const masterObjectAndBucket =
                         await metadata.getBucketAndObjectMDPromised(bucketName, objectKey, {}, log);
+
+                    if (!masterObjectAndBucket.obj) {
+                        return;
+                    }
+
                     const masterObject = JSON.parse(masterObjectAndBucket.obj);
                     await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterObject, log);
                 }
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -667,7 +667,12 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     }
 
                     const masterObject = JSON.parse(masterObjectAndBucket.obj);
-                    await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterObject, log);
+                    const versioningPreprocessingResult =
+                        await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterObject, log);
+
+                    if (versioningPreprocessingResult?.nullVersionId) {
+                        omVal.nullVersionId = versioningPreprocessingResult.nullVersionId;
+                    }
                 }
             },
             next => {
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -45,6 +45,7 @@ const { versioningPreprocessing }
 const {promisify} = require('util');
 
 const versioningPreprocessingPromised = promisify(versioningPreprocessing);
+metadata.getObjectMDPromised = promisify(metadata.getObjectMD);
 metadata.getBucketAndObjectMDPromised = promisify(metadata.getBucketAndObjectMD);
 
 const { CURRENT_TYPE, NON_CURRENT_TYPE, ORPHAN_DM_TYPE } = constants.lifecycleListing;
@@ -664,16 +665,14 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             },
             async () => {
                 if (versioning && !objMd) {
-                    const masterObjectAndBucket =
-                        await metadata.getBucketAndObjectMDPromised(bucketName, objectKey, {}, log);
+                    const masterMD = await metadata.getObjectMDPromised(bucketName, objectKey, {}, log);
 
-                    if (!masterObjectAndBucket.obj) {
+                    if (!masterMD) {
                         return;
                     }
 
-                    const masterObject = JSON.parse(masterObjectAndBucket.obj);
                     const versioningPreprocessingResult =
-                        await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterObject, log);
+                        await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterMD, log);
 
                     if (versioningPreprocessingResult?.nullVersionId) {
                         omVal.nullVersionId = versioningPreprocessingResult.nullVersionId;
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -621,7 +621,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
 
             if (isNull) {
                 if (!nullVersionCompatMode) {
-                  omVal.isNull2 = true;
+                    omVal.isNull2 = true;
                 }
 
                 omVal.isNull = isNull;
@@ -669,7 +669,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             },
             async () => {
                 // If we create a new version of an object (so objMd is null),
-                // we should make sure that the masterVersion is versionned.
+                // we should make sure that the masterVersion is versioned.
                 // If an object already exists, we just want to update the metadata
                 // of the existing object and not create a new one
                 if (versioning && !objMd) {
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -652,7 +652,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 });
             },
             next => {
-                versioningPreprocessing(bucketName, bucketInfo, objectKey, objMd, log, next);
+                if (!versionId) {
+                    return next();
+                }
+
+                return versioningPreprocessing(bucketName, bucketInfo, objectKey, objMd, log, next);
             },
             next => {
                 log.trace('putting object version', {
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -42,6 +42,10 @@ const quotaUtils = require('../api/apiUtils/quotas/quotaUtils');
 const { handleAuthorizationResults } = require('../api/api');
 const { versioningPreprocessing }
     = require('../api/apiUtils/object/versioning');
+const {promisify} = require('util');
+
+const versioningPreprocessingPromised = promisify(versioningPreprocessing);
+metadata.getBucketAndObjectMDPromised = promisify(metadata.getBucketAndObjectMD);
 
 const { CURRENT_TYPE, NON_CURRENT_TYPE, ORPHAN_DM_TYPE } = constants.lifecycleListing;
 
@@ -612,6 +616,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // To prevent this, the versionId field is only included in options when it is defined.
         if (versionId !== undefined) {
             options.versionId = versionId;
+
             // In the MongoDB metadata backend, setting the versionId option leads to the creation
             // or update of the version object, the master object is only updated if its versionId
             // is the same as the version. This can lead to inconsistencies when replicating objects
@@ -642,6 +647,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 if (!request.query?.accountId) {
                     return next();
                 }
+
                 return getCanonicalIdsByAccountId(request.query.accountId, log, (err, res) => {
                     if (err) {
                         return next(err);
@@ -651,12 +657,13 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     return next();
                 });
             },
-            next => {
-                if (!versionId) {
-                    return next();
+            async () => {
+                if (versioning && !objMd) {
+                    const masterObjectAndBucket =
+                        await metadata.getBucketAndObjectMDPromised(bucketName, objectKey, {}, log);
+                    const masterObject = JSON.parse(masterObjectAndBucket.obj);
+                    await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterObject, log);
                 }
-
-                return versioningPreprocessing(bucketName, bucketInfo, objectKey, objMd, log, next);
             },
             next => {
                 log.trace('putting object version', {
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -631,7 +631,6 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         }
 
         return async.series([
-            next => versioningPreprocessing(bucketName, bucketInfo, objectKey, objMd, log, next),
             // Zenko's CRR delegates replacing the account
             // information to the destination's Cloudserver, as
             // Vault admin APIs are not exposed externally.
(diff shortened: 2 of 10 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -650,6 +657,14 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     return next();
                 });
             },
+            async () => {
+                if (versioning && !objMd) {
```
francoisferrand, line 675: when does this `!objMd` case happen? Before this function is called, we already call `standardMetadataValidateBucketAndObj` (line 1655), which should return the object already...
(and looking at versioningPreprocessing, it seems that i knows how to handle objMd=nil as well as the case where object is/isn't the master?)

lib/routes/routeBackbeat.js
```diff
@@ -650,6 +657,14 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                    const masterObjectAndBucket =
+                        await metadata.getBucketAndObjectMDPromised(bucketName, objectKey, {}, log);
```
francoisferrand: that is 2 extra I/Os: to get the bucket then get the object.
We already retrieved that in `standardMetadataValidateBucketAndObj`, so
- we definitely already have the bucket -> no point retrieving it again
- if we try to get the same object, we will get the same result i guess; if not did we actually need the first one? (i.e. could we fix that call instead of making another one here)

lib/routes/routeBackbeat.js
```diff
@@ -650,6 +657,14 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                    const masterObject = JSON.parse(masterObjectAndBucket.obj);
+                    await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterObject, log);
```
francoisferrand: putMetadata can be used in 2 ways:
- in CRR, it is most of the time used to "create" (or replace) object: it similar to a regular putObject, but dealing only with metadata (CRR replicates data and metadata separately)
  --> there it may need to just replace the document (non-versionned case), create a new one (no previous object or versionned case), or create a new one + delete previous one (version-suspended case)
- in backbeat, it is used to "update" the metadata of an existing object: and it should never create a new document/revision
- (in CRR, there may also be "real" metadata updates on the source, which get replicated the way: so the call from CRR may also be used to udpate the MD of an existing object, as backbeat does..... what may help though it that CRR can only ever happen in version-enabled buckets)

Looking at the code in putObject, we mostly handle the first case, though we have a corner case "putObjectVersion", which is more related to the second case (it is used to restore cold objects, and thus will write the data + simply update the metadata of existing document).

In that case, we actually must not (and do not 😁) call `versioningPreprocessingPromised` : so it seems we should have something similar here, and not call `versioningPreprocessingPromised` in the "update metadata" case...

Maybe that is the purpose of the `!objMd` condition, but it is a bit hard to see...

francoisferrand: For larger context : I believe that we will eventually need to be able to differentiate "updates" from "puts", as there are some subbtle difference in handle race conditions/corner cases.

In particular, we may need to do this distinction to fix https://scality.atlassian.net/browse/ARTESCA-8449 (backbeat updates metadata by doing getMetadata...putMetadata, we multiple such sequences may happen in parallel). To fix this (which rarely happen), the idea is to introduce some some sort to `counter` and perform update with a condition on that counter: but it must be done _only_ when the semantic is that of updating existing MD, not when the update is actually the replacement of a version.

Not sure it is really needed for this specific fix, but best to know this in case your change may go in similar direction.

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -82,48 +82,44 @@ class BucketUtility {
...
-                            .map(object =>
-                                 this.s3.deleteObject({
-                                     Bucket: bucketName,
-                                     Key: object.Key,
-                                     VersionId: object.VersionId,
-                                 }).promise()
-                                 .then(() => object)))
-                )
-            );
+        const listedObjects = await this.s3.listObjectVersions(param).promise();
```
francoisferrand, line 93: how many objects are (typically) talking about in these tests?
could this affect the duration of the tests?

(or can we fix the backend so we can continue to delete in parallel?)

lib/routes/routeBackbeat.js
DarkIsDude: @francoisferrand when you do a PUT on PutMD and the object does not exists. This case can happens when it's an empty object

lib/routes/routeBackbeat.js
DarkIsDude: In this case I try to get the master / null version, so not the one that we want (objMD is undefined so we didn't find it). For the bucket I'll remove this extra call 🙏

lib/routes/routeBackbeat.js
```diff
@@ -611,6 +617,12 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // To prevent this, the versionId field is only included in options when it is defined.
         if (versionId !== undefined) {
             options.versionId = versionId;
+            omVal.versionId = versionId;
+
+            if (isNull) {
+                omVal.isNull = isNull;
```
francoisferrand, line 627: is there maybe `null2` field to handle as well ?
(somehow required for metadata backend, with the latest listing)

lib/routes/routeBackbeat.js
```diff
@@ -611,6 +617,12 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // To prevent this, the versionId field is only included in options when it is defined.
         if (versionId !== undefined) {
             options.versionId = versionId;
+            omVal.versionId = versionId;
```
francoisferrand, line 620: the semantics of putObject in metadata layer (either backends, i.e. metadata or mongodbClientInterface) are not clear or precisely defined : does setting this not interract (in some weird or unacceptable way?) the behavior of the `versionId` option?

lib/routes/routeBackbeat.js
francoisferrand: still not clear why we need the conditional, and not (more) systematically call versioningPreprocessing.
in particular, this function is responsible to handle version-suspended case (i.e. delete previous version). Or is this case objMD == nil a way to detect the insertion case (i.e. create a new version) instead of the udpate case (update -internal- metadata of an existing version, e.g. for replication status or transition)?

may help to have a comment explaining what use-case this "branch" handles.

> This case can happens when it's an empty object

do you mean "empty" as in "no data"? I don't understand how it would/should affect the behavior here :-/

lib/routes/routeBackbeat.js
```diff
@@ -650,6 +663,22 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                    if (!masterMD) {
+                        return;
+                    }
+
+                    const versioningPreprocessingResult =
+                        await versioningPreprocessingPromised(bucketName, bucketInfo, objectKey, masterMD, log);
+
+                    if (versioningPreprocessingResult?.nullVersionId) {
+                        omVal.nullVersionId = versioningPreprocessingResult.nullVersionId;
+                    }
```
francoisferrand, line 706: in other uses of versioningPreprocessing (in putObject / copyObject / putMpu) we set multiple (and different) fields:

```
			metadataStoreParams.versionId = options.versionId;
            metadataStoreParams.versioning = options.versioning;
            metadataStoreParams.isNull = options.isNull;
            metadataStoreParams.deleteNullKey = options.deleteNullKey;
            if (options.extraMD) {
                Object.assign(metadataStoreParams, options.extraMD);
            }
```

this gets translated (in `metadataStoreObject`) to this:
```
        if (versioning) {
            options.versioning = versioning;
        }
        if (versionId || versionId === '') {
            options.versionId = versionId;
        }

        if (deleteNullKey) {
            options.deleteNullKey = deleteNullKey;
        }

        const { isNull, nullVersionId, nullUploadId, isDeleteMarker } = params;
        md.setIsNull(isNull)
            .setNullVersionId(nullVersionId)
            .setNullUploadId(nullUploadId)
            .setIsDeleteMarker(isDeleteMarker);
        if (versionId && versionId !== 'null') {
            md.setVersionId(versionId);
        }
        if (isNull && !config.nullVersionCompatMode) {
            md.setIsNull2(true);
        }
```

→ so it seems many more fields may be set, to cover all cases?

francoisferrand (commented):
somehow I am not sure this covers all cases - and not confident everything is fully tests: there are tests, but logic is quite intricate with the different versioning cases + different listings/... and all the flags: isNull, isNull2, ...

→ would be worth reviewing the tests to see if we miss some cases (e.g. all the combinations of versioning disabled/enabled/suspended, creating vs updating an object, with previous version/nullVersion/no previous version...). May help to look at putObject tests for inspiration on the test case?
→ since we have multiple backend, and the separation is sometimes not clear (or not clearly documented), we should make sure this test is run both with mongo and metadata backends: is it the case?

lib/routes/routeBackbeat.js
```diff
@@ -650,6 +664,44 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                    if (versioningPreprocessingResult?.nullVersionId) {
+                        omVal.nullVersionId = versioningPreprocessingResult.nullVersionId;
+                        omVal.versionId = versioningPreprocessingResult.versionId;
+                        omVal.versioning = versioningPreprocessingResult.versioning;
+                        omVal.isNull = versioningPreprocessingResult.isNull;
+                        omVal.deleteNullKey = versioningPreprocessingResult.deleteNullKey;
+
+                        if (options.extraMD) {
+                            Object.assign(omVal, versioningPreprocessingResult.extraMD);
+                        }
```
francoisferrand: nit: (not sure the `|| {}`) is even needed, but seems best to avoid the extra cognitive complexity?
```suggestion
                        Object.assign(omVal, versioningPreprocessingResult.extraMD || {});
```

lib/routes/routeBackbeat.js
DarkIsDude: @francoisferrand if we have an objMd (so an object already exists), we don't want to call this function as this function will create a new version and we don't want that. So I would say it's more to detect indeed the insertion case and the update case.

Do you think a comment like `If we create a new version of an object (so objMd is null), we should make sure that the masterVersion is versionned. If an object already exists, we just want to update the metadata of the existing object and not create a new one` ?

The `versionning` is just an "optimisation" but can be removed 🙏.

lib/routes/routeBackbeat.js
DarkIsDude: That's good a point, thanks for pointing it. `versionId;versioning;isNull;` are already managed by the code before. So that don't make sense to manage them again here 🙏.

I added a the extraMD and the deleteNullKey !

lib/routes/routeBackbeat.js
DarkIsDude: As far as I checked, this seems not be the case. But maybe I'm missing something here 😬. Maybe the second review will have more insight

DarkIsDude:
> somehow I am not sure this covers all cases - and not confident everything is fully tests: there are tests, but logic is quite intricate with the different versioning cases + different listings/... and all the flags: isNull, isNull2, ...
> 
> → would be worth reviewing the tests to see if we miss some cases (e.g. all the combinations of versioning disabled/enabled/suspended, creating vs updating an object, with previous version/nullVersion/no previous version...). May help to look at putObject tests for inspiration on the test case? → since we have multiple backend, and the separation is sometimes not clear (or not clearly documented), we should make sure this test is run both with mongo and metadata backends: is it the case?

So I took time to review tests and path. With the condition `if (versioning && !objMd)` there is not a lot of path that can jump in that code. So I added some tests but I didn't find more of them. Keep in mind that I don't have an overview of all features / options

lib/routes/routeBackbeat.js
```diff
@@ -508,23 +515,22 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
-        // check if it's metadata only operation
+
         if (headers['x-scal-replication-content'] === 'METADATA') {
             if (!objMd) {
-                // if the target does not exist, return an error to
-                // backbeat, who will have to retry the operation as a
-                // complete replication
                 return callback(errors.ObjNotFound);
             }
-            // use original data locations and encryption info
```
francoisferrand, line 527: these comments do not seem useless at all, they should be kept

lib/routes/routeBackbeat.js
francoisferrand: did you perform such second review? (you can resolve this comment once you confirmed)

lib/routes/routeBackbeat.js
```diff
@@ -650,6 +667,45 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+            async () => {
+                // If we create a new version of an object (so objMd is null),
+                // we should make sure that the masterVersion is versionned.
+                // If an object already exists, we just want to update the metadata
+                // of the existing object and not create a new one
+                if (versioning && !objMd) {
+                    let masterMD;
+
+                    try {
+                        masterMD = await metadata.getObjectMDPromised(bucketName, objectKey, {}, log);
```
francoisferrand, line 679: this introduces an extra metadata IO in case of insert (first one was in `standardMetadataValidateBucketAndObj`, which returned -or not- the _version_ document) : this extra I/O can be skipped if we know the expected semantics of the `putMetadata` call (i.e. is it an insertion → can directly lookup the "master" in `standardMetadataValidateBucketAndObj` ; or an update → must lookup the version) -OR- update `standardMetadataValidateBucketAndObj` (and everything below) to be able to return both the version and/or master...

➡ maybe acceptable to keep the extra I/O for now, but we should create a ticket for tracking this, or make a note in https://scality.atlassian.net/browse/ARTESCA-8449

francoisferrand (approved):

lib/api/apiUtils/object/createAndStoreObject.js
ghost, line 1: This will probably be automated soon with some prettier solutions
I started this but it was on hold as we wanted to finish the unification first https://github.com/scality/cloudserver/pull/5877

tests/unit/routes/routeBackbeat.js
```diff
@@ -531,6 +559,67 @@ describe('routeBackbeat', () => {
...
+                callback(null, bucketInfo, notFoundObject);
+            });
+
+            routeBackbeat('127.0.0.1', mockRequest, mockResponse, log);
+            await endPromise;
+
+            sinon.assert.calledOnce(metadataGetObjectMDPromisedStub);
+            sinon.assert.calledTwice(metadataPutObjectMDStub);
+            sinon.assert.calledWith(
+                metadataPutObjectMDStub.firstCall,// Transform the non versioned object to a versioned one
```
ghost:
```suggestion
                metadataPutObjectMDStub.firstCall, // Transform the non versioned object to a versioned one
```

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -17,7 +17,8 @@ class BucketUtility {
 
     bucketExists(bucketName) {
         return this.s3
-            .headBucket({ Bucket: bucketName }).promise()
+            .headBucket({ Bucket: bucketName })
+            .promise()
```
ghost, line 21: These changes are unrelated to this PR and will likely conflict with Maha's, work, can we drop this? this will be handled by a prettier later

lib/routes/routeBackbeat.js
```diff
@@ -611,6 +617,16 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // To prevent this, the versionId field is only included in options when it is defined.
         if (versionId !== undefined) {
             options.versionId = versionId;
+            omVal.versionId = versionId;
+
+            if (isNull) {
+                if (!nullVersionCompatMode) {
+                  omVal.isNull2 = true;
+                }
```
ghost, line 625:
```suggestion
                if (!nullVersionCompatMode) {
                    omVal.isNull2 = true;
                }
```

lib/routes/routeBackbeat.js
```diff
@@ -650,6 +667,45 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     return next();
                 });
             },
+            async () => {
+                // If we create a new version of an object (so objMd is null),
+                // we should make sure that the masterVersion is versionned.
```
ghost:
```suggestion
                // we should make sure that the masterVersion is versioned.
```

lib/routes/routeBackbeat.js
```diff
@@ -650,6 +667,45 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                        masterMD = await metadata.getObjectMDPromised(bucketName, objectKey, {}, log);
+                    } catch (err) {
+                        if (err.is?.NoSuchKey) {
+                            log.debug('no master found for versioned object', {
+                                method: 'putMetadata',
+                                bucketName,
+                                objectKey,
+                            });
+                        } else {
+                            throw err;
```
ghost, line 688: I'm not sure we want to throw, the current stack is not using promises yet, so we tend to use more errors and calling the callbacks with it. I fear that a throw here will make the process crash, as we do not try/catch our code?


```suggestion
                            return next(err);
```

lib/routes/routeBackbeat.js
ghost: Side note, if we want to get both the master and the version, then this is just moving the problem, as we will still need two queries. Today we try to get the master version only for objects like delete markers, but if we provide a versionID and it's not there, then we just return `NoSuchVersion`. Unless we kind of batch or use operators like `$in`, this will probably lead to more I/Os in any cases. So, important that if we go with this approach, we do not affect all API calls outside of backbeat routes

The best would be to know in advance

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
DarkIsDude: I have mixed feeling about waiting on this one. Waiting for prettier (we don't know when we'll be able to add it) delay the tech debt... And apply this mindset widely introduce a frozen one that I'm not fan 😬. It's the same things as this one https://github.com/scality/cloudserver/pull/5913#pullrequestreview-3344451568

lib/routes/routeBackbeat.js
DarkIsDude: We can throw safely. We are in an `async` process. Async is able to switch from 'async/await' to 'callback'. Async will transform this throw in a callback one 🙏 

```
import async from "async";

function main() {
  async.series(
    [
      function (callback) {
        setTimeout(function () {
          console.log("Task 1");
          callback(null, "one");
        }, 200);
      },
      async function () {
        await new Promise((resolve) => {
          setTimeout(function () {
            console.log("Task 1.5");
            resolve();
          }, 150);
        });

        throw new Error("Error in Task 1.5");
      },
    ],
    function (err, results) {
      console.error(err);
      console.log(results);
    },
  );
}

main();
```

But I'm open minded as you mention that the current stack don't use promises yet. I would like to start to change that and move to a more modern stack by starting async/await syntax, in parallel of the SDK migration that introduce a lot async/await. Let me know

lib/routes/routeBackbeat.js
DarkIsDude: The one for location and encryption don't add any information ? Reading the code is enough to understand that we reuse the objMd ?

For the other one, the comment don't help to understand the code neither, he just describe what happens later. IMO explaining what happens in other component just add noise. If we do that everywhere, we'll have comment everywhere. A comment should be here to help to understand the code, some weird behaviour or exception ?

ghost (approved):

PR #5914, opened 2025-08-25
CLDSRV-717: Test S3C setup like in Integration

BourgoisMickael:
To catch errors faster than waiting to run Integration

Note: the commit [788a356](https://github.com/scality/cloudserver/pull/5914/commits/788a35632ccb610f4f8d9b93f6cf8f53c77cb7a8) will be reverted once all tests are fixed for S3C

### DIFF
--- a/.github/docker/setup-s3c.sh
+++ b/.github/docker/setup-s3c.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# setup S3C environment just like
+# https://github.com/scality/Integration/blob/development/9.5/tests/setup-environment/index.js
+
+set -e -o pipefail
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+VAULTCLIENT=$SCRIPT_DIR/../../node_modules/vaultclient/bin/vaultclient
+CONFIG=$SCRIPT_DIR/admin.json
+
+echo "Deleting and setting up S3C accounts like Integration (follow conf/authdata.json)"
+
+$VAULTCLIENT --config $CONFIG delete-account --name Bart || true
+$VAULTCLIENT --config $CONFIG create-account \
+    --name Bart \
+    --email sampleaccount1@sampling.com \
+    --accountid 123456789012 \
+    --canonicalid 79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be
+$VAULTCLIENT --config $CONFIG generate-account-access-key \
+    --name Bart \
+    --accesskey ACC1AK00000000000000 \
+    --secretkey ACC1SK0000000000000000000000000000000000
+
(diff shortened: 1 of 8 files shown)


anurag4DSB (approved):
Do we want to deploy vault as well if we want to do it like integration?

.github/docker/setup-s3c.sh
```diff
@@ -0,0 +1,62 @@
...
+    --canonicalid 79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2ba
+$VAULTCLIENT --config $CONFIG generate-account-access-key \
+    --name Replication \
+    --accesskey ACCREPAK000000000000 \
+    --secretkey ACCREPSK00000000000000000000000000000000
+
+echo "Copying s3c credentials to mem credentials"
+cp \
+    $SCRIPT_DIR/../../tests/functional/aws-node-sdk/lib/json/s3c_credentials.json \
+    $SCRIPT_DIR/../../tests/functional/aws-node-sdk/lib/json/mem_credentials.json
```
tcarmet, line 53: Is this file supposed to be added in the `.gitignore`?

.github/docker/setup-s3c.sh
```diff
@@ -0,0 +1,62 @@
...
+
+$VAULTCLIENT --config $CONFIG delete-account --name Bart || true
+$VAULTCLIENT --config $CONFIG create-account \
+    --name Bart \
+    --email sampleaccount1@sampling.com \
+    --accountid 123456789012 \
+    --canonicalid 79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be
+$VAULTCLIENT --config $CONFIG generate-account-access-key \
+    --name Bart \
+    --accesskey ACC1AK00000000000000 \
```
tcarmet, line 23: Since we are specifying those credentials in `s3c_credentials.json` should we read from it?

tcarmet (approved):
Pretty awesome call into executing this testsuite here 👌 

I have to state though that the setup is starting to look a bit similar to a workbench setup. I don't think workbench have all the necessary config capabilities/code to make it equal, but I think it's worth you having a quick look at it, and share some thoughts onto which direction we should go to eventually add those capabilities in.

BourgoisMickael:
> Do we want to deploy vault as well if we want to do it like integration?

@anurag4DSB Yes vault is deployed as well (the latest version with node 22)

.github/docker/setup-s3c.sh
BourgoisMickael: No this is executed in the CI

.github/docker/setup-s3c.sh
BourgoisMickael: Not necessarily, I've hardcoded here so we can easily rerun manually and quickly troubleshoot without having to look for creds. Same for Integration, it will be hardcoded

PR #5917, opened 2025-08-26
always evaluate policies in backbeat routes

welansari:
Internal services have the required permissions to access backbeat routes, there is no longer the need to skip policy evaluation here.

Issue: CLDSRV-731


DarkIsDude (approved):

SylvainSenechal (approved):

tests/unit/routes/routeBackbeat.js
```diff
@@ -1012,28 +1012,6 @@ describe('routeBackbeat authorization', () => {
                 const err = JSON.parse(response.end.getCall(0).args[0]);
                 assert.strictEqual(err.code, 'AccessDenied');
             });
-
-            it('should bypass policy evaluation', async () => {
```
francoisferrand, line 1016: maybe keep this test: it should evaluate policies (but not bucket policies) even when `bypassUserBucketPolicies` (i.e. on internal routes)

ghost (approved):

PR #5918, opened 2025-08-26
CLDSRV-732: Fix flakiness starting cloudserver in CI

BourgoisMickael:
The CI fails to wait for S3 after 40s recently

s3-stderr has those logs:
```
npm error code E403
npm error 403 403 Forbidden - GET https://registry.npmjs.org/nyc
npm error 403 In most cases, you or one of your dependencies are requesting
npm error 403 a package version that is forbidden by your security policy, or
npm error 403 on a server you do not have access to.
npm error A complete log of this run can be found in: /root/.npm/_logs/2025-08-26T12_00_55_746Z-debug-0.log
```

Build another image to include nyc and a command for coverage during tests

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -48,7 +48,18 @@ jobs:
         with:
           context: .
           push: true
+          target: production
           tags: ghcr.io/${{ github.repository }}:${{ github.event.inputs.tag }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Build and push test coverage image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          target: testcoverage
+          tags: ghcr.io/${{ github.repository }}:${{ github.event.inputs.tag }}-testcoverage
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -178,9 +178,25 @@ jobs:
         with:
           push: true
           context: .
+          target: production
           provenance: false
           tags: |
             ghcr.io/${{ github.repository }}:${{ github.sha }}
+          labels: |
+            git.repository=${{ github.repository }}
+            git.commit-sha=${{ github.sha }}
+          cache-from: type=gha,scope=cloudserver
+          cache-to: type=gha,mode=max,scope=cloudserver
+
+      - name: Build and push cloudserver image test coverage
+        uses: docker/build-push-action@v5
+        with:
+          push: true
+          context: .
+          target: testcoverage
+          provenance: false
+          tags: |
+            ghcr.io/${{ github.repository }}:${{ github.sha }}-testcoverage
           labels: |
             git.repository=${{ github.repository }}
             git.commit-sha=${{ github.sha }}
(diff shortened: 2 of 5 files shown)


docker-test-with-coverage.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
BourgoisMickael: managed by CI, this is a copy of what's in docker compose

leif-scality (approved):

.github/workflows/release.yaml
```diff
@@ -48,10 +48,21 @@ jobs:
         with:
           context: .
           push: true
+          target: production
           tags: ghcr.io/${{ github.repository }}:${{ github.event.inputs.tag }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+      - name: Build and push test coverage image
```
ghost, line 56: do we really want to create a new tagged image just for the coverage?

ghost (approved):
lgtm, I'm just not sure we need to change the release workflow here

.github/workflows/release.yaml
BourgoisMickael: It's there so you can run the docker compose from the CI for a specific releases version with coverage.

BourgoisMickael: Indeed we could skip coverage outside CI when targetting a specific version.
But the image is small, there's only 1 additinal layer for nyc and a change of CMD, that's 15s build time

.github/workflows/tests.yaml
```diff
@@ -187,6 +188,21 @@ jobs:
...
+          context: .
+          target: testcoverage
+          provenance: false
+          tags: |
+            ghcr.io/${{ github.repository }}:${{ github.sha }}-testcoverage
+          labels: |
+            git.repository=${{ github.repository }}
+            git.commit-sha=${{ github.sha }}
+          cache-from: type=gha,scope=cloudserver
+          cache-to: type=gha,mode=max,scope=cloudserver
```
tcarmet, line 204: To be safe and avoid mixing up caches with the production one.
```suggestion
          cache-from: type=gha,scope=cloudserver-test
          cache-to: type=gha,mode=max,scope=cloudserver-test
```

.github/workflows/release.yaml
```diff
@@ -48,10 +48,21 @@ jobs:
...
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          target: testcoverage
+          tags: ghcr.io/${{ github.repository }}:${{ github.event.inputs.tag }}-testcoverage
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
tcarmet, line 64:
```suggestion
          cache-from: type=gha,scope=cloudserver-test
          cache-to: type=gha,mode=max,scope=cloudserver-test
```

tmacro (approved):

Dockerfile
tcarmet, line 1: While the build with `target` looks quite clean, have you thought about adding `nyc` as a dev dependency (if not already there) and build cloudserver with `NODE_ENV=dev` as a `build-arg` which would then default to production.

I know both way works, but I know Vault has a [similar case already](https://github.com/scality/Vault/blob/development/7.70/.github/workflows/docker-build.yaml) and it would be nice if we keep it consistent across repositories.

Also would suffix it with `-test` in this case.

BourgoisMickael: In this case I don't want to install all devDependencies, only `nyc` is required and nothing else. It seems faster and smaller to apply this small layer on top of the production layer rather than replacing entirely the node_modules layers.

.github/workflows/tests.yaml
BourgoisMickael: We do actually want to mix caches to have faster CI. If the production stage added some dependencies in cache, I'd want to reuse them in the `testcoverage` stage

PR #5919, opened 2025-08-26
CLDSRV-717: Unify test routeBackbeatForReplication

BourgoisMickael:
Split from https://github.com/scality/cloudserver/pull/5907 to simplify review

You can review by commit.

### DIFF
--- a/tests/multipleBackend/routes/routeBackbeatForReplication.js
+++ b/tests/multipleBackend/routes/routeBackbeatForReplication.js
@@ -1628,7 +1628,7 @@ describe('backbeat routes for replication', () => {
                 bucket: bucketSource,
                 objectKey: keyName,
                 queryObj: { versionId: 'null' },
-                authCredentials: destinationAuthCredentials,
+                authCredentials: sourceAuthCredentials,
             }, (err, data) => {
                 if (err) {
                     return next(err);
--- a/tests/multipleBackend/routes/routeBackbeatForReplication.js
+++ b/tests/multipleBackend/routes/routeBackbeatForReplication.js
@@ -20,8 +20,31 @@ const destinationAuthCredentials = {
     secretKey: destinationCreds.secretAccessKey,
 };
 
+// Note: for S3C tests, those conf files needs to be modified beforehand
 const dstAccountInfo = require('../../../conf/authdata.json')
     .accounts.find(acc => acc.name === 'Replication');
+const srcAccountInfo = require('../../../conf/authdata.json')
+    .accounts.find(acc => acc.name === 'Bart');
+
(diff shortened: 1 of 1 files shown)


tests/multipleBackend/routes/routeBackbeatForReplication.js
```diff
@@ -340,13 +371,13 @@ describe('backbeat routes for replication', () => {
...
                     versionId,
-                    accountId: 'invalid',
+                    accountId: '888888888888', // Vault v1 differentiate InvalidAccountId from NoSuchEntity
                 },
                 authCredentials: destinationAuthCredentials,
                 requestBody: objMD,
             }, next),
         }, err => {
-            assert.strictEqual(err.code, 'AccountNotFound');
+            assert.strictEqual(err.code, process.env.S3_END_TO_END ? 'NoSuchEntity' : 'AccountNotFound');
```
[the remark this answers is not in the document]
BourgoisMickael: nah we should unify vault rather

jonathan-gramain (approved):

tests/multipleBackend/routes/routeBackbeatForReplication.js
```diff
@@ -1710,17 +1659,12 @@ describe('backbeat routes for replication', () => {
                 bucket: bucketSource,
                 objectKey: keyName,
                 queryObj: { versionId: 'null' },
-                authCredentials: destinationAuthCredentials,
+                authCredentials: sourceAuthCredentials,
```
ghost, line 1672: Why this change? It looks like we are on purpose getting the replicated null version as per the step name

BourgoisMickael: The bucket is `bucketSource`, not `bucketDestination`

BourgoisMickael: And I get an Access Denied when testing with S3C

BourgoisMickael: And the PUT into the `bucketDestination` is done below. If I try a GET on that bucket I get an `ObjNotFound`.

Now I think maybe we did not get any error in a zenko context because the `replication` credentials are using the zenko replication service account that can access every resources ?

BourgoisMickael: I've describe the test phases and renamed the steps to be clearer: [4e067fd](https://github.com/scality/cloudserver/pull/5919/commits/4e067fdbb908a498d9ab13c69d26d5300e466599)

The GET is done on the source to retrieve the object after lifecycle before replicating it to destination.

So the source credentials should be used on the source bucket

tcarmet (approved):

PR #5922, opened 2025-08-29
CLDSRV-734: Revert back from go to python supervisord

BourgoisMickael:
This will fix fluent-bit access to s3 log files

The go version is not a perfect replacement it has multiple changes that would require to update supervisor conf.

Breaking changes includes:

- log files name changes
  - process number starts count at 1 instead of 0
  - include full process_name with number in logfile
  - s3_1-1.log instead of s3-0.log
- no supervisorctl, replaced by superivsord ctl with less option
- no interactive supervisor ctl, if used without command it rewrite the superivsord.sock and then we lose access to supervisord ctl
- replaced priority with depends_on
- replaced %(ENV_X)s with envFiles

Overrall we’d like to avoid breaking changes until we resolve correct usage of the go version and apply it in every component at once.

### DIFF
--- a/images/federation/Dockerfile
+++ b/images/federation/Dockerfile
@@ -2,17 +2,12 @@ ARG CLOUDSERVER_VERSION=latest
 FROM ghcr.io/scality/cloudserver:${CLOUDSERVER_VERSION} AS builder
 
 ####################################################################################################
-FROM ghcr.io/scality/cloudserver:${CLOUDSERVER_VERSION} AS supervisord
-
-ARG SUPERVISORD_VERSION=0.7.3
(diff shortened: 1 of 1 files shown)


dvasilas (approved):

leif-scality (approved):

BourgoisMickael:
Will not merge for now, we will try to fix the divergence in supervisord.conf and try to use that same version into all our containers to have the same behavior everywhere.

tcarmet (approved):
LGTM but for full fluent-bit access does it need the `scality` group to exists? If so, can you double check if `usermode --login ${USER}` is also creating a `scality` group (not only the user), if not let's modify the user creation to ensure we have one.

BourgoisMickael:
Merging as we don't have time to tacke the breaking changes  for 9.5.1

PR #5924, opened 2025-09-01
🔧 Add nodemonde to enable automatic restart

DarkIsDude:
Issue: CLDSRV-736

# Pull request template

### Motivation and context

Improve DevX

### Related issues

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "mocha-junit-reporter": "^2.2.1",
     "mocha-multi-reporters": "^1.5.1",
     "node-mocks-http": "^1.16.1",
+    "nodemon": "^3.1.10",
     "nyc": "^15.1.0",
     "sinon": "^13.0.1",
     "tv4": "^1.3.0"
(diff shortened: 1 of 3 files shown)


francoisferrand (approved):

package.json
```diff
@@ -67,6 +67,7 @@
     "mocha-junit-reporter": "^2.2.1",
     "mocha-multi-reporters": "^1.5.1",
     "node-mocks-http": "^1.16.1",
+    "nodemon": "^3.1.10",
```
BourgoisMickael, line 70: can you run `yarn global add nodemon` on your laptop instead so it's not included in dependencies for CI ?

ghost: I think this defeats the purpose of using devDependencies? Otherwise why not other dependencies, to be installed globally?

ghost (approved):

package.json
BourgoisMickael: Other dependencies are required for testing in the CI, to run eslint, to run tests, but the github action doesn't use nodemon

BourgoisMickael: nodemon is something you want usually on your dev environment to use it with any repository

ghost: I already had issues with nodemon or other globally-installed projects like that when we use different (incompatible) options across projects, so I think it's better to have the dev env self sufficient by having it in the devDependency... So `yarn dev` always works

DarkIsDude: I kinda agree with William here. Each project should be independent and don't need to ask for anyone to install anything. For example when I did the onboarding, I had an issue with scubaclient to run cloud server. Indeed, Cloudserver has scuba client has a dependency, TS is a dependency of scuba client but not listed, so I need to install it globally... That's not very convinient. Also, adding it in dev dependency we are sure that everyone use the same version and we track usage of packages (CVE, stats, GitHub dependabot, ...).

francoisferrand: > TS is a dependency of scuba client but not listed
for context, there is a deeper issue with TSC: https://scality.slack.com/archives/C031XT2FDA9/p1727949229903809

hopefully should be solved naturally when we publish our packages to (GitHub) npm registry...

BourgoisMickael: Btw with node 22 there is a `--watch` mode directly integrated in to node: https://nodejs.org/en/blog/announcements/v22-release-announce#watch-mode-node---watch

Doc: https://nodejs.org/docs/latest-v22.x/api/cli.html#--watch

DarkIsDude: Are we all on Node22 ? Also this will not cover all our case (for example for Arsenal it'll not neither vault2). I can take time to migrate to it if we are all aligned

DarkIsDude: Cheked with Mickael, we can move forward.

PR #5926, opened 2025-09-02
fix site replication status not displayed for crr

welansari:
Replication status of each site is returned as a custom user metadata
when performing a HeadObject. Returning these headers for all location
types.

Issue: [CLDSRV-737](https://scality.atlassian.net/browse/CLDSRV-737)

[CLDSRV-737]: https://scality.atlassian.net/browse/CLDSRV-737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/utilities/collectResponseHeaders.js
+++ b/lib/utilities/collectResponseHeaders.js
@@ -104,10 +104,7 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
         responseMetaHeaders['x-amz-replication-status'] =
             objectMD.replicationInfo.status;
     }
-    if (objectMD.replicationInfo &&
-        // Use storageType to determine if user metadata is needed.
-        objectMD.replicationInfo.storageType &&
-        Array.isArray(objectMD.replicationInfo.backends)) {
+    if (Array.isArray(objectMD?.replicationInfo?.backends)) {
         objectMD.replicationInfo.backends.forEach(backend => {
             const { status, site, dataStoreVersionId } = backend;
             responseMetaHeaders[`x-amz-meta-${site}-replication-status`] =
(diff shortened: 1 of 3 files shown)


DarkIsDude (approved):

lib/utilities/collectResponseHeaders.js
```diff
@@ -104,11 +104,13 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg,
...
             objectMD.replicationInfo.status;
     }
-    if (objectMD.replicationInfo &&
-        // Use storageType to determine if user metadata is needed.
-        objectMD.replicationInfo.storageType &&
-        Array.isArray(objectMD.replicationInfo.backends)) {
+    if (Array.isArray(objectMD?.replicationInfo?.backends)) {
         objectMD.replicationInfo.backends.forEach(backend => {
+            const locationConstraint = config.getLocationConstraint(backend.site);
+            if (!config.includeCrrStatusMetadata &&
```
francoisferrand: do we need this configurable?

- With CRR, it does not really make sense to report the "replicated version id", since it is the same: not explicit, but probably handled already by the conditions below?
- If we set it to false, we keep the current behavior: which is not acceptable. So we should check with s3c team (and PO) if reporting the status would be acceptable...

welansari: > With CRR, it does not really make sense to report the "replicated version id", since it is the same: not explicit, but probably handled already by the conditions below?

It's not explicit but it is handled by the condition below, the dataStoreVersionID field is empty in CRR

> If we set it to false, we keep the current behavior: which is not acceptable. So we should check with s3c team (and PO) if reporting the status would be acceptable...

I wanted to keep the same behaviour for S3C that currently doesn't display the per site status. But yes we should check with the S3C team.

ghost (approved):

PR #5929, opened 2025-09-03
CLDSRV-740: Add healthcheck internal route s3 port

BourgoisMickael:
Bring back the healthcheck route like it was in development/7.70

needed by S3C s3-frontend

### DIFF
--- a/.github/docker/config.s3c.json
+++ b/.github/docker/config.s3c.json
@@ -22,7 +22,8 @@
     "clusters": 5,
     "kmsHideScalityArn": true,
     "healthChecks": {
-        "allowFrom": ["127.0.0.1", "::1"]
+        "allowFrom": ["127.0.0.1", "::1"],
+        "enableInternalRoute": true
     },
     "localCache": {
         "host": "localhost",
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -532,52 +532,35 @@ jobs:
     - name: Setup vault credentials like S3C Integration
       run: ./setup-s3c.sh
       working-directory: .github/docker
+    - name: Set config files env variables
+      run: |-
+        echo "S3_CONFIG_FILE=${{ github.workspace }}/.github/docker/config.s3c.json" >> $GITHUB_ENV
+        echo "S3_LOCATION_FILE=${{ github.workspace }}/tests/locationConfig/locationConfigS3C.json" >> $GITHUB_ENV
     - name: Run cloudserver-object tests
-      env:
-        S3_CONFIG_FILE: ${{ github.workspace }}/.github/docker/config.s3c.json
-        S3_LOCATION_FILE: ${{ github.workspace }}/tests/locationConfig/locationConfigS3C.json
       run: |-
         set -o pipefail;
         yarn run ft_awssdk_objects_misc | tee /tmp/artifacts/${{ matrix.job-name }}/ft_awssdk_objects_misc.log
     - name: Run cloudserver-version tests
-      env:
-        S3_CONFIG_FILE: ${{ github.workspace }}/.github/docker/config.s3c.json
-        S3_LOCATION_FILE: ${{ github.workspace }}/tests/locationConfig/locationConfigS3C.json
       run: |-
         set -o pipefail;
         yarn run ft_awssdk_versioning | tee /tmp/artifacts/${{ matrix.job-name }}/ft_awssdk_versioning.log
     - name: Run cloudserver-bucket tests
-      env:
-        S3_CONFIG_FILE: ${{ github.workspace }}/.github/docker/config.s3c.json
-        S3_LOCATION_FILE: ${{ github.workspace }}/tests/locationConfig/locationConfigS3C.json
       run: |-
         set -o pipefail;
         yarn run ft_awssdk_buckets | tee /tmp/artifacts/${{ matrix.job-name }}/ft_awssdk_buckets.log
     - name: Run cloudserver-routes (metadata) tests
-      env:
-        S3_CONFIG_FILE: ${{ github.workspace }}/.github/docker/config.s3c.json
-        S3_LOCATION_FILE: ${{ github.workspace }}/tests/locationConfig/locationConfigS3C.json
       run: |-
         set -o pipefail;
         yarn run ft_node_routes | tee /tmp/artifacts/${{ matrix.job-name }}/ft_node_routes.log
     - name: Run backbeat route tests
-      env:
-        S3_CONFIG_FILE: ${{ github.workspace }}/.github/docker/config.s3c.json
-        S3_LOCATION_FILE: ${{ github.workspace }}/tests/locationConfig/locationConfigS3C.json
       run: |-
         set -o pipefail;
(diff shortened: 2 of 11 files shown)


tcarmet (approved):

lib/utilities/healthcheckHandler.js
```diff
@@ -121,7 +121,8 @@ function healthcheckHandler(clientIP, req, res, log, statsClient) {
         });
     }
 
-    const deep = (req.url === '/ready');
+    // CLDSRV-740: S3C with the healthcheck on s3 port needs the internal route
```
francoisferrand: - Can we upgrade `s3frontend` to use the new route instead, to avoid introducing config flag and divergence? Seems the preferable approach here...
- If I remember correctly, we already have an option used to "enable" healthcheck on dedicated port (→ `/ready`, `/health`) or on the usual route (`/_/healthcheck`) -maybe something like "healthcheckPort"-. If we need to have separate behaviors, could we use the same setting?

francoisferrand: the flags are `metricsPort` and `metricsListenOn` I think

lib/utilities/healthcheckHandler.js
```diff
@@ -121,7 +121,8 @@ function healthcheckHandler(clientIP, req, res, log, statsClient) {
...
+    const deep = (req.url === '/ready') || (_config.healthcheckS3Internal && req.url === '/_/healthcheck/deep');
```
francoisferrand: I don't think this can actually happen: `healthchandler` is called only from `routeAdminRequest`, depending on the url:

```
        switch (req.url) {
            case '/live':
            case '/ready':
                healthcheckHandler(clientInfo.clientIP, req, res, log, statsClient);
                break;
```

tests/functional/healthchecks/test/checkRoutes.js
```diff
@@ -8,10 +8,11 @@ const async = require('async');
 const Redis = require('ioredis');
 
 const conf = require('../../config.json');
+const { config } = require('../../../../lib/Config');
 
 const redis = new Redis({
-    host: conf.localCache.host,
-    port: conf.localCache.port,
+    host: config.localCache.host || conf.localCache.host,
+    port: config.localCache.port || conf.localCache.port,
```
francoisferrand: that is a good catch, but we should `only` read from config?


```suggestion
    host: config.localCache.host
    port: config.localCache.port
```

francoisferrand: or maybe this was by design, to ensure the tests run with the right config.json file (and thus fully reproductible) : and not sure `Config` will pick the right config file?

lib/utilities/healthcheckHandler.js
BourgoisMickael: We don't want to change the `metricsPort`. For now we use `metricsPort` only to scrape metrics. And healthcheck is on the s3 port.

> Can we upgrade s3frontend to use the new route instead,

No because with the same port, internal services are prefixed with `/_/<internalServiceName` otherwise it's a regular S3 request. But We could use a name like `/_/healthcheck/ready`

.github/workflows/tests.yaml
```diff
@@ -574,6 +574,13 @@ jobs:
       run: |-
         set -o pipefail;
         yarn run ft_backbeat | tee /tmp/artifacts/${{ matrix.job-name }}/ft_backbeat.log
+    - name: Run healthchecks tests
+      env:
+        S3_CONFIG_FILE: ${{ github.workspace }}/.github/docker/config.s3c.json
+        S3_LOCATION_FILE: ${{ github.workspace }}/tests/locationConfig/locationConfigS3C.json
```
francoisferrand: relying on gha to set the parameters is not a good pattern, it will only make tests harder to run locally

lib/utilities/healthcheckHandler.js
francoisferrand: ok got this, it is on next chunk, with 2 different ways to call the function.
In that context:
- Either (and preferably) we can keep the same route on both cases, update s3frontend to use the /live and /ready path, and avoid the new config option
- Or we should make `deep` a parameter of the healtcheckHandler function, set appropriately (and differently) by both `internalHandlers` and `routeAdminRequest`, with no need for an option

tests/functional/healthchecks/test/checkRoutes.js
BourgoisMickael: It picks the default config location that shoud be the one from the root of the repo which has the same value.

It's for Integration mainly that it's interesting to rely on the lib/Config rather than a hardcoded test config as we might provide other values

.github/workflows/tests.yaml
BourgoisMickael: but that's how we do in this CI, we use env to define many parameters.

I could replace the many env in every steps with 1 setp that sets the env only once via `>> $GITHUB_ENV`

tests/functional/healthchecks/test/checkRoutes.js
```diff
@@ -8,10 +8,11 @@ const async = require('async');
 const Redis = require('ioredis');
 
 const conf = require('../../config.json');
+const { config } = require('../../../../lib/Config');
 
 const redis = new Redis({
-    host: conf.localCache.host,
-    port: conf.localCache.port,
+    host: config.localCache.host
+    port: config.localCache.port
```
BourgoisMickael:
```suggestion
    host: config.localCache.host,
    port: config.localCache.port,
```

.github/workflows/tests.yaml
BourgoisMickael: simplified here https://github.com/scality/cloudserver/pull/5929/commits/d5b643ce5759878826c6b05bd4c88ec321a9fccb

tmacro (approved):

.github/docker/config.s3c.json
```diff
@@ -22,7 +22,8 @@
     "clusters": 5,
     "kmsHideScalityArn": true,
     "healthChecks": {
-        "allowFrom": ["127.0.0.1", "::1"]
+        "allowFrom": ["127.0.0.1", "::1"],
+        "s3Internal": true
```
francoisferrand: nit: something like `enableInternalRoute` may be better?

lib/utilities/internalHandlers.js
```diff
@@ -13,6 +15,14 @@ const internalHandlers = {
     veeam: routeVeeam,
 };
 
+if (config.healthChecks && config.healthChecks.s3Internal) {
```
francoisferrand:
```suggestion
if (config.healthChecks?.s3Internal) {
```

lib/Config.js
```diff
@@ -1548,6 +1548,11 @@ class Config extends EventEmitter {
             this.healthChecks.allowFrom = defaultHealthChecks.allowFrom
                 .concat(config.healthChecks.allowFrom);
         }
+        /**
+         * CLDSRV-740: S3C with nginx s3-frontend needs the healthcheck on the
+         * same port as s3 to allow nginx upstream failover based on healthcheck
+         */
+        this.healthChecks.s3Internal = (config.healthChecks && config.healthChecks.s3Internal) || false;
```
```suggestion
        this.healthChecks.s3Internal = config.healthChecks?.s3Internal || false;
```

francoisferrand (approved):

PR #5935, opened 2025-09-05
CLDSRV-741: Bypass PutObj max size 5GB S3C-10336

BourgoisMickael:
We don't want to disable by default.
But in case a customer has problems with this new limit in S3C,
we want a way to easily disable the limit to give time to customer
to migrate its applications to MPU

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -1742,6 +1742,13 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+
+        /**
+         * S3C-10336: PutObject max size of 5GB is new in 9.5.1
+         * Provides a way to bypass the new validation if it breaks customer workflows
+         */
+        this.bypassMaxPutObjectSize = process.env.BYPASS_MAX_PUT_OBJECT_SIZE === 'true'
+            || config.bypassMaxPutObjectSize || false;
         return config;
     }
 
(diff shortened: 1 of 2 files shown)


jonathan-gramain (approved):

lib/api/objectPut.js
```diff
@@ -92,7 +93,9 @@ function objectPut(authInfo, request, streamingV4Params, log, callback) {
     if (Number.parseInt(size, 10) > constants.maximumAllowedUploadSize) {
         log.debug('Upload size exceeds maximum allowed for a single PUT',
             { size });
-        return callback(errors.EntityTooLarge);
+        if (!config.bypassMaxPutObjectSize) {
+            return callback(errors.EntityTooLarge);
+        }
```
ghost: I would add the debug log in the `if` to be consistent?

ghost (approved):

lib/Config.js
```diff
@@ -1742,6 +1742,13 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+
+        /**
+         * S3C-10336: PutObject max size of 5GB is new in 9.5.1
+         * Provides a way to bypass the new validation if it breaks customer workflows
+         */
```
DarkIsDude, line 1749:
```suggestion
```

DarkIsDude (approved):

lib/api/objectPut.js
BourgoisMickael: I left the debug out so we can easily see if objects exceeds size

BourgoisMickael: but I think it could be done by checking the bytesReceived directly as well

lib/Config.js
BourgoisMickael: this is documentation for the flag

PR #5939, opened 2025-09-10
Lowercase extractClientIPFromHeader & extractProtocolFromHeader

ghost:
This will reduce unnecessary computations in arsenal. See https://github.com/scality/Arsenal/pull/2502.

Issue: CLDSRV-743.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -507,6 +507,15 @@ function requestsConfigAssert(requestsConfig) {
             'extractProtocolFromHeader must be set if viaProxy is ' +
             'set to true and must be a string');
         }
+
+        // All headers in NodeJS are lowercase: to be exploitable
+        // we need to lowercase the value.
+        // eslint-disable-next-line no-param-reassign
+        requestsConfig.extractClientIPFromHeader =
+            requestsConfig.extractClientIPFromHeader?.toLowerCase();
+        // eslint-disable-next-line no-param-reassign
+        requestsConfig.extractProtocolFromHeader =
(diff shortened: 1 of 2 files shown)


welansari (approved):

lib/Config.js
```diff
@@ -506,6 +506,15 @@ function requestsConfigAssert(requestsConfig) {
...
             'set to true and must be a string');
+
+            // All headers in NodeJS are lowercase: to be exploitable
+            // we need to lowercase the value.
+            // eslint-disable-next-line no-param-reassign
+            requestsConfig.extractClientIPFromHeader =
+                requestsConfig.extractClientIPFromHeader.toLowerCase();
+            // eslint-disable-next-line no-param-reassign
+            requestsConfig.extractProtocolFromHeader =
+                requestsConfig.extractProtocolFromHeader.toLowerCase();
```
BourgoisMickael: They are not always present

```suggestion
            requestsConfig.extractClientIPFromHeader =
                requestsConfig.extractClientIPFromHeader?.toLowerCase();
            // eslint-disable-next-line no-param-reassign
            requestsConfig.extractProtocolFromHeader =
                requestsConfig.extractProtocolFromHeader?.toLowerCase();
```

BourgoisMickael: maybe this should be out of condition `viaProxy` as well

ghost: Done. This is actually not a valid case but as there is no strict enforcement, better to move it out to be future proof indeed.

BourgoisMickael (approved):

PR #5941, opened 2025-09-12
CLDSRV-744: add missing md5 checksum checks tests

leif-scality:
[CLDSRV-744](https://scality.atlassian.net/browse/CLDSRV-744): add missing md5 checksum checks

`validateChecksumsNoChunking` returns a `MissingChecksum` error because some S3 methods (e.g. [DeleteObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html)) may require a checksum.

The [DeleteObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) only mentions that it requires an MD5 but in practice it requires either MD5 or a checksum using one of the new algorithms, the newer SDK's send CRC64NVME only. The requirement will be added after the new algorithms are implemented.

Support for the newer algorithms (CRC-64/NVME, CRC-32, CRC-32C, SHA-1, SHA-256) will be added inside validateChecksumsNoChunking in another PR.

[CLDSRV-744]: https://scality.atlassian.net/browse/CLDSRV-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * validateChecksumsNoChunking - Validate the checksums of a request.
+ * @param {object} headers - http headers
+ * @param {string} body - http request body
+ * @return {string} - error message
+ */
+function validateChecksumsNoChunking(headers, body) {
+    if (headers['content-md5']) {
+        const md5 = crypto.createHash('md5').update(body, 'utf8').digest('base64');
```
BourgoisMickael: https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md#crypto-implement-cryptohash

https://nodejs.org/docs/latest-v24.x/api/crypto.html#cryptohashalgorithm-data-options

The `.hash` function is still marked as experimental but as RC, could be interesting to see if we better perf for small body

> A utility for creating one-shot hash digests of data. It can be faster than the object-based crypto.createHash() when hashing a smaller amount of data (<= 5MB) that's readily available. If the data can be big or if it is streamed, it's still recommended to use crypto.createHash() instead.

leif-scality: I could use `.hash` but given that the impacted endpoints are configuration, low req/sec methods I don't think we will see any difference.

lib/api/bucketPutACL.js
```diff
@@ -44,6 +45,12 @@ const monitoring = require('../utilities/monitoringHandler');
 function bucketPutACL(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'bucketPutACL' });
 
+    const err = validateChecksumsNoChunking(request.headers, request.post);
```
BourgoisMickael: cf: https://scality.atlassian.net/wiki/spaces/OS/pages/3317235725/Checking+object+integrity+in+Amazon+S3?focusedCommentId=3389947938

Is it the right place the validate checksum or should we rather perform the other validations first before doing CPU intensive validation ?

leif-scality: I can move it down, but I'd rather keep on top like in the other methods. I don't see how calculating the checksums could be exploited by an attacker

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -0,0 +1,20 @@
...
+        if (md5 !== headers['content-md5']) {
+            return `md5 mismatch, calculated ${md5} expected ${headers['content-md5']}`;
```
BourgoisMickael: prefer object
```suggestion
            return {
              error: 'md5 mismatch',
              details: {
                calculated: md5,
                expected: headers['content-md5']
              }
            };
```

```
logger.debug(err.error, err.details)
```

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -0,0 +1,27 @@
+const crypto = require('crypto');
+
+const ChecksumError = Object.freeze({
+    MD5Mismatch: 'md5 mismatch',
+    MissingChecksum: 'missing checksum',
```
BourgoisMickael:
```suggestion
    MD5Mismatch: 'MD5Mismatch',
    MissingChecksum: 'MissingChecksum',
```

I'd prefer the same string for key and value

lib/api/bucketPutACL.js
```diff
@@ -44,6 +45,12 @@ const monitoring = require('../utilities/monitoringHandler');
...
+    if (err && err.error !== ChecksumError.MissingChecksum) {
+        log.debug(err, { error: errors.BadDigest });
```
BourgoisMickael: I think the debug can be simplified into
```suggestion
        log.debug(err.error, err.details);
```

as the BadDigest will show in the response already

leif-scality: Does the BadDigest get logged after we exit the handler? If not how do will we know that we sent a BadRequest to the client?

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: ok

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/bucketPutACL.js
BourgoisMickael: I'm not sure, but there might be some debug logs to print the error sent to client

leif-scality: Before my changes we logged the BadDigest so I rather keep it

tcarmet (approved):

lib/api/objectPutRetention.js
```diff
@@ -26,6 +27,12 @@ const REPLICATION_ACTION = 'PUT_RETENTION';
 function objectPutRetention(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'objectPutRetention' });
 
+    const err = validateChecksumsNoChunking(request.headers, request.post);
+    if (err && err.error !== ChecksumError.MissingChecksum) {
+        log.debug(err, { error: errors.BadDigest });
```
ghost: I'm not sure we need to log the error that we return, better to have a string like this?
```suggestion
        log.debug('failed checksum validation', err);
```

Same for all other places

lib/api/multiObjectDelete.js
```diff
@@ -475,14 +475,10 @@ function multiObjectDelete(authInfo, request, log, callback) {
...
-            .update(request.post, 'utf8').digest('base64');
-        if (md5 !== request.headers['content-md5']) {
-            monitoring.promMetrics('DELETE', request.bucketName, 400,
-                'multiObjectDelete');
-            return callback(errors.BadDigest);
-        }
+    const err = validateChecksumsNoChunking(request.headers, request.post);
+    if (err && err.error !== ChecksumError.MissingChecksum) {
+        log.debug(err, { error: errors.BadDigest });
+        return callback(errors.BadDigest);
```
ghost: If we always have the same pattern, it would be better to just return the arsenal error in the function, or nothing, and if there is an error, then we call the callback.
The debug log can be moved there too. Should probably be an error one as it makes the whole API fail?

lib/api/apiUtils/integrity/validateChecksums.js
ghost: The `PutObjectRetention` is today (at least) called thousands of times per second with some solutions like Veeam. Even if this reduces later, this will remain high. I've seen platforms with 6,000 op/s on a 3 server architecture. The CPU impact could be significant, I suggest we properly test it during RC phase.

lib/api/bucketPutCors.js
```diff
@@ -33,14 +33,10 @@ function bucketPutCors(authInfo, request, log, callback) {
...
-    if (request.headers['content-md5']) {
-        const md5 = crypto.createHash('md5')
-            .update(request.post, 'utf8').digest('base64');
-        if (md5 !== request.headers['content-md5']) {
-            log.debug('bad md5 digest', { error: errors.BadDigest });
-            monitoring.promMetrics('PUT', bucketName, 400, 'putBucketCors');
-            return callback(errors.BadDigest);
-        }
+    const err = validateChecksumsNoChunking(request.headers, request.post);
+    if (err && err.error !== ChecksumError.MissingChecksum) {
```
williamlardier, line 37: If I understand correctly, if the error is `MissingChecksum`, ie the client sent a body but not the content-md5 header, we just ignore any validation?

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {object} headers - http headers
+ * @param {string} body - http request body
+ * @return {string} - error message
+ */
+function validateChecksumsNoChunking(headers, body) {
+    if (headers && headers['content-md5']) {
+        const md5 = crypto.createHash('md5').update(body, 'utf8').digest('base64');
+        if (md5 !== headers['content-md5']) {
+            return { error: ChecksumError.MD5Mismatch, details: { calculated: md5, expected: headers['content-md5'] } };
+        }
```
ghost, line 21: We currently do nothing of this object, when we handle errors. Should we just return the error and add a log in this function?

lib/api/multiObjectDelete.js
ghost: Also, according to the doc:
> Directory bucket - The Content-MD5 request header or a additional checksum request header (including x-amz-checksum-crc32, x-amz-checksum-crc32c, x-amz-checksum-sha1, or x-amz-checksum-sha256) is required for all Multi-Object Delete requests.

In the general case I think it's fine to ignore the `MissingChecksum` error, but here, I believe we want to enforce it, so return an error if the header is not set?

ghost: I see we have
```
    it('should accept request when content-md5 header is missing', done => {
```

in the test, so I guess that was expected, but that looks not standard. Given your implementation it means there is just no need to return the `MissingChecksum` error, we can just return nothing, if it's fine to not validate anything if the header is not present. That should simplify the processing in the APIs...

leif-scality: We cannot require the checksum header yet because we don't support the other algorithms today, so this endpoint would break with the newer SDK's if we require it, after we implement the newer algorithms we can return an error on `MissingChecksum`

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: I think the caller should do the logging.

In the next iteration I will add `SHA256Mismatch`, `CRC32Mismatch`, etc. I don't see why we would want to squash all the possible errors into `BadDigest`

lib/api/objectPutRetention.js
leif-scality: ok

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: > I suggest we properly test it during RC phase

What is the procedure to track this?

leif-scality: If the CPU cost is to elevated the only alternative I think is to disable it, but AWS does the check and looking at the body of the request it is fairly small so it should be ok 

https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectRetention.html

lib/api/objectPutRetention.js
leif-scality: done

lib/api/bucketPutACL.js
jonathan-gramain: There's another compelling reason to do the checksum earlier: we could avoid transcoding the string back to a `Buffer` just for the purpose of checksumming. I don't know for sure but I suspect the cost in CPU and memory of converting to a Buffer is not negligible, even compared to the cost of checksumming itself.

I am thinking we could precompute the MD5 checksum at the same time we aggregate the buffers, i.e. around https://github.com/scality/cloudserver/blob/d8b9b99f12beaa50bd245b76837ee7ae789399d9/lib/api/api.js#L252. But it would have to adapt to the hash algorithm(s) selected, I don't know for sure if we have that info at this point.

Another benefit could be that we could do generic validation against the headers for all request types, to reduce code duplication.

lib/api/objectPutRetention.js
```diff
@@ -26,6 +27,15 @@ const REPLICATION_ACTION = 'PUT_RETENTION';
 function objectPutRetention(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'objectPutRetention' });
 
+    if (config.integrityChecks.PutObjectRetention)
+    {
```
jonathan-gramain:
```suggestion
    if (config.integrityChecks.PutObjectRetention) {
```

leif-scality: Where does the String -> Buffer transcoding occur? [request.post is a string](https://github.com/scality/cloudserver/blob/development/9.0/lib/api/api.js#L246) and it is never converted to a Buffer, at least not for calculating checksums.

leif-scality: Some requests handle checksums differently so we cannot do a generic global check.

lib/api/bucketPutACL.js
leif-scality: removed the BadDigest

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: Added a variable in the config to disable the check in PutObjectRetention if needed

lib/api/objectPutRetention.js
leif-scality: Where are we transcoding the String back to a Buffer? The crypto hash.update takes a string and does no transcoding.

Not all endpoints handle the checksum in the same way so I think we should handle them inside each endpoint and not in api.js

lib/api/bucketPutACL.js
leif-scality: Where are we transcoding the String back to a Buffer? The crypto hash.update takes a string and does no transcoding.

Not all endpoints handle the checksum in the same way so I think we should handle them inside each endpoint and not in api.js

lib/api/objectPutRetention.js
jonathan-gramain: > The crypto hash.update takes a string and does no transcoding.

Are you sure of this? I assumed that it was always necessary to convert to a Buffer internally since hashes operate on bytes, not characters (however optimizations are likely possible in theory if the encoding used for hashing is the same as the native storage format of JS strings, but I'm not sure this is used in practice and internal formats may not be `utf-8` anyways, e.g. Python uses 2 bytes per character). I also asked Gemini, it seems to confirm my assumption, here's the first part of its answer:

> When passing data to the Node.js crypto module's hash.update() function, using a raw Buffer is generally faster and more performant than using a string. 🚀 This performance difference stems from how Node.js handles these two data types.
> 
> Key Performance Differences
> String Encoding Overhead: When you pass a string to hash.update(), Node.js must first convert that string into a sequence of bytes, or a Buffer. This conversion process requires additional CPU cycles and memory allocation to handle character encodings (like UTF-8, which is the default). For every string you pass, this encoding operation adds a small, but cumulatively significant, overhead.
> 
> Direct Binary Data: A Buffer, on the other hand, is already a direct representation of raw binary data. It's essentially a block of pre-allocated memory outside the V8 JavaScript engine's heap. This means when you pass a Buffer to hash.update(), the function can directly access the underlying bytes without any conversion step. It's already in the format the cryptographic hash function needs, leading to more efficient processing.
> [...]

jonathan-gramain: > Not all endpoints handle the checksum in the same way so I think we should handle them inside each endpoint and not in api.js

Do you know what are the differences in practice? I don't know about those, but maybe we could detect in some way the exceptions in the generic place? Just trying to see if we can do better, if it's indeed a pain to do it generically, no problem for me with doing it per call (but the performance point above may still apply).

lib/api/multiObjectDelete.js
ghost: Can we add a ticket number as TODO so we do not forget anything when this will be done later? As this conflicts with that is in the PR message, that will help reviewers understand the context till this is fixed.

lib/api/objectPutRetention.js
leif-scality: I added a commit that moves the check to `api.js` before the body gets transformed into a string.

Today the only non standard endpoint is DeleteObjects, it is the only method that requires a checksum (the requirement wont be added in this PR because we are missing the other algos).

After searching NodeJS does indeed decode into a new buffer [here](https://github.com/nodejs/node/blob/f6f8eb7c25ed770f613bc435b48a10e15bb9d24c/src/crypto/crypto_util.h#L64C5-L64C31), I thought it did it on the fly.

lib/Config.js
```diff
@@ -533,6 +533,21 @@ function bucketNotifAssert(bucketNotifConfig) {
     return bucketNotifConfig;
 }
 
+function parseIntegrityChecks(config) {
+    const integrityChecks = {
+        PutObjectRetention: true,
+    };
+
+    if (config.integrityChecks) {
+        if ('PutObjectRetention' in config.integrityChecks) {
```
jonathan-gramain: I'm thinking that it could be even simpler to configure a map of integrity checks to disable for all possible calls, and provide this possibility for other calls than `putObjectRetention` in case we need it. We would check the presence of a custom configuration with `config.integrityChecks[apiCall]` instead in the generic handler, and enable the checks by default.

jonathan-gramain (commented):
the POC LGTM, I'll review the final PR when completed, thanks for doing the change!

lib/Config.js
leif-scality: done

lib/api/multiObjectDelete.js
leif-scality: What is the conflict with the PR message?

I created this ticket [CLDSRV-747](https://scality.atlassian.net/browse/CLDSRV-747) to enable the requirement when we have the other checksums

lib/api/bucketPutNotification.js
```diff
@@ -1,5 +1,4 @@
 const async = require('async');
-
```
jonathan-gramain, line 2: can you cleanup the added/removed lines that were left over after the latest refactor?

tests/unit/api/api.js
```diff
@@ -112,4 +114,101 @@ describe('api.callApiMethod', () => {
...
+                    query: {},
+                    socket: { remoteAddress: '127.0.0.1', destroy: sandbox.stub() }
+                }, body);
+                
+                sandbox.stub(api, method).callsFake((userInfo, _request, log, cb) => {
+                    cb();
+                });
+
+                api.callApiMethod(method, requestWithBody, response, log, err => {
+                    assert(!err, `Unexpected error for ${method} with good checksum: ${err}`);
```
jonathan-gramain:
```suggestion
                    assert.ifError(err, `Unexpected error for ${method} with good checksum: ${err}`);
```

jonathan-gramain (approved):

tests/unit/api/api.js
leif-scality: done

lib/api/bucketPutNotification.js
leif-scality: done

leif-scality:
ping

leif-scality:
ping

leif-scality:
ping

PR #5942, opened 2025-09-12
CLDSRV-741: Override Key limit for S3C-10370

BourgoisMickael:
Key length will only be check in PutObject, CopyObject, initMPU.

To enforce KeyTooLong only when creating new objects. If objects already exists with key too long, they can still be reached.

See related PR https://github.com/scality/Arsenal/pull/2510

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -22,6 +22,7 @@ const {
 const { ValidLifecycleRules: supportedLifecycleRules } = require('arsenal').models;
 const { utapiVersion } = require('utapi');
 const { scaleMsPerDay } = s3middleware.objectUtils;
+const { constants: arsenalConstants } = require('arsenal');
 
 const constants = require('../constants');
 const {
@@ -1782,6 +1783,20 @@ class Config extends EventEmitter {
          */
         this.bypassMaxPutObjectSize = process.env.BYPASS_MAX_PUT_OBJECT_SIZE === 'true'
             || config.bypassMaxPutObjectSize || false;
+
+        /**
+         * S3C-10370: Before 9.5.1, there was no limit on the key length.
+         * The current limit is 915, provide a way to ignore (set to 0) or override the limit
+         */
+        this.objectKeyByteLimit = arsenalConstants.objectKeyByteLimit;
+        const overrideObjectKeyByteLimit =
+            process.env.OVERRIDE_OBJECT_KEY_BYTE_LIMIT || config.overrideObjectKeyByteLimit;
+        if (overrideObjectKeyByteLimit !== null && overrideObjectKeyByteLimit !== undefined) {
+            this.objectKeyByteLimit = parseInt(overrideObjectKeyByteLimit, 10);
+            assert(Number.isInteger(this.objectKeyByteLimit) && this.objectKeyByteLimit >= 0,
+                'bad config: overrideObjectKeyByteLimit must be a positive integer');
+        }
+
         return config;
     }
 
--- a/tests/functional/aws-node-sdk/test/object/get.js
+++ b/tests/functional/aws-node-sdk/test/object/get.js
@@ -227,6 +227,16 @@ describe('GET object', () => {
                 });
             });
 
+        it('should return NoSuchKey error when no such object even with key longer than 915 bytes',
+            done => {
+                s3.getObject({ Bucket: bucketName, Key: 'a'.repeat(2000) }, err => {
+                    assert.notEqual(err, null,
+                        'Expected failure but got success');
+                    assert.strictEqual(err.code, 'NoSuchKey');
+                    return done();
+                });
+            });
+
         describe('Additional headers: [Cache-Control, Content-Disposition, ' +
             'Content-Encoding, Expires, Accept-Ranges]', () => {
             describe('if specified in put object request', () => {
--- a/tests/functional/aws-node-sdk/test/object/get.js
+++ b/tests/functional/aws-node-sdk/test/object/get.js
@@ -210,9 +210,7 @@ describe('GET object', () => {
         'bucket name',
             done => {
                 s3.getObject({ Bucket: '', Key: 'somekey' }, err => {
-                    assert.notEqual(err, null,
-                        'Expected failure but got success');
-                    assert.strictEqual(err.code, 'MethodNotAllowed');
+                    checkError(err, 'MethodNotAllowed');
                     return done();
                 });
             });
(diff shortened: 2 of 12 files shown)


lib/Config.js
```diff
@@ -1749,6 +1749,17 @@ class Config extends EventEmitter {
          */
         this.bypassMaxPutObjectSize = process.env.BYPASS_MAX_PUT_OBJECT_SIZE === 'true'
             || config.bypassMaxPutObjectSize || false;
+
+        /**
+         * S3C-10370: Before 9.5.1, there was no limit on the key length.
+         * The current limit is 915, provide a way to ignore (set to 0) or override the limit
+         */
+        this.overrideKeyLimit = +(process.env.OVERRIDE_KEY_LIMIT || config.overrideKeyLimit) || null;
```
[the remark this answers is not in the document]
BourgoisMickael: no, null won't be coerces, the unary applies only to parenthesis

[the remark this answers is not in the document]
jonathan-gramain: I am also puzzled by the logic, how is it possible to set to 0 like the comment suggests?

Also generally speaking, it can be better style to split more complex logic into multiple `if` lines, rather than rely on a series of boolean operators, or rely too much on javascriptisms.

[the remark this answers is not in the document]
BourgoisMickael: You'd prefer something like ?
```suggestion
        this.overrideKeyLimit = process.env.OVERRIDE_KEY_LIMIT || config.overrideKeyLimit || null;
        if (this.overrideKeyLimit) {
            this.overrideKeyLimit = parseInt(this.overrideKeyLimit, 10);
        }
```

lib/Config.js
```diff
@@ -1749,6 +1749,18 @@ class Config extends EventEmitter {
...
+        /**
+         * S3C-10370: Before 9.5.1, there was no limit on the key length.
+         * The current limit is 915, provide a way to ignore (set to 0) or override the limit
+         */
+        this.overrideKeyLimit = process.env.OVERRIDE_KEY_LIMIT || config.overrideKeyLimit || null;
+        if (this.overrideKeyLimit !== null) {
+            this.overrideKeyLimit = parseInt(this.overrideKeyLimit, 10);
+            assert(Number.isInteger(this.overrideKeyLimit) && this.overrideKeyLimit >= 0,
+                'bad config: overrideKeyLimit must be a positive integer');
+        }
```
jonathan-gramain: For clarity, instead of overwriting the same attribute, it's best to use a const variable:

```suggestion
        const overrideKeyLimitStr = process.env.OVERRIDE_KEY_LIMIT || config.overrideKeyLimit || null;
        if (overrideKeyLimitStr !== null) {
            this.overrideKeyLimit = parseInt(overrideKeyLimitStr, 10);
            assert(Number.isInteger(this.overrideKeyLimit) && this.overrideKeyLimit >= 0,
                'bad config: overrideKeyLimit must be a positive integer');
        } else {
            this.overrideKeyLimit = null;
        }
```

tests/functional/aws-node-sdk/test/object/get.js
```diff
@@ -227,6 +227,16 @@ describe('GET object', () => {
                 });
             });
 
+        it('should return NoSuchKey error when no such object even with key longer than 915 bytes',
+            done => {
+                s3.getObject({ Bucket: bucketName, Key: 'a'.repeat(2000) }, err => {
+                    assert.notEqual(err, null,
+                        'Expected failure but got success');
```
jonathan-gramain: nitpick: I don't think the message is useful as is, maybe we could replace it by a more contextualized sentence about what was expected from the test (or removing the message completely is fine too).

jonathan-gramain (approved):

tests/functional/aws-node-sdk/test/object/get.js
BourgoisMickael: It matches the same pattern as the 2 tests above

lib/Config.js
BourgoisMickael: There [61178cc](https://github.com/scality/cloudserver/pull/5942/commits/61178ccffb6fe678970c5428a69171bfbdeccc15)

lib/Config.js
```diff
@@ -1749,6 +1749,19 @@ class Config extends EventEmitter {
          */
         this.bypassMaxPutObjectSize = process.env.BYPASS_MAX_PUT_OBJECT_SIZE === 'true'
             || config.bypassMaxPutObjectSize || false;
+
+        /**
+         * S3C-10370: Before 9.5.1, there was no limit on the key length.
+         * The current limit is 915, provide a way to ignore (set to 0) or override the limit
+         */
+        this.overrideKeyLimit = null;
```
francoisferrand: should this not be `0`, to be consistent with the value set below?

tests/functional/aws-node-sdk/test/object/get.js
francoisferrand: just `assert.notEqual(err, null)` is enough IMHO

tests/functional/aws-node-sdk/test/object/initiateMPU.js
```diff
@@ -52,6 +52,14 @@ describe('Initiate MPU', () => {
              })
         );
 
+        it('should return KeyTooLong error when key is longer than 915 bytes', done =>
+            s3.createMultipartUpload({ Bucket: bucket, Key: 'a'.repeat(916) }, err => {
+                assert.strictEqual(err.code, 'KeyTooLong');
```
francoisferrand, line 58: check not null before, to void accessing field of null :
```suggestion
				assert.notEqual(err, null);
                assert.strictEqual(err.code, 'KeyTooLong');
```

francoisferrand: or may use `checkError`

francoisferrand (approved):

lib/Config.js
BourgoisMickael: no 0 is used to ignore completely the limit

BourgoisMickael: cf: https://github.com/scality/Arsenal/pull/2510/files#diff-267a4c9594e1bbf54c8870e64da137edce475d5604c950f99ca05f55f0717165R1187

BourgoisMickael: if the value is null and not set, then the default 915 is used

francoisferrand: should we not explicitly set to `915` here by default, to make this explicit and reduce coupling/improve locality?

BourgoisMickael: It's already hardcoded in arsenal.

But I can put it in cloudserver as well.

Should we then remove the constant from arsenal and make it mandatory in cloudserver ?

tests/functional/aws-node-sdk/test/object/get.js
BourgoisMickael: [ec6b430](https://github.com/scality/cloudserver/pull/5942/commits/ec6b4305ec9e57da2ce25241c6298fffa916a370) now its consistent with the rest of the file

lib/Config.js
BourgoisMickael: Renamed the field to objectKeyByteLimit to add the default arsenal constant

PR #5950, opened 2025-09-24
Improvement/cldsrv 724 sur utapi tests

benzekrimaha:
Issue: CLDSRV-724

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,11 +19,12 @@
   },
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.908.0",
+    "@aws-sdk/credential-providers": "^3.864.0",
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
     "arsenal": "git+https://github.com/scality/Arsenal#8.2.43",
     "async": "2.6.4",
-    "aws-sdk": "^2.1692.0",
     "bucketclient": "scality/bucketclient#8.2.7",
     "bufferutil": "^4.0.8",
     "commander": "^12.1.0",
--- a/tests/functional/aws-node-sdk/lib/utility/bucket-util.js
+++ b/tests/functional/aws-node-sdk/lib/utility/bucket-util.js
@@ -1,27 +1,40 @@
-const AWS = require('aws-sdk');
-AWS.config.logger = console;
-const { S3 } = require('aws-sdk');
+const {
+    S3Client,
+    HeadBucketCommand,
+    CreateBucketCommand,
+    DeleteBucketCommand,
+    ListObjectVersionsCommand,
+    DeleteObjectCommand,
+    ListBucketsCommand,
+} = require('@aws-sdk/client-s3');
 const projectFixture = require('../fixtures/project');
 const getConfig = require('../../test/support/config');
 
 class BucketUtility {
-    constructor(profile = 'default', config = {}) {
+    constructor(profile = 'default', config = {}, unauthenticated = false) {
         const s3Config = getConfig(profile, config);
-
-        this.s3 = new S3(s3Config);
-        this.s3.config.setPromisesDependency(Promise);
-        this.s3.config.update({
-            maxRetries: 0,
-        });
+        if (unauthenticated) {
+            this.s3 = new S3Client({
+                ...s3Config,
+                maxAttempts: 0,
+                credentials: { accessKeyId: '', secretAccessKey: '' },
+                forcePathStyle: true,
+                signer: { sign: async request => request },
+             });
+        }
+       else {
+            this.s3 = new S3Client({
+                ...s3Config,
+                maxAttempts: 0,
+             });
+       }
     }
 
     bucketExists(bucketName) {
-        return this.s3
-            .headBucket({ Bucket: bucketName })
-            .promise()
+        return this.s3.send(new HeadBucketCommand({ Bucket: bucketName }))
             .then(() => true)
             .catch(err => {
-                if (err.code === 'NotFound') {
+                if (err.name === 'NotFound') {
                     return false;
                 }
                 throw err;
@@ -29,123 +42,105 @@ class BucketUtility {
     }
 
     createOne(bucketName) {
-        return this.s3
-            .createBucket({ Bucket: bucketName })
-            .promise()
-            .then(() => bucketName);
+        return this.s3.send(new CreateBucketCommand({ Bucket: bucketName }))
+            .then(() => bucketName)
+            .catch(err => {
+                throw err;
+            });
     }
 
     createOneWithLock(bucketName) {
-        return this.s3
-            .createBucket({
-                Bucket: bucketName,
-                ObjectLockEnabledForBucket: true,
-            })
-            .promise()
-            .then(() => bucketName);
+        return this.s3.send(new CreateBucketCommand({
+            Bucket: bucketName,
+            ObjectLockEnabledForBucket: true,
+        })).then(() => bucketName);
     }
 
     createMany(bucketNames) {
         const promises = bucketNames.map(bucketName =>
             this.createOne(bucketName),
         );
-
         return Promise.all(promises);
     }
 
     createRandom(nBuckets = 1) {
         if (nBuckets === 1) {
             const bucketName = projectFixture.generateBucketName();
-
             return this.createOne(bucketName);
         }
-
         const bucketNames = projectFixture
             .generateManyBucketNames(nBuckets)
-            .sort(() => 0.5 - Math.random()); // Simply shuffle array
-
+            .sort(() => 0.5 - Math.random());
         return this.createMany(bucketNames);
     }
 
     deleteOne(bucketName) {
-        return this.s3.deleteBucket({ Bucket: bucketName }).promise();
+        return this.s3.send(new DeleteBucketCommand({ Bucket: bucketName }));
     }
 
     deleteMany(bucketNames) {
         const promises = bucketNames.map(bucketName =>
             this.deleteOne(bucketName),
         );
-
         return Promise.all(promises);
     }
-
+    
     /**
      * Recursively delete all versions of all objects within the bucket
      * @param bucketName
      * @returns {Promise.<T>}
      */
-
-    async empty(bucketName, BypassGovernanceRetention = false) {
+    empty(bucketName, BypassGovernanceRetention = false) {
         const param = {
             Bucket: bucketName,
         };
 
-        const listedObjects = await this.s3.listObjectVersions(param).promise();
-
-        for (const version of listedObjects.Versions) {
-            if (version.Key.endsWith('/')) {
-                continue;
-            }
-
-            await this.s3
-                .deleteObject({
-                    Bucket: bucketName,
-                    Key: version.Key,
-                    VersionId: version.VersionId,
-                    ...(BypassGovernanceRetention && {
-                        BypassGovernanceRetention,
-                    }),
-                })
-                .promise();
-        }
-
-        for (const version of listedObjects.Versions) {
-            if (!version.Key.endsWith('/')) {
-                continue;
-            }
-
-            await this.s3
-                .deleteObject({
-                    Bucket: bucketName,
-                    Key: version.Key,
-                    VersionId: version.VersionId,
-                    ...(BypassGovernanceRetention && {
-                        BypassGovernanceRetention,
-                    }),
-                })
-                .promise();
-        }
-
-        for (const marker of listedObjects.DeleteMarkers) {
-            await this.s3
-                .deleteObject({
-                    Bucket: bucketName,
-                    Key: marker.Key,
-                    VersionId: marker.VersionId,
-                    ...(BypassGovernanceRetention && {
-                        BypassGovernanceRetention,
-                    }),
-                })
-                .promise();
-        }
+        return this.s3.send(new ListObjectVersionsCommand(param))
+            .then(data => Promise.all(
+                (data.Versions || [])
+                .filter(object => !object.Key.endsWith('/'))
+                .map(object =>
+                    this.s3.send(new DeleteObjectCommand({
+                        Bucket: bucketName,
+                        Key: object.Key,
+                        VersionId: object.VersionId,
+                        ...(BypassGovernanceRetention && { BypassGovernanceRetention }),
+                    })).then(() => object)
+                )
+                .concat((data.Versions || [])
+                    .filter(object => object.Key.endsWith('/'))
+                    .map(object =>
+                        this.s3.send(new DeleteObjectCommand({
+                            Bucket: bucketName,
+                            Key: object.Key,
+                            VersionId: object.VersionId,
+                            ...(BypassGovernanceRetention && { BypassGovernanceRetention }),
+                        }))
+                        .then(() => object)
+                    )
+                )
+                .concat((data.DeleteMarkers || [])
+                    .map(object =>
+                    this.s3.send(new DeleteObjectCommand({
+                        Bucket: bucketName,
+                        Key: object.Key,
+                        VersionId: object.VersionId,
+                        ...(BypassGovernanceRetention && { BypassGovernanceRetention }),
+                        }))
+                        .then(() => object)
+                    )
+                )
+            )
+        );
     }
 
     emptyMany(bucketNames) {
-        const promises = bucketNames.map(bucketName => this.empty(bucketName));
-
+        const promises = bucketNames.map(
+            bucketName => this.empty(bucketName)
+        );
         return Promise.all(promises);
     }
-
+    
     emptyIfExists(bucketName) {
         return this.bucketExists(bucketName).then(exists => {
             if (exists) {
@@ -159,15 +154,15 @@ class BucketUtility {
         const promises = bucketNames.map(bucketName =>
             this.emptyIfExists(bucketName),
         );
-
         return Promise.all(promises);
     }
 
     getOwner() {
-        return this.s3
-            .listBuckets()
-            .promise()
-            .then(data => data.Owner);
+        return this.s3.send(new ListBucketsCommand({}))
+            .then(data => data.Owner)
+            .catch(err => {
+                throw err;
+            });
     }
 }
 
--- a/tests/functional/aws-node-sdk/test/support/config.js
+++ b/tests/functional/aws-node-sdk/test/support/config.js
@@ -1,5 +1,6 @@
 const https = require('https');
-const AWS = require('aws-sdk');
+const http = require('http');
+const { NodeHttpHandler } = require('@smithy/node-http-handler');
 
 const { getCredentials } = require('./credentials');
 const { getAwsCredentials } = require('./awsConfig');
@@ -19,19 +20,45 @@ if (ssl && ssl.ca) {
 
 const DEFAULT_GLOBAL_OPTIONS = {
     httpOptions,
-    apiVersions: { s3: '2006-03-01' },
-    signatureCache: false,
-    sslEnabled: ssl !== undefined,
 };
+
 const DEFAULT_MEM_OPTIONS = {
     endpoint: `${transport}://127.0.0.1:8000`,
-    s3ForcePathStyle: true,
+    port: 8000,
+    forcePathStyle: true,
+    region: 'us-east-1', 
+    maxAttempts: 3,
+    requestHandler: new NodeHttpHandler({
+        connectionTimeout: 5000,
+        socketTimeout: 5000,
+        httpAgent: new (ssl ? https : http).Agent({
+            maxSockets: 200,
+            keepAlive: true,
+            keepAliveMsecs: 1000,
+        }),
+    }),
 };
-const DEFAULT_AWS_OPTIONS = {};
+
+const DEFAULT_AWS_OPTIONS = {
+    region: 'us-east-1',
+    maxAttempts: 3,
+    requestHandler: new NodeHttpHandler({
+        connectionTimeout: 5000,
+        socketTimeout: 5000,
+        httpAgent: new https.Agent({
+            maxSockets: 200,
+            keepAlive: true,
+            keepAliveMsecs: 1000,
+        }),
+    }),
+};
 
 function _getMemCredentials(profile) {
     const { accessKeyId, secretAccessKey } = getCredentials(profile);
-    return new AWS.Credentials(accessKeyId, secretAccessKey);
+    return {
+        accessKeyId,
+        secretAccessKey,
+    };
 }
 
 function _getMemConfig(profile, config) {
@@ -58,11 +85,11 @@ function _getAwsConfig(profile, config) {
     return awsConfig;
 }
 
-function getConfig(profile = 'default', config = {}) {
-    const fn = process.env.AWS_ON_AIR && process.env.AWS_ON_AIR === 'true'
-        ? _getAwsConfig : _getMemConfig;
-
-    return fn.apply(this, [profile, config]);
+function getConfig(profile, config) {
+    if (process.env.AWS_ON_AIR) {
+        return _getAwsConfig(profile, config);
+    }
+    return _getMemConfig(profile, config);
 }
 
 module.exports = getConfig;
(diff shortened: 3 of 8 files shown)


tests/sur/quota.js
```diff
@@ -690,35 +792,42 @@ function multiObjectDelete(bucket, keys, size, callback) {
...
-
         it('should decrease the inflights when performing multi object delete', done => {
             const bucket = 'quota-test-bucket10';
             const key = 'quota-test-object';
             const size = 400;
             return async.series([
-                next => createBucket(bucket, false, next),
+                next => {
+                    createBucket(bucket, false, next);
+                },
```
ghost:
```suggestion
                next => createBucket(bucket, false, next);
```

tests/sur/quota.js
```diff
@@ -690,35 +792,42 @@ function multiObjectDelete(bucket, keys, size, callback) {
...
-                }
-                ),
-                next => putObject(bucket, `${key}2`, size, err => {
-                    assert.ifError(err);
-                    return next();
-                }),
+                next => {
+                    putObject(bucket, `${key}1`, size, err => {
+                        assert.ifError(err);
+                        return next();
```
ghost, line 780: not related to your changes but a quick win could be to just call the next with the error, because an assert in a cb makes it harder to debug, and can even hide the error (as we crash)

tests/sur/quota.js
```diff
@@ -339,8 +395,52 @@ function multiObjectDelete(bucket, keys, size, callback) {
...
+                        // eslint-disable-next-line no-param-reassign
+                        delete args.request.headers['x-amz-checksum-sha256'];
+                        // eslint-disable-next-line no-param-reassign
+                        delete args.request.headers['x-amz-sdk-checksum-algorithm'];
+                        // eslint-disable-next-line no-param-reassign
+                        delete args.request.headers['content-md5'];
+                        
+                        // Log what headers we're sending
+                        // eslint-disable-next-line no-console
+                        console.log('Request headers after stripping:', Object.keys(args.request.headers));
```
ghost: We can probably remove this console log

tests/sur/quota.js
```diff
@@ -339,8 +395,52 @@ function multiObjectDelete(bucket, keys, size, callback) {
...
+                requestChecksumCalculation: 'WHEN_REQUIRED',
+                responseChecksumValidation: 'WHEN_REQUIRED',
+                checksumDisabled: true,
+                disableRequestCompression: true,
+                // Force the client to not add automatic headers
+                useGlobalEndpoint: false,
+            });
+            
+            // Add middleware to strip ALL checksum headers
+            s3Client.middlewareStack.add(
```
ghost: Can we maybe add a comment to explain why we need this? And maybe a ticket number if that could be removed, so we just support the standard headers

tests/sur/quota.js
```diff
@@ -196,27 +235,33 @@ function objectMPU(bucket, key, parts, partSize, callback) {
...
+    }))
+        .then(data => {
+            if (!s3Config.isQuotaInflightEnabled()) {
+                mockScuba.incrementBytesForBucket(bucket, -size);
+            }
+            return callback(null, data);
+        })
+        .catch(err => {
+            callback(err);
+        });
```
ghost:
```suggestion
        .catch(err => callback(err));
```

tests/sur/quota.js
```diff
@@ -196,27 +235,33 @@ function objectMPU(bucket, key, parts, partSize, callback) {
...
-            return s3Client.completeMultipartUpload(params, next);
+            return s3Client.send(new CompleteMultipartUploadCommand(params))
+                .then(() => next())
+                .catch(next);
         },
     ], err => {
-        if (!err && !s3Config.isQuotaInflightEnabled()) {
-            mockScuba.incrementBytesForBucket(bucket, parts * partSize);
+        if (!s3Config.isQuotaInflightEnabled()) {
+            mockScuba.incrementBytesForBucket(bucket, -(parts * partSize));
```
ghost: i'm not sure to get why we increment the bytes before the call then we decrement them, is there any issue with the test?

tests/sur/quota.js
```diff
@@ -57,109 +77,136 @@ function putObjectLockConfiguration(bucket, cb) {
...
+    }))
+        .then(data => {
+            if (!s3Config.isQuotaInflightEnabled()) {
+                mockScuba.incrementBytesForBucket(bucket, sourceSize);
+            }
+            return cb(null, data);
+        })
+        .catch(err => {
+            cb(err);
+        });
```
ghost:
```suggestion
        .catch(err => cb(err));
```

ghost (commented):
I just have one comment "blocking", it's the behavior change on one of the quota tests, i'm not sure this is needed

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -1,139 +1,146 @@
...
+                forcePathStyle: true,
+                signer: { sign: async request => request },
+             });
+        }
+       else {
+            this.s3 = new S3Client({
+                ...s3Config,
+                maxAttempts: 0,
+             });
+       }
```
DarkIsDude, line 30:
```suggestion
        if (unauthenticated) {
            s3Config = {
              ...s3Config,
              credentials: { accessKeyId: '', secretAccessKey: '' },
              forcePathStyle: true,
              signer: { sign: async request => request },
            };
        }

            this.s3 = new S3Client({
                ...s3Config,
                maxAttempts: 0,
             });
```

Pretty sure that `maxAttempts: 0,` is working also for unauthenticated client ?

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -1,139 +1,146 @@
...
         return this.createMany(bucketNames);
     }
 
     deleteOne(bucketName) {
-        return this.s3
-            .deleteBucket({ Bucket: bucketName }).promise();
+        return this.s3.send(new DeleteBucketCommand({ Bucket: bucketName }))
+            .catch(err => {
+                throw err;
+            });
```
DarkIsDude:
```suggestion
        return await this.s3.send(new DeleteBucketCommand({ Bucket: bucketName }));
```

One good practice when working with async/await is to always await with return, this will help to preserve the full stack trace (without the await, the stack trace will start at the caller, if you put the await here, the stack trace will display `deleteOne`

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -1,139 +1,146 @@
...
-                                     Key: object.Key,
-                                     VersionId: object.VersionId,
-                                 }).promise()
-                                 .then(() => object)))
-                )
-            );
+                    )
+                    .concat((data.DeleteMarkers || [])
+                        .map(object =>
+                        this.s3.send(new DeleteObjectCommand({
```
DarkIsDude: This indent is weird ?

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -1,139 +1,146 @@
...
+                            }))
+                                .then(() => object)
+                        ))
+            ));
     }
 
     emptyMany(bucketNames) {
-        const promises = bucketNames.map(
-            bucketName => this.empty(bucketName)
+            const promises = bucketNames.map(
```
DarkIsDude: Indent issue ?

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -1,139 +1,146 @@
...
+                bucketName => this.empty(bucketName)
         );
 
         return Promise.all(promises);
     }
-
     emptyIfExists(bucketName) {
```
DarkIsDude, line 144: I think our guidelines ask us to add a new line between function definition ?

tests/functional/aws-node-sdk/test/support/config.js
```diff
@@ -58,11 +85,11 @@ function _getAwsConfig(profile, config) {
     return awsConfig;
 }
 
-function getConfig(profile = 'default', config = {}) {
-    const fn = process.env.AWS_ON_AIR && process.env.AWS_ON_AIR === 'true'
-        ? _getAwsConfig : _getMemConfig;
-
-    return fn.apply(this, [profile, config]);
+function getConfig(profile, config) {
+    if (process.env.AWS_ON_AIR) {
```
DarkIsDude, line 89: Are you sure about that ? This one is essential `process.env.AWS_ON_AIR === 'true' as `AWS_ON_AIR = false` will consider it as `AWS_ON_AIR` in your case. An env var is always a string

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
+                             context.commandName === 'UploadPartCommand' ||
+                             context.commandName === 'CompleteMultipartUploadCommand';
+            
+            if (!needsMD5) {
+                return next(args);
+            }
+
+            const headers = args.request.headers;
+
+            // Remove any checksum headers
```
DarkIsDude:
```suggestion
```

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
+                    delete headers[header];
+                }
+            });
+
+            // Add MD5
+            if (args.request.body) {
+                const bodyContent = Buffer.from(args.request.body);
+                // Create a new hash instance for each request
+                headers['Content-MD5'] = createHash('md5').update(bodyContent).digest('base64');
+            }
```
DarkIsDude:
```suggestion
            if (args.request.body) {
                const bodyContent = Buffer.from(args.request.body);
                headers['Content-MD5'] = createHash('md5').update(bodyContent).digest('base64');
            }
```

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
 const MockUtapi = require('../utilities/mock/Utapi');
 const getConfig = require('../functional/aws-node-sdk/test/support/config');
 const WAIT_MS = 100;
 let s3Client = null;
 
+/**
+ * Creates an S3 client that uses MD5 checksums for DeleteObjects operations
+ * Based on https://github.com/aws/aws-sdk-js-v3/blob/main/supplemental-docs/MD5_FALLBACK.md
+ */
+function createS3ClientWithMD5(configuration = {}) {
```
DarkIsDude:
```suggestion
function createS3ClientWithMD5Checksum(configuration = {}) {
```

nits

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
 }
+
 function createBucket(bucket, cb) {
-    return s3Client.createBucket({
-        Bucket: bucket,
-    }, (err, data) => {
-        assert.ifError(err);
-        return cb(err, data);
-    });
+    return s3Client.send(new CreateBucketCommand({ Bucket: bucket }))
```
DarkIsDude: Do we need to return ? Seems that we use only the cb here ?

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
 function deleteObject(bucket, key, cb) {
-    return s3Client.deleteObject({
-        Bucket: bucket,
-        Key: key,
-    }, err => {
-        assert.ifError(err);
-        return cb(err);
-    });
+    return s3Client.send(new DeleteObjectCommand({ Bucket: bucket, Key: key }))
+        .then(() => cb())
```
DarkIsDude:
```suggestion
        .then(cb)
```

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
+
 function deleteBucket(bucket, cb) {
-    return s3Client.deleteBucket({
-        Bucket: bucket,
-    }, err => {
-        assert.ifError(err);
-        return cb(err);
-    });
+    return s3Client.send(new DeleteBucketCommand({ Bucket: bucket }))
+        .then(() => cb())
```
DarkIsDude:
```suggestion
        .then(cb)
```

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
-            Quiet: true,
-        },
+        Delete: deleteRequest,
     };
-    return s3Client.deleteObjects(params, err => {
-        assert.ifError(err);
-        return cb(err);
-    });
+    return s3Client.send(new DeleteObjectsCommand(params))
+        .then(() => cb())
```
DarkIsDude:
```suggestion
        .then(cb)
```

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
-        Bucket: bucket,
-        CopySource: `/${bucket}/${key}`,
-        Key: `${key}-copy`,
-    }, err => {
-        assert.ifError(err);
-        return cb(err);
-    });
+    const params = { Bucket: bucket, CopySource: `${bucket}/${key}`, Key: `${key}-copy` };
+    return s3Client.send(new CopyObjectCommand(params))
+        .then(() => cb())
```
DarkIsDude:
```suggestion
        .then(cb)
```

Same for other iterations

tests/utapi/awsNodeSdk.js
```diff
@@ -1,159 +1,194 @@
...
+
 function removeAllVersions(params, callback) {
     const bucket = params.Bucket;
     async.waterfall([
-        cb => s3Client.listObjectVersions(params, cb),
-        (data, cb) => deleteVersionList(data.DeleteMarkers, bucket,
-            err => cb(err, data)),
-        (data, cb) => deleteVersionList(data.Versions, bucket,
-            err => cb(err, data)),
+        cb => s3Client.send(new ListObjectVersionsCommand(params))
```
DarkIsDude:
```suggestion
        async() => s3Client.send(new ListObjectVersionsCommand(params)),
```

Async can understand the async/await pattern

tests/utapi/awsNodeSdk.js
```diff
@@ -162,33 +197,28 @@ function objectMPU(bucket, key, parts, partSize, callback) {
...
+
 function getObject(bucket, key, cb) {
-    return s3Client.getObject({
-        Bucket: bucket,
-        Key: key,
-    }, (err, data) => {
-        assert.ifError(err);
-        return cb(err, data);
-    });
+    return s3Client.send(new GetObjectCommand({ Bucket: bucket, Key: key }))
```
DarkIsDude: Same it's maybe not needed to return ?

tests/sur/quota.js
```diff
@@ -27,26 +43,30 @@ function createBucket(bucket, locked, cb) {
...
         config.ObjectLockEnabledForBucket = true;
     }
-    return s3Client.createBucket(config, (err, data) => {
-        assert.ifError(err);
-        return cb(err, data);
-    });
+    return s3Client.send(new CreateBucketCommand(config)) 
+        .then(data => cb(null, data))
+        .catch(err => {
+            assert.ifError(err);
```
DarkIsDude: I was curious about an assert in a catch and I think this will not work:

```

describe.only('Array', () => {
  it.only('should return -1 when the value is not present', done => {
    function test(cb) {
      new Promise((_resolve, reject) => {
        reject('C est pas vrai');
      })
        .then(cb)
        .catch(err => {
          assert.ifError(err);
          cb();
        });
    }

    test(done);
  });
});
```

The callback is never called ?

tests/sur/quota.js
```diff
@@ -57,109 +77,136 @@ function putObjectLockConfiguration(bucket, cb) {
...
+    }))
+        .then(data => {
+            if (!s3Config.isQuotaInflightEnabled()) {
+                mockScuba.incrementBytesForBucket(bucket, size);
+            }
+            return cb(null, data);
+        })
+        .catch(err => {
+            cb(err);
+        });
```
DarkIsDude:
```suggestion
        .catch(cb(err));
```

tests/sur/quota.js
```diff
@@ -57,109 +77,136 @@ function putObjectLockConfiguration(bucket, cb) {
...
         Bucket: bucket,
         Key: key,
         Body: Buffer.alloc(size),
-    });
+    };
 
-    request.on('build', () => {
-        request.httpRequest.headers['x-scal-s3-version-id'] = vID;
-    });
+    // Add custom header using middleware
```
DarkIsDude:
```suggestion
```

tests/sur/quota.js
```diff
@@ -57,109 +77,136 @@ function putObjectLockConfiguration(bucket, cb) {
...
-        return cb(err, data);
-    });
+    return s3Client.send(command)
+        .then(data => {
+            if (!s3Config.isQuotaInflightEnabled()) {
+                mockScuba.incrementBytesForBucket(bucket, 0);
+            }
+            cb(null, data);
+        })
+        .catch(err => cb(err));
```
DarkIsDude:
```suggestion
        .catch(cb);
```

tests/sur/quota.js
DarkIsDude: `catch(cb)

tests/sur/quota.js
```diff
@@ -296,35 +339,48 @@ function uploadPartCopy(bucket, key, partNumber, partSize, sleepDuration, keyToC
...
+        .then(data => {
+            callback(null, data);
+        })
+        .catch(err => {
+
+            if (!s3Config.isQuotaInflightEnabled()) {
+                mockScuba.incrementBytesForBucket(bucket, size);
+            }
+            return callback(err);
+        });
```
DarkIsDude, line 356:
```suggestion
        .catch(err => {
            if (!s3Config.isQuotaInflightEnabled()) {
                mockScuba.incrementBytesForBucket(bucket, size);
            }
            
            return callback(err);
        });
```

nits

tests/sur/quota.js
```diff
@@ -361,10 +461,12 @@ function multiObjectDelete(bucket, keys, size, callback) {
                 next => createBucket(bucket, false, next),
                 next => sendRequest(putQuotaVerb, '127.0.0.1:8000', `/${bucket}/?quota=true`,
                     JSON.stringify(quota), config).then(() => next()).catch(err => next(err)),
-                next => putObject(bucket, key, size, err => {
-                    assert.strictEqual(err.code, 'QuotaExceeded');
-                    return next();
-                }),
+                next => {
+                    putObject(bucket, key, size, err => {
+                        assert.strictEqual(err.Code, 'QuotaExceeded');
```
DarkIsDude: Same here, I'm not sure the callback will be called ?

DarkIsDude (changes requested):

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
benzekrimaha: good catch missed it here

tests/functional/aws-node-sdk/test/support/config.js
benzekrimaha: Actually it's either set or not at all this is the check we have across the whole code base so I'd rather keep it

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -136,7 +148,7 @@ class BucketUtility {
 
         return Promise.all(promises);
     }
-
+
```
ghost, line 143:
```suggestion

```

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -87,46 +101,44 @@ class BucketUtility {
...
-                                    Key: object.Key,
-                                    VersionId: object.VersionId,
-                                    ...(BypassGovernanceRetention && { BypassGovernanceRetention }),
-                                }).promise()
-                                .then(() => object)
-                            )
+                    )
+                    .concat((data.DeleteMarkers || [])
+                        .map(object =>
+                        this.s3.send(new DeleteObjectCommand({
```
ghost:
```suggestion
                        .map(object => this.s3.send(new DeleteObjectCommand({
```

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -1,81 +1,95 @@
...
+        return this.s3.send(new CreateBucketCommand({
             Bucket: bucketName,
             ObjectLockEnabledForBucket: true,
-        }).promise()
-        .then(() => bucketName);
+        }))
+            .then(() => bucketName)
+            .catch(err => {
+                throw err;
+            });
```
ghost: You can remove the .catch here, if we just re-throw inside

```suggestion
```

package.json
```diff
@@ -19,6 +19,8 @@
   },
   "homepage": "https://github.com/scality/S3#readme",
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.705.0",
```
ghost:
```suggestion
    "@aws-sdk/client-s3": "^3.908.0",
```

ghost (approved):
Some code style issues
When the code style comments from @DarkIsDude are addressed we should be good, so approving for my part

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
```diff
@@ -1,81 +1,95 @@
...
-
-        this.s3 = new S3(s3Config);
-        this.s3.config.setPromisesDependency(Promise);
-        this.s3.config.update({
-            maxRetries: 0,
-        });
+        if (unauthenticated) {
+            this.s3 = new S3Client({
+                ...s3Config,
+                maxAttempts: 0,
```
SylvainSenechal, line 19: Should be 1, although I guess probably 0 is interpreted as 1 too

tests/sur/quota.js
```diff
@@ -248,30 +269,28 @@ function uploadPartCopy(bucket, key, partNumber, partSize, sleepDuration, keyToC
...
-            return s3Client.uploadPart(uploadPartParams, (err, data) => {
-                if (err) {
-                    return next(err);
-                }
-                ETags[partNumber] = data.ETag;
-                return next();
-            });
+            return s3Client.send(new UploadPartCommand(uploadPartParams))
+                .then(data => {
+                    ETags[partNumber] = data.ETag;
```
SylvainSenechal, line 273: Just wanna make sure :
In the function UploadPartCopyCommand, you used
`data.CopyPartResult.ETag`
Are you sure that here, its not something like 
`data.UploadPartResult.ETag` ?

SylvainSenechal (commented):
Looks good to me, I'll let you address the last comments and Edouard approve it

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
benzekrimaha: 0 is indeed interpreted as 1 so keeping it that way as it's working across all tests

tests/sur/quota.js
benzekrimaha: Based on the documentation and tests passing it's directly retrieved with .Etag no object encapsulation here => https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/UploadPartCommandOutput/

tests/functional/aws-node-sdk/lib/utility/bucket-util.js
benzekrimaha: this is indeed accurate when the function is async , which is not the case here , in case of an error the stack trace will still show the error which is what we want / also these functions are used in tests only => we'll be able to see the api logs in case of debug. Still the catch is unnecessary so removing it

tests/utapi/awsNodeSdk.js
benzekrimaha: with the latest fixes made by leif on the api level we don't need this anymore so it has been removed as we only validate the md5 when provided now

tests/utapi/awsNodeSdk.js
benzekrimaha: .then(cb) would be problematic because it doesn't follow the error-first callback convention that Node.js expects.

DarkIsDude (approved):

SylvainSenechal (approved):

PR #5951, opened 2025-09-24
Improvement/cldsrv 724 kms tests migration

benzekrimaha:
Issue: CLDSRV-724

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   "dependencies": {
     "@aws-sdk/client-s3": "^3.908.0",
     "@aws-sdk/credential-providers": "^3.864.0",
+    "@aws-sdk/middleware-retry": "^3.374.0",
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
     "arsenal": "git+https://github.com/scality/Arsenal#8.2.43",
--- a/tests/functional/sse-kms-migration/arnPrefix.js
+++ b/tests/functional/sse-kms-migration/arnPrefix.js
@@ -34,8 +34,8 @@ describe('SSE KMS arnPrefix', () => {
                 ? bkt.kmsKeyInfo.masterKeyArn
                 : bkt.kmsKeyInfo.masterKeyId;
         }
-        await helpers.s3.createBucket(({ Bucket: bkt.name })).promise();
-        await helpers.s3.createBucket(({ Bucket: bkt.vname })).promise();
+        await helpers.s3.createBucket(({ Bucket: bkt.name }));
+        await helpers.s3.createBucket(({ Bucket: bkt.vname }));
         if (bktConf.deleteSSE) {
             await scenarios.deleteBucketSSEBeforeEach(bkt.name, log);
             await scenarios.deleteBucketSSEBeforeEach(bkt.vname, log);
@@ -46,12 +46,12 @@ describe('SSE KMS arnPrefix', () => {
                 Bucket: bkt.name,
                 ServerSideEncryptionConfiguration: helpers.hydrateSSEConfig({
                     algo: bktConf.algo, masterKeyId: bkt.kmsKey }),
-            }).promise();
+            });
             await helpers.s3.putBucketEncryption({
                 Bucket: bkt.vname,
                 ServerSideEncryptionConfiguration: helpers.hydrateSSEConfig({
                     algo: bktConf.algo, masterKeyId: bkt.kmsKey }),
-            }).promise();
+            });
         }
 
         // Put an object for each SSE conf in each bucket
@@ -77,7 +77,7 @@ describe('SSE KMS arnPrefix', () => {
     before('setup', async () => {
         console.log('Run arnPrefix',
             { profile: helpers.credsProfile, accessKeyId: helpers.s3.config.credentials.accessKeyId });
-        const allBuckets = (await helpers.s3.listBuckets().promise()).Buckets.map(b => b.Name);
+        const allBuckets = (await helpers.s3.listBuckets()).Buckets.map(b => b.Name);
         console.log('List buckets:', allBuckets);
         await helpers.MD.setup();
         copyKmsKey = (await helpers.createKmsKey(log)).masterKeyArn;
@@ -92,13 +92,13 @@ describe('SSE KMS arnPrefix', () => {
         } catch (e) { void e; }
 
         // init copy bucket
-        await helpers.s3.createBucket(({ Bucket: copyBkt })).promise();
-        await helpers.s3.createBucket(({ Bucket: mpuCopyBkt })).promise();
+        await helpers.s3.createBucket(({ Bucket: copyBkt }));
+        await helpers.s3.createBucket(({ Bucket: mpuCopyBkt }));
         await helpers.s3.putBucketEncryption({
             Bucket: copyBkt,
             ServerSideEncryptionConfiguration: helpers.hydrateSSEConfig({ algo: 'aws:kms', masterKeyId: copyKmsKey }),
-        }).promise();
-        await helpers.s3.putObject({ Bucket: copyBkt, Key: copyObj, Body: 'BODY(copy)' }).promise();
+        });
+        await helpers.s3.putObject({ Bucket: copyBkt, Key: copyObj, Body: 'BODY(copy)' });
 
         // Prepare every buckets with 1 object (for copy)
         await Promise.all(scenarios.testCases.map(async bktConf => this.initBucket(bktConf)));
@@ -192,7 +192,7 @@ describe('SSE KMS arnPrefix', () => {
             it('should encrypt MPU and put 2 encrypted parts', async () => {
                 const mpuKey = `${obj.name}-mpu`;
                 const mpu = await helpers.s3.createMultipartUpload(
-                    helpers.putObjParams(bkt.name, mpuKey, objConf, obj.kmsKey)).promise();
+                    helpers.putObjParams(bkt.name, mpuKey, objConf, obj.kmsKey));
                 const partsBody = [`${obj.body}-MPU1`, `${obj.body}-MPU2`];
                 const newParts = [];
                 for (const [index, body] of partsBody.entries()) {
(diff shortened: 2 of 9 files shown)


tests/functional/sse-kms-migration/beforeMigration.js
```diff
@@ -326,7 +326,7 @@ describe('SSE KMS before migration', () => {
             it(`should PutObject versioned with SSE ${obj.name}`, async () => {
                 // ensure versioned bucket is empty
                 await helpers.bucketUtil.empty(bkt.vname);
-                let { Versions } = await helpers.s3.listObjectVersions({ Bucket: bkt.vname }).promise();
+                let { Versions } = await helpers.s3.listObjectVersions({ Bucket: bkt.vname }) || [];
```
DarkIsDude: `Versions` can then be undefined ?

tests/functional/sse-kms-migration/cleanup.js
```diff
@@ -18,7 +18,7 @@ describe('SSE KMS Cleanup', () => {
     it('Empty and delete buckets for SSE KMS Migration', async () => {
         console.log('Run cleanup',
             { profile: helpers.credsProfile, accessKeyId: helpers.s3.config.credentials.accessKeyId });
-        const allBuckets = (await helpers.s3.listBuckets().promise()).Buckets.map(b => b.Name);
+        const allBuckets = ((await helpers.s3.listBuckets()).Buckets || []).map(b => b.Name);
```
DarkIsDude, line 21: We have same stuff for other tests (for example in beforeMigration). Seems like a debug ? Should we keep it ? Check the length ?

tests/functional/sse-kms-migration/scenarios.js
```diff
@@ -96,10 +96,12 @@ async function deleteBucketSSEBeforeEach(bktName, log) {
 }
 
 async function getBucketSSEError(Bucket) {
-    await assert.rejects(helpers.s3.getBucketEncryption({ Bucket }).promise(), err => {
```
DarkIsDude: Why we don't keep the reject assertion ?

tests/functional/sse-kms-migration/scenarios.js
```diff
@@ -233,6 +235,10 @@ async function copyObjectAndSSE(
 // check MPU headers against the MPU overview MD
 // because there is no migration for ongoing MPU
 function assertMPUSSEHeaders(actual, expected, algo) {
+    // eslint-disable-next-line no-console
```
DarkIsDude: To be deleted 🙏

tests/functional/sse-kms-migration/scenarios.js
```diff
@@ -270,24 +276,45 @@ async function mpuUploadPartCopy(
...
         CopySourceRange,
-    }).promise();
+    });
     testCase !== 'before' && assertMPUSSEHeaders(part, mpuOverviewMDSSE, algo);
     return part;
 }
 
 // before has no headers to assert
 async function mpuComplete({ UploadId, Bucket, Key }, { existingParts, newParts }, mpuOverviewMDSSE, algo, testCase) {
+    const extractETag = part => {
```
DarkIsDude, line 282: As we are in a test, we should prefer `assert` instead of throw ? Also `return part.CopyPartResult?.ETag || part.ETag || undefined` is better with a check that the result is not undefined ?

tests/functional/sse-kms-migration/scenarios.js
```diff
@@ -270,24 +276,45 @@ async function mpuUploadPartCopy(
...
+        }
+        // For regular uploadPart responses, ETag is at the top level
+        if (part.ETag) {
+            return part.ETag;
+        }
+        // If neither exists, throw an error for debugging
+        throw new Error(`Could not find ETag in part: ${JSON.stringify(part)}`);
+    };
+    
+    // Build the parts array with proper ETag extraction
```
DarkIsDude:
```suggestion
```

tests/functional/sse-kms-migration/helpers.js
```diff
@@ -14,20 +37,121 @@ function getKey(key) {
     return config.kmsHideScalityArn ? getKeyIdFromArn(key) : key;
 }
 
-// for Integration use default profile, in cloudserver use vault profile
+// For Integration use default profile, in cloudserver use vault profile
 const credsProfile = process.env.S3_END_TO_END === 'true' ? 'default' : 'vault';
-const s3config = getConfig(credsProfile, { signatureVersion: 'v4' });
-const s3 = new S3(s3config);
+
+// Create custom agents with specific pooling settings
```
DarkIsDude:
```suggestion
```

if you want to make it clear, you can change the name of the var `httpAgent` with `httpAgentWithSpecificPooling` if this is important when we read tests

tests/functional/sse-kms-migration/helpers.js
```diff
@@ -14,20 +37,121 @@ function getKey(key) {
...
+const s3Client = new S3Client(s3config);
+
+// Remove logger middleware if present
+if (s3Client.middlewareStack.identify().includes('loggerMiddleware')) {
+    s3Client.middlewareStack.remove('loggerMiddleware');
+}
+
 const bucketUtil = new BucketUtility(credsProfile);
 
+// Wrapper for SDK v3 commands to return promises directly
```
DarkIsDude:
```suggestion
```

Your code is clear enough

tests/functional/sse-kms-migration/helpers.js
```diff
@@ -14,20 +37,121 @@ function getKey(key) {
...
+                error.code === 'RequestTimeout'
+            )
+        ,
+        delayDecider: attempts => Math.min(1000 * Math.pow(2, attempts), 30000), // Exponential backoff
+    }),
+};
+
+const s3Client = new S3Client(s3config);
+
+// Remove logger middleware if present
```
DarkIsDude:
```suggestion
```

tests/functional/sse-kms-migration/helpers.js
```diff
@@ -51,29 +175,29 @@ const MD = {
 };
 
 async function getBucketSSE(Bucket) {
-    const sse = await s3.getBucketEncryption({ Bucket }).promise();
-    return sse
-        .ServerSideEncryptionConfiguration
-        .Rules[0]
-        .ApplyServerSideEncryptionByDefault;
+    try {
+        const sse = await s3Client.send(new GetBucketEncryptionCommand({ Bucket }));
```
DarkIsDude: Should you use your wrapper instead of `s3Client` ?

tests/functional/sse-kms-migration/arnPrefix.js
```diff
@@ -602,12 +602,12 @@ describe('KMS error', () => {
...
         {
             action: 'getObject', kmsAction: 'Decrypt',
-            fct: async () => helpers.s3.getObject({ Bucket, Key }).promise(),
+            fct: async () => helpers.s3.getObject({ Bucket, Key }) ,
         },
         {
             action: 'copyObject', detail: ' when getting from source', kmsAction: 'Decrypt',
             fct: async () =>
-                helpers.s3.copyObject({ Bucket, Key: 'copy', CopySource: `${Bucket}/${Key}` }).promise(),
+                helpers.s3.copyObject({ Bucket, Key: 'copy', CopySource: `${Bucket}/${Key}` }) ,
```
DarkIsDude:
```suggestion
                helpers.s3.copyObject({ Bucket, Key: 'copy', CopySource: `${Bucket}/${Key}` }),
```

You have some other iterations of that in your file

DarkIsDude (changes requested):

tests/functional/sse-kms-migration/beforeMigration.js
ghost: Should probably be like this to ensure we do not have an error trying to access `Versions` from an array that has no fields

```suggestion
                let { Versions } = await helpers.s3.listObjectVersions({ Bucket: bkt.vname }) || { Versions: [] };
```

tests/functional/sse-kms-migration/helpers.js
```diff
@@ -14,20 +37,121 @@ function getKey(key) {
...
+
+const s3config = {
+    ...getConfig(credsProfile, {}),
+    requestHandler: new NodeHttpHandler({
+        connectionTimeout: 120000,
+        socketTimeout: 120000,
+        httpAgent,
+        httpsAgent,
+    }),
+    maxAttempts: 8,
```
ghost: Usually we can keep 3, this should be enough (and one could argue it should be 1 for a test suite, as if we hide transient errors, we might miss real bugs; and in such case we could just remove the retry logic)

tests/functional/sse-kms-migration/cleanup.js
benzekrimaha: what do you mean by the same stuff , are you referring to the console logs ? as they have been there before they're being kept

tests/functional/sse-kms-migration/scenarios.js
benzekrimaha: the try-catch approach provides clearer information in this case thus the replacement

tests/functional/sse-kms-migration/helpers.js
```diff
@@ -14,20 +36,105 @@ function getKey(key) {
...
+        socketTimeout: 120000,
+        httpAgent,
+        httpsAgent,
+    }),
+    maxAttempts: 3,
+};
+
+const s3Client = new S3Client(s3config);
+
+if (s3Client.middlewareStack.identify().includes('loggerMiddleware')) {
```
SylvainSenechal, line 72: I think it deserves a comment, why are we doing this ? Is this removing some logs to make the test less verbose ?

package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "@aws-sdk/client-s3": "^3.705.0",
```
SylvainSenechal: nit: 705 is ~11 months old, you can use 910 if you want

SylvainSenechal (commented):
Nothing much to change here, but tests are failing

DarkIsDude (approved):

SylvainSenechal (approved):
lets gooo

PR #5952, opened 2025-09-25
CLDSRV-741: Validate notification event usage

BourgoisMickael:
To make sure end user doesn't expect to receive events
if cloudserver config doesn't support the feature

cf: https://github.com/scality/Arsenal/pull/2527

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -19,7 +19,7 @@ const {
     allowedUtapiEventFilterFields,
     allowedUtapiEventFilterStates,
 } = require('../constants');
-const { ValidLifecycleRules: supportedLifecycleRules } = require('arsenal').models;
+const { ValidLifecycleRules: supportedLifecycleRules, NotificationConfiguration } = require('arsenal').models;
 const { utapiVersion } = require('utapi');
 const { scaleMsPerDay } = s3middleware.objectUtils;
 const { constants: arsenalConstants } = require('arsenal');
@@ -1774,6 +1774,8 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+        // Restrict events to supported product features to be explicit about what can be notified to end user
+        NotificationConfiguration.restrictSupportedNotificationBasedOnLifecycle(this.supportedLifecycleRules);
 
         this.integrityChecks = parseIntegrityChecks(config);
 
--- a/tests/functional/aws-node-sdk/test/bucket/putBucketNotification.js
+++ b/tests/functional/aws-node-sdk/test/bucket/putBucketNotification.js
@@ -4,6 +4,7 @@ const { S3 } = require('aws-sdk');
 const checkError = require('../../lib/utility/checkError');
 const getConfig = require('../support/config');
 const BucketUtility = require('../../lib/utility/bucket-util');
+const { config } = require('../../../../../lib/Config');
 
 const bucket = 'mock-notification-bucket';
 
@@ -114,6 +115,35 @@ describe('aws-sdk test put notification configuration', () => {
                     done();
                 });
             });
+    });
+
+    describe('event validation', () => {
+        before(done => s3.createBucket({ Bucket: bucket }, done));
+
+        after(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+        const events = [
+            { supported: 'Transition', event: 's3:ObjectRestore:*' },
+            { supported: 'Transition', event: 's3:LifecycleTransition' },
+            { supported: 'Expiration', event: 's3:LifecycleExpiration:*' },
+        ];
+        events.forEach(({ supported, event }) => {
+            describe(`${event} event validation`, () => {
+                it(`should handle ${event} events based on lifecycle rules configuration`, done => {
+                    const params = getNotificationParams([event]);
+                    s3.putBucketNotificationConfiguration(params, err => {
+                        if (config.supportedLifecycleRules.some(rule => rule.includes(supported))) {
+                            // Should succeed when lifecycle rule is supported
+                            assert.ifError(err);
+                        } else {
+                            // Should fail when lifecycle rule is not supported
+                            checkError(err, 'MalformedXML', 400);
+                        }
+                        done();
+                    });
+                });
+            });
+        });
     });
 
     describe('cross origin requests', () => {
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.2.30",
+    "arsenal": "git+https://github.com/scality/Arsenal#8.2.32",
     "async": "2.6.4",
     "aws-sdk": "^2.1692.0",
     "bucketclient": "scality/bucketclient#8.2.5",
--- a/tests/functional/aws-node-sdk/test/object/mpuVersion.js
+++ b/tests/functional/aws-node-sdk/test/object/mpuVersion.js
@@ -97,6 +97,20 @@ function checkObjMdAndUpdate(objMDBefore, objMDAfter, props) {
     });
 }
 
+function clearUploadIdFromVersions(versions) {
+    if (!versions || versions.length === 0) {
+        return versions;
+    }
+
+    for (const version of versions) {
+        if (version.value) {
+            version.value.uploadId = undefined;
+        }
+    }
+
+    return versions;
+}
+
 // TODO: CLDSRV-721 RING 10 Support ObjectRestore (cold storage) with MD v1
 // The cold storage test suite is skipped as bad versionId while faking archive breaks after each bucket cleanup
 const describeSkipNullMdV1 = isNullKeyMetadataV1 || !hasColdStorage ? describe.skip : describe;
@@ -245,7 +259,7 @@ describe('MPU with x-scal-s3-version-id header', () => {
                         return next(err);
                     }),
                     next => metadata.listObject(bucketName, mdListingParams, log, (err, res) => {
-                        versionsBefore = res.Versions;
+                        versionsBefore = clearUploadIdFromVersions(res.Versions);
                         return next(err);
                     }),
                     next => putMPUVersion(s3, bucketName, objectName, '', next),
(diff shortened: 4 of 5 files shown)


lib/Config.js
```diff
@@ -817,6 +817,7 @@ class Config extends EventEmitter {
                 }
             }
         });
+        this.hasColdStorage = Object.values(this.locationConstraints).some(l => l.isCold);
```
francoisferrand: Not sure we should do this check: during setup, we don't know in which order bucket notif vs location creation may be done.

One could setup notification, and later create/update the locations.

Another use case would be to setup a "catch all" notification config, applied everywhere.

lib/api/apiUtils/bucket/getNotificationConfiguration.js
```diff
@@ -31,6 +31,45 @@ function getNotificationConfiguration(parsedXml) {
         error = error.addMetadataEntry('invalidArguments', invalidTargets);
         return { error };
     }
+
```
francoisferrand: There is already a validation in arsenal (if I recall correctly), where we traverse the whole list already.

Should the check not be added in there?

lib/api/apiUtils/bucket/getNotificationConfiguration.js
```diff
@@ -31,6 +31,45 @@ function getNotificationConfiguration(parsedXml) {
...
+            if (t.events.some(e => e.startsWith('s3:LifecycleExpiration'))) {
+                invalidEvents.push({
+                    ArgumentName: 's3:LifecycleExpiration',
+                    ArgumentValue: 'Expiration lifecycle rule is not supported',
+                });
+            }
+        });
+    }
+    if (invalidEvents.length > 0) {
+        const errDesc = 'Unable to validate the following event configurations';
```
francoisferrand: Does this match AWS error?

francoisferrand:
Generally, I don't think bucket notification is about "expecting" something, but about being ready to handle it: so not sure if this is really required.

More precisely here, there are 2 separate cases:
- if the condition is related to the "config" (ie it can be changed by the user or admin, like a location or system parameter): then we should not do such check IMO
- it would be consistent however to reject (not' implemented!) events which are disabled due to the way cloudserver is configured/deployed in the product (ie things that cannot be changed by end user)
- (eventually, we may have a third case where an admin would choose which feature to enable for an account... but we are not there yet)

lib/api/apiUtils/bucket/getNotificationConfiguration.js
```diff
@@ -31,6 +31,45 @@ function getNotificationConfiguration(parsedXml) {
         error = error.addMetadataEntry('invalidArguments', invalidTargets);
         return { error };
     }
+
+    // getting invalid events
+    const invalidEvents = [];
+    if (!config.hasColdStorage) {
```
francoisferrand: Restore/coldstorage is based on lifecycle transition; so these events should be rejected if transitions are not supported by the product.
(I don't think we have different flags for hot vs cold transition)

lib/Config.js
BourgoisMickael: I've added it as well in the live update by management path [945299e](https://github.com/scality/cloudserver/pull/5952/commits/945299e641757d2f869cc0053dab56e6814b42b3)

lib/api/apiUtils/bucket/getNotificationConfiguration.js
BourgoisMickael: But transition could also be used for non cold storage related transition (like infrequent access) even if we don't manage such thing now.

image

And cold storage could be targetted directly on:
- bucket creation with `--create-bucket-configuration LocationConstraint=eu-west-1`
- put object with `--storage-class`

lib/api/apiUtils/bucket/getNotificationConfiguration.js
BourgoisMickael: arsenal checks against all possible events, just like aws does with a generic error for the whole payload.

Here the validation is specific to cloudserver config with specific message, just like it's done for the target validation

lib/api/apiUtils/bucket/getNotificationConfiguration.js
BourgoisMickael: AWS doesn't have such error as they handle all events. This match the same message as invalid target to better help end user know that some specific events can't be used.

I did not want to return a 501 NotImplemented as the end user should rather remove the invalid events and then the request can pass successfully. (the error can be fixed and all the invalid arguments are listed in the error)

lib/api/apiUtils/bucket/getNotificationConfiguration.js
francoisferrand: - we do not support writing "directly" to cold storage
- consequently cold storage cannot be used as default location

in _**our**_ implementation, cold storage can thus only be used through transition.

If we want to make a check for s3:restore, the only "flag" we have now is if transitions are enabled (c.f. previous comment, checking the _current_ locations is not a good idea IMHO, as it may change over time)

lib/api/apiUtils/bucket/getNotificationConfiguration.js
francoisferrand: yes, but I wonder if this validation should not be done as well by arsenal: i.e. keep related things together.

the arsenal function could be passed the "supported lifecycle rules", just like we do for validating putBucetLifecycleConfiguration.

lib/api/apiUtils/bucket/getNotificationConfiguration.js
francoisferrand: maybe this could happen with AWS when policies are used to present use of some storage class?
not exactly the same usecase obviously, but maybe it relates?

lib/api/apiUtils/bucket/getNotificationConfiguration.js
francoisferrand: my point is not that we should handle update, but really that it seems a very valid use case to setup notification _before_ setting up the (cold) location.

generally, we should not restrict the API because _user_ has created some resources : but only because of _product_ configuration (i.e. something _we_ control, not the _end user_)

francoisferrand:
Overall, while I agree the problem is completely valid and must be addressed:

> S3C should reject bucket notifications related to unsupported features (restore, transition...)

I don't think the way it is approached is correct:

> To make sure end user doesn't expect to receive events if cloudserver config doesn't support the feature

We don't want to try to be smart and imagine what user is allowed to do : we really just want to be like amazon, and allow user to register for any notification that _may_ be generated by the product - even if they will never receive it because they don't perform the require configuration later on. In other words, bucket notifications is really saying "I want to be notified when XXX happens", and it does not provide any warantee that such event will happen.

→ This could be done similarly to how we configure the list of lifecycle "features", or possibly (as part of your change does) we could "derive/filter" that list of product-supported lifecycle workflows.
→ There is no reason to change the behavior on XDM, and prevent user from registering (for exemple) s3:restoreObject notifications

lib/api/apiUtils/bucket/getNotificationConfiguration.js
BourgoisMickael: I've changed the condition to rely on lifecycle transition to detect cold storage: [ead2d95](https://github.com/scality/cloudserver/pull/5952/commits/ead2d95145904eeb314f91ca775debf0761d159a)

lib/Config.js
BourgoisMickael: For a "catch all" notification config, end user should follow our documentation to know about what are supported events in the product.

lib/Config.js
```diff
@@ -1773,6 +1773,12 @@ class Config extends EventEmitter {
         }
 
         this.supportedLifecycleRules = parseSupportedLifecycleRules(config.supportedLifecycleRules);
+        /**
+         * We consider cold storage supported if Transition lifecycle rule is supported.
+         * In our implementation, it's the only way to move objects to cold storage (can't write directly to cold)
+         * Even if Transition could (rarely) be used without cold storage location (hot to hot transition)
```
```suggestion
         * Even if Transition could be used without cold storage location (hot transition), but both features are
         * enabled together at the moment.
```

lib/api/apiUtils/bucket/getNotificationConfiguration.js
francoisferrand: > I've changed the condition to rely on lifecycle transition to detect cold storage: [ead2d95](https://github.com/scality/cloudserver/pull/5952/commits/ead2d95145904eeb314f91ca775debf0761d159a)

👍

> There is already a validation in arsenal (if I recall correctly), where we traverse the whole list already.

The check in arsenal is already implemented and efficient, based on a `Set` which is build just once, and traversing the rules only once as well : I think it would be better to stick with the same approach, and possibly `compute` this set based on `SupportedLifecycleRules`.

This would make code simpler, faster, and ensures we have the exact same behavior as 7.x branches when transitions are disabled.

(since supported lifecycles rules are set on startup, we should simply build the "filtered" Set once on startup, then forget about it)

tests/functional/aws-node-sdk/test/bucket/putBucketNotification.js
```diff
@@ -116,6 +117,50 @@ describe('aws-sdk test put notification configuration', () => {
...
+    describe('event validation', () => {
+        before(done => s3.createBucket({ Bucket: bucket }, done));
+
+        after(done => s3.deleteBucket({ Bucket: bucket }, done));
+
+        describe('s3:ObjectRestore event validation', () => {
+            it('should handle s3:ObjectRestore events based on cold storage configuration', done => {
+                const params = getNotificationParams(['s3:ObjectRestore:*']);
+                s3.putBucketNotificationConfiguration(params, err => {
+                    if (config.hasColdStorage) {
```
francoisferrand: should we introduce the extra flag, or just look at the (existing) `supportedLifecycleRules` to see if transitions are supported?

francoisferrand: also may be merged with next case, it is just another case:
```
 { supported: 'Transition', event: 's3:ObjectRestore:*' },
```

BourgoisMickael: ~~I also use this flag in those other tests: https://github.com/scality/cloudserver/pull/5953~~

~~That's why I want the flag in the config so I can reuse in multiple places~~

lib/api/apiUtils/bucket/getNotificationConfiguration.js
BourgoisMickael: ~~Something like that [de2bbb1](https://github.com/scality/cloudserver/pull/5952/commits/de2bbb12c74b96dc14e3e507043eb8332d206f75)~~

Maybe I'll move that into arsenal directly

BourgoisMickael: Reworked in arsenal: https://github.com/scality/Arsenal/pull/2527

anurag4DSB (approved):

dvasilas (approved):

PR #5954, opened 2025-09-25
CLDSRV-741: Block internal routes veeam & workflowEngineOperator

BourgoisMickael:
For S3C those routes are not supported, we don't want to have them available to avoid unexpected behavior if used.

+ fix veeam route crash

### DIFF
--- a/.github/docker/config.s3c.json
+++ b/.github/docker/config.s3c.json
@@ -60,5 +60,6 @@
             "type": "dummy",
             "host": "localhost:6000"
         }
-    ]
+    ],
+    "enableVeeamRoute": false
 }
--- a/lib/utilities/internalHandlers.js
+++ b/lib/utilities/internalHandlers.js
@@ -11,9 +11,15 @@ const internalHandlers = {
     backbeat: routeBackbeat,
     report: reportHandler,
     metadata: routeMetadata,
-    'workflow-engine-operator': routeWorkflowEngineOperator,
-    veeam: routeVeeam,
 };
+
+if (config.workflowEngineOperator) {
+    internalHandlers['workflow-engine-operator'] = routeWorkflowEngineOperator;
+}
+
+if (config.enableVeeamRoute) {
+    internalHandlers.veeam = routeVeeam;
+}
 
 if (config.healthChecks.enableInternalRoute) {
     // CLDSRV-740: S3C with the healthcheck on s3 port needs the internal route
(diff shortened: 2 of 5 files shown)


nicolas2bert (approved):

.github/docker/config.s3c.json
```diff
@@ -60,5 +60,6 @@
             "type": "dummy",
             "host": "localhost:6000"
         }
-    ]
+    ],
+    "disableVeeamRoute": true
```
ghost: The /_/ path is supposed to be internal already, and only accessible through the right ingress rules (nginx proxy). Why is it crashing in S3C? Are we exposing those APIs?

lib/utilities/internalHandlers.js
```diff
@@ -11,10 +11,16 @@ const internalHandlers = {
     backbeat: routeBackbeat,
     report: reportHandler,
     metadata: routeMetadata,
-    'workflow-engine-operator': routeWorkflowEngineOperator,
-    veeam: routeVeeam,
 };
 
+if (config.workflowEngineOperator) {
+    internalHandlers['workflow-engine-operator'] = routeWorkflowEngineOperator;
+}
```
ghost, line 18: I believe this could be totally removed? It's not used anymore.
Maybe we can also introduce a "enableVeeamRoute" fflag instead of the negative one, so we are consistent across the internal routes, with a default to true, but if set to false (for s3c), then it's disabled?

.github/docker/config.s3c.json
BourgoisMickael: No the crash was a query like `/_/veeam` without anything else would crash, it got fixed with `|| ''`.

For S3C we want to disable completely the route as we don't support these veeam sosapi

lib/utilities/internalHandlers.js
BourgoisMickael: Well if it's to be removed there is also other code path to cleanup I guess.

I can reverse the veeam flag

BourgoisMickael: There [c090bb7](https://github.com/scality/cloudserver/pull/5954/commits/c090bb77333145bad996761974e2d0622465ab4b)

.github/docker/config.s3c.json
```diff
@@ -60,5 +60,6 @@
             "type": "dummy",
             "host": "localhost:6000"
         }
-    ]
+    ],
+    "enableVeeamRoute": true
```
ghost: false?

ghost (approved):

PR #5955, opened 2025-09-26
CLDSRV-746: Adding log fields to the backbeat routes in order to fill the analytics dashboard

fredmnl:
Adding action, account and userName fields to the backbeat routes. 

`_normalizeBackbeatRequest` was executed too late as it is the one filling in the `bucketName` in the `request` (among other things) and that must happen before this field is used to customize the logger.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -475,7 +475,7 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
  */
 
 /**
- * 
+ *
  * @param {string} accountId - account ID
  * @param {Log} log - logger instance
  * @param {CanonicalIdCallback} cb - callback function
@@ -1466,6 +1466,7 @@ function routeBackbeat(clientIP, request, response, log) {
     const contentLength = request.headers['x-amz-decoded-content-length'] || request.headers['content-length'];
     // eslint-disable-next-line no-param-reassign
     request.parsedContentLength = Number.parseInt(contentLength?.toString() ?? '', 10);
+    _normalizeBackbeatRequest(request);
     log.addDefaultFields({
         clientIP,
         url: request.url,
@@ -1478,7 +1479,7 @@ function routeBackbeat(clientIP, request, response, log) {
     });
 
     log.debug('routing request');
-    _normalizeBackbeatRequest(request);
+
     const requestContexts = prepareRequestContexts('objectReplicate', request);
 
     if (request.resourceType === 'expiration' || request.resourceType === 'batchdelete') {
@@ -1521,6 +1522,12 @@ function routeBackbeat(clientIP, request, response, log) {
             }
             // eslint-disable-next-line no-param-reassign
             request.accountQuotas = infos?.accountQuota;
+
+            // Add authentication info to logs for consistency with regular S3 routes
+            if (userInfo) {
+                log.addDefaultFields(getAuthLogFields(userInfo));
+            }
+
             // FIXME for now, any authenticated user can access API
             // routes. We should introduce admin accounts or accounts
             // with admin privileges, and restrict access to those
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

lib/routes/routeBackbeat.js
```diff
@@ -1475,10 +1476,26 @@ function routeBackbeat(clientIP, request, response, log) {
...
+        'expiration': 'BackbeatExpiration',
+        'batchdelete': 'BackbeatBatchDelete',
+        'data': 'BackbeatPutData',
+        'metadata': 'BackbeatPutMetadata',
+        'multiplebackenddata': 'BackbeatMultipleBackendData',
+        'multiplebackendmetadata': 'BackbeatMultipleBackendMetadata',
+        'index': 'BackbeatIndexOperation',
+        'api': 'BackbeatApi',
+        'lifecycle': 'BackbeatListLifecycle',
+    };
```
BourgoisMickael: Declare this mapping out of the routing to assign it only once and not per request.
___
For the value I'd remove the `Backbeat` prefix as you should already have from arsenal a field `internalServiceName` (https://github.com/scality/Arsenal/blob/13f05191a02fcba1be074b64e23e1be30c135137/lib/s3routes/routes.ts#L247C32-L247C51) that has the value `backbeat`.
___
This map mixes action and backend, `resourceType` is not enough to know the actual action, you might need to use the `method` and the query string `operation`.  For example metadata has a GET and a PUT.

Maybe you should have 2 fields, one for the backend, one for the action.

This map takes care of the routing and describe all the possible actions: https://github.com/scality/cloudserver/blob/c219cf19635b60e4ece20bb090221d46d1cd756d/lib/routes/routeBackbeat.js#L1368-L1403

Maybe the action could be added directly from inside each function separately, or use the function name in the routing:

```diff
                const route = backbeatRoutes[request.method][request.resourceType];
+               log.addDefaultFields({ action: route.name });
                return route(request, response, userInfo, log, err => {
```

lib/routes/routeBackbeat.js
```diff
@@ -1657,5 +1686,19 @@ function routeBackbeat(clientIP, request, response, log) {
         ));
 }
 
+function getAuthLogFields(userInfo) {
```
BourgoisMickael, line 1678: There's another one with unification

```js
                if (isRequesterASessionUser(userInfo)) {
                    authNames.sessionName = userInfo.getShortid().split(':')[1];
                }
```

But I think the session name is already present in userName

lib/routes/routeBackbeat.js
fredmnl: Using the function name sounds good to me.

BourgoisMickael: There are many routing branches with conditions, make sure to catch all of them. And there is a divergence in routing in branch 9.1

In 9.0:
- api: https://github.com/scality/cloudserver/blob/db0280e4c8fae7ab29d5eff7a3e4077031c7b2e6/lib/routes/routeBackbeat.js#L1535
- index routes: https://github.com/scality/cloudserver/blob/db0280e4c8fae7ab29d5eff7a3e4077031c7b2e6/lib/routes/routeBackbeat.js#L1416
- https://github.com/scality/cloudserver/blob/db0280e4c8fae7ab29d5eff7a3e4077031c7b2e6/lib/routes/routeBackbeat.js#L1630-L1638

___

In 9.1:
- api: https://github.com/scality/cloudserver/blob/38fca1ff4a50f0a7cc4304ab972c50156da52293/lib/routes/routeBackbeat.js#L1564
- non-object: https://github.com/scality/cloudserver/blob/38fca1ff4a50f0a7cc4304ab972c50156da52293/lib/routes/routeBackbeat.js#L1503-L1508
- multiplebackendmetadata https://github.com/scality/cloudserver/blob/38fca1ff4a50f0a7cc4304ab972c50156da52293/lib/routes/routeBackbeat.js#L1618-L1625
- https://github.com/scality/cloudserver/blob/38fca1ff4a50f0a7cc4304ab972c50156da52293/lib/routes/routeBackbeat.js#L1647
-

lib/routes/routeBackbeat.js
```diff
@@ -1657,5 +1675,50 @@ function routeBackbeat(clientIP, request, response, log) {
...
+            authNames.IAMdisplayName = userInfo.getIAMdisplayName();
+        }
+    }
+    if (typeof userInfo.isRequesterASessionUser === 'function' && userInfo.isRequesterASessionUser()) {
+        authNames.sessionName = userInfo.getShortid().split(':')[1];
+    }
+    return authNames;
+}
+
+function routeNameFromRequest(request) {
```
BourgoisMickael:
```suggestion
function actionFromRequest(request) {
```

```suggestion
function actionFromRequestRoute(request) {
```

BourgoisMickael (approved):

PR #5957, opened 2025-09-29
CLDSRV-741: Replace null with undefined in bucket

BourgoisMickael:
To make the bucket attributes payload smaller when features are not defined

Bucket payload any features looks like:

```json
{
  "name": "bkt9",
  "owner": "79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be",
  "ownerDisplayName": "Bart",
  "creationDate": "2025-09-29T10:33:41.359Z",
  "mdBucketModelVersion": 16,
  "transient": false,
  "deleted": false,
  "serverSideEncryption": null,
  "locationConstraint": "dc-1",
  "uid": "ce6dbe03-22b4-4354-8668-d6fecbd5dadd",
  "tags": [],
  "quotaMax": "0"
}
```

Dropped fields when null:
```diff
<   "versioningConfiguration": null,
<   "readLocationConstraint": null,
<   "cors": null,
<   "replicationConfiguration": null,
<   "lifecycleConfiguration": null,
<   "bucketPolicy": null,
<   "isNFS": false,
<   "ingestion": null,
<   "azureInfo": null,
```

___

Only sse remain null as it breaks many places if removed

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -165,20 +165,21 @@ function bucketLevelServerSideEncryption(bucket, headers, log, cb) {
  */
 function createBucket(authInfo, bucketName, headers,
     locationConstraint, log, cb) {
+    // Prefer using undefined instead of null for unused properties so they are not present in metadata payload
     log.trace('Creating bucket');
     assert.strictEqual(typeof bucketName, 'string');
     const canonicalID = authInfo.getCanonicalID();
     const ownerDisplayName =
         authInfo.getAccountDisplayName();
     const creationDate = new Date().toJSON();
-    const isNFSEnabled = headers['x-scal-nfs-enabled'] === 'true';
+    const isNFSEnabled = headers['x-scal-nfs-enabled'] === 'true' || undefined;
     const headerObjectLock = headers['x-amz-bucket-object-lock-enabled'];
     const objectLockEnabled
         = headerObjectLock && headerObjectLock.toLowerCase() === 'true';
     const bucket = new BucketInfo(bucketName, canonicalID, ownerDisplayName,
-        creationDate, BucketInfo.currentModelVersion(), null, null, null, null,
-        null, null, null, null, null, null, null, null, null, isNFSEnabled,
-        null, null, objectLockEnabled);
+        creationDate, BucketInfo.currentModelVersion(), undefined, undefined, undefined, undefined,
+        undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, isNFSEnabled,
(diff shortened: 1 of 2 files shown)


lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -165,20 +165,21 @@ function bucketLevelServerSideEncryption(bucket, headers, log, cb) {
...
     locationConstraint, log, cb) {
+    // Prefer using undefined instead of null for unused properties so they are not present in metadata payload
     log.trace('Creating bucket');
     assert.strictEqual(typeof bucketName, 'string');
     const canonicalID = authInfo.getCanonicalID();
     const ownerDisplayName =
         authInfo.getAccountDisplayName();
     const creationDate = new Date().toJSON();
-    const isNFSEnabled = headers['x-scal-nfs-enabled'] === 'true';
+    const isNFSEnabled = headers['x-scal-nfs-enabled'] === 'true' || undefined;
```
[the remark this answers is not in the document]
BourgoisMickael: This is on purpose to drop the boolean field if false

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -165,20 +165,21 @@ function bucketLevelServerSideEncryption(bucket, headers, log, cb) {
  */
 function createBucket(authInfo, bucketName, headers,
     locationConstraint, log, cb) {
+    // Prefer using undefined instead of null for unused properties so they are not present in metadata payload
```
DarkIsDude, line 168: Why this comment ? Maybe make more sense to put that in that class `BucketInfo` ? And/or add a check in the constructor to transform null to undefined ?

DarkIsDude (approved):

lib/api/apiUtils/bucket/bucketCreation.js
BourgoisMickael: The null be default was removed from BucketInfo some time ago and was not replaced by undefined. I guess we might want to leave the option to the called to use null or undefined.

BourgoisMickael: What do you think @williamlardier ?

ghost: I would say I agree with Mickael here, but as we change how the function is called at the same time, it can be a bit risky. We could have some logic relying explicitly on "null", but here, best is to trust the tests, if they pass we are good.

So better to mention, in the `BucketInfo` class, that we should not pass null but undefined to get the default values?

Foir the `null` I agree to keep it if provided, and not fallback to `undefined`.

ghost (approved):

PR #5965, opened 2025-10-02
Remove extra `/` from backbeat proxy path

welansari:
Issue: CLDSRV-755

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -1508,7 +1508,7 @@ function routeBackbeat(clientIP, request, response, log) {
             return responseJSONBody(errors.MethodNotAllowed, null, response,
                 log);
         }
-        const path = request.url.replace('/_/backbeat/api', '/_/');
+        const path = request.url.replace('/_/backbeat/api/', '/_/');
         const { host, port } = config.backbeat;
         const target = `http://${host}:${port}${path}`;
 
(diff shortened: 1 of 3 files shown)


ghost (approved):

lib/routes/routeBackbeat.js
```diff
@@ -1450,7 +1450,7 @@ function routeIndexingAPIs(request, response, userInfo, log, callback) {
 }
 
 function routeBackbeatAPIProxy(request, response, requestContexts, log) {
-    const path = request.url.replace('/_/backbeat/api', '/_/');
+    const path = request.url.replace('/_/backbeat/api/', '/_/');
```
francoisferrand: the same thing seems to be in `development/9.0`, c.f. [1] (and possibly older branches):
- should we not fix it there as well?
- why did we not have the issue before? is it a subtle behavior change in some dependency we bump'ed?

[1] https://github.com/scality/cloudserver/blame/e3d9501177c4ffced964e75ba462b1748861b9cc/lib/routes/routeBackbeat.js#L1511C38-L1511C38

francoisferrand (approved):

lib/routes/routeBackbeat.js
welansari: We always had this bug actually but it used to work as the `http-proxy` lib automatically removed the extra `/` in the proxied path, the lib was bumped in 9.1 but for future safety i agree that it should be fixed in 9.0 as well.

PR #5967, opened 2025-10-10
Install required version of typing-extensions

benzekrimaha:
The build step for the pykmip image is currently failing with : 
error: typing-extensions 4.12.2 is installed but typing-extensions>=4.13.2 is required by {'cryptography'}, this PR makes sure we install a version compatible 

Issue: CLDSRV-757

### DIFF
--- a/.github/pykmip/Dockerfile
+++ b/.github/pykmip/Dockerfile
@@ -13,6 +13,8 @@ RUN apk add --no-cache \
         build-base \
         curl \
         git && \
+    pip3 install --upgrade pip && \
+    pip3 install --upgrade typing-extensions>=4.13.2 && \
     git clone https://github.com/openkmip/pykmip.git && \
     cd pykmip && \
     python3 setup.py install && \


.github/pykmip/Dockerfile
```diff
@@ -13,6 +13,8 @@ RUN apk add --no-cache \
         build-base \
         curl \
         git && \
+    pip3 install --upgrade pip && \
+    pip3 install --upgrade typing-extensions>=4.13.2 && \
```
DarkIsDude, line 17: Who is installing the old and wrong version ?

benzekrimaha: typing-extensions is a python dependency , for more info => https://www.geeksforgeeks.org/python/introduction-to-python-typing-extensions-module/#what-is-the-typingextensions-module, the version installed with the current image used was not the one needed by cryptography as mentioned above thus the need to install it here

welansari (approved):

.github/pykmip/Dockerfile
DarkIsDude: 🙏 Thanks for clarifying. My point was more to find a more "cleaner" solution. Maybe we can open a PR to the package https://github.com/openkmip/pykmip to then be able to delete this fix ? Also why do we need to update pip ? If the module is packaged with the image, we should then maybe update the python:3.10 to use a more recent version ?

benzekrimaha: this is a quick fix to deblock the ci , we can indeed bump the image but we might end up with other dependencies breaking , as this is about one package only this seems like a good approach to me. for pip it's just a good practice to update it before installing a package.

DarkIsDude: 🤝 all good for me then if you have in mind to improve that later 🙏

DarkIsDude (approved):

PR #5968, opened 2025-10-13
Handle 404 from Scuba backend

ghost:
For Veeam reoutes, this will set 0 as the metric. The value is correct: if scuba returns a 404 error, it means there is no metric for the resource.

Issue: CLDSRV-758

### DIFF
--- a/tests/sur/routeVeeam.js
+++ b/tests/sur/routeVeeam.js
@@ -383,6 +383,111 @@ function makeVeeamRequest(params, callback) {
                 return done();
             }));
         });
+
+        it('GET capacity.xml should return 200 when scubaclient returns 404 (post-install scenario)', done => {
+            // This test simulates the post-install scenario where scubaclient returns 404
+            // because no metrics are available yet. By not calling scuba.incrementBytesForBucket,
+            // the mock scuba server will return 404 for this bucket.
+            
+            async.waterfall([
+                next => makeVeeamRequest({
+                    method: 'PUT',
+                    bucket: TEST_BUCKET,
+                    objectKey: '.system-d26a9498-cb7c-4a87-a44a-8ae204f5ba6c/capacity.xml',
+                    headers: {
+                        'content-length': testCapacity.length,
+                        'content-md5': testCapacityMd5,
+                        'x-scal-canonical-id': testArn,
+                    },
+                    authCredentials: veeamAuthCredentials,
+                    requestBody: testCapacity,
+                }, (err, response) => {
+                    if (err) {
+                        return done(err);
+                    }
+                    assert.strictEqual(response.statusCode, 200);
+                    return next();
+                }),
+                next => makeVeeamRequest({
+                    method: 'GET',
+                    bucket: TEST_BUCKET,
+                    objectKey: '.system-d26a9498-cb7c-4a87-a44a-8ae204f5ba6c/capacity.xml',
+                    headers: {
+                        'x-scal-canonical-id': testArn,
+                    },
+                    authCredentials: veeamAuthCredentials,
+                }, (err, response) => {
+                    if (err) {
+                        return done(err);
+                    }
+                    // Critical assertion: for 404 from scubaclient (no metrics yet),
+                    // should return 200 with static capacity data (Used=0)
+                    assert.strictEqual(response.statusCode, 200,
+                        'should return 200 when scubaclient returns 404 (no metrics available)');
+                    // Should return capacity.xml with static data
+                    assert(response.body.includes('<CapacityInfo>'),
+                        'should return capacity.xml content');
+                    assert(response.body.includes('<Used>0</Used>'),
+                        'Used should be 0 from static bucket metadata');
+                    return next();
+                }),
+            ], err => {
+                assert.ifError(err);
+                return done();
+            });
+        });
+
+        it('GET system.xml should return 200 even when scubaclient is down', done => {
+            // system.xml doesn't use scubaclient, so it should always work
+            // This test stops scuba to verify system.xml is independent of utilization metrics
+            async.waterfall([
+                next => makeVeeamRequest({
+                    method: 'PUT',
+                    bucket: TEST_BUCKET,
+                    objectKey: '.system-d26a9498-cb7c-4a87-a44a-8ae204f5ba6c/system.xml',
+                    headers: {
+                        'content-length': testSystem.length,
+                        'content-md5': testSystemMd5,
+                        'x-scal-canonical-id': testArn,
+                    },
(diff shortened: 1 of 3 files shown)


tests/sur/routeVeeam.js
```diff
@@ -383,6 +383,111 @@ function makeVeeamRequest(params, callback) {
                 return done();
             }));
         });
+
+        it('GET capacity.xml should return 200 when scubaclient returns 404 (post-install scenario)', done => {
+            // This test simulates the post-install scenario where scubaclient returns 404
+            // because no metrics are available yet. By not calling scuba.incrementBytesForBucket,
+            // the mock scuba server will return 404 for this bucket.
```
DarkIsDude, line 390: Should we do something to make sure it's the case ? My point is that if someone change the mock server, this test will continue to pass 🤷

tests/unit/routes/veeam-routes.js
```diff
@@ -0,0 +1,569 @@
...
+            assert(!utilizationStub.called, 'should not call UtilizationService for system.xml');
+            assert(response.writeHead.calledWith(200), 'should return 200 for system.xml');
+            assert(response.end.called, 'response should be ended');
+            done();
+        });
+    });
+
+    it('should verify the post-install scenario: 404 returns 200 with Used=0', done => {
+        // This test reproduces the post-install scenario where scubaclient returns 404
+        // because no metrics are available yet
```
DarkIsDude, line 216: I'm not sure it's very pertinent to put it here. You describe this case already in the `lib`. If we want to know why we have this case, we can check the code already.

tests/unit/routes/veeam-routes.js
```diff
@@ -0,0 +1,569 @@
...
+            // Verify we logged a warning
+            assert(logWarnSpy.calledOnce, 'should log warning for 404');
+            
+            // The critical assertion: for 404, we should return 200 with static capacity data
+            assert(response.writeHead.calledWith(200),
+                'should return 200 with static capacity data for 404');
+            
+            assert(response.end.called, 'response should be ended');
+            
+            // Verify it's specifically handling 404 gracefully
```
DarkIsDude:
```suggestion
            assert(logWarnSpy.calledOnce, 'should log warning for 404');
            assert(response.writeHead.calledWith(200),
                'should return 200 with static capacity data for 404');
            assert(response.end.called, 'response should be ended');
```

tests/unit/routes/veeam-routes.js
```diff
@@ -0,0 +1,569 @@
...
+        const error404 = new Error('Not Found');
+        error404.response = { status: 404 };
+        utilizationStub.callsArgWith(4, error404);
+
+        const request = createRequest();
+        const response = createResponse();
+
+        getVeeamFile(request, response, bucketMd, log);
+
+        // Give async callback time to execute
```
DarkIsDude: As you use this logic in each test, we can create a function `const giveAsyncCallbackTimeToExecute = setImmediate` and the in the test

```
giveAsyncCallbackTimeToExecute(() => {
  ...
});
```

Like that your logic will be clean in each tests

tests/unit/routes/veeam-routes.js
```diff
@@ -0,0 +1,569 @@
...
+        error500.response = { status: 500 };
+        utilizationStub.callsArgWith(4, error500);
+
+        const request = createRequest();
+        const response = createResponse();
+
+        getVeeamFile(request, response, bucketMd, log);
+
+        setImmediate(() => {
+            // For 500 errors, we should return error to client
```
DarkIsDude:
```suggestion
```

You already mention that in the `it`

tests/unit/routes/veeam-routes.js
```diff
@@ -0,0 +1,569 @@
...
+        errorConn.code = 'ECONNREFUSED';
+        utilizationStub.callsArgWith(4, errorConn);
+
+        const request = createRequest();
+        const response = createResponse();
+
+        getVeeamFile(request, response, bucketMd, log);
+
+        setImmediate(() => {
+            // For connection errors, we should return error to client
```
DarkIsDude: Same

tests/unit/routes/veeam-routes.js
```diff
@@ -0,0 +1,569 @@
...
+        const metadataError = new Error('Metadata service error');
+        metadataStub.callsArgWith(2, metadataError);
+
+        const request = createRequest();
+        const response = createResponse();
+
+        getVeeamFile(request, response, bucketMd, log);
+
+        setImmediate(() => {
+            // Metadata errors are returned via responseXMLBody
```
DarkIsDude:
```suggestion
```

tests/unit/routes/veeam-routes.js
```diff
@@ -0,0 +1,569 @@
...
+    });
+
+    it('should handle missing bucket metadata', done => {
+        const request = createRequest();
+        const response = createResponse();
+
+        listVeeamFiles(request, response, null, log);
+
+        setImmediate(() => {
+            // Should return NoSuchBucket error
```
DarkIsDude: It's not asserted?

DarkIsDude (approved):

tests/sur/routeVeeam.js
```diff
@@ -383,6 +383,111 @@ function makeVeeamRequest(params, callback) {
...
+                        'Used should be 0 from static bucket metadata');
+                    return next();
+                }),
+            ], err => {
+                assert.ifError(err);
+                return done();
+            });
+        });
+
+        it('GET system.xml should return 200 even when scubaclient is down', done => {
```
SylvainSenechal, line 440: this new test is not directly related to the issue, right ?
It's just testing the logic in veeam/get.js that is not using scuba because of the `!isSystemXML(request)` right ?

SylvainSenechal (approved):
Nothing much to add from Edouard's review, I still very much prefer functional tests over mocks, and I find mocks hard to read, but it's better than nothing 

Nits : some variables could be replaced by constants in veeam-route.js : 'test-bucket', 'getBucket'. 
I also prefer to not use default parameter for the object key in createRequest, but not very important

tests/sur/routeVeeam.js
ghost: Already tracked by this ticket: https://scality.atlassian.net/browse/CLDSRV-528
A bit too early to add scuba yet, we need to drop count items first

tests/sur/routeVeeam.js
ghost: I just added missing tests here yes, not related, it's for the coverage

PR #5974, opened 2025-10-15
💚 CLDSRV-765 Fix ceph flaky tests

DarkIsDude:
# Pull request template

### Motivation and context

### Related issues

https://scality.atlassian.net/browse/CLDSRV-765

### Sign your work

as soon as we can.

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/copyPart.js
+++ b/tests/functional/aws-node-sdk/test/object/copyPart.js
@@ -580,7 +580,7 @@ describe('Object Part Copy', () => {
                 });
             });
 
-            it.only('should not corrupt object if overwriting an existing part by copying a part ' +
+            it('should not corrupt object if overwriting an existing part by copying a part ' +
             'while the MPU is being completed', async () => {
                 const finalObjETag = '"db77ebbae9e9f5a244a26b86193ad818-1"';
                 process.stdout.write('Putting first part in MPU test"');
@@ -621,20 +621,8 @@ describe('Object Part Copy', () => {
                             },
                         }).promise()
                         .catch(async err => {
-                            // The completeMPU delete part in a 2 steps
-                            // (first mark it as deletion with `delete: true` in Mongo, then delete it).
-                            // At the same time, the uploadPartCopy update the same data in Mongo.
-                            // In that case, the completeMPU fail with an InternalError (DeleteConflict in the logs)
                             const raceConditionOccurred = err?.code === 'InternalError'
                                 && err?.message === 'conflict deleting MPU parts metadata';
-
-                            console.info({
-                                raceConditionOccurred,
-                                err,
-                                message: err?.message,
-                                description: err.description,
-                                code: err?.code,
-                            });
 
                             if (raceConditionOccurred) {
                                 return Promise.resolve(null);
--- a/tests/functional/aws-node-sdk/test/object/copyPart.js
+++ b/tests/functional/aws-node-sdk/test/object/copyPart.js
@@ -609,18 +609,43 @@ describe('Object Part Copy', () => {
                     process.stdout.write(
                       'Overwriting first part in MPU test and completing MPU at the same time',
                     );
+
+                    const completeMPU = () => s3
+                      .completeMultipartUpload({
+                        Bucket: destBucketName,
+                        Key: randomDestObjName,
+                        UploadId: uploadId,
+                        MultipartUpload: {
+                          Parts: [{ ETag: etag, PartNumber: 1 }],
+                        },
+                      }).promise();
+
                     const [completeRes, uploadRes] = await Promise.all([
-                      s3
-                        .completeMultipartUpload({
-                          Bucket: destBucketName,
-                          Key: randomDestObjName,
-                          UploadId: uploadId,
-                          MultipartUpload: {
-                            Parts: [{ ETag: etag, PartNumber: 1 }],
-                          },
-                        })
-                        .promise()
-                        .catch(err => {
+                      completeMPU()
+                        .catch(async err => {
+                          // The completeMPU delete part in a 2 steps
+                          // (first mark it as deletion with `delete: true` in Mongo, then delete it).
+                          // At the same time, the uploadPartCopy update the same data in Mongo.
+                          // In that case, the completeMPU fail with an InternalError (DeleteConflict in the logs)
+                          const raceConditionOccurred = err && err.code === 'InternalError';
+
+                          if (raceConditionOccurred) {
+                            process.stdout.write('Retrying the complete MPU');
+
+                            try {
+                              await completeMPU();
+                            } catch (e) {
+                              if (e.code === 'NoSuchUpload') {
+                                process.stdout.write('Race condition done');
+                                return Promise.resolve(null);
+                              }
+
+                              throw e;
+                            }
+                 
+                            throw new Error('Expected error on the retry of complete MPU');
+                          }
+
                           throw err;
                         }),
                       s3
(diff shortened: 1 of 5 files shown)


tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -610,19 +610,41 @@ describe('Object Part Copy', () => {
...
-                          },
-                        })
-                        .promise()
-                        .catch(err => {
-                          console.info('the error is in the completeMultipartUpload');
+                      completeMPU()
+                        .catch(async err => {
+                          // The completeMPU delete part in a 2 steps
+                          // (first mark it as deletion with `delete: true` in Mongo, then delete it).
+                          // At the same time, the uploadPartCopy update the same data in Mongo.
```
francoisferrand: what is described above is the race condition is test is trying to reproduced (i.e. copyPart at the same time as completeMpu), right?

but not clear which other race condition we test here : would be best to name it like below (something like `completeMPUFinishedLater`), to better describe what really happened...

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -610,19 +610,41 @@ describe('Object Part Copy', () => {
...
+                          // In that case, the completeMPU fail with an InternalError (DeleteConflict in the logs)
+                          const raceConditionOccurred = err && err.code === 'InternalError';
+
+                          if (raceConditionOccurred) {
+                            process.stdout.write('Retrying the complete MPU');
+
+                            try {
+                              await completeMPU();
```
francoisferrand: if we issue another compelteMPU, there is a significant change we actually run after every else : so this may still fail (not hitting the race condition we are trying to reproduce) ?

it seems to me that if we try to reproduce a race condition and cannot be "sure" to reproduce, we should make the test statistic: i.e.
* Handle all possible "outcomes", whether it hits the specific race condition we want to reproduce or not (if one outcome is not possible, e.g. both calls succeeded and the object is still there → should fail immediately);
* Ideally be able to precisely detect which "path" we went through (i.e. which case/timinng of the race condition);
* Retry the whole scenario (i.e. both calls!) until we covered all the all such path (with a max number of cases)

or am I missing something?

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -609,18 +609,43 @@ describe('Object Part Copy', () => {
...
+
+                    const completeMPU = () => s3
+                      .completeMultipartUpload({
+                        Bucket: destBucketName,
+                        Key: randomDestObjName,
+                        UploadId: uploadId,
+                        MultipartUpload: {
+                          Parts: [{ ETag: etag, PartNumber: 1 }],
+                        },
+                      }).promise();
```
SylvainSenechal: Side question : will these changes overlap with what Maha is currently doing ? I saw in her PR she removed all the .promise() on the s3 calls

tests/functional/aws-node-sdk/test/object/copyPart.js
```diff
@@ -609,18 +609,43 @@ describe('Object Part Copy', () => {
...
-                        })
-                        .promise()
-                        .catch(err => {
+                      completeMPU()
+                        .catch(async err => {
+                          // The completeMPU delete part in a 2 steps
+                          // (first mark it as deletion with `delete: true` in Mongo, then delete it).
+                          // At the same time, the uploadPartCopy update the same data in Mongo.
+                          // In that case, the completeMPU fail with an InternalError (DeleteConflict in the logs)
+                          const raceConditionOccurred = err && err.code === 'InternalError';
```
SylvainSenechal: yeah be careful, these err.code will become err.name after the sdk migration

BourgoisMickael:
Can we target `development/9.0`

tests/functional/aws-node-sdk/test/object/copyPart.js
DarkIsDude: @SylvainSenechal indeed, with the SDK migration, regarding the order of merge, this will be impacted

tests/functional/aws-node-sdk/test/object/copyPart.js
DarkIsDude: @francoisferrand I'll answer this one too https://github.com/scality/cloudserver/pull/5974#discussion_r2433107324. Let's take the following "processes".

- completeProcess start
- objectPutProcess start
- completeProcess mark the mpu as completing
- completeProcess complete the upload
- completeProcess delete old data: step1 mark data to deletion
- objectPutProcess put the new data
- completeProcess delete old data: step2 delete data
- objectPutProcess check if the complete is in progress, it's the case so don't clean other data
- completeProcess end
- objectPutProcess end

The test is here to test this part of the code : https://github.com/scality/cloudserver/blob/735303b2e13dbfd5ceb3d6fb15273980441305cc/lib/api/objectPutCopyPart.js#L375. So the test is here to test the following case: `objectPutProcess check if the complete is in progress, it's the case so don't clean data`. This one is working pretty fine 🙏.

We have a flaky test because this trigger an another race condition:
- completeProcess delete old data: step1 mark data to deletion
- objectPutProcess put the new data
- completeProcess delete old data: step2 delete data

When the step1 add `delete: true` to mongo, the objectPutProcess remove that (overwrite the object metadata) and the step2 can't handle it correctly (as the number of deleted item is 0). So the completeProcess trigger an internal error https://github.com/scality/cloudserver/blob/development/9.1/lib/api/completeMultipartUpload.js#L542.

After checking with Maha, we considered that the best fix was to add the retry and manage that in the test. To be honest, I'm not sure that in the CI we test the race condition of the test: to be sure that we test it, we should add a spy on the `isCompleteMPUInProgress` function and add some sleep to make sure processes are doing exactly what we want. But we don't want (we can't as CloudServer is running in a dedicated process in the CI?) to add sleep/spy on functional test. So even with statictic test, I'll never be sure that the test is testing what it should. I can do it if you want, but it's more or less the same as catching the error IMO.

We should also consider that this PR don't fix the second race condition. If we want to fix it, we should or consider the Mickael's PR (https://github.com/scality/Arsenal/pull/2497), or change CloudServer to update only some part of the objMd (to don't remove the `delete: true`) or consider this race condition as acceptable (and I think it is).

Sorry for the long answer 😬. Available for a call if you want 🙏

tests/functional/aws-node-sdk/test/object/copyPart.js
ghost: To me this is not an acceptable "fix". The issue is on the API side, not on the test. To me this test effectively tries to test a race condition with 2 possible paths:

- `completeMPU` wins and then uploadPartCopy should fail with `NoSuchKey` or `NoSuchUpload`. Then `completeMPUFinishedEarlier` handle the cleanup automatically.
- `uploadPartCopy` wins. Here upload succeeds, completeMPU succeeds as well, and both API pass.

Here this is a thrid issue that is coming from the API: they clash in the database. `CompleteMPU` start by deleting old data, then uploadPartCopy overwrites the same metadata, then we try to finish deletion, and we cannot find the flag, so we crash with the `InternalError`. But this error could also have many other reasons. The current check is not OK here in that sense.


We do not want to hide bugs by altering our tests, especially when the error here is not something we can discriminate. At the very least we should have a specific error description we would expect here, to ensure this is the right path.

ghost (changes requested):

SylvainSenechal:
I'll let others review. I feel like this change is ok, but I'm not too sure. The test does a double upload part, and checks for a race condition that may or may not occur, and then later we check that the completeMpu is ok. So I'm not exactly sure what we are testing anymore : The completeMpu, or the race condition behavior. It feels like its doing both at the same time.
I feel like this could deserve 2 tests :
A first one with a complete mpu without the race condition, where we upload first, then complete mpu and checks that it works.
A second where we force in a loop that raceCondition to happen, and then checks that things behave the way we want we that race occurs

tmacro (approved):

tests/functional/aws-node-sdk/test/object/copyPart.js
francoisferrand: @williamlardier @DarkIsDude Did you align on an acceptable fix?

The PR has been updated since the comment, with no response to the comment, and the status is not clear: has it simply not yet been reviewed, or does the comment still stand?

DarkIsDude: Yes we did. I'm waiting a review from @williamlardier to confirm it but the PR reflects our discussion on slack 🙏

francoisferrand: @DarkIsDude seems there are conflicts. Are you rebasing/merging after the review, or will you wait for the review to do it?

ghost (approved):

PR #5980, opened 2025-10-22
CLDSRV-750: add Server Access Logs

leif-scality:


francoisferrand:
@nicolas2bert @leif-scality  Is this supposed to land in a patch release on 9.1 ?

lib/Config.js
```diff
@@ -574,6 +574,27 @@ function parseIntegrityChecks(config) {
     return integrityChecks;
 }
 
+function parseServerAccessLogs(config) {
+    const res = {
+        enabled: true,
```
nicolas2bert: Should it be false by default?

config.json
```diff
@@ -149,5 +149,9 @@
     },
     "integrityChecks": {
         "objectPutRetention": true
+    },
+    "serverAccessLogs": {
+        "enabled": true,
```
nicolas2bert: Can we also make sure the API relies on this toggle and returns Not Implemented if enabled: false?

lib/Config.js
```diff
@@ -574,6 +574,27 @@ function parseIntegrityChecks(config) {
...
+        outputFile: './logs/server-access.log',
```
dvasilas:
```suggestion
        outputFile: '/logs/server-access.log',
```

lib/Config.js
dvasilas: Yes, let's set it to false by default so that `access.log` is not created in Artesca.

lib/server.js
```diff
@@ -122,6 +123,20 @@ class S3Server {
...
+            enabled: false,
+            startTime: requestStartTime,
+        };
+        // eslint-disable-next-line no-param-reassign
+        res.serverAccessLog = {};
+
+        res.on('finish', () => {
+            // eslint-disable-next-line no-param-reassign
+            req.serverAccessLog.endTime = process.hrtime.bigint();
+            logServerAccess(req.serverAccessLog, req, res);
```
dvasilas: We log every request, and we end up logging healthcheck requests
```
{"name":"ServerAccessLogger","time":1762877454094,"action":null,"accountName":null,"accountDisplayName":null,"userName":null,"clientPort":46630,"httpMethod":"GET","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":16.621336,"httpURL":"/_/healthcheck/deep","startTime":"14621972622003","requester":null,"operation":"REST.GET.UNKNOWN","requestURI":"GET /_/healthcheck/deep HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"16","turnAroundTime":null,"referer":null,"userAgent":"Wget/1.24.5","versionID":null,"signatureVersion":null,"cipherSuite":null,"authenticationType":null,"hostHeader":"10.160.105.56","tlsVersion":null,"aclRequired":null,"bucketOwner":null,"bucketName":null,"req_id":"7e243b2d294f92a03ca9","bytesSent":null,"clientIP":"::ffff:127.0.0.1","httpCode":200,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":null,"raftSessionID":null,"level":"info","message":"SERVER_ACCESS_LOG","hostname":"dvasilas-L01-store-1","pid":89}
```

One idea to avoid that could be not to write to access log at all for internal routes, using the url to detect them:

```
if (!req.url.startsWith('/_/')) {
      res.on('finish', () => {
          ...
  }
```

That way we wil also filter out requests to the backbeat routes.

dvasilas (commented):
(partial review, but I have some comments that can be addressed)

What about tests?
I think we should have tests that verify that what gets written to `server-access.log` is what is expected, for every API.
It is ok for me to add tests separately in another PR if you prefer, we just need to create the ticket so that we don't forget.

config.json
```diff
@@ -149,5 +149,9 @@
     },
     "integrityChecks": {
         "objectPutRetention": true
+    },
+    "serverAccessLogs": {
+        "enabled": false,
```
dvasilas, line 154: Could we have two types of "enabled"?
1. "Are we writing the `server-access.log`" (default `false` for Artesca). We could call this field `"generateAcessLogFile"`
2. "Is the feature enabled in Federation" (`false` by default, will be configured by Federation). When this is `false`: (1) The API returns `not implemented`, and (2) Cloudserver always sets `loggingEnabled` to `false`. That way, we can disable the feature in Federation.

leif-scality: In what case we would generate the file and not write to it?
can the config file be modified in Federation?

dvasilas: > In what case we would generate the file and not write to it?

For Artesca we should not generate it and not write to it

> can the config file be modified in Federation?

Yes, the `config.json` in this repo is not actually used in Federation.
Federation has a template to generate its own: https://github.com/scality/Federation/blob/development/9.5/roles/run-s3/templates/config.json.j2
( so when bumping cloudserver, we should also update this template)

leif-scality: > For Artesca we should not generate it and not write to it

This is already the case when `enabled = false`

lib/server.js
leif-scality: done

config.json
leif-scality: done

lib/utilities/serverAccessLogger.js
dvasilas, line 1: Typo in the name of the file, there are 3 `s`.

lib/utilities/serverAccesssLogger.js
```diff
@@ -0,0 +1,315 @@
...
+    let remoteIP = null;
+    if (request.headers) {
+        // Check for forwarded IP headers (proxy/load balancer scenarios)
+        const headerRemoteIP = request.headers['x-forwarded-for'] ||
+            request.headers['x-real-ip'] ||
+            request.headers['x-client-ip'] ||
+            request.headers['cf-connecting-ip']; // Cloudflare
+
+        // x-forwarded-for can contain multiple IPs, take the first one
+        if (headerRemoteIP && headerRemoteIP.includes(',')) {
```
dvasilas: What if there is only 1 IP and there is no `,` ?
What about something like
```
if (headerRemoteIP) {
    remoteIP = headerRemoteIP.includes(',')
        ? headerRemoteIP.split(',')[0].trim()
        : headerRemoteIP;
}
```

lib/api/multipartDelete.js
```diff
@@ -55,6 +55,10 @@ function multipartDelete(authInfo, request, log, callback) {
                 log.addDefaultFields({
                     bytesDeleted: partSizeSum,
                 });
+                if (request.serverAccessLogs) {
+                    // eslint-disable-next-line no-param-reassign
+                    request.serverAccessLogs.analyticsBytesDeleted = partSizeSum;
```
dvasilas:
```suggestion
                if (request.serverAccessLog) {
                    // eslint-disable-next-line no-param-reassign
                    request.serverAccessLog.analyticsBytesDeleted = partSizeSum;
```

lib/api/objectDelete.js
```diff
@@ -111,6 +111,10 @@ function objectDeleteInternal(authInfo, request, log, isExpiration, cb) {
                     log.end().addDefaultFields({
                         bytesDeleted: objMD['content-length'],
                     });
+                    if (request.serverAccessLogs) {
+                        // eslint-disable-next-line no-param-reassign
+                        request.serverAccessLogs.analyticsBytesDeleted = objMD['content-length'];
```
dvasilas:
```suggestion
                    if (request.serverAccessLog) {
                        // eslint-disable-next-line no-param-reassign
                        request.serverAccessLog.analyticsBytesDeleted = objMD['content-length'];
```

lib/Config.js
```diff
@@ -574,6 +574,31 @@ function parseIntegrityChecks(config) {
...
 
+function parseServerAccessLogs(config) {
+    const res = {
+        enabled: false,
+        outputFile: '/logs/server-access.log',
+    };
+
+    if (config && config.serverAccessLogs) {
+        if ('enabled' in config.serverAccessLogs) {
+            assert(typeof config.serverAccessLogs.enabled === 'boolean');
```
dvasilas:
```suggestion
            assert(typeof config.serverAccessLogs.enabled === 'boolean', 'bad config: serverAccessLogs.enabled not boolean');
```

lib/Config.js
```diff
@@ -574,6 +574,31 @@ function parseIntegrityChecks(config) {
...
+            res.enabled = config.serverAccessLogs.enabled;
+        }
+
+        if ('outputFile' in config.serverAccessLogs) {
+            assert(typeof config.serverAccessLogs.outputFile === 'string');
```
dvasilas:
```suggestion
            assert(typeof config.serverAccessLogs.outputFile === 'string', 'bad config: serverAccessLogs.outputFile not string');
```

package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.2.37",
+    "arsenal": "git+https://github.com/scality/Arsenal#improvement/ARSN-531-export-server-access-log-fields",
```
dvasilas: Reminder to fix before merging.

lib/server.js
```diff
@@ -122,6 +123,24 @@ class S3Server {
...
+                startTime: requestStartTime,
+            };
+
+            // eslint-disable-next-line no-param-reassign
+            res.serverAccessLog = {};
+
+            res.on('finish', () => {
+                // eslint-disable-next-line no-param-reassign
+                req.serverAccessLog.endTime = process.hrtime.bigint();
+                logServerAccess(req.serverAccessLog, req, res);
```
dvasilas: The first argument is a bit redundant, you could change the function to `logServerAccess(req, res)`.

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,330 @@
...
+
+    serverAccessLogger.info('SERVER_ACCESS_LOG', {
+        // Analytics
+        action: params.analyticsAction || null,
+        accountName: params.analyticsAccountName || null,
+        accountDisplayName: authInfo ? authInfo.getAccountDisplayName() : null,
+        userName: params.analyticsUserName || null,
+        clientPort: req.socket.remotePort || null,
+        httpMethod: req.method || null,
+        bytesDeleted: params.analyticsBytesDeleted || null,
```
dvasilas: Not sure if adding the `analytics` prefix to some fields is useful.
The situation in which it would make sense to have it is if we had a field with the same name and we wanted to distinguish between analytics and server access (`analyticsAction` and `serverAccessAction`).
Do you think we can remove it?

lib/utilities/serverAccesssLogger.js
```diff
@@ -0,0 +1,330 @@
...
+}
+
+function logServerAccess(params, req, res) {
+    const errorCode = res.serverAccessLog.errorCode;
+    const endTurnAroundTime = res.serverAccessLog.endTurnAroundTime;
+    const requestID = res.serverAccessLog.requestID;
+    const bytesSent = res.serverAccessLog.bytesSent;
+    const authInfo = params.authInfo;
+
+    serverAccessLogger.info('SERVER_ACCESS_LOG', {
```
dvasilas: I think we don't use this message; we could send an empty message instead to reduce a bit the size of each log
```suggestion
    serverAccessLogger.info('', {
```

lib/metadata/metadataUtils.js
```diff
@@ -316,8 +319,8 @@ function standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log,
  */
 function standardMetadataValidateBucket(params, actionImplicitDenies, log, callback) {
     const { bucketName } = params;
-    return metadata.getBucket(bucketName, log, (err, bucket) => {
-        storeServerAccessLogInfo(params.request, params.authInfo, bucket);
+    return metadata.getBucket(bucketName, log, (err, bucket, raftSessionId) => { // Extract raft session id
```
dvasilas:
```suggestion
    return metadata.getBucket(bucketName, log, (err, bucket, raftSessionId) => {
```

leif-scality: fixed

lib/utilities/serverAccesssLogger.js
leif-scality: removed

lib/server.js
leif-scality: removed

lib/Config.js
leif-scality: done

lib/Config.js
leif-scality: done

lib/api/objectDelete.js
leif-scality: done

lib/api/multipartDelete.js
leif-scality: done

lib/utilities/serverAccesssLogger.js
leif-scality: done

lib/utilities/serverAccessLogger.js
leif-scality: fixed

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,332 @@
+const { Werelogs } = require('werelogs');
```
nicolas2bert: In general I prefer the setup logic to be handled at the deployment level. Meaning at the application level, if dir OR file missing and serverAccessLogs is enabled, we disable server access logging or to give more visibility we can decide to stop the process since both features bucket access log and analytics will be broken.

It has a clear, explicit behavior: if the admin wants access logs, they must: create the directory, create the file, set serverAccessLogs.enabled = true

This avoids implicit side effects like silently creating a directory, silently creating a file, silently logging somewhere unexpected.
I prefer our admin/deployer to have full control.

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,332 @@
...
+    const outputFile = config.serverAccessLogs.outputFile || DEFAULT_OUTPUT_FILE;
+    const logDir = path.dirname(outputFile);
+
+    try {
+        if (!fs.existsSync(logDir)) {
+            fs.mkdirSync(logDir, { recursive: true });
+        }
+    } catch (error) {
+        // Fall back to logger-only logging if directory creation fails
+        logger.warn('Failed to create ServerAccess log directory, falling back to console logging:', error.message);
```
nicolas2bert: why having this fallback option? Don't the logs get mixed with application logs?

lib/metadata/metadataUtils.js
```diff
@@ -10,6 +10,28 @@ const { onlyOwnerAllowed } = require('../../constants');
...
+
+    request.serverAccessLog.raftSessionID = raftSessionId;
+
+    if (bucket) {
+        request.serverAccessLog.bucketOwner = bucket.getOwner();
+        request.serverAccessLog.bucketName = bucket.getName();
+    }
+
+    if (bucket && bucket.getBucketLoggingStatus() && bucket.getBucketLoggingStatus().getLoggingEnabled()) {
+        request.serverAccessLog.enabled = true;
```
nicolas2bert, line 28: what is the difference between those 2 booleans
- `request.serverAccessLog.enabled` and 
- `request.serverAccessLog.loggingEnabled`?

nicolas2bert: Looking at it more, the naming seems confusing.
Should we call `request.serverAccessLog.enabled`, `request.serverAccessLog.bucketLoggingEnabled`
and 
`request.serverAccessLog.loggingEnabled` called `request.serverAccessLog.BucketLoggingInfos` or similar since it is not a boolean.

lib/server.js
```diff
@@ -122,6 +123,24 @@ class S3Server {
         monitoringClient.httpActiveRequests.inc();
         const requestStartTime = process.hrtime.bigint();
 
+        // Skip server access logs for heartbeat and backbeat.
+        if (!req.url.startsWith('/_/')) {
```
nicolas2bert: Should we also skip if `!config.serverAccessLogs || !config.serverAccessLogs.enabled`?
That way we don't need to compute anything if serverAccessLog is disabled.

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,332 @@
...
+function calculateTurnAroundTime(startTurnAroundTime, endTurnAroundTime) {
+    if (!startTurnAroundTime || !endTurnAroundTime) {
+        return null;
+    }
+
+    return ((endTurnAroundTime - startTurnAroundTime) / 1_000_000n).toString();
+}
+
+function logServerAccess(req, res) {
+    const params = req.serverAccessLog;
```
nicolas2bert, line 366: Lets make sure that `req.serverAccessLog` and `res.serverAccessLog` and `serverAccessLogger` are not `undefined`

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,332 @@
...
+    const apiWerelogs = new Werelogs({
+        level: config.serverAccessLogs.logLevel || 'info',
+        dump: config.serverAccessLogs.dumpLevel || 'error',
+        streams: [{ level: 'trace', stream: serverAccessLogStream }]
+    });
+    logger.info('ServerAccessLogger created successfully');
+    return new apiWerelogs.Logger('ServerAccessLogger');
+}
+
+var serverAccessLogger = {
```
nicolas2bert: var serverAccessLogger;

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,332 @@
...
+const path = require('path');
+const logger = require('./logger');
+
+const DEFAULT_OUTPUT_FILE = '/logs/api-operations.log';
+const SERVER_ACCESS_LOG_FORMAT_VERSION = '0';
+
+function createServerAccessLogger() {
+    if (!config.serverAccessLogs || !config.serverAccessLogs.enabled) {
+        logger.warn('ServerAccessLogs disabled returning no-op logger');
+        return {
```
nicolas2bert: return;

Also change the warn logging to `ServerAccessLogs disabled`

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,332 @@
+const { Werelogs } = require('werelogs');
+const { config } = require('../Config');
```
nicolas2bert: We should test functions in this file lib/utilities/serverAccessLogger.js

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,332 @@
...
+const fs = require('fs');
+const path = require('path');
+const logger = require('./logger');
+
+const DEFAULT_OUTPUT_FILE = '/logs/api-operations.log';
+const SERVER_ACCESS_LOG_FORMAT_VERSION = '0';
+
+function createServerAccessLogger() {
+    if (!config.serverAccessLogs || !config.serverAccessLogs.enabled) {
```
nicolas2bert: Base on this check I assume that if config.serverAccessLogs is responsable to enable/disable for both the bucket logging and analytics feature.

leif-scality: yes

lib/utilities/serverAccessLogger.js
leif-scality: removed noop logger

lib/utilities/serverAccessLogger.js
leif-scality: removed noop logger

lib/utilities/serverAccessLogger.js
leif-scality: done

lib/utilities/serverAccessLogger.js
leif-scality: removed

lib/utilities/serverAccessLogger.js
leif-scality: I removed the directory creation, but I left the file creation because logrotate may delete, rename, or recreate the file so I think we should handle all those cases

lib/metadata/metadataUtils.js
leif-scality: `request.serverAccessLog.enabled = true` means that bucket logging is enabled and   `request.serverAccessLog.loggingEnabled` is the actual bucket logging configuration, this is how it is called in AWS  https://docs.aws.amazon.com/AmazonS3/latest/API/API_LoggingEnabled.html

lib/server.js
leif-scality: done

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,395 @@
...
+
+        if (!this.stream) {
+            logger.error('ServerAccessLogger: stream unavailable');
+            return;
+        }
+
+        if (!this.stream.write(data)) {
+            logger.warn('ServerAccessLogger: backpressure buffer full');
+            this._waitingDrain = true;
+            this.stream.once('drain', () => { this._waitingDrain = false; });
```
nicolas2bert, line 147: Potential race condition
- if backpressure happens (write() returns false), `_waitingDrain` is set to true.
- if the stream closes before 'drain' fires, the callback that resets _waitingDrain is never called.
-  _reopenStream() creates a new stream, but `_waitingDrain` is still true.
- all future writes are dropped permanently.


I would reset `_waitingDrain` when you clean/reopen

leif-scality: I will  reset _waitingDrain in cleanOldStream

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,395 @@
...
+    if (bytesSent) {
+        return bytesSent;
+    }
+
+    if (!res) {
+        return null;
+    }
+
+    const len = res.getHeader('Content-Length');
+    return len || null;
```
nicolas2bert: what if Content-Length is 0, should we log it as 0 and not null?
A test should have caught this potential bug.

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,395 @@
...
+var hostname = os.hostname();
+
+class ServerAccessLogger {
+    constructor(filename) {
+        this.filename = filename;
+        this.stat = undefined;
+        this._waitingDrain = false;
+        this._reopenStream();
+
+        setInterval(() => { this._checkFileRotated(); }, SERVER_ACCESS_LOGGER_CHECK_FILE_ROTATE_INTERVAL_MS);
```
nicolas2bert: `setInterval` is not cleared, it might not be an issue since we create one ServerAccessLogger instance per process, right?

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,430 @@
...
+                }
+
+                if (this.stream) {
+                    this._cleanOldStream(true);
+                }
+
+                this._stat = stat;
+                this.stream = fs.createWriteStream(this._filename, {
+                    encoding: 'utf-8',
+                    flush: true,
```
nicolas2bert, line 114: Flush or not flush, that is the/my question.
If I understand it correctly `flush: true` improves durability because it fsyncs every write.
However, this also makes the internal buffer drain more slowly, which increases the chance of backpressure (write() returning false) and can cause some log entries to be skipped.

leif-scality: We don't flush after every write, only before close 
https://nodejs.org/api/fs.html#fscreatewritestreampath-options

lib/utilities/serverAccessLogger.js
leif-scality: I assume the logger will be delete at the end of the process so no need to clear it. But if this stop being the case yes we should clean it

nicolas2bert (approved):

lib/metadata/metadataUtils.js
dvasilas: I don't think we should set `request.serverAccessLog.enabled = true` based on the configuration of a specific bucket.
( we should only set `request.serverAccessLog.loggingEnabled` here )

`request.serverAccessLog.enabled` is the global "is the feature enabled?" and should be set only by the cloudserver config.

It is possible that we globally enable the feature -> set logging enable for some buckets -> globally disable the feature. In that case we do not want to write to the access logs file.

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,430 @@
...
+        objectKey: params.objectKey || null,    // AWS "Key" field
+
+        // Scality server access logs extra fields
+        logFormatVersion: SERVER_ACCESS_LOG_FORMAT_VERSION,
+        loggingEnabled: params.enabled,
+        loggingTargetBucket: params.loggingEnabled ? params.loggingEnabled.TargetBucket : null,
+        loggingTargetPrefix: params.loggingEnabled ? params.loggingEnabled.TargetPrefix : null,
+        awsAccessKeyID: authInfo ? authInfo.getAccessKey() : null,
+        raftSessionID: params.raftSessionID || null,
+    })  }\n`);
```
dvasilas:
```suggestion
    }) }\n`);
```

lib/api/bucketDeleteCors.js
```diff
@@ -20,62 +17,40 @@ const requestType = 'bucketDeleteCors';
...
-            pushMetric('deleteBucketCors', log, {
-                authInfo,
-                bucket: bucketName,
-            });
-            monitoring.promMetrics(
-                'DELETE', bucketName, '204', 'deleteBucketCors');
-            return callback(err, corsHeaders);
+            pushMetric(METRICS_ACTION, log, { authInfo, bucket: bucketName });
+            monitoring.promMetrics('DELETE', bucketName, '204', METRICS_ACTION);
+            return callback(null , corsHeaders);
```
dvasilas:
```suggestion
            return callback(null, corsHeaders);
```

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,430 @@
+const { config } = require('../Config');
+const fs = require('fs');
+const os = require('os');
+const logger = require('./logger');
+
+const SERVER_ACCESS_LOG_FORMAT_VERSION = '0';
+
+var serverAccessLogger = undefined;
+var hostname = os.hostname();
```
dvasilas: I think you can use `const`.

leif-scality: done

lib/utilities/serverAccessLogger.js
leif-scality: done

dvasilas (commented):
I think only this comment remains for me: https://github.com/scality/cloudserver/pull/5980/files#r2545813184

lib/utilities/serverAccessLogger.js
```diff
@@ -0,0 +1,430 @@
...
+        // eslint-disable-next-line camelcase
+        req_id: requestID || null,              // AWS "Request ID" field
+        bytesSent: getBytesSent(res, bytesSent),
+        clientIP: getRemoteIPFromRequest(req),  // AWS 'Remote IP' field
+        httpCode: res.statusCode || null,       // AWS "HTTP Status" field
+        objectKey: params.objectKey || null,    // AWS "Key" field
+
+        // Scality server access logs extra fields
+        logFormatVersion: SERVER_ACCESS_LOG_FORMAT_VERSION,
+        loggingEnabled: params.enabled,
```
dvasilas, line 427:
```suggestion
        loggingEnabled: params.loggingEnabled,
```

leif-scality: loggingEnabled is already https://docs.aws.amazon.com/AmazonS3/latest/API/API_LoggingEnabled.html

dvasilas: Oh I see, I misunderstood. It's ok I think.

dvasilas (approved):

leif-scality:
ping

leif-scality:
ping

leif-scality:
ping

leif-scality:
ping

leif-scality:
ping

leif-scality:
ping

PR #5991, opened 2025-10-29
CLDSRV-770: parseRateLimitConfig unit tests

tmacro:
Adds unit tests for the config loading code already in the feature branch.
Small deviation from the design to note: `burstCapacity` is optionally configurable at the cluster level rather than being a constant.
I thought this change was easy enough and would help us out when trying to determine the default value during testing.

### DIFF
--- a/tests/unit/api/apiUtils/rateLimit/config.js
+++ b/tests/unit/api/apiUtils/rateLimit/config.js
@@ -1,0 +1,84 @@
+const assert = require('assert');
+
+const { parseRateLimitConfig } = require('../../../../../lib/api/apiUtils/rateLimit/config');
+
+describe('test parseRateLimitConfig', () => {
+    const testCases = [
+        {
+            desc: 'should return an empty config if given one',
+            input: {},
+            expected: {},
+        },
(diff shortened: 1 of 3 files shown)


tcarmet (approved):

tests/unit/api/apiUtils/rateLimit/config.js
```diff
@@ -0,0 +1,84 @@
...
+                    bucketSize: 5000,
+                }
+            },
+        },
+        {
+            desc: '[rps] should throw an error if requestsPerSecond isn\'t an object',
+            input: {
+                requestsPerSecond: 'foo'
+            },
+            throws: true,
```
BourgoisMickael, line 46: Can you be more explicit on the error thrown ?

BourgoisMickael (approved):

tests/unit/api/apiUtils/rateLimit/config.js
tmacro: I'm not really sure how without tying it pretty closely to the implementation.

PR #5992, opened 2025-10-30
CLDSRV-773: Add originOp for empty delete markers

BourgoisMickael:
Delete markers with no previous versions are missing the originOp.

This will now include the originOp all the time and cover 3 cases:
- not a delete marker
- delete marker on top of existing object
- delete marker for a non existing object (NEW)


nicolas2bert (approved):
Test will be e2e in integration?

dvasilas (approved):

BourgoisMickael:
> Test will be e2e in integration?

No specific tests for this.

Scuba has this PR https://github.com/scality/scuba/pull/342 to handle this.

This might fix a bucket notification edge case that might need to be tested in Integration for that.

PR #5996, opened 2025-11-10
🥅(CLDSRV-779) Manage error on putBucketReplication properly

DarkIsDude:
# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
--- a/lib/api/apiUtils/bucket/getReplicationConfiguration.js
+++ b/lib/api/apiUtils/bucket/getReplicationConfiguration.js
@@ -9,9 +9,15 @@ function getReplicationConfiguration(xml, log, cb) {
         if (err) {
             return cb(err);
         }
+
         const validator = new ReplicationConfiguration(result, log, config);
         const configErr = validator.parseConfiguration();
-        return cb(configErr || null, validator.getReplicationConfiguration());
+
+        if (configErr) {
+            return cb(configErr);
+        }
+
+        return cb(null, validator.getReplicationConfiguration());
     });
 }
 
--- a/tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
+++ b/tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
@@ -377,3 +377,61 @@ describe('aws-node-sdk test putBucketReplication configuration rules', () => {
             return checkError(config, 'MalformedXML', done);
         });
 });
+
+describe('aws-node-sdk test putBucketReplication CORS', () => {
+    let s3;
+    const bucket = 'source-bucket-cors';
+
+    beforeEach(done => {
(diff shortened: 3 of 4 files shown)


lib/api/apiUtils/bucket/getReplicationConfiguration.js
SylvainSenechal, line 1: Wait, this change is related to your other Arsenal pr, right ? Does it needs an arsenal bump in package json

tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
```diff
@@ -377,3 +377,61 @@ describe('aws-node-sdk test putBucketReplication configuration rules', () => {
             return checkError(config, 'MalformedXML', done);
         });
 });
+
+describe('aws-node-sdk test putBucketReplication CORS', () => {
```
SylvainSenechal, line 381: Are these new tests related to your changes 🤔 
You changed getReplicationConfiguration, but the tests aren't calling this function

DarkIsDude: Yes and no. This PR resolve a 500 errors that we have. Changes in the other PR fix an another issue. If Arsenal is merged after this one, I'll bump arsenal in an another PR to have this change as soon as possible 🙏

SylvainSenechal (approved):

tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
```diff
@@ -377,3 +377,61 @@ describe('aws-node-sdk test putBucketReplication configuration rules', () => {
...
+                    'arn:aws:iam::account-id:role/dest-resource',
+                Rules: [],
+            },
+        };
+        const request = s3.putBucketReplication(replicationParams);
+        request.on('build', () => {
+            request.httpRequest.headers.Origin = 'http://example.com';
+        });
+        request.send(err => {
+            assertError(err, 'MalformedXML');
```
SylvainSenechal, line 433: Should you add some asserts on err.response.headers ? I ask because the test is called "should return cors headers on malformed xml",  but it feels more like a "should return malformedXML on malformed replication configuration" 🤔

.editorconfig
francoisferrand, line 1: did you really mean to add this file?

(we don't generally have these files, maybe worth discussing before adding them here - so we also add them elsewhere... somehow also related to the discussion/analysis on coding style https://scality.atlassian.net/browse/OS-1014)

tests/functional/aws-node-sdk/test/bucket/putBucketReplication.js
francoisferrand: I concur : would be nice to also have a unit test calling `getReplicationConfiguration()` in case of error...

francoisferrand (approved):

.editorconfig
DarkIsDude: Yes, my goal was to add it. It's standard file that respect our repo (like eslint config, prettier, gitignore, ...). I'll add a comment to align all repo 🙏

PR #5997, opened 2025-11-12
CLDSRV-759: bucketGetRateLimit

tmacro:


tcarmet (approved):

anurag4DSB (approved):

tests/functional/aws-node-sdk/test/rateLimit/tooling.js
```diff
@@ -0,0 +1,81 @@
...
+    );
+    signer.addAuthorization(awsCredentials, new Date());
+
+    const url = `http://${host}${path}`;
+    const options = {
+        method: request.method,
+        headers: request.headers,
+    };
+
+    if (method !== 'GET' && method !== 'DELETE') {
```
jonathan-gramain, line 38: It seems a bit odd to attach a body for all other types of requests (i.e. to have a negative check), I'd rather check `if (method === 'PUT' || method === 'POST')` instead.

tests/functional/aws-node-sdk/test/rateLimit/tooling.js
```diff
@@ -0,0 +1,81 @@
...
+
+        if (xmlResult && xmlResult.Error) {
+            throw xmlResult;
+        }
+
+        // If XML parsing failed or no Error in XML, try JSON
+        try {
+            const json = JSON.parse(text);
+            if (json.error) {
+                throw json;
```
jonathan-gramain, line 63: This `throw` would be caught by the below `catch`, right? Not sure this is what you intended to do.

Also, do you want to throw an error in 100% of cases where` !request.ok` is true? In case it is so, the code doesn't throw in all cases (but maybe you did this on purpose so asking, in which case an extra comment after the `try/catch` block could help understanding this intent).

tests/functional/aws-node-sdk/test/bucket/getBucketRateLimit.js
```diff
@@ -0,0 +1,60 @@
...
+            assert.fail('Expected NoSuchRateLimitConfig error');
+        } catch (err) {
+            assert.strictEqual(err.Error.Code[0], 'NoSuchRateLimitConfig');
+        }
+    });
+
+    it('should return NoSuchBucket error when bucket does not exist', async () => {
+        try {
+            await sendRateLimitRequest('GET', '127.0.0.1:8000', '/nonexistentbucket/?rate-limit');
+        } catch (err) {
```
jonathan-gramain, line 50: Maybe add an `assert.fail` here too like in the above test

jonathan-gramain (approved):

PR #6001, opened 2025-11-18
⬆️(CLDSRV-779) bump arsenal version

DarkIsDude:
Issue: CLDSRV-779


tests/unit/api/bucketDeleteLifecycle.js
```diff
@@ -50,7 +50,7 @@ describe('deleteBucketLifecycle API', () => {
             next => metadata.getBucket(bucketName, log, next),
         ], (err, results) => {
             assert.equal(err, null, `Expected success, got error: ${err}`);
-            const bucket = results[2];
+            const bucket = results[2][0];
```
francoisferrand, line 53: seems weird that only this line is updated : shouldn't there be a change as well in cloudserver code, if API changed in arsenal?

DarkIsDude: Indeed, it's weird. There is a change in Arsenal where a new value was added. Here it's failling because we use it here in the async context with the next directly.

Most of the time in the code we do something like `metadata.getBucket(bucketName, log, (err, md) => {`.

DarkIsDude: @francoisferrand

francoisferrand (approved):

benzekrimaha (approved):

PR #6007, opened 2025-11-18
CLDSRV-778: Add cleanup job for internal memory ratelimit config

anurag4DSB:
Add cleanup jon to avoid memory leaks.

### DIFF
--- a/lib/api/apiUtils/rateLimit/cleanup.js
+++ b/lib/api/apiUtils/rateLimit/cleanup.js
@@ -1,0 +1,62 @@
+const { expireCounters, expireCachedConfigs } = require('./cache');
+const { rateLimitCleanupInterval } = require('../../../../constants');
+
+let cleanupInterval = null;
+
+/**
+ * Start periodic cleanup of expired rate limit counters and cached configs
+ *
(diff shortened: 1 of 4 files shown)


tmacro (approved):

lib/api/apiUtils/rateLimit/cleanup.js
```diff
@@ -0,0 +1,62 @@
...
+    if (!options.skipUnref) {
+        cleanupInterval.unref();
+    }
+}
+
+/**
+ * Stop the periodic cleanup job
+ * Used for graceful shutdown or testing
+ */
+function stopCleanupJob(log) {
```
nicolas2bert, line 49: I don't see it used.

nicolas2bert (approved):

lib/api/apiUtils/rateLimit/cleanup.js
anurag4DSB: We will re-evaluate in https://scality.atlassian.net/browse/CLDSRV-785 and see if we still need it and remove if not needed.

PR #6009, opened 2025-11-18
CLDSRV-783: implement worker sync with Token Reservation System

anurag4DSB:
## Architecture of Token Reservation System

**Token Reservation Flow:**
1. Workers maintain a local token buffer (50 tokens per worker)
2. Background job refills tokens every 100ms (async, non-blocking)
3. Redis atomically grants tokens using GCRA enforcement
4. Request handling consumes tokens from local buffer (in-memory, fast)
5. When tokens run low, worker requests more from Redis

**Key Benefits:**
- **Keeps Redis out of hot path**: Token consumption is pure in-memory
- **Atomic enforcement**: Redis Lua script enforces node-level quota atomically
- **Dynamic work-stealing**: Busy workers automatically get more tokens, idle workers get none
- **Finite buffer prevents over-bursting**: 50-token limit (5 seconds @ 10 req/s avg)


lib/api/apiUtils/rateLimit/client.js
```diff
@@ -5,12 +5,22 @@ const Redis = require('ioredis');
 const { config } = require('../../../Config');
 
 const updateCounterScript = fs.readFileSync(`${__dirname  }/updateCounter.lua`).toString();
+const reconcileCounterScript = fs.readFileSync(`${__dirname  }/reconcileCounter.lua`).toString();
```
anurag4DSB: this was removed in the latest commit

lib/api/apiUtils/rateLimit/client.js
```diff
@@ -65,6 +76,50 @@ class RateLimitClient {
...
+     * Used by token reservation system to request capacity in advance.
+     *
+     * @param {string} bucketName - Bucket name
+     * @param {number} requested - Number of tokens requested
+     * @param {number} interval - Interval per request in ms
+     * @param {number} burstCapacity - Burst capacity in ms
+     * @param {RateLimitClient~grantTokens} cb - Callback
+     */
+    grantTokens(bucketName, requested, interval, burstCapacity, cb) {
+        const key = `throttling:bucket:${bucketName}:rps`;
```
jonathan-gramain: Since the client is named RateLimitClient, would it be better to prefix keys with `ratelimit:`? I also find it more precise.

Also not sure I understand the suffix `rps`, as the key seems to store a timestamp, and it sounds like "requests per second" or similar. If it stores a timestamp, I'd rather suffix it for what it represents (`:emptyAt`?).

lib/api/apiUtils/rateLimit/grantTokens.lua
```diff
@@ -0,0 +1,64 @@
...
+-- Check if request fits within burst capacity
+-- Allow if: expectedTime + cost <= arrivedAt + burstCapacity
+local allowAt = expectedTime + cost
+local burstLimit = arrivedAt + burstCapacity
+
+if allowAt <= burstLimit then
+    -- Full request allowed
+    local newEmptyAt = expectedTime + cost
+    redis.call('SET', key, newEmptyAt)
+    redis.call('PEXPIRE', key, burstCapacity + 10000) -- TTL = burst + 10s buffer
```
jonathan-gramain: You could normally set the expiration time using a single SET command:

```suggestion
    redis.call('SET', key, newEmptyAt, 'PX', burstCapacity + 10000) -- TTL = burst + 10s buffer
```

lib/api/apiUtils/rateLimit/grantTokens.lua
```diff
@@ -0,0 +1,64 @@
...
+local cost = requested * interval
+
+-- Check if request fits within burst capacity
+-- Allow if: expectedTime + cost <= arrivedAt + burstCapacity
+local allowAt = expectedTime + cost
+local burstLimit = arrivedAt + burstCapacity
+
+if allowAt <= burstLimit then
+    -- Full request allowed
+    local newEmptyAt = expectedTime + cost
```
jonathan-gramain: This is already computed as `allowAt`, you can probably reuse the same variable

lib/api/apiUtils/rateLimit/grantTokens.lua
```diff
@@ -0,0 +1,64 @@
...
+
+    if availableCapacity > 0 then
+        -- Grant partial tokens
+        local granted = math.floor(availableCapacity / interval)
+
+        if granted > 0 then
+            local actualCost = granted * interval
+            local newEmptyAt = expectedTime + actualCost
+            redis.call('SET', key, newEmptyAt)
+            redis.call('PEXPIRE', key, burstCapacity + 10000)
```
jonathan-gramain: same comment about merging the two commands into one

lib/api/apiUtils/rateLimit/grantTokens.lua
```diff
@@ -0,0 +1,64 @@
...
+    -- available = (arrivedAt + burstCapacity - expectedTime) / interval
+    local availableCapacity = burstLimit - expectedTime
+
+    if availableCapacity > 0 then
+        -- Grant partial tokens
+        local granted = math.floor(availableCapacity / interval)
+
+        if granted > 0 then
+            local actualCost = granted * interval
+            local newEmptyAt = expectedTime + actualCost
```
jonathan-gramain: nitpick: a more fitting name may be `grantedCost` instead of `actualCost`

lib/api/apiUtils/rateLimit/grantTokens.lua
```diff
@@ -0,0 +1,64 @@
...
+
+    return requested
+else
+    -- Request exceeds capacity, grant partial tokens
+    -- available = (arrivedAt + burstCapacity - expectedTime) / interval
+    local availableCapacity = burstLimit - expectedTime
+
+    if availableCapacity > 0 then
+        -- Grant partial tokens
+        local granted = math.floor(availableCapacity / interval)
```
jonathan-gramain: nitpick: naming precision suggestion `granted` -> `grantedTokens`

lib/api/apiUtils/rateLimit/refillJob.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+async function refillTokenBuckets() {
+    const tokenBuckets = getAllTokenBuckets();
+
+    if (tokenBuckets.size === 0) {
+        return {
+            checked: 0,
+            refilled: 0,
+        };
+    }
```
jonathan-gramain: It may not be needed, unless it's an optimization to avoid calling `Promise.all` (not sure it's worth it, but fine with it)

lib/api/apiUtils/rateLimit/refillJob.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+async function startRefillJob() {
+    logger.info('Starting token refill job', {
+        refillIntervalMs: REFILL_INTERVAL_MS,
+        cleanupIntervalMs: CLEANUP_INTERVAL_MS,
+    });
+
+    const tick = async () => {
+        try {
+            const stats = await refillTokenBuckets();
```
jonathan-gramain: We may want to guard against running multiple refill jobs concurrently, in case one takes more than 100ms, at least to avoid snowballing, and maybe to avoid race conditions too.

One way could be to trade the `setInterval` by a `setTimeout` called after the refill completes.

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -0,0 +1,277 @@
...
+
+            // Request tokens from Redis (atomic GCRA enforcement)
+            const granted = await new Promise((resolve, reject) => {
+                redisClient.grantTokens(
+                    this.bucketName,
+                    requested,
+                    interval,
+                    burstCapacity,
+                    (err, result) => {
+                        if (err) return reject(err);
```
jonathan-gramain:
```suggestion
                        if (err) {
                            return reject(err);
                        }
```

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -0,0 +1,277 @@
...
+     * Record request timestamp for rate calculation
+     */
+    recordRequest() {
+        const now = Date.now();
+        this.requestTimestamps.push(now);
+
+        // Keep only last 1 second of timestamps
+        const cutoff = now - 1000;
+        while (this.requestTimestamps.length > 0 && this.requestTimestamps[0] < cutoff) {
+            this.requestTimestamps.shift();
```
jonathan-gramain: This operation could be slow-ish, if you have e.g. 1000 requests timestamps already in the array (with a client doing 1000 req/s) - you could instead use a `Set`, which can be iterated in insertion order, so deletions from it would be done in reasonable time, unnoticeable in practice.

jonathan-gramain (commented):
partial review, will finish tomorrow

lib/api/apiUtils/rateLimit/client.js
tmacro: :+1:  `ratelimit` makes sense.
The `rps` suffix is more for namespacing so we can add `bandwidth` etc.

lib/api/apiUtils/rateLimit/tokenBucket.js
tmacro: I took a stab at it, let me know if it's what your thinking about.

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -166,12 +168,21 @@ class WorkerTokenBucket {
...
+            this.requestCounter = 0;
+        }
+
+        this.requestTimestamps.add(now * 1000 + this.requestCounter);
 
         // Keep only last 1 second of timestamps
-        const cutoff = now - 1000;
-        while (this.requestTimestamps.length > 0 && this.requestTimestamps[0] < cutoff) {
-            this.requestTimestamps.shift();
+        const cutoff = now - 1000 * 1000;
```
jonathan-gramain: I think you meant `(now - 1000) * 1000`

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -182,17 +193,16 @@ class WorkerTokenBucket {
      */
     getCurrentRate() {
         const now = Date.now();
-        const cutoff = now - 1000;
+        const cutoff = now - 1000 * 1000;
```
jonathan-gramain: same

jonathan-gramain (commented):
Yes it's what I had in mind! LGTM except for the cutoff computation

lib/api/apiUtils/rateLimit/grantTokens.lua
jonathan-gramain, line 1: You'd want to modify the comment at line 7 to change `throttling` to `ratelimit` as well

lib/api/apiUtils/rateLimit/client.js
jonathan-gramain: The point I wanted to make is more about what the key itself represents compared to the actual suffix used, since I don't understand what `rps` means, it may be ok (if not crystal clear) hence my question. I believe the suffix should describe the type of contents associated to this particular key (as I understand it, it's a timestamp).

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -0,0 +1,291 @@
...
+            // Calculate GCRA parameters
+            const nodes = config.rateLimiting.nodes || 1;
+            const workers = config.clusters || 1;
+            const interval = calculateInterval(this.limitConfig.limit, nodes, workers);
+            const burstCapacitySeconds =
+                config.rateLimiting.bucket?.defaultConfig?.requestsPerSecond?.burstCapacity || 1;
+            const burstCapacity = burstCapacitySeconds * 1000;
+
+            // Request tokens from Redis (atomic GCRA enforcement)
+            const granted = await new Promise((resolve, reject) => {
```
BourgoisMickael: `util.promisify` could be used

lib/api/apiUtils/rateLimit/cache.js
```diff
@@ -63,16 +39,98 @@ function expireCachedConfigs(now) {
     return toRemove.length;
 }
 
+/**
+ * Record a request for load tracking
+ * @param {string} key - Bucket key (e.g., "ratelimit:bucket:mybucket:rps:emptyAt")
+ */
+function recordRequest(key) {
+    const now = Date.now();
+    let timestamps = requestTimestamps.get(key);
```
jonathan-gramain, line 48: You may want to also use a `Set` for this one (consider creating common code for both implementations if they are close enough)

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -0,0 +1,291 @@
...
+                    }
+                );
+            });
+
+            // Add granted tokens to buffer
+            this.tokens += granted;
+            this.refillCount++;
+
+            const duration = Date.now() - startTime;
+            this.lastRefillTime = Date.now();
```
BourgoisMickael:
```suggestion
            this.lastRefillTime = Date.now();
            const duration = this.lastRefillTime - startTime;
```

lib/server.js
```diff
@@ -360,6 +365,13 @@ class S3Server {
             // Start rate limit cleanup job
             if (this.config.rateLimiting?.enabled) {
                 startCleanupJob(log);
+                // Start token refill job for token reservation system
+                startRefillJob().catch(err => {
```
BourgoisMickael: can you pass the logger as parameter here instead of the other module having to instantiate a new logger ?

lib/server.js
```diff
@@ -360,6 +365,13 @@ class S3Server {
             // Start rate limit cleanup job
             if (this.config.rateLimiting?.enabled) {
```
BourgoisMickael: Those jobs should be started before the webservers above are started, to be sure to catch every requests at startup

lib/api/apiUtils/rateLimit/refillJob.js
```diff
@@ -0,0 +1,124 @@
...
+        checked++;
+
+        // Trigger async refill if needed (non-blocking)
+        const promise = bucket.refillIfNeeded().then(() => {
+            // Check if refill actually happened
+            if (bucket.refillCount > 0) {
+                refilled++;
+            }
+        }).catch(err => {
+            logger.error('Token refill error', {
```
BourgoisMickael: can we be more precise on the log message that it comes from `rateLimit.refillJob`

lib/api/apiUtils/rateLimit/refillJob.js
```diff
@@ -0,0 +1,124 @@
...
+
+const logger = new werelogs.Logger('S3');
+
+let refillTimer = null;
+
+// Refill interval in milliseconds (how often to check and refill buckets)
+const REFILL_INTERVAL_MS = 100;
+
+// Cleanup interval for expired buckets (every 10 seconds)
+const CLEANUP_INTERVAL_MS = 10000;
```
BourgoisMickael: Can we have env variable on those so we can override to tweak the values easily to test is other configs can be worth ?

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -0,0 +1,291 @@
...
+class WorkerTokenBucket {
+    constructor(bucketName, limitConfig, log) {
+        this.bucketName = bucketName;
+        this.limitConfig = limitConfig;
+        this.log = log;
+
+        // Token buffer configuration
+        this.tokens = 0;
+        this.bufferSize = 50; // Max tokens to hold (5 seconds @ 10 req/s)
+        this.refillThreshold = 20; // Trigger refill when below this
```
BourgoisMickael: same here can we have env variables to override to be able to easily tweak settings

BourgoisMickael (approved):

lib/api/apiUtils/rateLimit/cache.js
tmacro: It doesn't look like this code is actually called from anywhere so I'll remove it.

tcarmet (approved):

PR #6013, opened 2025-11-21
CLDSRV-750: format server access timestamps

leif-scality:
change timestamp format to `s.ms` expected by clickhouse

### DIFF
--- a/lib/utilities/serverAccessLogger.js
+++ b/lib/utilities/serverAccessLogger.js
@@ -358,6 +358,15 @@ function calculateElapsedMS(startTime, onCloseEndTime) {
     return Number(onCloseEndTime - startTime) / 1_000_000;
 }
 
+function timestampToDateTime643(unixMS) {
+    if (unixMS === undefined || unixMS === null) {
+        return null;
+    }
+
+    // clickhouse DateTime64(3) expects "seconds.milliseconds" (string type).
+    return (unixMS / 1000).toFixed(3);
+}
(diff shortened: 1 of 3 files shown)


fredmnl (approved):
I'm ok with this. 

Also considering whether we could use a human-readable format as well? Such as ISO8601 (with ms, which is not the default). I don't know if CH accepts it as is though.

tcarmet (approved):

dvasilas (commented):
Considering bumping the version in package.json in the PR, so you can release a new cloudserver version once this is merged.

lib/utilities/serverAccessLogger.js
```diff
@@ -372,7 +381,7 @@ function logServerAccess(req, res) {
 
     serverAccessLogger.write(`${JSON.stringify({
         // Werelog fields
-        time: Date.now(),
+        time: Date.now() / 1000,
```
dvasilas: Why not `formatStartTimeUnixMS(Date.now())` ?

leif-scality: Done, and changed the name to `timestampToClickhouseTimestamp`

dvasilas: :+1:

dvasilas (approved):
.

leif-scality:
ping

leif-scality:
ping

PR #6015, opened 2025-11-25
CLDSR-789: bucket logging check source and target bucket are in the same account

leif-scality:
Check that source and destination bucket are in the same account as documented in the AWS docs [Logging requests with server access logging - Amazon Simple Storage Service](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html) 

> Both the source and destination buckets must be in the same AWS Region and owned by the same account

### DIFF
--- a/lib/api/bucketPutLogging.js
+++ b/lib/api/bucketPutLogging.js
@@ -7,6 +7,9 @@ const monitoring = require('../utilities/monitoringHandler');
 const { errorInstances } = require('arsenal');
 const { config } = require('../Config');
 
+const ERROR_MSG_SOURCE_TARGET_BUCKET_OWNER_MISMATCH = 
+    'The owner for the bucket to be logged and the target bucket must be the same.';
+
 function bucketPutLogging(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'bucketPutLogging' });
 
@@ -43,9 +46,14 @@ function bucketPutLogging(authInfo, request, log, callback) {
                 return next(null, bucket);
             }
 
-            return metadata.getBucket(loggingEnabled.TargetBucket, log, err => {
+            return metadata.getBucket(loggingEnabled.TargetBucket, log, (err, targetBucket) => {
                 if (err) {
                     return next(errorInstances.InvalidTargetBucketForLogging.customizeDescription(err.description));
+                }
+
+                if (targetBucket.getOwner() !== bucket.getOwner()) {
+                    return next(errorInstances.InvalidTargetBucketForLogging
+                        .customizeDescription(ERROR_MSG_SOURCE_TARGET_BUCKET_OWNER_MISMATCH));
                 }
 
                 return next(null, bucket);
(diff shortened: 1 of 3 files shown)


leif-scality:
ping

tcarmet (approved):

lib/api/bucketPutLogging.js
```diff
@@ -43,11 +43,15 @@ function bucketPutLogging(authInfo, request, log, callback) {
...
             }
 
-            return metadata.getBucket(loggingEnabled.TargetBucket, log, err => {
+            return metadata.getBucket(loggingEnabled.TargetBucket, log, (err, targetBucket) => {
                 if (err) {
                     return next(errorInstances.InvalidTargetBucketForLogging.customizeDescription(err.description));
                 }
 
+                if (targetBucket.getOwner() !== bucket.getOwner()) {
+                    return next(errorInstances.AccessDenied);
```
BourgoisMickael: I've tried on AWS here is the error it returns instead of a simple AccessDenied

```xml
<Error>
    <Code>InvalidTargetBucketForLogging</Code>
    <Message>The owner for the bucket to be logged and the target bucket must be the same</Message>
    <TargetBucket>mickael-bourgois-test-many-files</TargetBucket>
    <RequestId>74007FWX2254Y8CM</RequestId>
    <HostId>59bqrnW/7k8iuRY155mmj2gPPHnJixF7dpqH7ME3OmFuS87qYxIotn7P6Pd7CvGMxmBd6kQNkNwKYL+uG2Y18fw+Gz5J/qTN</HostId>
</Error>
```

BourgoisMickael: Note that functional tests should be able to run against aws with the env variable `AWS_ON_AIR` using your aws credentials profile instead of a cloudserver to ensure the same behavior

leif-scality: Fixed, and updated functional tests to pass with AWS_ON_AIR

tests/functional/aws-node-sdk/test/bucket/putBucketLogging.js
```diff
@@ -151,6 +153,53 @@ describe('PUT bucket logging', () => {
...
+                    s3.putBucketLogging({
+                        Bucket: bucketName,
+                        BucketLoggingStatus: crossAccountConfig,
+                    }, err => {
+                        assert(err, 'Expected error but found none');
+                        assert.strictEqual(err.code, 'InvalidTargetBucketForLogging');
+                        assert.strictEqual(err.statusCode, 400);
+
+                        // Clean up the other account's bucket
+                        otherAccountBucketUtility.deleteOne(otherAccountTargetBucket)
```
BourgoisMickael: cleanup should ideally always be performed in an `after` hook so it is always executed even if the above assert fails. 

Otherwise on test failure that would leave the other bucket existing.

For such case you can encapsulate the test in a `describe` to handle the other bucket in hooks instead of in the test itself

BourgoisMickael (approved):

tests/functional/aws-node-sdk/test/bucket/putBucketLogging.js
leif-scality: moved the test to a new describe and handle the cleanup of both buckets

PR #6018, opened 2025-11-26
 CLDSRV-792: Split bucket logging API control from writing server-access.logs file

dvasilas:
Adds independent control over bucket logging API availability, separate from server access logging using an enum.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -574,9 +574,16 @@ function parseIntegrityChecks(config) {
     return integrityChecks;
 }
 
+const serverAccessLogsModes = {
+    DISABLED: 'DISABLED',
+    LOG_ONLY: 'LOG_ONLY',
+    ENABLED: 'ENABLED',
+};
+
 function parseServerAccessLogs(config) {
+    const validModes = Object.values(serverAccessLogsModes);
     const res = {
-        enabled: false,
+        mode: serverAccessLogsModes.DISABLED,
         outputFile: '/logs/server-access.log',
         highWaterMarkBytes: 10485760,
         retryReopenDelayMS: 1000,
@@ -585,7 +592,7 @@ function parseServerAccessLogs(config) {
 
     if (config && config.serverAccessLogs) {
         const settings = [
-            { key: 'enabled', type: 'boolean' },
+            { key: 'mode', type: 'string' },
             { key: 'outputFile', type: 'string' },
             { key: 'highWaterMarkBytes', type: 'number' },
             { key: 'retryReopenDelayMS', type: 'number' },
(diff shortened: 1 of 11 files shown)


tcarmet (approved):

leif-scality (changes requested):
I think we should do something like this
```
serverAccessLogs.enabledConfigRoutes: bool // Returns NotImplemented if false
serverAccessLogs.enabled: bool // Doesn't write to the file if false
```
You can disable the logging in `server.js`

BourgoisMickael (commented):
Why would we write the log file without having the put-bucket-logging enabled ?

This is kind of confusing, according to your table, with 2 booleans and 1 combination that's a misconfig, it looks like a tri-state / three-state.

Should you rather have 1 `state` or `mode` variable with an enum value:
- `DISABLED` (or absence of the config)
- `LOG` or `LOG_ONLY`
- `API` or `FULL` or `ENABLED`

dvasilas:
> Why would we write the log file without having the put-bucket-logging enabled ?

Because we use the same file for s3 analytics.

The default state when installing S3C should be
- We write the file so that s3 analytics can use it
- Bucket logging APIs are disabled

Having an enum makes sense indeed, rather than using 2 variables to represent 3 state.
I will go with that.

jonathan-gramain (approved):

leif-scality (approved):
Would have used constants instead of strings, but if not the rest is ok

dvasilas:
> Would have used constants instead of strings.

Done

lib/Config.js
```diff
@@ -574,9 +574,20 @@ function parseIntegrityChecks(config) {
     return integrityChecks;
 }
 
+const serverAccessLogsModes = {
+    DISABLED: 'disabled',
+    LOG_ONLY: 'log-only',
+    ENABLED: 'enabled',
```
BourgoisMickael:
```suggestion
    DISABLED: 'DISABLED',
    LOG_ONLY: 'LOG_ONLY',
    ENABLED: 'ENABLED',
```

I'd prefer we keep same value and key for enums

lib/Config.js
```diff
@@ -574,9 +574,20 @@ function parseIntegrityChecks(config) {
...
+};
+
 function parseServerAccessLogs(config) {
+    const validModes = [
+        serverAccessLogsModes.DISABLED,
+        serverAccessLogsModes.LOG_ONLY,
+        serverAccessLogsModes.ENABLED,
+    ];
```
BourgoisMickael:
```suggestion
    const validModes = Object.values(serverAccessLogsModes);
```

BourgoisMickael (approved):

lib/Config.js
dvasilas: Done.

lib/Config.js
dvasilas: :+1:

BourgoisMickael (approved):

PR #6019, opened 2025-12-02
CLDSRV-794: fix service user auth for bucket rate limit APIs + variable bug in lua scripts + arsenal update

anurag4DSB:
Authors: @tmacro and @anurag4DSB

This pull request introduces improvements to the rate limiting logic, especially for bucket rate limit API actions, and refines how service users are handled in authorization and validation flows. The main changes focus on ensuring rate-limiting checks are skipped for specific API actions and authorized service users, and fixing a variable name bug in the rate limit token grant logic(Lua script).

Per requirements, we should be able to call the bucket rate limit APIs without assuming role (like SUR admin APIs) using the normal credentials of a service user.

Dependency update:

Updated the arsenal dependency to version 8.2.41 in package.json. This update includes the exported bucket rate limit APIs' action map.

### DIFF
--- a/lib/metadata/metadataUtils.js
+++ b/lib/metadata/metadataUtils.js
@@ -5,8 +5,9 @@ const metadata = require('./wrapper');
 const BucketInfo = require('arsenal').models.BucketInfo;
 const { isBucketAuthorized, isObjAuthorized } =
     require('../api/apiUtils/authorization/permissionChecks');
+const { isRateLimitServiceUser } = require('../api/apiUtils/authorization/serviceUser');
 const bucketShield = require('../api/apiUtils/bucket/bucketShield');
-const { onlyOwnerAllowed } = require('../../constants');
+const { onlyOwnerAllowed, rateLimitApiActions } = require('../../constants');
 const { actionNeedQuotaCheck, actionWithDataDeletion } = require('arsenal/build/lib/policyEvaluator/RequestContext');
 const { processBytesToWrite, validateQuotas } = require('../api/apiUtils/quotas/quotaUtils');
 const { config } = require('../Config');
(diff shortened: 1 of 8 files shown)


lib/metadata/metadataUtils.js
```diff
@@ -182,10 +184,21 @@ function validateBucket(bucket, params, log, actionImplicitDenies = {}) {
...
         return errors.NoSuchBucket;
     }
+
     const canonicalID = authInfo.getCanonicalID();
     if (!Array.isArray(requestType)) {
         requestType = [requestType];
     }
+
+    // Skip checking bucket ownership if the requesting user is the rate limit service user
+    // and the requestType is Get/Put/DeleteBucketRateLimit.
```
BourgoisMickael, line 194: Should we also skip other internal usages such as: backbeat assumed role used for lifecycle / replication ? As it should not be blocked even if there are too many user requests.

+ since recently backbeat use another internal port (8003) already to bypass bucket policies

BourgoisMickael (approved):

nicolas2bert (approved):

lib/metadata/metadataUtils.js
anurag4DSB: Thank you so much, great catch.

It will be handled in https://scality.atlassian.net/browse/CLDSRV-797
We will merge this PR as we need it for the preview.

anurag4DSB:
@tmacro feel free to merge this PR after you have a look.

anurag4DSB:
/approve
(I know bert-e is not enabled on feature branch, but commenting for approval of author)

tmacro (approved):

PR #6020, opened 2025-12-02
CLDSRV-780: ServerAccessLogs json schema and functional tests

leif-scality:
- Defined a json schema for the server access logs
- Updated unit and logger output to match the schema
- Implemented functional tests, to check that cloudserver writes logs correctly to the output file

### DIFF
--- a/.github/docker/docker-compose.yaml
+++ b/.github/docker/docker-compose.yaml
@@ -8,6 +8,7 @@ services:
       - ${HOME}/.aws/credentials:/root/.aws/credentials
       - /tmp/artifacts/${JOB_NAME}:/artifacts
       - /tmp/coverage/${JOB_NAME}:/coverage/test
+      - /logs:/logs
     environment:
       - CI=true
       - ENABLE_LOCAL_CACHE=true
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -182,6 +182,7 @@ const api = {
             });
         }
         if (request.serverAccessLog) {
+            request.serverAccessLog.bucketName = request.bucketName;
             request.serverAccessLog.objectKey = request.objectKey;
             request.serverAccessLog.analyticsAction = actionLog;
         }
--- a/lib/server.js
+++ b/lib/server.js
@@ -409,7 +409,8 @@ class S3Server {
                 }
             }
 
-           try {
+            try {
+                logger.info('ServerAccessLogger config', { config: _config.serverAccessLogs });
                 if (_config.serverAccessLogs.mode === serverAccessLogsModes.LOG_ONLY
                     || _config.serverAccessLogs.mode === serverAccessLogsModes.ENABLED) {
                     var serverAccessLogger = new ServerAccessLogger(
--- a/lib/utilities/serverAccessLogger.js
+++ b/lib/utilities/serverAccessLogger.js
@@ -181,78 +181,92 @@ function getRemoteIPFromRequest(request) {
     return remoteIP;
 }
 
-function getOperation(req) {
-    const methodToResType = Object.freeze({
-        'bucketDelete': 'BUCKET',
-        'bucketDeleteCors': 'BUCKET',
-        'bucketDeleteEncryption': 'BUCKET',
-        'bucketDeleteWebsite': 'BUCKET',
-        'bucketGet': 'BUCKET',
-        'bucketGetACL': 'BUCKET',
-        'bucketGetCors': 'BUCKET',
-        'bucketGetObjectLock': 'BUCKET',
-        'bucketGetVersioning': 'VERSIONING',
-        'bucketGetWebsite': 'BUCKET',
-        'bucketGetLocation': 'BUCKET',
-        'bucketGetEncryption': 'BUCKET',
-        'bucketHead': 'BUCKET',
-        'bucketPut': 'BUCKET',
-        'bucketPutACL': 'BUCKET',
-        'bucketPutCors': 'BUCKET',
-        'bucketPutVersioning': 'VERSIONING',
-        'bucketPutTagging': 'BUCKET',
-        'bucketDeleteTagging': 'BUCKET',
-        'bucketGetTagging': 'BUCKET',
-        'bucketPutWebsite': 'BUCKET',
-        'bucketPutReplication': 'BUCKET',
-        'bucketGetReplication': 'BUCKET',
-        'bucketDeleteReplication': 'BUCKET',
-        'bucketDeleteQuota': 'BUCKET',
-        'bucketPutLifecycle': 'BUCKET',
-        'bucketUpdateQuota': 'BUCKET',
-        'bucketGetLifecycle': 'BUCKET',
-        'bucketDeleteLifecycle': 'BUCKET',
-        'bucketPutPolicy': 'BUCKETPOLICY',
-        'bucketGetPolicy': 'BUCKETPOLICY',
-        'bucketGetQuota': 'BUCKET',
-        'bucketDeletePolicy': 'BUCKETPOLICY',
-        'bucketPutObjectLock': 'BUCKET',
-        'bucketPutNotification': 'BUCKET',
-        'bucketGetNotification': 'BUCKET',
-        'bucketPutEncryption': 'BUCKET',
-        'bucketPutLogging': 'LOGGING_STATUS',
-        'bucketGetLogging': 'LOGGING_STATUS',
-        // 'corsPreflight': '',
-        'completeMultipartUpload': 'OBJECT',
-        'initiateMultipartUpload': 'OBJECT',
-        'listMultipartUploads': 'OBJECT',
-        'listParts': 'OBJECT',
-        'metadataSearch': 'OBJECT',
-        'multiObjectDelete': 'OBJECT',
-        'multipartDelete': 'OBJECT',
-        'objectDelete': 'OBJECT',
-        'objectDeleteTagging': 'OBJECT',
-        'objectGet': 'OBJECT',
-        'objectGetACL': 'OBJECT',
-        'objectGetLegalHold': 'OBJECT',
-        'objectGetRetention': 'OBJECT',
-        'objectGetTagging': 'OBJECT',
-        'objectCopy': 'OBJECT',
-        'objectHead': 'OBJECT',
-        'objectPut': 'OBJECT',
-        'objectPutACL': 'OBJECT',
-        'objectPutLegalHold': 'OBJECT',
-        'objectPutTagging': 'OBJECT',
-        'objectPutPart': 'OBJECT',
-        'objectPutCopyPart': 'OBJECT',
-        'objectPutRetention': 'OBJECT',
-        'objectRestore': 'OBJECT',
-        // 'serviceGet': '',
-        // 'websiteGet': '',
-        // 'websiteHead': '',
-    });
+// eslint-disable-next-line max-len
+// https://github.com/awslabs/glue-extensions-for-iceberg/blob/52bdb2908216a85859fd76a45981d0326d016a2f/spark/src/main/scala/software/amazon/glue/s3a/audit/S3LogVerbs.java
+// https://github.com/open-io/swift/blob/ff518e9907f74b5a2565973a260f36386b5d5cbf/etc/s3-default.cfg.in#L78
+// https://stackoverflow.com/questions/42707878/amazon-s3-logs-operation-definition
+const methodToResType = Object.freeze({
+    'bucketDelete': 'BUCKET',
+    'bucketDeleteCors': 'CORS',
+    'bucketDeleteEncryption': 'ENCRYPTION',
+    'bucketDeleteWebsite': 'WEBSITE',
+    'bucketGet': 'BUCKET',
+    'bucketGetACL': 'ACL',
+    'bucketGetCors': 'CORS',
+    'bucketGetObjectLock': 'OBJECT',
+    'bucketGetVersioning': 'VERSIONING',
+    'bucketGetWebsite': 'WEBSITE',
+    'bucketGetLocation': 'LOCATION',
+    'bucketGetEncryption': 'ENCRYPTION',
+    'bucketHead': 'BUCKET',
+    'bucketPut': 'BUCKET',
+    'bucketPutACL': 'ACL',
+    'bucketPutCors': 'CORS',
+    'bucketPutVersioning': 'VERSIONING',
+    'bucketPutTagging': 'TAGGING',
+    'bucketDeleteTagging': 'TAGGING',
+    'bucketGetTagging': 'TAGGING',
+    'bucketPutWebsite': 'WEBSITE',
+    'bucketPutReplication': 'REPLICATION',
+    'bucketGetReplication': 'REPLICATION',
+    'bucketDeleteReplication': 'REPLICATION',
+    'bucketDeleteQuota': 'QUOTA',
+    'bucketPutLifecycle': 'LIFECYCLE',
+    'bucketUpdateQuota': 'QUOTA',
+    'bucketGetLifecycle': 'LIFECYCLE',
+    'bucketDeleteLifecycle': 'LIFECYCLE',
+    'bucketPutPolicy': 'BUCKETPOLICY',
+    'bucketGetPolicy': 'BUCKETPOLICY',
+    'bucketGetQuota': 'QUOTA',
+    'bucketDeletePolicy': 'BUCKETPOLICY',
+    'bucketPutObjectLock': 'OBJECT',
+    'bucketPutNotification': 'NOTIFICATION',
+    'bucketGetNotification': 'NOTIFICATION',
+    'bucketPutEncryption': 'ENCRYPTION',
+    'bucketPutLogging': 'LOGGING_STATUS',
+    'bucketGetLogging': 'LOGGING_STATUS',
+    // 'corsPreflight': '',
+    'completeMultipartUpload': 'UPLOAD',
+    'initiateMultipartUpload': 'UPLOADS',
+    'listMultipartUploads': 'UPLOADS',
+    'listParts': 'UPLOAD',
+    'metadataSearch': 'OBJECT',
+    'multiObjectDelete': 'OBJECT',
+    'multipartDelete': 'UPLOAD',
+    'objectDelete': 'OBJECT',
+    'objectDeleteTagging': 'TAGGING',
+    'objectGet': 'OBJECT',
+    'objectGetACL': 'ACL',
+    'objectGetLegalHold': 'LEGALHOLD',
+    'objectGetRetention': 'OBJECT_LOCK_RETENTION',
+    'objectGetTagging': 'TAGGING',
+    'objectCopy': 'COPY',
+    'objectHead': 'OBJECT',
+    'objectPut': 'OBJECT',
+    'objectPutACL': 'ACL',
+    'objectPutLegalHold': 'LEGALHOLD',
+    'objectPutTagging': 'TAGGING',
+    'objectPutPart': 'PART',
+    'objectPutCopyPart': 'COPY',
+    'objectPutRetention': 'OBJECT_LOCK_RETENTION',
+    'objectRestore': 'OBJECT',
+    'serviceGet': 'SERVICE', // ListBuckets
+    'websiteGet': 'WEBSITE',
+    'websiteHead': 'WEBSITE',
+});
 
-    return `REST.${req.method}.${methodToResType[req.apiMethod] ? methodToResType[req.apiMethod] : 'UNKNOWN'}`;
+function getOperation(req) {
+    const resourceType = methodToResType[req.apiMethod];
+    if (!resourceType) {
+        process.emitWarning('Unknown apiMethod for server access log', {
+            type: 'ServerAccessLogWarning',
+            code: 'UNKNOWN_API_METHOD',
+            detail: `apiMethod=${req.apiMethod}, method=${req.method}, url=${req.url}`
+        });
+        return `REST.${req.method}.UNKNOWN`;
+    }
+
+    return `REST.${req.method}.${resourceType}`;
 }
 
 function getRequester(authInfo) {
@@ -284,33 +298,43 @@ function getURI(request) {
     return requestURI;
 }
 
+const objectSizePutMethods = Object.freeze({
+    'objectPut': true,
+    'objectPutPart': true,
+});
+
+const objectSizeGetMethods = Object.freeze({
+    'objectGet': true,
+});
+
 function getObjectSize(request, response) {
-    const objectSizePutMethods = Object.freeze({
-        'objectPut': true,
-        'objectPutPart': true,
-    });
-
-    const objectSizeGetMethods = Object.freeze({
-        'objectGet': true,
-    });
-
     // If it is a PUT get the Content-Length from the request, if it is a GET get it from the response.
     if (request && response && objectSizeGetMethods[request.apiMethod]) {
         const len = response.getHeader('Content-Length');
-        if (len === undefined || len === null) {
+        if (len === undefined || len === null || len === '') {
             return null;
         }
 
-        return len;
+        const numLen = Number(len);
+        if (isNaN(numLen)) {
+            return null;
+        }
+
+        return numLen;
     }
 
     if (request && objectSizePutMethods[request.apiMethod]) {
         const len = request.headers['content-length'];
-        if (len === undefined || len === null) {
+        if (len === undefined || len === null || len === '') {
             return null;
         }
 
-        return len;
+        const numLen = Number(len);
+        if (isNaN(numLen)) {
+            return null;
+        }
+
+        return numLen;
     }
 
     return null;
@@ -387,50 +411,50 @@ function logServerAccess(req, res) {
         pid: process.pid,
 
         // Analytics
-        action: params.analyticsAction || null,
-        accountName: params.analyticsAccountName || null,
+        action: params.analyticsAction === undefined ? null : params.analyticsAction,
+        accountName: params.analyticsAccountName === undefined ? null : params.analyticsAccountName,
         accountDisplayName: authInfo ? authInfo.getAccountDisplayName() : null,
-        userName: params.analyticsUserName || null,
-        clientPort: req.socket.remotePort || null,
-        httpMethod: req.method || null,
-        bytesDeleted: params.analyticsBytesDeleted || params.analyticsBytesDeleted === 0 ? 0 : null,
-        bytesReceived: req.parsedContentLength || 0,
-        bodyLength: parseInt(req.headers['content-length'], 10) || 0,
-        contentLength: req.parsedContentLength || 0,
+        userName: params.analyticsUserName === undefined ? null : params.analyticsUserName,
+        clientPort: req.socket.remotePort === undefined ? null : req.socket.remotePort,
+        httpMethod: req.method === undefined ? null : req.method,
+        bytesDeleted: params.analyticsBytesDeleted === undefined ? null : params.analyticsBytesDeleted,
+        bytesReceived: req.parsedContentLength === undefined ? null : req.parsedContentLength,
+        bodyLength: req.headers['content-length'] === undefined ? null : parseInt(req.headers['content-length'], 10),
+        contentLength: req.parsedContentLength === undefined ? null : req.parsedContentLength,
         // eslint-disable-next-line camelcase
         elapsed_ms: calculateElapsedMS(params.startTime, params.onCloseEndTime),
-        httpURL: req.url || null,
+        httpURL: req.url === undefined ? null : req.url,
 
         // AWS access server logs fields https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html
         startTime: timestampToDateTime643(params.startTimeUnixMS), // AWS "Time" field
         requester: getRequester(authInfo),
         operation: getOperation(req),
         requestURI: getURI(req),
-        errorCode: errorCode || null,
+        errorCode: errorCode === undefined ? null : errorCode,
         objectSize: getObjectSize(req, res),
         totalTime: calculateTotalTime(params.startTime, params.onFinishEndTime),
         turnAroundTime: calculateTurnAroundTime(params.startTurnAroundTime, endTurnAroundTime),
-        referer: req.headers.referer || null,
-        userAgent: req.headers['user-agent'] || null,
-        versionID: req.query ? req.query.versionId || null : null,
+        referer: req.headers.referer === undefined ? null : req.headers.referer,
+        userAgent: req.headers['user-agent'] === undefined ? null : req.headers['user-agent'],
+        versionID: !req.query ? null : req.query.versionId === undefined ? null : req.query.versionId,
         signatureVersion: authInfo ? authInfo.getAuthVersion() : null,
         cipherSuite: req.socket.encrypted ? req.socket.getCipher()['standardName'] : null,
         authenticationType: authInfo ? authInfo.getAuthType() : null,
-        hostHeader: req.headers.host || null,
+        hostHeader: req.headers.host === undefined ? null : req.headers.host,
         tlsVersion: req.socket.encrypted ? req.socket.getCipher()['version'] : null,
         aclRequired: null,       // TODO: CLDSRV-774
         // hostID: null,         // NOT IMPLEMENTED
         // accessPointARN: null, // NOT IMPLEMENTED
 
         // Shared between AWS access server logs and Analytics logs
-        bucketOwner: params.bucketOwner || null,
-        bucketName: params.bucketName || null,  // AWS "Bucket" field
+        bucketOwner: params.bucketOwner === undefined ? null : params.bucketOwner,
+        bucketName: params.bucketName === undefined ? null : params.bucketName, // AWS "Bucket" field
         // eslint-disable-next-line camelcase
-        req_id: requestID || null,              // AWS "Request ID" field
+        req_id: requestID === undefined ? null : requestID, // AWS "Request ID" field
         bytesSent: getBytesSent(res, bytesSent),
         clientIP: getRemoteIPFromRequest(req),  // AWS 'Remote IP' field
-        httpCode: res.statusCode || null,       // AWS "HTTP Status" field
-        objectKey: params.objectKey || null,    // AWS "Key" field
+        httpCode: res.statusCode === undefined ? null : res.statusCode, // AWS "HTTP Status" field
+        objectKey: params.objectKey === undefined ? null : params.objectKey, // AWS "Key" field
 
         // Scality server access logs extra fields
         logFormatVersion: SERVER_ACCESS_LOG_FORMAT_VERSION,
@@ -438,7 +462,7 @@ function logServerAccess(req, res) {
         loggingTargetBucket: params.loggingEnabled ? params.loggingEnabled.TargetBucket : null,
         loggingTargetPrefix: params.loggingEnabled ? params.loggingEnabled.TargetPrefix : null,
         awsAccessKeyID: authInfo ? authInfo.getAccessKey() : null,
-        raftSessionID: params.raftSessionID || null,
+        raftSessionID: params.raftSessionID === undefined ? null : params.raftSessionID,
     })}\n`);
 }
 
--- a/schema/server_access_log.schema.json
+++ b/schema/server_access_log.schema.json
@@ -1,0 +1,248 @@
+{
+    "$id": "https://github.com/scality/cloudserver/tree/development/9.2/schema/lib/server_access_log.schema.json",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Server Access Log",
+    "description": "A server access log. AWS fields are documented in https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html.",
+    "type": "object",
+    "properties": {
+        "time": {
+            "description": "Epoch timestamp in seconds, recorded when the log record is created.",
+            "type": "integer",
+            "minimum": 0
+        },
+        "hostname": {
+            "description": "Hostname of the worker as returned by os.hostname().",
+            "type": "string"
+        },
+        "pid": {
+            "description": "PID of the worker.",
+            "type": "integer",
+            "minimum": 0
+        },
+        "action": {
+            "description": "S3 API action name.",
+            "type": "string"
+        },
+        "accountName": {
+            "description": "Requester account display name.",
+            "type": ["string", "null"]
+        },
+        "accountDisplayName": {
+            "description": "Requester account display name.",
+            "type": ["string", "null"]
+        },
+        "userName": {
+            "description": "Requester IAM display name, or null if the requester is not an IAM user.",
+            "type": ["string", "null"]
+        },
+        "clientPort": {
+            "description": "Requester remote connection port.",
+            "type": "integer",
+            "maximum": 65535,
+            "minimum": 1
+        },
+        "httpMethod": {
+            "description": "Request HTTP method.",
+            "type": "string",
+            "enum": ["GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE", "POST", "PATCH", "CONNECT"]
+        },
+        "bytesDeleted": {
+            "description": "For DeleteObject: size in bytes of the deleted object, for DeleteObjects: sum of all the deleted objects.",
+            "type": ["integer", "null"],
+            "minimum": 0
+        },
+        "bytesReceived": {
+            "description": "For PutObject and UploadPart: size of the object in bytes.",
+            "type": "integer",
+            "minimum": 0
+        },
+        "bodyLength": {
+            "description": "Size in bytes of the request body, copied from the HTTP Content-Length header.",
+            "type": "integer",
+            "minimum": 0
+        },
+        "contentLength": {
+            "description": "Size in bytes of the request content.",
+            "type": "integer",
+            "minimum": 0
+        },
+        "elapsed_ms": {
+            "description": "Total duration of the request in milliseconds. The timer starts when the server first routes the request and stops when the request completes or is closed prematurely.",
+            "type": "number",
+            "minimum": 0
+        },
+        "httpURL": {
+            "description": "Request URL string. This contains only the URL that is present in the actual HTTP request.",
+            "type": "string"
+        },
+        "startTime": {
+            "description": "Timestamp formatted as: 'seconds.milliseconds', recorded when the server first routes the request. Represents the AWS server access log 'Time' field. String type compatible with Clickhouse DateTime64(3) type.",
+            "type": "string"
+        },
+        "requester": {
+            "description": "AWS server access log 'Requester' field. From AWS 'The canonical user ID of the requester, or a - for unauthenticated requests. If the requester was an IAM user, this field returns the requester's IAM user name along with the AWS account that the IAM user belongs to. This identifier is the same one used for access control purposes.'. We don't use null instead of '-' when the requester is missing.",
+            "type": ["string", "null"]
+        },
+        "operation": {
+            "description": "AWS server access log 'Operation' field. From AWS 'The operation listed here is declared as SOAP.operation, REST.HTTP_method.resource_type, WEBSITE.HTTP_method.resource_type, or BATCH.DELETE.OBJECT, or S3.action.resource_type for S3 Lifecycle and logging. For Compute checksum job requests, the operation is listed as S3.COMPUTE.OBJECT.CHECKSUM.'.",
+            "type": "string"
+        },
+        "requestURI": {
+            "description": "AWS server access log 'Request URI' field. From AWS 'The Request-URI part of the HTTP request message.'.",
+            "type": "string"
+        },
+        "errorCode": {
+            "description": "AWS server access log 'Error Code' field. From AWS 'The Amazon S3 Error responses , or - if no error occurred.'. We use null to signal no error.",
+            "type": ["string", "null"]
+        },
+        "objectSize": {
+            "description": "AWS server access log 'Object Size' field.",
+            "type": ["integer", "null"]
+        },
+        "totalTime": {
+            "description": "AWS server access log 'Total Time' field. From AWS 'The number of milliseconds that the request was in flight from the server's perspective. This value is measured from the time that your request is received to the time that the last byte of the response is sent. Measurements made from the client's perspective might be longer because of network latency.'.",
+            "type": "string"
+        },
+        "turnAroundTime": {
+            "description": "AWS server access log 'Turn Around Time' field. From AWS 'The number of milliseconds that Amazon S3 spent processing your request. This value is measured from the time that the last byte of your request was received until the time that the first byte of the response was sent.'.",
+            "type": ["string", "null"]
+        },
+        "referer": {
+            "description": "AWS server access log 'Referer' field. From AWS 'The value of the HTTP Referer header, if present. HTTP user-agents (for example, browsers) typically set this header to the URL of the linking or embedding page when making a request.'.",
+            "type": ["string", "null"]
+        },
+        "userAgent": {
+            "description": "AWS server access log 'User Agent' field. From AWS 'The value of the HTTP User-Agent header.'.",
+            "type": ["string", "null"]
+        },
+        "versionID": {
+            "description": "AWS server access log 'Version ID' field. From AWS 'The version ID of the object being copied, or - if the x-amz-copy-source header didn't specify a versionId parameter as part of the copy source.'. We use null to signal no versionId.",
+            "type": ["string", "null"]
+        },
+        "signatureVersion": {
+            "description": "AWS server access log 'Signature Version' field. From AWS 'The signature version, SigV2 or SigV4, that was used to authenticate the request, or a - for unauthenticated requests.'. We use null for unauthenticated requests.",
+            "type": ["string", "null"]
+        },
+        "cipherSuite": {
+            "description": "AWS server access log 'Cipher Suite' field. From AWS 'The Transport Layer Security (TLS) cipher that was negotiated for an HTTPS request, or a - for HTTP.'. We use null for HTTP.",
+            "type": ["string", "null"]
+        },
+        "authenticationType": {
+            "description": "AWS server access log 'Authentication Type' field. From AWS 'The type of request authentication used: AuthHeader for authentication headers, QueryString for query strings (presigned URLs), or a - for unauthenticated requests.'. We use null for unauthenticated requests.",
+            "type": ["string", "null"]
+        },
+        "hostHeader": {
+            "description": "AWS server access log 'Host Header' field. From AWS 'The endpoint that was used to connect to Amazon S3.'.",
+            "type": "string"
+        },
+        "tlsVersion": {
+            "description": "AWS server access log 'TLS Version' field. From AWS 'The Transport Layer Security (TLS) version negotiated by the client. The value is one of following: TLSv1.1, TLSv1.2, TLSv1.3, or - if TLS wasn't used.'. We use null if TLS was not used.",
+            "type": ["string", "null"]
+        },
+        "aclRequired": {
+            "description": "AWS server access log 'ACL Required' field.",
+            "type": ["null"]
+        },
+        "bucketOwner": {
+            "description": "AWS server access log 'Bucket Owner' field. From AWS 'The canonical user ID of the owner of the source bucket. The canonical user ID is another form of the AWS account ID'.",
+            "type": ["string", "null"]
+        },
+        "bucketName": {
+            "description": "AWS server access log 'Bucket' field. Null for ListBuckets.",
+            "type": ["string", "null"]
+        },
+        "req_id": {
+            "description": "AWS server access log 'Request ID' field. Matches the req_id field in logs.",
+            "type": "string"
+        },
+        "bytesSent": {
+            "description": "AWS server access log 'Bytes Sent' field. From AWS 'The number of response bytes sent, excluding HTTP protocol overhead, or - if zero.'. We 0 and null are possible values in our implementation.",
+            "type": ["integer", "null"]
+        },
+        "clientIP": {
+            "description": "AWS server access log 'Remote IP' field. From AWS 'The apparent IP address of the requester. Intermediate proxies and firewalls might obscure the actual IP address of the machine that's making the request.'.",
+            "type": "string"
+        },
+        "httpCode": {
+            "description": "AWS server access log 'HTTP Status' field. From AWS 'The numeric HTTP status code of the response.'.",
+            "type": "integer"
+        },
+        "objectKey": {
+            "description": "AWS server access log 'Key' field. From AWS 'The key (object name) part of the request.'.",
+            "type": ["string", "null"]
+        },
+        "logFormatVersion": {
+            "description": "Version of the server access log schema.",
+            "type": ["string"],
+            "const": "0"
+        },
+        "loggingEnabled": {
+            "description": "True if the target bucket has bucket logging configured and enabled.",
+            "type": "boolean"
+        },
+        "loggingTargetBucket": {
+            "description": "Target bucket where server access logs should be uploaded. https://docs.aws.amazon.com/AmazonS3/latest/API/API_LoggingEnabled.html",
+            "type": ["string", "null"]
+        },
+        "loggingTargetPrefix": {
+            "description": "Prefix used when creating the log object in the Target bucket. https://docs.aws.amazon.com/AmazonS3/latest/API/API_LoggingEnabled.html",
+            "type": ["string", "null"]
+        },
+        "awsAccessKeyID": {
+            "description": "Requester AWS access key ID.",
+            "type": ["string", "null"]
+        },
+        "raftSessionID": {
+            "description": "Raft session ID.",
+            "type": ["integer", "null"]
+        }
+    },
+    "additionalProperties": false,
+    "required": [
+        "time",
+        "hostname",
+        "pid",
+        "action",
+        "accountName",
+        "accountDisplayName",
+        "userName",
+        "clientPort",
+        "httpMethod",
+        "bytesDeleted",
+        "bytesReceived",
+        "bodyLength",
+        "contentLength",
+        "elapsed_ms",
+        "httpURL",
+        "startTime",
+        "requester",
+        "operation",
+        "requestURI",
+        "errorCode",
+        "objectSize",
+        "totalTime",
+        "turnAroundTime",
+        "referer",
+        "userAgent",
+        "versionID",
+        "signatureVersion",
+        "cipherSuite",
+        "authenticationType",
+        "hostHeader",
+        "tlsVersion",
+        "aclRequired",
+        "bucketOwner",
+        "bucketName",
+        "req_id",
+        "bytesSent",
+        "clientIP",
+        "httpCode",
+        "objectKey",
+        "logFormatVersion",
+        "loggingEnabled",
+        "loggingTargetBucket",
+        "loggingTargetPrefix",
+        "awsAccessKeyID",
+        "raftSessionID"
+    ]
+}
--- a/tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
+++ b/tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
@@ -1,0 +1,2652 @@
+const assert = require('assert');
+const fs = require('fs');
+const path = require('path');
+const tv4 = require('tv4');
+
+const withV4 = require('../support/withV4');
+const BucketUtility = require('../../lib/utility/bucket-util');
+const { config, serverAccessLogsModes } = require('../../../../../lib/Config');
+
+const TEST_CONFIG = {
+    MAX_LOG_WAIT_RETRIES: 50,
+    LOG_POLL_DELAY_MS: 100,
+};
+
+// Load the JSON schema
+const schemaPath = path.join(__dirname, '../../../../../schema/server_access_log.schema.json');
+const schema = JSON.parse(fs.readFileSync(schemaPath, 'utf8'));
+
+function truncateLogFileIfExists(filePath) {
+    if (fs.existsSync(filePath)) {
+        fs.truncateSync(filePath, 0);
+    }
+}
+
+async function waitForLogs(filePath, expectedLines, maxRetries, delayMs) {
+    for (let attempt = 0; attempt < maxRetries; attempt++) {
+        const logEntries = fs.readFileSync(filePath, 'utf8');
+        const lines = logEntries.trim().split('\n').filter(line => line.length > 0);
+        if (lines.length >= expectedLines) {
+            try {
+                return lines.map(line => JSON.parse(line));
+            } catch (err) {
+                // FIXME(CLDSRV-800): readFileSync may read partial lines making JSON.parse fail, so we need to retry.
+                if (attempt == maxRetries) {
+                    throw new Error(`Failed to read log entries from ${filePath} after ${maxRetries} attempts: ${err}`);
+                }
+            }
+        }
+        await sleep(delayMs);
+    }
+    throw new Error(`Failed to read log entries from ${filePath} after ${maxRetries} attempts`);
+}
+
+async function waitForAction(filePath, action, maxRetries, delayMs) {
+    for (let attempt = 0; attempt < maxRetries; attempt++) {
+        const logEntries = fs.readFileSync(filePath, 'utf8');
+        const lines = logEntries.trim().split('\n').filter(line => line.length > 0);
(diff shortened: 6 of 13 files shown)


schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "description": "Total duration of the request in milliseconds. The timer starts when the server first routes the request and stops when the request completes or is closed prematurely.",
+            "type": "number",
+            "minimum": 0
+        },
+        "httpURL": {
+            "description": "Request URL string. This contains only the URL that is present in the actual HTTP request.",
+            "type": "string"
+        },
+        "startTime": {
+            "description": "Timestamp in seconds.milliseconds, recorded when the server first routes the request. Represents the AWS server access log 'Time' field. String type because to be compatible with clickhouse DateTime64(3) type.",
```
dvasilas:
```suggestion
            "description": "Timestamp in seconds.milliseconds, recorded when the server first routes the request. Represents the AWS server access log 'Time' field. String type to be compatible with clickhouse DateTime64(3) type.",
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "description": "Represents the AWS server access log 'Key' field.",
+            "type": ["string", "null"]
+        },
+        "logFormatVersion": {
+            "description": "Version of the server access log schema.",
+            "type": ["string"],
+            "const": "0"
+        },
+        "loggingEnabled": {
+            "description": "If true it means that the target bucket has bucket logging configured and enabled.",
```
dvasilas:
```suggestion
            "description": "True if the target bucket has bucket logging configured and enabled.",
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "description": "Request URL string. This contains only the URL that is present in the actual HTTP request.",
+            "type": "string"
+        },
+        "startTime": {
+            "description": "Timestamp in seconds.milliseconds, recorded when the server first routes the request. Represents the AWS server access log 'Time' field. String type because to be compatible with clickhouse DateTime64(3) type.",
+            "type": "string",
+            "minimum": 0
+        },
+        "requester": {
+            "description": "Represents the AWS server access log 'Requester' field.",
```
dvasilas:
```suggestion
            "description": "Represents the AWS server access log 'Requester' field. https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html",
```

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,1083 @@
...
+                };
+                const expectedOperations = ['REST.PUT.BUCKET', 'REST.PUT.OBJECT', 'REST.GET.TAGGING'];
+                return { method, methodName: 'objectGetTagging', expectedOperations };
+            })(),
+            (() => {
+                // This operation tests copying an object.
+                const method = async () => {
+                    await s3.createBucket({ Bucket: bucketName }).promise();
+                    await s3.putObject({ Bucket: bucketName, Key: objectKey, Body: 'test data' }).promise();
+                    await s3.copyObject({
```
dvasilas, line 2072: AWS doc says
> A copy operation involves a GET and a PUT. For that reason, we log two records when performing a copy operation.

Do we have a different behavior for copies?

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,1083 @@
...
+                assert.strictEqual(logEntries.length, operation.expectedOperations.length,
+                    `Expected ${operation.expectedOperations.length} log entries, got ${logEntries.length}`);
+
+                for (let logEntryIdx = 0, operationIdx = 0;
+                    operationIdx < operation.expectedOperations.length;
+                    logEntryIdx++, operationIdx++) {
+                    const result = tv4.validateResult(logEntries[logEntryIdx], schema);
+                    assert.strictEqual(result.valid, true,
+                        `Log entry should match schema: ${JSON.stringify(result.error)}`);
+                    assert.strictEqual(logEntries[logEntryIdx].operation, operation.expectedOperations[operationIdx],
```
dvasilas: Do you think we can add checks for more fields (maybe not this PR) ?
I am thinking at least `bytesReceived`, `bytesDeleted`, `objectSize`, `versionID`.

There are some checks that you can probably add easily in this PR: `bucketName`, `bucketOwner` (?), `objectKey`, `loggingEnabled`, `loggingTargetBucket`, `loggingTargetPrefix`, `awsAccessKeyID`, `httpMethod`, `httpCode`.

My point is: let's try to have as much test coverage as possible for the log record.

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: I am guessing AWS does a GET + PUT in the background to do the copy, I suppose metadata does the copy without recalling the AWS API.

There is only one API call to https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html so only one log entry

The trace 
- 1 PUT
- 1 COPY
- 2 DeleteObject

So the copy is there
```
{"time":1764767027,"hostname":"xps","pid":182,"action":"CreateBucket","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41824,"httpMethod":"PUT","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.27713,"httpURL":"/test-server-access-log-bucket","startTime":"1764767027.212","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.PUT.BUCKET","requestURI":"PUT /test-server-access-log-bucket HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":null,"bucketName":"test-server-access-log-bucket","req_id":"d1608b0bae41b38012da","bytesSent":null,"clientIP":"::ffff:127.0.0.1","httpCode":200,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"PutObject","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41826,"httpMethod":"PUT","bytesDeleted":null,"bytesReceived":9,"bodyLength":9,"contentLength":9,"elapsed_ms":0.407388,"httpURL":"/test-server-access-log-bucket/test-object-key","startTime":"1764767027.214","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.PUT.OBJECT","requestURI":"PUT /test-server-access-log-bucket/test-object-key HTTP/1.1","errorCode":null,"objectSize":9,"totalTime":"0","turnAroundTime":null,"referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","bucketName":"test-server-access-log-bucket","req_id":"e17dcc57bb40863715d2","bytesSent":null,"clientIP":"::ffff:127.0.0.1","httpCode":200,"objectKey":"test-object-key","logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"CopyObject","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41840,"httpMethod":"PUT","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.43388,"httpURL":"/test-server-access-log-bucket/test-object-key-copy","startTime":"1764767027.215","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.PUT.COPY","requestURI":"PUT /test-server-access-log-bucket/test-object-key-copy HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","bucketName":"test-server-access-log-bucket","req_id":"007c6d149d4cad3261cf","bytesSent":185,"clientIP":"::ffff:127.0.0.1","httpCode":200,"objectKey":"test-object-key-copy","logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"ListBuckets","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41852,"httpMethod":"GET","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.246037,"httpURL":"/","startTime":"1764767027.217","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.GET.UNKNOWN","requestURI":"GET / HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":null,"bucketName":null,"req_id":"8880cae3184492467b5f","bytesSent":385,"clientIP":"::ffff:127.0.0.1","httpCode":200,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"ListMultipartUploads","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41860,"httpMethod":"GET","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.358717,"httpURL":"/test-server-access-log-bucket?uploads","startTime":"1764767027.218","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.GET.UPLOADS","requestURI":"GET /test-server-access-log-bucket?uploads HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","bucketName":"test-server-access-log-bucket","req_id":"0c26c04dd745bab9f564","bytesSent":284,"clientIP":"::ffff:127.0.0.1","httpCode":200,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"ListObjects","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41872,"httpMethod":"GET","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.327174,"httpURL":"/test-server-access-log-bucket?versions","startTime":"1764767027.219","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.GET.BUCKET","requestURI":"GET /test-server-access-log-bucket?versions HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","bucketName":"test-server-access-log-bucket","req_id":"5536b252b849a73e993e","bytesSent":983,"clientIP":"::ffff:127.0.0.1","httpCode":200,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"DeleteObject","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41880,"httpMethod":"DELETE","bytesDeleted":9,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.380872,"httpURL":"/test-server-access-log-bucket/test-object-key?versionId=null","startTime":"1764767027.220","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.DELETE.OBJECT","requestURI":"DELETE /test-server-access-log-bucket/test-object-key?versionId=null HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":"null","signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","bucketName":"test-server-access-log-bucket","req_id":"5d2520863f41a1d58f2d","bytesSent":null,"clientIP":"::ffff:127.0.0.1","httpCode":204,"objectKey":"test-object-key","logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"DeleteObject","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41886,"httpMethod":"DELETE","bytesDeleted":9,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.279938,"httpURL":"/test-server-access-log-bucket/test-object-key-copy?versionId=null","startTime":"1764767027.222","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.DELETE.OBJECT","requestURI":"DELETE /test-server-access-log-bucket/test-object-key-copy?versionId=null HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":"null","signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","bucketName":"test-server-access-log-bucket","req_id":"2c7f5939924bada9a4c4","bytesSent":null,"clientIP":"::ffff:127.0.0.1","httpCode":204,"objectKey":"test-object-key-copy","logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
{"time":1764767027,"hostname":"xps","pid":182,"action":"DeleteBucket","accountName":"Bart","accountDisplayName":"Bart","userName":null,"clientPort":41902,"httpMethod":"DELETE","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":0.234398,"httpURL":"/test-server-access-log-bucket","startTime":"1764767027.222","requester":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","operation":"REST.DELETE.BUCKET","requestURI":"DELETE /test-server-access-log-bucket HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"0","turnAroundTime":"0","referer":null,"userAgent":"aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"127.0.0.1:8000","tlsVersion":null,"aclRequired":null,"bucketOwner":"79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be","bucketName":"test-server-access-log-bucket","req_id":"8d56bebe404988d04d4d","bytesSent":null,"clientIP":"::ffff:127.0.0.1","httpCode":204,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"accessKey1","raftSessionID":null}
```

dvasilas: Makes sense.

schema/server_access_log.schema.json
leif-scality: done

schema/server_access_log.schema.json
leif-scality: done

schema/server_access_log.schema.json
leif-scality: done

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: I now check more properties.
- Time properties are not tested because we can't control them, the type is checked by the schema
- Bytes*/contentLength are not tested because I think they may change so better do it after. The schema checks their types
- VersionID we don't know what version value is going to be used 
- TLS is not tested because we don't have an HTTPS setup
```js
        // List of all the log properties.
        // Commented properties are set by the test or not tested.
        // UNKNOWN: The property is not accessible by the test or is non determistic.
        // DYNAMIC: The propery is set depending on the operation.
        // STATIC: Shared among all tests.
        // TODO: The property need to be tested.
        const commonProperties = {
            // 'time': '', // UNKNOWN
            // 'hostname': '', // UNKNOWN
            // 'pid': '', // UNKNOWN
            'action': 'REQUIRED', // DYNAMIC
            'accountName': 'Bart', // STATIC
            'accountDisplayName': 'Bart', // STATIC
            'userName': null, // TODO: Add test with IAM user to get a non null userName.
            // 'clientPort': '', // UNKNOWN
            'httpMethod': 'REQUIRED', // DYNAMIC
            // 'bytesDeleted': '', // TODO
            // 'bytesReceived': '', // TODO
            // 'bodyLength': '', // TODO
            // 'contentLength': '', // TODO
            // 'elapsed_ms': '', // UNKNOWN
            // 'httpURL': '', // TODO
            // 'startTime': '', // UNKNOWN
            'requester': '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be', // STATIC
            'operation': 'REQUIRED', // DYNAMIC
            // 'requestURI': '', // TODO
            'errorCode': null,  // DYNAMIC
            // 'objectSize': '', // TODO
            // 'totalTime': '', // UNKNOWN
            // 'turnAroundTime': '', // UNKNOWN
            'referer': null, // TODO: Add test that sets the referer.
            // 'userAgent': // UNKNOWN
            // 'versionID': '', // UNKNOWN
            'signatureVersion': 'SigV4', // STATIC
            'cipherSuite': null, // TODO: Add https tests.
            'authenticationType': 'AuthHeader', // STATIC
            // 'hostHeader': '', // UNKNOWN
            'tlsVersion': null, // TODO: Add https tests.
            'aclRequired': null, // TODO: Add https tests.
            'bucketOwner': '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be', // DYNAMIC
            bucketName, // DYNAMIC
            // 'req_id': '', // UNKNOWN
            // 'bytesSent': '', // TODO
            // 'clientIP': '', // UNKNOWN
            'httpCode': 200, // DYNAMIC
            'objectKey': null, // DYNAMIC
            'logFormatVersion': '0', // STATIC
            'loggingEnabled': false, // DYNAMIC
            'loggingTargetBucket': null, // DYNAMIC
            'loggingTargetPrefix': null, // DYNAMIC
            'awsAccessKeyID': 'accessKey1', // STATIC
            'raftSessionID': null, // UNKNOWN
        };
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "minimum": 0
+        },
+        "httpURL": {
+            "description": "Request URL string. This contains only the URL that is present in the actual HTTP request.",
+            "type": "string"
+        },
+        "startTime": {
+            "description": "Timestamp in seconds.milliseconds, recorded when the server first routes the request. Represents the AWS server access log 'Time' field. String type to be compatible with clickhouse DateTime64(3) type.",
+            "type": "string",
+            "minimum": 0
```
dvasilas: Minimum is probably not needed here.

dvasilas (approved):
Don't you need an Arsenal bump?

schema/server_access_log.schema.json
leif-scality: The startTime cannot be negative, so it is just an extra sanity check

dvasilas: Yes, I mean that it may not apply because statTime is a string. Not sure.

.github/docker/docker-compose.yaml
```diff
@@ -8,6 +8,7 @@ services:
       - ${HOME}/.aws/credentials:/root/.aws/credentials
       - /tmp/artifacts/${JOB_NAME}:/artifacts
       - /tmp/coverage/${JOB_NAME}:/coverage/test
+      - /logs:/logs
```
tcarmet, line 11: I understand why you are doing this, but it's a bit annoying when we want to execute those locally to either remind ourselves to create this file, or don't forget to add the steps you added in the CI to make sure we create the `/logs` directory.

Is there a way to reduce a bit those? either use `/tmp/logs` on the host or a directory in the repo that you keep with `.gitkeep` but ignore everything inside with `.gitignore` configuration. Just so that's it's a bit less problematic for local execution.

leif-scality: The tests read the `config.json` file so you can modify `serverAccessLogs.outputFile` to read from where ever you want.
I wrote the CI code for it to work with the default config. But if you change the config and both the tests and the server are in sync it should work

tcarmet: So having something like `/tmp/logs:/logs` would break it, because both cloudserver and the tests need to read the logs from the same file path?

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
+{
+    "$id": "https://scality.com/cloudserver/server_access_log.schema.json",
```
jonathan-gramain: I would rather see a github.com link pointing to the actual file in a given branch of the Cloudserver repository here

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Server Access Log",
+    "description": "A server access log",
+    "type": "object",
+    "properties": {
+        "time": {
+            "description": "Epoch timestamp in seconds, recorded when the log record is created.",
```
jonathan-gramain, line 9: nitpick: some descriptions end with a period, some not, they should be consistent. I would think no period is better if they are not complete sentences.

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "description": "PID of the worker",
+            "type": "integer",
+            "minimum": 0
+        },
+        "action": {
+            "description": "S3 API action name.",
+            "type": "string"
+        },
+        "accountName": {
+            "description": "Requester account display name.",
```
jonathan-gramain, line 27:
```suggestion
            "description": "Requester account name",
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "maximum": 65535,
+            "minimum": 1
+        },
+        "httpMethod": {
+            "description": "Request HTTP method.",
+            "type": "string",
+            "enum": ["GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE", "POST", "PATCH", "CONNECT"]
+        },
+        "bytesDeleted": {
+            "description": "For DeleteObject bytesDeleted is the size in bytes of the deleted object, for DeleteObjects it is the sum of all the deleted objects.",
```
jonathan-gramain: suggestion for a bit less verbosity and more accuracy:

```suggestion
            "description": "For DeleteObject: size in bytes of the deleted object, for DeleteObjects: byte size sum of all the deleted objects",
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "type": ["integer", "null"],
+            "minimum": 0
+        },
+        "bytesReceived": {
+            "description": "Size of the object in bytes of PutObject and UploadPart",
```
jonathan-gramain:
```suggestion
            "description": "For PutObject and UploadPart: size of the object in bytes",
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "description": "Total duration of the request in milliseconds. The timer starts when the server first routes the request and stops when the request completes or is closed prematurely.",
+            "type": "number",
+            "minimum": 0
+        },
+        "httpURL": {
+            "description": "Request URL string. This contains only the URL that is present in the actual HTTP request.",
+            "type": "string"
+        },
+        "startTime": {
+            "description": "Timestamp in seconds.milliseconds, recorded when the server first routes the request. Represents the AWS server access log 'Time' field. String type to be compatible with clickhouse DateTime64(3) type.",
```
jonathan-gramain:
```suggestion
            "description": "Timestamp formatted as: 'seconds.milliseconds', recorded when the server first routes the request. Represents the AWS server access log 'Time' field. String type compatible with Clickhouse DateTime64(3) type.",
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+            "type": "string",
+            "minimum": 0
+        },
+        "requester": {
+            "description": "Represents the AWS server access log 'Requester' field. https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html",
```
jonathan-gramain: I think you could get rid of "Represents the" for this and all following values, should be self explanatory without it.

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+        "authenticationType": {
+            "description": "Represents the AWS server access log 'Authentication Type' field.",
+            "type": ["string", "null"]
+        },
+        "hostHeader": {
+            "description": "Represents the AWS server access log 'Host Header' field.",
+            "type": "string"
+        },
+        "tlsVersion": {
+            "description": "Represents the AWS server access log 'Tls Version' field.",
```
jonathan-gramain:
```suggestion
            "description": "Represents the AWS server access log 'TLS Version' field.",
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+        "bucketOwner": {
+            "description": "Represents the AWS server access log 'Bucket Owner' field.",
+            "type": ["string", "null"]
+        },
+        "bucketName": {
+            "description": "Represents the AWS server access log 'Bucket' field.",
+            "type": "string"
+        },
+        "req_id": {
+            "description": "Represents the AWS server access log 'Request ID' field. Matches the req_id of the stdout logs.",
```
jonathan-gramain: I think "stdout" is an implementation detail (and it ends up in log files on S3C transparently) so I'd rather just say "matches the req_id field in logs"

lib/utilities/serverAccessLogger.js
```diff
@@ -182,74 +182,75 @@ function getRemoteIPFromRequest(request) {
...
         'bucketPutLogging': 'LOGGING_STATUS',
         'bucketGetLogging': 'LOGGING_STATUS',
         // 'corsPreflight': '',
-        'completeMultipartUpload': 'OBJECT',
-        'initiateMultipartUpload': 'OBJECT',
-        'listMultipartUploads': 'OBJECT',
-        'listParts': 'OBJECT',
+        'completeMultipartUpload': 'UPLOAD',
+        'initiateMultipartUpload': 'UPLOAD',
+        'listMultipartUploads': 'UPLOADS',
```
jonathan-gramain: Checking if it's not a typo and you mean plural here, since others are all singular.

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
jonathan-gramain, line 1: Nitpick: generally speaking, it should be doable to merge tests for groups of API functions that put, get and delete something, in a single test that would put, then get, then delete the API object. Just to reduce the number of tests since I believe the test coverage would be identical.

Not blocking for me, just a suggestion.

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,2562 @@
...
+                    logEntryIdx++, operationIdx++) {
+                    const result = tv4.validateResult(logEntries[logEntryIdx], schema);
+                    assert.strictEqual(result.valid, true,
+                        `Log entry should match schema: ${JSON.stringify(result.error)}`);
+
+                    const properties = operation.expected[operationIdx];
+                    for (const [key, val] of Object.entries(properties)) {
+                        assert.strictEqual(logEntries[logEntryIdx][key], val,
+                            `Invalid value for ${key}, action ${properties.action}`);
+                    }
```
jonathan-gramain, line 2647: Could this be replaced by:

```
assert.deepStrictEqual(logEntries[logEntryIdx], operation.expected[operationIdx])
```

A failed assertion would show a diff, so I think it would give the same amount of debug info, if not more.

lib/server.js
```diff
@@ -396,7 +396,8 @@ class S3Server {
                 }
             }
 
-           try {
+            try {
+                logger.info('ServerAccessLogger config', { 'config': _config.serverAccessLogs });
```
jonathan-gramain:
```suggestion
                logger.info('ServerAccessLogger config', { config: _config.serverAccessLogs });
```

.github/docker/docker-compose.yaml
leif-scality: In the CI if you don't change the test config.json to read from `/tmp/logs` and create the `/tmp/logs` folder, yes it will break.

In the CI the clouserver container and the tests don't read the same config file, so cloudserver would write to its `/logs` in its container, and the changes would be visible in `/tmp/logs` in the host filesystem. For this to work you would have to update the test config.json for it to read from `/tmp/logs` and create the `/tmp/logs` folder before the test.

The only way for the test to work locally and in the CI is to change the config.json to `/tmp/logs` but then you need to change the value in Federation back to `/logs` and you also have to create the folder in all cases because cloudserver  only creates the file not the folder.

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,248 @@
...
+        },
+        "tlsVersion": {
+            "description": "Represents the AWS server access log 'Tls Version' field.",
+            "type": ["string", "null"]
+        },
+        "aclRequired": {
+            "description": "Represents the AWS server access log 'ACL Required' field.",
+            "type": ["null"]
+        },
+        "bucketOwner": {
```
nicolas2bert, line 146: `bucketOwner` and `bucketName` should never be `null`.
This is what is breaking integration tests.

For some operations like CreateBucket, deleteObject**s**, uploadPart, the bucket

here are the server access logs for:
- deleteobjects:
```    
{"time":1764928092,"hostname":"nicolas2be-952-store-4","pid":81,"action":"DeleteObjects","accountName":"jul","accountDisplayName":"jul","userName":null,"clientPort":51762,"httpMethod":"POST","bytesDeleted":null,"bytesReceived":129,"bodyLength":129,"contentLength":129,"elapsed_ms":183.005601,"httpURL":"/bucket0?delete","startTime":"1764928092.696","requester":"b4e50bcfa5a588f74c82becb4476a0d7442892ba27d05baa847f0528b1076f72","operation":"REST.POST.OBJECT","requestURI":"POST /bucket0?delete HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"182","turnAroundTime":"100","referer":null,"userAgent":"aws-cli/1.38.10 md/Botocore#1.37.10 ua/2.1 os/macos#24.5.0 md/arch#arm64 lang/python#3.13.2 md/pyimpl#CPython m/N cfg/retry-mode#legacy botocore/1.37.10","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"51.21.225.168:8080","tlsVersion":null,"aclRequired":null,"bucketOwner":null,"bucketName":null,"req_id":"151a943d0741b0e6144c","bytesSent":183,"clientIP":"10.160.120.218","httpCode":200,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"GRRX6ENBPIXH0XXUPZ1P","raftSessionID":null}
```

- CreateBucket:

```
{"time":1764841467,"hostname":"nicolas2be-952-store-1","pid":109,"action":"CreateBucket","accountName":"jul","accountDisplayName":"jul","userName":null,"clientPort":31794,"httpMethod":"PUT","bytesDeleted":null,"bytesReceived":0,"bodyLength":0,"contentLength":0,"elapsed_ms":268.683911,"httpURL":"/bucket0","startTime":"1764841467.173","requester":"b4e50bcfa5a588f74c82becb4476a0d7442892ba27d05baa847f0528b1076f72","operation":"REST.PUT.BUCKET","requestURI":"PUT /bucket0 HTTP/1.1","errorCode":null,"objectSize":null,"totalTime":"268","turnAroundTime":"224","referer":null,"userAgent":"aws-cli/1.38.10 md/Botocore#1.37.10 ua/2.1 os/macos#24.5.0 md/arch#arm64 lang/python#3.13.2 md/pyimpl#CPython m/N cfg/retry-mode#legacy botocore/1.37.10","versionID":null,"signatureVersion":"SigV4","cipherSuite":null,"authenticationType":"AuthHeader","hostHeader":"51.21.225.168:8080","tlsVersion":null,"aclRequired":null,"bucketOwner":null,"bucketName":null,"req_id":"cbfa47daf640b6b6ec00","bytesSent":null,"clientIP":"10.160.120.218","httpCode":200,"objectKey":null,"logFormatVersion":"0","loggingEnabled":false,"loggingTargetBucket":null,"loggingTargetPrefix":null,"awsAccessKeyID":"GRRX6ENBPIXH0XXUPZ1P","raftSessionID":null}
```

lib/utilities/serverAccessLogger.js
leif-scality: It is upload in this case, but I fixed other issues in the map

https://github.com/awslabs/glue-extensions-for-iceberg/blob/52bdb2908216a85859fd76a45981d0326d016a2f/spark/src/main/scala/software/amazon/glue/s3a/audit/S3LogVerbs.java#L36

schema/server_access_log.schema.json
leif-scality: Fixed by b361882ec996f2b5e1be60e79974cebfe916e674

dvasilas (commented):
We should also include tests about error cases (`AccessDenied`, `NoSuchBucket`, ..).

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: No, because some fields we don't control like timestamps.

Output with the suggested change 
```
+ actual - expected

  {
    accountDisplayName: 'Bart',
    accountName: 'Bart',
    aclRequired: null,
    action: 'DeleteBucket',
    authenticationType: 'AuthHeader',
    awsAccessKeyID: 'accessKey1',
+   bodyLength: 0,
    bucketName: 'xxx',
    bucketOwner: null,
+   bytesDeleted: null,
+   bytesReceived: 0,
+   bytesSent: 197,
    cipherSuite: null,
+   clientIP: '::ffff:127.0.0.1',
+   clientPort: 52912,
+   contentLength: 0,
+   elapsed_ms: 0.32016,
    errorCode: 'NoSuchBucket',
+   hostHeader: '127.0.0.1:8000',
+   hostname: 'xps',
    httpCode: 404,
    httpMethod: 'DELETE',
+   httpURL: '/xxx',
    logFormatVersion: '0',
    loggingEnabled: false,
    loggingTargetBucket: null,
    loggingTargetPrefix: null,
    objectKey: null,
+   objectSize: null,
    operation: 'REST.DELETE.BUCKET',
+   pid: 10496,
    raftSessionID: null,
    referer: null,
+   req_id: '5061ab0fdc4b9cf4db78',
+   requestURI: 'DELETE /xxx HTTP/1.1',
    requester: '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be',
    signatureVersion: 'SigV4',
+   startTime: '1764951661.826',
+   time: 1764951661,
    tlsVersion: null,
+   totalTime: '0',
+   turnAroundTime: '0',
+   userAgent: 'aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise',
    userName: null,
+   versionID: null
  }

      + expected - actual

         "aclRequired": [null]
         "action": "DeleteBucket"
         "authenticationType": "AuthHeader"
         "awsAccessKeyID": "accessKey1"
      -  "bodyLength": 0
         "bucketName": "xxx"
         "bucketOwner": [null]
      -  "bytesDeleted": [null]
      -  "bytesReceived": 0
      -  "bytesSent": 197
         "cipherSuite": [null]
      -  "clientIP": "::ffff:127.0.0.1"
      -  "clientPort": 52912
      -  "contentLength": 0
      -  "elapsed_ms": 0.32016
         "errorCode": "NoSuchBucket"
      -  "hostHeader": "127.0.0.1:8000"
      -  "hostname": "xps"
         "httpCode": 404
         "httpMethod": "DELETE"
      -  "httpURL": "/xxx"
         "logFormatVersion": "0"
         "loggingEnabled": false
         "loggingTargetBucket": [null]
         "loggingTargetPrefix": [null]
         "objectKey": [null]
      -  "objectSize": [null]
         "operation": "REST.DELETE.BUCKET"
      -  "pid": 10496
         "raftSessionID": [null]
         "referer": [null]
      -  "req_id": "5061ab0fdc4b9cf4db78"
      -  "requestURI": "DELETE /xxx HTTP/1.1"
         "requester": "79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be"
         "signatureVersion": "SigV4"
      -  "startTime": "1764951661.826"
      -  "time": 1764951661
         "tlsVersion": [null]
      -  "totalTime": "0"
      -  "turnAroundTime": "0"
      -  "userAgent": "aws-sdk-nodejs/2.1692.0 linux/v22.14.0 promise"
         "userName": [null]
      -  "versionID": [null]
       }
      
```

lib/server.js
leif-scality: fixed

schema/server_access_log.schema.json
leif-scality: done

schema/server_access_log.schema.json
leif-scality: done

schema/server_access_log.schema.json
leif-scality: done

schema/server_access_log.schema.json
leif-scality: Changed to the correct description

schema/server_access_log.schema.json
leif-scality: Added periods to all the lines

schema/server_access_log.schema.json
leif-scality: replaced with "https://github.com/scality/cloudserver/tree/development/9.2/schema/lib/server_access_log.schema.json"

schema/server_access_log.schema.json
leif-scality: done

schema/server_access_log.schema.json
leif-scality: done

schema/server_access_log.schema.json
leif-scality: done

leif-scality:
> We should also include tests about error cases (`AccessDenied`, `NoSuchBucket`, ..).

I added one test with `NoSuchBucket` to check the `errorCode` is set

schema/server_access_log.schema.json
leif-scality: Forgot it was a string, you are right

leif-scality: fixed

leif-scality:
ping

.github/docker/docker-compose.yaml
tcarmet: okkk got it, thank you for the clarification! ❤️

leif-scality:
ping

leif-scality:
> > We should also include tests about error cases (`AccessDenied`, `NoSuchBucket`, ..).
> 
> I added one test with `NoSuchBucket` to check the `errorCode` is set

Also added error test for PutObject and GetObject

tcarmet (approved):

lib/api/api.js
```diff
@@ -182,6 +182,7 @@ const api = {
             });
         }
         if (request.serverAccessLog) {
+            request.serverAccessLog.bucketName = request.bucketName;
```
anurag4DSB, line 185: Curious about the reason for this move.
Why did we change the location where we add the bucket name to server access logs?

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,247 @@
...
+            "description": "AWS server access log 'Host Header' field.",
+            "type": "string"
+        },
+        "tlsVersion": {
+            "description": "AWS server access log 'TLS Version' field.",
+            "type": ["string", "null"]
+        },
+        "aclRequired": {
+            "description": "AWS server access log 'ACL Required' field.",
+            "type": ["null"]
```
anurag4DSB, line 144: For all the "AWS server access logs 'XXX' field." descriptions, have we considered using actual descriptions as AWS docs change without version control? Or is it intentional that we want engineers to refer to https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html for what the fields actually represent?
It's worth adding the actual description, and we can add a code comment that these are AWS server access log fields.
Example for `aclRequired`, `AWS server access log 'aclRequired' field indicating whether ACL authorization was required.` would be more helpful than `"AWS server access log 'ACL Required' field."`

Also if we want to be AWS compatible, the proper description will also help us check if we are actually compatible. For example for `aclRequired` allows null, but the actual AWS schema should accept "Yes" or "-".

Example correction:
```
"aclRequired": {
    "description": "AWS server access log 'aclRequired' field indicating whether ACL authorization was required.",
    "type": "string",
    "enum": ["Yes", "-"]
}
```

if we want to allow nulls in our own internal representation, we could explicitly combine them with something like this

```
"aclRequired": {
    "description": "AWS server access log 'aclRequired' field indicating whether ACL authorization was required.",
    "type": ["string", "null"],
    "enum": ["Yes", "-"]
}
```

anurag4DSB: I will let you check and change for others as well.

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,2645 @@
...
+
+async function waitForLogs(filePath, expectedLines, maxRetries = 50, delayMs = 100) {
+    for (let attempt = 0; attempt < maxRetries; attempt++) {
+        const logEntries = fs.readFileSync(filePath, 'utf8');
+        const lines = logEntries.trim().split('\n').filter(line => line.length > 0);
+        if (lines.length >= expectedLines) {
+            try {
+                return lines.map(line => JSON.parse(line));
+            } catch (err) {
+                // FIXME: readFileSync may read partial lines making JSON.parse fail, so we need to retry.
```
anurag4DSB: The FIXME comments acknowledge that readFileSync may read partial lines, causing JSON.parse to fail. This is a real race condition that will cause flaky tests.

We could use a more robust approach with file position tracking

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,247 @@
...
+            "type": "integer",
+            "maximum": 65535,
+            "minimum": 1
+        },
+        "httpMethod": {
+            "description": "Request HTTP method.",
+            "type": "string",
+            "enum": ["GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE", "POST", "PATCH", "CONNECT"]
+        },
+        "bytesDeleted": {
```
anurag4DSB, line 49: `bytesDeleted`, `bytesReceived`, `bodyLength`, `contentLength` all have type: `["integer", "null"]` with minimum: 0, wouldn't null values fail the minimum constraint?
We could do something like this":
```
 {
      "bytesDeleted": {
          "description": "For DeleteObject: size in bytes...",
          "oneOf": [
              {"type": "integer", "minimum": 0},
              {"type": "null"}
          ]
      }
  }
```

Or better, make these fields optional and omit them when null.

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,247 @@
+{
```
anurag4DSB, line 1: For future iterations, we could also add a pattern for these fields.

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,2645 @@
+const assert = require('assert');
```
anurag4DSB, line 1: We should add tests for fields `versionID` populated, `cipherSuite` and `tlsVersion`(if possible)

For my edification,  what happens when we log-rotate mid-request?

lib/utilities/serverAccessLogger.js
```diff
@@ -182,74 +182,78 @@ function getRemoteIPFromRequest(request) {
...
-        'bucketPutNotification': 'BUCKET',
-        'bucketGetNotification': 'BUCKET',
-        'bucketPutEncryption': 'BUCKET',
+        'bucketPutObjectLock': 'OBJECT',
+        'bucketPutNotification': 'NOTIFICATION',
+        'bucketGetNotification': 'NOTIFICATION',
+        'bucketPutEncryption': 'ENCRYPTION',
         'bucketPutLogging': 'LOGGING_STATUS',
         'bucketGetLogging': 'LOGGING_STATUS',
         // 'corsPreflight': '',
```
anurag4DSB: WHy is this commented out?

anurag4DSB: Could we add a fallback or/and a JIra ticket here?
`methodToResType[req.apiMethod] || 'UNKNOWN'`

lib/utilities/serverAccessLogger.js
```diff
@@ -182,74 +182,78 @@ function getRemoteIPFromRequest(request) {
...
         'objectRestore': 'OBJECT',
-        // 'serviceGet': '',
-        // 'websiteGet': '',
-        // 'websiteHead': '',
+        'serviceGet': 'SERVICE', // ListBuckets
+        'websiteGet': 'WEBSITE',
+        'websiteHead': 'WEBSITE',
     });
 
     return `REST.${req.method}.${methodToResType[req.apiMethod] ? methodToResType[req.apiMethod] : 'UNKNOWN'}`;
```
anurag4DSB: Out of scope for this PR, but as I faced it as well, if req.apiMethod is not in the map, it logs REST.GET.UNKNOWN which is valid per schema, but doesn't help debugging. We should log a warning.

something like this
```
.
.
.
      const resourceType = methodToResType[req.apiMethod];
      if (!resourceType) {
          logger.warn('Unknown apiMethod for server access log', {
              apiMethod: req.apiMethod,
              method: req.method,
              url: req.url,
          });
          return `REST.${req.method}.UNKNOWN`;
      }

      return `REST.${req.method}.${resourceType}`;
  }
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,247 @@
...
+    "$id": "https://github.com/scality/cloudserver/tree/development/9.2/schema/lib/server_access_log.schema.json",
```
anurag4DSB, line 2: Another nit pick: The schema doesn't document which fields are populated for which operations. For example:
  - objectKey is null for bucket operations
  - bytesDeleted is only set for DELETE operations
  - versionID is only set for versioned bucket requests

we could add a top-level examples array showing complete log entries for different operation types, or add conditional validation using if/then/else:
```
  {
      "$id": "...",
      "$schema": "...",
      "title": "Server Access Log",
      "description": "A server access log",
      "type": "object",
      "properties": { /* ... */ },
      "required": [ /* ... */ ],
      "examples": [
          {
              "description": "Example PutObject log",
              "value": { /* complete log entry */ }
          },
          {
              "description": "Example GetObject log",
              "value": { /* complete log entry */ }
          }
      ]
  }
```

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,2645 @@
...
+const schemaPath = path.join(__dirname, '../../../../../schema/server_access_log.schema.json');
+const schema = JSON.parse(fs.readFileSync(schemaPath, 'utf8'));
+
+function truncateLogFileIfExists(filePath) {
+    if (fs.existsSync(filePath)) {
+        fs.truncateSync(filePath, 0);
+    }
+}
+
+async function waitForLogs(filePath, expectedLines, maxRetries = 50, delayMs = 100) {
```
anurag4DSB: There are magic numbers here, lets explicitly extract these at module level, 
```
  const TEST_CONFIG = {
      MAX_LOG_WAIT_RETRIES: 50,
      LOG_POLL_DELAY_MS: 100,
      LOG_READ_TIMEOUT_MS: 5000,
  };
```

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,2645 @@
...
+                            operation: 'REST.PUT.CORS',
+                            action: 'PutBucketCors',
+                            httpMethod: 'PUT',
+                        },
+                        {
+                            ...commonProperties,
+                            operation: 'REST.DELETE.CORS',
+                            action: 'DeleteBucketCors',
+                            httpCode: 204,
+                            httpMethod: 'DELETE',
```
anurag4DSB, line 231: NIT: There is a lot of code duplication, worth extracting into a helper function.
Something like this
```
function expectedCreateBucket(overrides = {}) {
      return {
          ...commonProperties,
          operation: 'REST.PUT.BUCKET',
          bucketOwner: null,
          action: 'CreateBucket',
          httpMethod: 'PUT',
          ...overrides,
      };
  }

  // Usage:
  expected: [
      expectedCreateBucket(),
      expectedPutObject({ objectKey }),
      expectedDeleteObject({ objectKey, httpCode: 204 }),
  ]
```

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,2645 @@
...
+        });
+
+        afterEach(async () => {
+            const lastAction = await cleanupBuckets(s3, bucketName);
+            await waitForAction(logFilePath, lastAction);
+            truncateLogFileIfExists(logFilePath);
+        });
+
+        for (const operation of operations) {
+            it(`check ${operation.methodName} log entry`, async () => {
```
anurag4DSB: Worth being more descriptive here

```suggestion
            it(`should log correct ${operation.methodName} operation with all required fields`, async () => {
```

schema/server_access_log.schema.json
```diff
@@ -0,0 +1,247 @@
...
+        "httpCode",
+        "objectKey",
+        "logFormatVersion",
+        "loggingEnabled",
+        "loggingTargetBucket",
+        "loggingTargetPrefix",
+        "awsAccessKeyID",
+        "raftSessionID"
+    ]
+}
```
anurag4DSB, line 248: The schema doesn't specify "additionalProperties": false, allowing undocumented fields to be added.
This ensures strict validation and prevents accidental field additions.
```
  {
      "properties": { /* ... */ },
      "required": [ /* ... */ ],
      "additionalProperties": false
  }
```

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -0,0 +1,2645 @@
...
+// - [ ] Cloudserver returns PutBucketNotification action for PutBucketNotificationConfiguration (same for the Get)
+// - We skip QUOTA methods because they are not part of the AWS API.
+// - We skip objectRestore because it is not supported in CloudServer.
+describe('Server Access Logs - File Output', async () => {
+    withV4(async sigCfg => {
+        const bucketUtil = new BucketUtility('default', sigCfg);
+        bu = bucketUtil;
+        const s3 = bucketUtil.s3;
+        const logFilePath = config.serverAccessLogs.outputFile;
+        const bucketName = 'test-server-access-log-bucket';
```
anurag4DSB, line 106: If tests run in parallel or are interrupted, could this cause conflicts?

lib/utilities/serverAccessLogger.js
```diff
@@ -387,58 +391,58 @@ function logServerAccess(req, res) {
...
-        bytesDeleted: params.analyticsBytesDeleted || params.analyticsBytesDeleted === 0 ? 0 : null,
-        bytesReceived: req.parsedContentLength || 0,
-        bodyLength: parseInt(req.headers['content-length'], 10) || 0,
-        contentLength: req.parsedContentLength || 0,
+        userName: params.analyticsUserName === undefined ? null : params.analyticsUserName,
+        clientPort: req.socket.remotePort === undefined ? null : req.socket.remotePort,
+        httpMethod: req.method === undefined ? null : req.method,
+        bytesDeleted: params.analyticsBytesDeleted === undefined ? null : params.analyticsBytesDeleted,
+        bytesReceived: req.parsedContentLength === undefined ? null : req.parsedContentLength,
+        bodyLength: req.headers['content-length'] === undefined ? null : parseInt(req.headers['content-length'], 10),
```
anurag4DSB, line 422: Very nitpicking here: There are some hard-to-read and sometimes inconsistent complex null coalescing here.
Could we standardize them?
example: 

```
  accountDisplayName: authInfo?.getAccountDisplayName() ?? null,
  clientPort: req.socket?.remotePort ?? null,
  bytesDeleted: params.analyticsBytesDeleted ?? null,
```

anurag4DSB (dismissed):
As this is a new standard which will be used by future updates to S3 analytics, we should also write a document to guide engineers on whats the right way to add new fields to server access logs. This will be helpful for everyone to adhere to the standards thought by the creators of server access logs.

anurag4DSB:
Removing blocker, as per discussion in daily, might be addressed later in other tickets.

lib/api/api.js
leif-scality: Calls like CreateBucket don't call the helpers so bucketName was always null.

https://github.com/scality/cloudserver/pull/6020#discussion_r2597646221

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: We prevent flaky test by retrying, cloudserver should not take more than 5 seconds to write the logs 
I already create a ticket to track this  CLDSRV-800

schema/server_access_log.schema.json
leif-scality: ?

lib/utilities/serverAccessLogger.js
leif-scality: corsPreflight is not an S3 operation so it will output `UNKNOWN`

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: The tests expect the server-access.log file to be empty, if tests are run in parallel this is no longer true so they will break.

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: > We should add tests for fields versionID populated, cipherSuite and tlsVersion(if possible)

cipherSuite and tlsVersion require the client connect using https, I don't know if we  currently have the setup to do it. 

> For my edification, what happens when we log-rotate mid-request?

cloudserver keeps writing to the old file until it detects the new file, fluent bit also keeps reading the old file so we don't lose logs. By default cloudserver checks for file changes every 10secs and fluentbit keeps reading for 30seconds, so we should switch to the new file before fluentbit stops reading

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: Yeah looking back we can do better, the log format should change so I can add the improvements when I update the tests in the next iteration

schema/server_access_log.schema.json
leif-scality: > bucketOwner and bucketName should never be null.

For `ListBuckets` they are going to be `null`

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
anurag4DSB: Could we add the ticket number to the comment please?
(for people just looking at the code)

lib/utilities/serverAccessLogger.js
```diff
@@ -301,7 +305,7 @@ function getObjectSize(request, response) {
             return null;
         }
 
-        return len;
+        return Number(len);
```
BourgoisMickael: It's not very nice to encapsulate a primitive into its object constructor instance.

Besides the constructor might convert `''` to `0` instead of `NaN`. Is it the expected behavior to default to 0 for empty string ?

lib/utilities/serverAccessLogger.js
```diff
@@ -301,7 +305,7 @@ function getObjectSize(request, response) {
...
     }
 
     if (request && objectSizePutMethods[request.apiMethod]) {
```
BourgoisMickael, line 326: Not related to this PR, but those constant objects `objectSizePutMethods` & `objectSizeGetMethods` should be declared only once outside the function to avoid reallocation of constant objects for each requests.

lib/utilities/serverAccessLogger.js
```diff
@@ -182,74 +182,78 @@ function getRemoteIPFromRequest(request) {
 }
 
 function getOperation(req) {
+    // eslint-disable-next-line max-len
+    // https://github.com/awslabs/glue-extensions-for-iceberg/blob/52bdb2908216a85859fd76a45981d0326d016a2f/spark/src/main/scala/software/amazon/glue/s3a/audit/S3LogVerbs.java
+    // https://github.com/open-io/swift/blob/ff518e9907f74b5a2565973a260f36386b5d5cbf/etc/s3-default.cfg.in#L78
+    // https://stackoverflow.com/questions/42707878/amazon-s3-logs-operation-definition
     const methodToResType = Object.freeze({
```
BourgoisMickael: This constant object should be declared only once outside the function to avoid reallocation for each requests. (it's a quit big object it can have a small impact)

lib/utilities/serverAccessLogger.js
BourgoisMickael: I wonder if we should even use `process.emitWarning` to produce the warning directly into stderr to have more chance to keep it for long and notice it rather than having it disappear in regular log rotation.

As we should notice and fix this find of issue fast

BourgoisMickael (commented):
getBytesSent could be simplfied to 

```js

function getBytesSent(res, bytesSent) {
    if (bytesSent !== undefined && bytesSent !== null) {
        return bytesSent;
    }
    return res?.getHeader('Content-Length') || null;
}
```

schema/server_access_log.schema.json
leif-scality: The minimum is only applied when the value is an integer

lib/utilities/serverAccessLogger.js
leif-scality: Changed to 
```
    if (request && response && objectSizeGetMethods[request.apiMethod]) {
        const len = Number(response.getHeader('Content-Length'));
        if (len === undefined || len === null || len === '') {
            return null;
        }

        const numLen = Number(len);
        if (isNaN(numLen)) {
            return null;
        }

        return numLen;
    }

```

lib/utilities/serverAccessLogger.js
leif-scality: added `process.emitWarning`

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: done

schema/server_access_log.schema.json
leif-scality: inlined descriptions

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: added constant

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
leif-scality: done

schema/server_access_log.schema.json
leif-scality: done

lib/utilities/serverAccessLogger.js
leif-scality: moved variable to global scope

lib/utilities/serverAccessLogger.js
leif-scality: moved variable to global scope

leif-scality:
ping

lib/utilities/serverAccessLogger.js
BourgoisMickael: That doesn't change the use of `Number`, this double the usage. And `Number` cannot return null or undefined. It returns a number

PR #6021, opened 2025-12-03
Serve capabilities from config if present

francoisferrand:
The capabilities returned by getCapabilities API will not be directly
set in config by ZKOP, so that they are simply _served_ by Cloudserver.

Old fields are keps for backwards compatibility, but the goal is really
that ZKOP is free to put _any_ content there, as needed by pensieve-api:
this will in particular help to minimize the changes required when
adding a new kind of location.

Issue: CLDSRV-599

### DIFF
--- a/tests/unit/utils/reportHandler.js
+++ b/tests/unit/utils/reportHandler.js
@@ -1,0 +1,290 @@
+const assert = require('assert');
+const sinon = require('sinon');
+const { getCapabilities } = require('../../../lib/utilities/reportHandler');
+
+describe('reportHandler.getCapabilities', () => {
+    const sandbox = sinon.createSandbox();
+    const originalEnv = process.env.LOCAL_VOLUME_CAPABILITY;
+
+    afterEach(() => {
+        sandbox.restore();
+        if (originalEnv === undefined) {
+            delete process.env.LOCAL_VOLUME_CAPABILITY;
+        } else {
+            process.env.LOCAL_VOLUME_CAPABILITY = originalEnv;
+        }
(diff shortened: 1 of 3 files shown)


tests/unit/utils/reportHandler.js
```diff
@@ -0,0 +1,290 @@
...
+            process.env.LOCAL_VOLUME_CAPABILITY = 'false';
+            const cfg = {
+                capabilities: {
+                    locationTypeLocal: true,
+                },
+                supportedLifecycleRules: ['Expiration'],
+            };
+            const caps = getCapabilities(cfg);
+
+            // locationTypeLocal should be false due to env variable
```
DarkIsDude:
```suggestion
```

DarkIsDude: Same for next one 🙏

DarkIsDude (approved):

tests/unit/utils/reportHandler.js
```diff
@@ -0,0 +1,290 @@
...
+        it('should not enable managedLifecycle if managedLifecycle is disabled', () => {
+            const cfg = {
+                capabilities: {
+                    managedLifecycle: false,
+                },
+                supportedLifecycleRules: ['Expiration', 'Transition'],
+            };
+            const caps = getCapabilities(cfg);
+
+            // managedLifecycle should remain true
```
SylvainSenechal:
```suggestion
            // managedLifecycle should remain false
```

SylvainSenechal (approved):

PR #6022, opened 2025-12-08
adding header handling in routeVeam + unit tests

benzekrimaha:
Issue: CLDSRV-806


lib/routes/routeVeeam.js
```diff
@@ -67,8 +89,32 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
         // Download relies on GETs calls with auth in query parameters, that can be
         // checked if 'X-Amz-Credential' is included.
         // Deletion requires that the tags of the object are returned.
-        if (requestQueryParams && Object.keys(requestQueryParams).length > 0
-            && !(method === 'GET' && (requestQueryParams['X-Amz-Credential'] || ('tagging' in requestQueryParams)))) {
+        const queryKeys = Object.keys(requestQueryParams || {});
```
DarkIsDude: Here we parse 3 times the requestQueryParams with same logic (lowerCase then compare). We can maybe do eveything in one shot if this route is performance sensitive ?

lib/routes/routeVeeam.js
```diff
@@ -67,8 +89,32 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
...
+        const isPresignedGet = method === 'GET'
+            && (requestQueryParams
+                && (_hasQueryKeyCaseInsensitive(requestQueryParams, 'X-Amz-Credential')
+                    || _hasQueryKeyCaseInsensitive(requestQueryParams, 'tagging')));
+
+        if (!isPresignedGet && queryKeys.length > 0) {
+            const unsupportedKeys = queryKeys
+                .filter(key => {
```
DarkIsDude: We can do a some and return directly instead of continuing the iteration ?

lib/routes/routeVeeam.js
```diff
@@ -67,8 +89,32 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
...
+                    const lowerKey = key.toLowerCase();
+                    if (allowedSdkQueryKeys.has(lowerKey)) {
+                        return false;
+                    }
+                    if (lowerKey.startsWith('x-amz-sdk-')) {
```
DarkIsDude: Why this one and not the array at the top ? Why are we more permissive here ?

.github/pykmip/Dockerfile
```diff
@@ -4,22 +4,22 @@ FROM python:3.10-alpine
...
 
 RUN apk add --no-cache \
-        libressl && \
+        ca-certificates \
+        openssl && \
     apk add --no-cache --virtual .build-deps \
         python3-dev \
         libffi-dev \
-        libressl-dev \
+        openssl-dev \
```
francoisferrand: why switch from `libressl` to `openssl` ?
AFAIU we actually want to use libressl, which is the more secure & modern lib...

lib/routes/routeVeeam.js
```diff
@@ -67,8 +89,32 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
...
+            if (unsupportedKeys.length > 0) {
+                return errorInstances.InvalidRequest
+                    .customizeDescription('The Veeam SOSAPI folder does not support this action.');
+            }
+        }
+
+        const xIdValue = _getQueryValueCaseInsensitive(requestQueryParams, 'x-id');
+        if (xIdValue && xIdValue !== apiToAction[method]) {
             return errorInstances.InvalidRequest
                 .customizeDescription('The Veeam SOSAPI folder does not support this action.');
```
SylvainSenechal: Is veeam as strict as AWS regarding errors messages, or can we customize it some more ? : 
"the x-id value you provided in the query parameter: xxx, is unexpected, it should be be xxx for this api"

lib/routes/routeVeeam.js
```diff
@@ -67,8 +89,32 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
         // Download relies on GETs calls with auth in query parameters, that can be
         // checked if 'X-Amz-Credential' is included.
         // Deletion requires that the tags of the object are returned.
```
SylvainSenechal: This older comment is mentioning that this logic is used to allow certains apis to go through. 

It mentions the Deletion apis requiring tagging. But the logic later checks that the value of method is GET. Will this work when the method is DELETE ? I see in your tests you added 3 tests that are all PutObject, I say let's test it with a DeleteObject too

SylvainSenechal: More general comment and follow up : I find it hard to reason about both the previous logic and the new one, as these logic try to fit all apis at the same time, with long if checking for method type, different headers. For example, the tagging check I only useful for the DeleteObjects if I understood correctly. 
What would happen if if later we had another constraint to deal with that only applies to one of the 5 apis   

Considering that we have a nice enum list "apiToAction" of all possible apis that this router will handle, I would add a function with an exhaustive switch statement to treat all 5 apis separately.

I also understand that we may not have time to do this

SylvainSenechal: I think the comment about the delete is missleading, i think this comment is still talking about the Get api, and saying that the get method should return tags in order to be able to perform the deletion later. But the way and the place where its written makes it look like we want to allow DELETE apis. anyways

lib/routes/routeVeeam.js
SylvainSenechal: Possibly this, and also considering an early return based on the length of queryKeys or requestQueryParams ?

lib/routes/routeVeeam.js
```diff
@@ -67,8 +89,32 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
...
-        if (requestQueryParams && Object.keys(requestQueryParams).length > 0
-            && !(method === 'GET' && (requestQueryParams['X-Amz-Credential'] || ('tagging' in requestQueryParams)))) {
+        const queryKeys = Object.keys(requestQueryParams || {});
+        const isPresignedGet = method === 'GET'
+            && (requestQueryParams
+                && (_hasQueryKeyCaseInsensitive(requestQueryParams, 'X-Amz-Credential')
+                    || _hasQueryKeyCaseInsensitive(requestQueryParams, 'tagging')));
+
+        if (!isPresignedGet && queryKeys.length > 0) {
```
SylvainSenechal: With the previous logic, having !isPresignedGet && queryKeys.length > 0 was enough to return an error. 
Can we return an error directly here, and do the check unsupportedKeys right after, because here the 2 are mixed but I think it's not necessary ?

SylvainSenechal (changes requested):
Some comments, I think no need for a bigger refactoring, but some simplifications could be possible ? + adding a test with method different than GET

lib/routes/routeVeeam.js
```diff
@@ -67,8 +89,32 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
...
+                    const lowerKey = key.toLowerCase();
+                    if (allowedSdkQueryKeys.has(lowerKey)) {
+                        return false;
+                    }
+                    if (lowerKey.startsWith('x-amz-sdk-')) {
+                        return false;
+                    }
+                    return true;
+                });
+            if (unsupportedKeys.length > 0) {
```
SylvainSenechal: Wait so here, if the api ever receive any query different than the ones listed in allowedSdkQueryKeys, we will block the apis ? Isn't it risky because it forces us to be exhaustive in the list of allowed keys, are we sure this list is now complete ?

tests/unit/internal/routeVeeam.js
```diff
@@ -77,6 +77,44 @@ describe('RouteVeeam: checkBucketAndKey', () => {
...
+            'test',
+            '.system-d26a9498-cb7c-4a87-a44a-8ae204f5ba6c/system.xml',
+            { 'x-id': 'PutObject' },
+            'PUT',
+            log,
+        );
+        assert.strictEqual(err, undefined);
+    });
+
+    it('should allow AWS SDK auxiliary x-amz-sdk-* query params on PUT for system.xml', () => {
```
SylvainSenechal, line 143: nit: could add a test with a wrong param x-iamwrong ?

lib/routes/routeVeeam.js
benzekrimaha: The startsWith('x-amz-sdk-') check is a defensive fallback: AWS SDKs occasionally adds new telemetry keys under that namespace, and I wanted to avoid blocking a future patch release that introduces, say, x-amz-sdk-language if we’d rather force every key to be enumerated explicitly, I can drop the prefix guard and expand the array whenever we encounter a new one. let me know what you think

lib/routes/routeVeeam.js
benzekrimaha: As nothing was specified I'd rather keep it that way in this PR , we can work on better error logging later on as today we only have this kind of errors

lib/routes/routeVeeam.js
benzekrimaha: Before we touched it, the route rejected every non-GET request that carried any query parameters.That’s why our PUT started failing as soon as the SDK v3 appended ?x-id=…. The change simply teaches the validator that x-id (and the handful of SDK telemetry keys) are acceptable, while keeping the old rule for everything else. For the list I tried to make it as exhaustive as possible with everything I could find on official documentations

tests/unit/internal/routeVeeam.js
benzekrimaha: we're already testing a wrong value here => https://github.com/scality/cloudserver/pull/6022/commits/51810dd8fdf5181efc87e5056ff93223acc6e0e4#diff-7d9febf34141ec8f0c48580a4e9e899f15f34227d9ecc39fcda6c5392e87f581R108, isn't it over testing to add other headers as we already know that we reject them directly ?

lib/routes/routeVeeam.js
benzekrimaha: Before we touched it, the route rejected every non-GET request that carried any query parameters.That’s why our PUT started failing as soon as the SDK v3 appended ?x-id=…. The change simply teaches the validator that x-id (and the handful of SDK telemetry keys) are acceptable, while keeping the old rule for everything else.

lib/routes/routeVeeam.js
benzekrimaha: indeed now as soon as we have an unsupportedKey we return

.github/pykmip/Dockerfile
```diff
@@ -13,13 +13,12 @@ RUN apk add --no-cache \
         build-base \
         curl \
         git && \
-    pip3 install --upgrade pip && \
-    pip3 install --upgrade typing-extensions>=4.13.2 && \
     git clone https://github.com/openkmip/pykmip.git && \
     cd pykmip && \
-    python3 setup.py install && \
+    pip3 install . && \
+    cd / && \
```
francoisferrand: why do we need to `cd /` ?
everything afterwards uses absolute path...

.github/pykmip/Dockerfile
```diff
@@ -13,13 +13,12 @@ RUN apk add --no-cache \
         build-base \
         curl \
         git && \
-    pip3 install --upgrade pip && \
-    pip3 install --upgrade typing-extensions>=4.13.2 && \
     git clone https://github.com/openkmip/pykmip.git && \
     cd pykmip && \
-    python3 setup.py install && \
+    pip3 install . && \
```
francoisferrand, line 21: what is the reason for the build failure? Is there a new release of pykmip?
should the fix not include specifying the version of pykmip that we checkout, so we avoid unexpected behavior changes (or incompatibility when it requires a newer version of python...) ?

lib/routes/routeVeeam.js
```diff
@@ -29,6 +29,13 @@ const apiToAction = {
     LIST: 'ListObjects',
 };
 
+const allowedSdkQueryKeys = new Set([
+    'x-id',
+    'x-amz-user-agent',
+    'x-amz-sdk-invocation-id',
+    'x-amz-sdk-request',
```
francoisferrand: not needed, these are covered by the generic `lowerKey.startsWith('x-amz-sdk-')` condition you added

lib/routes/routeVeeam.js
```diff
@@ -67,8 +74,33 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
...
-            && !(method === 'GET' && (requestQueryParams['X-Amz-Credential'] || ('tagging' in requestQueryParams)))) {
+        const originalQuery = requestQueryParams || {};
+        const normalizedQueryKeys = new Map(
+            Object.keys(originalQuery).map(key => [key.toLowerCase(), key])
+        );
+        const isPresignedGet = method === 'GET'
+            && (normalizedQueryKeys.has('x-amz-credential')
+                || normalizedQueryKeys.has('tagging'));
+
+        if (!isPresignedGet && normalizedQueryKeys.size > 0) {
```
francoisferrand: I guess "extra headers" should not be supported either with presigned URL
(looking at the earlier code, it was -slightly incorrectly- trying to check that there is no header unless it is `X-Amz-Credential` or `tagging` on GET)

....but indeed, there are quite a few extra headers to check in the presigned case (X-Amz-Expires, X-Amz-Signature... c.f. extractQueryParams in arsenal)

→ I wonder if it really makes sense to be strict only for non-presigned Get requests: either we should be strict in any case, or never ?

lib/routes/routeVeeam.js
```diff
@@ -67,8 +74,33 @@ function checkBucketAndKey(bucketName, objectKey, requestQueryParams, method, lo
...
+            const hasUnsupportedKeys = Array.from(normalizedQueryKeys.keys())
+                .some(lowerKey => {
```
francoisferrand: this is very inefficient: instead of just going through the list one, you
- create a new map (converting & storing every header in lowercase)
- build yet another array of the keys
- iterate over every entry + explicitly check if presigned and/or x-id

```suggestion
let presigned = false;
for [key, value] in Object.entries(requestQueryParams || {}) {
    const normalizedKey = key.toLowerCase();
    if(normalizedKey == "x-id" && value !== apiToAction[method]) {
        return errorInstances.InvalidRequest
            .customizeDescription('The Veeam SOSAPI folder does not support this action.');
    } 
    if (normalizedKey == 'x-amz-credential' || normalizedKey == 'tagging') {
        presigned = true;
    }
    if (normalizedKey != 'x-amz-user-agent' && !normalizedKey.startsWith('x-amz-sdk-')) {
        return errorInstances.InvalidRequest
            .customizeDescription('The Veeam SOSAPI folder does not support this action.');
    }
}
```

tests/unit/internal/routeVeeam.js
francoisferrand: looking at all the branches in the code, more tests are needed : e.g. having "presigned" headers, on GET or otherwise, different case for headers, ...

francoisferrand:
branch is not up-to-date with upstream development/9.0 : needs to be rebase and resolve conflicts...

.github/pykmip/Dockerfile
benzekrimaha: The build was failing in the python3 setup.py install step when building pykmip from the GitHub repo on python:3.10-alpine (the setup.py invocation was exiting with status 1). There isn’t a new PyPI release of pykmip; we’re building from the GitHub repo because the latest fixes we need are only there (the last tagged release is still 0.10.0).

lib/routes/routeVeeam.js
benzekrimaha: Shud this be looked upon in this PR or rather create a ticket to check this later as we're not altering the previous behaviour here ?

DarkIsDude (approved):

lib/routes/routeVeeam.js
francoisferrand: we are altering the behavior anyway : if we decide to be more lenient, no reason to try to fix the behavior and be "sometimes strict but less strict than before" (which also risks breaking if the SDK "randomly" adds yet more headers we do not support)

The question is really : what are we protecting against with these headers? Are there specific headers that MUST NOT be passed?
(naively, I don't think there are headers when doing a s3 get object request which are expected to modify the semantics : so just ignoring headers should be fine.... unless Veeam "spec" requires otherwise)

francoisferrand: so I guess we should double check the veeam spec ;
and using the same validation as regular get object : no more, no less...

SylvainSenechal (approved):

.github/pykmip/Dockerfile
francoisferrand: if we need a specific SHA1, we can (and should!) specify it anyway; and add a comment explaining why we don't use the release / ...

francoisferrand (approved):

PR #6024, opened 2025-12-10
CLDSRV-808: PUT/GET bucket logging fix test cleanup

leif-scality:
buckets where not deleted when the `before` failed. Replaced the `before` with `beforeEach` and added an `after` global call

### DIFF
--- a/tests/functional/aws-node-sdk/test/bucket/putBucketLogging.js
+++ b/tests/functional/aws-node-sdk/test/bucket/putBucketLogging.js
@@ -38,6 +38,21 @@ const validLoggingConfigWithGrants = {
 
 const itSkipIfAWS = process.env.AWS_ON_AIR ? it.skip : it;
 
+function cleanUp(bucketUtil, cb) {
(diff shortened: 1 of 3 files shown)


BourgoisMickael (approved):

tests/functional/aws-node-sdk/test/bucket/putBucketLogging.js
```diff
@@ -57,6 +57,21 @@ describe('PUT bucket logging', () => {
             });
         }
 
+        after(done => {
+            Promise.all([
```
dvasilas: nit: You could use a helper function to avoid duplication.

dvasilas (approved):

tests/functional/aws-node-sdk/test/bucket/putBucketLogging.js
leif-scality: added helper

leif-scality:
ping

PR #6025, opened 2025-12-11
CLDSRV-809: allow tokens before redis interaction

anurag4DSB:
- Add rate limit fields to server access logs (optional fields for monitoring)
- Fix fail-open behavior at startup: workers now start with full token buffer (50 tokens) instead of 0, preventing request throttling before Redis connection is established

Note: this is default behaviour, which might be modified per strategy used for Redis unavailability which is out of scope of this PR and will be handled in another PR.


anurag4DSB:
rebased with latest dev/9.2 and pushed

BourgoisMickael (approved):

tmacro (approved):

PR #6026, opened 2025-12-11
CLDSRV-810: Use rate limiting config in token bucket class

anurag4DSB:
Use the right config for rate limiting


anurag4DSB:
rebased and pushed.

BourgoisMickael (approved):

tmacro (approved):

PR #6031, opened 2025-12-12
CLDSRV-796: Handle redis unavailability for rate limit

tmacro:


anurag4DSB (approved):

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -107,13 +107,22 @@ class WorkerTokenBucket {
...
+                    this.bucketName,
+                    requested,
+                    interval,
+                    burstCapacity,
+                );
+            } else {
+                // Connection to redis has failed in some way.
+                // Client will be reconnecting in the background.
+                // We grant the requested amount of tokens anyway to avoid degrading service availability.
+                this.log.warn(
```
anurag4DSB, line 123: we should also log the bucket name here

anurag4DSB: should we also add to default fields, rateLimited: false or something, like we do when we rate limit

like this: https://github.com/scality/cloudserver/blob/development/9.2/lib/api/api.js#L386

tmacro: I added the bucket name. I don't think the `rateLimited` field makes sense here as we're not making accept/deny decisions.

tcarmet (approved):

PR #6032, opened 2025-12-15
CLDSRV-814: prevent premature codecov results

tcarmet:
Prevent premature codecov results to be published on PR status before all tests successfully run and pass.

`development/9.0` contains 9 codecov uploads, `development/9.2` contains 10, should be fine to set it up as 9 starting from `9.0` and if needed we can bump it.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,10 @@
 ---
+codecov:
+  require_ci_to_pass: true
+  notify:
+    wait_for_ci: true
+    after_n_builds: 9
+
 parsers:
   javascript:
     enable_partials: yes


codecov.yml
```diff
@@ -1,4 +1,10 @@
 ---
+codecov:
+  require_ci_to_pass: true
+  notify:
+    wait_for_ci: true
+    after_n_builds: 9
```
francoisferrand, line 6: Is this needed since you also specify "wait for ci" ?

(Generally, I dislike this option as it creates a risk that we forget to update it, and possibly lose the notification if the workflow is reworked to reduce number of "reports")

codecov.yml
```diff
@@ -1,4 +1,10 @@
 ---
+codecov:
+  require_ci_to_pass: true
+  notify:
```
francoisferrand, line 4: Does this affect GitHub comment?

There is another setting explicitly affecting it if I remember correctly, but it has fewer options.

codecov.yml
tcarmet: Yes it is needed, if I only leave `wait_for_ci` it will get a single report, and wait for the job of that particular report to finish, instead of waiting for the whole workflow to finish.

And yes it does create a bit of maintenance, but honestly in my experience number of reports tend to go up, instead of down. So I don't think it's that much of a big deal.

codecov.yml
tcarmet: It does, the github comment will wait for those conditions to be reached before being sent now.

anurag4DSB (approved):

tmacro (approved):

PR #6033, opened 2025-12-17
✨ CLDSRV-812: ListObjectsV2 create x-amz-optional-attributes header

DarkIsDude:
# Pull request template

## Description

Support a new header `x-amz-object-optional-attributes` to be able to retrieve in the ListObjectsV2 user metadata.

### Motivation and context

https://github.com/scality/citadel/pull/301/changes

https://scality.atlassian.net/browse/CLDSRV-812

### Related issues

https://github.com/scality/Arsenal/pull/2581 is needed as the bump of Arsenal in Vault.

A new MR will be opened to update the response of `ListObjectsV2` and return asked fields.

- [x] https://github.com/scality/Arsenal/pull/2581 bump Arsenal


tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+
+    describe('z-amz-optional-attributes header', () => {
+        it('should return an error if the user does not have the permission', done => {
+            const authInfoNoPerm = makeAuthInfo('accessKey2'); // A new user without specific permissions
+            const testGetRequest = Object.assign({
+                query: {},
+                url: baseUrl,
+            }, baseGetRequest);
+            testGetRequest.headers['x-amz-optional-object-attributes'] = 'x-amz-meta-department';
+
```
francoisferrand: nit: multiple empty lines

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+                [
+                    next => bucketPut(authInfo, testPutBucketRequest, log, next),
+                    (_, next) => bucketGet(authInfo, testGetRequest, log, next),
+                    (result, _, next) => parseString(result, next),
+                ],
+                (err, result) => {
+                    assert.strictEqual(err, null);
+                    assert.strictEqual(result.ListBucketResult.$.xmlns, 'http://s3.amazonaws.com/doc/2006-03-01/');
+                    done();
+                },
```
francoisferrand: nit: inconsistent formatting of async.xxx
```suggestion
            async.waterfall([
                next => bucketPut(authInfo, testPutBucketRequest, log, next),
                (_, next) => bucketGet(authInfo, testGetRequest, log, next),
                (result, _, next) => parseString(result, next),
            ], (err, result) => {
                assert.strictEqual(err, null);
                assert.strictEqual(result.ListBucketResult.$.xmlns, 'http://s3.amazonaws.com/doc/2006-03-01/');
                done();
            },
```

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+            async.series([
+                next => bucketPut(authInfo, testPutBucketRequest, log, next),
+                next =>  bucketGet(authInfoNoPerm, testGetRequest, log, next),
+            ], err => {
+                assert.strictEqual(err.is.AccessDenied, true);
+                done();
+            });
+        });
+
+        it('should ignore the missing permission if the header contains only RestoreStatus', done => {
```
francoisferrand: nit: it does not really "ignore" the missing permission, it should not even try to check/require it...


```suggestion
        it('should return RestoreStatus without extra permission', done => {
```

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+            const testGetRequest = Object.assign({
```
francoisferrand: while we should have a test with this purpose, looking at the code I think it does not really test anything: permission-wise, we have the exact same test as `should return valid xml if the user have the permission` (the only difference between the tests is the attribute we ask for, but no difference in "auth" setup/mock or assertion on the requested permissions)

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+            ],
+            (err, result) => {
+                assert.strictEqual(err, null);
+                assert.strictEqual(result.ListBucketResult.$.xmlns, 'http://s3.amazonaws.com/doc/2006-03-01/');
+                done();
+            });
+        });
+
+        it('should return an error if the user does not have all permissions for multiple attributes', done => {
+            const authInfoNoPerm = makeAuthInfo('accessKey2');
```
francoisferrand: here as well, same auth setup as `should return an error if the user does not have the permission` : making this test redundant...

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+            });
+        });
+
+        it('should return an error for a mix of valid and invalid attributes', done => {
+            const testGetRequest = Object.assign({
+                query: {},
+                url: baseUrl,
+            }, baseGetRequest);
+            testGetRequest.headers['x-amz-optional-object-attributes'] = 'RestoreStatus,InvalidAttribute';
+
```
francoisferrand, line 395: completely unrelated, but I wonder if we need to be case-insensitive when matching attributes (esp. `RestoreStatus` which is capitalized) : we should check/test AWS behavior (and verify if user metadata "labels" are case sensitive or not), and handle case-senstitiveness accordingly

May be done in a followup ticket however

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+                done();
+            });
+        });
+
+        it('should handle attributes with leading/trailing whitespace', done => {
+            const testGetRequest = Object.assign({
+                query: {},
+                url: baseUrl,
+            }, baseGetRequest);
+            testGetRequest.headers['x-amz-optional-object-attributes'] = 'x-amz-meta-foo';
```
francoisferrand: should have leading/trailing whitespaces


```suggestion
            testGetRequest.headers['x-amz-optional-object-attributes'] = '   x-amz-meta-foo   ';
```

NOTE: may be worth testing as well with `RestoreStatus`, since we will not pick the data form the same place...

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
...
+
+            async.waterfall([
+                next => bucketPut(authInfo, testPutBucketRequest, log, next),
+                (_, next) => bucketGet(authInfo, testGetRequest, log, next),
+                (result, _, next) => parseString(result, next),
+            ],
+            (err, result) => {
+                assert.strictEqual(err, null);
+                assert.strictEqual(result.ListBucketResult.$.xmlns, 'http://s3.amazonaws.com/doc/2006-03-01/');
+                done();
```
francoisferrand, line 420: as such, this test (and the next) do not test much : since we don't have a way to check that "every" requested field is returned

I guess this will be fixed in next PR, but may be best to add a `//TODO` for now to ensure we don't forget to update the test in next PR

francoisferrand (approved):

tests/unit/api/bucketGet.js
```diff
@@ -333,4 +333,186 @@ describe('bucketGet API V2', () => {
             });
         });
     });
+
+    describe('z-amz-optional-attributes header', () => {
```
francoisferrand:
```suggestion
    describe('x-amz-optional-attributes header', () => {
```

tests/unit/api/bucketGet.js
DarkIsDude: That's the idea behind the ignore with `should not even try to check/require it`. I change the code but didn't use RestoreStatus as we don't.

tests/unit/api/bucketGet.js
DarkIsDude: The idea of tests is to tests each "path" of the code. Here we have 2 differents cases (coming from product) that I'm trying to reflect in the code. If you have the permission, then you can ask for user metadata, if not, you should be able to request only RestoreStatus. So even the result is the same, the code path is not the same and we test this condition `if (requestedAttributes.filter(attr => attr != 'RestoreStatus').length > 0) {`

tests/unit/api/bucketGet.js
DarkIsDude: Agree on this one, can make sense to remove it. As we are in unit test, I prefer to try to be exhaustif but I can follow your comment 🙏

tests/unit/api/bucketGet.js
DarkIsDude: That's a really good question. AWS is case sensitive on RestoreStatus. For user metadata, AWS is not case sensitive (everything go to lower case). So let's do it now 🙏

tests/unit/api/bucketGet.js
DarkIsDude: No, in next PR I create new test as the logic tested is different and a test should test only one thing. Even if the result is the same, the logic tested here is to make sure that with several attributes, the GET is working and nothing break

lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -252,6 +252,18 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
                 generateRequestContext('bypassGovernanceRetention');
             requestContexts.push(checkUserGovernanceBypassRequestContext);
         }
+    } else if (apiMethodAfterVersionCheck === 'bucketGet') {
+        requestContexts.push(generateRequestContext(apiMethodAfterVersionCheck));
+
+        const optionalAttributesHeader = request.headers['x-amz-optional-object-attributes'];
+        const requestedAttributes = optionalAttributesHeader ?
+            optionalAttributesHeader.split(',').map(
+                header => header !== 'RestoreStatus' ? header.toLowerCase() : header
```
SylvainSenechal: If I understand correctly, the reason why RestoreStatus is not lower-cased is because of AWS standard ? why not leaving the other params the same ? Because we store them lower-cased ?

lib/api/bucketGet.js
```diff
@@ -284,6 +284,15 @@ function bucketGet(authInfo, request, log, callback) {
     const params = request.query;
     const bucketName = request.bucketName;
     const v2 = params['list-type'];
+
+    const optionalAttributes =
+        request.headers['x-amz-optional-object-attributes']?.split(',').map(attr => attr.trim()) ?? [];
+    if (optionalAttributes.some(attr => !attr.startsWith('x-amz-meta-') && attr != 'RestoreStatus')) {
+        return callback(
+            errorInstances.InvalidArgument.customizeDescription('Invalid header x-amz-optional-object-attributes')
```
SylvainSenechal: Is this the official aws error ? If so we can leave it, but if its a custom one that you choose, why not add the value of the problematic value

tests/unit/api/bucketGet.js
SylvainSenechal, line 1: Ok for the tests in general, but how come you never assert on value returned in result.ListBucketResult ? 
I mean, shouldn't you do a content check check ?

SylvainSenechal: Ah, I guess the feature is not finished and there will be some more work later, which explain why optionalAttributes is not used in bucketGet ?

lib/api/apiUtils/authorization/prepareRequestContexts.js
francoisferrand: for what I understand, you added this since user metadata are expected to be lower case in AWS (did not check, but I think I saw you mention this). In that case,
- what happens (on AWS) if an upper case metadata is used? It is rejected, just "converted" to lower case?
- how does our API behave? If 
- Does this apply to the prefix (`x-amz-meta-`) as well? e.g. if we only consider lower case prefix, no need to convert to lower case for filtering user-metadata properties...

(overall, trying to say that we don't need to be too defensive here: if spec & API both only support lower user-metadata, no need to spend cycle trying to accomodate, we can just be strict and expect user to behave correctly)

tests/unit/api/bucketGet.js
francoisferrand: Testing the different code path is fine (i.e. RestoreStatus vs x-amz-meta-department)

My issue is that the test mentions _permissions_, which is not tested at all in this test : i.e. we do not verify that the right permissions/actions were requested, use some mocks, [...] to ensure that the custom permission is not required in case only `RestoreStatus` is present...

tests/unit/api/bucketGet.js
francoisferrand: it does not test GET is working (only that it does not crash), as it does not validate which permissions are requested

lib/api/apiUtils/authorization/prepareRequestContexts.js
DarkIsDude: Indeed, user metadata are lowercase. AWS will convert a non lowercase usermetadata to a lowercase (we are doing the same). And yes, this apply to the prefix also.

lib/api/bucketGet.js
DarkIsDude: Good catch, the error returned by AWS is `Invalid attribute name specified` so I guess I must do the same 👏

tests/unit/api/bucketGet.js
DarkIsDude: Indeed 👍

tests/unit/api/bucketGet.js
DarkIsDude: I'll add some spy to be sure then 👍

SylvainSenechal (approved):

PR #6035, opened 2025-12-18
Override an unversioned object originOp when moving it to a version

delthas:
This prevents triggering a possible s3:ObjectCreated:Put bucket notification when moving this object to a version, when deleting it (when creating a deletion marker).


delthas:
Moved to #6036

PR #6036, opened 2025-12-18
Override an unversioned object originOp when moving it to a version

delthas:
This prevents triggering a possible s3:ObjectCreated:Put bucket notification when moving this object to a version, when deleting it (when creating a deletion marker).

Issue: CLDSRV-816

### DIFF
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -98,6 +98,7 @@ function _storeNullVersionMD(bucketName, objKey, nullVersionId, objMD, log, cb) 
             isNull2: true,
         });
     }
+    nullVersionMD.originOp = 's3:StoreNullVersion';
     metadata.putObjectMD(bucketName, objKey, nullVersionMD, { versionId }, log, err => {
         if (err) {
             log.debug('error from metadata storing null version as new version',
--- a/tests/unit/api/objectDelete.js
+++ b/tests/unit/api/objectDelete.js
@@ -7,7 +7,7 @@ const services = require('../../../lib/services');
 const { bucketPut } = require('../../../lib/api/bucketPut');
 const bucketPutACL = require('../../../lib/api/bucketPutACL');
 const constants = require('../../../constants');
-const { cleanup, DummyRequestLogger, makeAuthInfo } = require('../helpers');
+const { cleanup, DummyRequestLogger, makeAuthInfo, versioningTestUtils} = require('../helpers');
 const objectPut = require('../../../lib/api/objectPut');
 const { objectDelete, objectDeleteInternal } = require('../../../lib/api/objectDelete');
 const objectGet = require('../../../lib/api/objectGet');
@@ -16,6 +16,7 @@ const mpuUtils = require('../utils/mpuUtils');
 const metadataswitch = require('../metadataswitch');
 const { fakeMetadataArchive } = require('../../functional/aws-node-sdk/test/utils/init');
 const bucketPutNotification = require('../../../lib/api/bucketPutNotification');
+const bucketPutVersioning = require('../../../lib/api/bucketPutVersioning');
 
 const any = sinon.match.any;
 const originalDeleteObject = services.deleteObject;
@@ -31,6 +32,8 @@ const earlyDate = new Date();
 const lateDate = new Date();
 earlyDate.setMinutes(earlyDate.getMinutes() - 30);
 lateDate.setMinutes(lateDate.getMinutes() + 30);
+
+const enableVersioningRequest = versioningTestUtils.createBucketPutVersioningReq(bucketName, 'Enabled');
 
 function testAuth(bucketOwner, authUser, bucketPutReq, objPutReq, objDelReq,
     log, cb) {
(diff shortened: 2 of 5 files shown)


delthas:
Tested locally and with the full workflow on an ARTESCA, working for both manual deletion and expiration workflows.

lib/api/apiUtils/object/versioning.js
```diff
@@ -98,6 +98,7 @@ function _storeNullVersionMD(bucketName, objKey, nullVersionId, objMD, log, cb)
             isNull2: true,
         });
     }
+    nullVersionMD.originOp = 's3:ObjectNullVersionStoredAsNewVersion';
```
francoisferrand: maybe something shorter: `StoreNullVersion` ?
(minimal optimization, but anything we add ends up pilling in metadata.... and with billions of objects, it quickly adds up)

lib/api/apiUtils/object/versioning.js
francoisferrand, line 83: in which situation is this function called? i.e. 
- if it called only when we have a "master-only" version, to "migrate it" to a proper version → fine
- if however it may be called in other cases, like when (over)writing an object (for exemple in version-suspended buckets ?), then we must keep the existing originOp in that case

Did you perform such kind of "impact analysis"? Are we sure this is only called to perform an _extra_ put, for migration purpose, and never to create/replace the new version itself?

→ would be great to check the tests, and (if needed) update them to check that the proper originOp is set in each case (`putObject`, `copyObject`, `putMultipart` and `backeatPutMetadata` in the different combination of versioned mode & pre-existing object, ...)

lib/api/apiUtils/object/versioning.js
```diff
@@ -98,6 +98,7 @@ function _storeNullVersionMD(bucketName, objKey, nullVersionId, objMD, log, cb)
             isNull2: true,
         });
     }
```
francoisferrand, line 100: this was missed in the existing tests : so we should introduce -or update- a test (unit and/or functional, in line with how it is already tested) to make sure this behavior will stay.

delthas: OK, added a test that checks for versionId, isNull, and originOp. https://github.com/scality/cloudserver/pull/6036/files#diff-dffdecd8dd3ea359b81fdebca3575a93b16e025dcfca47e1f609282f31b70764R425

Do we want to test anything else in that test? (Apart from check the existing originOp in existing put/copy/...)

lib/api/apiUtils/object/versioning.js
delthas: OK, added some more tests.

tests/unit/api/objectPut.js
```diff
@@ -980,6 +980,30 @@ describe('objectPut API with versioning', () => {
         });
     });
 
+    it('should set originOp when moving null version', done => {
```
SylvainSenechal: I think it's worth it to document more either this test or the one in objectDelete.js. 
Here you added a comment "M was moved to V", I understand it because I read the Jira ticket where you give a lot of context, but someone else probably wouldn't understand. 

This new StoreNullVersion doesn't look very documented either, so this test could be a an opportunity to write a few line about the bug that happened before and why we added StoreNullVersion to fix it

SylvainSenechal (approved):

tests/unit/api/objectDelete.js
```diff
@@ -371,3 +374,59 @@ describe('objectDelete API', () => {
         });
     });
 });
+
+
```
francoisferrand: nit: remove extra empty line

```suggestion
```

tests/unit/api/objectDelete.js
```diff
@@ -371,3 +374,59 @@ describe('objectDelete API', () => {
...
+    });
+    const testDeleteRequest = new DummyRequest({
+        bucketName,
+        namespace,
+        objectKey,
+        headers: {},
+        url: `/${bucketName}/${objectKey}`,
+    });
+
+    it('should store modified originOp when moving object null version', done => {
```
francoisferrand: maybe better describe the scenario, to explain this happens when putting a delete marker on top of a "master-only null version"


```suggestion
    it('should store modified originOp when storing null object due to object marker', done => {
```

...or we could make this more general, and consider this test actually tests this overall process of "upgrading" the null-version object (in case an object marker is put on top of it) : i.e. we do not just want to verify that the object has the proper originOp, but also that it has all proper metadata and "null" version id (which is mostly what you do here, actually)

tests/unit/api/objectPut.js
francoisferrand: > This new StoreNullVersion doesn't look very documented either, so this test could be a an opportunity to write a few line about the bug that happened before and why we added StoreNullVersion to fix it

this is not a bug, this is a "known" corner-case when moving from the "master-only bucket format" (non-versioned) to the "versioned bucket format". And this is documented, in `docs/ARCHITECTURE.rst`, in section "Preserving Existing Null Versions in Versioning-Enabled Bucket".

tests/unit/api/objectDelete.js
```diff
@@ -371,3 +374,59 @@ describe('objectDelete API', () => {
...
+            next => objectDelete(authInfo, testDeleteRequest, log, next),
+            async () => {
+                const calls = metadataswitch.putObjectMD.getCalls();
+                sinon.assert.calledWith(calls[calls.length - 2],
+                    bucketName, objectKey, sinon.match({
+                        versionId: sinon.match.truthy,
+                        isNull: true,
+                        originOp: 's3:StoreNullVersion',
+                    }), any, any, any);
+            },
```
francoisferrand, line 428: just like the other test, I guess we should also check that the second call created a delete marker with the expected originOp (`s3:ObjectRemoved:DeleteMarkerCreated`)

tests/unit/api/objectPut.js
```diff
@@ -980,6 +980,30 @@ describe('objectPut API with versioning', () => {
...
+            async () => {
+                // M was moved to V, with originOp overridden to prevent bucket notifications.
+                const calls = metadata.putObjectMD.getCalls();
+                sinon.assert.calledWith(calls[calls.length - 2],
+                    bucketName, objectName, sinon.match({
+                        originOp: 's3:StoreNullVersion',
+                    }), any, any, any);
+            },
+            async () => {
+                // New V was created with the right originOp.
```
francoisferrand:
```suggestion
                // New version document was created with the right originOp.
```

tests/unit/api/objectPut.js
```diff
@@ -980,6 +980,30 @@ describe('objectPut API with versioning', () => {
...
     });
 
+    it('should set originOp when moving null version', done => {
+        async.series([
+            next => bucketPut(authInfo, testPutBucketRequest, log, next),
+            next => objectPut(authInfo, testPutObjectRequest, undefined, log, next),
+            next => bucketPutVersioning(authInfo, enableVersioningRequest, log, next),
+            next => objectPut(authInfo, testPutObjectRequest, undefined, log, next),
+            async () => {
+                // M was moved to V, with originOp overridden to prevent bucket notifications.
```
francoisferrand:
```suggestion
                // Old master-only document was moved to a proper version, with originOp overridden to prevent unexpected bucket notifications.
```

francoisferrand (approved):

tests/unit/api/objectPut.js
delthas: Slightlly clarified the test name.

tests/unit/api/objectDelete.js
delthas: OK, clarified the test name accordingly.

PR #6037, opened 2025-12-29
add nested xml input format for bucket quota

SylvainSenechal:
Issue: CLDSRV-818

### DIFF
--- a/lib/api/bucketUpdateQuota.js
+++ b/lib/api/bucketUpdateQuota.js
@@ -8,7 +8,10 @@ const monitoring = require('../utilities/monitoringHandler');
 const { parseString } = require('xml2js');
 
 function validateBucketQuotaProperty(requestBody, next) {
-    const quota = requestBody.quota;
+    let quota = requestBody.quota;
+    if (quota === undefined) {
+        quota = requestBody.QuotaConfiguration?.Quota;
+    }
     const quotaValue = parseInt(quota, 10);
     if (Number.isNaN(quotaValue)) {
         return next(errorInstances.InvalidArgument.customizeDescription('Quota Value should be a number'));
@@ -27,7 +30,7 @@ function parseRequestBody(requestBody, next) {
         }
         return next(null, jsonData);
     } catch {
-        return parseString(requestBody, (xmlError, xmlData) => {
+        return parseString(requestBody, { explicitArray: false }, (xmlError, xmlData) => {
             if (xmlError) {
                 return next(errorInstances.InvalidArgument.customizeDescription('Request body must be a JSON object'));
             }
--- a/lib/api/bucketUpdateQuota.js
+++ b/lib/api/bucketUpdateQuota.js
@@ -22,20 +22,26 @@ function validateBucketQuotaProperty(requestBody, next) {
     return next(null, quotaValue);
 }
 
-function parseRequestBody(requestBody, next) {
-    try {
-        const jsonData = JSON.parse(requestBody);
-        if (typeof jsonData !== 'object') {
-            throw new Error('Invalid JSON');
-        }
-        return next(null, jsonData);
-    } catch {
-        return parseString(requestBody, { explicitArray: false }, (xmlError, xmlData) => {
-            if (xmlError) {
+function parseRequestBody(requestBody, contentType, next) {
+    switch (contentType) {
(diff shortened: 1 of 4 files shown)


lib/api/bucketUpdateQuota.js
SylvainSenechal, line 1: @francoisferrand I tried the curl command provided in the doc we gave to the client : 
`curl -X PUT "http://localhost:8000/testb/?quota=true" --user "accessKey1:verySecretKey1"   --aws-sigv4 "aws:amz:us-east-1:s3" -d '{"quota" : 32}'`

In that case, curl will not set the content-type header. So I think it's tricky to modify the logic around the body parsing by using content type, because clients can't be expected to set that header so no matter what we will have to use a fallback which is basically the code that we already have with the try/catch

lib/api/bucketUpdateQuota.js
```diff
@@ -8,7 +8,12 @@ const monitoring = require('../utilities/monitoringHandler');
 const { parseString } = require('xml2js');
 
 function validateBucketQuotaProperty(requestBody, next) {
-    const quota = requestBody.quota;
+    let quota = requestBody.quota;
+    if (quota === undefined) {
+        // Make the api compatible with both non-nested XML and XML with rootNode.
+        // XML with root node is correct XML and allows integration with Smithy client.
```
DarkIsDude:
```suggestion
```

DarkIsDude (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@zenko/cloudserver",
-  "version": "9.2.14",
+  "version": "9.3.0-preview.1",
```
SylvainSenechal, line 3: wanna do this now, because this change is for the cloudserver quota client feature, and I need an image with this new code to run in the test pipeline. Otherwise can't merge the new client

lib/api/bucketUpdateQuota.js
francoisferrand: maybe we should check the `contentType` if set (and default to JSON if not) ?

```
switch(contentType) {
case 'xml':
   return parseString(requestBody, { explicitArray: false }, (xmlError, xmlData) => { [...] };

case 'json':
default:
   const jsonData = JSON.parse(requestBody);
   return next(null, jsonData);
}
```

could be a followup ticket, or just this one as it is probably simple:
- probably a good time to make this change: 9.3 already has "risky" changes, and we have time before we ship it
- should be fine if we keep JSON as fallback (according to the docs which were shared)
- esp. quota API is not widely used at the moment
- but we need to test it explicitly to avoid any surprise

tests/functional/aws-node-sdk/test/bucket/updateBucketQuota.js
```diff
@@ -28,6 +28,16 @@ describe('Test update bucket quota', () => {
     it('should update the quota', () => sendRequest('PUT', 
         '127.0.0.1:8000', `/${bucket}/?quota=true`, JSON.stringify(quota)));
 
+    it('should update quota with XML format', async () => {
+        try {
+            const xmlQuota = '<QuotaConfiguration><Quota>3000</Quota></QuotaConfiguration>';
+            await sendRequest('PUT', '127.0.0.1:8000', `/${bucket}/?quota=true`, xmlQuota);
```
francoisferrand: should it not be something like this:
(may be the same, but closer to how AWS documents its own api...)

```suggestion
            await sendRequest('PUT', '127.0.0.1:8000', `/${bucket}/?quota`, xmlQuota);
```

lib/api/bucketUpdateQuota.js
SylvainSenechal: yes let's make that change and add it in 9.3

tests/functional/aws-node-sdk/test/bucket/updateBucketQuota.js
SylvainSenechal: the api is written with quota=true in the client doc, the existing tests are also using quota=true.
Upon testing it seems that all of these will work : 
{bucketName}/quota=true
{bucketName}/quota=false
{bucketName}/quota=
{bucketName}/quota

But considering the existing documentation and code, I think we can keep using quota=true

francoisferrand (approved):

PR #6041, opened 2026-01-05
CLDSRV-815: Reduce server access log size 

dvasilas:
Reduces server access log size:
  1. Omit `null`/`undefined` fields
  2. Removes `clientPort` and `httpURL` (not used by S3 analytics or bucket logging)
  3. Removes `accountDisplayName` (value is always identical to `accountName`). Fluentbit will copy `accountName` to `accountDisplayName` 

Results to ~25% size reduction.


lib/utilities/serverAccessLogger.js
```diff
@@ -413,17 +426,14 @@ function logServerAccess(req, res) {
         // Analytics
         action: params.analyticsAction === undefined ? null : params.analyticsAction,
         accountName: params.analyticsAccountName === undefined ? null : params.analyticsAccountName,
-        accountDisplayName: authInfo ? authInfo.getAccountDisplayName() : null,
```
BourgoisMickael, line 416: is there actually no difference at all , no matter if we use iam user, role, or internal account (and same for artesca) ? 

Then we could drop the duplicate from regular logs as well

lib/utilities/serverAccessLogger.js
```diff
@@ -391,6 +391,19 @@ function timestampToDateTime643(unixMS) {
...
 }
 
+/**
+ * Omits null and undefined fields from an object.
+ * Only processes own properties (not inherited).
+ * Preserves all other falsy values (0, false, "", etc.)
+ * @param {Object} obj - Object to filter
+ * @returns {Object} - New object with null/undefined fields removed
+ */
+function omitNullFields(obj) {
```
BourgoisMickael: this function produce multiple additional transformations for each requests only for logs formatting.

Could we rely only on `JSON.stringify` that already drops `undefined` fields and replace every `null` usage by `undefined` usage so it's done out of the box (with a `|| undefined` for every field) ?

lib/utilities/serverAccessLogger.js
dvasilas: Regular logs do not have this duplication; only `accountName` exists.

Here is how the values were being set:
`accountName`:
https://github.com/scality/cloudserver/blob/d4147d8580676cd440c0ca68a97c35ac12f43e53/lib/api/api.js#L253
->
https://github.com/scality/cloudserver/blob/d4147d8580676cd440c0ca68a97c35ac12f43e53/lib/api/api.js#L262
->
https://github.com/scality/cloudserver/blob/d4147d8580676cd440c0ca68a97c35ac12f43e53/lib/utilities/serverAccessLogger.js#L415

`accountDisplayName`:
https://github.com/scality/cloudserver/blob/d4147d8580676cd440c0ca68a97c35ac12f43e53/lib/utilities/serverAccessLogger.js#L416

So 
`accountName` comes from `userInfo.getAccountDisplayName()`
and `accountDisplayName` comes from `request.serverAccessLog.authInfo.getAccountDisplayName()`

and
https://github.com/scality/cloudserver/blob/d4147d8580676cd440c0ca68a97c35ac12f43e53/lib/api/api.js#L241

so they come from calling the same method on the same object (I didn't try to test all scenarios).

lib/utilities/serverAccessLogger.js
dvasilas: Done.

BourgoisMickael (approved):

jonathan-gramain (approved):

PR #6043, opened 2026-01-08
 ✨(CLDSRV-813) update CloudServer XML ListObjectsV2 to support optional attributes

DarkIsDude:
### Related issues

https://scality.atlassian.net/browse/CLDSRV-813

### DIFF
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -264,25 +264,36 @@ function processMasterVersions(bucketName, listParams, list) {
 
 function processOptionalAttributes(item, optionalAttributes) {
     const xml = [];
+    const userMetadata = new Set();
 
-    for (const key of Object.keys(item)) {
-        if (key.startsWith('x-amz-meta-')) {
-            if (optionalAttributes.includes('x-amz-meta-*') || optionalAttributes.includes(key)) {
-                xml.push(`<${key}>${item[key]}</${key}>`);
-            }
+    for (const attribute of optionalAttributes) {
+        switch (attribute) {
+            case 'RestoreStatus':
+                xml.push('<RestoreStatus>');
+                xml.push(`<IsRestoreInProgress>${!!item.restoreStatus?.inProgress}</IsRestoreInProgress>`);
+
+                if (item.restoreStatus?.expiryDate) {
+                    xml.push(`<RestoreExpiryDate>${item.restoreStatus?.expiryDate}</RestoreExpiryDate>`);
+                }
+
+                xml.push('</RestoreStatus>');
+                break;
+            case 'x-amz-meta-*':
+                for (const key of Object.keys(item.userMetadata)) {
+                    userMetadata.add(key);
+                }
+                break;
+            default:
+                if (item.userMetadata?.[attribute]) {
+                    userMetadata.add(attribute);
+                }
         }
     }
 
-    if (optionalAttributes.includes('RestoreStatus')) {
-        xml.push('<RestoreStatus>');
-        xml.push(`<IsRestoreInProgress>${!!item.restoreStatus?.inProgress}</IsRestoreInProgress>`);
-
-        if (item.restoreStatus?.expiryDate) {
-            xml.push(`<RestoreExpiryDate>${item.restoreStatus?.expiryDate}</RestoreExpiryDate>`);
-        }
-
-        xml.push('</RestoreStatus>');
+    for (const key of userMetadata) {
+        xml.push(`<${key}>${item.userMetadata[key]}</${key}>`);
     }
+
 
     return xml;
 }
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -294,7 +294,6 @@ function processOptionalAttributes(item, optionalAttributes) {
         xml.push(`<${key}>${item.userMetadata[key]}</${key}>`);
     }
 
-
     return xml;
 }
 
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -138,6 +138,7 @@ function processVersions(bucketName, listParams, list) {
         const objectKey = escapeXmlFn(item.key);
         const isLatest = lastKey !== objectKey;
         lastKey = objectKey;
+
         xml.push(
             v.IsDeleteMarker ? '<DeleteMarker>' : '<Version>',
             `<Key>${objectKey}</Key>`,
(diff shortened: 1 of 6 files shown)


lib/api/bucketGet.js
```diff
@@ -253,20 +262,48 @@ function processMasterVersions(bucketName, listParams, list) {
...
+function processOptionalAttributes(item, optionalAttributes) {
+    const xml = [];
+
+    for (const key of Object.keys(item)) {
+        if (key.startsWith('x-amz-meta-')) {
+            if (optionalAttributes.includes('x-amz-meta-*') || optionalAttributes.includes(key)) {
+                xml.push(`<${key}>${item[key]}</${key}>`);
+            }
+        }
+    }
```
francoisferrand: it would be more efficient to go the other way round:
- e.g. go through every `optionalAttributes` (which are more likely not as numerous as the fields in our metadata objects)
- in particular, this also ensure we keep the "usual" case (e.g. without optionalAttributes) optimized.
- it avoids the repeated lookups in `optionalAttributes`

```
for(const attr  in optionalAttributes) {
   switch (attr) {
   case 'RestoreStatus':
     [...]
	 break;

   case 'x-amz-meta-*':
     for ((key, value) in item.getUserMetadata()) {
        xml.push(`<${key}>${item[key]}</${key}>`);
     }
     break;

   default:
     const v = item[attr];
     if (v !== undefined) {
         xml.push(....)
     }
     break;
   }

```

The `x-amz-meta-*` case is a bit more concerning, since
- it needs to go through every field to get the every user metadata (but there is already the `ObjectMD.getUserMetadata()` function to do this)
- there could be an issue if both `x-amz-meta-*` and specific fields are specified ; but this should be rejected when validating the request?

This approach is also more future-proof, as it will allow to ad more fields by simply adding them to the `optionalAttributes` list during validation (i.e. less coupling, only need to update one place if we add more fields later)

tests/functional/aws-node-sdk/test/object/mpuVersion.js
```diff
@@ -101,14 +101,15 @@ function checkObjMdAndUpdate(objMDBefore, objMDAfter, props) {
     });
 }
 
-function clearUploadIdFromVersions(versions) {
+function clearUploadIdFromVersionsAndRestoreStatus(versions) {
```
francoisferrand:
```suggestion
function clearUploadIdAndRestoreStatusFromVersions(versions) {
```

tests/functional/aws-node-sdk/test/object/mpuVersion.js
```diff
@@ -101,14 +101,15 @@ function checkObjMdAndUpdate(objMDBefore, objMDAfter, props) {
...
     if (!versions || versions.length === 0) {
         return versions;
     }
 
     for (const version of versions) {
         if (version.value) {
             version.value.uploadId = undefined;
+            version.value.restoreStatus = undefined;
```
francoisferrand, line 112: why do we need to clear `RestoreStatus` from all of these tests?

`RestoreStatus` is only set by cold storage, so will not be set in most case : so seems best to keep this "as is" and handle the cold-storage tests

tests/functional/aws-node-sdk/test/object/putVersion.js
```diff
@@ -34,6 +34,21 @@ function putObjectVersion(s3, params, vid, next) {
...
 }
 
+function clearRestoreStatus(versions) {
+    if (!versions || versions.length === 0) {
+        return versions;
+    }
+
+    for (const version of versions) {
+        if (version.value) {
+            version.value.restoreStatus = undefined;
```
francoisferrand, line 44: why do we need to clear RestoreStatus in every test?

all APIs should behave the same as before, so you should not need to tweak the tests
(and resetting `restoreStatus` may be expected -if anywhere- only in tests where you do actually invoke the list object v2 api with optional attributes)

tests/unit/api/bucketGet.js
```diff
@@ -514,5 +514,99 @@ describe('bucketGet API V2', () => {
                 done();
             });
         });
+
```
francoisferrand, line 442: the `x-amz-optional-attributes header` tests from previous PR should be updated: some tests where actually not really testing much (or duplicated) since the result was not observable...

tests/unit/api/bucketGet.js
```diff
@@ -514,5 +514,99 @@ describe('bucketGet API V2', () => {
                 done();
             });
```
francoisferrand, line 440: would be nice to have some functional tests as well : i.e. tests using the AWS sdk to retrieve the values from a running cloudserver service.

As an extra benefit, it would also provide an exemple of how to use the API.

tests/unit/api/bucketGet.js
DarkIsDude, line 1: Add a new test on lowerCase (all usermetadata are lowercase)

tests/functional/aws-node-sdk/test/object/mpuVersion.js
DarkIsDude: The issue is that we call directly `metadata.listObject` in tests. Regarding arsenal PR, the restoreStatus attribute will be returned as soon as we have an archive field.

And we have a diff. The issue is that the versionBefore do a `fakeMetadataArchive` that update the object. So the version before contains it. This is not done in the second one (versionAfter) and so we have this diff. Regarding the test, that make sense and our API don't change, here we have this diff just because we don't call our API but internal arsenal function.

tests/functional/aws-node-sdk/test/object/putVersion.js
DarkIsDude: https://github.com/scality/cloudserver/pull/6043#discussion_r2686454126

lib/api/bucketGet.js
```diff
@@ -138,6 +138,7 @@ function processVersions(bucketName, listParams, list) {
         const objectKey = escapeXmlFn(item.key);
         const isLatest = lastKey !== objectKey;
         lastKey = objectKey;
+
```
benzekrimaha, line 141: maybe for the future but IMO it's better to have a separate linting/prettier commit to not have both changes in the same one

tests/unit/api/bucketGet.js
DarkIsDude: That's done, thanks for the suggestion. It was a good idea 🙏

lib/api/bucketGet.js
DarkIsDude: I think the PR is small enough to integrate it. I like to update my code to new standard when I work on a file. The idea is for me to don't refactor a full file in one shot but when I work on a piece of code I do it. Thanks to that I reduce the tech debt and do it piece by piece (instead of one big PR dedicated to it). If it's a blocker for you, I can change the PR but if not, I would like to keep it 🙏. Let me know

benzekrimaha: Not a blocker , and no separate PR needed , just a separate commit would be better IMO in the future. And yes it is something we all need to be doing indeed when working on legacy code as it helps a lot

tests/functional/aws-node-sdk/test/bucket/get.js
```diff
@@ -490,5 +498,176 @@ describe('GET Bucket - AWS.S3.listObjects', () => {
...
+                        'x-amz-meta-*,RestoreStatus,x-amz-meta-department',
+                    );
+                    throw new Error('Request should have been rejected');
+                } catch (err) {
+                    assert.strictEqual(err.statusCode, 403);
+                    assert.strictEqual(err.code, 'AccessDenied');
+                }
+            });
+
+            it('should return an XML if the header is only RestoreStatus even without permission', async () => {
```
benzekrimaha:
```suggestion
            it('should always return an XML when the header is RestoreStatus, async () => {
```

francoisferrand (approved):

benzekrimaha (approved):
LGTM

PR #6044, opened 2026-01-14
Cloudserver adaptation post arsenal refacto

benzekrimaha:
Issue: CLDSRV-826

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
     "@smithy/node-http-handler": "^3.0.0",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.3.0-preview.1",
+    "arsenal": "git+https://github.com/scality/Arsenal#8.3.0",
     "async": "2.6.4",
     "bucketclient": "scality/bucketclient#8.2.7",
     "bufferutil": "^4.0.8",
--- a/tests/functional/raw-node/test/GCP/bucket/get.js
+++ b/tests/functional/raw-node/test/GCP/bucket/get.js
@@ -1,9 +1,16 @@
 const assert = require('assert');
 const async = require('async');
 const arsenal = require('arsenal');
+const {
+    ListObjectsCommand,
+    HeadBucketCommand,
+    CreateBucketCommand,
+    DeleteBucketCommand,
+    PutObjectCommand,
+    DeleteObjectCommand,
+} = require('@aws-sdk/client-s3');
 const { GCP } = arsenal.storage.data.external.GCP;
-const { makeGcpRequest } = require('../../../utils/makeRequest');
-const { gcpRequestRetry, genUniqID } = require('../../../utils/gcpUtils');
+const { genUniqID, gcpRetry } = require('../../../utils/gcpUtils');
 const { getRealAwsConfig } =
     require('../../../../aws-node-sdk/test/support/awsConfig');
 const { listingHardLimit } = require('../../../../../../constants');
@@ -18,95 +25,90 @@ const gcpClient = new GCP(config);
 function populateBucket(createdObjects, callback) {
     process.stdout.write(
         `Putting ${createdObjects.length} objects into bucket\n`);
-    async.mapLimit(createdObjects, 10,
-    (object, moveOn) => {
-        makeGcpRequest({
-            method: 'PUT',
-            bucket: bucketName,
-            objectKey: object,
-            authCredentials: config.credentials,
-        }, err => {
-            moveOn(err);
-        });
-    }, err => {
-        if (err) {
-            process.stdout
-                .write(`err putting objects ${err.code}\n`);
+    async.mapLimit(
+        createdObjects,
+        10,
+        async object => {
+            const command = new PutObjectCommand({
+                Bucket: bucketName,
+                Key: object,
+            });
+            await gcpClient.send(command);
+        },
+        err => {
+            if (err) {
+                process.stdout
+                    .write(`err putting objects ${err}\n`);
+            }
+            return callback(err);
         }
-        return callback(err);
-    });
+    );
 }
 
 function removeObjects(createdObjects, callback) {
     process.stdout.write(
         `Deleting ${createdObjects.length} objects from bucket\n`);
-    async.mapLimit(createdObjects, 10,
-    (object, moveOn) => {
-        makeGcpRequest({
-            method: 'DELETE',
-            bucket: bucketName,
-            objectKey: object,
-            authCredentials: config.credentials,
-        }, err => moveOn(err));
-    }, err => {
-        if (err) {
-            process.stdout
-                .write(`err deleting objects ${err.code}\n`);
+    async.mapLimit(
+        createdObjects,
+        10,
+        async object => {
+            const command = new DeleteObjectCommand({
+                Bucket: bucketName,
+                Key: object,
+            });
+            await gcpClient.send(command);
+        },
+        err => {
+            if (err) {
+                process.stdout
+                    .write(`err deleting objects ${err}\n`);
+            }
+            return callback(err);
         }
-        return callback(err);
-    });
+    );
 }
 
 describe('GCP: GET Bucket', function testSuite() {
     this.timeout(180000);
 
-    before(done => {
-        gcpRequestRetry({
-            method: 'PUT',
-            bucket: bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
-                process.stdout.write(`err in creating bucket ${err}\n`);
-            }
-            return done(err);
-        });
+    before(async () => {
+        await gcpRetry(
+            gcpClient,
+            new CreateBucketCommand({ Bucket: bucketName }),
+        );
     });
 
-    after(done => {
-        gcpRequestRetry({
-            method: 'DELETE',
-            bucket: bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
-                process.stdout.write(`err in deleting bucket ${err}\n`);
-            }
-            return done(err);
-        });
+    after(async () => {
+        await gcpRetry(
+            gcpClient,
+            new DeleteBucketCommand({ Bucket: bucketName }),
+        );
     });
 
     describe('without existing bucket', () => {
-        it('should return 404 and NoSuchBucket', done => {
+        it('should return 404 and NoSuchBucket', async () => {
             const badBucketName = `nonexistingbucket-${genUniqID()}`;
-            gcpClient.getBucket({
-                Bucket: badBucketName,
-            }, err => {
+            try {
+                const command = new HeadBucketCommand({
+                    Bucket: badBucketName,
+                });
+                await gcpClient.send(command);
+                assert.fail('Expected NoSuchBucket error, but got success');
+            } catch (err) {
                 assert(err);
-                assert.strictEqual(err.$metadata?.httpStatusCode, 404);
-                assert.strictEqual(err.name, 'NoSuchBucket');
-                return done();
-            });
+                const statusCode = err.$metadata && err.$metadata.httpStatusCode;
+                assert.strictEqual(statusCode, 404);
+                const errorName = err.name === 'NotFound' ? 'NoSuchBucket' : err.name;
+                assert.strictEqual(errorName, 'NoSuchBucket');
+            }
         });
 
-        it('should return 200', done => {
-            gcpClient.listObjects({
+        it('should return 200', async () => {
+            const command = new ListObjectsCommand({
                 Bucket: bucketName,
-            }, (err, res) => {
-                assert.equal(err, null, `Expected success, but got ${err}`);
-                assert.strictEqual(res.$metadata?.httpStatusCode, 200);
-                return done();
             });
+            const res = await gcpClient.send(command);
+            assert.strictEqual(res.$metadata?.httpStatusCode, 200);
         });
     });
 
@@ -119,27 +121,22 @@ describe('GCP: GET Bucket', function testSuite() {
 
             after(done => removeObjects(createdObjects, done));
 
-            it(`should list all ${smallSize} created objects`, done => {
-                gcpClient.listObjects({
+            it(`should list all ${smallSize} created objects`, async () => {
+                const command = new ListObjectsCommand({
                     Bucket: bucketName,
-                }, (err, res) => {
-                    assert.equal(err, null, `Expected success, but got ${err}`);
-                    assert.strictEqual(res.Contents.length, smallSize);
-                    return done();
                 });
+                const res = await gcpClient.send(command);
+                assert.strictEqual(res.Contents.length, smallSize);
             });
 
             describe('with MaxKeys at 10', () => {
-                it('should list MaxKeys number of objects', done => {
-                    gcpClient.listObjects({
+                it('should list MaxKeys number of objects', async () => {
+                    const command = new ListObjectsCommand({
                         Bucket: bucketName,
                         MaxKeys: 10,
-                    }, (err, res) => {
-                        assert.equal(err, null,
-                            `Expected success, but got ${err}`);
-                        assert.strictEqual(res.Contents.length, 10);
-                        return done();
                     });
+                    const res = await gcpClient.send(command);
+                    assert.strictEqual(res.Contents.length, 10);
                 });
             });
         });
@@ -152,15 +149,12 @@ describe('GCP: GET Bucket', function testSuite() {
 
             after(done => removeObjects(createdObjects, done));
 
-            it('should list at max 1000 of objects created', done => {
-                gcpClient.listObjects({
+            it('should list at max 1000 of objects created', async () => {
+                const command = new ListObjectsCommand({
                     Bucket: bucketName,
-                }, (err, res) => {
-                    assert.equal(err, null, `Expected success, but got ${err}`);
-                    assert.strictEqual(res.Contents.length,
-                        listingHardLimit);
-                    return done();
                 });
+                const res = await gcpClient.send(command);
+                assert.strictEqual(res.Contents.length, listingHardLimit);
             });
 
             describe('with MaxKeys at 1001', () => {
@@ -175,17 +169,13 @@ describe('GCP: GET Bucket', function testSuite() {
                 //
                 // Actual behavior: it returns a list longer than 1000 objects when
                 // max-keys is greater than 1000
-                it.skip('should list at max 1000, ignoring MaxKeys', done => {
-                    gcpClient.listObjects({
+                it.skip('should list at max 1000, ignoring MaxKeys', async () => {
+                    const command = new ListObjectsCommand({
                         Bucket: bucketName,
                         MaxKeys: 1001,
-                    }, (err, res) => {
-                        assert.equal(err, null,
-                            `Expected success, but got ${err}`);
-                        assert.strictEqual(res.Contents.length,
-                            listingHardLimit);
-                        return done();
                     });
+                    const res = await gcpClient.send(command);
+                    assert.strictEqual(res.Contents.length, listingHardLimit);
                 });
             });
         });
--- a/tests/functional/raw-node/test/GCP/bucket/getVersioning.js
+++ b/tests/functional/raw-node/test/GCP/bucket/getVersioning.js
@@ -1,106 +1,64 @@
 const assert = require('assert');
-const async = require('async');
 const arsenal = require('arsenal');
+const {
+    PutBucketVersioningCommand,
+    GetBucketVersioningCommand,
+    CreateBucketCommand,
+    DeleteBucketCommand,
+} = require('@aws-sdk/client-s3');
 const { GCP } = arsenal.storage.data.external.GCP;
-const { makeGcpRequest } = require('../../../utils/makeRequest');
-const { gcpRequestRetry, genUniqID } = require('../../../utils/gcpUtils');
+const { genUniqID, gcpRetry } = require('../../../utils/gcpUtils');
 const { getRealAwsConfig } =
     require('../../../../aws-node-sdk/test/support/awsConfig');
 
 const credentialOne = 'gcpbackend';
 const verEnabledObj = 'Enabled';
 const verDisabledObj = 'Suspended';
-const xmlEnable =
-    '<?xml version="1.0" encoding="UTF-8"?>' +
-    '<VersioningConfiguration>' +
-    '<Status>Enabled</Status>' +
-    '</VersioningConfiguration>';
-const xmlDisable =
-    '<?xml version="1.0" encoding="UTF-8"?>' +
-    '<VersioningConfiguration>' +
-    '<Status>Suspended</Status>' +
-    '</VersioningConfiguration>';
 
 describe('GCP: GET Bucket Versioning', () => {
     const config = getRealAwsConfig(credentialOne);
     const gcpClient = new GCP(config);
 
-    beforeEach(function beforeFn(done) {
+    beforeEach(async function beforeFn() {
         this.currentTest.bucketName = `somebucket-${genUniqID()}`;
-        gcpRequestRetry({
-            method: 'PUT',
-            bucket: this.currentTest.bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
-                process.stdout.write(`err in creating bucket ${err.code}\n`);
-            }
-            return done(err);
-        });
+        await gcpRetry(
+            gcpClient,
+            new CreateBucketCommand({
+                Bucket: this.currentTest.bucketName,
+            }),
+        );
     });
 
-    afterEach(function afterFn(done) {
-        gcpRequestRetry({
-            method: 'DELETE',
-            bucket: this.currentTest.bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
-                process.stdout.write(`err in deleting bucket ${err.code}\n`);
-            }
-            return done(err);
-        });
+    afterEach(async function afterFn() {
+        await gcpRetry(
+            gcpClient,
+            new DeleteBucketCommand({
+                Bucket: this.currentTest.bucketName,
+            }),
+        );
     });
 
-    it('should verify bucket versioning is enabled', function testFn(done) {
-        return async.waterfall([
-            next => makeGcpRequest({
-                method: 'PUT',
-                bucket: this.test.bucketName,
-                authCredentials: config.credentials,
-                queryObj: { versioning: '' },
-                requestBody: xmlEnable,
-            }, err => {
-                if (err) {
-                    process.stdout.write(`err in setting versioning ${err.code}`);
-                }
-                return next(err);
-            }),
-            next => {
-                gcpClient.getBucketVersioning({
-                    Bucket: this.test.bucketName,
-                }, (err, res) => {
-                    assert.equal(err, null,
-                        `Expected success, but got err ${err}`);
-                    assert.deepStrictEqual(res.Status, verEnabledObj);
-                    return next();
-                });
-            },
-        ], err => done(err));
+    it('should verify bucket versioning is enabled', async function testFn() {
+        await gcpClient.send(new PutBucketVersioningCommand({
+            Bucket: this.test.bucketName,
+            VersioningConfiguration: { Status: 'Enabled' },
+        }));
+
+        const res = await gcpClient.send(new GetBucketVersioningCommand({
+            Bucket: this.test.bucketName,
+        }));
+        assert.deepStrictEqual(res.Status, verEnabledObj);
     });
 
-    it('should verify bucket versioning is disabled', function testFn(done) {
-        return async.waterfall([
-            next => makeGcpRequest({
-                method: 'PUT',
-                bucket: this.test.bucketName,
-                authCredentials: config.credentials,
-                queryObj: { versioning: '' },
-                requestBody: xmlDisable,
-            }, err => {
-                if (err) {
-                    process.stdout.write(`err in setting versioning ${err}`);
-                }
-                return next(err);
-            }),
-            next => gcpClient.getBucketVersioning({
-                Bucket: this.test.bucketName,
-            }, (err, res) => {
-                assert.equal(err, null,
-                    `Expected success, but got err ${err}`);
-                assert.deepStrictEqual(res.Status, verDisabledObj);
-                return next();
-            }),
-        ], err => done(err));
+    it('should verify bucket versioning is disabled', async function testFn() {
+        await gcpClient.send(new PutBucketVersioningCommand({
+            Bucket: this.test.bucketName,
+            VersioningConfiguration: { Status: 'Suspended' },
+        }));
+
+        const res = await gcpClient.send(new GetBucketVersioningCommand({
+            Bucket: this.test.bucketName,
+        }));
+        assert.deepStrictEqual(res.Status, verDisabledObj);
     });
 });
--- a/tests/functional/raw-node/test/GCP/bucket/head.js
+++ b/tests/functional/raw-node/test/GCP/bucket/head.js
@@ -1,9 +1,13 @@
 const assert = require('assert');
 const arsenal = require('arsenal');
 const { GCP } = arsenal.storage.data.external.GCP;
-const { gcpRequestRetry, genUniqID } = require('../../../utils/gcpUtils');
+const { genUniqID, gcpRetry } = require('../../../utils/gcpUtils');
 const { getRealAwsConfig } =
     require('../../../../aws-node-sdk/test/support/awsConfig');
+const {
+    CreateBucketCommand,
+    DeleteBucketCommand,
+} = require('@aws-sdk/client-s3');
 
 const credentialOne = 'gcpbackend';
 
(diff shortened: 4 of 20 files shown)


tests/functional/raw-node/test/GCP/bucket/get.js
```diff
@@ -15,6 +16,37 @@ const bigSize = listingHardLimit + 1;
 const config = getRealAwsConfig(credentialOne);
 const gcpClient = new GCP(config);
 
+gcpClient.listObjects = (params, callback) => {
+    const command = new ListObjectsCommand(params);
+    return gcpClient.send(command)
+        .then(data => callback(null, data))
+        .catch(err => {
+            if (err && err.$metadata && err.$metadata.httpStatusCode &&
```
SylvainSenechal: no need to check if err exists here. I think err.$metadata.httpstatus code will always be defined.
Probably same for statusCode, gotta call the api once to verify but I feel like we shouldn't have to check if its defined or not

tests/functional/raw-node/test/GCP/object/completeMpu.js
```diff
@@ -100,6 +101,19 @@ describe('GCP: Complete MPU', function testSuite() {
     before(done => {
         config = getRealAwsConfig(credentialOne);
         gcpClient = new GCP(config);
+        gcpClient.listObjects = (params, callback) => {
```
SylvainSenechal: Just to understand : this was available in Arsenal GcpService but you just removed it and now you are redefining it in all these files, why not keeping it in Arsenal ? I think the GcpService already has several functions that are used for tests only 🤔 
Or is it because you want to reassign the statusCode ?

SylvainSenechal (dismissed):
Don't forget to use real arsenal version in package json
Just one comment to address about the gcpClient.listObjects, maybe can consider a utility function at least, or even better keep it in Arsenal

tests/functional/raw-node/test/GCP/object/completeMpu.js
SylvainSenechal: Upon looking : consider updating the gcpClientRetry function : check err.$metadata.statuscode == 429 on top of the existing check err.statusCode == 429. 
This way, can keep listObjects in Arsenal. (unless there are other reason I didn't see)

benzekrimaha: even if we added the `$metadata.httpStatusCode` check inside gcpClientRetry, we still need bespoke error munging in these raw-node tests. Rather than keeping a half-working helper in Arsenal plus a bunch of local overrides, I’d rather keep the shared GcpService minimal and let the tests own whatever extra plumbing they require.

tests/functional/raw-node/test/GCP/bucket/get.js
```diff
@@ -15,6 +16,35 @@ const bigSize = listingHardLimit + 1;
...
+            }
+            if (err.$metadata && err.$metadata.httpStatusCode === 404) {
+                // eslint-disable-next-line no-param-reassign
+                err.name = 'NoSuchBucket';
+            }
+            return callback(err);
+        }
+        return callback(null, res);
+    });
+
```
francoisferrand: why keep these "legacy" functions, and not migrate the actual tests to the new SDK api?
since this PR is focused, it does not seem like a large change...

tests/functional/raw-node/test/GCP/object/completeMpu.js
```diff
@@ -100,6 +101,19 @@ describe('GCP: Complete MPU', function testSuite() {
...
+            const command = new ListObjectsCommand(params);
```
francoisferrand: same as above, should migrate each test? (and avoid duplicating this piece of code)

francoisferrand (commented):
in addition to the comments overall, there is something weird in these test: we use GCP client from arsenal, yet at the same time there is a re-implementation of GCP without the sdk (c.f. `makeGcpRequest`, `retryGcpRequest`...)

this should probably be changed, to do things consistently: either with GCP client (to maximize code reuse), ~~or with  a custom implementation,~~ or directly with the GCP sdk (if we want to ensure the test code is independent form the production code, and failure would thus only happen due to change in production code)

tests/functional/raw-node/test/GCP/bucket/get.js
```diff
@@ -18,95 +25,104 @@ const gcpClient = new GCP(config);
...
-        gcpRequestRetry({
-            method: 'PUT',
-            bucket: bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
-                process.stdout.write(`err in creating bucket ${err}\n`);
-            }
-            return done(err);
+        const command = new CreateBucketCommand({
```
francoisferrand: there used to be a `gcpRequestRetry()` : no problem to stop using it, but don't we need to setup retry somehow? or is it useless and handled in the client now?

tests/functional/raw-node/test/GCP/bucket/get.js
```diff
@@ -18,95 +25,104 @@ const gcpClient = new GCP(config);
...
+            Bucket: bucketName,
         });
+        gcpClient.send(command)
+            .then(() => done())
+            .catch(err => {
+                process.stdout.write(`err in creating bucket ${err}\n`);
+                return done(err);
+            });
```
francoisferrand:
```
    before(async () => {
        const command = new CreateBucketCommand({Collapse commentComment on line 
            Bucket: bucketName,
        });
        await gcpClient.send(command);
    });
```

tests/functional/raw-node/test/GCP/bucket/get.js
```diff
@@ -18,95 +25,104 @@ const gcpClient = new GCP(config);
...
+        const command = new DeleteBucketCommand({
+            Bucket: bucketName,
         });
+        gcpClient.send(command)
+            .then(() => done())
+            .catch(err => {
+                process.stdout.write(`err in deleting bucket ${err}\n`);
+                return done(err);
+            });
     });
```
francoisferrand:
```
    after(async () => {
        const command = new DeleteBucketCommand({
            Bucket: bucketName,
        });
        await gcpClient.send(command)
    });
```

tests/functional/raw-node/test/GCP/bucket/getVersioning.js
```diff
@@ -1,106 +1,105 @@
...
-    '<?xml version="1.0" encoding="UTF-8"?>' +
-    '<VersioningConfiguration>' +
-    '<Status>Suspended</Status>' +
-    '</VersioningConfiguration>';
 
 describe('GCP: GET Bucket Versioning', () => {
     const config = getRealAwsConfig(credentialOne);
     const gcpClient = new GCP(config);
 
     beforeEach(function beforeFn(done) {
```
francoisferrand: make function async and use await when sending command (same for `afterEach`)

tests/functional/raw-node/test/GCP/bucket/getVersioning.js
```diff
@@ -1,106 +1,105 @@
...
+                });
+                return gcpClient.send(command)
+                    .then(res => {
+                        assert.deepStrictEqual(res.Status, verEnabledObj);
+                        return next();
+                    })
+                    .catch(err => next(err));
             },
         ], err => done(err));
     });
```
francoisferrand: almost every line is changed, best use async and shorten the test:
```
    it('should verify bucket versioning is enabled', async () => {
           await gcpClient.send(new PutBucketVersioningCommand({
                Bucket: this.test.bucketName,
                VersioningConfiguration: { Status: 'Enabled' },
           }));

           const res = gcpClient.send(new GetBucketVersioningCommand({
                Bucket: this.test.bucketName,
           }));
           assert.deepStrictEqual(res.Status, verEnabledObj);
    });
```

tests/functional/raw-node/test/GCP/bucket/getVersioning.js
```diff
@@ -1,106 +1,105 @@
...
-            }),
-            next => gcpClient.getBucketVersioning({
-                Bucket: this.test.bucketName,
-            }, (err, res) => {
-                assert.equal(err, null,
-                    `Expected success, but got err ${err}`);
-                assert.deepStrictEqual(res.Status, verDisabledObj);
-                return next();
-            }),
+            // Disable versioning using the official SDK client
```
francoisferrand: same

tests/functional/raw-node/test/GCP/bucket/head.js
```diff
@@ -33,33 +37,25 @@ describe('GCP: HEAD Bucket', () => {
...
-                authCredentials: config.credentials,
-            }, 0, (err, res) => {
-                if (err) {
-                    return done(err);
-                }
-                this.currentTest.bucketObj = {
-                    MetaVersionId: res.headers['x-goog-metageneration'],
-                };
-                return done();
+            const cmd = new CreateBucketCommand({
```
francoisferrand: can we drop the "retry", or should we replace it with some other mean to retry?

tests/functional/raw-node/test/GCP/bucket/head.js
```diff
@@ -33,33 +37,25 @@ describe('GCP: HEAD Bucket', () => {
...
+                Bucket: this.currentTest.bucketName,
             });
+            gcpClient.send(cmd)
+                .then(() => done())
+                .catch(err => done(err));
         });
 
         afterEach(function afterFn(done) {
```
francoisferrand: async

tests/functional/raw-node/test/GCP/bucket/putVersioning.js
```diff
@@ -1,110 +1,93 @@
...
 
-    beforeEach(function beforeFn(done) {
-        this.currentTest.bucketName = `somebucket-${genUniqID()}`;
-        gcpRequestRetry({
-            method: 'PUT',
-            bucket: this.currentTest.bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
+    before(done => {
```
francoisferrand: * changing everything below, use `async` and `await`
* can we drop the retry?

tests/functional/raw-node/test/GCP/object/completeMpu.js
francoisferrand: * in _tests_, we should not need a generic wrapper to remap errors : each test is a single scenario, we can update that test to check the excepted result (both error or success)
* even if we really absolutely need a helper, just introduce a regular function but do not modify the gcpClient object
* can't we migrate each test instead of keeping this yet another AWS sdk v2 lookalike api?

tests/functional/raw-node/test/GCP/object/deleteMpu.js
```diff
@@ -40,23 +43,50 @@ describe('GCP: Abort MPU', function testSuite() {
...
+                        .then(() => next())
+                        .catch(err => {
+                            process.stdout.write(
+                                `err in creating bucket (attempt ${attempt + 1}) ${err}\n`);
+                            return next(err);
+                        });
+                },
+            err => {
+                if (err && (err.name === 'SlowDown'
+                    || err.$metadata?.httpStatusCode === 429)
```
francoisferrand: bad indent, this is not at the same indent level as 'err' (nested expression)

```suggestion
                            || err.$metadata?.httpStatusCode === 429)
```

or better yet just keep it on previous line, the line does not seem that long?

tests/functional/raw-node/test/GCP/object/getTagging.js
```diff
@@ -1,65 +1,96 @@
...
+                        process.stdout.write(
+                            `SlowDown creating bucket, retrying in ${delay}ms (attempt ${attempt + 1})\n`);
+                        return setTimeout(() => createBucket(attempt + 1), delay);
+                    }
+                    process.stdout.write(`err in creating bucket ${err}`);
+                    return done(err);
+                });
+        }
+
+        createBucket(0);
```
francoisferrand: can't this be written much more cleanly with `async.retry` or even just modifying the `gcpRequestRetry` to take both a command and client and repeatedly send the command if needed... (that function would not really be specific to gcp, but provide a good separation between retry logic and the test)

tests/functional/raw-node/utils/gcpUtils.js
```diff
@@ -2,38 +2,8 @@ const async = require('async');
 const assert = require('assert');
 const { v4: uuidv4 } = require('uuid');
 
-const { makeGcpRequest } = require('./makeRequest');
-
 const genUniqID = () => uuidv4().replace(/-/g, '');
 
-function gcpRequestRetry(params, retry, callback) {
```
francoisferrand, line 9: - should still keep the retry logic isolated in some helper, instead of re-implementing directly in the tests
- helper may need to support both async & callbacks (depending on where it is used)

```
async function gcpRetry(gcpClient, cmd, retry, cb) {
   if (cb) {
      return callbackify(() => gcpRetry(gcpClient, cmd, retry))(cb);
   }
   [...]
   return await gcpCient.send(cmd)
}
```

tests/functional/raw-node/utils/gcpUtils.js
```diff
@@ -2,38 +2,8 @@ const async = require('async');
...
-        if (err) {
-            if (retry <= maxRetries && err.statusCode === 429) {
-                return gcpClientRetry(fn, params, callback, retry + 1);
-            }
-            return callback(err);
-        }
-        return callback(null, res);
-    });
-}
-
```
francoisferrand, line 36: if we really need helper (listObjects, ...) they should be added here

package.json
```diff
@@ -31,7 +31,7 @@
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
     "@smithy/node-http-handler": "^3.0.0",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.3.0-preview.1",
+    "arsenal": "git+https://github.com/scality/Arsenal#aed443f02efebeb3f5c2e74786baadb5c19eaded",
```
francoisferrand: should be updated with arsenal 8.3.0 now that it is released

DarkIsDude (approved):
Not a lot to say, code is clean with some async/await, congrats 🙏

tests/functional/raw-node/test/GCP/bucket/get.js
```diff
@@ -18,95 +25,98 @@ const gcpClient = new GCP(config);
...
-            moveOn(err);
-        });
-    }, err => {
-        if (err) {
-            process.stdout
-                .write(`err putting objects ${err.code}\n`);
+    async.mapLimit(
+        createdObjects,
+        10,
+        (object, moveOn) => {
```
francoisferrand: nit: async module supports async functions, so the same can be written with async/await:

```
    async.mapLimit(
        createdObjects,
        10,
        async (object) => await gcClient.send(new PutObjectCommand(...)),
        callback,
    );
```

tests/functional/raw-node/utils/gcpUtils.js
```diff
@@ -1,37 +1,49 @@
...
+        shouldRetry = defaultShouldRetry,
+        getDelayMs = attempt => Math.pow(2, attempt) * 1000,
+    } = retryOptions || {};
+
+    let lastError;
+
+    for (let attempt = 0; attempt < maxAttempts; attempt++) {
+        try {
+            const cmd = typeof makeCommand === 'function' ?
+                makeCommand() : makeCommand;
```
francoisferrand: why do we need to support a _function_? There does not seem to be any case where the command is dynamically built.
can't we _always_ receive the actual command to send, and just use it?

(and if we really needed to use a function, no point supporting both a function and a command)

tests/functional/raw-node/test/GCP/bucket/get.js
```diff
@@ -18,95 +25,98 @@ const gcpClient = new GCP(config);
...
-        }, 0, err => {
-            if (err) {
-                process.stdout.write(`err in creating bucket ${err}\n`);
-            }
-            return done(err);
-        });
+    before(async () => {
+        await gcpRetry(
+            gcpClient,
+            () => new CreateBucketCommand({
```
francoisferrand: what is the point of passing a _function_ instead of just the _command_ object?
can't we keep things simple?

tests/functional/raw-node/test/GCP/bucket/head.js
```diff
@@ -12,66 +16,63 @@ describe('GCP: HEAD Bucket', () => {
     const gcpClient = new GCP(config);
 
     describe('without existing bucket', () => {
-        beforeEach(function beforeFn(done) {
+        beforeEach(function beforeFn() {
```
francoisferrand: not aync and no callback : will never terminate?

```suggestion
        beforeEach(async function beforeFn() {
```

tests/functional/raw-node/test/GCP/bucket/head.js
```diff
@@ -12,66 +16,63 @@ describe('GCP: HEAD Bucket', () => {
...
+        afterEach(async function afterFn() {
+            const cmd = new DeleteBucketCommand({
+                Bucket: this.currentTest.bucketName,
             });
+            try {
+                await gcpClient.send(cmd);
+            } catch (err) {
+                process.stdout
+                    .write(`err deleting bucket: ${err.code}\n`);
+            }
```
francoisferrand: the log is not needed (handled by Jest already), and more importantly this catch actually "eats" the error : so failure will not cause the test to fail (as it did before)


```suggestion
            await gcpClient.send(cmd);
```

tests/functional/raw-node/test/GCP/object/delete.js
```diff
@@ -17,45 +22,39 @@ describe('GCP: DELETE Object', function testSuite() {
...
+            gcpClient,
+            () => new DeleteBucketCommand({ Bucket: bucketName }),
+            null,
+            err => {
+                if (err) {
+                    process.stdout.write(`err in deleting bucket ${err}\n`);
+                }
+                return done(err);
+            },
+        );
```
francoisferrand:
```suggestion
after(async () => {
        await gcpRetry(
            gcpClient,
            new DeleteBucketCommand({ Bucket: bucketName }),
       )
});
```

tests/functional/raw-node/test/GCP/object/delete.js
```diff
@@ -17,45 +22,39 @@ describe('GCP: DELETE Object', function testSuite() {
...
+            const cmd = new PutObjectCommand({
+                Bucket: bucketName,
+                Key: objectKey,
             });
+            gcpClient.send(cmd)
+                .then(() => done())
+                .catch(err => {
+                    process.stdout.write(`err in creating object ${err}\n`);
+                    return done(err);
+                });
```
francoisferrand:
```
        beforeEach(async () => {
            const cmd = new PutObjectCommand({
                Bucket: bucketName,
                Key: objectKey,
            });
            await gcpClient.send(cmd);
        });
```

tests/functional/raw-node/test/GCP/object/deleteTagging.js
```diff
@@ -55,50 +47,63 @@ function assertObjectMetaTag(params, callback) {
...
+            expectedMetaMeta
+        );
+
+        const cmd = new PutObjectCommand({
+            Bucket: bucketName,
+            Key: this.currentTest.key,
+            Metadata: metadata,
+        });
+
+        gcpClient.send(cmd)
```
francoisferrand: use await

tests/functional/raw-node/test/GCP/object/deleteTagging.js
```diff
@@ -108,16 +113,17 @@ describe('GCP: DELETE Object Tagging', function testSuite() {
...
-            method: 'DELETE',
-            bucket: bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
-                process.stdout.write(`err in deleting bucket ${err}`);
-            }
-            return done(err);
-        });
+        gcpRetry(
```
francoisferrand: await gcpRetry(..)

tests/functional/raw-node/test/GCP/object/get.js
```diff
@@ -14,77 +20,70 @@ describe('GCP: GET Object', function testSuite() {
...
                 Bucket: bucketName,
                 Key: this.test.key,
-            }, (err, res) => {
-                assert.equal(err, null,
-                    `Expected success, got error ${err}`);
-                assert.strictEqual(res.ETag, this.test.ETag);
-                assert.strictEqual(res.VersionId, this.test.uploadId);
-                return done();
             });
+            gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/get.js
```diff
@@ -14,77 +20,70 @@ describe('GCP: GET Object', function testSuite() {
...
-            }, err => {
-                if (err) {
-                    process.stdout.write(`err in deleting object ${err}\n`);
-                }
-                return done(err);
+            const cmd = new DeleteObjectCommand({
+                Bucket: bucketName,
+                Key: this.currentTest.key,
             });
+            gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/getTagging.js
```diff
@@ -1,65 +1,77 @@
...
-            headers: tagHeader,
-        }, (err, res) => {
-            if (err) {
+
+        const putCmd = new PutObjectCommand({
+            Bucket: bucketName,
+            Key: this.currentTest.key,
+        });
+
+        gcpClient.send(putCmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/getTagging.js
```diff
@@ -69,16 +81,21 @@ describe('GCP: GET Object Tagging', () => {
...
-            if (err) {
-                process.stdout.write(`err in deleting bucket ${err}`);
-            }
-            return done(err);
-        });
+        if (!bucketCreated) {
+            done();
+            return;
+        }
+        gcpRetry(
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/head.js
```diff
@@ -14,58 +19,50 @@ describe('GCP: HEAD Object', function testSuite() {
...
 
     after(done => {
-        gcpRequestRetry({
-            method: 'DELETE',
-            bucket: bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
+        const cmd = new DeleteBucketCommand({ Bucket: bucketName });
+        gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/head.js
```diff
@@ -14,58 +19,50 @@ describe('GCP: HEAD Object', function testSuite() {
...
-                bucket: bucketName,
-                objectKey: this.currentTest.key,
-                authCredentials: config.credentials,
-            }, (err, res) => {
-                if (err) {
+            const cmd = new PutObjectCommand({
+                Bucket: bucketName,
+                Key: this.currentTest.key,
+            });
+            gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/head.js
```diff
@@ -14,58 +19,50 @@ describe('GCP: HEAD Object', function testSuite() {
...
-            makeGcpRequest({
-                method: 'DELETE',
-                bucket: bucketName,
-                objectKey: this.currentTest.key,
-                authCredentials: config.credentials,
+            if (!this.currentTest.key) {
+                done();
+                return;
+            }
+            gcpClient.deleteObject({
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/head.js
```diff
@@ -75,16 +72,20 @@ describe('GCP: HEAD Object', function testSuite() {
...
                 Bucket: bucketName,
                 Key: this.test.key,
-            }, (err, res) => {
-                assert.equal(err, null,
-                    `Expected success, got error ${err}`);
-                assert.strictEqual(res.ETag, this.test.ETag);
-                assert.strictEqual(res.VersionId, this.test.uploadId);
-                return done();
             });
+            gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/put.js
```diff
@@ -14,38 +18,31 @@ describe('GCP: PUT Object', function testSuite() {
...
 
     after(done => {
-        gcpRequestRetry({
-            method: 'DELETE',
-            bucket: bucketName,
-            authCredentials: config.credentials,
-        }, 0, err => {
-            if (err) {
+        const cmd = new DeleteBucketCommand({ Bucket: bucketName });
+        gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/put.js
```diff
@@ -57,20 +54,19 @@ describe('GCP: PUT Object', function testSuite() {
...
-                bucket: bucketName,
-                objectKey: this.currentTest.key,
-                authCredentials: config.credentials,
-            }, 0, (err, res) => {
-                if (err) {
+            const cmd = new PutObjectCommand({
+                Bucket: bucketName,
+                Key: this.currentTest.key,
+            });
+            gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/putTagging.js
```diff
@@ -2,60 +2,55 @@ const assert = require('assert');
...
-            bucket: bucketName,
-            objectKey: this.currentTest.key,
-            authCredentials: config.credentials,
-        }, (err, res) => {
-            if (err) {
+        const cmd = new PutObjectCommand({
+            Bucket: bucketName,
+            Key: this.currentTest.key,
+        });
+        gcpClient.send(cmd)
```
francoisferrand: await

tests/functional/raw-node/test/GCP/object/putTagging.js
```diff
@@ -65,16 +60,17 @@ describe('GCP: PUT Object Tagging', () => {
...
+            gcpClient,
+            () => new DeleteBucketCommand({ Bucket: bucketName }),
+            null,
+            err => {
+                if (err) {
+                    process.stdout.write(`err in deleting bucket ${err}`);
+                }
+                return done(err);
+            },
+        );
```
francoisferrand:
```suggestion
    after(async () => {
         await gcpRetry(
            gcpClient,
            new DeleteBucketCommand({ Bucket: bucketName }),
        );
```

tests/functional/raw-node/test/GCP/object/upload.js
```diff
@@ -32,26 +34,25 @@ describe('GCP: Upload Object', function testSuite() {
...
-            }),
-        err => done(err));
+        async.eachSeries(
+            Object.values(bucketNames),
+            (bucket, next) => gcpRetry(
+                gcpClient,
+                () => new CreateBucketCommand({ Bucket: bucket.Name }),
+                null,
+                next,
+            ),
```
francoisferrand:
```suggestion
            async (bucket) => await gcpRetry(
                gcpClient,
                () => new CreateBucketCommand({ Bucket: bucket.Name }),
            ),
```

tests/functional/raw-node/utils/gcpUtils.js
```diff
@@ -141,13 +153,20 @@ function setBucketClass(storageClass) {
         '</CreateBucketConfiguration>';
 }
 
+function listBucketObjects(gcpClient, params, cb) {
```
francoisferrand: do we really need this helper?
it is really just sending the command, we can migrate the 2 places where it is called, like everywhere else...

benzekrimaha: this change was made based on this comment here : https://github.com/scality/cloudserver/pull/6044#discussion_r2737561559 to avoid having to redefine the functions everytime. switching back to the commands on the tests level.

tests/functional/raw-node/test/GCP/bucket/head.js
francoisferrand: @benzekrimaha this one really looks like a bug : either it should be an async function, or it should call the continuation callback?

tests/functional/raw-node/test/GCP/bucket/head.js
```diff
@@ -12,66 +16,58 @@ describe('GCP: HEAD Bucket', () => {
...
-                bucket: this.currentTest.bucketName,
-                authCredentials: config.credentials,
-            }, 0, err => {
-                if (err) {
-                    process.stdout
-                        .write(`err deleting bucket: ${err.code}\n`);
-                }
-                return done(err);
+        afterEach(async function afterFn() {
+            const cmd = new DeleteBucketCommand({
```
francoisferrand: missing retry (got removed)

tests/functional/raw-node/test/GCP/object/completeMpu.js
```diff
@@ -87,52 +90,73 @@ function emptyBucket(gcpClient, bucketName, cb) {
                 Bucket: bucketName,
                 Key: object.Key,
             };
-            return gcpClientRetry(gcpClient.deleteObject.bind(gcpClient), deleteParams, next);
+            return gcpClient.deleteObject(deleteParams, next);
         }, cb);
     });
 }
 
+function waitForBucketReady(gcpClient, bucketName) {
```
francoisferrand: function should be marked `async` for clarity

tests/functional/raw-node/test/GCP/object/completeMpu.js
```diff
@@ -87,52 +90,73 @@ function emptyBucket(gcpClient, bucketName, cb) {
...
+    const cmd = new HeadBucketCommand({ Bucket: bucketName });
+    return gcpRetry(gcpClient, cmd, {
```
francoisferrand: technically not needed, but best to `await` here to ensure the call stack is correct (when debugging) : c.f. https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration#Step-5-%E2%80%94-Don%E2%80%99t-forget-to-await-returns
 
```suggestion
    return await gcpRetry(gcpClient, cmd, {
```

tests/functional/raw-node/test/GCP/object/completeMpu.js
```diff
@@ -87,52 +90,73 @@ function emptyBucket(gcpClient, bucketName, cb) {
...
+            buckets,
+            async bucket => {
+                const cmd = new CreateBucketCommand({ Bucket: bucket.Name });
+                try {
+                    await gcpRetry(gcpClient, cmd);
+                    await waitForBucketReady(gcpClient, bucket.Name);
+                } catch (err) {
                     process.stdout.write(`err in creating bucket ${err}\n`);
+                    throw err;
                 }
```
francoisferrand: nit: not really needed, the exception will be automatically propagated anyway

tests/functional/raw-node/test/GCP/object/completeMpu.js
```diff
@@ -87,52 +90,73 @@ function emptyBucket(gcpClient, bucketName, cb) {
...
+                    });
                 });
-            }),
-        done);
+                const cmd = new DeleteBucketCommand({ Bucket: bucket.Name });
+                try {
+                    await gcpRetry(gcpClient, cmd);
+                } catch (err) {
+                    process.stdout.write(`err in deleting bucket ${err}\n`);
+                    throw err;
```
francoisferrand: nit: not really needed, the exception will be automatically propagated anyway

tests/functional/raw-node/test/GCP/object/copy.js
```diff
@@ -63,41 +55,38 @@ describe('GCP: COPY Object', function testSuite() {
...
                 },
-                authCredentials: config.credentials,
-            }, (err, res) => {
-                if (err) {
-                    process.stdout.write(`err in creating object ${err}\n`);
-                }
-                this.currentTest.contentHash = res.headers['x-goog-hash'];
-                return done(err);
             });
+            gcpClient.send(cmd)
```
francoisferrand: use await (every line is changed anyway, best make it clean)

tests/functional/raw-node/test/GCP/object/deleteMpu.js
```diff
@@ -37,55 +41,59 @@ describe('GCP: Abort MPU', function testSuite() {
     let config;
     let gcpClient;
 
-    before(done => {
+    function waitForBucketReady(bucketName) {
```
francoisferrand: this function is duplicated, should be moved to utils...

tests/functional/raw-node/test/GCP/object/deleteMpu.js
```diff
@@ -37,55 +41,59 @@ describe('GCP: Abort MPU', function testSuite() {
...
+        const buckets = Object.values(bucketNames);
+        await async.eachSeries(
+            buckets,
+            async bucket => {
+                await gcpRetry(
+                    gcpClient,
+                    new CreateBucketCommand({ Bucket: bucket.Name }),
+                );
+                await waitForBucketReady(bucket.Name);
+            },
```
francoisferrand, line 56: nit: maybe we should have a helper as well `createBuckets` ?

tests/functional/raw-node/test/GCP/object/initiateMpu.js
```diff
@@ -2,105 +2,146 @@ const assert = require('assert');
...
+    const isRetryableCreateError = err => err && (
+        err.name === 'NoSuchBucket'
+        || err.name === 'NotFound'
+        || err.$metadata?.httpStatusCode === 404
+        || err.name === 'SlowDown'
+        || err.$metadata?.httpStatusCode === 429
+    );
 
-    before(done => {
+    function waitForBucketReady(bucketName) {
```
francoisferrand: 3rd implementation

tests/functional/raw-node/test/GCP/object/initiateMpu.js
```diff
@@ -2,105 +2,146 @@ const assert = require('assert');
...
+        const cmd = new HeadBucketCommand({ Bucket: bucketName });
+        return gcpRetry(gcpClient, cmd, {
+            maxAttempts: 6,
+            shouldRetry: err => err && (
+                err.name === 'NoSuchBucket'
+                || err.name === 'NotFound'
+                || err.$metadata?.httpStatusCode === 404
+                || err.name === 'SlowDown'
+                || err.$metadata?.httpStatusCode === 429
+            ),
```
francoisferrand: duplicated with `isRetryableCreateError` above

tests/functional/raw-node/test/GCP/object/upload.js
```diff
@@ -29,84 +33,79 @@ describe('GCP: Upload Object', function testSuite() {
     let config;
     let gcpClient;
 
-    before(done => {
+    function waitForBucketReady(bucketName) {
```
francoisferrand: 4th definition

tests/functional/raw-node/test/GCP/object/upload.js
```diff
@@ -29,84 +33,79 @@ describe('GCP: Upload Object', function testSuite() {
...
-            (bucket, next) => gcpClient.listObjects({
-                Bucket: bucket.Name,
-            }, (err, res) => {
-                assert.equal(err, null,
-                    `Expected success, but got error ${err}`);
-                async.map(res.Contents, (object, moveOn) => {
-                    const deleteParams = {
+    after(async () => {
+        const buckets = Object.values(bucketNames);
+        await async.eachSeries(
```
francoisferrand, line 53: this cleanup code is duplicated, should be refactored into a helper function

tests/functional/raw-node/utils/gcpUtils.js
```diff
@@ -1,49 +1,125 @@
...
+        maxAttempts = 3,
+        shouldRetry = defaultShouldRetry,
+        getDelayMs = attempt => Math.pow(2, attempt) * 1000,
+    } = retryOptions || {};
+
+    let lastError;
+
+    for (let attempt = 0; attempt < maxAttempts; attempt++) {
+        try {
+
```
francoisferrand, line 21: useless empty line

tests/functional/raw-node/utils/gcpUtils.js
```diff
@@ -1,49 +1,125 @@
...
+    || err.name === 'NotFound'
+    || err.$metadata?.httpStatusCode === 404
+    || err.name === 'SlowDown'
+    || err.$metadata?.httpStatusCode === 429
+    || (typeof err.message === 'string'
+        && (err.message.includes('NoSuchBucket')
+            || err.message.includes('unable to complete upload')))
+);
+
+async function gcpUploadWithRetry(gcpClient, params, retryOptions) {
```
francoisferrand: this function is duplicated with the one above and `createMultipartUploadWithRetry` : should have a generic `async function gcpRetryCall(fn, retryOptions, cb)` which calls `fn` a few times, and can be used to implement both `gcpRetry` and the others

tests/functional/raw-node/utils/gcpUtils.js
```diff
@@ -1,49 +1,125 @@
...
+    const retryDelayMs = attempt => (attempt + 1) * 1000;
+    const isRetryableCreateError = err => err && (
+        err.name === 'NoSuchBucket'
+        || err.name === 'NotFound'
+        || err.$metadata?.httpStatusCode === 404
+        || err.name === 'SlowDown'
+        || err.$metadata?.httpStatusCode === 429
+    );
+
+    function createMultipartUploadWithRetry(attempt, cb) {
```
francoisferrand: duplicated in tests as well...

DarkIsDude (approved):
LGTM

tests/functional/raw-node/test/GCP/object/completeMpu.js
```diff
@@ -97,42 +100,38 @@ describe('GCP: Complete MPU', function testSuite() {
...
+            async bucket => {
+                await new Promise((resolve, reject) => {
+                    emptyBucket(gcpClient, bucket.Name, err => {
+                        if (err) {
+                            reject(err);
+                            return;
+                        }
+                        resolve();
+                    });
                 });
```
francoisferrand: nit: prefer promisify

```suggestion
                await promisify(emptyBucket)(gcpClient, bucket.Name);
```

francoisferrand (approved):

PR #6045, opened 2026-01-15
🎨 Async/Await migration on bucketGet and objectGetLegalHold

DarkIsDude:
Issue: CLDSRV-823

### Motivation and context

Opening a first PR to migrate from callback to async/await syntax. A TAD is opened.

https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration

This is a first PR to start the migration. As mention in the TAD, the idea is to migrate to async/await when working on a specific part of our component. Maybe a common goal of one PR per squad per sprint is a good idea.

All idea or remarks are welcome 🙏

### DIFF
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -1,3 +1,4 @@
+const { promisify } = require('util');
 const querystring = require('querystring');
 const { errors, errorInstances, versioning, s3middleware } = require('arsenal');
 const constants = require('../../constants');
@@ -280,8 +281,7 @@ function processOptionalAttributes(item, optionalAttributes) {
     return xml;
 }
 
-function handleResult(listParams, requestMaxKeys, encoding, authInfo,
-    bucketName, list, corsHeaders, log, callback) {
+function handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName, list, log) {
     // eslint-disable-next-line no-param-reassign
     listParams.maxKeys = requestMaxKeys;
     // eslint-disable-next-line no-param-reassign
@@ -297,7 +297,7 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo,
     pushMetric('listBucket', log, { authInfo, bucket: bucketName });
     monitoring.promMetrics('GET', bucketName, '200', 'listBucket');
 
-    return callback(null, res, corsHeaders);
+    return res;
 }
 
 /**
@@ -306,11 +306,9 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo,
  *                               requester's info
  * @param  {object} request - http request object
  * @param  {function} log - Werelogs request logger
- * @param  {function} callback - callback to respond to http request
- *  with either error code or xml response body
- * @return {undefined}
+ * @return {Promise<object>} - object containing xml and additionalResHeaders
  */
-function bucketGet(authInfo, request, log, callback) {
+async function bucketGet(authInfo, request, log) {
     const params = request.query;
     const bucketName = request.bucketName;
     const v2 = params['list-type'];
@@ -328,9 +326,9 @@ function bucketGet(authInfo, request, log, callback) {
     }
 
     if (v2 !== undefined && Number.parseInt(v2, 10) !== 2) {
-        return callback(errorInstances.InvalidArgument.customizeDescription('Invalid ' +
-            'List Type specified in Request'));
+        throw errorInstances.InvalidArgument.customizeDescription('Invalid List Type specified in Request');
     }
+
     if (v2) {
         log.addDefaultFields({ action: 'ListObjectsV2' });
         if (request.serverAccessLog) {
@@ -348,18 +346,14 @@ function bucketGet(authInfo, request, log, callback) {
     const encoding = params['encoding-type'];
     if (encoding !== undefined && encoding !== 'url') {
         monitoring.promMetrics('GET', bucketName, 400, 'listBucket');
-        return callback(errorInstances.InvalidArgument.customizeDescription('Invalid ' +
-            'Encoding Method specified in Request'));
+        throw errorInstances.InvalidArgument.customizeDescription('Invalid Encoding Method specified in Request');
     }
 
     const requestMaxKeys = params['max-keys'] ? Number.parseInt(params['max-keys'], 10) : 1000;
     if (Number.isNaN(requestMaxKeys) || requestMaxKeys < 0) {
         monitoring.promMetrics('GET', bucketName, 400, 'listBucket');
-        return callback(errors.InvalidArgument);
+        throw errors.InvalidArgument;
     }
-    // AWS only returns 1000 keys even if max keys are greater.
-    // Max keys stated in response xml can be greater than actual
-    // keys returned.
     const actualMaxKeys = Math.min(constants.listingHardLimit, requestMaxKeys);
 
     const metadataValParams = {
@@ -388,51 +382,67 @@ function bucketGet(authInfo, request, log, callback) {
         listParams.marker = params.marker;
     }
 
-    standardMetadataValidateBucket(metadataValParams, request.actionImplicitDenies, log, (err, bucket) => {
-        const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
+    let error;
+    let bucket;
 
-        if (err) {
-            log.debug('error processing request', { error: err });
-            monitoring.promMetrics('GET', bucketName, err.code, 'listBucket');
-            return callback(err, null, corsHeaders);
-        }
+    try {
+        const standardMetadataValidateBucketPromised = promisify(standardMetadataValidateBucket);
+        bucket = await standardMetadataValidateBucketPromised(metadataValParams, request.actionImplicitDenies, log);
+    } catch (err) {
+        error = err;
+    }
 
-        if (params.versions !== undefined) {
-            listParams.listingType = 'DelimiterVersions';
-            delete listParams.marker;
-            listParams.keyMarker = params['key-marker'];
-            listParams.versionIdMarker = params['version-id-marker'] ?
-                versionIdUtils.decode(params['version-id-marker']) :
-                undefined;
-        }
+    const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
 
-        if (!requestMaxKeys) {
-            const emptyList = {
-                CommonPrefixes: [],
-                Contents: [],
-                Versions: [],
-                IsTruncated: false,
-            };
-            return handleResult(listParams, requestMaxKeys, encoding, authInfo,
-                bucketName, emptyList, corsHeaders, log, callback);
-        }
+    if (error) {
+        log.debug('error processing request', { error });
+        monitoring.promMetrics('GET', bucketName, error.code, 'listBucket');
+        error.additionalResHeaders = corsHeaders;
+        throw error;
+    }
+    if (params.versions !== undefined) {
+        listParams.listingType = 'DelimiterVersions';
+        delete listParams.marker;
+        listParams.keyMarker = params['key-marker'];
+        listParams.versionIdMarker = params['version-id-marker'] ?
+            versionIdUtils.decode(params['version-id-marker']) :
+            undefined;
+    }
+    if (!requestMaxKeys) {
+        const emptyList = {
+            CommonPrefixes: [],
+            Contents: [],
+            Versions: [],
+            IsTruncated: false,
+        };
+        const res = handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName, emptyList, log);
+        return [res, corsHeaders];
+    }
 
-        return services.getObjectListing(bucketName, listParams, log, (err, list) => {
-            if (err) {
-                log.debug('error processing request', { error: err });
-                monitoring.promMetrics('GET', bucketName, err.code, 'listBucket');
-                return callback(err, null, corsHeaders);
-            }
+    let list;
+    try {
+        list = await promisify(services.getObjectListing)(bucketName, listParams, log);
+    } catch (err) {
+        log.debug('error processing request', { error: err });
+        monitoring.promMetrics('GET', bucketName, err.code, 'listBucket');
 
-            return handleResult(listParams, requestMaxKeys, encoding, authInfo,
-                bucketName, list, corsHeaders, log, callback);
-        });
-    });
-    return undefined;
+        err.additionalResHeaders = corsHeaders;
+        throw err;
+    }
+
+    const res = handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName, list, log);
+    return [res, corsHeaders];
 }
 
 module.exports = {
     processVersions,
     processMasterVersions,
-    bucketGet,
+    bucketGet: (...args) => {
+        const callback = args.at(-1);
+        const argsWithoutCallback = args.slice(0, -1);
+
+        bucketGet(...argsWithoutCallback)
+            .then(result => callback(null, ...result))
+            .catch(err => callback(err, null, err.additionalResHeaders));
+    },
 };
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -308,7 +308,13 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName
  * @param  {function} log - Werelogs request logger
  * @return {Promise<object>} - object containing xml and additionalResHeaders
  */
-async function bucketGet(authInfo, request, log) {
+async function bucketGet(authInfo, request, log, callback) {
+    if (callback) {
+        return bucketGet(authInfo, request, log)
+            .then(result => callback(null, ...result))
+            .catch(err => callback(err, null, err.additionalResHeaders));
+    }
+
     const params = request.query;
     const bucketName = request.bucketName;
     const v2 = params['list-type'];
@@ -437,12 +443,5 @@ async function bucketGet(authInfo, request, log) {
 module.exports = {
     processVersions,
     processMasterVersions,
-    bucketGet: (...args) => {
-        const callback = args.at(-1);
-        const argsWithoutCallback = args.slice(0, -1);
-
-        bucketGet(...argsWithoutCallback)
-            .then(result => callback(null, ...result))
-            .catch(err => callback(err, null, err.additionalResHeaders));
-    },
+    bucketGet,
 };
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -388,24 +388,19 @@ async function bucketGet(authInfo, request, log, callback) {
         listParams.marker = params.marker;
     }
 
-    let error;
     let bucket;
 
     try {
         const standardMetadataValidateBucketPromised = promisify(standardMetadataValidateBucket);
         bucket = await standardMetadataValidateBucketPromised(metadataValParams, request.actionImplicitDenies, log);
     } catch (err) {
-        error = err;
+        log.debug('error processing request', { err });
+        monitoring.promMetrics('GET', bucketName, err.code, 'listBucket');
+        throw err;
     }
 
     const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
 
-    if (error) {
-        log.debug('error processing request', { error });
-        monitoring.promMetrics('GET', bucketName, error.code, 'listBucket');
-        error.additionalResHeaders = corsHeaders;
-        throw error;
-    }
     if (params.versions !== undefined) {
         listParams.listingType = 'DelimiterVersions';
         delete listParams.marker;
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -306,7 +306,9 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName
  *                               requester's info
  * @param  {object} request - http request object
  * @param  {function} log - Werelogs request logger
+ * @param  {function} callback - callback optional to keep backward compatibility
  * @return {Promise<object>} - object containing xml and additionalResHeaders
+ * @throws {Error}
  */
 async function bucketGet(authInfo, request, log, callback) {
     if (callback) {
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -328,9 +328,7 @@ async function bucketGet(authInfo, request, log, callback) {
             .map(attr => attr !== 'RestoreStatus' ? attr.toLowerCase() : attr)
         ?? [];
     if (optionalAttributes.some(attr => !attr.startsWith('x-amz-meta-') && attr != 'RestoreStatus')) {
-        return callback(
-            errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified')
-        );
+        throw errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified');
     }
 
     if (v2 !== undefined && Number.parseInt(v2, 10) !== 2) {
--- a/lib/api/objectGetLegalHold.js
+++ b/lib/api/objectGetLegalHold.js
@@ -24,7 +24,13 @@ const standardMetadataValidateBucketAndObjPromised = (metadataValParams, actionI
  * @param {object} log - Werelogs logger
  * @return {Promise<object>} - object containing xml and additionalResHeaders
  */
-async function objectGetLegalHold(authInfo, request, log) {
+async function objectGetLegalHold(authInfo, request, log, callback) {
+    if (callback) {
+        return objectGetLegalHold(authInfo, request, log)
+            .then(result => callback(null, ...result))
+            .catch(err => callback(err, null, err.additionalResHeaders));
+    }
+
     log.debug('processing request', { method: 'objectGetLegalHold' });
 
     const { bucketName, objectKey, query } = request;
--- a/lib/api/objectGetLegalHold.js
+++ b/lib/api/objectGetLegalHold.js
@@ -1,4 +1,3 @@
-const async = require('async');
 const { errors, errorInstances, s3middleware } = require('arsenal');
 
 const { decodeVersionId, getVersionIdResHeader }
@@ -10,15 +9,22 @@ const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 
 const { convertToXml } = s3middleware.objectLegalHold;
 
+const standardMetadataValidateBucketAndObjPromised = (metadataValParams, actionImplicitDenies, log) =>
+    new Promise((resolve, reject) => {
+        standardMetadataValidateBucketAndObj(metadataValParams, actionImplicitDenies, log, (err, bucket, objectMD) => {
+            if (err) {return reject(err);}
+            return resolve({ bucket, objectMD });
+        });
+    });
+
 /**
  * Returns legal hold status of object
  * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
  * @param {object} request - http request object
  * @param {object} log - Werelogs logger
- * @param {function} callback - callback to server
- * @return {undefined}
+ * @return {Promise<object>} - object containing xml and additionalResHeaders
  */
-function objectGetLegalHold(authInfo, request, log, callback) {
+async function objectGetLegalHold(authInfo, request, log) {
     log.debug('processing request', { method: 'objectGetLegalHold' });
 
     const { bucketName, objectKey, query } = request;
@@ -26,7 +32,7 @@ function objectGetLegalHold(authInfo, request, log, callback) {
     const decodedVidResult = decodeVersionId(query);
     if (decodedVidResult instanceof Error) {
         log.trace('invalid versionId query', { versionId: query.versionId, error: decodedVidResult });
-        return process.nextTick(() => callback(decodedVidResult));
+        throw decodedVidResult;
     }
     const versionId = decodedVidResult;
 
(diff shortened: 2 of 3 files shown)


lib/api/bucketGet.js
```diff
@@ -280,92 +277,86 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo,
...
-    standardMetadataValidateBucket(metadataValParams, request.actionImplicitDenies, log, (err, bucket) => {
-        const corsHeaders = collectCorsHeaders(request.headers.origin,
-            request.method, bucket);
-        if (err) {
-            log.debug('error processing request', { error: err });
-            monitoring.promMetrics(
-                'GET', bucketName, err.code, 'listBucket');
-            return callback(err, null, corsHeaders);
+        const { error, bucket } = await validateBucket(metadataValParams, request.actionImplicitDenies, log);
+        const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
```
SylvainSenechal: I see it was already like this before migration, but can we check the error directly after calling validateBucket, instead of having this collectCorsHeaders call in between ?

lib/api/objectGetLegalHold.js
```diff
@@ -15,26 +14,21 @@ const { convertToXml } = s3middleware.objectLegalHold;
  * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
  * @param {object} request - http request object
  * @param {object} log - Werelogs logger
- * @param {function} callback - callback to server
- * @return {undefined}
+ * @return {Promise<object>} - object containing xml and additionalResHeaders
  */
-function objectGetLegalHold(authInfo, request, log, callback) {
+async function objectGetLegalHold(authInfo, request, log, callback) {
```
SylvainSenechal, line 20: Isn't it weird to have an async function that takes a callback 🤔  Don't we need to change the code where the function is called too (I think you know what you're doing but still wanna know)

lib/api/bucketGet.js
DarkIsDude: It's here because it's needed in the callback `return callback(error, null, corsHeaders);`

lib/api/objectGetLegalHold.js
DarkIsDude: That'a good point. When we are doing the migration it'll be needed to support both case. Here we need the callback to stay compliant and wait for arsenal to support async/await. We can of course change the upstream but then you'll have huge PR

SylvainSenechal (approved):

lib/api/objectGetLegalHold.js
delthas: It seems to me that in any case, we would have to change all callers to become also async functions and _await_ the result of this function? So if we are changing them to become async we can also drop the callback at that time, too?

Maybe a way to progressively async/await is to start with the bottom of the call stack and go up? i.e.:
- HTTP handler for the route -> turn that into async first
- cloudserver functions like objectGet -> then, turn this into async, no callback needed
- arsenal functions -> last, turn those into async, no callback needed?

If we start with objectGet and make objectGet async, we already have to edit the HTTP handlers to be `async` and call `await`, no?

lib/api/objectGetLegalHold.js
```diff
@@ -44,71 +38,69 @@ function objectGetLegalHold(authInfo, request, log, callback) {
...
+            location: objectMD ? objectMD.dataStoreName : undefined,
+        });
+        const verCfg = bucket.getVersioningConfiguration();
+        additionalResHeaders['x-amz-version-id'] = getVersionIdResHeader(verCfg, objectMD);
+
+        return callback(null, xml, additionalResHeaders);
+    } catch (err) {
+        const additionalResHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
+        err.additionalResHeaders = additionalResHeaders;
+        return callback(err);
```
delthas: If we are to keep the callbacks, it would perhaps be nice to make the function purely async without callbacks, e.g. `async function objectGetLegalHold(authInfo, request, log)`, then have some util that turns that function into a callback func?

Like some "`callbackify`" util func that would take a generic `async function(arg1, arg2, arg3, ...) -> (ret1, ret2, ...)` and return an `async function(arg1, arg2, arg3, ..., callback(err, ret1, ret2, ...)) -> void`. (That awaits the result, if no exception calls the callback passing the args, if theres is an exception catches it and passes that as callback err).)

Then we could just make the async function purely async, no callbacks, and just have `module.exports = callbackify(objectGetLegalHold)` when exporting it.

lib/api/bucketGet.js
```diff
@@ -8,14 +8,9 @@ const escapeForXml = s3middleware.escapeForXml;
...
-const { generateToken, decryptToken }
-    = require('../api/apiUtils/object/continueToken');
+const { generateToken, decryptToken } = require('../api/apiUtils/object/continueToken');
 
-// do not url encode the continuation tokens
-const skipUrlEncoding = new Set([
-    'ContinuationToken',
-    'NextContinuationToken',
-]);
+const xmlParamsToSkipUrlEncoding = new Set(['ContinuationToken', 'NextContinuationToken',]);
```
francoisferrand: Not: extra comma at end of array
```suggestion
const xmlParamsToSkipUrlEncoding = new Set(['ContinuationToken', 'NextContinuationToken']);
```

lib/api/bucketGet.js
```diff
@@ -122,17 +117,16 @@ function processVersions(bucketName, listParams, list) {
...
 
-    const escapeXmlFn = listParams.encoding === 'url' ?
-        querystring.escape : escapeForXml;
+    const escapeXmlFn = listParams.encoding === 'url' ? querystring.escape : escapeForXml;
     xmlParams.forEach(p => {
         if (p.value) {
             const val = p.tag !== 'NextVersionIdMarker' || p.value === 'null' ?
-                p.value : versionIdUtils.encode(p.value);
+                p.value :
+                versionIdUtils.encode(p.value);
```
francoisferrand, line 126: Does not seem related to async/await: kind of dilutes the message here (ie show casing the migration)

francoisferrand: (significant part of this change seems to be in that case...)

lib/api/bucketGet.js
```diff
@@ -374,26 +365,30 @@ function bucketGet(authInfo, request, log, callback) {
...
+            log.debug('error processing request', { error: err });
+            monitoring.promMetrics('GET', bucketName, err.code, 'listBucket');
+            return callback(err, null, corsHeaders);
+        }
+
+        const res = handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName, list, log);
+        return callback(null, res, corsHeaders);
+    } catch (err) {
+        log.error('unhandled error in bucketGet', { error: err });
+        return callback(errors.InternalError);
```
francoisferrand: this is a pattern which must not be done, it creates the risk to invoke the callback multiple times : which is very hard to debug...

```
		return callback(null, res, corsHeaders);         // → continuation callback eventually throws
    } catch (err) {                                                        // → caught here
        log.error('unhandled error in bucketGet', { error: err });
        return callback(errors.InternalError);              // → calls the callback again here
    }
```

i.e. never call Javascript continuation callback from within the `try/catch` block

lib/api/bucketGet.js
```diff
@@ -280,92 +277,86 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo,
...
-            'List Type specified in Request'));
-    }
-    if (v2) {
-        log.addDefaultFields({
-            action: 'ListObjectsV2',
-        });
-        if (request.serverAccessLog) {
-            // eslint-disable-next-line no-param-reassign
-            request.serverAccessLog.analyticsAction = 'ListObjectsV2';
+async function bucketGet(authInfo, request, log, callback) {
```
francoisferrand, line 313: this API seems to be only-halfway migrated: it is async, but also takes a callback (and really uses the callback, not just a quick trampoline or wrapper which we can eventually remove

```suggestion
async function bucketGet(authInfo, request, log, callback) {
   if (callback) {
      return bucketGet(authInfo, request, log).then(res => callback(null, res), callback)
   }

  [...]
  if (...) {
     throw ...
  }
  [...]
  return result;
}
```

lib/api/objectGetLegalHold.js
francoisferrand: the purpose of keeping the "dual interface" (async+callback) is to allow progressive migration:
1. first we migrate the function
   - callers does not need to be migrated: if the callback is passed, it behaves as before
2. callers can be migrated though: when no callback is passed, the function is a regular async function
3. eventually, we can remove the callback

Note however that this should not be done by calling the callback inside try block (c.f. earlier comment) : the body should be fully migrated to async, with some isolated trickery to automatically "callbackify" the function if a callback is passed (so step 3 is really removing a few lines)

lib/api/bucketGet.js
DarkIsDude: indeed, I re-read the whole file and try to simplify at the same time 🙏

francoisferrand: can we split the PR, to keep the focus?

lib/api/bucketGet.js
```diff
@@ -267,9 +249,25 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo,
     }
     pushMetric('listBucket', log, { authInfo, bucket: bucketName });
     monitoring.promMetrics('GET', bucketName, '200', 'listBucket');
-    return callback(null, res, corsHeaders);
+    return res;
 }
 
+const validateBucket = (params, denials, log) => new Promise(resolve => {
```
francoisferrand: - we should use promisify instead of manually recoding it?
- is it worth introducing a wrapper? Esp. in such case where it is called just once, I feel it just kind of "hides" the call to `standardMetadataValidateBucket` without much value; and I fear such helper would end up being harder to remove eventually (once migration is complete), because it would require changes the code in multiple places when not needed anymore [vs just searching for `standardMetadataValidateBucket` and updating all uses]

```suggestion
const validateBucket = promisify(standardMetadataValidateBucket);
```

I.e. overall, since this is a temporary state, I would rather not try to make the migrated function look "perfect", but rather embrace the fact this is a transition, and temporarily keep promisify calls in the middle of these...

lib/api/bucketGet.js
```diff
@@ -344,56 +333,65 @@ function bucketGet(authInfo, request, log, callback) {
...
-        });
-    });
-    return undefined;
+    const { error, bucket } = await validateBucket(metadataValParams, request.actionImplicitDenies, log);
+    const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
+
+    if (error) {
+        log.debug('error processing request', { error });
+        monitoring.promMetrics('GET', bucketName, error.code, 'listBucket');
+        error.additionalResHeaders = corsHeaders;
```
francoisferrand: this was not done before: as we kind of "abused" the continuation callback semantics with an 3rd parameter on error...
→ so I wonder if this is a good exemple? or it is on purpose, to show how to handle such tricks as well?
→ to minimize risk, should we first do a refactor to stop doing this (without migrating to async/await); then proceed with a more streamlined migration to async/await?

I don't see how this is handled in "caller" code though (i.e. code "receiving" these CORS/additional headers on error), did I miss it do we already support such use of the `additionalResHeaders` field ?

lib/api/objectGetLegalHold.js
```diff
@@ -44,71 +38,68 @@ function objectGetLegalHold(authInfo, request, log, callback) {
...
-                bucket: bucketName,
-                keys: [objectKey],
-                versionId: objectMD ? objectMD.versionId : undefined,
-                location: objectMD ? objectMD.dataStoreName : undefined,
+                return resolve({ bucket, objectMD });
             });
-            const verCfg = bucket.getVersioningConfiguration();
-            additionalResHeaders['x-amz-version-id'] =
-                getVersionIdResHeader(verCfg, objectMD);
+    });
```
francoisferrand:
```suggestion
    const { bucket, objectMD } = await promisify(standardMetadataValidateBucketAndObj)(
            metadataValParams, request.actionImplicitDenies, log);
```

lib/api/bucketGet.js
francoisferrand: need to update the output parameters as well (and this will be already a step when doing TS migration)

lib/api/bucketGet.js
```diff
@@ -344,56 +333,65 @@ function bucketGet(authInfo, request, log, callback) {
...
+
+    const res = handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName, list, log);
+    return [res, corsHeaders];
 }
 
 module.exports = {
     processVersions,
     processMasterVersions,
-    bucketGet,
+    bucketGet: (...args) => {
```
francoisferrand: should we go that road?
- "rewiring" functions in imports/exports makes the code difficult to read IMHO, as the function is not called the same way it is "defined"
- may not be an issue for this specific function, but this approach does not allow progressively migrating calls: even if the body of the function was updated already, it is still exported only as async or as continuation callback...
- the slicing is dangerous, as it is based on the actual arguments of the call, not the parameters of the function. So if some code calls without callbacK (esp. likely while migrating...) we end up dropping the last (non-callback) parameter

lib/api/bucketGet.js
DarkIsDude: Does it make really sense ? `+160 −171` the PR is pretty small already ?

lib/api/bucketGet.js
DarkIsDude: 🤦 thanks for the spot. I forgot to add that in the callback 🙏.

I'm not sure we should do a first refactor, it'll add delay and it's the goal of a refactor IMO ? But I guess I can add that in our docs on Confluence ? So yes the goal is to make a "good" example.

lib/api/bucketGet.js
DarkIsDude: My goal was also to try to lighten the function as it's already a 100 lines function but I agree with you that it's maybe not the best choices 🙏.

lib/api/bucketGet.js
DarkIsDude: That's a good point. So what is your suggestion ? Should I create a new wrapper `bucketGet` and rename the function `bucketGetAsync` ? Or integrate that directly in the function with a `if (callback) { bucketGet(...).then().catch() }` ? I'm not a big fan of this if as we introduce callback in an async/await code (but can be acceptable during migration I guess?). Or any other idea?

The idea of the slice was to avoid repeating the same arguments in the async and in the callback function but indeed the tradeoff is what you mention. Anyway if some code calls without callback, we'll have an issue in any case.

lib/api/objectGetLegalHold.js
```diff
@@ -44,71 +46,68 @@ function objectGetLegalHold(authInfo, request, log, callback) {
...
 }
 
-module.exports = objectGetLegalHold;
+module.exports = (...args) => {
+    const callback = args.at(-1);
+    const argsWithoutCallback = args.slice(0, -1);
+
+    objectGetLegalHold(...argsWithoutCallback)
+        .then(result => callback(null, ...result))
+        .catch(err => callback(err, null, err.additionalResHeaders));
```
maeldonn, line 112:
```suggestion
        .catch(err => callback(err, null, err.additionalResHeaders ?? {}));
````

It's just defensive, maybe useless...

Comment (maeldonn) [reply lib/api/objectGetLegalHold.js:112]: The same thing can be found in the other file if you need to modify it. 🤷‍♂️

File: lib/api/objectGetLegalHold.js [hunk 13]
```diff
@@ -10,31 +9,34 @@ const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 
 const { convertToXml } = s3middleware.objectLegalHold;
 
+const standardMetadataValidateBucketAndObjPromised = (metadataValParams, actionImplicitDenies, log) =>
```
maeldonn: Instead of doing one custom promisify for each file that is migrated, we can do the same thing in another way:

https://nodejs.org/docs/latest-v22.x/api/util.html#custom-promisified-functions

Doing it at function level will facilitate its migration and reduce impacts.

lib/api/objectGetLegalHold.js
```diff
@@ -10,31 +9,34 @@ const collectCorsHeaders = require('../utilities/collectCorsHeaders');
...
+    });
+
 /**
  * Returns legal hold status of object
  * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
  * @param {object} request - http request object
  * @param {object} log - Werelogs logger
- * @param {function} callback - callback to server
- * @return {undefined}
+ * @return {Promise<object>} - object containing xml and additionalResHeaders
```
maeldonn, line 18: Consider adding @throws documentation for better visibility on error handling.

lib/api/bucketGet.js
```diff
@@ -344,56 +318,74 @@ function bucketGet(authInfo, request, log, callback) {
...
-            return handleResult(listParams, requestMaxKeys, encoding, authInfo,
-                bucketName, list, corsHeaders, log, callback);
-        });
-    });
-    return undefined;
+    let error;
+    let bucket;
+
+    try {
+        const standardMetadataValidateBucketPromised = promisify(standardMetadataValidateBucket);
```
maeldonn, line 394: In a migration context, I think it's easier to keep all promisified functions at the top of the file.

lib/api/bucketGet.js
```diff
@@ -344,56 +318,74 @@ function bucketGet(authInfo, request, log, callback) {
...
+        bucket = await standardMetadataValidateBucketPromised(metadataValParams, request.actionImplicitDenies, log);
+    } catch (err) {
+        error = err;
```
maeldonn: The goal of this variable is to get cors headesr before throwing the error but it adds complexity. You could call collectCorsHeaders in the catch to avoid this.

lib/api/bucketGet.js
maeldonn: And for performance of course

lib/api/bucketGet.js
```diff
@@ -278,28 +261,25 @@ function handleResult(listParams, requestMaxKeys, encoding, authInfo,
  * @param  {function} log - Werelogs request logger
  * @param  {function} callback - callback to respond to http request
```
maeldonn:
```suggestion
```

lib/api/bucketGet.js
```diff
@@ -182,31 +178,19 @@ function processMasterVersions(bucketName, listParams, list) {
...
-            tag: 'NextContinuationToken',
-            value: generateToken(list.NextContinuationToken),
-        });
-        xmlParams.push({
-            tag: 'KeyCount',
-            value: list.Contents ? list.Contents.length : 0,
-        });
+        xmlParams.push({ tag: 'StartAfter', value: listParams.startAfter || '' });
+        xmlParams.push({ tag: 'FetchOwner', value: `${listParams.fetchOwner}` });
+        xmlParams.push({ tag: 'ContinuationToken', value: generateToken(listParams.continuationToken) || '', });
```
maeldonn: trailing comma 👀

lib/api/objectGetLegalHold.js
```diff
@@ -44,71 +46,68 @@ function objectGetLegalHold(authInfo, request, log, callback) {
...
+        throw errorInstances.InvalidRequest.customizeDescription('Bucket is missing Object Lock Configuration');
+    }
+
+    const { legalHold } = objectMD;
+    const xml = convertToXml(legalHold);
+    if (xml === '') {
+        throw errors.NoSuchObjectLockConfiguration;
+    }
+
+    const additionalResHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
```
maeldonn, line 91: To keep the initial behavior, this function should be call earlier to do the err.additionalResHeader trick.

lib/api/bucketGet.js
```diff
@@ -344,56 +318,74 @@ function bucketGet(authInfo, request, log, callback) {
...
+            Contents: [],
+            Versions: [],
+            IsTruncated: false,
+        };
+        const res = handleResult(listParams, requestMaxKeys, encoding, authInfo, bucketName, emptyList, log);
+        return [res, corsHeaders];
+    }
+
+    let list;
+    try {
```
maeldonn, line 424: Not a big fan of multiple try/catch in a single function, could it be simplified ?

lib/api/bucketGet.js
francoisferrand: If it were just any feature or PR, indeed that would be acceptable (though I personally prefer to avoid the noise of useless reformatting)...

but in this specific case, we want the PR to showcase the migration : so best to keep it focused and show _only_ the migration.

We want every body to review this PR, so they have a better sense of what we are getting into; and so we should really cut the noise.

lib/api/bucketGet.js
```diff
@@ -308,21 +288,15 @@ function bucketGet(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'bucketGet' });
     const encoding = params['encoding-type'];
     if (encoding !== undefined && encoding !== 'url') {
-        monitoring.promMetrics(
-            'GET', bucketName, 400, 'listBucket');
-        return callback(errorInstances.InvalidArgument.customizeDescription('Invalid ' +
-            'Encoding Method specified in Request'));
+        monitoring.promMetrics('GET', bucketName, 400, 'listBucket');
+        throw errorInstances.InvalidArgument.customizeDescription('Invalid Encoding Method specified in Request');
```
francoisferrand, line 355: line is quite long (115 chars), do we already have an alignment of changing max line length?

I really would rather avoid having such discussion in this PR which is about async/await...

lib/api/bucketGet.js
francoisferrand: does promisify not already handle "caching" of promisified functions? So performance should be fine either way (keeping in the code avoids the "confusion" which may be associated with duplicated symbols, and makes followup changes/migration easier as only a single line needs to be changed once that function is upgraded to async)

lib/api/bucketGet.js
francoisferrand: > You could call collectCorsHeaders in the catch to avoid this.

this would duplicate the call to `collectCorsHeaders` (leading to higher maintenance), as it is needed in the success path as well

lib/api/bucketGet.js
francoisferrand: > I'm not a big fan of this if as we introduce callback in an async/await code (but can be acceptable during migration I guess?)

clearly not ideal, but seems to me the best compromise for the migration (and _**only**_ for the migration):
- works seamlessly: callback code continue working as before, and (caller) code can be migrated to its "final" form directly (avoids having 2 phases)
- no cognitive load of having multiple function names
- easy to remove (just drop the trampoline at the beginning of the function, no need to change every call site)

> Anyway if some code calls without callback, we'll have an issue in any case.
yes and no. Calling without callback is fine if we `await`, which is were we want to go...
Esp. some functions are called in many modules: so we may want to migrate that function, but only progressively update every place where it is invoked...

lib/api/objectGetLegalHold.js
francoisferrand: we should also add this trick to the confluence page (in the tips & tricks part)

lib/api/objectGetLegalHold.js
```diff
@@ -44,71 +46,68 @@ function objectGetLegalHold(authInfo, request, log, callback) {
...
+    let bucket, objectMD;
+
+    try {
+        ({ bucket, objectMD } = await standardMetadataValidateBucketAndObjPromised(
+            metadataValParams,
+            request.actionImplicitDenies,
+            log,
+        ));
+    } catch (err) {
+        log.trace('request authorization failed', { method: 'objectGetLegalHold', error: err });
```
francoisferrand, line 58: nit: (not sure we change this in this PR)

is this trace really needed? It seems to me such "authorization failed" log should be in standardMetadataValidateBucket (and may be there already...), which would make this code much more palatable...

lib/api/bucketGet.js
DarkIsDude: I'm not a big fan on this mindset. Migrating to async/await is also the opportunity to review formatting and improving the code and asking to separate PR will be harder and will not encourage the migration. We can then say "Yeah but it's the first one", but if it's the first one and the example, then let make how to do it from A to Z... Also the new PR is really small...

But my mantra is to say that reviewer have always last word, so let's agree to disagree and move forward. The new PR is open.

lib/api/bucketGet.js
DarkIsDude: https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration#%E2%9A%A0%EF%B8%8F-Limitation%3A-Multiple-callback-arguments added it here

lib/api/bucketGet.js
DarkIsDude: 🤝 let's go

lib/api/objectGetLegalHold.js
DarkIsDude: Right now it's indeed useless. If we are in the catch, we have an err ? I'm not a big fan of pre-defensive code (defensive code that we are not sure to trigger) as it's most of the time dead code.

lib/api/bucketGet.js
DarkIsDude: To be honest I don't have preference here. Moving it here or at the top should not impact performance as indeed there is a cache (thanks AI). Regarding confusion, I'm not sure that change a lot as `standardMetadataValidateBucketPromised` is pretty clear IMO, same for single line, the PR will be more or less the same. @maeldonn if you have a strong opinion let's reopen the comment and change it

lib/api/objectGetLegalHold.js
DarkIsDude: Added. I don't know if we should. That more or less the goal async/await function. They can throw. And adding all the throw case can be very longue as we are in a handler. What I did is enough for you ?

lib/api/bucketGet.js
DarkIsDude: Agree, here we can easily DRY so better to keep it as it is.

lib/api/objectGetLegalHold.js
DarkIsDude: @maeldonn can you double check ? The err was not added on all errors but only one some. I rechecked and I think I followed the same path as before. Did I miss something ?

lib/api/bucketGet.js
DarkIsDude: CI is passing so I guess yes ? For me if the CI is ok, then I respect our code standard ?

lib/api/objectGetLegalHold.js
DarkIsDude: IMO trace are useful when working locally. If it's there, it's because it was needed for someone one day to debug so I prefer to keep it.

lib/api/bucketGet.js
DarkIsDude: What is your suggestion ? I didn't find a better approach. The first one is useful to catch error but continue the flow  if it's ok (and specially collect headers even if an error).

The second one is classic. I can put everything in the same one but in that case I'll duplicate the collect headers (as I need bucket in the happy path). Didn't find something better.

Multiple try/catch is fine but the error management is more or less the same. I can extract that in a new function if you prefer ?

lib/api/objectGetLegalHold.js
DarkIsDude: Thanks for this trick. Discover it, that's perfect for what we are doing. You can review this commit https://github.com/scality/cloudserver/pull/6045/changes/872cfa351090e1d3716753b4859c821c8f5cdba6 and the confluence page where I added your idea 👏  https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration#%E2%9A%A0%EF%B8%8F-Limitation%3A-Multiple-callback-arguments

lib/api/bucketGet.js
maeldonn: If there is no bypass it's ok for me 🙂

lib/api/objectGetLegalHold.js
maeldonn: Sorry I think I confused with the bucketGet.js file... Just ignore my comment 😅

maeldonn (approved):

lib/api/bucketGet.js
```diff
@@ -336,47 +336,56 @@ function bucketGet(authInfo, request, log, callback) {
         listParams.marker = params.marker;
     }
 
-    standardMetadataValidateBucket(metadataValParams, request.actionImplicitDenies, log, (err, bucket) => {
-        const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
+    let error;
+    let bucket;
```
francoisferrand, line 391: `bucket` _will not_ be set in case `standardMetadataValidateBucketPromised` throws:
- is it already the behavior of standardMetadataValidateBucketPromised' continuation callback, or did it _always_ provide the `bucket` ? In that case promisify does not work... 😢
- does `collectCorsHeaders` work when bucket is not provider?

DarkIsDude: That's a good question. I remember encounter an issue with that some weeks (not for this PR).

The answer can be found inside the `collectCorsHeaders`:

```
    // NOTE: Because collecting CORS headers requires making a call to
    // metadata to retrieve the bucket's CORS configuration, we opt not to
    // return the CORS headers if the request encounters an error before
    // the api method retrieves the bucket from metadata (an example
    // being if a request is not properly authenticated). This is a slight
    // deviation from AWS compatibility, but has the benefit of avoiding
    // additional backend calls for an invalid request. Also, we anticipate
    // that the preflight OPTIONS route will serve most client needs regarding
    // CORS.
    if (!origin || !bucket) {
        return {};
    }
```

So the answer is yes, `collectCorsHeaders` will continue to work but no CORS will be returned. So I'm not sure to understand your first point. The bucket will be defined the same as before (can be, if all goes right), and will not if we have an error. Can you help me to clarify 🙏 ?

francoisferrand: my point was that setting `bucket` outside of the try/catch so it could be used was _fishy_, since it would not be set in case of exception: either the bucket is not needed by collectCorsHeader, or it would not work, or there was a trick in the (preexisting) callback from `standardMetadataValidateBucket` to _always_ pass the bucket, even on error.

Now, from what you found in `collectCorsHeaders()`, this code can actually be greatly simplified, since we don't need to call `collectCorsHeaders` in case of error (it would do nothing)

```suggestion
let bucket
try {
    const standardMetadataValidateBucketPromised = promisify(standardMetadataValidateBucket);
    bucket = await standardMetadataValidateBucketPromised(metadataValParams, request.actionImplicitDenies, log);
} catch (err) {
    log.debug('error processing request', { error });
    monitoring.promMetrics('GET', bucketName, error.code, 'listBucket');
    throw error;
}

const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
```

francoisferrand: (my first point was that **_if_** it worked the callback of `standardMetadataValidateBucket` always received the bucket, even on error, then we could not use promisify directly and would have to do some tricks, e.g. in case `standardMetadataValidateBucket` looked like this:

```js
function standardMetadataValidateBucket(...., cb) {
    [...]
    if (err) {
       return cb(err, bucket);  // failed!! but we still return the bucket
    }
    return cb(undefined, bucket);  // success
}
```

DarkIsDude: Ok got it thanks. I had in mind to simplify the function but this one change the behaviour of the code and don't reflect what was done before. As we said in a previous conversation that we should only focus on async/await and don't change anything else, I didn't implement it. Anyway, it's done 🙏

lib/api/bucketGet.js
```diff
@@ -388,47 +388,51 @@ function bucketGet(authInfo, request, log, callback) {
...
-            return callback(err, null, corsHeaders);
-        }
+    try {
+        const standardMetadataValidateBucketPromised = promisify(standardMetadataValidateBucket);
+        bucket = await standardMetadataValidateBucketPromised(metadataValParams, request.actionImplicitDenies, log);
+    } catch (err) {
+        log.debug('error processing request', { err });
+        monitoring.promMetrics('GET', bucketName, err.code, 'listBucket');
+        throw err;
+    }
```
francoisferrand, line 400: not an issue for this PR, but I wonder if we can write this "more cleanly" (e.g. avoid the separate `let` variable, and make the exception "wrapping" more seamless) with a combination of promise and await? e.g. something like this:

```javascript
const bucket = await standardMetadataValidateBucketPromised(metadataValParams, request.actionImplicitDenies, log).catch(err => {
	log.debug('error processing request', { err });
    monitoring.promMetrics('GET', bucketName, err.code, 'listBucket');
    throw err;
});
```

(ideally, I'd rather avoid such "exception wrapping", but may not be possible in every case...)

what do you think: Does it work? Is there some  Is this a "pattern" we may or should use?
or should we rather a single try/catch over the whole function in such case (which would avoid the duplicated "catch" block for log & prometheus)?

lib/api/objectGetLegalHold.js
```diff
@@ -40,60 +45,69 @@ function objectGetLegalHold(authInfo, request, log, callback) {
...
+        authInfo,
+        bucket: bucketName,
+        keys: [objectKey],
+        versionId: objectMD ? objectMD.versionId : undefined,
+        location: objectMD ? objectMD.dataStoreName : undefined,
     });
+    const verCfg = bucket.getVersioningConfiguration();
+    additionalResHeaders['x-amz-version-id'] = getVersionIdResHeader(verCfg, objectMD);
+
+    return [xml, additionalResHeaders];
```
francoisferrand, line 103: I think we should explicitly add this corner-case in the TAD, to warn user when migrating functions which pass multiple values to callback?

francoisferrand (approved):

lib/api/objectGetLegalHold.js
DarkIsDude: https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration#%E2%9A%A0%EF%B8%8F-Limitation%3A-Multiple-callback-arguments it's already here 🙏

lib/api/bucketGet.js
DarkIsDude: That's a good question and it depends a lot on each case. I prefer for example the try/catch than the mix of promise/await but it's a pure personal opinion. IMO the industry try to move to a more try/catch approach but it's just a feeling. Anyway what you suggest should work.

It's a "migration". So some pattern can not be perfect... A good option to avoid the try/catch can be to type of errors and use that in one catch block (more or less what we did for AWS error). In that case I would suggest to keep the try/catch and don't try to complexify the migration. It's a well balanced between moving forward with the migration and improving our code.

PR #6050, opened 2026-01-16
CLDSRV-821: add backbeat calls to logs

leif-scality:
This PR makes cloudserver write backbeat requests to the server access logs file.

- The `loggingEnabled` field is **always** set to `false` to prevent the log from entering the server-access clickhouse table and it being consumed by log courier.
- The `operation` field is **always** set to `REST.${req.method}.BACKBEAT`
- The `action` field is set to the backbeat route or `BACKBEAT_INVALID` if the route is not valid or does not exists.

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -114,6 +114,10 @@ function _respondWithHeaders(response, payload, extraHeaders, log, callback) {
         'content-type': 'application/json',
         'content-length': Buffer.byteLength(body),
     }, extraHeaders);
+    if (response.serverAccessLog) {
+        // eslint-disable-next-line no-param-reassign
+        response.serverAccessLog.bytesSent = Buffer.byteLength(body);
+    }
     response.writeHead(200, httpHeaders);
     response.end(body, 'utf8', () => {
         log.end().info('responded with payload', {
@@ -1572,6 +1576,10 @@ function routeBackbeat(clientIP, request, response, log) {
     const contentLength = request.headers['x-amz-decoded-content-length'] || request.headers['content-length'];
     // eslint-disable-next-line no-param-reassign
     request.parsedContentLength = Number.parseInt(contentLength?.toString() ?? '', 10);
+
+    log.debug('routing request');
+    _normalizeBackbeatRequest(request);
+
     log.addDefaultFields({
         clientIP,
         url: request.url,
@@ -1582,9 +1590,17 @@ function routeBackbeat(clientIP, request, response, log) {
         bytesReceived: request.parsedContentLength || 0,
         bodyLength: parseInt(request.headers['content-length'], 10) || 0,
     });
+    if (request.serverAccessLog) {
+        // eslint-disable-next-line no-param-reassign
+        request.serverAccessLog.bucketName = request.bucketName;
+        // eslint-disable-next-line no-param-reassign
+        request.serverAccessLog.objectKey = request.objectKey;
+        // eslint-disable-next-line no-param-reassign
+        request.serverAccessLog.backbeat = true;
+        // eslint-disable-next-line no-param-reassign
+        request.serverAccessLog.analyticsAction = 'BACKBEAT_INVALID';
+    }
 
-    log.debug('routing request');
-    _normalizeBackbeatRequest(request);
     const requestContexts = prepareRequestContexts('objectReplicate', request);
 
     if (request.resourceType === 'expiration' || request.resourceType === 'batchdelete') {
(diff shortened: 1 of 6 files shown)


francoisferrand:
Should we log backbeat calls in Bucket Logging?
I am really just asking, because backbeat routes are often kind of "internal" calls (typically used for inter-site replication or lifecycle), and as such not sure what the expected semantics is: do we want the user to actually see everything that is happening on the platform, or only "user-facing" operations (like AWS I guess) ?

scality-gdoumergue:
Hello @leif-scality, can you confirm that backbeat calls will still be seen in Grafana dashboards, for CS and customers to troubleshoot and measure backbeat-backed features performance? Thank you.

lib/server.js
```diff
@@ -126,11 +126,13 @@ class S3Server {
...
 
-        // Skip server access logs for heartbeat and backbeat.
+        // Skip server access logs for heartbeat.
         const isLoggingEnabled = _config.serverAccessLogs
             && (_config.serverAccessLogs.mode === serverAccessLogsModes.LOG_ONLY
                 || _config.serverAccessLogs.mode === serverAccessLogsModes.ENABLED);
-        if (isLoggingEnabled && !req.url.startsWith('/_/')) {
+        const isInternalRoute = req.url.startsWith('/_');
+        const isBackbeatRoute = req.url.startsWith('/_/backbeat/');
+        if (isLoggingEnabled && isInternalRoute ? isBackbeatRoute : true) {
```
dvasilas: nit: Could be refactored to
``` suggestion
        if (isLoggingEnabled && (!isInternalRoute || isBackbeatRoute);) {
```

lib/routes/routeBackbeat.js
```diff
@@ -1642,6 +1658,16 @@ function routeBackbeat(clientIP, request, response, log) {
...
     const isObjectRequest = _isObjectRequest(request);
 
+    if (request.serverAccessLog) {
+        let route = backbeatRoutes[request.method][request.resourceType];
+        if (useMultipleBackend && request.resourceType !== 'multiplebackendmetadata') {
+            route = backbeatRoutes[request.method][request.resourceType][request.query.operation];
+        }
+
+        // eslint-disable-next-line no-param-reassign
+        request.serverAccessLog.analyticsAction = route.name;
```
dvasilas:
``` suggestion
        request.serverAccessLog.analyticsAction = route?.name ?? 'BACKBEAT_INVALID';
```

dvasilas (commented):
Let's add tests for backbeat routes to `tests/unit/utils/serverAccessLogger.js`?

nicolas2bert:
@scality-gdoumergue about 
> Hello @leif-scality, can you confirm that backbeat calls will still be seen in Grafana dashboards, for CS and customers to troubleshoot and measure backbeat-backed features performance? Thank you.


yes, however this improvement https://scality.atlassian.net/browse/RD-1246 won't be there yet. We will be working on it after it is merged.

dvasilas:
> Should we log backbeat calls in Bucket Logging? I am really just asking, because backbeat routes are often kind of "internal" calls (typically used for inter-site replication or lifecycle), and as such not sure what the expected semantics is: do we want the user to actually see everything that is happening on the platform, or only "user-facing" operations (like AWS I guess) ?

@francoisferrand we will not log internal backbeat calls in bucket logging. We add them to the `server-access.logs` file so that they appear in S3 analytics Grafana dashboards, but `loggingEnabled` (controls bucket logging) will be set to `false`.

lib/routes/routeBackbeat.js
leif-scality: Normally route is never undefined or null because it is checked above, but added the check just in case

leif-scality:
@dvasilas added unit tests

nicolas2bert (approved):

lib/utilities/serverAccessLogger.js
```diff
@@ -257,6 +257,10 @@ const methodToResType = Object.freeze({
 
 function getOperation(req) {
     const resourceType = methodToResType[req.apiMethod];
+
+    if (req.serverAccessLog && req.serverAccessLog.backbeat) {
+        return `REST.${req.method}.BACKBEAT`;
```
dvasilas, line 262: I am wondering about `REST.${req.method}.BACKBEAT`.
Backbeat is the name of an S3C service. End-users of service provider customers may have no idea what backbeat is.
Maybe we could use something like `REST.${req.method}.INTERNAL` ?
@nicolas2bert  @davidtencer what do you think?

dvasilas: Actually nevermind, I forgot that these will not end up in the bucket logging logs.

dvasilas (approved):

leif-scality:
ping

leif-scality:
ping

leif-scality:
ping

PR #6051, opened 2026-01-19
CLDSRV-827: Fix contentLength and objectSize in access logs 

dvasilas:
The `contentLength` field in S3 analytics captures egress (response size) for `GetObject` operations and ingress (request size) for `PutObject`.

In server access logs, `contentLength` was the HTTP request body size (bytes received from client).
This broke backwards compatibility with S3 analytics, and caused an Integration test to fail.

---

The fix is to update `contentLength` to capture transfer size (egress for GET, ingress for PUT).

This PR includes a second fix for `objectSize` to always reflect full object size (and not  sizes in range get operations).

---

Intregration bulid with this fix: https://github.com/scality/Integration/actions/runs/21141099093

### DIFF
--- a/lib/api/objectGet.js
+++ b/lib/api/objectGet.js
@@ -139,6 +139,11 @@ function objectGet(authInfo, request, returnTagCount, log, callback) {
 
         const objLength = (objMD.location === null ?
                            0 : parseInt(objMD['content-length'], 10));
+        // Store full object size for server access logs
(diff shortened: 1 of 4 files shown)


nicolas2bert (approved):

anurag4DSB (approved):

dvasilas:
:robot:

PR #6055, opened 2026-01-20
CLDSRV-836 Fix deep healthcheck fail status

BourgoisMickael:
Now fail if ALL backends of only one client fails.

Previously, the deep healthcheck would fail if ALL backends/locations
failed globally across all clients (data, metadata, vault, kms).

This change modifies the logic to fail if ANY client has ALL its
backends/locations failing. This ensures:

1. For data backend with multiple sproxyd location constraints:
   - Returns HTTP 200 if at least ONE location is healthy
   - Returns HTTP 500 only if ALL locations fail

2. Each client (data, metadata, vault, kms) is evaluated independently
   - If ALL locations of the data client fail, overall check fails
   - If ALL locations of metadata fail, overall check fails
   - etc.

The new logic uses:
- `results.some()` to check across clients
- `keys.every()` within each client to check all its locations

### DIFF
--- a/lib/utilities/healthcheckHandler.js
+++ b/lib/utilities/healthcheckHandler.js
@@ -56,20 +56,30 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
         }
     });
     async.parallel(clientTasks, (err, results) => {
-        let fail = false;
         // obj will be an object of the healthcheck results of
         // every backends. No errors were returned directly to
         // async.parallel in order to complete the check, so a
         // manual check makes S3 return 500 error if any backend failed
         // other than aws_s3 or azure
-        const obj = results.reduce((obj, item) => Object.assign(obj, item), {});
-        // fail only if *all* backend fail, so that we can still serve the ones which are working
-        fail = Object.keys(obj).every(k =>
-            // if there is an error from an external backend,
-            // only return a 500 if it is on startup
-            // (flightCheckOnStartUp set to true)
-            obj[k].error && (flightCheckOnStartUp || !obj[k].external)
-        );
+        // Check if any client has ALL its backends/locations failing.
+        // Each result in the array represents one client (data, metadata, vault, kms)
+        // with its backend locations as keys.
+        // If ALL backend/locations of ANY client have errors, the overall check fails.
+        const { obj, fail } = results.reduce((acc, clientResult) => {
+            Object.assign(acc.obj, clientResult);
+
+            // Check if ALL backends/locations of this client have errors
+            const keys = Object.keys(clientResult);
+            // eslint-disable-next-line no-param-reassign
+            acc.fail ||= keys.length > 0 && keys.every(k =>
+                // if there is an error from an external backend,
+                // only return a 500 if it is on startup
+                // (flightCheckOnStartUp set to true)
+                clientResult[k].error && (flightCheckOnStartUp || !clientResult[k].external)
+            );
+
+            return acc;
+        }, { obj: {}, fail: false });
         if (fail) {
             return cb(errors.InternalError, obj);
         }
--- a/tests/unit/healthchecks/clientCheck.js
+++ b/tests/unit/healthchecks/clientCheck.js
@@ -1,0 +1,235 @@
+const assert = require('assert');
+const sinon = require('sinon');
+const { errors } = require('arsenal');
+
+const { clientCheck } = require('../../../lib/utilities/healthcheckHandler');
+const { DummyRequestLogger } = require('../helpers');
+const { data } = require('../../../lib/data/wrapper');
+const metadata = require('../../../lib/metadata/wrapper');
+const vault = require('../../../lib/auth/vault');
+const kms = require('../../../lib/kms/wrapper');
+
+const log = new DummyRequestLogger();
+
+describe('clientCheck - failure detection logic', () => {
+    let dataStub;
+    let metadataStub;
+    let vaultStub;
+    let kmsStub;
+
+    beforeEach(() => {
+        // Create stubs for all client checkHealth methods
+        dataStub = sinon.stub(data, 'checkHealth');
+        metadataStub = sinon.stub(metadata, 'checkHealth');
+        vaultStub = sinon.stub(vault, 'checkHealth');
+        kmsStub = sinon.stub(kms, 'checkHealth');
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    it('should succeed when all backends are healthy', done => {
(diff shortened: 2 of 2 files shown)


leif-scality (approved):

tcarmet (approved):

jonathan-gramain (approved):

anurag4DSB (approved):
This behaviour needs to be documented in official documentation so customers and CS know what to expect + release notes are a must for this change in behaviour

lib/utilities/healthcheckHandler.js
```diff
@@ -56,20 +56,30 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
...
+                return false;
+            }
+            // Check if ALL backends/locations of this client have errors
+            return keys.every(k =>
+                // if there is an error from an external backend,
+                // only return a 500 if it is on startup
+                // (flightCheckOnStartUp set to true)
+                clientResult[k].error && (flightCheckOnStartUp || !clientResult[k].external)
+            );
+        });
```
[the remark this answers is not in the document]
francoisferrand: should have extra tests indeed

lib/utilities/healthcheckHandler.js
```diff
@@ -56,20 +56,30 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
         }
     });
     async.parallel(clientTasks, (err, results) => {
-        let fail = false;
         // obj will be an object of the healthcheck results of
         // every backends. No errors were returned directly to
         // async.parallel in order to complete the check, so a
         // manual check makes S3 return 500 error if any backend failed
         // other than aws_s3 or azure
         const obj = results.reduce((obj, item) => Object.assign(obj, item), {});
```
francoisferrand: should this be merged with the next traversal?
the 2 iterations made here make the code kind of hard to read IMHO, as we build `obj` and then do not really use it (until the `return` call, that is)

Maybe something like this would be more readable:

```suggestion
        const obj = {};
        const fail = results.reduce((result, item) => {
            Object.assign(obj, item);

            keys = Object.keys(item);
            return result || (keys.length > 0 && keys.every(k =>
                item[k].error && (flightCheckOnStartUp || !item[k].external)
            ));
        }, false)
```

lib/utilities/healthcheckHandler.js
```diff
@@ -56,20 +56,30 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
...
-            obj[k].error && (flightCheckOnStartUp || !obj[k].external)
-        );
+        // Check if any client has ALL its backends/locations failing.
+        // Each result in the array represents one client (data, metadata, vault, kms)
+        // with its backend locations as keys.
+        // If ALL backend/locations of ANY client have errors, the overall check fails.
+        const fail = results.some(clientResult => {
+            const keys = Object.keys(clientResult);
+            // If this client has no keys, skip it
+            if (keys.length === 0) {
```
francoisferrand: is this a real case, i.e. we have some "clients" which return an empty result ?

francoisferrand:
> This behaviour needs to be documented in official documentation so customers and CS know what to expect + release notes are a must for this change in behavior

- Does this actually change behavior in S3C? When discussing, our understanding was that this actually does not change the behavior (since each "client" returns a single result) : so we would just go back to the existing behavior of S3C.
- The case with multiple result is multi-location backend in Zenko, which will indeed return one result per location. In that case, the behavior did not change either with respect to that client.
- In the end, we expect the only change of behavior (at product level) is that in Artesca, cloudserver may now fail healthcheck when either _every_ vault or _every_ mongodb is dead  (since k8s handles the routing), instead of when when every vault and every mongo is done (as today). So the change is really just a corner case, which nobody will ever notice in Artesca IMHO.

lib/utilities/healthcheckHandler.js
BourgoisMickael: There [421f6c7](https://github.com/scality/cloudserver/pull/6055/commits/421f6c747d554cc6c2e7c70aa3ccc6f468fd411f)

lib/utilities/healthcheckHandler.js
BourgoisMickael: no there should not be

lib/utilities/healthcheckHandler.js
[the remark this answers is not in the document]
francoisferrand: @BourgoisMickael not sure you seen this one?

we changed the behavior, but no test: so something is missing (probably already was), and we should add the missing test(s)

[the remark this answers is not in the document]
BourgoisMickael: There are some tests: [3f5af34](https://github.com/scality/cloudserver/pull/6055/commits/3f5af34d8ca16181b7d5cd23cee0a859da67ed1f)

lib/utilities/healthcheckHandler.js
```diff
@@ -61,25 +61,25 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
...
-                return false;
-            }
+        const { obj, fail } = results.reduce((acc, clientResult) => {
+            Object.assign(acc.obj, clientResult);
+
             // Check if ALL backends/locations of this client have errors
-            return keys.every(k =>
+            const keys = Object.keys(clientResult);
+            // eslint-disable-next-line no-param-reassign
+            acc.fail = acc.fail || (keys.length > 0 && keys.every(k =>
```
francoisferrand: nit:
```suggestion
            acc.fail ||= keys.length > 0 && keys.every(k =>
                // if there is an error from an external backend,
                // only return a 500 if it is on startup
                // (flightCheckOnStartUp set to true)
                clientResult[k].error && (flightCheckOnStartUp || !clientResult[k].external)
            );
```

tests/unit/healthchecks/clientCheck.js
```diff
@@ -0,0 +1,235 @@
...
+            assert.ifError(err);
+            assert.deepStrictEqual(result, {
+                metadata: { code: 200, message: 'OK' },
+                vault: { code: 200, message: 'OK' },
+                kms: { code: 200, message: 'OK' },
+            });
+            done();
+        });
+    });
+
```
francoisferrand, line 168: for completeness, should we have a test as well when metadata or vault fails?

francoisferrand (approved):

tests/unit/healthchecks/clientCheck.js
BourgoisMickael: i don't think so, all client are handled generically, there is no more cases to test for specific clients

BourgoisMickael:
ping

PR #6060, opened 2026-01-26
Support the new API GetObjectAttributes

maeldonn:
Issue: CLDSRV-817

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -279,6 +279,14 @@ const constants = {
     rateLimitDefaultConfigCacheTTL: 30000, // 30 seconds
     rateLimitDefaultBurstCapacity: 1,
     rateLimitCleanupInterval: 10000, // 10 seconds
+    // Supported attributes for the GetObjectAttributes 'x-amz-optional-attributes' header.
+    supportedGetObjectAttributes: new Set([
+        'StorageClass',
+        'ObjectSize',
+        'ObjectParts',
+        'Checksum',
+        'ETag',
+    ]),
 };
 
 module.exports = constants;
--- a/lib/api/apiUtils/object/parseAttributesHeader.js
+++ b/lib/api/apiUtils/object/parseAttributesHeader.js
@@ -1,0 +1,25 @@
+const { errorInstances } = require('arsenal');
+const { supportedGetObjectAttributes } = require('../../../../constants');
+
+/**
+ * parseAttributesHeaders - Parse and validate the x-amz-object-attributes header
+ * @param {object} headers - request headers
+ * @returns {Set<string>} - set of requested attribute names
+ * @throws {Error} - InvalidRequest if header is missing/empty, InvalidArgument if attribute is invalid
+ */
+function parseAttributesHeaders(headers) {
+  const attributes = headers['x-amz-object-attributes']?.split(',').map(attr => attr.trim()) ?? [];
+  if (attributes.length === 0) {
+    throw errorInstances.InvalidRequest.customizeDescription(
+      'The x-amz-object-attributes header specifying the attributes to be retrieved is either missing or empty',
+    );
+  }
+
+  if (attributes.some(attr => !supportedGetObjectAttributes.has(attr))) {
+    throw errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified.');
+  }
+
+  return new Set(attributes);
+}
+
+module.exports = parseAttributesHeaders;
--- a/lib/api/objectGetAttributes.js
+++ b/lib/api/objectGetAttributes.js
@@ -1,0 +1,170 @@
+const { promisify } = require('util');
+const { errors } = require('arsenal');
+const { standardMetadataValidateBucketAndObj } = require('../metadata/metadataUtils');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
+const parseAttributesHeaders = require('./apiUtils/object/parseAttributesHeader');
+const { decodeVersionId, getVersionIdResHeader } = require('./apiUtils/object/versioning');
+const { checkExpectedBucketOwner } = require('./apiUtils/authorization/bucketOwner');
+const { pushMetric } = require('../utapi/utilities');
+const { getPartCountFromMd5 } = require('./apiUtils/object/partInfo');
+
+const checkExpectedBucketOwnerPromise = promisify(checkExpectedBucketOwner);
+const validateBucketAndObj = promisify(standardMetadataValidateBucketAndObj);
+
+const OBJECT_GET_ATTRIBUTES = 'objectGetAttributes';
+
+/**
+ * buildXmlResponse - Build XML response for GetObjectAttributes
+ * @param {object} objMD - object metadata
+ * @param {Set<string>} requestedAttrs - set of requested attribute names
+ * @returns {string} XML response
+ */
+function buildXmlResponse(objMD, requestedAttrs) {
+  const xml = [];
+  xml.push(
+    '<?xml version="1.0" encoding="UTF-8"?>',
+    '<GetObjectAttributesResponse>',
+  );
+
+  for (const attribute of requestedAttrs) {
+    switch (attribute) {
+      case 'ETag':
+        xml.push(`<ETag>${objMD['content-md5']}</ETag>`);
+        break;
+      case 'ObjectParts': {
+        const partCount = getPartCountFromMd5(objMD);
+        if (partCount) {
+          xml.push(
+            '<ObjectParts>',
+            `<PartsCount>${partCount}</PartsCount>`,
+            '</ObjectParts>',
+          );
+        }
+        break;
+      }
+      case 'StorageClass':
+        xml.push(`<StorageClass>${objMD['x-amz-storage-class']}</StorageClass>`);
+        break;
+      case 'ObjectSize':
+        xml.push(`<ObjectSize>${objMD['content-length']}</ObjectSize>`);
+        break;
+    }
+  }
+
+  xml.push('</GetObjectAttributesResponse>');
+  return xml.join('');
+}
+
+/**
+ * objectGetAttributes - Retrieves all metadata from an object without returning the object itself
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback optional to keep backward compatibility
+ * @returns {Promise<object>} - { xml, responseHeaders }
+ * @throws {ArsenalError} NoSuchVersion - if versionId specified but not found
+ * @throws {ArsenalError} NoSuchKey - if object not found
+ * @throws {ArsenalError} MethodNotAllowed - if object is a delete marker
+ */
+async function objectGetAttributes(authInfo, request, log, callback) {
+  if (callback) {
+    return objectGetAttributes(authInfo, request, log)
+      .then(result => callback(null, result.xml, result.responseHeaders))
+      .catch(err => callback(err, null, err.responseHeaders ?? {}));
+  }
+
+  log.trace('processing request', { method: OBJECT_GET_ATTRIBUTES });
+  const { bucketName, objectKey, headers, actionImplicitDenies } = request;
+
+  const versionId = decodeVersionId(request.query);
+  if (versionId instanceof Error) {
+    log.debug('invalid versionId query', {
+      method: OBJECT_GET_ATTRIBUTES,
+      versionId: request.query.versionId,
+      error: versionId,
+    });
+    throw versionId;
+  }
+
+  const metadataValParams = {
+    authInfo,
+    bucketName,
+    objectKey,
+    versionId,
+    getDeleteMarker: true,
+    requestType: request.apiMethods || OBJECT_GET_ATTRIBUTES,
+    request,
+  };
+
+  let bucket, objectMD;
+  try {
+    ({ bucket, objectMD } = await validateBucketAndObj(metadataValParams, actionImplicitDenies, log));
+    await checkExpectedBucketOwnerPromise(headers, bucket, log);
+  } catch (err) {
+    log.debug('error validating bucket and object', {
+      method: OBJECT_GET_ATTRIBUTES,
+      bucket: bucketName,
+      key: objectKey,
+      versionId,
+      error: err,
+    });
+    throw err;
+  }
+
+  const responseHeaders = collectCorsHeaders(headers.origin, request.method, bucket);
+
+  if (!objectMD) {
+    log.debug('object not found', {
+      method: OBJECT_GET_ATTRIBUTES,
+      bucket: bucketName,
+      key: objectKey,
+      versionId,
+    });
+    const err = versionId ? errors.NoSuchVersion : errors.NoSuchKey;
+    err.responseHeaders = responseHeaders;
+    throw err;
+  }
+
+  responseHeaders['x-amz-version-id'] = getVersionIdResHeader(bucket.getVersioningConfiguration(), objectMD);
+  responseHeaders['Last-Modified'] = objectMD['last-modified'] && new Date(objectMD['last-modified']).toUTCString();
+
+  if (objectMD.isDeleteMarker) {
+    log.debug('attempt to get attributes of a delete marker', {
+      method: OBJECT_GET_ATTRIBUTES,
+      bucket: bucketName,
+      key: objectKey,
+      versionId,
+    });
+    responseHeaders['x-amz-delete-marker'] = true;
+    const err = errors.MethodNotAllowed;
+    err.responseHeaders = responseHeaders;
+    throw err;
+  }
+
+  const requestedAttrs = parseAttributesHeaders(headers);
+
+  if (requestedAttrs.has('Checksum')) {
+    log.debug('Checksum attribute requested but not implemented', {
+      method: OBJECT_GET_ATTRIBUTES,
+      bucket: bucketName,
+      key: objectKey,
+      versionId,
+    });
+    const err = errors.NotImplemented.customizeDescription('Checksum attribute is not implemented');
+    err.responseHeaders = responseHeaders;
+    throw err;
+  }
+
+  pushMetric(OBJECT_GET_ATTRIBUTES, log, {
+    authInfo,
+    bucket: bucketName,
+    keys: [objectKey],
+    versionId: objectMD?.versionId,
+    location: objectMD?.dataStoreName,
+  });
+
+  const xml = buildXmlResponse(objectMD, requestedAttrs);
+  return { xml, responseHeaders };
+}
+
+module.exports = objectGetAttributes;
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
     "@smithy/node-http-handler": "^3.0.0",
-    "arsenal": "git+https://github.com/scality/arsenal#8.3.3",
+    "arsenal": "git+https://github.com/scality/Arsenal#8.3.4",
     "async": "2.6.4",
     "bucketclient": "scality/bucketclient#8.2.7",
     "bufferutil": "^4.0.8",
--- a/tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
+++ b/tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
@@ -1,0 +1,254 @@
+const assert = require('assert');
+const {
+    CreateBucketCommand,
+    DeleteBucketCommand,
+    DeleteObjectCommand,
+    PutObjectCommand,
+    GetObjectAttributesCommand,
+    CreateMultipartUploadCommand,
+    UploadPartCommand,
+    CompleteMultipartUploadCommand,
+} = require('@aws-sdk/client-s3');
+
+const withV4 = require('../support/withV4');
+const BucketUtility = require('../../lib/utility/bucket-util');
+
+const bucket = 'testbucket';
+const key = 'testobject';
+const body = 'hello world!';
+const expectedMD5 = 'fc3ff98e8c6a0d3087d515c0473f8677';
+
+describe('objectGetAttributes', () => {
+    withV4(sigCfg => {
+        let bucketUtil;
+        let s3;
+
+        before(() => {
+            bucketUtil = new BucketUtility('default', sigCfg);
+            s3 = bucketUtil.s3;
+        });
+
+        beforeEach(async () => {
+            await s3.send(new CreateBucketCommand({ Bucket: bucket }));
+            await s3.send(new PutObjectCommand({ Bucket: bucket, Key: key, Body: body }));
+        });
+
+        afterEach(async () => {
+            await s3.send(new DeleteObjectCommand({ Bucket: bucket, Key: key }));
+            await s3.send(new DeleteBucketCommand({ Bucket: bucket }));
+        });
+
+        it('should fail with a wrong bucket owner header', async () => {
+            try {
+                await s3.send(new GetObjectAttributesCommand({
+                    Bucket: bucket,
+                    Key: key,
+                    ObjectAttributes: ['ETag'],
+                    ExpectedBucketOwner: 'wrongAccountId',
+                }));
+                assert.fail('Expected AccessDenied error');
+            } catch (err) {
+                assert.strictEqual(err.name, 'AccessDenied');
+                assert.strictEqual(err.message, 'Access Denied');
+            }
+        });
+
+        it('should fail because attributes header is missing', async () => {
+            try {
+                await s3.send(new GetObjectAttributesCommand({
+                    Bucket: bucket,
+                    Key: key,
+                    ObjectAttributes: [],
+                }));
+                assert.fail('Expected InvalidArgument error');
+            } catch (err) {
+                assert.strictEqual(err.name, 'InvalidArgument');
+                assert.strictEqual(err.message, 'Invalid attribute name specified.');
+            }
+        });
+
+        it('should fail because attribute name is invalid', async () => {
+            try {
+                await s3.send(new GetObjectAttributesCommand({
+                    Bucket: bucket,
+                    Key: key,
+                    ObjectAttributes: ['InvalidAttribute'],
+                }));
+                assert.fail('Expected InvalidArgument error');
+            } catch (err) {
+                assert.strictEqual(err.name, 'InvalidArgument');
+                assert.strictEqual(err.message, 'Invalid attribute name specified.');
+            }
+        });
+
+        it('should return NoSuchKey for non-existent object', async () => {
+            try {
+                await s3.send(new GetObjectAttributesCommand({
+                    Bucket: bucket,
+                    Key: 'nonexistent',
+                    ObjectAttributes: ['ETag'],
+                }));
+                assert.fail('Expected NoSuchKey error');
+            } catch (err) {
+                assert.strictEqual(err.name, 'NoSuchKey');
+                assert.strictEqual(err.message, 'The specified key does not exist.');
+            }
+        });
+
+        it('should return all attributes', async () => {
+            const data = await s3.send(new GetObjectAttributesCommand({
+                Bucket: bucket,
+                Key: key,
+                ObjectAttributes: ['ETag', 'ObjectParts', 'StorageClass', 'ObjectSize'],
+            }));
+
+            assert.strictEqual(data.ETag, expectedMD5);
+            assert.strictEqual(data.StorageClass, 'STANDARD');
+            assert.strictEqual(data.ObjectSize, body.length);
+            assert.strictEqual(data.ObjectParts, undefined, "ObjectParts shouldn't be present for non-MPU object");
+            assert(data.LastModified, 'LastModified should be present');
+        });
+
+        it('should return ETag', async () => {
+            const data = await s3.send(new GetObjectAttributesCommand({
+                Bucket: bucket,
+                Key: key,
+                ObjectAttributes: ['ETag'],
+            }));
+
+            assert.strictEqual(data.ETag, expectedMD5);
+        });
+
+        it('should fail with NotImplemented when Checksum is requested', async () => {
+            try {
+                await s3.send(new GetObjectAttributesCommand({
+                    Bucket: bucket,
+                    Key: key,
+                    ObjectAttributes: ['Checksum'],
+                }));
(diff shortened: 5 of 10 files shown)


constants.js
```diff
@@ -279,6 +279,14 @@ const constants = {
     rateLimitDefaultConfigCacheTTL: 30000, // 30 seconds
     rateLimitDefaultBurstCapacity: 1,
     rateLimitCleanupInterval: 10000, // 10 seconds
+    // Metadata allowed to be returned by getObjectAttributes API
+    allowedObjectAttributes: new Set([
```
maeldonn: Should we handle our custom metadata ?

lib/api/apiUtils/object/parseAttributesHeader.js
```diff
@@ -0,0 +1,26 @@
+const { errorInstances } = require('arsenal');
+const { allowedObjectAttributes } = require('../../../../constants');
+
+function parseAttributesHeaders(headers) {
```
francoisferrand, line 10: there is already a very similar function in @DarkIsDude PR, which efficiently handles validation & parsing: should we not "merge" both code (i.e. typically make your code extend his) ?

tests/unit/api/objectGetAttributes.js
francoisferrand, line 1: it's good (and necessary) to have unit test, but we should also have functional test to verify we can indeed use the API as expected through the AWS sdk.

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
...
+      },
+      next => {
+        if (!objMD) {
+          const e = versionId ? errors.NoSuchVersion : errors.NoSuchKey;
+          return next(e);
+        }
+
+        if (objMD.isDeleteMarker) {
+          responseHeaders['x-amz-delete-marker'] = true;
+          return next(errors.MethodNotAllowed);
```
francoisferrand: is this really the behavior of AWS S3 ?
A deleteMarker is a version as well, seems weird we would not be able to get information about it (though obviously there is much less data to retrieve)

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
+const { waterfall } = require('async');
```
SylvainSenechal: Hey I haven't read the pr at all, but just seing this rapidly : I think you can try use more modern javascript syntax if possible, maybe you did it to stay aligned with other apis file style, but I think it can be more modern

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
...
+ */
+function buildXmlResponse(objMD, attributes) {
+  const attrResp = {};
+
+  if (attributes.includes('ETag')) {
+    attrResp.ETag = objMD['content-md5'];
+  }
+
+  if (attributes.includes('Checksum')) {
+    attrResp.Checksum = {};
```
maeldonn: Does not seem to be handled...

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,348 @@
+const assert = require('assert');
+const AWS = require('aws-sdk');
+const S3 = AWS.S3;
```
francoisferrand:
```suggestion
const { s3 } = require('aws-sdk');
```

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+const xml2js = require('xml2js');
+const getConfig = require('../support/config');
+const { removeAllVersions, versioningEnabled } = require('../../lib/utility/versioning-util.js');
+
+const bucket = 'testbucket';
+const key = 'testobject';
+const body = 'hello world!';
+
+const sendRequest = async (method, host, path, headers = {}) => {
```
francoisferrand:
```suggestion
async function sendRequest(method, host, path, headers = {}) {
```

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+  const endpoint = new AWS.Endpoint(host);
+  const request = new AWS.HttpRequest(endpoint);
+  request.method = method.toUpperCase();
+  request.path = path;
+  request.headers.Host = host;
+  request.headers['X-Amz-Date'] = new Date().toISOString().replace(/[:\-]|\.\d{3}/g, '');
+  request.headers['X-Amz-Content-SHA256'] = AWS.util.crypto.sha256('', 'hex');
+  request.region = 'us-east-1';
+
```
francoisferrand: this is not the way to use AWS SDK. It should look like something like the following, and thus no need for a wrapper...

```
const client = new s3.Client({
    endpoint: "...",
    [...]
});
client.send(new GetObjectAttributes({
    [...]
});
```

francoisferrand (changes requested):

lib/api/objectGetAttributes.js
maeldonn: Yes, this is normal behavior...

constants.js
francoisferrand: yes, with the same additional permission @DarkIsDude added for ListObjectv2
→ should be done in a separate PR I guess

DarkIsDude: @maeldonn we need to handle indeed user metadata (and RestoreStatus?). Does it make sense for you @maeldonn ?

lib/api/apiUtils/object/parseAttributesHeader.js
```diff
@@ -0,0 +1,32 @@
...
+  const raw = headers['x-amz-object-attributes'] || '';
+
+  const attributes = raw
+    .split(',')
+    .map(s => s.trim())
+    .filter(s => s !== '');
+
+  if (attributes.length === 0) {
+    throw errorInstances.InvalidRequest.customizeDescription(
+      'The x-amz-object-attributes header specifying the attributes to be retrieved is either missing or empty',
```
DarkIsDude, line 14: Does this description is the one that return AWS ? One of our goal is to be AWS compliant, even on them

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
...
+ * validateBucketAndObjPromise - Promisified wrapper for standardMetadataValidateBucketAndObj
+ * @param {object} params - validation parameters
+ * @param {boolean} actionImplicitDenies - whether action has implicit denies
+ * @param {object} log - Werelogs logger
+ * @returns {Promise<{bucket: BucketInfo, objMD: object}>} - bucket and object metadata
+ * @throws {Error} - rejects with error from standardMetadataValidateBucketAndObj
+ */
+function validateBucketAndObjPromise(params, actionImplicitDenies, log) {
+  return new Promise((resolve, reject) => {
+    standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log, (err, bucket, objMD) => {
```
DarkIsDude: Can we use promisify ? I'm not sure as the callback receive several value but good point to check ?

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * objectGetAttributes - Retrieves all metadata from an object without returning the object itself
+ * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
+ * @param {object} request - http request object
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback to server
+ * @return {undefined}
+ */
+async function objectGetAttributes(authInfo, request, log, callback) {
```
DarkIsDude, line 69: We should use callbackify instead of mixing async/callback 🙏

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
...
+  log.trace('processing request', { method: OBJECT_GET_ATTRIBUTES });
+  const { bucketName, objectKey, headers, actionImplicitDenies } = request;
+
+  let responseHeaders = {};
+
+  const versionId = decodeVersionId(request.query);
+  if (versionId instanceof Error) {
+    log.debug('invalid versionId query', { versionId: request.query.versionId, error: versionId });
+    throw versionId;
```
DarkIsDude, line 86: ⏫ for example here you throw but without calling the callback, you'll have a weird behaviour. Make your function fully async and then export a callbackify for example

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
...
+    const { bucket, objMD } = await validateBucketAndObjPromise(metadataValParams, actionImplicitDenies, log);
+    await checkExpectedBucketOwnerPromise(headers, bucket, log);
+
+    responseHeaders = collectCorsHeaders(headers.origin, request.method, bucket);
+    if (objMD) {
+      responseHeaders['x-amz-version-id'] = getVersionIdResHeader(bucket.getVersioningConfiguration(), objMD);
+      responseHeaders['Last-Modified'] = objMD['last-modified'] && new Date(objMD['last-modified']).toUTCString();
+    }
+
+    if (!objMD) {
```
DarkIsDude: I would move this one before the previous if ?

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,146 @@
...
+      authInfo,
+      bucket: bucketName,
+      keys: [objectKey],
+      versionId: objMD?.versionId,
+      location: objMD?.dataStoreName,
+    });
+
+    const xml = buildXmlResponse(objMD, attributes);
+    return callback(null, xml, responseHeaders);
+  } catch (err) {
```
DarkIsDude: By making your function full async, you'll avoid this try/catch level in your function as it's handle with callbackify

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,272 @@
+const assert = require('assert');
+const { S3 } = require('aws-sdk');
+const getConfig = require('../support/config');
+
+const bucket = 'testbucket';
+const key = 'testobject';
+const body = 'hello world!';
+const expectedMD5 = 'fc3ff98e8c6a0d3087d515c0473f8677';
+
+describe('Test get object attributes', () => {
```
DarkIsDude:
```suggestion
describe('objectGetAttributes', () => {
```

We know that it's a test as we are in test and we usually put the file name or the endpoint, you can check other tests

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,272 @@
...
+    await s3.createBucket({ Bucket: bucket }).promise();
+    await s3.putObject({ Bucket: bucket, Key: key, Body: body }).promise();
+  });
+
+  afterEach(async () => {
+    await s3.deleteObject({ Bucket: bucket, Key: key }).promise();
+    await s3.deleteBucket({ Bucket: bucket }).promise();
+  });
+
+  it('should fail because a bad bucket owner', async () => {
```
DarkIsDude: That's not really english right ? `should fail with a wrong bucket owner param` for example ?

DarkIsDude: I'll let you reword your following `because` ?

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,272 @@
...
+        Key: key,
+        ObjectAttributes: ['ETag', 'Checksum', 'ObjectParts', 'StorageClass', 'ObjectSize'],
+      })
+      .promise();
+
+    assert.strictEqual(data.ETag, expectedMD5);
+    assert.strictEqual(data.StorageClass, 'STANDARD');
+    assert.strictEqual(data.ObjectSize, body.length);
+    assert(data.Checksum, 'Checksum should be present');
+    assert(!data.ObjectParts, "ObjectParts shouldn't be present for non-MPU object");
```
DarkIsDude: I guess we'll prefer to compare it to undefined ? Because if `data.ObjectParts` is false, then your test pass ?

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,272 @@
...
+    assert(data.LastModified, 'LastModified should be present');
```
DarkIsDude: You can compare the value to a string for example ?

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,272 @@
...
+  it("shouldn't return ObjectParts", async () => {
+    const data = await s3
+      .getObjectAttributes({
+        Bucket: bucket,
+        Key: key,
+        ObjectAttributes: ['ObjectParts'],
+      })
+      .promise();
+
+    // ObjectParts may be empty for non-MPU objects
```
DarkIsDude: Instead of a comment you can update your `it` => `should not return ObjectParts for non-MPU objects`

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -0,0 +1,272 @@
...
+  let s3;
+  const mpuKey = 'mpuObject';
+  const partSize = 5 * 1024 * 1024; // Minimum part size is 5MB
+  const partCount = 3;
+
+  before(async () => {
+    const config = getConfig('default', { signatureVersion: 'v4' });
+    s3 = new S3(config);
+
+    // Create bucket
```
DarkIsDude:
```suggestion
```

You just describe your code (same for following comments)

tests/functional/aws-node-sdk/test/versioning/objectGetAttributes.js
```diff
@@ -0,0 +1,145 @@
...
+  it('should return MethodNotAllowed for delete marker', async () => {
+    await s3
+      .putObject({
+        Bucket: bucket,
+        Key: key,
+        Body: body,
+      })
+      .promise();
+
+    // Delete creates a delete marker
```
DarkIsDude: Same you just describe your code here

tests/unit/api/objectGetAttributes.js
DarkIsDude, line 1: I guess you can move your code to async/await after reworking the lib/api to full async/await ?

tests/unit/api/objectGetAttributes.js
```diff
@@ -0,0 +1,672 @@
...
+      bucketName,
+      namespace,
+      objectKey: objectName,
+      headers: {},
+      url: `/${bucketName}/${objectName}`,
+      query: {},
+      actionImplicitDenies: false,
+    };
+
+    bucketPut(authInfo, testPutBucketRequest, log, () => {
```
DarkIsDude: Should we move that in a beforeEach ? As you always do that ?

DarkIsDude (changes requested):

lib/api/apiUtils/object/parseAttributesHeader.js
maeldonn: Yes, it is

lib/api/objectGetAttributes.js
maeldonn: Promisify expect a standard callback (callback(error, singleParam)), that's not the case here...
Since standardMetadataValidateBucketAndObj is used everywhere, a clean way to promisify this function would be to override the util.promisify.custom symbol.

https://nodejs.org/docs/latest-v22.x/api/util.html#custom-promisified-functions

lib/api/objectGetAttributes.js
maeldonn: It's not intentional, it should have been in the try block...

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
maeldonn: I think it's enough here. I have another test to check that Last Modified is a valid date.

tests/unit/api/objectGetAttributes.js
maeldonn: After async/await migration it will be easier to do that 👷‍♂️

constants.js
maeldonn: As we said, we will not process RestoreStatus but we will add user metadata in another ticket.

lib/api/apiUtils/object/parseAttributesHeader.js
maeldonn: Will be updated in another PR (process user metadata in GetObjectAttributes).

francoisferrand: why? it means we are duplicating the code now instead of just rebasing this PR on top of the other and handling it just once...

maeldonn: Because user metadata is already handled in the listObjectV2 API. That's not the case here. That would take me to an intermediate step where I should add unnecessary complexity....

Edit: Done [here](https://github.com/scality/cloudserver/pull/6070/changes/260e8b3ba37ab9d29ae4f9d3594e7a0cf818a8ca)

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,154 @@
...
+    throw versionId;
+  }
+
+  const metadataValParams = {
+    authInfo,
+    bucketName,
+    objectKey,
+    versionId,
+    getDeleteMarker: true,
+    requestType: request.apiMethods || OBJECT_GET_ATTRIBUTES,
```
SylvainSenechal, line 95: nit: Is the || needed ? I feel like request.apiMethods would always be the same (and equal to OBJECT_GET_ATTRIBUTES) ? Curious to know in what situation request.apiMethods is undefined

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,154 @@
...
+
+  if (objMD.isDeleteMarker) {
+    log.debug('attempt to get attributes of a delete marker', {
+      method: OBJECT_GET_ATTRIBUTES,
+      bucket: bucketName,
+      key: objectKey,
+      versionId,
+    });
+    responseHeaders['x-amz-delete-marker'] = true;
+    const err = errors.MethodNotAllowed;
```
SylvainSenechal, line 139: @maeldonn @francoisferrand  
I think per AWS standards : 
if versionId was not provided but it's a deleteMarker, we should return 404
```suggestion
    const err = versionId ? errors.MethodNotAllowed : errors.NoSuchKey;
```

SylvainSenechal:
Tested locally, looked good

lib/api/objectGetAttributes.js
maeldonn: As we discussed, this is already handled on line 119.

constants.js
```diff
@@ -275,6 +275,14 @@ const constants = {
         'bucketPutLogging',
         'bucketGetLogging',
     ],
+    // Metadata allowed to be returned by getObjectAttributes API
+    allowedObjectAttributes: new Set([
```
DarkIsDude:
```suggestion
    // Accepted (or allowed) values that can be required in the GetObjectAttributes API through the x-amz-optional-attributes header
    allowedValuesForXAmzOptionalAttributesInGetObjectAttributes: new Set([
```

package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@azure/storage-blob": "^12.28.0",
     "@hapi/joi": "^17.1.1",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.2.43",
+    "arsenal": "git+https://github.com/scality/Arsenal#feature/ARSN-549/get-object-attributes",
```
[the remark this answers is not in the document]
DarkIsDude: I guess you can bump ?

tests/unit/api/objectGetAttributes.js
```diff
@@ -0,0 +1,488 @@
...
+    fn(...args, (err, ...results) => {
+        if (err) {
+            reject(err);
+        } else {
+            resolve(results);
+        }
+    });
+});
+
+const bucketPutAsync = promisify(bucketPut);
```
DarkIsDude, line 36: Should we use the custom trick that you did then ?

DarkIsDude: My idea is too move your wonderful function in an util and then use the promisify.custom

package.json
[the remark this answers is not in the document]
maeldonn: I'm waiting for all the reviews first. I will do it just before merging the PR. That's why I haven't resolved this comment 😉

tests/unit/api/objectGetAttributes.js
maeldonn: I created this utility function as I use these functions only in my tests. If it were in an API file, I would have used promisify.custom. I don't think it's worth it here because it would just unnecessarily increase the number of files to review and the PR size.

Concerning this utility function, I can move it to an utility file, but it must remain intended for testing. The problem is that I return an array instead of an object as in the standard, we therefore lose the type intelligence on the return.

constants.js
maeldonn: Description is updated to be more descriptive.

DarkIsDude (commented):
LGTM. Congrats

DarkIsDude (approved):

SylvainSenechal (approved):
approved, gonna check the next pr

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,154 @@
...
+ * @returns {string} XML response
+ */
+function buildXmlResponse(objMD, attributes) {
+  const attrResp = {};
+
+  if (attributes.includes('ETag')) {
+    attrResp.ETag = objMD['content-md5'];
+  }
+
+  // NOTE: Checksum is not implemented
```
francoisferrand: if we don't implement it (which is fine), then we should reject it in the request.

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,154 @@
...
+/**
+ * buildXmlResponse - Build XML response for GetObjectAttributes
+ * @param {object} objMD - object metadata
+ * @param {array} attributes - requested attributes
+ * @returns {string} XML response
+ */
+function buildXmlResponse(objMD, attributes) {
+  const attrResp = {};
+
+  if (attributes.includes('ETag')) {
```
francoisferrand: we make lots of requests on this array : would be much more efficient to use a map or just build a plain object (since anyway we don't have a generic logic to "map" requested attributes to response fields)

i.e. either we should make it generic (by instroducing a const map `attribute → response field`) and this function is just traversing the list of requesting attributes generically ; or we build (earlier) some kind of object where each field represent an attribute (and can eventually rely on typing)

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,154 @@
...
+ * @param {object} log - Werelogs logger
+ * @param {function} callback - callback optional to keep backward compatibility
+ * @returns {Promise<object>} - { xml, responseHeaders }
+ * @throws {ArsenalError} NoSuchVersion - if versionId specified but not found
+ * @throws {ArsenalError} NoSuchKey - if object not found
+ * @throws {ArsenalError} MethodNotAllowed - if object is a delete marker
+ */
+async function objectGetAttributes(authInfo, request, log, callback) {
+  if (callback) {
+    return objectGetAttributes(authInfo, request, log)
```
francoisferrand, line 71: nit: not proper with this implementation, but do we actually _need_ this callback form? Since we are introducing the function, it is not already used, so can't we directly use only the async form?

(such callback trampoline was supposed to be used for migration, not in new code)

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,154 @@
...
+      key: objectKey,
+      versionId,
+      error: err,
+    });
+    throw err;
+  }
+
+  const responseHeaders = collectCorsHeaders(headers.origin, request.method, bucket);
+
+  if (!objMD) {
```
francoisferrand: can this really happen?
would we not have an exception in `validateBucketAndObj` and throw (line 107) if objMD is false-ish ?

lib/metadata/metadataUtils.js
```diff
@@ -281,6 +282,16 @@ function standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log,
         return callback(null, bucket, objMD);
     });
 }
+
+standardMetadataValidateBucketAndObj[promisify.custom] = (params, action, log) => new Promise((resolve, reject) => {
```
francoisferrand: this is already in Edouard's PR about migration to async I think, should pay attention to avoid duplicating when merging : i.e. only in the first PR merged...
(implementing it exactly the same way & at the same place would help, as it would allow `git` to act as gatekeeper, by having a conflict)

lib/api/objectGetAttributes.js
maeldonn: Using the async version directly involves making changes in arsenal that are outside of the scope of this feature.

lib/api/objectGetAttributes.js
maeldonn: The function return objMD as undefined when the object doesn't exist...

lib/metadata/metadataUtils.js
maeldonn: Updated to match Edouard PR

lib/api/objectGetAttributes.js
francoisferrand: Did you test this is indeed the behavior of AWS?

Looking at the [documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html#API_GetObjectAttributes_ResponseElements), the x-amz-delete-marker attribute is mentioned in the context of a 200 response:

> If the action is successful, the service sends back an HTTP 200 response.
> The response returns the following HTTP headers.
> [...]
> [x-amz-delete-marker](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html#API_GetObjectAttributes_ResponseSyntax)

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,160 @@
...
+      versionId,
+    });
+    const err = versionId ? errors.NoSuchVersion : errors.NoSuchKey;
+    err.responseHeaders = responseHeaders;
+    throw err;
+  }
+
+  responseHeaders['x-amz-version-id'] = getVersionIdResHeader(bucket.getVersioningConfiguration(), objectMD);
+  responseHeaders['Last-Modified'] = objectMD['last-modified'] && new Date(objectMD['last-modified']).toUTCString();
+
```
francoisferrand, line 130: the API supports a [x-amz-expected-bucket-owner](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html#API_GetObjectAttributes_RequestSyntax) header, which is currently not implemented?
→ please create a followup ticket

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,160 @@
...
+const { getPartCountFromMd5 } = require('./apiUtils/object/partInfo');
+
+const checkExpectedBucketOwnerPromise = promisify(checkExpectedBucketOwner);
+const validateBucketAndObj = promisify(standardMetadataValidateBucketAndObj);
+
+const OBJECT_GET_ATTRIBUTES = 'objectGetAttributes';
+const ATTRIBUTE_HANDLERS = {
+  ETag: objMD => objMD['content-md5'],
+  ObjectParts: objMD => {
+    const partCount = getPartCountFromMd5(objMD);
```
francoisferrand: the number of parts can be retrieved simply from `objMD.locations` field

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,160 @@
...
+    return partCount ? { PartsCount: partCount } : undefined;
```
francoisferrand: in the API, this is supposed to return each part
- i.e. array of partNumber and size
- checksum of each part: not implemented yet, this is ongoing work... but we should create the ticket (or add it in the ticket about supporting object checksum through this API)
- pagination with configurable [x-amz-max-parts](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html#API_GetObjectAttributes_RequestSyntax)

if we don't do it now, do we have a followup ticket?

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,160 @@
...
+    if (requestedAttrs.has(attr)) {
+      const value = handler(objMD);
+      if (value !== undefined) {
+        attrResp[attr] = value;
+      }
+    }
+  }
+
+  const builder = new xml2js.Builder();
+  return builder.buildObject({ GetObjectAttributesResponse: attrResp });
```
francoisferrand: all APIs build XML by hand (for performance reasons I think).
Is there a reason not to do the same here?

(this actually creates extra friction in next PR, to make code reusable...)

lib/api/objectGetAttributes.js
maeldonn: Yes I tested. No, it's not. The documentation is misleading, I don't think the two sentences are related.

lib/api/objectGetAttributes.js
maeldonn: This is already implemented and covered with a test.

lib/api/objectGetAttributes.js
maeldonn: It doesn't really simplify. We will use the locations field when implementing the checksum part.

lib/api/objectGetAttributes.js
maeldonn: This only works when a checksum is specified for each part, which isn't supported in CloudServer. Further details can be found in the follow-up ticket.

lib/api/objectGetAttributes.js
maeldonn: Updated to build XML by hand.

lib/api/objectGetAttributes.js
```diff
@@ -0,0 +1,170 @@
+const { promisify } = require('util');
```
francoisferrand, line 1: nit: just noticed this is indented with 2 spaces only, while we use 4-space indent overall...

francoisferrand (approved):

lib/api/objectGetAttributes.js
maeldonn: Yes used 2 space indentation on new files as it described here https://scality.atlassian.net/wiki/spaces/OS/pages/3726508042/Formatting+baseline+for+our+JS+TS+projects

PR #6061, opened 2026-01-26
CLDSRV-839: Multi delete raftSession in logs

BourgoisMickael:
- Include raft in server access logs
- Replace `REST.POST.OBJECT` => `REST.POST.MULTI_OBJECT_DELETE`
- Add internal `BATCH.DELETE.OBJECT` lines for each key

### DIFF
--- a/lib/api/multiObjectDelete.js
+++ b/lib/api/multiObjectDelete.js
@@ -31,6 +31,7 @@ const versionIdUtils = versioning.VersionID;
 const { data } = require('../data/wrapper');
 const logger = require('../utilities/logger');
 const { processBytesToWrite, validateQuotas } = require('./apiUtils/quotas/quotaUtils');
+const { logBatchDeleteObject } = require('../utilities/serverAccessLogger');
 
 /*
    Format of xml request:
@@ -371,14 +372,24 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                             callback(err, objMD, deleteInfo, result.versionId));
                 },
             ], (err, objMD, deleteInfo, versionId) => {
+                const requestID = log.getSerializedUids();
+
                 if (err === skipError) {
+                    // Object doesn't exist - log without object size (AWS behavior)
+                    logBatchDeleteObject(request, requestID, entry.key, null, null);
                     return moveOn();
                 } else if (err === objectLockedError) {
                     errorResults.push({ entry, error: errors.AccessDenied, objectLocked: true });
+                    // Log locked object with size if available
+                    const objectSize = objMD && objMD['content-length'] ? objMD['content-length'] : null;
+                    logBatchDeleteObject(request, requestID, entry.key, objectSize, errors.AccessDenied);
                     return moveOn();
                 } else if (err) {
                     log.error('error deleting object', { error: err, entry });
                     errorResults.push({ entry, error: err });
+                    // Log error case with size if available
+                    const objectSize = objMD && objMD['content-length'] ? objMD['content-length'] : null;
+                    logBatchDeleteObject(request, requestID, entry.key, objectSize, err);
                     return moveOn();
                 }
                 if (deleteInfo.deleted && objMD['content-length']) {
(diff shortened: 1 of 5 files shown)


lib/api/multiObjectDelete.js
```diff
@@ -408,6 +419,9 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
                     entry, isDeleteMarker,
                     deleteMarkerVersionId,
                 });
+                // Log successful deletion with object size
+                const objectSize = objMD && objMD['content-length'] ? objMD['content-length'] : null;
+                logBatchDeleteObject(request, requestID, entry.key, objectSize, null);
```
leif-scality: The log should be done in `onFinish` in `server.js` for the timers to be correct.

lib/utilities/serverAccessLogger.js
```diff
@@ -231,7 +231,7 @@ const methodToResType = Object.freeze({
     'listMultipartUploads': 'UPLOADS',
     'listParts': 'UPLOAD',
     'metadataSearch': 'OBJECT',
-    'multiObjectDelete': 'OBJECT',
+    'multiObjectDelete': 'MULTI_OBJECT_DELETE',
```
leif-scality, line 260: is this what AWS sets?

lib/api/multiObjectDelete.js
```diff
@@ -371,14 +372,24 @@ function getObjMetadataAndDelete(authInfo, canonicalID, request,
...
+                    // Log locked object with size if available
+                    const objectSize = objMD && objMD['content-length'] ? objMD['content-length'] : null;
+                    logBatchDeleteObject(request, requestID, entry.key, objectSize, errors.AccessDenied);
                     return moveOn();
                 } else if (err) {
                     log.error('error deleting object', { error: err, entry });
                     errorResults.push({ entry, error: err });
+                    // Log error case with size if available
+                    const objectSize = objMD && objMD['content-length'] ? objMD['content-length'] : null;
+                    logBatchDeleteObject(request, requestID, entry.key, objectSize, err);
```
leif-scality: Why do we want to log this operation if it will not be consumed by log courrier ?
Does AWS log all the deletes when doing a `multiObjectDelete`?

lib/utilities/serverAccessLogger.js
leif-scality: yes

lib/api/multiObjectDelete.js
BourgoisMickael: internal requests (both batch delete and `REST.COPY.*_GET` doesn't hold any timing information

BourgoisMickael: But maybe instead of logging immediately I could store the objects and queue them in memory to log everything at the end with this regular log. But that would put a little more memory usage

BourgoisMickael: Change the behavior to queue each internal request to log them at the end of request: [f7a31fb](https://github.com/scality/cloudserver/pull/6061/commits/f7a31fb31429a1b9b7114a2ff37972b04ebda07b)

lib/utilities/serverAccessLogger.js
```diff
@@ -472,13 +494,101 @@ function logServerAccess(req, res) {
...
+            }
+        }
+    }
+
+    serverAccessLogger.write(`${JSON.stringify(logEntry)}\n`);
 }
 
 module.exports = {
     logServerAccess,
+    logBatchDeleteObject,
```
dvasilas: Probably does not need to be exported?

dvasilas (approved):

tcarmet (approved):

lib/utilities/serverAccessLogger.js
BourgoisMickael: yep, dropped there [86d9120](https://github.com/scality/cloudserver/pull/6061/commits/86d912057d4c366df42f13c57606a43c6aeb1eae)

PR #6063, opened 2026-01-28
CLDSRV-839: Fix Copy access logs (source, destination)

BourgoisMickael:
- separate source and destination server access logging config
- log access to source object:
  - `REST.COPY.OBJECT_GET`
  - `REST.COPY.PART_GET`


lib/metadata/metadataUtils.js
```diff
@@ -17,22 +17,49 @@ const {
...
+ * @param {object} [options] - Optional configuration
+ * @param {boolean} [options.copySource=false] - If true, stores information in request.sourceServerAccessLog
+ *   instead of request.serverAccessLog. Used for copy operations to preserve both source and destination
+ *   bucket information separately. When false (default), stores in request.serverAccessLog.
+ * @return {undefined}
+ */
+function storeServerAccessLogInfo(request, bucket, raftSessionId, options = {}) {
     /* eslint-disable no-param-reassign */
 
     if (!request || !request.serverAccessLog) {
```
dvasilas, line 34: If `!request.serverAccessLog == true` does this mean that the destination bucket does not have bucket logging enabled?
But the source may have logging enabled, so we shouldn't return ?

BourgoisMickael: No this field is for the global configurable activation of server access logs in cloudserver: https://github.com/scality/cloudserver/blob/39ec017c0ea6e3b3e0f63cd4771147e77a2ffc2e/lib/server.js#L135-L158

For bucket specific activation of bucket loging there are the flags `request.serverAccessLog.enabled` and `request.serverAccessLog.loggingEnabled`

dvasilas: oh right, thanks.

lib/api/objectCopy.js
```diff
@@ -599,6 +618,16 @@ function objectCopy(authInfo, request, sourceBucket,
         const corsHeaders = collectCorsHeaders(request.headers.origin,
             request.method, destBucketMD);
 
+        // Initialize the queue for internal log request logging
+        initializeInternalLogRequestQueue(request);
+        // Queue the source-side access log (REST.COPY.OBJECT_GET)
+        queueInternalLogRequest(request, {
+            operation: 'REST.COPY.OBJECT_GET',
```
dvasilas, line 631: Not sure `REST.COPY.OBJECT_GET` (and `PART_GET`) fits well with `REST.HTTP_method.resource_type` as AWS described the `operation` field in documentation: https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html#log-record-fields

BourgoisMickael: Those are the actual methods used by aws in bucket logging (even if they are not documented)

BourgoisMickael: Example: https://eu-north-1.console.aws.amazon.com/s3/object/mick-mpu?region=eu-north-1&prefix=tstlogs/2026-01-28-03-38-06-44898754B296BB2E

dvasilas (approved):

tcarmet (approved):

PR #6064, opened 2026-01-28
CLDSRV-839: Fix UploadPart server access logs & objectSize

BourgoisMickael:
Fix UploadPart missing from access logs, and missing objectSize.


leif-scality (approved):

dvasilas (approved):
An additional improvement would be to add `objectSize` to tests (but could be done later).

PR #6070, opened 2026-02-04
Return user metadata in GetObjectAttributes API

maeldonn:
Issue: CLDSRV-844


lib/api/apiUtils/authorization/prepareRequestContexts.js
```diff
@@ -261,6 +261,23 @@ function prepareRequestContexts(apiMethod, request, sourceBucket,
...
+                generateRequestContext('objectGetVersionAttributes'),
+            );
+        } else {
+            requestContexts.push(
+                generateRequestContext('objectGet'),
+                generateRequestContext('objectGetAttributes'),
+            );
+        }
+
+        const attributes = request.headers['x-amz-object-attributes']?.split(',') ?? [];
```
SylvainSenechal, line 277: If I remember correctly, you had a function in the previous pr to get/split those attributes, or am i wrong ?

SylvainSenechal: ah it's the parseAttributesHeaders function just on the next file, yeah shouldn't you use parseAttributesHeaders or is there a problem with using it in this specific context ?

lib/api/apiUtils/object/parseAttributesHeader.js
```diff
@@ -1,21 +1,29 @@
 const { errorInstances } = require('arsenal');
-const { supportedGetObjectAttributes } = require('../../../../constants');
 
 /**
- * parseAttributesHeaders - Parse and validate the x-amz-object-attributes header
- * @param {object} headers - request headers
- * @returns {string[]} - array of valid attribute names
- * @throws {Error} - InvalidRequest if header is missing/empty, InvalidArgument if attribute is invalid
+ * Parse and validate attribute headers from a request.
```
SylvainSenechal: nit : 
- you can specify it's comma separated headers only that are parsed (or make the split parameter an argument of the function but this is probably going too far), or just rename the function parseCsvAttributesHeaders ?
- since you already took the time to document the function, maybe you could provide an example of header and parsed result after the @throws documentation

lib/api/apiUtils/object/parseAttributesHeader.js
```diff
@@ -1,21 +1,29 @@
...
  */
-function parseAttributesHeaders(headers) {
-  const attributes = headers['x-amz-object-attributes']?.split(',').map(attr => attr.trim()) ?? [];
-  if (attributes.length === 0) {
+function parseAttributesHeaders(headers, headerName, supportedAttributes, isRequired = false) {
+  const attributes =
+    headers[headerName]
+      ?.split(',')
+      .map(attr => attr.trim())
+      .map(attr => (supportedAttributes.has(attr) ? attr : attr.toLowerCase())) ?? [];
```
SylvainSenechal: wait so it means even if the attribute is not in supportedAttribute, it will be returned as lower cased ?
Shouldn't this be a filter instead ? 

I mean, ok I guess the overall function works because later there is the attributes.some() and has() thing, but could it be clearer

SylvainSenechal: Anyways, maybe the funciton is good enough as it is, but if you're able to make it a bit clearer it's nice, cause it's doing multiple thing that aren't completely obvious from the function name, it looks like it's gonna do some generic parsing, then ends up doing lowercasing and having a special case for x-amz-meta. Maybe renaming function name and intermediates variables help

lib/api/objectGetAttributes.js
```diff
@@ -47,10 +48,33 @@ function buildXmlResponse(objMD, attributes) {
...
   return builder.buildObject({ GetObjectAttributesResponse: attrResp });
 }
 
+/**
+ * parseUserMetadata - Extract user metadata from object metadata
+ * @param {object} objMD - object metadata
+ * @param {string[]} attributes - requested attributes
+ * @returns {object} - object containing requested user metadata key-value pairs
+ */
+function parseUserMetadata(objMD, attributes) {
```
SylvainSenechal: It would've been nicer if attributes was already filtered, to have the function be more specific. Also can consider a classic for loop, I don't think users usually set a lot of different metadata, but it can still be a few kb of metadata, and this way no need for eslint-disable

This is longer but not very hard to read compared with the filter with triple boolean. Maybe with this loop there is no need to filter the attributes into metadataAttributes


```suggestion
function extractUserMetadata(objMD, metadataAttributes) {
  const result = {};
  
  if (metadataAttributes.includes('x-amz-meta-*')) {
    for (const key of Object.keys(objMD)) {
      if (key.startsWith('x-amz-meta-') && objMD[key]) {
        result[key] = objMD[key];
      }
    }
  } else {
    for (const key of metadataAttributes) {
      if (objMD[key]) {
        result[key] = objMD[key];
      }
    }
  }
  
  return result;
}```

File: tests/functional/aws-node-sdk/test/object/objectGetAttributes.js [hunk 5]
```diff
@@ -262,3 +263,196 @@ describe('Test get object attributes with multipart upload', () => {
...
+        key1: 'value1',
+        key2: 'value2',
+        key3: 'value3',
+      },
+    }).promise();
+
+    const response = await getObjectAttributesWithUserMetadata(s3, {
+      Bucket: bucket,
+      Key: key,
+    }, 'x-amz-meta-*');
```
SylvainSenechal: can you add a test to see what happens if user request both 
x-amz-meta-*, and another x-amz-meta-something 

maybe also when x-amz-meta-* is requested twice, or when a specific metadata is requested twice

SylvainSenechal (changes requested):
some comments to address, 2 tests to add

lib/api/apiUtils/object/parseAttributesHeader.js
maeldonn: If it's not a supported attribute:
1. It's user metadata so it's ok
2. It's something else -> will throw an exception

lib/api/apiUtils/authorization/prepareRequestContexts.js
maeldonn: We want to check rights before starting to validate headers

lib/api/apiUtils/object/parseAttributesHeader.js
maeldonn: Added an example section in jsdoc

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
maeldonn: Added the two tests 🙌

SylvainSenechal (approved):

lib/api/apiUtils/object/parseAttributesHeader.js
francoisferrand: this seems uselessly innefficient: we traverse the list 3 times to do the same thing...
- trim() can be done in the same iteration, no need to the extra strings/list allocation
- isn't user metadata already mandated (by spec) to be lower case? do we do the same lower case conversion in other APIs? if not, not need to be more accommodating here
- if an attribute is in the `supportedAttributes`, there is no point making another check for the prefix (and rechecking later it is a supported attribute, if there is no prefix)

```suggestion
  const attributes =
    headers[headerName]?.split(',').map(attr => {
		const res = attr.trim())
		if (!supportedAttributes.has(res) && !res.startsWith('x-amz-meta-')) {
            throw errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified.');
        }
        return res;
    });
```

lib/api/bucketGet.js
```diff
@@ -332,16 +333,12 @@ function bucketGet(authInfo, request, log, callback) {
...
-            .map(attr => attr !== 'RestoreStatus' ? attr.toLowerCase() : attr)
-        ?? [];
-    if (optionalAttributes.some(attr => !attr.startsWith('x-amz-meta-') && attr != 'RestoreStatus')) {
-        return callback(
-            errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified')
-        );
+    let optionalAttributes;
+    try {
+        const headerName = 'x-amz-optional-object-attributes';
+        optionalAttributes = parseAttributesHeaders(request.headers, headerName, new Set(['RestoreStatus']));
```
francoisferrand: The `Set` should not be allocated on every call, should be a constant

lib/api/objectGetAttributes.js
```diff
@@ -47,10 +48,37 @@ function buildXmlResponse(objMD, attributes) {
...
   return builder.buildObject({ GetObjectAttributesResponse: attrResp });
 }
 
+/**
+ * extractUserMetadata - Extract requested user metadata from object metadata
+ * @param {object} objMD - object metadata
+ * @param {string[]} attributes - requested attributes
+ * @returns {object} - object containing requested user metadata key-value pairs
+ */
+function extractUserMetadata(objMD, attributes) {
```
francoisferrand: there is already this logic in ListObjectv2, must be factorized...

lib/api/objectGetAttributes.js
```diff
@@ -137,7 +165,7 @@ async function objectGetAttributes(authInfo, request, log, callback) {
     throw err;
   }
 
-  const attributes = parseAttributesHeaders(headers);
+  const attributes = parseAttributesHeaders(headers, 'x-amz-object-attributes', supportedGetObjectAttributes, true);
```
francoisferrand: this `true` here makes it hard to understand what is happening, and it can just as well be done here (it does not add any value to have a conditional behavior inside the `parseAttributesHeaders` function):

```suggestion
  const attributes = parseAttributesHeaders(headers, 'x-amz-object-attributes', supportedGetObjectAttributes);
  if (attributes.length === 0) {
      throw ....
  }
```

francoisferrand (changes requested):

lib/api/apiUtils/object/parseAttributesHeader.js
```diff
@@ -1,25 +1,43 @@
...
-    );
+function parseAttributesHeaders(headers, headerName, supportedAttributes) {
+  const rawValue = headers[headerName];
+  if (rawValue === null || rawValue === undefined) {
+    return new Set();
   }
 
-  if (attributes.some(attr => !supportedGetObjectAttributes.has(attr))) {
-    throw errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified.');
+  const result = new Set();
```
DarkIsDude: You can move this one before the if and return result line 21 ? No need to `new Set()` twice ?

lib/api/objectGetAttributes.js
```diff
@@ -131,7 +136,14 @@ async function objectGetAttributes(authInfo, request, log, callback) {
     throw err;
   }
 
-  const requestedAttrs = parseAttributesHeaders(headers);
+  const attrHeader = headers['x-amz-object-attributes'];
+  if (attrHeader === undefined) {
```
DarkIsDude: We don't have something in the validation for that ? Because we can fail really really really fast for that, even before getting the bucket

tests/unit/api/objectGetAttributes.js
```diff
@@ -348,6 +348,179 @@ describe('objectGetAttributes API with multipart upload', () => {
...
+        const [xml] = await objectGetAttributesAsync(authInfo, testGetRequest, log);
+        const result = await parseStringPromise(xml);
+        const response = result.GetObjectAttributesResponse;
+
+        assert.strictEqual(response.ETag[0], expectedMD5);
+        assert.strictEqual(response.ObjectSize[0], String(body.length));
+        assert.strictEqual(response['x-amz-meta-custom'][0], 'custom-value');
+    });
+
+    it('should not include x-amz-meta-* marker in response when wildcard is used', async () => {
```
DarkIsDude: I'm not sure it's useful to have a dedicated test for that, we can probably add it somewhere already

lib/api/apiUtils/object/parseAttributesHeader.js
maeldonn: Updated to return the `result` variable in the early return.

lib/api/objectGetAttributes.js
maeldonn: It was in validation before, but I was asked to move it outside since it only for GetObjectAttributes.

DarkIsDude: 🤷 Ok, don't know why but if it was asked, let's go

DarkIsDude (approved):

tests/unit/api/objectGetAttributes.js
maeldonn: Why ? Can you develop your thoughts ?

DarkIsDude: The test name can be `it should return all user metadata in response when wildcard`, it's a test that you already have. You can change with `should return only all user metadata` if you want, but you can assert this test in a previous one.

lib/api/apiUtils/object/extractUserMetadata.js
```diff
@@ -0,0 +1,23 @@
...
+ * @returns {object} - object containing requested user metadata key-value pairs
+ */
+function extractUserMetadata(metadata, attributes) {
+    const result = {};
+
+    const isWildcard = attributes.has('x-amz-meta-*');
+    const sourceKeys = isWildcard ? Object.keys(metadata) : attributes;
+
+    for (const key of sourceKeys) {
+        const isValidKey = isWildcard ? key.startsWith('x-amz-meta-') : true;
```
francoisferrand: instead of checking `isWildcard` for every key, then further checking `isValidKey`, this could be moved to the "sourceKeys" computation to avoid extra cycles & improve readability....

```suggestion
    const sourceKeys = isWildcard ? Object.keys(metadata).find(key => key.startsWith('x-amz-meta-')) : attributes;

    for (const key of sourceKeys) {
       if (metadata[key] != null) {
```

...and maybe can be further optimized, since `metadata[key] != null` will always be true in that case

lib/api/apiUtils/object/parseAttributesHeader.js
```diff
@@ -1,25 +1,43 @@
...
   }
 
-  if (attributes.some(attr => !supportedGetObjectAttributes.has(attr))) {
-    throw errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified.');
+  for (const rawAttr of rawValue.split(',')) {
+    let attr = rawAttr.trim();
+
+    if (!supportedAttributes.has(attr)) {
+      attr = attr.toLowerCase();
+    }
```
francoisferrand: do we need to handle "lower case" conversion?
is it the behavior for other API (listObjevtv2, get user metadata...), or are these -and AWS- 'strictly' case-sensitive?

lib/api/objectGetAttributes.js
benzekrimaha: if we're checking wether it's missing or empty shouldn't we go for if(!attrHeader) more generic instead ?

tests/functional/aws-node-sdk/test/object/objectGetAttributes.js
```diff
@@ -252,3 +254,269 @@ describe('Test get object attributes with multipart upload', () => {
...
+                Bucket: bucket,
+                Key: key,
+            }, 'ETag,x-amz-meta-custom,ObjectSize');
+
+            assert.strictEqual(response.ETag, expectedMD5);
+            assert.strictEqual(response.ObjectSize, body.length);
+            assert.strictEqual(response['x-amz-meta-custom'], 'custom-value');
+        });
+
+        it('should return all metadata when wildcard is combined with specific metadata key', async () => {
```
benzekrimaha:
```suggestion
        it('should return all metadata once wildcard is provided', async () => {
```

tests/unit/api/objectGetAttributes.js
```diff
@@ -348,6 +348,166 @@ describe('objectGetAttributes API with multipart upload', () => {
...
+        const testGetRequest = createGetAttributesRequest(['x-amz-meta-foo', 'x-amz-meta-bar']);
+        const [xml] = await objectGetAttributesAsync(authInfo, testGetRequest, log);
+        const result = await parseStringPromise(xml);
+        const response = result.GetObjectAttributesResponse;
+
+        assert.strictEqual(response['x-amz-meta-foo'][0], 'foo-value');
+        assert.strictEqual(response['x-amz-meta-bar'][0], 'bar-value');
+    });
+
+    it('should return only all user metadata when x-amz-meta-* is requested', async () => {
```
benzekrimaha:
```suggestion
    it('should return all user metadata when x-amz-meta-* is requested', async () => {
```

benzekrimaha (approved):
couple comments but nothing blocking for me , approving

lib/api/objectGetAttributes.js
maeldonn: This is the behavior of AWS -> empty header leads to an InvalidArgument exception.

lib/api/bucketGet.js
```diff
@@ -248,34 +252,21 @@ function processMasterVersions(bucketName, listParams, list) {
...
-                for (const key of Object.keys(item.userMetadata)) {
-                    userMetadata.add(key);
-                }
-                break;
-            default:
-                if (item.userMetadata?.[attribute]) {
-                    userMetadata.add(attribute);
-                }
+
+    if (optionalAttributes.has('RestoreStatus')) {
```
francoisferrand: the previous code was done purposely for efficiency to avoid repeatedly iterating through every attribute : not really an improvement here...

esp. ideally we want to reuse the whole logic (i.e. same single pass iteration), not just `extractUserMetadata`. Especially we don't really care if extra attributes are supported in this "generation" stage, as the "supported" attributes will be validated earlier when receiving the API (some the `optionalAttributes` should only contain the allowed parameters.

(but this requires following the same pattern as all APIs, and generate XML manually instead of using `xml2js.Builder`)

lib/api/apiUtils/object/parseAttributesHeader.js
maeldonn: It's the behavior of listObjectV2 for user metadata.

lib/api/bucketGet.js
maeldonn: Reverted to the listObjectV2 implementation and created an utility function for the two APIs.

lib/api/bucketGet.js
```diff
@@ -150,7 +153,7 @@ function processVersions(bucketName, listParams, list) {
             `<ID>${v.Owner.ID}</ID>`,
             `<DisplayName>${v.Owner.DisplayName}</DisplayName>`,
             '</Owner>',
-            ...processOptionalAttributes(v, listParams.optionalAttributes),
+            ...buildAttributesXml(v, v.userMetadata, listParams.optionalAttributes),
             `<StorageClass>${v.StorageClass}</StorageClass>`,
```
francoisferrand, line 161: for the sake of performance, best to keep using push : "unpack" will create an extra copy...

```
xml.push(...)
buildAttributesXml(v, v.userMetadata, listParams.optionalAttributes, xml)
xml.push(...)
```

lib/api/bucketGet.js
```diff
@@ -231,7 +234,7 @@ function processMasterVersions(bucketName, listParams, list) {
             );
         }
 
-        xml.push(...processOptionalAttributes(v, listParams.optionalAttributes));
+        xml.push(...buildAttributesXml(v, v.userMetadata, listParams.optionalAttributes));
```
francoisferrand: same here, and trivial to integrate

lib/api/objectGetAttributes.js
```diff
@@ -20,39 +20,32 @@ const OBJECT_GET_ATTRIBUTES = 'objectGetAttributes';
...
-        }
-        break;
-      }
-      case 'StorageClass':
-        xml.push(`<StorageClass>${objMD['x-amz-storage-class']}</StorageClass>`);
-        break;
-      case 'ObjectSize':
-        xml.push(`<ObjectSize>${objMD['content-length']}</ObjectSize>`);
-        break;
+    ...buildAttributesXml(objMD, userMetadata, requestedAttrs),
```
francoisferrand: best to just use `push` and avoid repeated copies to unpack/join strings....

lib/api/objectGetAttributes.js
```diff
@@ -20,39 +20,32 @@ const OBJECT_GET_ATTRIBUTES = 'objectGetAttributes';
...
+/**
+ * getUserMetadata - Retrieves all object user metadata
+ * @param {object} objMD - object metadata
+ * @returns {object} - user metadata key-value pairs
+ */
+function getUserMetadata(objMD) {
+  const metadata = {};
+
+  for (const key in objMD) {
+    if (key.startsWith('x-amz-meta-') && objMD[key] !== undefined) {
```
francoisferrand: do we actually need the second condition?
user metadata are only added on demand, so if "empty" (i.e. undefined) we should probably return them anyway?

or is there a specific way that we end up with such undefined custom metadata that the user did not ask for?

(we will need however to properly "serialize" undefined to an empty string, though)

francoisferrand (approved):

francoisferrand:
don't forget to rebase to "squash" the fixup before merging, and probably add extra commit to bump package.json for 9.3.3 release 🙏

PR #6073, opened 2026-02-04
🎨 format on bucketGet and objectGetLegalHold for async/await

DarkIsDude:
Issue: CLDSRV-823

# Pull request template

### Motivation and context

### Related issues

current repository

### Sign your work

as soon as we can.

### DIFF
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -179,9 +179,9 @@ function processMasterVersions(bucketName, listParams, list) {
     if (listParams.v2) {
         xmlParams.push({ tag: 'StartAfter', value: listParams.startAfter || '' });
         xmlParams.push({ tag: 'FetchOwner', value: `${listParams.fetchOwner}` });
-        xmlParams.push({ tag: 'ContinuationToken', value: generateToken(listParams.continuationToken) || '', });
-        xmlParams.push({ tag: 'NextContinuationToken', value: generateToken(list.NextContinuationToken), });
-        xmlParams.push({ tag: 'KeyCount', value: list.Contents ? list.Contents.length : 0, });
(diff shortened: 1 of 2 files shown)


lib/api/bucketGet.js
```diff
@@ -182,31 +177,19 @@ function processMasterVersions(bucketName, listParams, list) {
...
-            tag: 'NextContinuationToken',
-            value: generateToken(list.NextContinuationToken),
-        });
-        xmlParams.push({
-            tag: 'KeyCount',
-            value: list.Contents ? list.Contents.length : 0,
-        });
+        xmlParams.push({ tag: 'StartAfter', value: listParams.startAfter || '' });
+        xmlParams.push({ tag: 'FetchOwner', value: `${listParams.fetchOwner}` });
+        xmlParams.push({ tag: 'ContinuationToken', value: generateToken(listParams.continuationToken) || '', });
```
maeldonn: Some trailing commas are still there. This should not be the case with a formatter.

maeldonn (commented):
What the goal of this PR ? We should wait the validation of the prettier configuration before starting formatting files. It can lead to unnecessary conflicts...

DarkIsDude:
> What the goal of this PR ? We should wait the validation of the prettier configuration before starting formatting files. It can lead to unnecessary conflicts...

You have the reason of this PR here https://github.com/scality/cloudserver/pull/6045#discussion_r2709947094. Prettier can fix some stuff but not all of them and I hope that we'll not do a one shot with prettier but first add prettier and will validate only file that are in the diff and after some months migrate remaining one, because if we do that, we'll have conflicts on all PR. Also prettier will not be able to fix everything and we'll always need to do some manual work.

francoisferrand (approved):

maeldonn (approved):

PR #6077, opened 2026-02-06
Improvement/cldsrv 846 migrate bucket logging tests

benzekrimaha:
Issue: CLDSRV-846


bin/ensureServiceUser
DarkIsDude, line 1: We don't have test for that ?

DarkIsDude (approved):

bin/ensureServiceUser
benzekrimaha: This file is actually used for cmd line commands , when I checked we indeed don't have tests for it as not really needed

delthas (approved):

PR #6081, opened 2026-02-17
CLDSRV-848: add missing checksum algorithms

leif-scality:
Add CRC32, CRC32C, CRC64NVME, SHA1, SHA256 checksum support for non streaming methods (all methods except PutObject and PutPart).


unholy0X (approved):

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,42 +1,226 @@
...
     MissingChecksum: 'MissingChecksum',
+    AlgoNotSupported: 'AlgoNotSupported',
+    AlgoNotSupportedSDK: 'AlgoNotSupportedSDK',
+    MultipleChecksumTypes: 'MultipleChecksumTypes',
+    MissingCorresponding: 'MissingCorresponding',
+    InvalidAlgoValue: 'InvalidAlgoValue',
 });
 
+const algorithms = {
+    'crc64nvme': {
```
jonathan-gramain: Normally you can remove all the quotes from property names

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,42 +1,226 @@
...
+        'digest': async data => {
+            const input = Buffer.isBuffer(data) ? data : Buffer.from(data);
+            const crc = new CrtCrc64Nvme();
+            crc.update(input);
+            const result = await crc.digest();
+            return Buffer.from(result).toString('base64');
+        },
+        'checkExpected': expected => {
```
jonathan-gramain: What about naming this callback `isValidDigest` (I understand you mean to check the "expected" digest, as if the digest is expected to match the contents' digest, but doesn't come obvious when reading the code, I think `isValidDigest` is more appropriate as it implies it returns true/false if the digest is a valid digest or not based on the expected type, without actually checking it against any content.

I don't know if it matters, but optionally we could also check that characters are valid within the expected encoding.

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,42 +1,226 @@
...
+
+function uint32ToBase64(num) {
+    const buf = Buffer.alloc(4);
+    buf.writeUInt32BE(num, 0);
+    return buf.toString('base64');
+}
+
+async function validateXAmzChecksums(headers, body) {
+    const checksumHeaders = Object.keys(headers).filter(header => header.startsWith('x-amz-checksum-'));
+    const xAmzCheckumCnt = checksumHeaders.length;
```
jonathan-gramain:
```suggestion
    const xAmzChecksumCnt = checksumHeaders.length;
```

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,42 +1,226 @@
...
+        return { error: ChecksumError.MissingChecksum, details: null };
+    }
+
+    // No x-amz-sdk-checksum-algorithm we expect one x-amz-checksum-[crc64nvme, crc32, crc32C, sha1, sha256].
+    const algo = checksumHeaders[0].split('-')[3];
+    if (typeof algo !== 'string') {
+        return { error: ChecksumError.AlgoNotSupported, details: { algorithm: algo } };
+    }
+
+    if (algo in algorithms === false) {
```
jonathan-gramain: nitpick: it's more typical to avoid explicit boolean value checks

```suggestion
    if (!(algo in algorithms)) {
```

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,42 +1,226 @@
...
 const { config } = require('../../../Config');
 
 const ChecksumError = Object.freeze({
     MD5Mismatch: 'MD5Mismatch',
+    XAmzMismatch: 'XAmzMismatch',
     MissingChecksum: 'MissingChecksum',
+    AlgoNotSupported: 'AlgoNotSupported',
+    AlgoNotSupportedSDK: 'AlgoNotSupportedSDK',
+    MultipleChecksumTypes: 'MultipleChecksumTypes',
+    MissingCorresponding: 'MissingCorresponding',
```
jonathan-gramain: Not a fan of this particular error name, could we either re-use `MissingChecksum` (is it useful to distinguish them?), or alternatively specify that it refers to a checksum like above: `ChecksumTypeMismatch` for example.

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,42 +1,226 @@
 const crypto = require('crypto');
+const { Crc32 } = require('@aws-crypto/crc32');
+const { Crc32c } = require('@aws-crypto/crc32c');
+const { CrtCrc64Nvme } = require('@aws-sdk/crc64-nvme-crt');
 const { errors: ArsenalErrors } = require('arsenal');
 const { config } = require('../../../Config');
 
 const ChecksumError = Object.freeze({
     MD5Mismatch: 'MD5Mismatch',
```
jonathan-gramain, line 31: Instead of creating internal error types that are not used for any logic nor returned to the client, we could instead directly construct the error when we see it and return it as is with an appropriate descriptive error message (i.e. construct directly an error with `InvalidRequest.customizeDescription(...)`). We can keep the special errors that we need special handling for (`MissingChecksum` etc.)

In the unit tests, you can simply check that the error is `InvalidRequest`, and if you want to be fancy, also check a part of the error message.

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,42 +1,226 @@
...
+async function defaultValidationFunc(request, body, log) {
+    const err = await validateChecksumsNoChunking(request.headers, body);
+    if (!err) {
+        return null;
     }
 
-    return null;
+    log.debug('failed checksum validation', { method: request.apiMethod }, err);
+
+    switch (err.error) {
```
jonathan-gramain, line 196: See my suggestion about returning directly errors to avoid having to do this conversion.

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -62,14 +249,13 @@ const methodValidationFunc = Object.freeze({
  * @param {object} log - logger
  * @return {object} - error
  */
-function validateMethodChecksumNoChunking(request, body, log) {
+async function validateMethodChecksumNoChunking(request, body, log) {
     if (config.integrityChecks[request.apiMethod]) {
         const validationFunc = methodValidationFunc[request.apiMethod];
         if (!validationFunc) {
-            return null;
+            return null; //await defaultValidationFunc2(request, body, log);
```
jonathan-gramain: Why adding this comment?

tests/functional/raw-node/test/xAmzChecksum.js
```diff
@@ -0,0 +1,319 @@
...
+        doTest(
+            headers,
+            methods[0],
+            400,
+            ['InvalidRequest', 'Expecting a single x-amz-checksum- header. Multiple checksum Types are not allowed.'],
+            done,
+        );
+    });
+
+    itSkipIfAWS('should respond InvalidRequest with if invalid x-amz-checksum- algorithm', done => {
```
jonathan-gramain:
```suggestion
    itSkipIfAWS('should respond InvalidRequest with invalid x-amz-checksum- algorithm', done => {
```

tests/functional/raw-node/test/xAmzChecksum.js
```diff
@@ -0,0 +1,319 @@
...
+                            },
+                            authCredentials
+                        ),
+                        res => {
+                            let data = '';
+                            res.on('data', chunk => {
+                                data += chunk;
+                            });
+                            res.on('end', () => {
+                                assert(!data.includes('BadDigest'));
```
jonathan-gramain, line 297: Could we more simply check that `res.statusCode == 200`?

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: renamed to `isValidDigest` and added base64 regex test

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: removed

tests/functional/raw-node/test/xAmzChecksum.js
leif-scality: done

tests/functional/raw-node/test/xAmzChecksum.js
leif-scality: The request is going to fail, because it is not a complete request, so we only check that it is not a BadDigest meaning that we passed the checksum checks. 

I added missing asserts to cover the other errors (InvalidDigest, ...)

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: The internal errors are have more details than the AWS errors(used for logging), they are also easier to test.
I prefer having the building of the AMZ error outside the validation function.

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: The internal errors are have more details than the AWS errors(used for logging), they are also easier to test.
I prefer having the building of the AMZ error outside the validation function.

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,75 +1,293 @@
...
+            }
+            if (expected.length != 44) {
+                return false;
+            }
+            return base64Regex.test(expected);
+        },
+    }
+});
+
+function uint32ToBase64(num) {
```
BourgoisMickael: can you move this function up so it's easier to understand as it's used in the code above

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,75 +1,293 @@
...
+            if (expected.length != 12) {
+                return false;
+            }
+            return base64Regex.test(expected);
+        },
+    },
+    crc32: {
+        digest: data => {
+            const input = Buffer.isBuffer(data) ? data : Buffer.from(data);
+            return uint32ToBase64(new Crc32().update(input).digest() >>> 0);
```
BourgoisMickael: Can you include a comment about the `>>> 0`  to describe it's used to coerce into an unsigned 32-bit number

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,75 +1,293 @@
...
+            return Buffer.from(result).toString('base64');
+        },
+        isValidDigest: expected => {
+            if (typeof expected !== 'string') {
+                return false;
+            }
+            if (expected.length != 12) {
+                return false;
+            }
+            return base64Regex.test(expected);
```
BourgoisMickael: could be simplified into 1 liner:
```suggestion
            return typeof expected === 'string' && expected.length === 12 && base64Regex.test(expected);
```

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,75 +1,293 @@
...
+        return {
+            error: ChecksumError.MissingCorresponding,
+            details: { expected: headers['x-amz-sdk-checksum-algorithm'] }
+        };
+    } else if (xAmzChecksumCnt === 0) {
+        return { error: ChecksumError.MissingChecksum, details: null };
+    }
+
+    // No x-amz-sdk-checksum-algorithm we expect one x-amz-checksum-[crc64nvme, crc32, crc32C, sha1, sha256].
+    const algo = checksumHeaders[0].split('-')[3];
```
BourgoisMickael: You could avoid allocating an array from `split` by using `slice` since you know the full prefix

```suggestion
    const HEADER_PREFIX = 'x-amz-checksum-';
    // ...

    const checksumHeaders = Object.keys(headers).filter(header => header.startsWith(HEADER_PREFIX));
    // ...
    const algo = checksumHeaders[0].slice(HEADER_PREFIX.length);
```

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,75 +1,293 @@
...
+    if (typeof algo !== 'string') {
+        return { error: ChecksumError.AlgoNotSupported, details: { algorithm: algo } };
+    }
```
BourgoisMickael: This check could be removed.

It's checked with the `algo in algorithms` below already.

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: changed to slice

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,75 +1,249 @@
...
 const ChecksumError = Object.freeze({
     MD5Mismatch: 'MD5Mismatch',
+    MD5Invalid: 'MD5Invalid',
+    XAmzMismatch: 'XAmzMismatch',
     MissingChecksum: 'MissingChecksum',
+    AlgoNotSupported: 'AlgoNotSupported',
+    AlgoNotSupportedSDK: 'AlgoNotSupportedSDK',
+    MultipleChecksumTypes: 'MultipleChecksumTypes',
+    MissingCorresponding: 'MissingCorresponding',
+    InvalidAlgoValue: 'InvalidAlgoValue',
```
jonathan-gramain: This error seems to be used not when the algorithm is invalid, but when the checksum is malformed. So I suggest to rename the error to `MalformedChecksum`.

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -1,75 +1,249 @@
...
+
 const ChecksumError = Object.freeze({
     MD5Mismatch: 'MD5Mismatch',
+    MD5Invalid: 'MD5Invalid',
+    XAmzMismatch: 'XAmzMismatch',
     MissingChecksum: 'MissingChecksum',
+    AlgoNotSupported: 'AlgoNotSupported',
+    AlgoNotSupportedSDK: 'AlgoNotSupportedSDK',
+    MultipleChecksumTypes: 'MultipleChecksumTypes',
+    MissingCorresponding: 'MissingCorresponding',
```
jonathan-gramain, line 38: nitpick: rename to `MissingExpectedChecksum`?

jonathan-gramain (approved):
Comments on error names, otherwise LGTM

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: I prefer `MissingCorresponding` because it points to this specific error `x-amz-sdk-checksum-algorithm specified, but no corresponding x-amz-checksum-* or x-amz-trailer headers were found.`.

`MissingExpectedChecksum` is not as precise

lib/api/apiUtils/integrity/validateChecksums.js

BourgoisMickael (approved):

PR #6082, opened 2026-02-23
Improvement/cldsrv 561 add tests for create and store object

benzekrimaha:
This PR strengthens coverage around createAndStoreObject, the core path used by putObject, putRestoredObject, and putObjectVersion, with emphasis on cold-storage restore and archived-object edge cases.
The goal is to improve confidence in behavior-heavy flows
Issue: CLDSRV-561

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/objectOverwrite.js
+++ b/tests/functional/aws-node-sdk/test/object/objectOverwrite.js
@@ -1,12 +1,49 @@
 const assert = require('assert');
 const {
     PutObjectCommand,
+    PutBucketVersioningCommand,
     HeadObjectCommand,
     GetObjectCommand,
 } = require('@aws-sdk/client-s3');
 
 const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
+const { fakeMetadataArchive, fakeMetadataTransition, getMetadata, initMetadata } = require('../utils/init');
+
+const coldStateScenarios = [
+    {
+        name: 'transition in progress',
+        transitionInProgress: true,
+    },
+    {
+        name: 'archived',
+        archiveState: {
+            archiveInfo: { archiveId: 'archive-1' },
+            restoreRequestedAt: new Date(0).toISOString(),
+            restoreRequestedDays: 5,
+        },
+    },
+    {
+        name: 'restored (not expired)',
+        archiveState: {
+            archiveInfo: { archiveId: 'archive-restored' },
+            restoreRequestedAt: new Date(0).toISOString(),
+            restoreRequestedDays: 5,
+            restoreCompletedAt: new Date(Date.now() - 60 * 60 * 1000).toISOString(),
+            restoreWillExpireAt: new Date(Date.now() + 24 * 60 * 60 * 1000).toISOString(),
+        },
+    },
+    {
+        name: 'restored (expired)',
+        archiveState: {
+            archiveInfo: { archiveId: 'archive-expired' },
+            restoreRequestedAt: new Date(0).toISOString(),
+            restoreRequestedDays: 5,
+            restoreCompletedAt: new Date(Date.now() - 48 * 60 * 60 * 1000).toISOString(),
+            restoreWillExpireAt: new Date(Date.now() - 24 * 60 * 60 * 1000).toISOString(),
+        },
+    },
+];
 
 const objectName = 'someObject';
 const firstPutMetadata = {
@@ -30,6 +67,7 @@ describe('Put object with same key as prior object', () => {
             bucketUtil = new BucketUtility('default', sigCfg);
             s3 = bucketUtil.s3;
             bucketName = await bucketUtil.createRandom(1);
+            await initMetadata();
         });
 
         beforeEach(async () => {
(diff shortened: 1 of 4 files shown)


tests/functional/aws-node-sdk/test/object/objectOverwrite.js
```diff
@@ -66,5 +73,57 @@ describe('Put object with same key as prior object', () => {
                 const bodyText = await res.Body.transformToString();
                 assert.deepStrictEqual(bodyText, 'Much different');
             });
+
+        it('should replace archived object in non-versioned bucket', async () => {
```
francoisferrand: is the behavior of putObject really specific when the object is archived? or is this just to handle the case  where there are actually no "data" associated  (e.g. `locations` is empty) associated with the object?

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
```diff
@@ -66,5 +73,57 @@ describe('Put object with same key as prior object', () => {
...
+            await fakeMetadataArchivePromise(bucketName, objectName, undefined, {
+                archiveInfo: { archiveId: 'archive-1' },
+                restoreRequestedAt: new Date(0).toISOString(),
+                restoreRequestedDays: 5,
+            });
```
francoisferrand: probably more than one case? should we have similar test cases (or a single table-driven test) for every "stage" of the cold object: transition in progress, archived, restored, expired ?

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
```diff
@@ -66,5 +73,57 @@ describe('Put object with same key as prior object', () => {
...
+
+            await s3.send(new PutObjectCommand({
+                Bucket: bucketName,
+                Key: objectName,
+                Body: 'overwrite archived data',
+                Metadata: secondPutMetadata,
+            }));
+
+            const currentMD = await getMetadataPromise(bucketName, objectName, undefined);
+            assert.strictEqual(currentMD.archive, undefined);
```
francoisferrand: Isn't there a special flag we need to pass to Arsenal/Metadata in that case, to ensure this operation can be picked'up by Backbeat in the oplog? (not sure, may not be needed in this specific case... or for some version-suspended case?)

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
```diff
@@ -66,5 +73,57 @@ describe('Put object with same key as prior object', () => {
...
+        });
+
+        it('should overwrite archived current object in versioned bucket', async () => {
```
francoisferrand: in a versioned bucket, there is no "overwrite" : it just creates another revision

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
```diff
@@ -66,5 +73,57 @@ describe('Put object with same key as prior object', () => {
...
+                Bucket: bucketName,
+                Key: objectName,
+                Body: 'versioned second payload',
+                Metadata: secondPutMetadata,
+            }));
+            assert(secondPutRes.VersionId);
+
+            const currentVersionMD = await getMetadataPromise(bucketName, objectName, undefined);
+            assert.strictEqual(currentVersionMD.archive, undefined);
+        });
```
francoisferrand, line 168: I feel we are missing may test cases : e.g.
- bucket can be version suspended
- previous version can be a "null version" (created before versioning was enabled)
- there is PutObject, but also our own extension "PutObject with version id" which is used when restoring an object : which creates many of the code paths in this function. (not sure how much is really tested, and probably should not go in "objectOverwrite" case : but we need to review if such cases are missing)

(but code coverage says we cover all these, so must be tested somewhere already, somehow...)

francoisferrand (commented):
Looking at code coverage (https://app.codecov.io/github/scality/cloudserver/pull/6082/blob/lib/api/apiUtils/object/createAndStoreObject.js?dropdown=coverage), we can confirm `createAndStoreObject` is already reletively well "covered".

There are some gaps indeed, but this PR does not cover them.

However, the missing parts is also sometimes more subtle, and requires checking the actual behavior : like do we set the right arsenal flag (needOplogUpdate)...

So it is not just adding tests:
- need to dig deeper to identify the specific behaviors which must be implemented by this function (like when to require oplog update, the `originOp`, etc...)
- then either find the tests which already cover these cases (there are some, c.f. coverage) and update them; or indeed add some extra tests to validate these specific behaviors

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
benzekrimaha: Good point. I added an explicit unit case for “no location and no archive info” and verified it does **not** trigger the archived replacement oplog path. This distinguishes the archive-specific behavior from generic “no data location” situations.

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
benzekrimaha: Yes, and this is now covered with a table-driven functional set in objectOverwrite: transition-in-progress, archived, restored (not expired), and restored (expired).

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
benzekrimaha: Yes it is covered at behavior level by asserting the metadata-store options in unit tests (needOplogUpdate / originOp) for archived replacement vs non-archived cases. This validates the flagging logic used to generate the oplog signal expected by Backbeat.

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
benzekrimaha: Agreed. I aligned the test intent and assertions accordingly: in versioned buckets we now assert that a **new** version is created and the previous one is preserved, rather than describing it as overwrite semantics.

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
benzekrimaha: Covered now through the combined suite:
- version-suspended/null-current behavior in objectOverwrite,
- restore-specific PutObject with x-scal-s3-version-id scenarios in putVersion, and unit assertions for restore metadata behavior (originOp, metadata preservation, restore-attempt cleanup, ingestion version-id handling).

SylvainSenechal (approved):
wouldn't we want (some of) these tests to be in zenko, to avoid relying on mocks ? I guess should be possible, running mongo commands to directly inspect database data

francoisferrand:
> wouldn't we want (some of) these tests to be in zenko, to avoid relying on mocks ? I guess should be possible, running mongo commands to directly inspect database data

some (if not all) of these tests are already in Zenko, as part of _functional_ / _e2e_ tests : i.e. can we restore an object (in the various cases)? do we get the proper bucket notification? ... But IMHO "too late" (in the dev process) to inspect mongo content in Zenko tests.

relying on mocks is good to test corner cases, and ideally we would have integration tests to cross-check the actual backend works as expected: i.e. running some tests against mongo to verify it behaves as expected. How relevant (or not) the integration tests are really depends on the specific case: e.g. if we are already sure mongo (or metadata) write the object MD as we provide them, then good enough to verify we call this metadata layer with the right information....

But we should try to keep these tests as close to the code as possible: i.e. we want to make sure Cloudserver behaves as expected with mongo backend, should be done here (but integration between backbeat and cloudserver would probably be in Zenko)

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
```diff
@@ -66,5 +74,138 @@ describe('Put object with same key as prior object', () => {
                 const bodyText = await res.Body.transformToString();
                 assert.deepStrictEqual(bodyText, 'Much different');
             });
+
+        [
+            {
+                name: 'transition in progress',
+                setup: () => fakeMetadataTransitionPromise(bucketName, objectName, undefined),
```
francoisferrand: the difference between these "setup" is really just the last parameter of `fakeMetadataArchivePromise` : instead of putting a `setup()` function, it would be more readable to specify that parameter only (with a meaningful name) and keep the whole logic in the test body

tests/functional/aws-node-sdk/test/object/objectOverwrite.js
```diff
@@ -1,12 +1,19 @@
...
     GetObjectCommand,
 } = require('@aws-sdk/client-s3');
 
 const withV4 = require('../support/withV4');
 const BucketUtility = require('../../lib/utility/bucket-util');
+const { fakeMetadataArchive, fakeMetadataTransition, getMetadata, initMetadata } = require('../utils/init');
+const fakeMetadataArchivePromise = promisify(fakeMetadataArchive);
+const fakeMetadataTransitionPromise = promisify(fakeMetadataTransition);
+const getMetadataPromise = promisify(getMetadata);
+const initMetadataPromise = promisify(initMetadata);
```
francoisferrand: as per our "async migration" process, should we not take the chance to promisify these functions directly (not update every place they are used, but make them optionally async) ?

(not really the case we discussed -this is not the code we modify-, but we don't take the step we will never migrate such code...)

tests/unit/api/createAndStoreObject.js
```diff
@@ -0,0 +1,629 @@
...
+        it('should create object successfully', done => {
+            const request = new DummyRequest({
+                bucketName,
+                namespace: 'default',
+                objectKey,
+                headers: { 'content-type': 'text/plain' },
+                url: `/${bucketName}/${objectKey}`,
+            }, Buffer.from('test data', 'utf8'));
+
+            createAndStoreObject(bucketName, testBucket, objectKey, null,
```
francoisferrand: should be promisified, to write "modern" async code?

francoisferrand (approved):

tests/functional/aws-node-sdk/test/object/putVersion.js
```diff
@@ -48,13 +48,13 @@ async function putObjectVersion(s3, params, vid, next) {
...
-    try {
+
+    const promise = (async () => {
         const res = await s3.send(command);
-        next(null, res);
-    } catch (err) {
-        next(err);
-    }
+        return res;
+    })();
```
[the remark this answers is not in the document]
benzekrimaha: fixed

benzekrimaha:
@bert-e create_pull_requests

PR #6085, opened 2026-02-23
Update LastModified date format to ISO string in XML response

ZiyangLinScality:
Issue [CLDSRV-857](https://scality.atlassian.net/browse/CLDSRV-857)

When browsing objects created by Veeam SOSAPI, the UI shows "An error occurred while loading data" even though the API returns a 200 response with valid XML.

Root cause: The LastModified field in the response uses RFC 2822 format (Mon, 23 Feb 2026 08:24:20 GMT) instead of ISO 8601 (2026-02-23T08:24:20.000Z). AWS SDK v3 strictly expects RFC 3339/ISO 8601 and throws TypeError: Invalid RFC3339 timestamp format during deserialization, causing the entire response to fail.

[CLDSRV-857]: https://scality.atlassian.net/browse/CLDSRV-857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/routes/veeam/list.js
+++ b/lib/routes/veeam/list.js
@@ -38,7 +38,7 @@ function buildXMLResponse(request, arrayOfFiles, versioned = false) {
         value: {
             IsDeleteMarker: false,
             IsNull: true,
-            LastModified: file['Last-Modified'],
+            LastModified: new Date(file['Last-Modified']).toISOString(),
             // Generated ETag alrady contains quotes, removing them here
             ETag: file.ETag.substring(1, file.ETag.length - 1),
             Size: file['Content-Length'],
--- a/tests/unit/routes/veeam-routes.js
+++ b/tests/unit/routes/veeam-routes.js
@@ -493,6 +493,44 @@ describe('Veeam routes - LIST request handling', () => {
         });
     });
 
+    it('should emit LastModified in ISO 8601 format in XML body', done => {
+        const request = createRequest();
+        const response = createResponse();
+
+        const xmlChunks = [];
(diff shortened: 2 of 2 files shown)


benzekrimaha (approved):

DarkIsDude (commented):
We should have test somehow 🙏

lib/routes/veeam/list.js
```diff
@@ -38,7 +38,7 @@ function buildXMLResponse(request, arrayOfFiles, versioned = false) {
         value: {
             IsDeleteMarker: false,
             IsNull: true,
-            LastModified: file['Last-Modified'],
+            LastModified: new Date(file['Last-Modified']).toISOString(),
```
anurag4DSB, line 41: We have unit tests for listVeeamFiles in `tests/unit/routes/veeam-routes.js` but none of them inspect the XML body. They only assert on status code and response.end being called. That means no existing test would have caught this bug, and nothing currently protects against it regressing.

I'd like to see a test added in the (suggesting) 'Veeam routes - LIST request handling' block that captures the XML output and asserts <LastModified> is ISO 8601. Something like `should emit LastModified in ISO 8601 format`

This is exactly the gap that let the original bug ship undetected.

anurag4DSB (commented):
Lets add the test and then we should be good.

anurag4DSB (approved):

DarkIsDude (approved):

PR #6088, opened 2026-02-25
👷(CLDSRV-860) Monitor async await migration

DarkIsDude:
https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration

### DIFF
--- a/.github/scripts/check-diff-async.mjs
+++ b/.github/scripts/check-diff-async.mjs
@@ -1,6 +1,6 @@
 /**
  * Check that all new/modified functions in the current git diff use async/await.
- * Fails with exit code 1 if any additions introduce callback-style functions or .then() chains.
+ * Fails with exit code 1 if any additions introduce callback-style functions.
  *
  * Usage: node scripts/check-diff-async.mjs
  * In CI: runs against the current PR diff (files changed vs base branch)
@@ -8,7 +8,7 @@
 import { execSync } from 'node:child_process';
 import { Project, SyntaxKind } from 'ts-morph';
 
-const CALLBACK_PARAM_PATTERN = /^(cb|callback|next|done|err)$/i;
+const CALLBACK_PARAM_PATTERN = /^(cb|callback|next|done)$/i;
 
 function getChangedJsFiles() {
     const base = process.env.GITHUB_BASE_REF
@@ -112,20 +112,6 @@ for (const sourceFile of project.getSourceFiles()) {
             });
         }
     }
-
-    const propertyAccesses = sourceFile.getDescendantsOfKind(SyntaxKind.PropertyAccessExpression);
-    for (const access of propertyAccesses) {
-        if (access.getName() !== 'then') continue;
-        const line = access.getStartLineNumber();
-        if (addedLines.has(line)) {
-            violations.push({
-                file: filePath,
-                line,
-                type: 'then-chain',
-                detail: 'use await instead of .then()',
-            });
-        }
-    }
 }
 
 if (violations.length === 0) {
@@ -137,6 +123,6 @@ console.error(`✗ Found ${violations.length} async/await violation(s) in the di
 for (const v of violations) {
     console.error(`  ${v.file}:${v.line} [${v.type}] ${v.detail}`);
 }
-console.error('\nNew code must use async/await instead of callbacks or .then() chains.');
+console.error('\nNew code must use async/await instead of callbacks.');
 console.error('See the async/await migration guide in CONTRIBUTING.md for help.');
 process.exit(1);
--- a/.github/scripts/check-diff-async.mjs
+++ b/.github/scripts/check-diff-async.mjs
@@ -1,0 +1,142 @@
+/**
+ * Check that all new/modified functions in the current git diff use async/await.
+ * Fails with exit code 1 if any additions introduce callback-style functions or .then() chains.
+ *
+ * Usage: node scripts/check-diff-async.mjs
+ * In CI: runs against the current PR diff (files changed vs base branch)
+ */
+import { execSync } from 'node:child_process';
+import { Project, SyntaxKind } from 'ts-morph';
+
+const CALLBACK_PARAM_PATTERN = /^(cb|callback|next|done|err)$/i;
+
+function getChangedJsFiles() {
+    const base = process.env.GITHUB_BASE_REF
+        ? `origin/${process.env.GITHUB_BASE_REF}`
+        : 'HEAD';
+    const output = execSync(`git diff --name-only --diff-filter=ACMR ${base} -- '*.js'`, {
+        encoding: 'utf8',
+    }).trim();
+
+    return output ? output.split('\n').filter(f => f.endsWith('.js')) : [];
+}
+
+/**
+ * Get added line numbers for a file in the current diff.
+ */
+function getAddedLineNumbers(filePath) {
+    const base = process.env.GITHUB_BASE_REF
+        ? `origin/${process.env.GITHUB_BASE_REF}`
+        : 'HEAD';
+    const diff = execSync(`git diff ${base} -- ${filePath}`, { encoding: 'utf8' });
+    const addedLines = new Set();
+    let currentLine = 0;
+
+    for (const line of diff.split('\n')) {
+        const hunkMatch = line.match(/^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@/);
+
+        if (hunkMatch) {
+            currentLine = parseInt(hunkMatch[1], 10) - 1;
+            continue;
+        }
+
+        if (line.startsWith('+') && !line.startsWith('+++')) {
+            currentLine++;
+            addedLines.add(currentLine);
+        } else if (!line.startsWith('-')) {
+            currentLine++;
+        }
+    }
+
+    return addedLines;
+}
+
+const changedFiles = getChangedJsFiles();
+if (changedFiles.length === 0) {
+    console.log('No changed JS files to check.');
+    process.exit(0);
+}
+
+console.log(`Checking ${changedFiles.length} changed JS file(s) for async/await compliance...\n`);
+
+const project = new Project({
+    compilerOptions: { allowJs: true, noEmit: true },
+    skipAddingFilesFromTsConfig: true,
+});
+
+const filesToCheck = changedFiles.filter(f =>
+    !f.startsWith('tests/') &&
+    !f.startsWith('node_modules/') &&
+    (
+        f.startsWith('lib/') ||
+        f.startsWith('bin/') ||
+        !f.includes('/')
+    )
+);
+if (filesToCheck.length === 0) {
+    console.log('No source JS files in diff (tests and node_modules excluded).');
+    process.exit(0);
+}
+
+project.addSourceFilesAtPaths(filesToCheck);
+
+const violations = [];
+
+for (const sourceFile of project.getSourceFiles()) {
+    const filePath = sourceFile.getFilePath().replace(process.cwd() + '/', '');
+    const addedLines = getAddedLineNumbers(filePath);
+
+    if (addedLines.size === 0) continue;
+
+    const functions = [
+        ...sourceFile.getDescendantsOfKind(SyntaxKind.FunctionDeclaration),
+        ...sourceFile.getDescendantsOfKind(SyntaxKind.FunctionExpression),
+        ...sourceFile.getDescendantsOfKind(SyntaxKind.ArrowFunction),
+        ...sourceFile.getDescendantsOfKind(SyntaxKind.MethodDeclaration),
+    ];
+
+    for (const fn of functions) {
+        if (fn.isAsync()) continue;
+
+        const startLine = fn.getStartLineNumber();
+        if (!addedLines.has(startLine)) continue;
+
+        const params = fn.getParameters();
+        const lastParam = params[params.length - 1];
+        if (lastParam && CALLBACK_PARAM_PATTERN.test(lastParam.getName())) {
+            violations.push({
+                file: filePath,
+                line: startLine,
+                type: 'callback',
+                detail: `function has callback parameter '${lastParam.getName()}'`,
+            });
+        }
+    }
+
+    const propertyAccesses = sourceFile.getDescendantsOfKind(SyntaxKind.PropertyAccessExpression);
+    for (const access of propertyAccesses) {
+        if (access.getName() !== 'then') continue;
+        const line = access.getStartLineNumber();
+        if (addedLines.has(line)) {
+            violations.push({
+                file: filePath,
+                line,
+                type: 'then-chain',
+                detail: 'use await instead of .then()',
+            });
+        }
+    }
+}
+
+if (violations.length === 0) {
+    console.log('✓ All new code in the diff uses async/await.');
+    process.exit(0);
+}
+
+console.error(`✗ Found ${violations.length} async/await violation(s) in the diff:\n`);
+for (const v of violations) {
+    console.error(`  ${v.file}:${v.line} [${v.type}] ${v.detail}`);
+}
+console.error('\nNew code must use async/await instead of callbacks or .then() chains.');
+console.error('See the async/await migration guide in CONTRIBUTING.md for help.');
+process.exit(1);
--- a/.github/scripts/check-diff-async.mjs
+++ b/.github/scripts/check-diff-async.mjs
@@ -5,7 +5,7 @@
  * Usage: node scripts/check-diff-async.mjs
  * In CI: runs against the current PR diff (files changed vs base branch)
  */
-import { execSync } from 'node:child_process';
+import { execFileSync } from 'node:child_process';
 import { Project, SyntaxKind } from 'ts-morph';
 
 const CALLBACK_PARAM_PATTERN = /^(cb|callback|next|done)$/i;
@@ -14,9 +14,14 @@ function getChangedJsFiles() {
     const base = process.env.GITHUB_BASE_REF
         ? `origin/${process.env.GITHUB_BASE_REF}`
         : 'HEAD';
-    const output = execSync(`git diff --name-only --diff-filter=ACMR ${base} -- '*.js'`, {
-        encoding: 'utf8',
-    }).trim();
+    const output = execFileSync('git', [
+        'diff',
+        '--name-only',
+        '--diff-filter=ACMR',
+        base,
+        '--',
+        '*.js',
+    ], { encoding: 'utf8' }).trim();
 
     return output ? output.split('\n').filter(f => f.endsWith('.js')) : [];
 }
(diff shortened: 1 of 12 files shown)


.github/scripts/check-diff-async.mjs
```diff
@@ -0,0 +1,142 @@
...
+                line: startLine,
+                type: 'callback',
+                detail: `function has callback parameter '${lastParam.getName()}'`,
+            });
+        }
+    }
+
+    const propertyAccesses = sourceFile.getDescendantsOfKind(SyntaxKind.PropertyAccessExpression);
+    for (const access of propertyAccesses) {
+        if (access.getName() !== 'then') continue;
```
DarkIsDude: We don't allow then but we ask them in the migration. What do you think ?

.github/scripts/count-async-functions.mjs
```diff
@@ -0,0 +1,89 @@
...
+console.log(`Total functions:      ${totalFunctions}`);
+console.log(`Async functions:      ${asyncFunctions} (${migrationPercent}%)`);
+console.log(`Callback functions:   ${callbackFunctions}`);
+console.log(`Remaining .then():    ${thenChains}`);
+console.log('');
+console.log(`Migration: ${asyncFunctions}/${totalFunctions} functions (${migrationPercent}%)`);
+
+if (process.env.GITHUB_STEP_SUMMARY) {
+    const { writeFileSync, appendFileSync } = await import('node:fs');
+    appendFileSync(process.env.GITHUB_STEP_SUMMARY, [
```
francoisferrand, line 87: do we really want a _summary_ in _every_ PR?
for monitoring the progress, this may be done in a separate job (eg. nightly on on dev branches) which builds a report giving the status on the "latest branch" (as a badge in the README, as a GitHub page, ....)

.github/workflows/tests.yaml
```diff
@@ -96,13 +96,29 @@ jobs:
     - name: Install python deps
       run: pip install flake8
     - name: Lint Javascript
-      run: yarn run --silent lint -- --max-warnings 0
+      run: yarn run --silent lint
```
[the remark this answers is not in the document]
francoisferrand: kind of agree with Claude here, we don't want to disable all warnings, but only the `prefer-await-to-than` and `callback-return`... but we want to get a report on these...

- could we instead run eslint multiple times? First without these checks but with strict warning; then only on these "migration" checks (and ignoring warnings).
- another option may be to create a custom eslint reporter, to generate a separate report AND filter out (=ignore) irrelevant warnings ? c.f. [1]
- maybe we could use a custom reporter to automatically handle the `check-diff-async` part (i.e. ignore defects not in diff).... and maybe (probably if I am being optimistic...) this is a feature of eslint or an existing plugin that we may leverage?

[1] https://stackoverflow.com/questions/66899779/configure-eslint-to-report-only-plugin-specific-errors

.github/scripts/check-diff-async.mjs
francoisferrand: do we actually need this extra "check"?
- I guess we would only use `then()` because we have a callback function... which should be reported already
- even after the migration, do you thing we may have some cases where it makes sense to use the raw `.then()` instead of `await` ?

.github/scripts/check-diff-async.mjs
francoisferrand, line 1: should these scripts not be in the `Guidelines` repo, so they can be reused?

(and if needed, it may include a github action as well)

francoisferrand, line 1: we should have tests for these scripts, to make it easier to test/debug them (even if they are tools).
(this will be easier to implement if they are in a kind of dedicated repo -like `Guidelines`- and simply _used_ here)

note: for testing against git/diff, maybe we could use https://github.com/kiegroup/mock-github, as we do for testing CI workflows

.github/scripts/check-diff-async.mjs
```diff
@@ -0,0 +1,142 @@
+/**
+ * Check that all new/modified functions in the current git diff use async/await.
+ * Fails with exit code 1 if any additions introduce callback-style functions or .then() chains.
```
francoisferrand: can't we use https://github.com/tj-actions/eslint-changed-files for this purpose?
seems quite similar...

or run eslint like this: `eslint --fix $(git diff --name-only HEAD | xargs)
` (c.f. https://stackoverflow.com/questions/54511168/enable-eslint-only-for-edited-files)

.github/scripts/count-async-functions.mjs
```diff
@@ -0,0 +1,89 @@
...
+
+const project = new Project({
+    compilerOptions: {
+        allowJs: true,
+        noEmit: true,
+    },
+    skipAddingFilesFromTsConfig: true,
+});
+
+project.addSourceFilesAtPaths([
```
francoisferrand: shouldn't this be some parameters of the script (or auto-discovered from package.json/...), to allow reusing this script?

.github/scripts/count-async-functions.mjs
```diff
@@ -0,0 +1,89 @@
...
+        const params = fn.getParameters();
+        const lastParam = params[params.length - 1];
+        if (lastParam && CALLBACK_PARAM_PATTERN.test(lastParam.getName())) {
+            callbackFunctions++;
+        }
+    }
+
+    const propertyAccesses = sourceFile.getDescendantsOfKind(SyntaxKind.PropertyAccessExpression);
+    for (const access of propertyAccesses) {
+        if (access.getName() === 'then') {
```
francoisferrand, line 64: same as above: not sure it makes sense to try to "eradicate" `then()` (but we can reporting them 😁)

.github/scripts/count-async-functions.mjs
```diff
@@ -0,0 +1,89 @@
...
+            thenChains++;
+        }
+    }
+}
+
+const migrationPercent = totalFunctions > 0
```
francoisferrand: this is not correct : not all functions will be (ever) be async.

the migration progress is really `1 - callbackFunctions / (callbackFunctions_before_migration)`....

as a "kind of proxy", we may compute `callbackFunctions / (asyncFunctions + callbackFunctions)` ; but not sure it is that useful...

francoisferrand: (completion is really just when `callbackFunctions == 0`)

francoisferrand (changes requested):

.github/scripts/check-diff-async.mjs
DarkIsDude: After the migration we should not have then 🙏. I guess we can remove this check indeed and find something on how to monitor it if it's needed later

.github/scripts/check-diff-async.mjs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+/**
+ * Get added line numbers for a file in the current diff.
+ */
+function getAddedLineNumbers(filePath) {
+    const base = process.env.GITHUB_BASE_REF
+        ? `origin/${process.env.GITHUB_BASE_REF}`
+        : 'HEAD';
+    const diff = execSync(`git diff ${base} -- ${filePath}`, { encoding: 'utf8' });
```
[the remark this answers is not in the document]
DarkIsDude: That's file as it's only use in the CI and so in a secure context.

.github/scripts/count-async-functions.mjs
DarkIsDude: IMO we can keep it simple as compute the report in each PR. Not a big deal, it's only some seconds so no finance impact.

.github/scripts/check-diff-async.mjs
DarkIsDude: Right now as it's the first repo, I would keep them here. When will be migrated to an another repo, can be. Maybe with a claude skills also 🤷

.github/workflows/tests.yaml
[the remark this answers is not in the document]
DarkIsDude: Thanks for the spot. I changed the CI to disable some rules at runtime

.github/scripts/check-diff-async.mjs
DarkIsDude: I don't want to run it only on the file but also on the line. We have pretty big file with big function and changing all of them can be a mess 😬. What do you think @francoisferrand ?

.github/scripts/count-async-functions.mjs
```diff
@@ -0,0 +1,89 @@
...
+    skipAddingFilesFromTsConfig: true,
+});
+
+project.addSourceFilesAtPaths([
+    'lib/**/*.js',
+    'index.js',
+    'dataserver.js',
+    'mdserver.js',
+    'managementAgent.js',
+    'bin/**/*.js'
```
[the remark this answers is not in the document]
DarkIsDude: It's not needed right now.

.github/scripts/count-async-functions.mjs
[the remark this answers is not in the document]
DarkIsDude: .then are pretty rare and don't exists in the code. Let's start simple

.github/scripts/count-async-functions.mjs
DarkIsDude: Changed. The goal here is just to have an indicator about progress (maybe to be able to follow that over time) not to have something really accurate. Reaching 100% is not the goal, the real goal is the 0... I would like to keep the calcul simple

.github/scripts/count-async-functions.mjs
francoisferrand: I was not thinking about build time or finance, but really about relevance:
- a summary is "inside" the build, so people will not go look at it - not will it show them how much progress they made (in their PR)
- for "monitoring", i.e. if someone want to (periodically...) check the status : not sure it is very practical to just go and look inside the last finished build

the summary does not hurt I agree, but I'd rather have something more dedicated to their purpose, like
- a report in a "static" place : GitHub page, badge in the Readme, periodic reporting job...
- a comment on the PR indicating how this specific PR affected the async migration (like _"Great, you removed 10 async calls! Migration is now 53% complete, keep pushing!"_)

.github/scripts/check-diff-async.mjs
francoisferrand: > Right now as it's the first repo, I would keep them here

why do you say that?
Do we expect to have to iterate on these scripts, and need "closeness" with the code? Or do you think the script are kind of tailored to cloudserver, and would not work as such in other code base?

.github/scripts/check-diff-async.mjs
```diff
@@ -0,0 +1,133 @@
...
+        ...sourceFile.getDescendantsOfKind(SyntaxKind.FunctionExpression),
+        ...sourceFile.getDescendantsOfKind(SyntaxKind.ArrowFunction),
+        ...sourceFile.getDescendantsOfKind(SyntaxKind.MethodDeclaration),
+    ];
+
+    for (const fn of functions) {
+        if (fn.isAsync()) continue;
+
+        const startLine = fn.getStartLineNumber();
+        if (!addedLines.has(startLine)) continue;
```
francoisferrand, line 107: is this the right filter?
This would consider only functions whose "header" is changed : and skip if only an update is made inside the existing function's body...

should we not search that there are no addedLines between `fn.getStartLineNumber()` and `fn.getEndLineNumber()` (or however it is called)

DarkIsDude: That's a good comment that I didn't notice. Anyway I suggest to keep current behaviour. We can start "slowly" and not too aggressive. This will allow dev to be familiar with async/await and introduce not too much change at the start. I'll add a comment in the guideline to setup your suggestion in some months ?

DarkIsDude: https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/2025-10-30+-+Async+Await+migration#Next-steps

.github/scripts/count-async-functions.mjs
DarkIsDude: That's a good point 🙏. Let's start simple again and I'll add your comment. We don't know how this will be use and what is needed from the team/dev. So I suggest to not waste time on that and iterate when we have feedback ?

.github/scripts/check-diff-async.mjs
delthas: I understand the iterative approach, but I think it would be really too limited to only require await migrations on signature changes -- they change very rarely, not sure signature only is significant enough for a first iteration.

delthas (approved):
LGTM on the whole (agree we can iterate later)

.github/scripts/count-async-functions.mjs
```diff
@@ -0,0 +1,100 @@
...
+
+const migrationPercent = (asyncFunctions + callbackFunctions) > 0
+    ? ((asyncFunctions / (asyncFunctions + callbackFunctions)) * 100).toFixed(1)
+    : '0.0';
+
+console.log('=== Async/Await Migration Progress ===');
+console.log(`Total functions:      ${totalFunctions}`);
+console.log(`Async functions:      ${asyncFunctions} (${asyncFunctionPercent}%)`);
+console.log(`Callback functions:   ${callbackFunctions}`);
+console.log(`Remaining .then():    ${thenChains}`);
```
francoisferrand, line 82: should we also count "dual-style" functions (async + cb)?

.github/scripts/count-async-functions.mjs
francoisferrand: > We don't know how this will be use and what is needed from the team/dev

That is not really true : this work was started because you wanted to solve a problem, i.e. _gaining visibility on progress of the migration_ I think. 
And my point here is really simple : does the summary approach give you that visibility?

.github/workflows/tests.yaml
```diff
@@ -95,14 +97,31 @@ jobs:
         cache: pip
     - name: Install python deps
       run: pip install flake8
-    - name: Lint Javascript
-      run: yarn run --silent lint -- --max-warnings 0
+    - name: Lint Javascript (strict, excluding async migration rules)
+      run: |
+        yarn run --silent lint -- --max-warnings 0 --rule "promise/prefer-await-to-then: off" --rule "n/callback-return: off"
```
francoisferrand, line 102: should these rule exceptions not be added in the package.json instead?
so that developer can run yarn lint locally as well...

or do we expect developers to run lint with this rule, and be smart about ignoring them?

eslint.config.mjs
```diff
@@ -67,5 +71,7 @@ export default [...compat.extends('@scality/scality'), {
         "quote-props": "off",
         "mocha/no-exclusive-tests": "error",
         "no-redeclare": ["error", { "builtinGlobals": false }],
+        "promise/prefer-await-to-then": "warn",
+        "n/callback-return": "warn",
```
francoisferrand, line 75: - kind of redundant with the dedicated linter you added
- in CI : disabled
- in dev : since we migrate "function by function", there will probably be lots of warnings there... which may make the lint practically useless (too hard to use, too many warnings to ignore) for devs?

i.e. should we really add these now?

.github/scripts/count-async-functions.mjs
DarkIsDude: So I'm now using CodeQL to extract same. We'll be able to follow each "detection" and the one closed. To be honest I thought also to just remove it. I'm not sure it's useful for me as the CI now enforce that so from my point of view it's enough. Moreover will change the detection to have something more and more aggresive. What do you think ?

.github/scripts/check-diff-async.mjs
DarkIsDude: Yes, I think we'll need to iterate over them and I don't know how the migration will be adopted by developer. Also I don't know when/if we'll add that on other repo. That's why I want to first put it here and when we'll decide in 4L to move that on a new repo, migrate this script at this time as it's easier to iterate it on only one repo.

.github/scripts/check-diff-async.mjs
DarkIsDude: Thanks for your feedback both. 

image

image

image

Seems that we'll start with only function declaration and iterate when the team will be familiar with it 🙏

.github/scripts/count-async-functions.mjs
DarkIsDude: I don't think it's useful right now ?

.github/workflows/tests.yaml
DarkIsDude: That's a good question. I would answer with smart dev to make sure we don't miss anything in the lint... But I can define a new script if you prefer and think it's a bad idea.

eslint.config.mjs
DarkIsDude: I agree that `lots of warnings` but I think it's important to have a visual reminder...

.github/scripts/count-async-functions.mjs
```diff
@@ -0,0 +1,120 @@
...
+        const params = fn.getParameters();
+        const lastParam = params[params.length - 1];
+        if (lastParam && CALLBACK_PARAM_PATTERN.test(lastParam.getName())) {
+            callbackFunctions++;
+        }
+    }
+
+    const propertyAccesses = sourceFile.getDescendantsOfKind(SyntaxKind.PropertyAccessExpression);
+    for (const access of propertyAccesses) {
+        if (access.getName() === 'then') {
```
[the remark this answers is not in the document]
DarkIsDude: Adding a `then` property on our codebase it's not something that we are doing and it's pretty fine to use that

.github/workflows/tests.yaml
```diff
@@ -95,14 +97,32 @@ jobs:
...
```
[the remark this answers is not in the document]
DarkIsDude: It's intented as we use then during the migration process but he should eventually reach the 0 at the end of the migration

.github/scripts/count-async-functions.mjs
francoisferrand: we can keep it, that way we can see if it is useful, if the information is useful but not at the right place, if codeql is enough/appropriate, and if they provide the same "measure" (double blind validation!)

.github/workflows/tests.yaml
francoisferrand: being smart about ignoring 1 warning is a thing, but may be difficult if there are many... which I guess would be the case here?
→ it would lead to warning fatigue, not running the tool and/or missing other (mandatory!) warnings

Maybe could be the other way, i.e. default to off in eslint config, and we can anyway enable them explicitely with `--rule "promise/prefer-await-to-then: on"` (or an extra target "super lint") ?

francoisferrand (approved):

eslint.config.mjs
francoisferrand: how "lots" ? if it makes the tool unusable, nobody will use it....and there won't be a visual either :/

DarkIsDude: Depends about the file you are but from my tests and check (eslint and count number of warning per file), it's not that much. Let's see how this will evolve

PR #6091, opened 2026-03-03
CLDSRV-851: Read TLS info from proxy headers in access logs

dvasilas:
When cloudserver is behind a TLS-terminating proxy (e.g. nginx), `cipherSuite` and `tlsVersion` are `null` because the socket is not encrypted.
Fall back to `x-ssl-cipher` and `x-ssl-protocol` headers set by the proxy.


leif-scality (approved):

nicolas2bert (approved):

dvasilas:
:robot:

BourgoisMickael (approved):

PR #6094, opened 2026-03-06
CLDSRV-863: add checksums to PutObject and UploadPart

leif-scality:


nicolas2bert (commented):
## Review: CLDSRV-863 — Add checksums to PutObject and UploadPart

**PR Status:** This looks like a work-in-progress commit (`tmp2`) that should not be merged in its current state. There are several correctness bugs, missing error handling, and debug artifacts throughout.

---

### Critical — Bugs that will cause crashes or data corruption

**1. Double-callback / crash in `prepareStream2` (`prepareStream.js:69`)**

In the `STREAMING-AWS4-HMAC-SHA256-PAYLOAD` branch, when `streamingV4Params` is not an object, the function calls `errCb()` via the error log path but then `return null`. The caller in `storeObject.js:72` checks for null and calls `cbOnce(errors.InvalidArgument)` — so the callback fires twice. Either call `errCb` and return (no null), or return null without calling `errCb` and let the caller handle it. Pick one.

**2. Missing error handler on `ChecksumTransform` (all branches in `prepareStream.js`)**

`TrailingChecksumTransform` correctly gets `.on('error', errCb)` at line 87, but `ChecksumTransform` never gets an error handler in any of the four branches. If `_flush()` fails (e.g., CRC digest throws), the error is unhandled and will crash the process. Add `checksumedStream.on('error', errCb)` in every branch.

**3. Race condition on `validateChecksum()` (`storeObject.js:92`)**

`validateChecksum()` reads `this.digest`, which is set in `_flush()`. The `data.put` callback fires when the backend finishes consuming data, but `_flush()` runs asynchronously when the stream ends. If `_flush()` hasn't resolved its promise yet, `this.digest` is `undefined` and:
- Non-trailer path: `undefined !== expectedDigest` silently passes if `expectedDigest` is also `undefined`
- Trailer path: `undefined !== trailerChecksumValue` produces a wrong mismatch error

The fix is to either validate inside `_flush()` and emit an error on the stream, or listen for the `'finish'` event before calling `validateChecksum()`.

**4. Data leak on checksum failure (`storeObject.js:95-98`)**

When `validateChecksum()` fails, the code returns `cbOnce(errors.BadDigest)` but the data has already been stored by `data.put`. Compare with `checkHashMatchMD5` just below — it calls `data.batchDelete` to clean up stored data on MD5 mismatch. The checksum failure path should do the same, otherwise failed uploads leak storage.

---

### Significant — Should fix before merge

**5. Unnecessary default CRC64 hashing (`validateChecksums.js:218`)**

When no checksum header and no trailer are present, `getChecksumDataFromHeaders` returns `{ algorithm: 'crc64nvme', isTrailer: false, expected: undefined }`. This causes the entire request body to be hashed with CRC64 even when there's nothing to validate against, wasting CPU on every non-checksummed request. Should return a sentinel (e.g., `{ algorithm: null }`) and skip hashing.

**6. Raw error string (`validateChecksums.js:200`)**

`{ error: 'invalid x-amz-trailer' }` uses a raw string instead of a `ChecksumError` enum value. `arsenalErrorFromChecksumError` won't match this in its switch statement and will fall through to undefined behavior.

**7. Extra argument in `_transform` callback (`ChecksumTransform.js:71`)**

```js
callback(null, input, encoding);
```
The `Transform._transform` callback signature is `callback(error, chunk)`. The third `encoding` argument is not part of the API. While Node currently ignores it, it's incorrect. Use `callback(null, input)`.

**8. `veeam/utils.js:38` still uses old `prepareStream`**

The Veeam utils path was not migrated to `prepareStream2`. There's a bare `// TODO` but no tracking issue. This path will miss checksum validation.

---

### Cleanup — Must address before merge

- **Commit message is `tmp2`** — rewrite with a proper description
- **Commented-out `console.log`** in `createAndStoreObject.js:215,218,220,255`, `storeObject.js:90-91,94`, `prepareStream.js:85,113`, `ChecksumTransform.js:69` — remove all
- **Commented-out imports/code** in `storeObject.js:4,6,63-67` — remove dead code
- **Double semicolon** at `validateChecksums.js:205`
- **Dev note comments** — `V4Transform.js:9` ("Do we use this one or vaults?"), `validateChecksumHeaders.js:5` ("FIXME: merge this"), `validateChecksumHeaders.js:36` (inline question) — resolve or remove
- **Unresolved FIXMEs** in `ChecksumTransform.js:28` (missing trailer handling), `prepareStream.js:57,69` (error handling inconsistency)
- **No tests** — This is a significant behavior change (new stream transform, new validation logic) with zero test additions

---

### Design Observations

The overall approach — adding a `ChecksumTransform` that computes checksums during streaming and validates after storage — is sound. The `getChecksumDataFromHeaders` extraction is a good refactor. However, the validation-after-storage pattern has an inherent problem: you must clean up stored data on failure. Consider validating in `_flush()` and emitting a stream error before data reaches the backend, which would avoid the cleanup problem entirely.

The `TrailingChecksumTransform` changes to parse trailers after the zero-length chunk are well-structured, though the 1024-byte limit on trailer lines should be documented or tested.

Review by Claude Code

PR #6095, opened 2026-03-09
CLDSRV-853: Rate limit config for accounts

tmacro:


lib/api/apiUtils/rateLimit/config.js
```diff
@@ -125,6 +151,91 @@ const { calculateInterval } = require('./gcra');
...
+ * @property {object} defaultConfig - Default config applied if no resource specific configuration is found.
+ * @property {number} configCacheTTL - Number of milliseconds to cache per resource configs
+ * @property {number} defaultBurstCapacity - Default used if resource does not specify a burst capacity.
+*/
+
+const rateLimitClassConfigSchema = Joi.object({
+    defaultConfig: Joi.object({
+        requestsPerSecond: Joi.object({
+            limit: Joi.number().integer().min(0).required(),
+            burstCapacity: Joi.number().positive(),
```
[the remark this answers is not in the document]
tmacro: Intended behavior change

lib/api/apiUtils/rateLimit/config.js
```diff
@@ -48,6 +48,32 @@ const { calculateInterval } = require('./gcra');
  *       "defaultBurstCapacity": 1
  *     },
  *
+ *     // Optional: Account-level rate limiting configuration
+ *     "account": {
+ *       // Optional: Global default rate limit for all accounts
+ *       // Can be overridden per-bucket via PutBucketRateLimit API
```
anurag4DSB: Unless I remmeber incorrectly, only limit can be over written by PutBucketRateLimit.

If thats still the case, could we clarify that by mentioning that in the comment and put the comment just above "limit"

anurag4DSB (approved):

lib/api/apiUtils/rateLimit/config.js
```diff
@@ -51,7 +51,7 @@ const { calculateInterval } = require('./gcra');
  *     // Optional: Account-level rate limiting configuration
  *     "account": {
  *       // Optional: Global default rate limit for all accounts
- *       // Can be overridden per-bucket via PutBucketRateLimit API
+ *       // Can be overridden per-account via UpdateAccountLimits API
```
anurag4DSB: Same for update account limits API

lib/api/apiUtils/rateLimit/config.js
```diff
@@ -125,6 +151,91 @@ const { calculateInterval } = require('./gcra');
...
 
+/**
+ * RateLimitClassConfig
+ * @typedef {Object} RateLimitClassConfig
+ * @property {object} defaultConfig - Default config applied if no resource specific configuration is found.
+ * @property {number} configCacheTTL - Number of milliseconds to cache per resource configs
+ * @property {number} defaultBurstCapacity - Default used if resource does not specify a burst capacity.
+*/
+
+const rateLimitClassConfigSchema = Joi.object({
```
BourgoisMickael, line 162: should you validate some max value to avoid any misconfig breaking the system ?

BourgoisMickael (approved):

lib/api/apiUtils/rateLimit/config.js
tmacro: I think it's fine. Joi already ensure the number is "safe" from a javascript perspective.

lib/api/apiUtils/rateLimit/config.js
```diff
@@ -125,6 +151,91 @@ const { calculateInterval } = require('./gcra');
...
+const rateLimitConfigSchema = Joi.object({
+    enabled: Joi.boolean().default(false),
+    serviceUserArn: Joi.string().required(),
+    nodes: Joi.number().integer().positive().default(1),
+    tokenBucketBufferSize: Joi.number().integer().positive().default(50),
+    tokenBucketRefillThreshold: Joi.number().integer().positive().default(20),
+    bucket: rateLimitClassConfigSchema,
+    account: rateLimitClassConfigSchema,
+    error: Joi.object({
+        statusCode: Joi.number().integer().min(400).max(599).default(503),
```
anurag4DSB, line 182: afterthought: Shouldn't the default be 429?

tmacro: My thought was that `503 Slowdown` matches AWS behavior so we keep that as the default, but then override it at the federation level when deploying to avoid the load balancer issue.

anurag4DSB: Okay fair, I was only thinking about nginx blacklisting or local monitoring deployments.
I am okay :)

tmacro:
I'm closing this PR in favor of an approach using joi to parse the config.

PR #4155, opened 2022-01-10
improvement/CLDSRV-91 Bump Arsenal to 8.1.18

KillianG:
Bump arsenal to 8.1.18 to bring fix for ARTESCA-3109

PR #6097, opened 2026-03-09
Improvement/CLDSRV-825 drop ceph support

benzekrimaha:
Issue: CLDSRV-825


francoisferrand:
this PR seems to remove only the _tests_ : while the goal of the ticket was to remove supporting code (corner cases) as well:

> remove corner cases to support it (in production code!)

AFAIR, there were at least a few of these (which is consistent with the tests we had), can you double check?

lib/utilities/reportHandler.js
```diff
@@ -72,7 +70,6 @@ function getCapabilities(cfg = config) {
         caps.locationTypeDigitalOcean &&= cfg.supportedLocationTypes.has('location-do-spaces-v1');
         caps.locationTypeSproxyd &&= cfg.supportedLocationTypes.has('location-scality-sproxyd-v1');
         caps.locationTypeNFS &&= cfg.supportedLocationTypes.has('location-nfs-mount-v1');
-        caps.locationTypeCephRadosGW &&= cfg.supportedLocationTypes.has('location-ceph-radosgw-s3-v1');
```
francoisferrand, line 75: This is used to ensure consistency, Cloudserver does not have the "source of truth" : it comes from the ZenkoVersion CR (through ZKOP)
→ should not be kept
→ either we should keep this (to ensure consistency, even if ZenkoVersion indicates the version is supported), or enforce the new lack-of-support of cloudserver (i.e. remove 'location-ceph-radosgw-s3-v1' from the reported location types and unset locationTypeCephRadosGW)

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAwsVersioning.js
```diff
@@ -206,7 +205,7 @@ describeSkipIfNotMultiple('AWS backend delete object w. versioning: ' +
                     key, versionId: 'null', resultType: deleteVersion },
                     err => next(err, awsVerId)),
                 (awsVerId, next) => {
-                    const wanted = isCEPH ? 'NoSuchKey' : 'NoSuchVersion';
+                    const wanted = 'NoSuchVersion';
```
francoisferrand: should be inlined, no reason to keep the variable now

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAzure.js
```diff
@@ -30,7 +29,7 @@ const nonExistingId = process.env.AWS_ON_AIR ?
     'MhhyTHhmZ4cxSi4Y9SMe5P7UJAz7HLJ9' :
     '3939393939393939393936493939393939393939756e6437';
 
-describeSkipIfNotMultipleOrCeph('Multiple backend delete object from Azure',
+describe.skip('Multiple backend delete object from Azure',
```
francoisferrand: should not be skipped with multi-backend, it seems definition of `describeSkipIfNotMultipleOrCeph` was broken at some point → describeSkipIfNotMultiple

(if it does not work, create a followup ticket and reference it here)

francoisferrand: This applies to every place where `describeSkipIfNotMultipleOrCeph` was used, please fix them all to use `describeSkipIfNotMultiple` instead

tests/functional/aws-node-sdk/test/multipleBackend/objectPutCopyPart/objectPutCopyPartAzure.js
```diff
@@ -126,7 +124,7 @@ function assertCopyPart(infos, cb) {
     ], cb);
 }
 
-describeSkipIfNotMultipleOrCeph('Put Copy Part to AZURE', function describeF() {
+describe.skip('Put Copy Part to AZURE', function describeF() {
```
francoisferrand: same, should be be run with multiple data backend (even if it uses another definition of `describeSkipIfNotMultipleOrCeph` which you removed from line 23)

tests/functional/aws-node-sdk/test/multipleBackend/put/putGcp.js
```diff
@@ -235,7 +235,7 @@ describeFn() {
     });
 });
 
-describeSkipIfNotMultipleOrCeph('MultipleBackend put object' +
+describe.skip('MultipleBackend put object' +
                                 'based on bucket location', () => {
```
francoisferrand:
```suggestion
describe.skip('MultipleBackend put object based on bucket location', () => {
```

tests/functional/aws-node-sdk/test/object/websiteGetWithACL.js
```diff
@@ -26,14 +26,12 @@ const aclEquivalent = {
 };
 
 const aclTests = [
-    // CEPH: test_website_private_bucket_list_private_index_blockederrordoc
```
francoisferrand, line 29: this is weird ; I don't understand what these comments stand for, are they really related to Ceph?

francoisferrand (changes requested):
- as already said, need to remove specific tweaks from the actual code (api handlers...)
- some tests are always skipped (not sure why) while they should run for multiple backend

tests/functional/aws-node-sdk/test/multipleBackend/delete/deleteAzure.js
francoisferrand: note: this was introduced in https://github.com/scality/cloudserver/pull/4155/changes/c3385a4225bd28ced47511646f7f6ac202a70a4b, with not much explanation either in commit, PR or [ticket](https://scality.atlassian.net/browse/CLDSRV-93)

tests/functional/aws-node-sdk/test/object/websiteGetWithACL.js
benzekrimaha: Those comments weren’t about Ceph-specific behavior in Cloudserver IMO. They were legacy cross-references to Ceph’s S3 test case names (e.g. test_website_private_bucket_list_private_index_blockederrordoc), added when aligning our website ACL tests with theirs. 
The cases themselves are generic website endpoint + ACL tests (bucket/index/error doc visibility). The “CEPH:” prefix was only a test-name reference and was confusing, so I've removed them.

DarkIsDude (changes requested):
Can you check the CI 🙏 ?

lib/utilities/reportHandler.js
DarkIsDude: I'm not sure to understand your point @francoisferrand

DarkIsDude (commented):
I found some CEPH reference in the product doc that we should also clean. The CI is still unstable

DarkIsDude (approved):

lib/Config.js
```diff
@@ -1880,7 +1874,9 @@ class Config extends EventEmitter {
 
         if (config.capabilities) {
             if (config.capabilities.locationTypes) {
-                this.supportedLocationTypes = new Set(config.capabilities.locationTypes);
+                this.supportedLocationTypes = new Set(
```
francoisferrand: why?
this should not be needed, unless there is an issue in tests.... so the tests should be updated instead: we should not adapt production code to tests, but tests to match how it happens in production

tests/multipleBackend/multipartUpload.js
```diff
@@ -531,7 +523,7 @@ describe('Multipart Upload API with AWS Backend', function mpuTestSuite() {
                 })).then(() => {
                     assert.fail('Expected an error listing parts of aborted MPU');
                 }).catch(err => {
-                    const wantedError = isCEPH ? 'NoSuchKey' : 'NoSuchUpload';
+                    const wantedError = 'NoSuchUpload';
```
francoisferrand: variable probably not needed here

tests/unit/utils/reportHandler.js
```diff
@@ -38,9 +37,12 @@ describe('reportHandler.getCapabilities', () => {
                 secureChannelOptimizedPath: true,
                 s3cIngestLocation: true,
                 nfsIngestLocation: false,
-                cephIngestLocation: false,
                 awsIngestLocation: false,
             });
+            assert.strictEqual(
+                Object.prototype.hasOwnProperty.call(caps, 'locationTypeCephRadosGW'),
```
francoisferrand: useless: redundant with `assert.deepStrictEqual` above, can never fail...

tests/unit/utils/reportHandler.js
```diff
@@ -63,6 +65,31 @@ describe('reportHandler.getCapabilities', () => {
                 locationTypeS3Custom: false,
                 customCapability: 'test-value',
             });
+            assert.strictEqual(
+                Object.prototype.hasOwnProperty.call(caps, 'locationTypeCephRadosGW'),
```
francoisferrand: useless: redundant with assert.deepStrictEqual above, can never fail...

francoisferrand (commented):
a few small things, but more importantly it seems to me this cleanup is not complete: this is modifying only the _tests_, but we keep the code handling specific corner cases for cepth (which may be either/both in cloudserver and arsenal).

this code _MUST_ be cleaned as well, otherwise we just continue maintaining the same code, just with less coverage

benzekrimaha:
@francoisferrand  Thanks for the careful review — I agree with your point that removing tests without removing related code would leave us with untested legacy behavior.
I re-audited this PR specifically with that objective:
I listed CEPH-conditioned test removals (isCEPH / itSkipCeph / describeSkipIfCeph / CEPH-only expected error variants).
I mapped those removed tests to the corresponding runtime paths.
I removed the CEPH-specific runtime handling that was still present.
What was actually CEPH-specific in CloudServer:
=> Capability-key cleanup in reportHandler getCapabilities() (locationTypeCephRadosGW / cephIngestLocation) was the remaining CEPH-specific logic.
This has now been removed, and the matching unit test coverage was removed accordingly.
**_What remains and why:_**
The remaining behaviors touched by removed CEPH-conditioned tests are generic S3/MPU API semantics, not CEPH branches. Examples:
Internal MPU metadata NoSuchKey mapped to public NoSuchUpload.
Multipart delete behavior controlled by legacyAwsBehavior.
Standard versioning/object-lock semantics (NoSuchKey / NoSuchVersion / MethodNotAllowed).
These paths are still covered by active non-CEPH tests.
**So this is not a tests-only cleanup anymore:**
CEPH-specific runtime handling in CloudServer has been removed.
Remaining logic is generic compatibility behavior with active coverage.
_If you want, I can also do the same explicit mapping pass on Arsenal in a follow-up to ensure both repos are aligned._

lib/utilities/reportHandler.js
```diff
@@ -74,9 +74,6 @@ function getCapabilities(cfg = config) {
         caps.locationTypeLocal &&= cfg.supportedLocationTypes.has('location-file-v1');
     }
 
-    delete caps.locationTypeCephRadosGW;
-    delete caps.cephIngestLocation;
```
francoisferrand, line 78: this was an earlier comment, we need to keep these!
cloudserver will not support ceph anymore, so we don't want to advertise that we still support it

francoisferrand:
> Capability-key cleanup in reportHandler getCapabilities() (locationTypeCephRadosGW / cephIngestLocation) was the remaining CEPH-specific logic.

this is not "ceph-specific" logic, this is logic which was introduced earlier on purpose to correctly report that we DO NOT support ceph anymore. So it must be kept.

> I can also do the same explicit mapping pass on Arsenal in a follow-up to ensure both repos are aligned.

yes, as discussed on Monday we need to check in Arsenal as well : where the actual specifics of each (data) backend are implemented

francoisferrand (approved):

PR #6108, opened 2026-03-13
impr(CLDSRV-852): Refactor WorkerTokenBucket

tmacro:
- Update RateLimitClient.grantToken() params
- Cleanup calculateInterval helper
- Refactor WorkerTokenBucket
  - Change constructor to accept resourceClass, resourceId, measure
  - Precalculate interval for reuse in constructor
  - Add hasCapacity() method
  - Add updateLimit() method to encasulate limit change logic
  - Remove unused mechanism to record request durations
  - Simplify refill logic

Note for revewers: S3 API unit tests are still expected to fail

### DIFF
--- a/lib/api/apiUtils/rateLimit/tokenBucket.js
+++ b/lib/api/apiUtils/rateLimit/tokenBucket.js
@@ -16,32 +16,44 @@ const { instance: redisClient } = require('./client');
 const { config } = require('../../../Config');
 const { calculateInterval } = require('./gcra');
 
-// Map of bucket name -> WorkerTokenBucket instance
+// Map of `${resourceClass}:${resourceId}:${measure}` -> WorkerTokenBucket instance
 const tokenBuckets = new Map();
 
 /**
- * Per-bucket token bucket for a single worker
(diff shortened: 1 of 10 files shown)


BourgoisMickael (approved):

lib/api/apiUtils/rateLimit/tokenBucket.js
```diff
@@ -16,32 +16,44 @@ const { instance: redisClient } = require('./client');
 const { config } = require('../../../Config');
 const { calculateInterval } = require('./gcra');
 
-// Map of bucket name -> WorkerTokenBucket instance
+// Map of `${resourceClass}:${resourceId}:${measure}` -> WorkerTokenBucket instance
 const tokenBuckets = new Map();
 
 /**
  * Per-bucket token bucket for a single worker
```
anurag4DSB: The class JSDoc still says "Per-bucket token bucket for a single worker". The class now handles any resource class (bucket, account, etc.) and the worker framing was removed from the architecture (rate limiting is now per-node). Does the comment still match the class's role?

anurag4DSB (approved):

PR #6112, opened 2026-03-16
CLDSRV-863: Checksums for PutObject and UploadPart

leif-scality:


lib/auth/streamingV4/ChecksumTransform.js
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    _flush(callback) {
+        Promise.resolve(this.algo.digestFromHash(this.hash))
+            .then(digest => { this.digest = digest; })
+            .then(() => callback(), err => {
```
BourgoisMickael: prefer using `catch` for error handling

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -30,9 +35,18 @@ class TrailingChecksumTransform extends Transform {
...
      */
     _flush(callback) {
-        if (!this.streamClosed) {
+        if (!this.streamClosed && this.readingTrailer && this.trailerBuffer.length === 0) {
+            // Nothing came after "0\r\n", don't fail.
+            // If the x-amz-trailer header was present then the trailer is required and ChecksumTransform will fail.
+            return callback();
+        } else if (!this.streamClosed && this.readingTrailer && this.trailerBuffer.length !== 0) {
+            this.log.error('stream ended without trailer "\r\n"');
+            return callback(errors.IncompleteBody.customizeDescription(
```
BourgoisMickael: you could use `errorInstances` when using `customizeDescription` to avoid instantiating twice an error

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -453,3 +459,274 @@ describe('validateMethodChecksumNoChunking', () => {
...
+    // Valid-format digests (correct length and base64, content not verified by getChecksumDataFromHeaders)
+    const validDigests = {
+        crc32: 'AAAAAA==',     // 8 chars
+        crc32c: 'AAAAAA==',     // 8 chars
+        sha1: 'AAAAAAAAAAAAAAAAAAAAAAAAAAA=', // 28 chars
+        sha256: 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=', // 44 chars
+        crc64nvme: 'AAAAAAAAAAA=', // 12 chars
+    };
+
+    it('no headers: returns crc64nvme with isTrailer=false and expected=undefined', () => {
```
BourgoisMickael: prefer starting tests with `should`

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -453,3 +459,274 @@ describe('validateMethodChecksumNoChunking', () => {
...
+        const result = arsenalErrorFromChecksumError({ error: ChecksumError.MissingChecksum, details: null });
+        assert.strictEqual(result, null);
+    });
+
+    it('XAmzMismatch (crc32): returns BadDigest mentioning CRC32', () => {
+        const result = arsenalErrorFromChecksumError({
+            error: ChecksumError.XAmzMismatch,
+            details: { algorithm: 'crc32', calculated: 'a', expected: 'b' },
+        });
+        assert(result.is.BadDigest);
```
BourgoisMickael: I don't like that `is` helper is tests because if test fails you compare booleans (`false != true`) instead of names (`InternalError != BadDigest`), so you don't have much details.

Can we compre on error name with strings ?

lib/auth/streamingV4/ChecksumTransform.js
leif-scality: claude complains about double calling callback, once in the `then` and again in the `catch` if the first callback throws

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: done

leif-scality: done

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: done

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: done

lib/auth/streamingV4/ChecksumTransform.js
BourgoisMickael: you could also use only 1 then

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
```diff
@@ -0,0 +1,610 @@
...
+const msgMalformedTrailer = 'The request contained trailing data that was not well-formed' +
+    ' or did not conform to our published schema.';
+const msgSdkMissingTrailer = 'x-amz-sdk-checksum-algorithm specified, but no corresponding' +
+    ' x-amz-checksum-* or x-amz-trailer headers were found.';
+
+// Create the 24 common protocol-scenario tests for a given URL factory.
+// urlFn() is called lazily at test runtime so that uploadId is available.
+function makeScenarioTests(urlFn) {
+    itSkipIfAWS(
+        'testS3PutNoChecksum: signed sha256 in x-amz-content-sha256, no x-amz-checksum header -> 200 OK',
```
BourgoisMickael: same here, prefer starting tests with `should`

tests/unit/api/apiUtils/object/prepareStream.js
```diff
@@ -0,0 +1,200 @@
...
+    accessKey: 'AKIAIOSFODNN7EXAMPLE',
+    signatureFromRequest: 'abc123',
+    region: 'us-east-1',
+    scopeDate: '20210101',
+    timestamp: '20210101T000000Z',
+    credentialScope: '20210101/us-east-1/s3/aws4_request',
+};
+
+describe('prepareStream return value shape', () => {
+    it('returns { error: null, stream: ChecksumTransform } for UNSIGNED-PAYLOAD', () => {
```
BourgoisMickael: same here prefer starting test with `should`

tests/unit/api/apiUtils/object/prepareStream.js
```diff
@@ -0,0 +1,200 @@
...
+        const request = makeRequest({
+            'x-amz-content-sha256': 'STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER',
+        });
+        const result = prepareStream(request, null, log, () => {});
+        assert.strictEqual(result.error.message, 'BadRequest');
+        assert.strictEqual(result.stream, null);
+    });
+});
+
+describe('prepareStream STREAMING-AWS4-HMAC-SHA256-PAYLOAD', () => {
```
BourgoisMickael: prefer one global `prepareStream` describe, and then other describes inside

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
leif-scality: done

tests/unit/api/apiUtils/object/prepareStream.js
leif-scality: done

tests/unit/api/apiUtils/object/prepareStream.js
leif-scality: done

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
leif-scality: done

tests/unit/api/apiUtils/object/prepareStream.js
leif-scality: done

tests/unit/api/apiUtils/object/prepareStream.js
leif-scality: done

lib/api/apiUtils/object/storeObject.js
```diff
@@ -58,31 +59,67 @@ function checkHashMatchMD5(stream, hashedStream, dataRetrievalInfo, log, cb) {
...
+
+            // ChecksumTransform._flush computes the digest asynchronously for
+            // some algorithms (e.g. crc64nvme). writableFinished is true once
+            // _flush has called its callback, guaranteeing this.digest is set.
+            // Stream piping ordering means this is virtually always true here,
+            // but we wait for 'finish' explicitly to be safe.
+            if (checksumedStream.stream.writableFinished) {
+                return doValidate();
+            }
+            checksumedStream.stream.once('finish', doValidate);
```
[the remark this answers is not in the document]
leif-scality: double-delete: not a risk — `finish` and `error` are mutually exclusive in the Node.js stream lifecycle, so both handlers can't fire for the same stream.

lib/api/apiUtils/object/storeObject.js
```diff
@@ -58,31 +59,67 @@ function checkHashMatchMD5(stream, hashedStream, dataRetrievalInfo, log, cb) {
...
+
+    const checksumedStream = prepareStream(stream, streamingV4Params, log, errCb);
+    if (checksumedStream.error) {
+        log.debug('dataStore failed to prepare stream', checksumedStream);
+        return process.nextTick(() => cbOnce(checksumedStream.error));
     }
-    const dataStream = stripTrailingChecksumStream(dataStreamTmp, log, cbOnce);
     return data.put(
-        cipherBundle, dataStream, size, objectContext, backendInfo, log,
+        cipherBundle, checksumedStream.stream, size, objectContext, backendInfo, log,
```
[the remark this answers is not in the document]
leif-scality: .headers: checked — no backend in `lib/data/` or Arsenal reads `.headers` off the stream passed to `data.put`.

lib/auth/streamingV4/ChecksumTransform.js
leif-scality: done

tests/unit/api/apiUtils/object/storeObject.js
```diff
@@ -0,0 +1,303 @@
...
+        });
+    }
+
+    // Stub data.batchDelete to succeed.
+    function batchDeleteSucceeds() {
+        batchDeleteStub.callsFake((keys, a, b, log2, cb) => cb(null));
+    }
+
+    describe('normal behaviour', () => {
+        it('calls data.put with the stream returned by prepareStream', done => {
```
BourgoisMickael: start with `should`

tests/unit/auth/ChecksumTransform.js
```diff
@@ -0,0 +1,182 @@
...
+
+    before(async () => {
+        expectedDigests = {};
+        for (const algo of algos) {
+            expectedDigests[algo] = await Promise.resolve(algorithms[algo].digest(testData));
+        }
+    });
+
+    for (const algo of algos) {
+        it(`[${algo}] passes data through unchanged`, async () => {
```
BourgoisMickael: start with `should`

tests/unit/auth/ChecksumTransform.js
```diff
@@ -0,0 +1,182 @@
...
+            assert.deepStrictEqual(err, errors.InternalError);
+            done();
+        });
+        stream.write(testData);
+        stream.end();
+        stream.resume();
+    });
+});
+
+describe('ChecksumTransform validateChecksum — non-trailer mode (isTrailer=false)', () => {
```
BourgoisMickael: prefer using one global describe and then others inside

tests/unit/api/apiUtils/object/storeObject.js
leif-scality: done

tests/unit/auth/ChecksumTransform.js
leif-scality: done

tests/unit/auth/ChecksumTransform.js
leif-scality: done

BourgoisMickael (approved):

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -30,9 +38,16 @@ class TrailingChecksumTransform extends Transform {
...
     _flush(callback) {
-        if (!this.streamClosed) {
+        if (!this.streamClosed && this.readingTrailer && this.trailerBuffer.length === 0) {
+            // Nothing came after "0\r\n", don't fail.
+            // If the x-amz-trailer header was present then the trailer is required and ChecksumTransform will fail.
+            return callback();
+        } else if (!this.streamClosed && this.readingTrailer && this.trailerBuffer.length !== 0) {
+            this.log.error('stream ended without trailer "\r\n"');
+            return callback(incompleteBodyError);
+        } else if (!this.streamClosed && !this.readingTrailer) {
```
jonathan-gramain: We could simplify the flow to avoid repeating conditions, e.g.:

```
if (streamClosed) {
    // return ok
}
if (!readingTrailer) {
    // log and return incomplete body error
}
if (trailerBuffer.length === 0) {
    // no trailer: ok
}
// log and return incomplete body error
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -20,6 +24,10 @@ class TrailingChecksumTransform extends Transform {
         this.bytesToDiscard = 0; // when trailing \r\n are present, we discard them but they can be in different chunks
         this.bytesToRead = 0; // when a chunk is advertised, the size is put here and we forward all bytes
         this.streamClosed = false;
```
jonathan-gramain: It's not related to this PR, but how about renaming this variable with a more specific name like `this.receivedLastChunk`? I got confused as it can easily be mistaken for a nodejs stream state. But feel free to ignore.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -66,6 +81,49 @@ class TrailingChecksumTransform extends Transform {
                 continue;
             }
 
+            // after the 0-size chunk, read the trailer line (e.g. "x-amz-checksum-crc32:YABb/g==")
+            if (this.readingTrailer) {
+                const combined = Buffer.concat([this.trailerBuffer, chunk]);
+                const lineBreakIndex = combined.indexOf('\r\n');
+                if (lineBreakIndex === -1) {
+                    if (combined.byteLength > 1024) {
```
jonathan-gramain: I believe 1024 is more than enough if all we want to support is a checksum trailer, but if we need to accept and ignore other headers, I believe we should allow larger values, I saw 8KB is the default for nginx so that may be a good value. Even if we only support and accept a checksum trailer, I think it doesn't hurt to stick with the default to avoid surprises in the future.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -66,6 +81,49 @@ class TrailingChecksumTransform extends Transform {
                 continue;
             }
 
+            // after the 0-size chunk, read the trailer line (e.g. "x-amz-checksum-crc32:YABb/g==")
+            if (this.readingTrailer) {
```
jonathan-gramain: This block is a bit complicated and doesn't seem to follow the HTTP spec which passes a `Trailers` header to know which trailers to expect at the end. Also we don't support multiple trailers. Maybe that's okay, just raising the point.

Wondering: would it be appropriate to rely on node.js http module trailer handling (https://nodejs.org/docs/latest-v22.x/api/http.html#messagetrailers) ? I don't know if those are automatically populated based on the "Trailers" header value, but if that works it would become much simpler and we would naturally be able to ignore extra unsupported trailers, or do whatever error handling we need without the nitty gritty details of parsing.

jonathan-gramain (commented):
Partial review (stopped at commit "emit trailer in TrailingChecksumTransform"), will continue tomorrow

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: The only trailer sent by the `aws-chunked` encoding is only the checksum, the limit should be `'x-amz-content-sha256:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA='.length`, but I added more because AWS accepts if the client sends whitespace before or after the `:`.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: The SDK's send `Content-Encoding: aws-chunked`, that can be different from HTTP `chunked` when using signatures, the SDK's also don't set the `Trailer` header. For this two reasons it is not possible to use the nodejs chunked parser. 

AWS also only expects the checksum trailer only, if a different trailer appears before x-amz-checksum- it returns MalformedTrailer and if it arrives after it is ignored.

I also tried forcing nodejs to use the chunked parser for aws-chunked but it is part of the private API so it is not really possible.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html#trailing-checksums

jonathan-gramain: OK, so be it, thanks for the details

lib/auth/streamingV4/trailingChecksumTransform.js
jonathan-gramain: It could be even cleaner with an FSM when state management becomes a bit more complex, by maintaining an explicit state variable and acting based on its value.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -66,6 +81,49 @@ class TrailingChecksumTransform extends Transform {
...
+                    this.readingTrailer = false;
+                    this.streamClosed = true;
+                    return callback();
+                }
+                let trailerLine = fullTrailer.toString();
+                // Some clients terminate the trailer with \n\r\n instead of
+                // just \r\n, producing a trailing \n in the parsed line.
+                if (trailerLine.endsWith('\n')) {
+                    trailerLine = trailerLine.slice(0, -1);
+                }
```
jonathan-gramain: Do we need this? I would think the `trim()` call on the value would suffice to get rid of the extra `\n`.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -66,6 +81,49 @@ class TrailingChecksumTransform extends Transform {
...
+                if (lineBreakIndex === -1) {
+                    if (combined.byteLength > 1024) {
+                        this.log.error('trailer line too long');
+                        return callback(errors.MalformedTrailerError);
+                    }
+                    // The trailer is not complete yet, continue.
+                    this.trailerBuffer = combined;
+                    return callback();
+                }
+                this.trailerBuffer = Buffer.alloc(0);
```
jonathan-gramain: Maybe it could be cleaner to set it to `null` as we shouldn't use this variable anymore at this point.

Even in addition to this, we could check if it is null in place of using an extra variable `readingTrailer`, and initialize it when it's time to read the trailer, although I'm not sure it would make things more understandable (just one less variable to manage).

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -141,6 +177,94 @@ async function validateXAmzChecksums(headers, body) {
...
+
+        return null;
+    };
+
+    const checksumHeaders = Object.keys(headers).filter(header => header.startsWith('x-amz-checksum-'));
+    const xAmzChecksumCnt = checksumHeaders.length;
+    if (xAmzChecksumCnt > 1) {
+        return { error: ChecksumError.MultipleChecksumTypes, details: { algorithms: checksumHeaders } };
+    }
+
```
jonathan-gramain, line 209: Nitpick suggestion for clarity:
- add `const checksumHeader = checksumHeaders[0];`
- thereafter, check if a header is present via `if (checksumHeader)` and use it directly, instead of checking the count and using `checksumHeaders[0]`

lib/auth/streamingV4/ChecksumTransform.js
```diff
@@ -0,0 +1,94 @@
...
+                this.digest = digest;
+                callback();
+            }, err => {
+                this.log.error('failed to compute checksum digest', { error: err, algorithm: this.algoName });
+                callback(errors.InternalError);
+            });
+    }
+
+    _transform(chunk, encoding, callback) {
+        const input = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk);
```
jonathan-gramain, line 88: You can probably assume that it's a `Buffer`, I don't think we would send string contents to this stream in any case.

lib/auth/streamingV4/ChecksumTransform.js
```diff
@@ -0,0 +1,94 @@
...
+        this.hash.update(input, encoding);
```
jonathan-gramain: Since `input` is a buffer, `encoding` will be ignored, so better leave it out (if you wanted to support strings, then you would have to pass `encoding` to the `Buffer.from` call above, but as I said I don't think we would have to deal with strings).

lib/auth/streamingV4/ChecksumTransform.js
```diff
@@ -0,0 +1,94 @@
...
+        this.algo = algorithms[algoName];
+        this.hash = this.algo.createHash();
+        this.digest = undefined;
+        this.expectedDigest = expectedDigest;
+        this.isTrailer = isTrailer;
+        this.trailerChecksumName = undefined;
+        this.trailerChecksumValue = undefined;
+    }
+
+    setExpectedChecksum(name, value) {
```
jonathan-gramain, line 19: I feel like whether the checksum comes from the checksum header or a trailer, they should be treated the same way when it comes to comparing with the computed checksum.
I understand that you defer validation of the checksum value itself to `validateChecksum`, but maybe it would be best to validate it directly here, then update `this.expectedChecksum` so the following logic of `validateChecksum` can be agnostic of where the checksum comes from. (If you really need to defer the error reporting to `validateChecksum`, you could set an error flag `this.checksumError` or similar and check it at the beginning of `validateChecksum`, but it may not be needed)

jonathan-gramain (commented):
Done up to commit [add ChecksumTransform to calculate and verify stream checksums](https://github.com/scality/cloudserver/pull/6112/changes/fd0af1c99f0f2f2bc1d1a97c41a059884412a8d3)

lib/api/apiUtils/object/prepareStream.js
```diff
@@ -1,50 +1,114 @@
...
+        log.debug('prepareStream invalid checksum headers', checksumAlgo);
+        return { error: arsenalErrorFromChecksumError(checksumAlgo), stream: null };
     }
 
-    const trailingChecksumTransform = new TrailingChecksumTransform(log);
-    trailingChecksumTransform.on('error', errCb);
-    stream.pipe(trailingChecksumTransform);
-    trailingChecksumTransform.headers = stream.headers;
-    return trailingChecksumTransform;
+    let stream = request;
```
jonathan-gramain: I don't think you benefit from creating and overwriting this temporary variable, you could just use `request` or middleware streams like `v4Transform` directly where `stream` is used.

lib/api/apiUtils/object/storeObject.js
```diff
@@ -58,31 +59,67 @@ function checkHashMatchMD5(stream, hashedStream, dataRetrievalInfo, log, cb) {
...
+                        log.error('dataStore failed to delete old data', { error: deleteErr });
+                    }
+                    cbOnce(streamErr);
+                });
+            };
+
+            const doValidate = () => {
+                const checksumErr = checksumedStream.stream.validateChecksum();
+                if (checksumErr) {
+                    log.debug('failed checksum validation stream', checksumErr);
```
jonathan-gramain:
```suggestion
                    log.debug('failed checksum validation stream', { error: checksumErr });
```

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: reworked as FSM

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: reworked as FSM

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: removed in new FSM, trim removes `\n`

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: reworked as FSM

lib/auth/streamingV4/ChecksumTransform.js
leif-scality: The thing is that when we call `setExpectedChecksum` we don't know if `_flush` has been called and finished, so `this.digest` may not be ready. In `dataStore` we make sure the checksumed stream is finished before calling `validateChecksum`, so we know `this.digest` is ready.

Let me know if this responds your concern or if I missed something

tests/unit/api/apiUtils/object/storeObject.js
```diff
@@ -0,0 +1,302 @@
...
+                'x-amz-checksum-crc32': 'AAAAAA==',
+            }, 'hello world');
+            dataStore({}, null, request, 0, null, {}, log, err => {
+                assert.strictEqual(err.message, 'BadDigest');
+                assert(batchDeleteStub.calledOnce);
+                done();
+            });
+        });
+
+        it('should not delete stored data when checksum validation passes', done => {
```
jonathan-gramain, line 176: The test is marked as "checksum validation passes" but it seems there is no checksum passed at all, is it meant to pass a valid `x-amz-checksum-crc32` header matching the `hello world` payload for example?

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
```diff
@@ -0,0 +1,608 @@
...
+const algos = [
+    { name: 'crc32', wrongDigest: 'AAAAAA==' },
+    { name: 'crc32c', wrongDigest: 'AAAAAA==' },
+    { name: 'crc64nvme', wrongDigest: 'AAAAAAAAAAA=' },
+    { name: 'sha1', wrongDigest: 'AAAAAAAAAAAAAAAAAAAAAAAAAAA=' },
+    { name: 'sha256', wrongDigest: 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=' },
+];
+
+// Build the raw chunked body for STREAMING-UNSIGNED-PAYLOAD-TRAILER
+function buildTrailerBody(algoName, wrongDigest) {
```
jonathan-gramain: nitpick on the variable name `wrongDigest`: technically this function doesn't care if the digest is correct or not, so it could just be called `digest` (meaning it could also potentially be used to build correct bodies)

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
```diff
@@ -0,0 +1,608 @@
...
+    assert.ifError(err);
+    assert.strictEqual(res.statusCode, 400, `expected 400, got ${res.statusCode}: ${res.body}`);
+    assert(res.body.includes('BadDigest'), `missing BadDigest in "${res.body}"`);
+    done();
+}
+
+// Constants for protocol scenario tests
+
+const trailerContent = Buffer.from('trailer content'); // 15 bytes, hex 'f'
+const trailerContentSha256 = '4x74k2oA6j6knXzpDwogNkS6E3MM49tPpJMjfD+ES68=';
```
jonathan-gramain:
```suggestion
const trailerContentSha256 = crypto.createHash('sha256').update(trailerContent).digest('base64');
```

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
```diff
@@ -0,0 +1,608 @@
...
+}
+
+const msgMalformedTrailer = 'The request contained trailing data that was not well-formed' +
+    ' or did not conform to our published schema.';
+const msgSdkMissingTrailer = 'x-amz-sdk-checksum-algorithm specified, but no corresponding' +
+    ' x-amz-checksum-* or x-amz-trailer headers were found.';
+
+// Create the 24 common protocol-scenario tests for a given URL factory.
+// urlFn() is called lazily at test runtime so that uploadId is available.
+function makeScenarioTests(urlFn) {
```
jonathan-gramain, line 148: Could be worth adding some tests with larger synthetic bodies (e.g. 10MB of `A`s), both with correct and wrong checksum, to check that streaming checksumming works.

jonathan-gramain (commented):
Will review the new FSM tomorrow

lib/auth/streamingV4/ChecksumTransform.js
jonathan-gramain: What I meant is that every sanity check done in the `this.isTrailer` case (https://github.com/scality/cloudserver/pull/6112/changes#diff-5ea0da6c64111c8a94b664848cc7c8e9201b530aa77f0a6e962a905af8055f9eR25) except the final digest check could be done within `setExpectedChecksum`, where the function would:
- on success, set `this.expectedChecksum` to be validated by `validateChecksum` at the end, i.e. using the same code path than for non-trailer checksum
- on error, either return a validation error immediately or mark the error in an attribute `this.checksumError` (for example) to be checked and returned first thing by `validateChecksum`

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -1,12 +1,213 @@
...
+    step(data, push, emit, log) {
+        let idx = 0;
+
+        while (idx < data.byteLength) {
+            switch (this.state) {
+                case State.READ_CHUNK_LEN: {
+                    const byte = data[idx++];
+                    if (byte === CR && this.bufferOffset === 0) {
+                        log.error('unexpected CR at start of chunk length');
+                        return errorInstances.InvalidArgument;
```
BourgoisMickael: If you don't use `customizeDescription` you can use `errors` to have a new instance with the current stacktrace, if passing an error with a fresh stacktrace matters.

Otherwise you'd have a pre instanciated error without `.stack` trace. https://github.com/scality/Arsenal/blob/362f7908b6126709add96ee981d5235d96bab58c/lib/errors/index.ts#L196-L214

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -1,12 +1,213 @@
...
+                    }
+
+                    // Check it is HEX.
+                    if (!/^[0-9a-fA-F]+$/.test(chunkLenStr)) {
+                        log.error('invalid chunk size', { chunkLenStr });
+                        return errorInstances.InvalidArgument;
+                    }
+                    const chunkLen = parseInt(chunkLenStr, 16);
+
+                    if (chunkLen > maximumAllowedPartSize) {
```
BourgoisMickael, line 94: Should we rather use `maximumAllowedUploadSize` ? It's for `PutObject`, even if it's the same value, the `maximumAllowedPartSize` is for MPU (with lower value for `MPU_TESTING`)

tests/unit/auth/TrailingChecksumTransform.js
```diff
@@ -227,3 +260,280 @@ describe('TrailingChecksumTransform class', () => {
         chunkedReader.pipe(trailingChecksumTransform);
     });
 });
+
+describe('TrailingChecksumTransform trailer parsing and emitting', () => {
```
BourgoisMickael: this can be inside the above top-level describe, to keep only 1 top-level describe

BourgoisMickael (approved):

tests/unit/auth/TrailingChecksumTransform.js
```diff
@@ -407,6 +416,30 @@ describe('TrailingChecksumTransform trailer parsing and emitting', () => {
...
+            assert.deepStrictEqual(err, errors.IncompleteBody);
+        });
+
+        it('should return IncompleteBody when stream ends in READ_CHUNK_DATA_LF (trailing CR seen, no LF)', 
+            async () => {
+            const err = await expectError(['5\r\nhello\r']);
+            assert.deepStrictEqual(err, errors.IncompleteBody);
+        });
+
+        it('should return IncompleteBody when stream ends in READ_TRAILER_CHECKSUM_LF (trailer CR seen, no LF)',
```
jonathan-gramain: nitpick: trailer -> trailing

tests/unit/auth/TrailingChecksumTransform.js
```diff
@@ -444,5 +477,63 @@ describe('TrailingChecksumTransform trailer parsing and emitting', () => {
...
+            assert.strictEqual(trailers.length, 1);
+        });
+
+        it('should return InvalidArgument when chunk data CR is not followed by LF', async () => {
+            // 1 byte of data 'A', then '\r' then 'x' instead of LF
+            const err = await expectError(['1\r\nA\rx']);
+            assert.deepStrictEqual(err, errors.InvalidArgument);
+        });
+
+        it('should return InvalidArgument when trailer CR is not followed by LF', async () => {
```
jonathan-gramain: trailer -> trailing

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -1,10 +1,207 @@
...
+const CR = '\r'.charCodeAt(0);
+const LF = '\n'.charCodeAt(0);
+
+const State = Object.freeze({
+    READ_CHUNK_LEN: 0,
+    READ_CHUNK_LEN_LF: 1,
+    READ_CHUNK_DATA: 2,
+    READ_CHUNK_DATA_CR: 3,
+    READ_CHUNK_DATA_LF: 4,
+    COMPLETED: 5,
```
jonathan-gramain: just to make it more explicit

```suggestion
    COMPLETED_NO_TRAILER: 5,
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -1,10 +1,207 @@
...
+    READ_TRAILER_CHECKSUM: 6,
+    READ_TRAILER_CHECKSUM_LF: 7,
+    END: 8,
```
jonathan-gramain: same idea

```suggestion
    COMPLETED_WITH_TRAILER: 8,
```

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -1,10 +1,207 @@
...
+    step(data, push, emit, log) {
+        let idx = 0;
+
+        while (idx < data.byteLength) {
+            switch (this.state) {
+                case State.READ_CHUNK_LEN: {
+                    const byte = data[idx++];
+                    if (byte === CR && this.bufferOffset === 0) {
+                        log.error('unexpected CR at start of chunk length');
+                        return errorInstances.InvalidArgument;
```
jonathan-gramain: minor suggestion for simplification: defer the "empty length" special error case to the length parsing stage, which I find a bit cleaner in terms of responsibility:
1. accumulate until `\r\n`
2. parse and validate what was obtained

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -1,10 +1,207 @@
...
+                    this.state = State.READ_CHUNK_LEN;
+
+                    continue;
+                }
+                case State.COMPLETED:
+                    // A byte arrived after "0\r\n" — transition to trailer reading.
+                    // bufferOffset is reset here so the shared buffer is clean for the trailer.
+                    this.state = State.READ_TRAILER_CHECKSUM;
+                    this.bufferOffset = 0;
+                // falls through
```
jonathan-gramain: I wouldn't advise relying on fall through (a legacy from C language, can cause subtle bugs), here a `continue` would work fine, letting the FSM switch normally to the next state.

lib/auth/streamingV4/trailingChecksumTransform.js
```diff
@@ -1,10 +1,207 @@
...
+                    this.chunkRemaining = chunkLen;
+                    this.state = State.READ_CHUNK_DATA;
+
+                    continue;
+                }
+                case State.READ_CHUNK_DATA: {
+                    // subarray clamps to buffer bounds, so toPush may be shorter than chunkRemaining
+                    // when the chunk spans multiple _transform calls.
+                    const toPush = data.subarray(idx, idx + this.chunkRemaining);
+                    push(toPush);
```
jonathan-gramain, line 115: It seems flow control is lacking here, which could cause memory overuse or exhaustion: `push` is the stream's push function, which may return `false` indicating that backpressure is needed (i.e. should wait for the `drain` event to continue pushing).

Basically, if `push` returns false, instead of `continue`, you would listen for the transform stream's `drain` event (with `once`), with a handler that re-invokes the FSM `step` function, and then `return`.

I'm thinking it could be more easily integrated if the FSM logic is directly embedded within the transform stream, rather than in a separate class.

jonathan-gramain (changes requested):

lib/auth/streamingV4/trailingChecksumTransform.js
jonathan-gramain: BTW, it's not related to this PR since, if I'm not mistaken, it's an issue that was present in the original `TrailingChecksumTransform` class.

I am wondering if we shouldn't create a separate ticket for this, as it could be a potential issue affecting releases containing https://scality.atlassian.net/browse/S3C-9732, when the clients use `STREAMING-UNSIGNED-PAYLOAD-TRAILER`.

If my understanding is correct, since we pipe the input chunked stream to this transform stream, we lose backpressure capability, which may cause memory ballooning.

jonathan-gramain: I asked claude to double check, its answer:

> you do not need to check the return value of this.push() inside _transform.
>
> Why: In a Node.js Transform stream, backpressure is managed by the framework through the callback mechanism. The stream will not call _transform again until the previous callback() is invoked. If the internal readable buffer is full, the stream simply stops calling _transform, which in turn stops pulling from the writable side. This is the designed flow control mechanism.
>
> The false return from push() is informational (the internal buffer has reached highWaterMark), but in _transform it's the framework's responsibility to act on it — not yours. The Node.js  docs explicitly show _transform implementations calling this.push() without checking the return value. 
>
> The only place where checking push() matters is in _read() of a custom Readable, where you are the data producer and need to know when to stop. In _transform, the stream infrastructure is the producer (via _write → _transform), so it handles this automatically.

So if that is correct, we should be good. The doc is not crystal clear about the behavior of transform streams though.

jonathan-gramain (approved):

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: I know indexing an empty array does not throw in JS, but I don't feel comfortable doing it

jonathan-gramain: You can do:

```
const checksumHeader = xAmzChecksumCnt > 0 ? checksumHeaders[0] : undefined;
```

My point was more about using a separate variable since we have zero or one checksum, it clarifies the logic.

lib/auth/streamingV4/ChecksumTransform.js
leif-scality: I had issues in the past with encoding, so I prefer to be explicit and always check

lib/auth/streamingV4/ChecksumTransform.js
leif-scality: True, removed the encoding argument

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: done

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: I also did some research, and the Transfrom will detect that we need to backoff and stop calling `_transform` automatically. We would have to manage this if we were using raw `read` and `write` calls

lib/api/apiUtils/object/prepareStream.js
leif-scality: removed

lib/api/apiUtils/object/storeObject.js
leif-scality: done

tests/unit/api/apiUtils/object/storeObject.js
leif-scality: fixed, added checksum

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
leif-scality: fixed with refactor of trailer body creation

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
leif-scality: fixed with refactor of trailer body creation

tests/functional/raw-node/test/checksumPutObjectUploadPart.js
leif-scality: done

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: done

tests/unit/auth/TrailingChecksumTransform.js
leif-scality: done

tests/unit/auth/TrailingChecksumTransform.js
leif-scality: done

tests/unit/auth/TrailingChecksumTransform.js
leif-scality: done

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: done

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: done

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: done

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: done

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: done

lib/auth/streamingV4/trailingChecksumTransform.js
leif-scality: this transform is used for both cases

lib/auth/streamingV4/ChecksumTransform.js
leif-scality: I understand, but if we return an error then we need to handle it in the `trailer` event handler not great, and setting a variable and then checking it in `validateChecksum` is not a improvement in my opinion

leif-scality:
ping

leif-scality:
ping

PR #6116, opened 2026-03-18
CLDSRV-835: Retry tcp flakiness with concurrency

BourgoisMickael:
```
Multi-Object Versioning Delete Success / With default signature
  "before each" hook for "should batch delete 1000 objects quietly":
Error: read ECONNRESET
  at TCP.onStreamRead (node:internal/stream_base_commons:216:20)
```

The other option would be to change the http agent on the s3 client to use agentkeepalive lib and define freeSocketTimeout to be less than cloudserver keep alive (5s).

### DIFF
--- a/tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
+++ b/tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
@@ -56,16 +56,31 @@ describe('Multi-Object Versioning Delete Success', function success() {
                 objects.push(`${key}${i}`);
             }
 
-            // Create objects in batches of 20 concurrently
+            // Create objects in batches of 20 concurrently. Fast connections in a
+            // batch can sit idle while Promise.all waits for slower ones; if the
+            // server closes them (keepAliveTimeout 5s), the next batch gets ECONNRESET.
+            // Retrying recovers from that transient race.
+            const putWithRetry = async (params, attempt = 0) => {
+                try {
+                    return await s3.send(new PutObjectCommand(params));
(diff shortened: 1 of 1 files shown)


tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
```diff
@@ -56,16 +56,31 @@ describe('Multi-Object Versioning Delete Success', function success() {
...
 
-            // Create objects in batches of 20 concurrently
+            // Create objects in batches of 20 concurrently. Fast connections in a
+            // batch can sit idle while Promise.all waits for slower ones; if the
+            // server closes them (keepAliveTimeout 5s), the next batch gets ECONNRESET.
+            // Retrying recovers from that transient race.
+            const putWithRetry = async (params, attempt = 0) => {
+                try {
+                    return await s3.send(new PutObjectCommand(params));
+                } catch (err) {
```
dvasilas, line 66: nit: We could filter and retry only on `ECONNRESET`.

dvasilas (approved):

tests/functional/aws-node-sdk/test/versioning/multiObjectDelete.js
BourgoisMickael: There could be multiple network error arising depending on the race case, for example `ECONNRESET` or `TimeoutError: socket hang up`.

We should not have any case where we have a permanent error in here, so I'd say it's not that bad if we retry 3 times any kind of error

tmacro (approved):

PR #6121, opened 2026-03-27
🐛 use SUR date instead of the infostore one to have an update LastMod…

DarkIsDude:
Issue: CLDSRV-878

# Pull request template

## Description

https://scality.atlassian.net/browse/CLDSRV-631 this PR introduce a regression. The associated cronjob was removed https://github.com/scality/zenko-operator/pull/540/changes#diff-c76a6b4555478f74e4c871dbf370d4e6e5ae54c4b8590b0581b1cc0894e83ae5R55 but the date used is still the one associated with it. The goal of this PR is to use the date from SUR.

```
curl -v --insecure https://10.160.99.90:8443/zenko/s3/test2/.system-d26a9498-cb7c-4a87-a44a-8ae204f5ba6c/capacity.xml?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=FRFIWTEZRPYUSHVY6WEP%2F20260327%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20260327T155955Z&X-Amz-Expires=900&X-Amz-Security-Token=eyJzYWx0IjoiQ0Z2bzd2Y29BY1ZZRHBZTXJ4Y0dpM2ExZ1NYcVBGaGpWdis5QThhU0NSMD0iLCJ0YWciOiJXaHNFWDUyMWlQWEFkWnV1WnRCOXNnPT0iLCJjaXBoZXJ0ZXh0IjoienBFRmpzeHMvYmRORzZZbU5mWFpJbGlHS01yWFliR3RaZW9NcUx0N3JXVjVJR1JnR0F1RExlZVdGSkUwdHVBNjBPbWpqeUhINkszWkp1Q1hkVWt5VGhrcUVsWStYTUpCdE5haWRjOHZBek9uUjJnc2p1L1pKRUdVUnZnMmVHMjRXS2JPZUQra1h0Q25nQWpFTjFWTWRsd25OcTBUMmw4WWNMaVBrY2owa1BaWnB2M2I0MEZCaVFsbFowOThnN2JSanhTZ2daVEtQNnNpY1hrYU0rK2tZbjMrMHhlS2NjcEFxWENWR3l6NjNxZWFuclFqNW1Ta2FDVHF1QVlCQlVGT0hnVnFWNFIzTjdiRUpCUTNSb1JGRGIzei9BaW85WERVcDE4RVNaZHVTaGxEN0ZaaktMVEJWL1hCNkRWNHd3bnRsMWRGOVRzZHlHOFZiWHFZSVN6elJ2SEpDeXlYTnNHNTBlUjNGQVNCSWZWTlc1QXlpOHk4cVVkOTFCMVBBUGxZaVJ1T284Z0twOWtvb0NFS2hCNUprSGc1Tzl2RFhRVEdoUkxvdnplaUIzbkRYTnIrL1FhZ0dmZ0tzSEMzZVIyT2hCYnhTUU50NVBrb012bGVodnh1dlNPZ01ucFFFczVudzZTeHBtMDlZVHRwOUJZazk4Q1IxcUZnVlJBMEZDTVA4a2dwU3VmbHh3dFNMZkhIVmNkN2lCWHh5MEJDaGVMLzE4NkZkMXhDaFVYT1A0d3o0Yjd4OHZKV1JYZm82WStWZlU2a1Q5cW52WGhOSmlJeGtmalU0aUgwT2k4bTAyU2lUeVRIdGJjMlBFK1dwaTVZL0xpbFpudjMwNVJlbWR1SDNLZStKTlpBS3FSdVYvM3ZxMnczdWlER1djMVh4SWJHNytxTWhlaWZVcnRENWsvMjg2ekY5UWxSTmpwYVZHdGZ6R2tVS3VQbk82NENubkt4UFcxc05wQyt4WEljUVV0WVF4MjlFYTlDSndZZEtlazJtYXVneXdUOE1zZFA5SkI5WVRTZHRqMzRWcXNuZ2tQMVdlZWZBMTZFYjdWNU5EZm5RS0pobjFpdUVhaGpVY1ViVll4L0ZOTEVBYW1wSlltN0plNjdDeEs0dmluOTI5QXpUMHJJand5a3lNam5Fa0x3VjVrdDV5WEJ1TmNPa1lUc3o5bzd3Tnp6Zlp3MnI4cHY5cVc0YXBhcnpxNHVOcTFXdjdZeG1VY2dLRExwYlFRc1VQWW5SREdkZDNpcDNzeUR3ZXcxMzZ5V2tMOGJYU0JFUXIxdHZmL1BRQTFRdUR6Q0s1TmtzWGd6cW0vdXorZ0FzQVVub1R2cTh1RzNjaGJnR0hzalFsa29MUjFsWTJGQ2t1bHF3NTlCUjJkVzJaRmptdFQ3aTJkWnk5T2pWZFRDUFc2QXM5N1Q4T3d5SXlKUytVZ3Q2YzA2a1JVZ2dNVHhETFBpQ2FUbDFJbS9rZ3YxTmZzVWg0aFhabFBvdnkrYlJHTWFrRW5sMXdQYXVzSzlIZHMxR3FvcGVkSjlNK0svSXhqeHpkS3B4YU1ZZUcwSDBVbGVnSkRQSjl0TmRhVE15MXZqZ0F4VzlXWm9qSkpMSXNXWHhDQzlIai95MGoySFN5WGFLWGdqYnZtMlFOcWFLckFHN1lXRUVSQkdQeG1HWktwRHlhcmFLdGVaQzVNWXU4ckNVdVdTc1U3ekQrMFlHMU9CVWlMTTJyUXdWQmdSWFZ2QlRRUlVUZTJPbFplUVNZY1dZNUg2aWNON1dYZ3pEdndOUm5SeWIvVzM5R2hUMGFCL0diYTlYRVI0K0NLT1A5UXgifQ%3D%3D&X-Amz-Signature=4ca11d205c394debb177a26b8e292493a1478a91364ffd96b306628db0e0ddc2&X-Amz-SignedHeaders=host
```

With the fix `< last-modified: Fri, 27 Mar 2026 16:04:16 GMT`
Without the fix `< last-modified: Wed, 25 Mar 2026 14:12:11 GMT`

### DIFF
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -17,13 +17,13 @@ async function getVeeamFile(request, response, bucketMd, log) {
     }
 
     if ('tagging' in request.query) {
-        return respondWithData(request, response, log, bucketMd,
+        return await respondWithData(request, response, log, bucketMd,
             buildHeadXML('<Tagging><TagSet></TagSet></Tagging>'));
     }
 
     try {
         const result = await buildVeeamFileData(request, bucketMd, log, 'getVeeamFile');
-        return respondWithData(request, response, log, result.bucketData, result.xmlContent, result.modified);
+        return await respondWithData(request, response, log, result.bucketData, result.xmlContent, result.modified);
     } catch (err) {
         return responseXMLBody(err, null, response, log);
     }
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -1,9 +1,8 @@
 const xml2js = require('xml2js');
 const { errors } = require('arsenal');
 const metadata = require('../../metadata/wrapper');
-const { respondWithData, buildHeadXML, getFileToBuild, isSystemXML } = require('./utils');
+const { respondWithData, buildHeadXML, getFileToBuild, isSystemXML, fetchCapacityMetrics } = require('./utils');
 const { responseXMLBody } = require('arsenal/build/lib/s3routes/routesUtils');
-const UtilizationService = require('../../../lib/utilization/instance');
 
 /**
  * Returns system.xml or capacity.xml files for a given bucket.
@@ -55,31 +54,10 @@ function getVeeamFile(request, response, bucketMd, log) {
         };
 
         if (!isSystemXML(request.objectKey)) {
-            const bucketKey = `${bucketMd._name}_${new Date(bucketMd._creationDate).getTime()}`;
-            return UtilizationService.getUtilizationMetrics('bucket', bucketKey, null, {}, (err, bucketMetrics) => {
+            return fetchCapacityMetrics(bucketMd, request, log, 'getVeeamFile', (err, bucketMetrics) => {
                 if (err) {
-                    const statusCode = err.response?.status || err.statusCode || err.code;
-
-                    // Only handle 404 gracefully (no metrics available yet, e.g. post-install)
-                    if (statusCode === 404) {
-                        log.warn('UtilizationService returned 404 when fetching capacity metrics', {
-                            method: 'getVeeamFile',
-                            bucket: request.bucketName,
-                            error: err.message || err.code,
-                        });
-                        return finalizeRequest();
-                    }
-
-                    log.error('error fetching capacity metrics from UtilizationService', {
-                        method: 'getVeeamFile',
-                        bucket: request.bucketName,
-                        error: err.message || err.code,
-                        statusCode,
-                    });
-
                     return responseXMLBody(errors.InternalError, null, response, log);
                 }
-
                 return finalizeRequest(bucketMetrics);
             });
         }
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -1,7 +1,5 @@
-const xml2js = require('xml2js');
 const { errors } = require('arsenal');
-const metadata = require('../../metadata/wrapper');
-const { respondWithData, buildHeadXML, getFileToBuild, isSystemXML, fetchCapacityMetrics } = require('./utils');
+const { respondWithData, buildHeadXML, buildVeeamFileData } = require('./utils');
 const { responseXMLBody } = require('arsenal/build/lib/s3routes/routesUtils');
 
 /**
@@ -17,51 +15,18 @@ function getVeeamFile(request, response, bucketMd, log) {
     if (!bucketMd) {
         return responseXMLBody(errors.NoSuchBucket, null, response, log);
     }
+
     if ('tagging' in request.query) {
         return respondWithData(request, response, log, bucketMd,
             buildHeadXML('<Tagging><TagSet></TagSet></Tagging>'));
     }
-    return metadata.getBucket(request.bucketName, log, (err, data) => {
+
+    return buildVeeamFileData(request, bucketMd, log, 'getVeeamFile', (err, result) => {
         if (err) {
-            return responseXMLBody(errors.InternalError, null, response, log);
+            return responseXMLBody(err, null, response, log);
         }
-        const finalizeRequest = bucketMetrics => {
-            const fileToBuild = getFileToBuild(request, data._capabilities?.VeeamSOSApi);
-            if (fileToBuild.error) {
-                return responseXMLBody(fileToBuild.error, null, response, log);
-            }
 
-            // Extract the last modified date, but do not include it when computing
-            // the file's ETag (md5)
-            const modified = bucketMetrics?.date || new Date();
-            delete fileToBuild.value.LastModified;
-
-            // The SOSAPI metrics are dynamically computed using the SUR backend.
-            if (bucketMetrics && !fileToBuild.value.CapacityInfo?.Used) {
-                fileToBuild.value.CapacityInfo.Used = Number(bucketMetrics.bytesTotal);
-                fileToBuild.value.CapacityInfo.Available =
-                    Number(fileToBuild.value.CapacityInfo.Capacity) - Number(bucketMetrics.bytesTotal);
-                // TODO CLDSRV-633 when SUR backend supports realtime metrics: it will
-                // report the real last cseq/date processed by SUR, instead of the current date,
-                // ensuring no issue in a SOSAPI context. We should use this information.
-            }
-
-            const builder = new xml2js.Builder({
-                headless: true,
-            });
-            return respondWithData(request, response, log, data,
-                buildHeadXML(builder.buildObject(fileToBuild.value)), modified);
-        };
-
-        if (!isSystemXML(request.objectKey)) {
-            return fetchCapacityMetrics(bucketMd, request, log, 'getVeeamFile', (err, bucketMetrics) => {
-                if (err) {
-                    return responseXMLBody(errors.InternalError, null, response, log);
-                }
-                return finalizeRequest(bucketMetrics);
-            });
-        }
-        return finalizeRequest();
+        return respondWithData(request, response, log, result.bucketData, result.xmlContent, result.modified);
     });
 }
 
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -34,7 +34,7 @@ function getVeeamFile(request, response, bucketMd, log) {
 
             // Extract the last modified date, but do not include it when computing
             // the file's ETag (md5)
-            const modified = bucketMetrics?.date || fileToBuild.value.LastModified;
+            const modified = bucketMetrics?.date || new Date();
             delete fileToBuild.value.LastModified;
 
             // The SOSAPI metrics are dynamically computed using the SUR backend.
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -34,7 +34,7 @@ function getVeeamFile(request, response, bucketMd, log) {
 
             // Extract the last modified date, but do not include it when computing
             // the file's ETag (md5)
-            const modified = fileToBuild.value.LastModified;
+            const modified = bucketMetrics?.date || fileToBuild.value.LastModified;
             delete fileToBuild.value.LastModified;
 
             // The SOSAPI metrics are dynamically computed using the SUR backend.
@@ -53,15 +53,14 @@ function getVeeamFile(request, response, bucketMd, log) {
             return respondWithData(request, response, log, data,
                 buildHeadXML(builder.buildObject(fileToBuild.value)), modified);
         };
+
         if (!isSystemXML(request.objectKey)) {
             const bucketKey = `${bucketMd._name}_${new Date(bucketMd._creationDate).getTime()}`;
             return UtilizationService.getUtilizationMetrics('bucket', bucketKey, null, {}, (err, bucketMetrics) => {
                 if (err) {
-                    // Handle errors from UtilizationService/scubaclient
-                    // axios errors have status in err.response.status
                     const statusCode = err.response?.status || err.statusCode || err.code;
+
                     // Only handle 404 gracefully (no metrics available yet, e.g. post-install)
-                    // For 404, continue with static capacity data (Used=0 from bucket metadata)
                     if (statusCode === 404) {
                         log.warn('UtilizationService returned 404 when fetching capacity metrics', {
                             method: 'getVeeamFile',
@@ -70,14 +69,17 @@ function getVeeamFile(request, response, bucketMd, log) {
                         });
                         return finalizeRequest();
                     }
+
                     log.error('error fetching capacity metrics from UtilizationService', {
                         method: 'getVeeamFile',
                         bucket: request.bucketName,
                         error: err.message || err.code,
                         statusCode,
                     });
+
                     return responseXMLBody(errors.InternalError, null, response, log);
                 }
+
                 return finalizeRequest(bucketMetrics);
             });
         }
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -11,7 +11,7 @@ const { responseXMLBody } = require('arsenal/build/lib/s3routes/routesUtils');
  * @param {object} log - logger object
  * @returns {undefined} -
  */
-function getVeeamFile(request, response, bucketMd, log) {
+async function getVeeamFile(request, response, bucketMd, log) {
     if (!bucketMd) {
         return responseXMLBody(errors.NoSuchBucket, null, response, log);
     }
@@ -21,13 +21,12 @@ function getVeeamFile(request, response, bucketMd, log) {
             buildHeadXML('<Tagging><TagSet></TagSet></Tagging>'));
     }
 
-    return buildVeeamFileData(request, bucketMd, log, 'getVeeamFile', (err, result) => {
-        if (err) {
-            return responseXMLBody(err, null, response, log);
-        }
-
+    try {
+        const result = await buildVeeamFileData(request, bucketMd, log, 'getVeeamFile');
         return respondWithData(request, response, log, result.bucketData, result.xmlContent, result.modified);
-    });
+    } catch (err) {
+        return responseXMLBody(err, null, response, log);
+    }
 }
 
 module.exports = getVeeamFile;
--- a/lib/routes/veeam/get.js
+++ b/lib/routes/veeam/get.js
@@ -22,7 +22,7 @@ async function getVeeamFile(request, response, bucketMd, log) {
     }
 
     try {
-        const result = await buildVeeamFileData(request, bucketMd, log, 'getVeeamFile');
+        const result = await buildVeeamFileData(request, bucketMd, log);
         return await respondWithData(request, response, log, result.bucketData, result.xmlContent, result.modified);
     } catch (err) {
         return responseXMLBody(err, null, response, log);
--- a/lib/routes/veeam/head.js
+++ b/lib/routes/veeam/head.js
@@ -1,7 +1,5 @@
-const xml2js = require('xml2js');
 const { errors } = require('arsenal');
-const metadata = require('../../metadata/wrapper');
-const { getResponseHeader, buildHeadXML, getFileToBuild, isSystemXML, fetchCapacityMetrics } = require('./utils');
+const { getResponseHeader, buildVeeamFileData } = require('./utils');
 const { responseXMLBody, responseContentHeaders } = require('arsenal/build/lib/s3routes/routesUtils');
 
 /**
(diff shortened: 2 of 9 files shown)


lib/routes/veeam/get.js
```diff
@@ -34,7 +34,7 @@ function getVeeamFile(request, response, bucketMd, log) {
 
             // Extract the last modified date, but do not include it when computing
             // the file's ETag (md5)
-            const modified = fileToBuild.value.LastModified;
+            const modified = bucketMetrics?.date || fileToBuild.value.LastModified;
```
DarkIsDude: Should we remove `fileToBuild.value.LastModified`. IMO yes ?

lib/routes/veeam/head.js
DarkIsDude, line 1: We don't call the SUR backend on this route. So we can't use the new date. It's not an issue as Veeam is not doing head but can lead to weird behaviour. We have the same "issue" in the list.

DarkIsDude: Please be advised that the UI is using the List/Head...

DarkIsDude: - HEAD: Clear-cut: add the UtilizationService call for capacity.xml, same pattern as GET. HEAD is contractually required to return the same headers as GET — inconsistent Last-Modified between GET and HEAD would be a spec violation.

- LIST: This is the tricky one. Without calling UtilizationService, there are really only two options:
1. Keep the metadata date — Stale, and actively harmful: if the LIST LastModified for capacity.xml doesn't change, Veeam may cache the entry and skip re-fetching via GET, which defeats the entire purpose of using the metrics date.
2. Use new Date().toISOString() in LIST for capacity.xml — Always returns current time. This signals to Veeam: "this file is always potentially fresh, always re-fetch via GET/HEAD." Since GET/HEAD now return the accurate bucketMetrics.date, Veeam gets the right value on the actual fetch. Capacity metrics do change constantly, so this is semantically honest.

Option 2 is the pragmatic fix: no extra service call, no stale caching problem, and the accurate date is still served on GET/HEAD.

lib/routes/veeam/get.js
francoisferrand: we would still need a "default" value though, we cannot pass "undefined" as the date?

lib/routes/veeam/get.js
```diff
@@ -53,15 +53,14 @@ function getVeeamFile(request, response, bucketMd, log) {
             return respondWithData(request, response, log, data,
                 buildHeadXML(builder.buildObject(fileToBuild.value)), modified);
         };
+
         if (!isSystemXML(request.objectKey)) {
             const bucketKey = `${bucketMd._name}_${new Date(bucketMd._creationDate).getTime()}`;
             return UtilizationService.getUtilizationMetrics('bucket', bucketKey, null, {}, (err, bucketMetrics) => {
                 if (err) {
-                    // Handle errors from UtilizationService/scubaclient
-                    // axios errors have status in err.response.status
```
francoisferrand, line 61: I think this comment was actually useful : we never have `err.response.status` in our code base, kind of weird to see it checked in addition to the more usual (in our code!) err.statusCode and err.code

lib/routes/veeam/head.js
francoisferrand: nothing currently implemented: this is the same code as before, right?

Fetching metrics in HEAD and returning current date in LIST seem ok to me, though Veeam may still do weird things: e.g. it may not like that LIST returns a date which is newer than the GET/HEAD call it makes next... Could we make a call to utilization api for LIST as well?

tests/unit/api/apiUtils/objectLockHelpers.js
```diff
@@ -172,7 +172,7 @@ describe('objectLockHelpers: calculateRetainUntilDate', () => {
         };
         const date = moment();
         const expectedRetainUntilDate
-            = date.add(mockConfigWithYears.years * 365, 'days');
+            = date.add(mockConfigWithYears.years * 365 * 86400000, 'ms');
```
francoisferrand, line 175: why this change? 'days' is more readable than 864000000ms...

lib/routes/veeam/get.js
DarkIsDude: As this logic is not updated anymore, should the default be new Date() then ?

lib/routes/veeam/head.js
DarkIsDude: On the HEAD / List nothing is implemented indeed. We can make a to the utilization API of course but I'm worried about the performance ? It's only for 2 files right now (system / capacity), so I think it's fine but if veeam call it every seconds it can be an issue 🤷. What do you think @francoisferrand ? Should I also validate this with someone (if we choose the new Date for list).

tests/unit/api/apiUtils/objectLockHelpers.js
DarkIsDude: If you try to run test locally you have an error because it's not on the same timezone. The goal is to fix that 🙏

lib/routes/veeam/get.js
DarkIsDude: Even if I can understand your point of view, this is/should be reflected in test (and it's the case). It's also a pretty standard piece of code when working with axios. This comment just describe the code, it's not tweak or anything weird ?

francoisferrand: it may be pretty standard when working with axios, but we don't actually use axios in cloudserver at all : it really only comes here from ScubaClient (i.e. we deal with a dependency of a dependency), which makes it kind of confusing...

but yes, hopefully it should be covered by tests...so we can't remove it by mistake (but not sure it is...)

lib/routes/veeam/head.js
francoisferrand: I guess we can make a call for list as well (esp. a single call should be enough for both, right?), I don't think it would cause an issue even if called every second...

@FCScality may have more info about veeam behavior...

lib/routes/veeam/get.js
francoisferrand: yes, maybe `new Date()` is better ; but not sure if we are "allowed" to do this: this could trigger some warning in veeam, if they see this and understand we are faking it... Or maybe I'm just too cautious :D

@FCScality do you think returning current time (by default) should be fine?

(or maybe Scuba API call guarantees we have a date, and we will always have a "valid" bucketMetrics here, and should return an error otherwise?)

tests/unit/api/apiUtils/objectLockHelpers.js
francoisferrand: weird...
could this be due to the use of `moment` above...and maybe fixe day just using plain old Date() instead?

DarkIsDude: That's not really weird. The root cause is that in production code we add milliseconds where here we add days. That's why moving from ms reproduce what is done in production and fix the issue.

Removing moment will not solve the issue if we still use days in test and not in production

lib/routes/veeam/get.js
francoisferrand: discussed with @FCScality : returning current date is a sane default → `new Date()` would be good.
(even better if we can guarantee that scuba returns a value though)

lib/routes/veeam/head.js
francoisferrand: we expect Veeam to call the API every 10 min, so extra call is fine I guess.

tests/unit/api/apiUtils/objectLockHelpers.js
francoisferrand: what I mean is it's weird to have an issue with _timezone_ : date mathematics should work fine from ms to days, and not be affected by timezone...

(adding `365 days` should be the same as adding `365 * 86400000 ms`)

lib/routes/veeam/head.js
DarkIsDude: Yes I'm on it :pray: . I also ask Florent confirmation this morning

lib/routes/veeam/get.js
DarkIsDude: Yes I was on it, Florent told me that this morning 🙏

lib/routes/veeam/get.js
DarkIsDude: It is

tests/unit/api/apiUtils/objectLockHelpers.js
DarkIsDude: moment.add(N, 'days') is DST-aware — it adjusts across DST transitions, so a "day" can be 23h or 25h. moment.add(N * 86400000, 'ms') is DST-unaware — it always adds exactly 86400000ms per day.

In the CI it's not an issue as we are not in a timezone with DST... That's not the case locally.

lib/routes/veeam/head.js
```diff
@@ -17,26 +17,45 @@ function headVeeamFile(request, response, bucketMd, log) {
...
+
     return metadata.getBucket(request.bucketName, log, (err, data) => {
         if (err) {
             return responseXMLBody(errors.InternalError, null, response, log);
         }
-        const fileToBuild = getFileToBuild(request, data._capabilities?.VeeamSOSApi);
-        if (fileToBuild.error) {
-            return responseXMLBody(fileToBuild.error, null, response, log);
+
+        const finalizeRequest = bucketMetrics => {
```
francoisferrand: seems very redundant with the code in veeam/get.js ?
can this be factorized, or is there a subtle difference which makes it too complicated for the benefit?

lib/routes/veeam/list.js
```diff
@@ -96,37 +96,53 @@ function listVeeamFiles(request, response, bucketMd, log) {
...
             return responseXMLBody(errors.InternalError, null, response, log);
         }
-        const filesToBuild = [];
-        const fieldsToGenerate = [];
-        if (data._capabilities?.VeeamSOSApi?.SystemInfo) {
-            fieldsToGenerate.push({
-                ...data._capabilities?.VeeamSOSApi?.SystemInfo,
-                name: `${validPath}system.xml`,
+
+        const buildAndRespond = bucketMetrics => {
```
francoisferrand: nit: should we move this function to async, so it is easier to write and read (i.e. let things flow in the right order : first thing → next thing) ? or use async.series/async.waterfall ?

lib/routes/veeam/list.js
```diff
@@ -96,37 +96,53 @@ function listVeeamFiles(request, response, bucketMd, log) {
...
+            // When `versions` is present, listing should return a versioned list
+            return respondWithData(request, response, log, data,
+                buildXMLResponse(request, filesToBuild, 'versions' in request.query));
+        };
+
         if (data._capabilities?.VeeamSOSApi?.CapacityInfo) {
-            fieldsToGenerate.push({
-                ...data._capabilities?.VeeamSOSApi?.CapacityInfo,
-                name: `${validPath}capacity.xml`,
+            return fetchCapacityMetrics(bucketMd, request, log, 'listVeeamFiles', (err, bucketMetrics) => {
```
francoisferrand: should this condition be handled directly in `fetchCapacityMetrics`? This would allow
- more logical code in handlers (fetch capacity → process it), without the "callback"
- could create a "default" bucketMetrics with the current date, so we don't need to repeat `bucketMetrics?.date || new Date()`

(it would also solve the previous comment, as a side benefit)

lib/routes/veeam/get.js
```diff
@@ -34,7 +33,7 @@ function getVeeamFile(request, response, bucketMd, log) {
 
             // Extract the last modified date, but do not include it when computing
             // the file's ETag (md5)
-            const modified = fileToBuild.value.LastModified;
+            const modified = bucketMetrics?.date || new Date();
             delete fileToBuild.value.LastModified;
```
francoisferrand: nit: we keep deleting `fileToBuild.value.LastModified` : should we not do this in getFileToBuild?

lib/routes/veeam/head.js
```diff
@@ -13,31 +11,23 @@ const { responseXMLBody, responseContentHeaders } = require('arsenal/build/lib/s
...
-        delete fileToBuild.value.LastModified;
-        // Recompute file content to generate appropriate content-md5 header
-        const builder = new xml2js.Builder({
-            headless: true,
-        });
-        const dataBuffer = Buffer.from(buildHeadXML(builder.buildObject(fileToBuild)));
-        return responseContentHeaders(null, {}, getResponseHeader(request, data,
-            dataBuffer, modified, log), response, log);
-    });
+
```
francoisferrand, line 18: nit: not really related to this PR, but I wonder why we need to make a call to getBucket even though we receive `bucketMD` as function parameter? do you know why or what is the difference?

lib/routes/veeam/utils.js
```diff
@@ -192,20 +199,123 @@ function getFileToBuild(request, data, inlineLastModified = false) {
...
+ * @param {object} request - incoming request
+ * @param {object} log - logger object
+ * @param {string} method - calling method name for log context
+ * @returns {Promise<object>} bucketMetrics always has at least a `date` field;
+ *   on a real 404 the date defaults to new Date()
+ */
+async function fetchCapacityMetrics(bucketMd, request, log, method) {
+    const bucketKey = `${bucketMd._name}_${new Date(bucketMd._creationDate).getTime()}`;
+    try {
+        return await new Promise((resolve, reject) =>
```
francoisferrand: prefer to use `promisify`

lib/routes/veeam/list.js
```diff
@@ -83,51 +83,75 @@ function buildXMLResponse(request, arrayOfFiles, versioned = false) {
...
     if (!('list-type' in request.query) && !('versions' in request.query)) {
         return responseXMLBody(errorInstances.InvalidRequest
             .customizeDescription('The Veeam folder does not support this action.'), null, response, log);
     }
-    return metadata.getBucket(request.bucketName, log, (err, data) => {
-        if (err) {
+
+    let data;
+    try {
+        data = await new Promise((resolve, reject) =>
```
francoisferrand: should we prefer promisify here?

lib/routes/veeam/utils.js
```diff
@@ -20,43 +23,45 @@ function _decodeURI(uri) {
...
-    });
-    dataStream.on('end', () => {
-        if (exceeded) {
-            log.error('data stream exceed announced size',
-                { parsedContentLength, overflow: cursor });
-            return callback(errors.InternalError);
-        } else {
-            return callback(null, value.toString());
-        }
+    return new Promise((resolve, reject) => {
```
francoisferrand: - use promisify
- in the guidelines, I think there was mention of never returning a promise (to ensure stack trace is correct)


```suggestion
    return await utils.promisify(() => { ... })();
```

lib/routes/veeam/utils.js
```diff
@@ -192,20 +199,123 @@ function getFileToBuild(request, data, inlineLastModified = false) {
...
+ */
+async function buildVeeamFileData(request, bucketMd, log, name) {
+    let data;
+    try {
+        data = await new Promise((resolve, reject) =>
+            metadata.getBucket(request.bucketName, log, (err, result) => {
+                if (err) { return reject(err); }
+                return resolve(result);
+            })
+        );
```
francoisferrand: shouldn't we use promisify?
```suggestion
        data = await utils.promisify(cb => metadata.getBucket(request.bucketName, log, cb))();
```

lib/routes/veeam/utils.js
```diff
@@ -192,20 +199,123 @@ function getFileToBuild(request, data, inlineLastModified = false) {
...
+        }
+    } else {
+        bucketMetrics = { date: new Date() };
+    }
+
+    const modified = bucketMetrics.date;
+    if (
+        bucketMetrics.bytesTotal !== undefined
+            && fileToBuild.value.CapacityInfo
+            && !fileToBuild.value.CapacityInfo.Used
```
francoisferrand: not really "pretty" ?

```suggestion
    if (bucketMetrics.bytesTotal !== undefined
            && fileToBuild.value.CapacityInfo
            && !fileToBuild.value.CapacityInfo.Used) {
```

lib/routes/veeam/utils.js
```diff
@@ -192,20 +199,123 @@ function getFileToBuild(request, data, inlineLastModified = false) {
...
+        if (statusCode === 404) {
+            log.warn('UtilizationService returned 404 when fetching capacity metrics', {
+                method,
+                bucket: request.bucketName,
+                error: err.message || err.code,
+            });
+            return { date: new Date() };
+        }
+        log.error('error fetching capacity metrics from UtilizationService', {
+            method,
```
francoisferrand: is it really worth the extra param?
esp. since werelogs will dump all log lines associated with the request on error: we will have the http request verb/path/...

lib/routes/veeam/utils.js
```diff
@@ -192,20 +199,123 @@ function getFileToBuild(request, data, inlineLastModified = false) {
...
+/**
+ * Builds Veeam file data (XML content + response metadata) for a given request.
+ *
+ * @param {object} request - incoming request
+ * @param {object} bucketMd - bucket metadata from the router
+ * @param {object} log - logger object
+ * @param {string} name - calling method name (for log context)
+ * @returns {Promise<object>} result with { xmlContent, dataBuffer, modified, bucketData }
+ */
+async function buildVeeamFileData(request, bucketMd, log, name) {
```
francoisferrand: `name` is not used

lib/routes/veeam/head.js
DarkIsDude: To be honest I don't know. I have the feeling that nothing is different 😬. Maybe the idea was to keep the handler as independant as possible ?

lib/routes/veeam/utils.js
DarkIsDude: 🤔 I'm not sure the usage of promisify here will add a value ? Finally it's the same ? Thanks for the await 🙏

lib/routes/veeam/utils.js
DarkIsDude: As you want 🙏. I prefer the other way as it's really a block. Prettier should resolve that soon 🙏

delthas (approved):

lib/routes/veeam/head.js
```diff
@@ -13,31 +11,23 @@ const { responseXMLBody, responseContentHeaders } = require('arsenal/build/lib/s
...
+    try {
+        const result = await buildVeeamFileData(request, bucketMd, log);
+        return responseContentHeaders(
```
francoisferrand, line 21:
```suggestion
        return await responseContentHeaders(
```

lib/routes/veeam/head.js
francoisferrand: let's create a followup ticket to investigate this... if we can remove a redundant metadata call, that would be a quick win...

lib/routes/veeam/utils.js
francoisferrand: it would avoid the ugly settle (probably keeping the jstutil.once, though):
```
        const settle = jsutil.once((err, result) => {
            if (err) { return reject(err); }
            return resolve(result);
        });
```

lib/routes/veeam/utils.js
```diff
@@ -20,43 +28,45 @@ function _decodeURI(uri) {
...
+    return new Promise((resolve, reject) => {
+        const value = Buffer.alloc(parsedContentLength);
+        const settle = jsutil.once((err, result) => {
+            if (err) { return reject(err); }
+            return resolve(result);
+        });
+        const dataStream = prepareStream(request, request.streamingV4Params, log, settle);
+        let cursor = 0;
+        let exceeded = false;
+        dataStream.on('data', data => {
```
francoisferrand: nit: could we use `pipeline` here?

francoisferrand (approved):

lib/routes/veeam/head.js
DarkIsDude: https://scality.atlassian.net/browse/CLDSRV-882

lib/routes/veeam/head.js
DarkIsDude: 😬 🙏

DarkIsDude: @francoisferrand that's not an async method ?

PR #6124, opened 2026-03-31
Improvement/cldsrv 875 integrate shared prettier

benzekrimaha:
Issue: CLDSRV-875


.github/workflows/tests.yaml
```diff
@@ -8,6 +8,10 @@ on:
     branches-ignore:
       - 'q/*/**'
 
+  pull_request:
```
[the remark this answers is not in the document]
benzekrimaha: Adressed by adding a workflow-level concurrency group, so push + pull_request runs for the same commit will cancel the older one.

package.json
```diff
@@ -1,148 +1,152 @@
 {
```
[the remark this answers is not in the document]
benzekrimaha: enforcing 4-space is cleaner and consistent. The tradeoff is that we will  need to run yarn format after yarn add/remove. That's a documented expectation, not a blocker.

CONTRIBUTING.md
```diff
@@ -15,3 +14,13 @@ yarn dev
 
 This will start the developement server, watching for changes and restarting
 as needed.
+
+### Formatting
```
DarkIsDude, line 18: Then I think we should ignore package.json ?

package.json
```diff
@@ -1,148 +1,152 @@
...
+        "ft_sse_cleanup": "cd tests/functional/sse-kms-migration && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 10000 cleanup.js --exit",
+        "ft_sse_before_migration": "cd tests/functional/sse-kms-migration && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 10000 cleanup.js beforeMigration.js --exit",
+        "ft_sse_migration": "cd tests/functional/sse-kms-migration && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 10000 migration.js --exit",
+        "ft_sse_arn": "cd tests/functional/sse-kms-migration && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 10000 cleanup.js arnPrefix.js --exit",
+        "lint": "eslint $(git ls-files '*.js')",
+        "lint_md": "mdlint $(git ls-files '*.md')",
+        "format:changed": "scripts/format-changed.sh",
```
DarkIsDude:
```suggestion
        "format:check:diff": "scripts/format-changed.sh",
```

Maybe ?

scripts/format-changed.sh
DarkIsDude, line 1: Maybe a quick comment to explain the goal of the script and how it's working ?

CONTRIBUTING.md
benzekrimaha: either add it to the ignore or we inforce the linting in it , it's a choice to make we can check with others what they prefer

package.json
benzekrimaha: tried to keep the same name as the sh , can indeed be changed

DarkIsDude (approved):

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,51 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  lint:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
francoisferrand: these are not the latest actions: upgrade instead of introducing old dependency

.github/workflows/tests.yaml
francoisferrand, line 1: large reformatting for a few _deleted_ lines make this very difficult to review...

scripts/format-changed.sh
```diff
@@ -0,0 +1,57 @@
...
+    DIFF_BASE="$DEFAULT_BASE"
+fi
+
+if ! git rev-parse --verify "${DIFF_BASE}^{commit}" >/dev/null 2>&1; then
+    if [[ -n "$BASE_REF" ]]; then
+        git fetch --no-tags --prune --depth=50 origin "${BASE_REF#origin/}" >/dev/null 2>&1 || true
+    else
+        git fetch --no-tags --prune --depth=50 origin "${DIFF_BASE#origin/}" >/dev/null 2>&1 || true
+    fi
+fi
```
francoisferrand: - instead of re-fecthing, best to set proper parameters in the workflow, during checkout
- "50" is completely arbitrary, and may not cover the whole range anyway... best to just pull the whole history (in the workflow)

package.json
francoisferrand: the script can be renamed as well

package.json
```diff
@@ -1,148 +1,152 @@
...
+        "format": "prettier --write .",
+        "format:check": "prettier --check .",
```
francoisferrand: do we really _all_ need these? (not many, but still quite redundant)

just a single `prettier` command is probably enough, and developer can pass additional flags to yarn : `yarn run prettier --check` ?

francoisferrand: and similarly the `--check` could be removed from the script, and extra args added to prettier call:
```
yarn run --silent prettier "${CHANGED[@]}" "$@"
```

so we have a generic `prettier:diff` command, which can be called either for reformatting or just checking:
```
yarn run prettier:diff --check
yarn run prettier:diff --format
```

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,51 @@
...
+name: lint
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name || github.sha }}
```
francoisferrand: `github.head_ref` and `github.ref_name` are _always_ set for pull_request, no point making things more complicated

```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref }}
```

.github/workflows/tests.yaml
```diff
@@ -74,85 +74,60 @@ permissions:
...
-    - name: Lint Javascript
-      run: yarn run --silent lint -- --max-warnings 0
-    - name: Lint Markdown
-      run: yarn run --silent lint_md
-    - name: Lint python
-      run: flake8 $(git ls-files "*.py")
-    - name: Lint Yaml
-      run: yamllint -c yamllint.yml $(git ls-files "*.yml")
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name || github.sha }}
```
francoisferrand: github.head_ref is never set on push, and github.ref_name is always set, no point making things more complicated


```suggestion
  group: ${{ github.workflow }}-${{ github.ref_name }}``

File: .github/workflows/tests.yaml [hunk 10]
```diff
@@ -209,8 +184,7 @@ jobs:
           push: true
           context: images/federation
           provenance: false
-          build-args:
-            CLOUDSERVER_VERSION=${{ github.sha }}
+          build-args: CLOUDSERVER_VERSION=${{ github.sha }}
```
francoisferrand: I think this was intentional, to clearly show each value

```suggestion
          build-args: |
            CLOUDSERVER_VERSION=${{ github.sha }}
```

scripts/format-changed.sh
```diff
@@ -0,0 +1,57 @@
...
+# 3) Listing changed files between base..HEAD, keeping only supported extensions.
+# 4) Running `prettier --check` on that filtered list.
+# 5) Exiting 0 when no supported files changed.
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT_DIR"
+
+BASE_SHA="${BASE_SHA:-}"
+BASE_REF="${BASE_REF:-${GITHUB_BASE_REF:-}}"
+DEFAULT_BASE="${DEFAULT_BASE:-origin/main}"
```
maeldonn: `main` is not an existing branch in this repository

.github/workflows/tests.yaml
benzekrimaha: I’ve reduced this to a minimal, targeted change in the workflow (no broad formatting pass).
In tests.yaml , only the intended lines were kept:
- concurrency group simplification
- build-args switched to multiline block format
- No unrelated reformatting remains.

.github/workflows/lint.yaml
francoisferrand, line 1: do not use a separate workflow : this just creates noise (more runs in actions, bert-e not properly displaying the build status...) and bugs (inconsistent triggers or concurrency).

if you really want to split the file, just make the "main" workflow invoke the other:
* use workflow_call trigger
  ```
  on:
    workflow_call:
  ```
* create a job in the regular "tests" workflow which _uses_ that workflow
  ```
  lint:
    uses: ./.github/workflows/lint.yaml
    secrets: inherit 
  ```

scripts/prettier-diff.sh
```diff
@@ -0,0 +1,75 @@
...
+# Usage examples:
+# - scripts/prettier-diff.sh --check
+# - scripts/prettier-diff.sh --format
+# - scripts/prettier-diff.sh --check --log-level warn
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT_DIR"
+
+BASE_SHA="${BASE_SHA:-}"
+BASE_REF="${BASE_REF:-${GITHUB_BASE_REF:-}}"
```
francoisferrand, line 17: In CI, BASE_REF is GITHUB_BASE_REF. No need for 2 variables, can use GITHUB_BASE_REF directly.

scripts/prettier-diff.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Purpose:
+# - Run Prettier on files changed in the current branch/PR.
```
francoisferrand, line 5: when running locally, there is a semantic question here: what do we mean by diff ?
(basically, this is the same question as `git diff`)

It can mean:
- the diff in the local workspace (easiest to implement and to reason with)
- the diff from the "upstream" branch (i.e. everything local or committed that is not in git upstream branch: which can be both the remote branch -what we push to- or the "parent" branch when stacking PRs)
- the diff from the development branch we based our work on and will eventually push to
- ...

To keep things simple, instead of choosing either or trying to automatically cover every case (and likely adding lot of complexity and more coupling) I would rather just rely on git diff (and allow passing params to git diff) : maximum flexibility and minimum work in the script...

scripts/prettier-diff.sh
```diff
@@ -0,0 +1,75 @@
...
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT_DIR"
+
+BASE_SHA="${BASE_SHA:-}"
+BASE_REF="${BASE_REF:-${GITHUB_BASE_REF:-}}"
+ORIGIN_HEAD="$(git symbolic-ref --quiet --short refs/remotes/origin/HEAD 2>/dev/null || true)"
+if [[ -n "$ORIGIN_HEAD" ]]; then
+	DEFAULT_BASE="${DEFAULT_BASE:-$ORIGIN_HEAD}"
+else
+	DEFAULT_BASE="${DEFAULT_BASE:-origin/development/9.3}"
```
francoisferrand, line 22: do not hard-code a branch. If we can't figure it out, just bail out...

scripts/prettier-diff.sh
```diff
@@ -0,0 +1,75 @@
...
+	DIFF_BASE="$DEFAULT_BASE"
+fi
+
+if ! git rev-parse --verify "${DIFF_BASE}^{commit}" >/dev/null 2>&1; then
+	echo "::error::Base ref ${DIFF_BASE} not found; cannot run Prettier."
+	echo "::error::Ensure the workflow checks out full history (fetch-depth: 0)."
+	exit 1
+fi
+
+mapfile -t CHANGED < <(git diff --name-only "${DIFF_BASE}..HEAD" --diff-filter=ACMRT \
```
francoisferrand, line 64: I suggest removing the logic above, and just simplify this to something like `git diff --name-only "$@"`
(will need to use `shift` after processing our own params)

→ can be used for most use-case locally
→ can directly pass `${{ github.event.pull_request.base.sha }}..HEAD` from the CI workflow (less coupling)

CONTRIBUTING.md
```diff
@@ -15,3 +14,16 @@ yarn dev
...
+
+Prettier is enforced in CI for changed files, including `package.json`.
+Since `yarn add`/`yarn remove` may rewrite `package.json` with 2-space
+indentation, run formatting before pushing dependency changes:
+
+```bash
+yarn format
```
maeldonn: Align with package.json

maeldonn (approved):

.github/workflows/tests.yaml
```diff
@@ -79,8 +79,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  lint:
+    uses: ./.github/workflows/lint.yaml
+    with:
+      git-diff-args: ${{ github.event_name == 'push' && format('{0}..{1}', github.event.before, github.sha) || 'HEAD~1..HEAD' }}
```
francoisferrand: not correct : should run on the diff from "upstream" to now.

for exemple: if we push 1 commit, then another 1 → script must run on the whole diff (2 commits), not just the last one

.github/workflows/lint.yaml
```diff
@@ -32,11 +27,8 @@ jobs:
       - name: Prettier (changed files)
         shell: bash
         run: |
-          if [[ -n "${{ inputs.git-diff-args }}" ]]; then
-            yarn run --silent prettier:diff --check ${{ inputs.git-diff-args }}
-          else
-            yarn run --silent prettier:diff --check
-          fi
+          MERGE_BASE=$(git merge-base HEAD origin/HEAD)
```
francoisferrand: `merge-base` should work, but not sure about origin/HEAD : may be ok if the latest development branch is indeed the "default" one, but could fail otherwise.

maybe a slightly safer way would be to list all development branches, something like

```
MERGE_BASE=$(git merge-base HEAD $(git branch --list 'development/*'))
```

francoisferrand: (to be tested, not sure we have all branches locally or if we need to add the `origin/` prefix)

.github/workflows/lint.yaml
francoisferrand: not resolved, still uses outdated GitHub Actions (v4)

DarkIsDude (approved):

.github/workflows/lint.yaml
benzekrimaha: Good point. I initially used `@v4` to stay aligned with the existing workflows and avoid introducing a different Actions major only in this new workflow.

That said, I agree we should not add new old-version usages if we intend to upgrade. I’d rather handle this consistently: either keep `@v4` here for now to avoid workflow-version drift, or update all `actions/checkout` / `actions/setup-node` usages together in a dedicated follow-up.

francoisferrand: what is the risk of having different versions in separate workflows?
the workflows (and jobs!) are actually independent, there is no drift here.

* here you are introducing a new workflow using deprecated actions which will stop working in a few days or weeks : which is not a good practice, just like we should not **introduce** dependencies with CVEs we should not 
* since the actions are outdated and will stop working soon, there is also a very significant risk that the other actions are updated while this PR is in review (it could even have been fixed already!) - which would not create any conflict, and leave us with partially migrated actions while we think migration was done

Bottom line : really no reason to use outdated actions in this new workflow, please update the PR

benzekrimaha: Done, I updated the new lint workflow to use the latest action. Note that I kept the change scoped to this new workflow to avoid turning this PR into a broader CI action migration.

francoisferrand (approved):

PR #6126, opened 2026-04-01
CLDSRV-881: fix missing CORS headers on 403 responses

tcarmet:
When a bucket has CORS configured, cloudserver was omitting `Access-Control-*` headers from 403 responses. Browsers treat such responses as opaque and block the error from the web app, even though AWS S3 returns CORS headers on error responses when a matching rule exists.

The existing `collectCorsHeaders` helper only runs inside an API handler, after bucket metadata has been loaded. Auth failures from Vault return before any handler runs, so the route-level error response path sent no CORS headers.

The fix wraps `callApiMethod`'s callback: on error, when the request has an `Origin` header and a known bucket, it fetches the bucket's CORS config and sets the matching `Access-Control-*` headers directly on the HTTP response before propagating the error. The extra metadata lookup only happens on failed requests with an `Origin` header - no cost for non-CORS traffic or successful requests.

Fixes: [CLDSRV-881](https://scality.atlassian.net/browse/CLDSRV-881)

[CLDSRV-881]: https://scality.atlassian.net/browse/CLDSRV-881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -81,6 +81,8 @@ const { isRequesterASessionUser } = require('./apiUtils/authorization/permission
 const checkHttpHeadersSize = require('./apiUtils/object/checkHttpHeadersSize');
 const constants = require('../../constants');
 const { config } = require('../Config.js');
+const metadata = require('../metadata/wrapper');
+const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const { validateMethodChecksumNoChunking } = require('./apiUtils/integrity/validateChecksums');
 const {
     requestNeedsRateCheck,
@@ -92,6 +94,91 @@ const {
 const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
 
 auth.setHandler(vault);
+
+// Detect CORS headers the handler already attached to its callback args.
+// Callback arity varies per route ((err, corsHeaders), (err, xml,
+// corsHeaders), (err, dataGetInfo, resMetaHeaders, range), ...) so we
+// scan the args instead of relying on a fixed position.
+function hasCorsHeaders(callbackArgs) {
+    for (const arg of callbackArgs) {
+        if (!arg || typeof arg !== 'object' || Array.isArray(arg)
+            || Buffer.isBuffer(arg)) {
+            continue;
+        }
+        if ('access-control-allow-origin' in arg) {
+            return true;
+        }
+    }
+    return false;
+}
+
+// Wrap the API callback so that, on error, we look up the bucket's CORS
+// configuration and set the matching Access-Control-* headers directly on
+// the HTTP response. This is needed because auth / pre-handler errors
+// return before the API handler retrieves the bucket (where the existing
+// collectCorsHeaders call lives), so the route-level error response path
+// otherwise sends no CORS headers - breaking cross-origin browser clients.
+// We only pay the extra metadata lookup when an Origin header is present,
+// matching AWS behavior and the existing contract of collectCorsHeaders.
+//
+// On the error path of a single request, this wrapper produces at most one
+// extra metadata.getBucket call - the fallback below. It only fires when:
+//   1. The handler ran and called metadata.getBucket itself (count = 1), AND
+//   2. It returned without surfacing CORS headers in its callback args
+//      (e.g. handler dropped the loaded bucket between waterfall steps,
+//      or the request did not match any CORS rule and the result was {}).
+// Combined with the handler's own call, total getBucket calls per failing
+// request stay capped at 2. Any future optimization must preserve or
+// improve that ceiling; the unit test in tests/unit/api/corsErrorHeaders.js
+// asserts it.
+function wrapCallbackWithErrorCorsHeaders(callback, request, response, log) {
+    function applyHeadersFromBucket(bucket) {
+        if (!bucket || response.headersSent) {
+            return;
+        }
+        const headers = collectCorsHeaders(
+            request.headers.origin, request.method, bucket);
+        Object.keys(headers).forEach(key => {
+            try {
+                response.setHeader(key, headers[key]);
+            } catch (e) {
+                log.debug('could not set CORS header on error', {
+                    header: key, error: e.message,
+                });
+            }
+        });
+    }
+
+    return (err, ...callbackArgs) => {
+        if (!err || !request.headers || !request.headers.origin
+            || !request.bucketName) {
+            return callback(err, ...callbackArgs);
+        }
+        // Fast path: most post-auth failures come back with corsHeaders
+        // already computed by the handler. The route will forward them
+        // to responseXMLBody/responseNoBody, so no extra lookup is needed.
+        if (hasCorsHeaders(callbackArgs)) {
+            return callback(err, ...callbackArgs);
+        }
+        // Fallback: either the bucket was never loaded (pre-handler
+        // failure like auth.server.doAuth denial, header-size check,
+        // copy-source parse error) or the handler returned without
+        // surfacing CORS headers. Re-fetch the bucket using the
+        // request's bucketName (the destination on copy operations) so
+        // we evaluate CORS against the right bucket.
+        return metadata.getBucket(request.bucketName, log, (mdErr, bucket) => {
+            if (mdErr) {
+                log.warn('could not fetch bucket CORS config for error response', {
+                    bucketName: request.bucketName,
+                    error: mdErr.code || mdErr.message,
+                });
+                return callback(err, ...callbackArgs);
+            }
+            applyHeadersFromBucket(bucket);
+            return callback(err, ...callbackArgs);
+        });
+    };
+}
 
 function checkAuthResults(authResults, apiMethod, log) {
     let returnTagCount = true;
@@ -346,6 +433,8 @@ const api = {
     callApiMethod(apiMethod, request, response, log, callback) {
         // Attach the apiMethod method to the request, so it can used by monitoring in the server
         request.apiMethod = apiMethod;
+        callback = wrapCallbackWithErrorCorsHeaders(
+            callback, request, response, log);
         // Array of end of API callbacks, used to perform some logic
         // at the end of an API.
         request.finalizerHooks = [];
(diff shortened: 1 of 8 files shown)


lib/api/api.js
```diff
@@ -92,6 +94,42 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+// otherwise sends no CORS headers - breaking cross-origin browser clients.
+// We only pay the extra metadata lookup when an Origin header is present,
+// matching AWS behavior and the existing contract of collectCorsHeaders.
+function wrapCallbackWithErrorCorsHeaders(request, response, log, callback) {
+    return (err, ...rest) => {
+        if (!err || !request.headers || !request.headers.origin
+            || !request.bucketName) {
+            return callback(err, ...rest);
+        }
+        return metadata.getBucket(request.bucketName, log, (mdErr, bucket) => {
```
dvasilas, line 169: Are there any cases in which api handlers pass `corsHeaders` to the callback on error (when they have already done the bucket lookup), and we can use them to avoid the `getBucket` ?

lib/api/api.js
```diff
@@ -92,6 +94,42 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+            if (!mdErr && bucket && !response.headersSent) {
```
dvasilas: We could maybe log a `warn` in case there is a `mdErr`.

lib/api/api.js
tcarmet: There are, good catch. The only annoying part is that the parameter placement in which they can be returned is a bit inconsistent but I've added a a little helper to help us identify the cases where there can be one. Also added a extra testcase for it.

lib/api/api.js
tcarmet: done

lib/api/api.js
tcarmet: Also your suggestion found a couple of pre-existing bug on `bucketGetCors` and `bucketGetWebsite` so nice :ok_hand:

lib/api/bucketGetCors.js
```diff
@@ -29,10 +29,7 @@ function bucketGetCors(authInfo, request, log, callback) {
         const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
         if (err) {
             monitoring.promMetrics('GET', bucketName, err.code, METRICS_ACTION);
-            if (err?.is?.AccessDenied) {
-                return callback(err, corsHeaders);
-            }
-            return callback(err);
+            return callback(err, null, corsHeaders);
```
dvasilas, line 32: The same issue seems to exist here: https://github.com/scality/cloudserver/blob/bb5ca88d45aec3e311f49a44978f38227fa8a2c2/lib/api/bucketGetLocation.js#L33

lib/api/api.js
```diff
@@ -92,6 +94,77 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+// Wrap the API callback so that, on error, we look up the bucket's CORS
+// configuration and set the matching Access-Control-* headers directly on
+// the HTTP response. This is needed because auth / pre-handler errors
+// return before the API handler retrieves the bucket (where the existing
+// collectCorsHeaders call lives), so the route-level error response path
+// otherwise sends no CORS headers - breaking cross-origin browser clients.
+// We only pay the extra metadata lookup when an Origin header is present,
+// matching AWS behavior and the existing contract of collectCorsHeaders.
+function wrapCallbackWithErrorCorsHeaders(request, response, log, callback) {
+    return (err, ...rest) => {
```
dvasilas: If the handler computes `corsHeaders` and gets back `{}` (bucket has no CORS config), and then calls `callback(err, null, {})`, the wrapper will still do `metadata.getBucket`, right? (but it does not need to, I think).

Not sure if it's a good idea, but we could cache the bucket in `request` (something like  `request._loadedBucket`) and first check there before `metadata.getBucket`

lib/api/bucketGetCors.js
tcarmet: good catch, thanks. There are other calls where corsHeaders is in the same position but on those cases the routing is correct. I think this should be the last one in the wrong position.

lib/api/api.js
tcarmet: Was able to confirm this with a extra test, indeed there would be 2 metadata call in this case. 

So yeah we could consider `{}` as there's no header, and skip the bucket call in this case, it seems to be implicitly true, or we cache the bucket as you stated.

I'm debating between both right now, I pushed a code with the bucket cache on `request`. seems like the safest option between the two. Let me know what you think.

dvasilas: Something I didn't consider when I suggested the cache approach: I think `objectCopy` and `objectPutCopyPart` call `standardMetadataValidateBucketAndObj` twice (once for the destination, once for the source), so the second call will overwrite the first, and we may not get the headers for the bucket we want.
You could confirm with a test.

tcarmet: yes confirmed with the tests, nice catch again. At the end I ended up droping that cache entirely and accepting up to 2 `metadata.getBucket` calls on the error path. I believe the trade-off is fine here because:

- This only fires on failing requests with an Origin header where the handler returned without surfacing CORS headers (no rule match, no CORS config, or handler dropped the bucket between waterfall steps. Small slice of overall traffic.
- Turns out going the other way and trusting `{}` as "no CORS to apply" would have regressed handlers that drop the bucket internally their collectCorsHeaders runs with bucket=undefined and returns `{}` even when a matching CORS rule exists. We'd silently lose CORS headers we currently apply.
- A smarter cache (map keyed by `bucketName` so `objectCopy` doesn't overwrite the destination with the source) would have preserved the optimization without the leak you flagged. So **that is an option** we could arguably go this way, but I honestly felt like it was a over-optimization for an error-path issue, so I went with the simpler code.

To sum things up I removed the cache, documented the behavior in case we decide  to optimize in the future and added/adapted the test cases (objectCopy and limiting number of calls to getBucket) so that if we do decide to mess with it, we ensure what we discussed is not lost and will fail if we handle it badly.

dvasilas: That's good with me. Thanks for taking the time to look at the optimization.

dvasilas (approved):

lib/api/api.js
```diff
@@ -92,6 +94,95 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+// asserts it.
+function wrapCallbackWithErrorCorsHeaders(request, response, log, callback) {
+    function applyHeadersFromBucket(bucket) {
+        if (!bucket || response.headersSent) {
+            return;
+        }
+        const headers = collectCorsHeaders(
+            request.headers.origin, request.method, bucket);
+        Object.keys(headers).forEach(key => {
+            if (headers[key] === undefined) {
```
jonathan-gramain: does `collectCorsHeaders` return undefined values? I would tend to think that it would be the responsibility of this function to return well formed headers.

lib/api/api.js
```diff
@@ -92,6 +94,95 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
 
 auth.setHandler(vault);
 
+// Detect CORS headers the handler already attached to its callback args.
+// Callback arity varies per route ((err, corsHeaders), (err, xml,
+// corsHeaders), (err, dataGetInfo, resMetaHeaders, range), ...) so we
+// scan the args instead of relying on a fixed position.
+function hasCorsHeaders(rest) {
```
jonathan-gramain: nitpick: not fond of `rest` as a name, what about `callbackArgs`.

Also wondering if there's a reason to check every argument, i.e. if the CORS headers shouldn't always be returned as the third callback argument as a convention?

lib/api/api.js
```diff
@@ -92,6 +94,95 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+        const headers = collectCorsHeaders(
+            request.headers.origin, request.method, bucket);
+        Object.keys(headers).forEach(key => {
+            if (headers[key] === undefined) {
+                return;
+            }
+            try {
+                response.setHeader(key, headers[key]);
+            } catch (e) {
+                log.debug('could not set cors header on error', {
```
jonathan-gramain: nitpick: other messages spell `CORS` with upper-case letters, you may want to use the same case for consistency (I think upper-case is most appropriate)

lib/api/api.js
```diff
@@ -92,6 +94,95 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+        // Fallback: either the bucket was never loaded (pre-handler
+        // failure like auth.server.doAuth denial, header-size check,
+        // copy-source parse error) or the handler returned without
+        // surfacing CORS headers. Re-fetch the bucket using the
+        // request's bucketName (the destination on copy operations) so
+        // we evaluate CORS against the right bucket.
+        return metadata.getBucket(request.bucketName, log, (mdErr, bucket) => {
+            if (mdErr) {
+                log.warn('could not fetch bucket CORS config for error '
+                    + 'response', {
```
jonathan-gramain: I though we accepted 120 columns for a while now, not sure but not wrapping the message would be more appropriate

```suggestion
                log.warn('could not fetch bucket CORS config for error response', {
```

lib/api/api.js
```diff
@@ -92,6 +94,95 @@ const monitoringMap = policies.actionMaps.actionMonitoringMapS3;
...
+// extra metadata.getBucket call - the fallback below. It only fires when:
+//   1. The handler ran and called metadata.getBucket itself (count = 1), AND
+//   2. It returned without surfacing CORS headers in its callback args
+//      (e.g. handler dropped the loaded bucket between waterfall steps,
+//      or the request did not match any CORS rule and the result was {}).
+// Combined with the handler's own call, total getBucket calls per failing
+// request stay capped at 2. Any future optimization must preserve or
+// improve that ceiling; the unit test in tests/unit/api/corsErrorHeaders.js
+// asserts it.
+function wrapCallbackWithErrorCorsHeaders(request, response, log, callback) {
```
jonathan-gramain: nitpick: since the function is a wrapper, it could be more appropriate to pass the callback object to wrap as first argument.

lib/api/api.js
tcarmet: There's no clear separation of the CORS headers, it's not always returned as the third callback, it's mix of the second or third depending on the routing it takes. (I added a comment already stating, maybe it wasn't clear. let me know)

Tackling a standard position would be a refactor of it's own IMHO not in scope right now.

jonathan-gramain (approved):

PR #6142, opened 2026-04-07
CLDSRV-774: Add aclRequired field to server access logs

dvasilas:
Add the `aclRequired` field to access logs.

From https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html#log-record-fields
> indicates whether the request required an access control list (ACL) for authorization. If the request required an ACL for authorization, the string is Yes. If no ACLs were required, the string is -.

### DIFF
--- a/lib/api/apiUtils/authorization/permissionChecks.js
+++ b/lib/api/apiUtils/authorization/permissionChecks.js
@@ -509,8 +509,10 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
     request, aclPermission, results, actionImplicitDenies) {
     const bucketPolicy = bucket.getBucketPolicy();
     let processedResult = results[requestType];
+    let aclRequired = false;
     if (!bucketPolicy || request?.bypassUserBucketPolicies) {
         processedResult = actionImplicitDenies[requestType] === false && aclPermission;
+        aclRequired = true;
     } else {
         const bucketPolicyPermission = checkBucketPolicy(bucketPolicy, requestType, canonicalID, arn,
             bucketOwner, log, request, actionImplicitDenies);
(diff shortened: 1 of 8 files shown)


leif-scality (approved):

lib/api/apiUtils/authorization/permissionChecks.js
```diff
@@ -525,9 +527,10 @@ function processBucketPolicy(requestType, bucket, canonicalID, arn, bucketOwner,
             processedResult = true;
         } else {
             processedResult = actionImplicitDenies[requestType] === false && aclPermission;
+            aclRequired = true;
         }
     }
-    return processedResult;
+    return { allowed: processedResult, aclRequired };
```
BourgoisMickael, line 533: you could update the `request.serverAccessLog.aclRequired` immediately in that function instead of returning the value

lib/metadata/metadataUtils.js
```diff
@@ -422,6 +424,12 @@ function standardMetadataValidateBucketAndObj(params, actionImplicitDenies, log,
         },
     ], (err, bucket, objMD, raftSessionId) => {
         storeServerAccessLogInfo(request, bucket, raftSessionId, params.serverAccessLogOptions);
+        if (isCopySource && request?.sourceServerAccessLog) {
```
BourgoisMickael: Should this logic rather be in `storeServerAccessLogInfo` ?

BourgoisMickael (approved):

lib/api/apiUtils/authorization/permissionChecks.js
dvasilas: I didn't do it like that because there is 1 caller of this function that never uses ACLs, and we should skip `aclRequired` there.

lib/metadata/metadataUtils.js
dvasilas: Yes, that's a bit better, done, thanks

dvasilas:
ping

PR #6153, opened 2026-04-16
CLDSRV-894: Report turn-around time for PutObject and UploadPart

dvasilas:
Turn-around time was initially left unreported for PutObject and UploadPart because these operations stream the request body to the storage backend as it arrives. For large objects, receiving and processing overlap, so the measured turn-around time only captures the tail-end confirmation latency rather than the full processing duration. We decided this was misleading and chose to report '-' instead.

However, testing against AWS shows that it does report turn-around time for these operations regardless of object size. This PR aligns our behavior with AWS.

### DIFF
--- a/lib/api/api.js
+++ b/lib/api/api.js
@@ -263,6 +263,19 @@ const api = {
                     request.serverAccessLog.analyticsUserName    = authNames.userName;
                 }
                 if (apiMethod === 'objectPut' || apiMethod === 'objectPutPart') {
+                    const setStartTurnAroundTime = () => {
+                        if (request.serverAccessLog) {
+                            request.serverAccessLog.startTurnAroundTime = process.hrtime.bigint();
+                        }
+                    };
+                    // For 0-byte uploads, downstream handlers do not consume
+                    // the request stream, so 'end' never fires. Set
+                    // startTurnAroundTime synchronously in that case.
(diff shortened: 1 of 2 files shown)


BourgoisMickael (approved):

lib/api/api.js
```diff
@@ -263,6 +263,19 @@ const api = {
                     request.serverAccessLog.analyticsUserName    = authNames.userName;
                 }
                 if (apiMethod === 'objectPut' || apiMethod === 'objectPutPart') {
+                    const setStartTurnAroundTime = () => {
+                        if (request.serverAccessLog) {
+                            request.serverAccessLog.startTurnAroundTime = process.hrtime.bigint();
+                        }
+                    };
+                    // For 0-byte uploads, downstream handlers do not consume
+                    // the request stream, so 'end' never fires. Set
```
jonathan-gramain, line 272: Strange, sounds like a bug at first that the request stream is not consumed, do you happen to know if there's a good reason to be so?

jonathan-gramain (approved):

lib/api/api.js
dvasilas: I think it's an optimization: https://github.com/scality/cloudserver/blob/029b436f36c786a4c74de7098c44750d795a9216/lib/api/apiUtils/object/createAndStoreObject.js#L217

PR #6160, opened 2026-04-23
CLDSRV-895: ListParts return part checksum

leif-scality:
- Return part checksum in ListParts

### DIFF
--- a/lib/api/listParts.js
+++ b/lib/api/listParts.js
@@ -14,6 +14,7 @@ const { pushMetric } = require('../utapi/utilities');
 const monitoring = require('../utilities/monitoringHandler');
 const { data } = require('../data/wrapper');
 const { setExpirationHeaders } = require('./apiUtils/object/expirationHeaders');
+const { algorithms } = require('./apiUtils/integrity/validateChecksums');
 
 /*
   Format of xml response:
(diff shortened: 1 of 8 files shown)


lib/api/listParts.js
```diff
@@ -64,6 +65,40 @@ function buildXML(xmlParams, xml, encodingFn) {
...
+        checksumAlgorithm: item.value.ChecksumAlgorithm,
+        checksumValue: item.value.ChecksumValue,
+    };
+}
+
+function getPartChecksumXML(checksumAlgorithm, checksumValue) {
+    if (!checksumAlgorithm || checksumValue === undefined) {
+        return undefined;
+    }
+    const algorithm = String(checksumAlgorithm).toLowerCase();
```
BourgoisMickael: Why does this need `String` convertion ? Which other type could it be ?

BourgoisMickael (approved):

lib/api/listParts.js
leif-scality: removed

lib/api/listParts.js
```diff
@@ -64,6 +65,40 @@ function buildXML(xmlParams, xml, encodingFn) {
...
+
+function getPartChecksum(item) {
+    return {
+        checksumAlgorithm: item.value.ChecksumAlgorithm,
+        checksumValue: item.value.ChecksumValue,
+    };
+}
+
+function getPartChecksumXML(checksumAlgorithm, checksumValue) {
+    if (!checksumAlgorithm || checksumValue === undefined) {
```
dvasilas:
```suggestion
    if (!checksumAlgorithm || !checksumValue) {
```

?

lib/api/listParts.js
```diff
@@ -271,13 +305,19 @@ function listParts(authInfo, request, log, callback) {
             ], xml, encodingFn);
 
             partListing.forEach(part => {
+                const partChecksumXML = mpuOverviewObj.checksumIsDefault ?
+                    undefined : getPartChecksumXML(
```
dvasilas: Should we also check if `(mpuOverviewObj.checksumAlgorithm && mpuOverviewObj.checksumType`) ?

lib/api/listParts.js
leif-scality: fixed

lib/api/listParts.js
leif-scality: fiexd

dvasilas (approved):

PR #6170, opened 2026-05-15
CLDSRV-898: handle checksums in CompleteMultipartUpload

leif-scality:
- Calculate and compare the final object checksum with the one sent by the headers
- Check that all parts have the correct checksum and checksum type
- stores the final checksum when FULL_OBJECT (COMPOSITE are going to be stored by https://scality.atlassian.net/browse/S3C-10399)
- lint `tests/unit/api/multipartUpload.js`

### DIFF
--- a/lib/api/apiUtils/integrity/validateChecksums.js
+++ b/lib/api/apiUtils/integrity/validateChecksums.js
@@ -36,7 +36,7 @@ const errMPUTypeInvalid = errorInstances.InvalidRequest.customizeDescription(
     'Value for x-amz-checksum-type header is invalid.',
 );
 const errMPUTypeWithoutAlgo = errorInstances.InvalidRequest.customizeDescription(
-    'The x-amz-checksum-type header can only be used ' + 'with the x-amz-checksum-algorithm header.',
+    'The x-amz-checksum-type header can only be used with the x-amz-checksum-algorithm header.',
 );
 
 // Methods that validate BOTH Content-MD5 and x-amz-checksum-* against the
--- a/lib/api/apiUtils/integrity/validateChecksums.js
+++ b/lib/api/apiUtils/integrity/validateChecksums.js
@@ -164,6 +164,20 @@ const algorithms = Object.freeze({
 });
 
 /**
+ * List per-algorithm `x-amz-checksum-<algo>` value headers, excluding the
+ * `x-amz-checksum-type` and `x-amz-checksum-algorithm` configuration
+ * headers.
+ *
+ * @param {object} headers - HTTP request headers (lowercased keys)
+ * @returns {string[]} matching header names
+ */
+function listChecksumValueHeaders(headers) {
+    return Object.keys(headers).filter(
+        h => h.startsWith('x-amz-checksum-') && h !== 'x-amz-checksum-type' && h !== 'x-amz-checksum-algorithm',
+    );
+}
+
+/**
  * Validate the x-amz-checksum-<algo> header against a buffered body.
  *
  * @param {object} headers - HTTP request headers (lowercased keys)
@@ -172,7 +186,7 @@ const algorithms = Object.freeze({
  *   null on success; otherwise a ChecksumError with details.
  */
 async function validateXAmzChecksums(headers, body) {
-    const checksumHeaders = Object.keys(headers).filter(header => header.startsWith('x-amz-checksum-'));
+    const checksumHeaders = listChecksumValueHeaders(headers);
     const xAmzChecksumCnt = checksumHeaders.length;
     if (xAmzChecksumCnt > 1) {
         return { error: ChecksumError.MultipleChecksumTypes, details: { algorithms: checksumHeaders } };
@@ -256,7 +270,7 @@ function getChecksumDataFromHeaders(headers) {
         return null;
     };
 
-    const checksumHeaders = Object.keys(headers).filter(header => header.startsWith('x-amz-checksum-'));
+    const checksumHeaders = listChecksumValueHeaders(headers);
     const xAmzChecksumCnt = checksumHeaders.length;
     if (xAmzChecksumCnt > 1) {
         return { error: ChecksumError.MultipleChecksumTypes, details: { algorithms: checksumHeaders } };
@@ -441,9 +455,7 @@ function arsenalErrorFromChecksumError(err) {
 function validateCompleteMultipartUploadChecksum(headers, finalChecksum) {
     // x-amz-checksum-type can be present in a completeMPU request so we drop it.
     // x-amz-checksum-algorithm is ignored by AWS in this context so we drop it.
-    const valueHeaders = Object.keys(headers).filter(
-        h => h.startsWith('x-amz-checksum-') && h !== 'x-amz-checksum-type' && h !== 'x-amz-checksum-algorithm',
-    );
+    const valueHeaders = listChecksumValueHeaders(headers);
 
     if (valueHeaders.length === 0) {
         return null;
--- a/lib/api/apiUtils/integrity/validateChecksums.js
+++ b/lib/api/apiUtils/integrity/validateChecksums.js
@@ -39,10 +39,9 @@ const errMPUTypeWithoutAlgo = errorInstances.InvalidRequest.customizeDescription
     'The x-amz-checksum-type header can only be used ' + 'with the x-amz-checksum-algorithm header.',
 );
 
+// Methods that validate BOTH Content-MD5 and x-amz-checksum-* against the
+// buffered request body. For these, x-amz-checksum-* is the body digest.
 const checksumedMethods = Object.freeze({
-    // CompleteMPU's x-amz-checksum-<algo> is the final-object checksum,
-    // not a body digest. Validated in completeMultipartUpload.js instead.
-    // 'completeMultipartUpload': true,
     multiObjectDelete: true,
     bucketPutACL: true,
     bucketPutCors: true,
@@ -61,6 +60,15 @@ const checksumedMethods = Object.freeze({
     objectPutRetention: true,
     objectPutTagging: true,
     objectRestore: true,
+});
+
+// Methods that validate Content-MD5 against the buffered request body but
+// do NOT treat x-amz-checksum-* as a body digest. CompleteMPU's
+// x-amz-checksum-<algo> is the asserted final-object checksum, not a
+// body digest, and is validated separately by
+// validateCompleteMultipartUploadChecksum.
+const md5OnlyMethods = Object.freeze({
+    completeMultipartUpload: true,
 });
 
 const ChecksumError = Object.freeze({
@@ -313,10 +321,37 @@ function getChecksumDataFromHeaders(headers) {
 }
 
 /**
+ * validateContentMd5 - Validate the Content-MD5 header against the
+ * buffered body.
+ * @param {object} headers - http headers
+ * @param {Buffer} body - http request body
+ * @return {object|null} - null on success; otherwise a ChecksumError with details.
+ */
+function validateContentMd5(headers, body) {
+    if (!('content-md5' in headers)) {
+        return { error: ChecksumError.MissingChecksum, details: null };
+    }
+    if (typeof headers['content-md5'] !== 'string') {
+        return { error: ChecksumError.MD5Invalid, details: { expected: headers['content-md5'] } };
+    }
+    if (headers['content-md5'].length !== 24) {
+        return { error: ChecksumError.MD5Invalid, details: { expected: headers['content-md5'] } };
+    }
+    if (!base64Regex.test(headers['content-md5'])) {
+        return { error: ChecksumError.MD5Invalid, details: { expected: headers['content-md5'] } };
+    }
+    const md5 = crypto.createHash('md5').update(body).digest('base64');
+    if (md5 !== headers['content-md5']) {
+        return { error: ChecksumError.MD5Mismatch, details: { calculated: md5, expected: headers['content-md5'] } };
+    }
+    return null;
+}
+
+/**
  * validateChecksumsNoChunking - Validate the checksums of a request.
  * @param {object} headers - http headers
  * @param {Buffer} body - http request body
- * @return {object} - error
+ * @return {Promise<object|null>} - null on success; otherwise a ChecksumError with details.
  */
 async function validateChecksumsNoChunking(headers, body) {
     if (!headers) {
@@ -325,23 +360,10 @@ async function validateChecksumsNoChunking(headers, body) {
 
     let md5Present = false;
     if ('content-md5' in headers) {
-        if (typeof headers['content-md5'] !== 'string') {
-            return { error: ChecksumError.MD5Invalid, details: { expected: headers['content-md5'] } };
+        const md5Err = validateContentMd5(headers, body);
+        if (md5Err) {
+            return md5Err;
         }
-
-        if (headers['content-md5'].length !== 24) {
-            return { error: ChecksumError.MD5Invalid, details: { expected: headers['content-md5'] } };
-        }
-
-        if (!base64Regex.test(headers['content-md5'])) {
-            return { error: ChecksumError.MD5Invalid, details: { expected: headers['content-md5'] } };
-        }
-
-        const md5 = crypto.createHash('md5').update(body).digest('base64');
-        if (md5 !== headers['content-md5']) {
-            return { error: ChecksumError.MD5Mismatch, details: { calculated: md5, expected: headers['content-md5'] } };
-        }
-
         md5Present = true;
     }
 
@@ -408,6 +430,65 @@ function arsenalErrorFromChecksumError(err) {
     }
 }
 
+/**
+ * Validate the optional x-amz-checksum-<algo> header on a CompleteMPU
+ * request against the computed final-object checksum.
+ *
+ * @param {object} headers - request.headers (lowercased keys)
+ * @param {object|null} finalChecksum - { algorithm, type, value } or null
+ * @returns {{error: string, details: object}|null}
+ */
+function validateCompleteMultipartUploadChecksum(headers, finalChecksum) {
+    // x-amz-checksum-type can be present in a completeMPU request so we drop it.
+    // x-amz-checksum-algorithm is ignored by AWS in this context so we drop it.
+    const valueHeaders = Object.keys(headers).filter(
+        h => h.startsWith('x-amz-checksum-') && h !== 'x-amz-checksum-type' && h !== 'x-amz-checksum-algorithm',
+    );
+
+    if (valueHeaders.length === 0) {
+        return null;
+    }
+
+    if (valueHeaders.length > 1) {
+        return { error: ChecksumError.MultipleChecksumTypes, details: { algorithms: valueHeaders } };
+    }
+
+    const headerName = valueHeaders[0];
+    const foundAlgo = headerName.slice('x-amz-checksum-'.length);
+    const foundValue = headers[headerName];
+
+    if (!(foundAlgo in algorithms)) {
+        return { error: ChecksumError.AlgoNotSupported, details: { algorithm: foundAlgo } };
+    }
+
+    const dashIdx = foundValue.lastIndexOf('-');
+    const hasCompositeSuffix = dashIdx > 0 && /^\d+$/.test(foundValue.slice(dashIdx + 1));
+    if (hasCompositeSuffix && !compositeAlgorithms.has(foundAlgo)) {
+        return { error: ChecksumError.MalformedChecksum, details: { algorithm: foundAlgo, expected: foundValue } };
+    }
+    if (!hasCompositeSuffix && !fullObjectAlgorithms.has(foundAlgo)) {
+        return { error: ChecksumError.MalformedChecksum, details: { algorithm: foundAlgo, expected: foundValue } };
+    }
+    const digestPart = hasCompositeSuffix ? foundValue.slice(0, dashIdx) : foundValue;
+    if (!algorithms[foundAlgo].isValidDigest(digestPart)) {
+        return { error: ChecksumError.MalformedChecksum, details: { algorithm: foundAlgo, expected: foundValue } };
+    }
+
+    if (!finalChecksum || finalChecksum.algorithm !== foundAlgo || finalChecksum.value !== foundValue) {
+        return {
+            error: ChecksumError.XAmzMismatch,
+            details: {
+                algorithm: foundAlgo,
+                expected: foundValue,
+                computedAlgorithm: finalChecksum && finalChecksum.algorithm,
+                computed: finalChecksum && finalChecksum.value,
+            },
+        };
+    }
+
+    return null;
+}
+
 async function defaultValidationFunc(request, body, log) {
     const err = await validateChecksumsNoChunking(request.headers, body);
     if (!err) {
@@ -418,6 +499,22 @@ async function defaultValidationFunc(request, body, log) {
         log.debug('failed checksum validation', { method: request.apiMethod }, err);
     }
 
+    return arsenalErrorFromChecksumError(err);
+}
+
+// Validate ONLY Content-MD5 against the buffered body. Used for methods
+// where x-amz-checksum-* has a non-body meaning (e.g. CompleteMPU's
+// final-object checksum assertion) and must not be re-validated here.
+function md5OnlyValidationFunc(request, body, log) {
+    // Content-MD5 is optional. If absent, nothing to validate.
+    if (!('content-md5' in request.headers)) {
+        return null;
+    }
+    const err = validateContentMd5(request.headers, body);
+    if (!err) {
+        return null;
+    }
+    log.debug('failed Content-MD5 validation', { method: request.apiMethod }, err);
     return arsenalErrorFromChecksumError(err);
 }
 
@@ -435,6 +532,10 @@ async function validateMethodChecksumNoChunking(request, body, log) {
 
     if (request.apiMethod in checksumedMethods) {
         return await defaultValidationFunc(request, body, log);
+    }
+
+    if (request.apiMethod in md5OnlyMethods) {
+        return md5OnlyValidationFunc(request, body, log);
     }
 
     return null;
@@ -551,13 +652,13 @@ const COMPOSITE_ALGOS = new Set(['crc32', 'crc32c', 'sha1', 'sha256']);
  * @returns {{ checksum: string, error: null }
  *   | { checksum: null, error: { code: string, details: object } }}
  */
-function computeCompositeMPUChecksum(algorithm, partChecksumsBase64) {
+async function computeCompositeMPUChecksum(algorithm, partChecksumsBase64) {
     if (!COMPOSITE_ALGOS.has(algorithm)) {
         return { checksum: null, error: { code: ChecksumError.MPUAlgoNotSupported, details: { algorithm } } };
     }
 
     const concat = Buffer.concat(partChecksumsBase64.map(c => Buffer.from(c, 'base64')));
-    const digest = algorithms[algorithm].digest(concat);
+    const digest = await algorithms[algorithm].digest(concat);
     return {
         checksum: `${digest}-${partChecksumsBase64.length}`,
         error: null,
@@ -599,4 +700,5 @@ module.exports = {
     getChecksumDataFromMPUHeaders,
     computeCompositeMPUChecksum,
     computeFullObjectMPUChecksum,
+    validateCompleteMultipartUploadChecksum,
 };
--- a/lib/api/apiUtils/integrity/validateChecksums.js
+++ b/lib/api/apiUtils/integrity/validateChecksums.js
@@ -29,8 +29,7 @@ const errTrailerNotSupported = errorInstances.InvalidRequest.customizeDescriptio
 );
 const errMPUAlgoNotSupported = errorInstances.InvalidRequest.customizeDescription(
     'Checksum algorithm provided is unsupported. ' +
-        'Please try again with any of the valid types: ' +
-        '[CRC32, CRC32C, CRC64NVME, SHA1, SHA256]',
+        'Please try again with any of the valid types: [CRC32, CRC32C, CRC64NVME, SHA1, SHA256]',
 );
 const errMPUTypeInvalid = errorInstances.InvalidRequest.customizeDescription(
     'Value for x-amz-checksum-type header is invalid.',
--- a/lib/api/apiUtils/integrity/validateChecksums.js
+++ b/lib/api/apiUtils/integrity/validateChecksums.js
@@ -40,7 +40,9 @@ const errMPUTypeWithoutAlgo = errorInstances.InvalidRequest.customizeDescription
 );
 
 const checksumedMethods = Object.freeze({
-    completeMultipartUpload: true,
+    // CompleteMPU's x-amz-checksum-<algo> is the final-object checksum,
+    // not a body digest. Validated in completeMultipartUpload.js instead.
+    // 'completeMultipartUpload': true,
     multiObjectDelete: true,
     bucketPutACL: true,
     bucketPutCors: true,
--- a/lib/api/completeMultipartUpload.js
+++ b/lib/api/completeMultipartUpload.js
@@ -26,7 +26,13 @@ const { setExpirationHeaders } = require('./apiUtils/object/expirationHeaders');
 const { validatePutVersionId } = require('./apiUtils/object/coldStorage');
 const { validateQuotas } = require('./apiUtils/quotas/quotaUtils');
 const { setSSEHeaders } = require('./apiUtils/object/sseHeaders');
-const { algorithms: checksumAlgorithms } = require('./apiUtils/integrity/validateChecksums');
+const {
+    algorithms: checksumAlgorithms,
+    arsenalErrorFromChecksumError,
+    computeCompositeMPUChecksum,
+    computeFullObjectMPUChecksum,
+    validateCompleteMultipartUploadChecksum,
+} = require('./apiUtils/integrity/validateChecksums');
 
 const versionIdUtils = versioning.VersionID;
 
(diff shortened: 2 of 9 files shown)


lib/api/completeMultipartUpload.js
```diff
@@ -26,12 +26,88 @@ const { setExpirationHeaders } = require('./apiUtils/object/expirationHeaders');
...
+
+    const storedByPartNumber = new Map();
+    storedParts.forEach(item => {
+        const partNumber = Number.parseInt(item.key.split(mpuSplitter)[1], 10);
+        storedByPartNumber.set(partNumber, item);
+    });
+
+    const parts = jsonList.Part || [];
+    for (let i = 0; i < parts.length; i++) {
+        const part = parts[i];
```
jonathan-gramain:
```suggestion
    for (const part of (jsonList.Part || [])) {
```

lib/api/completeMultipartUpload.js
```diff
@@ -26,12 +26,88 @@ const { setExpirationHeaders } = require('./apiUtils/object/expirationHeaders');
...
+        const partNumber = Number.parseInt(part.PartNumber[0], 10);
+
+        const presentTags = allChecksumXmlTags.filter(tag => part[tag]);
+
+        for (const tag of presentTags) {
+            if (tag !== expectedTag) {
+                const algoLabel = tag.replace(/^Checksum/, '').toLowerCase();
+                return errorInstances.BadDigest.customizeDescription(
+                    `The ${algoLabel} you specified for part ${partNumber} ` + 'did not match what we received.',
```
jonathan-gramain:
```suggestion
                    `The ${algoLabel} you specified for part ${partNumber} did not match what we received.`,
```

lib/api/completeMultipartUpload.js
```diff
@@ -26,12 +26,88 @@ const { setExpirationHeaders } = require('./apiUtils/object/expirationHeaders');
...
+                    'One or more of the specified parts could not be found.  ' +
+                        'The part may not have been uploaded, or the specified ' +
+                        "entity tag may not match the part's entity tag.",
+                );
+            }
+        } else if (requireForEachPart) {
+            return errorInstances.InvalidRequest.customizeDescription(
+                `The upload was created using a ${mpuAlgo} checksum. ` +
+                    'The complete request must include the checksum for each ' +
+                    `part. It was missing for part ${partNumber} in the request.`,
```
jonathan-gramain, line 120:
```suggestion
                    `part, it is missing for part ${partNumber} in the request.`,
```

tests/unit/api/completeMultipartUpload.js
jonathan-gramain, line 1: There's already a test file `multipartUpload.js`, wouldn't it be better to add the tests there? Or if a refactor into multiple files is beneficial, maybe port some related tests from there to this new file to avoid scattering the complete MPU tests into multiple files.

tests/unit/api/completeMultipartUpload.js
```diff
@@ -0,0 +1,502 @@
...
+const { bucketPut } = require('../../../lib/api/bucketPut');
+const initiateMultipartUpload = require('../../../lib/api/initiateMultipartUpload');
+const objectPutPart = require('../../../lib/api/objectPutPart');
+const completeMultipartUpload = require('../../../lib/api/completeMultipartUpload');
+const { validatePerPartChecksums } = completeMultipartUpload;
+const { validateMethodChecksumNoChunking } = require('../../../lib/api/apiUtils/integrity/validateChecksums');
+const DummyRequest = require('../DummyRequest');
+const { cleanup, DummyRequestLogger, makeAuthInfo } = require('../helpers');
+
+const SPLITTER = '..|..';
```
jonathan-gramain:
```suggestion
const SPLITTER = constants.splitter;
```

tests/unit/api/completeMultipartUpload.js
```diff
@@ -0,0 +1,502 @@
...
+const UPLOAD_ID = 'upload-id-1';
+
+// XML element name AWS uses for each algorithm in CompleteMultipartUpload's
+// per-part body.
+const TAG_BY_ALGO = {
```
jonathan-gramain: Shouldn't this be in constants too?

tests/unit/api/completeMultipartUpload.js
```diff
@@ -0,0 +1,502 @@
...
+                    };
+                    const err = validatePerPartChecksums(jsonList, stored, SPLITTER, mpuChecksum);
+                    assert(err);
+                    assert.strictEqual(err.is.InvalidPart, true);
+                    // AWS reuses its generic InvalidPart message — no algorithm
+                    // or part number in the wording.
+                    assert.strictEqual(
+                        err.description,
+                        'One or more of the specified parts could not be ' +
+                            'found.  The part may not have been uploaded, or ' +
```
jonathan-gramain: Is the double space intentional? Also not very fond of matching against such a long error message that may vary, but that's okay I think.

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -408,6 +425,86 @@ function arsenalErrorFromChecksumError(err) {
...
+ * @param {object|null} finalChecksum - { algorithm, type, value } or null
+ * @returns {{error: string, details: object}|null}
+ */
+function validateCompleteMultipartUploadChecksum(headers, finalChecksum) {
+    // `x-amz-checksum-type` and `x-amz-checksum-algorithm` are configuration
+    // headers (MPU completeness mode / SDK algorithm hint), not value
+    // headers. They must not count toward the "value header" tally.
+    const valueHeaders = Object.keys(headers).filter(
+        h => h.startsWith('x-amz-checksum-') && h !== 'x-amz-checksum-type' && h !== 'x-amz-checksum-algorithm',
+    );
```
jonathan-gramain: If the list of supported headers is known and a short list of supported algorithms, it may be cleaner to directly extract each of the possible ones, or otherwise filter like `[list-of-supported-headers].includes(h)`.

It would change the behavior (probably in a good way) if the client sends one or more unsupported checksums along with multiple valid ones, where we probably want to return `AlgoNotSupported` rather than `MultipleChecksumTypes` in priority in this case, but that's more a nitpick, either way should work.

lib/api/completeMultipartUpload.js
```diff
@@ -108,6 +114,83 @@ function validatePerPartChecksums(jsonList, storedParts, mpuSplitter, mpuChecksu
...
+    }
+
+    let result;
+    if (type === 'COMPOSITE') {
+        result = await computeCompositeMPUChecksum(
+            algorithm,
+            partInputs.map(p => p.value),
+        );
+    } else if (type === 'FULL_OBJECT') {
+        result = computeFullObjectMPUChecksum(algorithm, partInputs);
```
jonathan-gramain: Wondering what is the worst case run time of `computeFullObjectMPUChecksum`, to know if it shouldn't be an async function so it can yield to the event loop regularly? Assuming the worst case is 10K parts and it's just combining CRCs, I think it should not take more than one or a few ms, so should be fine I think, but better check if not done yet.

lib/api/completeMultipartUpload.js
```diff
@@ -490,56 +574,130 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
-                        dataLocations,
-                        [mpuOverviewKey],
-                        null,
-                        completeObjData,
-                        totalMPUSize,
-                    );
+                let nextCalled = false;
+                const callNext = (...args) => {
+                    if (nextCalled) {
+                        log.error('processParts: swallowed late callNext after next already invoked', {
```
jonathan-gramain: Not a big fan of this style of defensive programming: for me, either we know that next may be called multiple times by design and we accept it as part of the normal behavior (e.g. with `jsutil.once` wrapping the inner `next` callback where we know it may occur), or we should raise an exception (`async.*` typically does). This type of error will probably be lost in the logs without anyone noticing and finish as if `jsutil.once` was used in the first place, potentially hiding real issues.

The original code didn't have such error handling, so I think it should be fine to assume it will not happen if the logic introduced ensures a single callback call.

lib/api/completeMultipartUpload.js
leif-scality: A test was added in the previous PR for testing the worst case, and it takes a couple of milliseconds so it should be ok

lib/api/completeMultipartUpload.js
leif-scality: The current error message is identical to the one sent by AWS
```
<Error>
<Code>InvalidRequest</Code>
<Message>The upload was created using a sha256 checksum. The complete request must include the checksum for each part. It
  was missing for part 1 in the request.</Message>
<RequestId>2CMRHS7YQJ418XRD</RequestId><HostId>Z+nL7Rh7aglZj/H2D2aesp/j38QjuvN/wpgur86e7/0P6nohBelRVXaXY+mWSkQ
  dGaPIKHTmmg0iPUkaWjIWFTHmMJdNZolP</HostId>
</Error>
```

tests/unit/api/completeMultipartUpload.js
leif-scality: AWS returns two spaces. For all the final API/XML errors I return exactly what AWS returns 
```
 <Error><Code>InvalidPart</Code>
<Message>One or more of the specified parts could not be found.  The part may not have been uploaded, or the specified entity
  tag may not match the part's entity tag.</Message>
<UploadId>CTFNKyL2hI6n6irH0zbVWDzdPZ4n2ueJceRh1juCeuL2X5HOjrvCmXQMEqaoAatEWTEa3pWWxC7t9lOStMzjo0nJb4pv8Ct6oT
  Hv2n8mggVXRQ8RxiXyVyt3.3zpY98HVsZd.ozihhJ1HdUjLCkJtwJMQdNBd4fSdG9drS80vdg-</UploadId>
<PartNumber>1</PartNumber>
<ETag>0ebf9257a12e808d107b2ed1a826c122</ETag>
<RequestId>DAQAPVMCMSY4PDPV</RequestId>
<HostId>XS/2re3ieUQRTKdANLtZv14qyB2h3LjVHnmVrvjP0cj1PazPO16KkArQMtBLBy8S4mmLzQkuXZc=</HostId></Error>
```

tests/unit/api/completeMultipartUpload.js
leif-scality: The `algorithms` object contains the TAG of each algo already, this object is just for testing that the tag was not changed in the `algorithms` object

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: AWS checks `MultipleChecksumTypes` before `AlgoNotSupported`, the current behavior is the same as AWS.

AWS also ignores only  `x-amz-checksum-type` and `x-amz-checksum-algorithm`, they don't count them to the checksum count, and they also never trigger `AlgoNotSupported`. if we send `x-amz-checksum-BAD` on the other hand we get an  `AlgoNotSupported`. If we send `x-amz-checksum-BAD` + `x-amz-checksum-ZZZ` we get  `MultipleChecksumTypes`.

So the order is 
1. check no MultipleChecksumTypes
2. check no AlgoNotSupported
3. check actual checksum value, BadDigest if mismatch

I added a commit to ignore `x-amz-checksum-type` and `x-amz-checksum-algorithm` in the other functions, I didn't know this behavior existed.

lib/api/completeMultipartUpload.js
leif-scality: done

lib/api/completeMultipartUpload.js
leif-scality: done

tests/unit/api/completeMultipartUpload.js
leif-scality: done

tests/unit/api/completeMultipartUpload.js
leif-scality: moved the tests to `multipartUpload.js`.  This also triggered the new prettier linter in the file, I added it as a separate commit

lib/api/completeMultipartUpload.js
```diff
@@ -159,6 +335,29 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                        if (headerType) {
+                            const headerTypeUpper = headerType.toUpperCase();
+                            if (headerTypeUpper !== 'COMPOSITE' && headerTypeUpper !== 'FULL_OBJECT') {
+                                const typeErr = errorInstances.InvalidRequest.customizeDescription(
+                                    'Value for x-amz-checksum-type header is invalid.',
+                                );
+                                return next(typeErr, destBucket);
+                            }
+                            const mpuType = storedMetadata.checksumType;
+                            if (!mpuType || headerTypeUpper !== mpuType.toUpperCase()) {
```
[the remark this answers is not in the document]
leif-scality: done

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -316,30 +336,39 @@ function getChecksumDataFromHeaders(headers) {
  * @param {Buffer} body - http request body
  * @return {object} - error
  */
+// Validate a Content-MD5 header against the buffered body. Returns null on
+// success, an error object otherwise.
+function validateContentMd5(headers, body) {
```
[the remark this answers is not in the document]
leif-scality: done

tests/unit/api/multipartUpload.js
```diff
@@ -3070,420 +3243,1554 @@ describe('multipart upload in ingestion bucket', () => {
...
+            completeMultipartUpload(authInfo, req, log, completeErr => {
+                assert.ifError(completeErr);
+                done();
+            });
+        });
+    });
 
-        bucketPut(authInfo, bucketPutRequest, log, err => {
+    it('should reject CompleteMPU with a dedicated message when the MPU has no checksum mode', done => {
+        const services = require('../../../lib/services');
```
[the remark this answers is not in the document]
leif-scality: done

lib/api/completeMultipartUpload.js
leif-scality: Removed callNext. The promise's then(onResolve, onReject) plus the try/catch + returns in each handler already guarantee a single next call, so the guard was dead code. If a bug ever leaks through, async.waterfall will throw as you noted.

lib/api/completeMultipartUpload.js
```diff
@@ -379,55 +599,124 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                    uploadId,
+                    log,
+                ).then(
+                    ({ result, error }) => {
+                        if (error) {
+                            log.error('failing CompleteMPU due to final-object checksum error', {
+                                uploadId,
+                                error,
+                            });
+                            return callNext(error, destBucket);
```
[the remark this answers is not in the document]
leif-scality: fixed

tests/unit/api/multipartUpload.js
BourgoisMickael, line 1: This diff is too big, it could be in another PR by itself

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -162,7 +172,12 @@ const algorithms = Object.freeze({
  *   null on success; otherwise a ChecksumError with details.
  */
 async function validateXAmzChecksums(headers, body) {
-    const checksumHeaders = Object.keys(headers).filter(header => header.startsWith('x-amz-checksum-'));
+    const checksumHeaders = Object.keys(headers).filter(
+        header =>
+            header.startsWith('x-amz-checksum-') &&
+            header !== 'x-amz-checksum-type' &&
+            header !== 'x-amz-checksum-algorithm',
+    );
```
BourgoisMickael: This pattern reoccurs multiple times in this file, you can make a function for that

tests/functional/aws-node-sdk/test/object/completeMpuChecksum.js
```diff
@@ -0,0 +1,259 @@
...
+                        undefined,
+                        `COMPOSITE storage is deferred; HeadObject should not surface ${field}`,
+                    );
+                    assert.strictEqual(head.ChecksumType, undefined);
+                }
+            });
+        });
+
+        it(
+            'should return CRC64NVME/FULL_OBJECT on CompleteMPU response ' + 'when CreateMPU sent no checksum headers',
```
BourgoisMickael:
```suggestion
            'should return CRC64NVME/FULL_OBJECT on CompleteMPU response when CreateMPU sent no checksum headers',
```

tests/functional/aws-node-sdk/test/object/completeMpuChecksum.js
```diff
@@ -0,0 +1,259 @@
...
+                    await promise;
+                } catch (err) {
+                    caught = err;
+                }
+                assert(caught, 'expected CompleteMPU to reject');
+                assert.strictEqual(
+                    caught.name,
+                    'InvalidPart',
+                    `expected InvalidPart, got ${caught.name}: ${caught.message}`,
+                );
```
BourgoisMickael: You can use [`assert.rejects`](https://nodejs.org/docs/latest-v22.x/api/assert.html#assertrejectsasyncfn-error-message)


```js
            await assert.rejects(promise, err => {
                assert.strictEqual(
                    err.name,
                    'InvalidPart',
                    `expected InvalidPart, got ${err.name}: ${err.message}`,
                );
               });
```

leif-scality:
I forced push from the wrong branch, reverted the changes

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -467,6 +468,74 @@ describe('validateMethodChecksumNoChunking', () => {
...
+        it('should return BadDigest when Content-MD5 does not match the body', async () => {
+            config.integrityChecks.completeMultipartUpload = true;
+            const request = {
+                apiMethod: 'completeMultipartUpload',
+                headers: { 'content-md5': '1B2M2Y8AsgTpgAmY7PhCfg==' },
+            };
+            const log = { debug: sandbox.stub() };
+            const result = await validateMethodChecksumNoChunking(request, body, log);
+            assert.deepStrictEqual(result, ArsenalErrors.BadDigest);
+            assert(log.debug.calledOnce);
```
BourgoisMickael: debug logs should probably not be tested

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -467,6 +502,74 @@ describe('validateMethodChecksumNoChunking', () => {
...
+
+        it('should return null when Content-MD5 matches the body', async () => {
+            config.integrityChecks.completeMultipartUpload = true;
+            const request = {
+                apiMethod: 'completeMultipartUpload',
+                headers: { 'content-md5': correctMd5 },
+            };
+            const log = { debug: sandbox.stub() };
+            const result = await validateMethodChecksumNoChunking(request, body, log);
+            assert.strictEqual(result, null);
```
BourgoisMickael: In some tests above you use `assert.ifError(result)` and here `assert.strictEqual(result, null)`, you should stick to only 1 pattern and be consistent everywhere

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -467,6 +502,74 @@ describe('validateMethodChecksumNoChunking', () => {
...
+        });
+
+        it('should return BadDigest when Content-MD5 does not match the body', async () => {
+            config.integrityChecks.completeMultipartUpload = true;
+            const request = {
+                apiMethod: 'completeMultipartUpload',
+                headers: { 'content-md5': '1B2M2Y8AsgTpgAmY7PhCfg==' },
+            };
+            const log = { debug: sandbox.stub() };
```
BourgoisMickael: It would be better to have a full mocked logger like it's done in some other test files, to avoid breakin if we change from debug to another level

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: I will remove the log tests

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: I will remove the log tests

lib/api/completeMultipartUpload.js
jonathan-gramain: understood

tests/unit/api/completeMultipartUpload.js
jonathan-gramain: Could you construct it from `algorithms`? If `algorithm` is supposed to be a constant and we are concerned that it may be accidentally changed, what about using `Object.freeze` on it?

tests/unit/api/completeMultipartUpload.js
jonathan-gramain: okay fair enough (my little voice tells me that AWS didn't care that much about the beauty of their error messages :slightly_smiling_face: )

jonathan-gramain: I believe no sane application would match an error by their exact error message to the dot, but I'm good with sticking to AWS to avoid ourselves asking the question. Personally I wouldn't care though :stuck_out_tongue:

jonathan-gramain: Also I believe the error messages are not enforced by the API contract, hence may change at any time on AWS side, which would make our effort useless in the end. Just my two cents and my last comment on this :wink:

lib/api/apiUtils/integrity/validateChecksums.js
jonathan-gramain: Sounds good to me.

IMHO we don't necessarily have to mimic every AWS behavior when it comes to error handling in particular, as it takes time to reverse engineer those aspects with no clear benefit, and AWS implementation is not perfect and their choices can be debated too. Sure it became the _de facto_ standard and we need to stick to its way of doing things to avoid breaking existing applications, but not necessarily in all its nitty gritty details and quirks.

lib/api/completeMultipartUpload.js
```diff
@@ -106,6 +112,83 @@ function validatePerPartChecksums(jsonList, storedParts, mpuSplitter, mpuChecksu
...
+ * @param {array} filteredPartList - validated, ordered subset matching jsonList
+ * @param {object} storedMetadata - MPU overview metadata
+ * @param {string} mpuSplitter - splitter used in part keys
+ * @param {string} uploadId - for log context
+ * @param {object} log - werelogs logger
+ * @returns {object|null} { algorithm, type, value } or null
+ */
+async function computeFinalChecksum(storedParts, filteredPartList, storedMetadata, mpuSplitter, uploadId, log) {
+    const algorithm = storedMetadata.checksumAlgorithm;
+    const type = storedMetadata.checksumType;
```
jonathan-gramain, line 155: Deconstruction could be slightly more readable

```suggestion
    const { checksumAlgorithm: algorithm, checksumType: type } = storedMetadata;
```

lib/api/completeMultipartUpload.js
```diff
@@ -497,55 +581,117 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                );
+                return undefined;
 
-                const partsInfo = generateMpuPartStorageInfo(filteredPartsObj.partList);
-                if (partsInfo.error) {
-                    return next(partsInfo.error, destBucket);
-                }
-                const { keysToDelete, extraPartLocations } = filteredPartsObj;
-                const { aggregateETag, dataLocations, calculatedSize } = partsInfo;
+                function continueProcessParts() {
```
jonathan-gramain, line 653: Shouldn't this function be another step in the waterfall?

lib/api/completeMultipartUpload.js
```diff
@@ -497,55 +581,117 @@ function completeMultipartUpload(authInfo, request, log, callback) {
...
+                                return next(arsenalErrorFromChecksumError(checksumErr), destBucket);
+                            }
+                            return continueProcessParts(null);
+                        } catch (resolveErr) {
+                            log.error('unexpected throw after final-checksum compute', {
+                                uploadId,
+                                error: resolveErr,
+                            });
+                            return next(resolveErr, destBucket);
+                        }
```
jonathan-gramain, line 641: Is there a benefit in catching those errors since we already have a catch-all below

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -467,6 +468,74 @@ describe('validateMethodChecksumNoChunking', () => {
             assert.strictEqual(result, null);
         });
     });
+
+    // CompleteMPU is not in `checksumedMethods` (x-amz-checksum-* is the
+    // final-object checksum, not a body digest) but it still must validate
+    // Content-MD5 against the XML body when present.
+    describe('completeMultipartUpload (md5-only path)', () => {
```
jonathan-gramain, line 491: I think a lot of the unit tests could be turned into a declarative matrix for readability.

tests/unit/api/multipartUpload.js
```diff
@@ -4519,3 +4519,174 @@ describe('CompleteMultipartUpload final-object checksum storage', () => {
...
+                },
+                partBody,
+            );
+            objectPutPart(authInfo, partRequest, undefined, log, err => (err ? reject(err) : resolve()));
+        });
+    }
+
+    // Resolves with { xml, headers } so callers can inspect both the
+    // response body and the response headers.
+    function completeMpuP(uploadId, partChecksumXml = '') {
```
jonathan-gramain: It seems to me the test code is riddled with this type of local helpers, maybe we should consider a common layer of helpers for MPU construction?

tests/functional/aws-node-sdk/test/object/completeMpuChecksum.js
```diff
@@ -134,53 +134,126 @@ describe('CompleteMultipartUpload final-object checksum', () =>
...
-                new HeadObjectCommand({
-                    Bucket: bucket,
-                    Key: key,
-                    ChecksumMode: 'ENABLED',
-                }),
-            );
-            assert.strictEqual(head.ChecksumCRC64NVME, complete.ChecksumCRC64NVME);
-            assert.strictEqual(head.ChecksumType, 'FULL_OBJECT');
+        it(
+            'should return CRC64NVME/FULL_OBJECT on CompleteMPU response ' + 'when CreateMPU sent no checksum headers',
```
jonathan-gramain: Extra newline and concatenation doesn't look necessary.

tests/unit/api/multipartUpload.js
leif-scality: this was caused by the linter

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: created helper

tests/functional/aws-node-sdk/test/object/completeMpuChecksum.js
leif-scality: fixed

tests/functional/aws-node-sdk/test/object/completeMpuChecksum.js
leif-scality: done

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: done

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: used ifError everywhere

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: done

tests/unit/api/completeMultipartUpload.js
leif-scality: The goal of the test if to raise an error if we change the `xmlTag` for some reason in `algorithms`. If we construct it from algorithms then we will never detect the change. `algorithms` is already `Object.freeze`

tests/unit/api/completeMultipartUpload.js
leif-scality: Yeah the message can be improved, but I just copy and paste it to stay consistent. Also having the same error message may help the client if they google it, they will get more information than with a custom one

lib/api/apiUtils/integrity/validateChecksums.js
leif-scality: I needed to test a lot to prevent being more strict than AWS and introducing breaking changes. So I have a list of errors already so we might as well use it

tests/unit/api/multipartUpload.js
leif-scality: moved them to mpuUtils.js

tests/functional/aws-node-sdk/test/object/completeMpuChecksum.js
leif-scality: removed

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: This is an edge case, Other tests use for loops to test all checksums algorithms already

lib/api/completeMultipartUpload.js
leif-scality: integrated it with the waterfall

lib/api/completeMultipartUpload.js
leif-scality: fixed by the previous comment

lib/api/completeMultipartUpload.js
leif-scality: done

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -467,6 +468,74 @@ describe('validateMethodChecksumNoChunking', () => {
...
+            const result = await validateMethodChecksumNoChunking(request, body, log);
+            assert.strictEqual(result, null);
+        });
+
+        it('should NOT validate x-amz-checksum-* as a body digest (final-object semantics)', async () => {
+            // If we routed CompleteMPU through `defaultValidationFunc` like
+            // the other methods, this wrong x-amz-checksum-sha256 (treated
+            // as a body digest) would return BadDigest. The md5-only path
+            // must ignore it — the final-object validator handles it later.
+            config.integrityChecks.completeMultipartUpload = true;
```
jonathan-gramain: Wouldn't this potentially affect other unrelated tests and should be reverted after the test?

tests/unit/api/multipartUpload.js
```diff
@@ -4134,3 +4133,208 @@ describe('CompleteMultipartUpload body-checksum bypass', () => {
...
+        // computeCompositeMPUChecksum's deterministic output for these
+        // exact placeholder digests:
+        const expected = crypto
+            .createHash('sha256')
+            .update(Buffer.concat([d1, d2, d3].map(x => Buffer.from(x, 'base64'))))
+            .digest('base64');
+        assert.strictEqual(got.value, `${expected}-3`);
+    });
+
+    ['sha1', 'crc32', 'crc32c'].forEach(algo => {
```
jonathan-gramain: You can probably include the `sha256` test here, and if you don't want to or cannot check the actual value for non-sha256 algorithms, just condition this check to `algo === 'sha256'`. The sha256 tests uses three parts but repeats parts 1 and 3, not sure if this is useful or wanted.

tests/unit/api/multipartUpload.js
```diff
@@ -4134,3 +4133,208 @@ describe('CompleteMultipartUpload body-checksum bypass', () => {
...
+            { checksumAlgorithm: 'sha256', checksumType: 'WEIRD' },
+            splitter,
+            uploadId,
+            log,
+        );
+        assert.strictEqual(got, null);
+    });
+
+    it(
+        'should return null when underlying compute reports an error ' + '(crc64nvme COMPOSITE is not allowed)',
```
jonathan-gramain:
```suggestion
        'should return null when underlying compute reports an error (crc64nvme COMPOSITE is not allowed)',
```

tests/functional/aws-node-sdk/test/object/completeMpuChecksum.js
```diff
@@ -134,55 +134,58 @@ describe('CompleteMultipartUpload final-object checksum', () =>
             });
         });
 
-        it('should return CRC64NVME/FULL_OBJECT on response when CreateMPU sent no checksum headers', async () => {
-            const key = `complete-default-${Date.now()}`;
+        it(
+            'should return CRC64NVME/FULL_OBJECT on CompleteMPU response ' + 'when CreateMPU sent no checksum headers',
```
jonathan-gramain: The re-indentation doesn't look necessary here, or is it enforced by the new linter rules?

leif-scality: fixed

tests/unit/api/multipartUpload.js
leif-scality: fixed

tests/unit/api/multipartUpload.js
leif-scality: changed the test to use a list of [algo, expectedDigest]

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: This was old code, removed it. No longer relevant

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -549,13 +664,13 @@ const COMPOSITE_ALGOS = new Set(['crc32', 'crc32c', 'sha1', 'sha256']);
  * @returns {{ checksum: string, error: null }
  *   | { checksum: null, error: { code: string, details: object } }}
  */
-function computeCompositeMPUChecksum(algorithm, partChecksumsBase64) {
+async function computeCompositeMPUChecksum(algorithm, partChecksumsBase64) {
```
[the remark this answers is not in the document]
leif-scality: There is an await in the function...

tests/unit/api/multipartUpload.js
```diff
@@ -194,41 +272,33 @@ describe('Multipart Upload API', () => {
...
     });
 
-    it('should return an error on an initiate multipart upload call if ' +
-        'no destination bucket', done => {
-        initiateMultipartUpload(authInfo, initiateRequest,
-            log, err => {
-                assert(err.is.NoSuchBucket);
-                done();
-            });
+    it('should return an error on an initiate multipart upload call if ' + 'no destination bucket', done => {
```
BourgoisMickael:
```suggestion
    it('should return an error on an initiate multipart upload call if no destination bucket', done => {
```

tests/unit/api/multipartUpload.js
```diff
@@ -242,1780 +312,1854 @@ describe('Multipart Upload API', () => {
...
-                bucketName,
-                namespace,
-                objectKey,
-                headers: { host: `${bucketName}.s3.amazonaws.com` },
-                url: `/${objectKey}?partNumber=1&uploadId=${testUploadId}`,
-                query: {
-                    partNumber: '2',
-                    uploadId: testUploadId,
+    it(
+        'should complete a multipart upload even if etag is sent ' + 'in post body without quotes (a la Cyberduck)',
```
BourgoisMickael:
```suggestion
        'should complete a multipart upload even if etag is sent in post body without quotes (a la Cyberduck)',
```

BourgoisMickael: this happens in other places as well in this file

tests/unit/api/multipartUpload.js
leif-scality: linted the whole PR, 12 changes made

tests/unit/api/multipartUpload.js
```diff
@@ -3456,34 +3621,799 @@ describe('initiateMultipartUpload checksum headers', () => {
...
+                    const wrongDigest = SAMPLE_DIGESTS[wrongAlgo][0];
+                    const jsonList = {
+                        Part: [
+                            makeJsonPart(1, 'etag1', { [wrongTag]: wrongDigest }),
+                            makeJsonPart(2, 'etag2', { [tag]: d2 }),
+                        ],
+                    };
+                    const err = validatePerPartChecksums(jsonList, stored, splitter, mpuChecksum);
+                    assert(err);
+                    assert.strictEqual(err.is.BadDigest, true);
```
BourgoisMickael:
```suggestion
                    assert.strictEqual(err.message, 'BadDigest');
```

This would give more information in case of failure rather than boolean check `false !== true`

tests/unit/api/multipartUpload.js
```diff
@@ -3456,34 +3621,799 @@ describe('initiateMultipartUpload checksum headers', () => {
...
+                    // AWS-style message: "The {algo} you specified for part {N} did not match what we received."
+                    assert.strictEqual(
+                        err.description,
+                        `The ${wrongAlgo} you specified for part 1 did ` + 'not match what we received.',
```
BourgoisMickael:
```suggestion
                        `The ${wrongAlgo} you specified for part 1 did not match what we received.`,
```

tests/unit/api/multipartUpload.js
```diff
@@ -3456,34 +3621,799 @@ describe('initiateMultipartUpload checksum headers', () => {
...
+                it(missingLabel, () => {
+                    const jsonList = {
+                        Part: [makeJsonPart(1, 'etag1', { [tag]: d1 }), makeJsonPart(2, 'etag2')],
+                    };
+                    const err = validatePerPartChecksums(jsonList, stored, splitter, mpuChecksum);
+                    if (requiresPerPart) {
+                        assert(err);
+                        assert.strictEqual(err.is.InvalidRequest, true);
+                        assert(err.description.includes(algorithm));
+                        assert(err.description.includes('part 2 in the request'));
```
BourgoisMickael: you could use `assert.match` with regex to see better messages on error rather than boolean comparison

tests/unit/api/multipartUpload.js
leif-scality: I fixed other cases, the others ones where already there fixed some, some make the linter fail so they remain with '+'

tests/unit/api/multipartUpload.js
leif-scality: fixed where possible

tests/unit/api/multipartUpload.js
leif-scality: fixed

tests/unit/api/multipartUpload.js
leif-scality: fixed

tests/unit/api/completeMultipartUpload.js
jonathan-gramain: Then, why not using `Object.freeze` for each subcomponent of `algorithms` if the goal is to ensure that it's never modified, that would be probably a better answer than a test checking this. Just my two cents, no big deal otherwise.

jonathan-gramain (approved):

nicolas2bert (approved):

leif-scality:
ping

leif-scality:
ping

PR #6174, opened 2026-05-21
CLDSRV-901: Add functional test coverage for ListObjectsV2 optional object attributes

dvasilas:
This PR adds functional test coverage for `x-amz-optional-object-attributes` header on `ListObjectsV2` and `ListObjectVersions` (these tests will also run in scality/Integration).

This feature has been implemented for Artesca. For context see [requirements](https://github.com/scality/citadel/blob/development/1.0/docs/requirements/s3-apis/bucket-listobjectsv2-optional-object-attributes.md) and [design](https://github.com/scality/citadel/blob/development/1.0/docs/design/s3-apis/bucket-listobjectsv2-optional-object-attributes.md).

It also includes two fixes:
- If the request included `RestoreStatus` multiple times, then the response included multiple `<RestoreStatus>` blocks. The fix is to deduplicate `<RestoreStatus>` in the response (requests with multiple `RestoreStatus` are still allowed).
- Delete markers included optional attributes. The fix is to omit optional attributes from delete markers (this matches AWS behavior).

### DIFF
--- a/.github/docker/docker-compose.yaml
+++ b/.github/docker/docker-compose.yaml
@@ -107,7 +107,7 @@ services:
     volumes:
       - /tmp/artifacts/${JOB_NAME}/sproxyd:/logs
   vault:
-    image: ghcr.io/scality/vault:7.76.0
+    image: ghcr.io/scality/vault:7.89.0
     profiles: ['vault']
     user: root
     command: sh -c "chmod 400 tests/utils/keyfile && yarn start > /artifacts/vault.log 2> /artifacts/vault-stderr.log"
@@ -122,6 +122,8 @@ services:
       - ENABLE_LOCAL_CACHE=true
       - REDIS_HOST=0.0.0.0
       - REDIS_PORT=6379
+      # TODO: drop NODE_AUTH_TOKEN when VAULT-721 is done
+      - NODE_AUTH_TOKEN=dummy
     depends_on:
       - redis
   metadata-standalone:
--- a/lib/api/bucketGet.js
+++ b/lib/api/bucketGet.js
@@ -332,16 +332,19 @@ function bucketGet(authInfo, request, log, callback) {
     const bucketName = request.bucketName;
     const v2 = params['list-type'];
 
-    const optionalAttributes =
+    const optionalAttributes = new Set(
         request.headers['x-amz-optional-object-attributes']
             ?.split(',')
             .map(attr => attr.trim())
             .map(attr => attr !== 'RestoreStatus' ? attr.toLowerCase() : attr)
-        ?? [];
-    if (optionalAttributes.some(attr => !attr.startsWith('x-amz-meta-') && attr != 'RestoreStatus')) {
-        return callback(
-            errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified')
-        );
+        ?? []
+    );
+    for (const attr of optionalAttributes) {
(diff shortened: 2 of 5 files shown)


lib/api/bucketGet.js
```diff
@@ -266,7 +268,7 @@ function processOptionalAttributes(item, optionalAttributes) {
     const xml = [];
     const userMetadata = new Set();
 
-    for (const attribute of optionalAttributes) {
+    for (const attribute of new Set(optionalAttributes)) {
```
maeldonn: `new Set()` is called for every listed object. Could we do it once when parsing the header ?

dvasilas: Good point. Fixed in https://github.com/scality/cloudserver/pull/6174/commits/eebfbd9a964766d843c842fceed5e4ab1dbb8769.

maeldonn (approved):
LGTM

.github/docker/docker-compose.yaml
```diff
@@ -122,6 +122,8 @@ services:
       - ENABLE_LOCAL_CACHE=true
       - REDIS_HOST=0.0.0.0
       - REDIS_PORT=6379
+      # TODO: drop NODE_AUTH_TOKEN when VAULT-721 is done
```
DarkIsDude, line 125: We should mention that in the JIRA also to make sure the develop who will resolve the VAULT-721, will remove it from here 🙏. Why waiting to do it ?

lib/api/bucketGet.js
```diff
@@ -332,16 +334,19 @@ function bucketGet(authInfo, request, log, callback) {
...
             .map(attr => attr.trim())
             .map(attr => attr !== 'RestoreStatus' ? attr.toLowerCase() : attr)
-        ?? [];
-    if (optionalAttributes.some(attr => !attr.startsWith('x-amz-meta-') && attr != 'RestoreStatus')) {
-        return callback(
-            errorInstances.InvalidArgument.customizeDescription('Invalid attribute name specified')
-        );
+        ?? []
+    );
+    for (const attr of optionalAttributes) {
```
DarkIsDude, line 344: Are we sure this match the AWS behaviour ? Of course no duplicate blocks, but should we reject the request ?

DarkIsDude (approved):

lib/api/bucketGet.js
dvasilas: Checked with multiple `RestoreStatus`. AWS does not reject the request
```
aws s3api list-objects-v2 --bucket  cldsrv-probe-optattr-manual --optional-object-attributes RestoreStatus RestoreStatus --no-cli-pager
{
    "Contents": [
        {
            "Key": "obj1",
            "LastModified": "2026-05-21T10:15:25+00:00",
            "ETag": "\"74e11c9bddafedf3ed7cd8bea153e7c8\"",
            "ChecksumAlgorithm": [
                "CRC64NVME"
            ],
            "ChecksumType": "FULL_OBJECT",
            "Size": 19,
            "StorageClass": "STANDARD"
        }
    ],
    "RequestCharged": null,
    "Prefix": ""
}
```


Side note: We have a divergence with AWS. AWS does not include `RestoreStatus` in the response for `STANDARD` objects (but I didn't want to change this in this PR because it will break existing behavior in Artesca).

.github/docker/docker-compose.yaml
dvasilas: Done: https://scality.atlassian.net/browse/VAULT-721?focusedCommentId=475640

PR #6176, opened 2026-05-27
CLDSRV-908: CopyObject handle checksums

leif-scality:
- Forward a src object checksum to the dest object
- Recompute the checksum when required (x-amz-checksum-algorithm header set, src object MPU with COMPOSITE checksum, ...). If needed this will trigger a GET operation to download and calculate the object checksum
- Compute a CRC64NMVE for the dest object if the src object has no checksum (counts as a recompute so the same conditions as above apply)

```
  ┌─────┬─────────────────────┬────────────────────────────────────┬──────────────────────────────────────────────────────────────────┐
  │  #  │   Source checksum   │ x-amz-checksum-algorithm requested │                            Recompute?                            │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 1   │ none                │ none                               │ Yes (Compute default CRC64NVME)                                  │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 2   │ FULL_OBJECT, algo X │ none                               │ No — propagate as-is                                             │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 3   │ FULL_OBJECT, algo X │ X (same algo)                      │ No — propagate as-is                                             │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 4   │ FULL_OBJECT, algo X │ Y (different algo)                 │ Yes                                                              │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 5   │ COMPOSITE, algo X   │ none                               │ Yes (can't propagate a MPU format digest to a single-object dest)│
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 6   │ COMPOSITE, algo X   │ X (same algo)                      │ Yes                                                              │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 7   │ COMPOSITE, algo X   │ Y (different algo)                 │ Yes                                                              │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 8   │ none                │ any algo                           │ Yes (source has no digest, must compute)                         │
  ├─────┼─────────────────────┼────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ 9   │ any                 │ source is 0-byte                   │ Yes (special-cased — empty-bytes digest, no streaming)           │
  └─────┴─────────────────────┴────────────────────────────────────┴──────────────────────────────────────────────────────────────────┘
```

| Backend / scenario | Copy method | Old Copy method (pre-PR) |
|---|---|---|
| **file / mem / sproxyd (scality)** | `GET` + `PUT` + `DELETE` | `GET` + `PUT` + `DELETE` (same) |
| **aws_s3 / azure / gcp — propagate** (FULL_OBJECT src, algorithm absent or matching) | Native server-side copy (`CopyObject` / `beginCopyFromURL` / `copyObject`) — no GET | Native server-side copy (same) |
| **aws_s3 / azure / gcp — recompute** | `GET` + `PUT` + `DELETE` — streamed through CloudServer | Native server-side copy — no GET |
| **external, cross-type or external ↔ local** (e.g. aws → azure, aws → file) | `GET` + `PUT` + `DELETE` | `GET` + `PUT` + `DELETE` (same) |
| **external, any + SSE** | `GET` + `PUT` + `DELETE` | `GET` + `PUT` + `DELETE` (same) |
| **copy-to-self — propagate** (same location, no SSE, unversioned) | metadata-only — reuse location (no data) | metadata-only — reuse location (same) |
| **copy-to-self — recompute** (any backend) | `GET` only — reuse location (no PUT, no DELETE) | metadata-only — reuse location, no GET |
| **0-byte source** (same location / local) | metadata-only — no data | metadata-only — no data (same) |
| **0-byte source → different external backend** | `PUT` (`data.put(null)`) | `PUT` (`data.put(null)`) (same) |

### DIFF
--- a/lib/api/apiUtils/integrity/validateChecksums.js
+++ b/lib/api/apiUtils/integrity/validateChecksums.js
@@ -38,6 +38,13 @@ const errMPUTypeWithoutAlgo = errorInstances.InvalidRequest.customizeDescription
     'The x-amz-checksum-type header can only be used with the x-amz-checksum-algorithm header.',
 );
 
+// TODO(S3C-11278): Update with 'MD5', 'SHA512', 'XXHASH128', 'XXHASH3', 'XXHASH64' when they are introduced.
+// https://scality.atlassian.net/browse/S3C-11278
+const errCopyChecksumAlgoNotSupported = errorInstances.InvalidRequest.customizeDescription(
+    'Checksum algorithm provided is unsupported. Please try again with any of the valid types: ' +
+        '[CRC32, CRC32C, CRC64NVME, SHA1, SHA256]',
+);
+
 // Methods that validate BOTH Content-MD5 and x-amz-checksum-* against the
 // buffered request body. For these, x-amz-checksum-* is the body digest.
 const checksumedMethods = Object.freeze({
@@ -90,6 +97,7 @@ const ChecksumError = Object.freeze({
     MPUTypeInvalid: 'MPUTypeInvalid',
     MPUTypeWithoutAlgo: 'MPUTypeWithoutAlgo',
     MPUInvalidCombination: 'MPUInvalidCombination',
+    CopyChecksumAlgoNotSupported: 'CopyChecksumAlgoNotSupported',
 });
 
 const base64Regex = /^[A-Za-z0-9+/]*={0,2}$/;
@@ -438,6 +446,8 @@ function arsenalErrorFromChecksumError(err) {
                 `The ${err.details.type} checksum type cannot be used ` +
                     `with the ${err.details.algorithm.toUpperCase()} checksum algorithm.`,
             );
+        case ChecksumError.CopyChecksumAlgoNotSupported:
+            return errCopyChecksumAlgoNotSupported;
         default:
             return ArsenalErrors.BadDigest;
     }
@@ -699,6 +709,31 @@ function computeFullObjectMPUChecksum(algorithm, parts) {
     return { checksum: combinePartCrcs(parts, params.polyReversed, params.dim), error: null };
 }
 
+/**
+ * Read and validate the x-amz-checksum-algorithm header for CopyObject.
+ *
+ * @param {object} headers - request headers
+ * @returns {{ error: null, algorithm: string|null }
+ *   | { error: { error: string, details: { algorithm: string } }, algorithm: null }}
+ */
+function getCopyObjectChecksumAlgorithm(headers) {
+    const requested = headers['x-amz-checksum-algorithm'];
+    if (requested === undefined) {
+        return { error: null, algorithm: null };
+    }
+    const algorithm = requested.toLowerCase();
+    if (!algorithms[algorithm]) {
+        return {
+            error: {
+                error: ChecksumError.CopyChecksumAlgoNotSupported,
+                details: { algorithm: requested },
+            },
+            algorithm: null,
+        };
+    }
+    return { error: null, algorithm };
+}
+
 module.exports = {
     ChecksumError,
     defaultChecksumData,
@@ -712,4 +747,5 @@ module.exports = {
     computeCompositeMPUChecksum,
     computeFullObjectMPUChecksum,
     validateCompleteMultipartUploadChecksum,
+    getCopyObjectChecksumAlgorithm,
 };
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -36,6 +36,31 @@ const externalVersioningErrorMessage = 'We do not currently support putting ' +
 'a versioned object to a location-constraint of type AWS or Azure or GCP.';
 
 /**
+ * Compute the prior data locations that are orphaned.
+ *
+ * @param {Array} dataToDelete - prior data locations from versioningPreprocessing
+ * @param {Array} newDataGetInfo - new data locations stored in the new metadata
+ * @returns {Array|null} orphaned entries to batchDelete, or null if none
+ */
+function _orphanedDataLocations(dataToDelete, newDataGetInfo) {
+    // Identity is (dataStoreName, key): the same key under a different
+    // dataStoreName points to a different backend slot, so the old slot
+    // is an orphan even when the key string collides. dataStoreName is a
+    // location-constraint name (never contains '|'), so the delimiter is
+    // unambiguous.
+
+    // We need to normalize to support old legacy location names.
+    const normalize = loc => (typeof loc === 'string' ? { key: loc } : loc);
+    const locId = loc => {
+        const n = normalize(loc);
+        return `${n.dataStoreName}|${n.key}`;
+    };
+    const newIds = new Set((newDataGetInfo || []).map(locId));
+    const orphans = (dataToDelete || []).filter(loc => !newIds.has(locId(loc)));
+    return orphans.length > 0 ? orphans : null;
+}
+
+/**
  * Preps metadata to be saved (based on copy or replace request header)
  * @param {object} request - request
  * @param {object} sourceObjMD - object md of source object
@@ -547,7 +572,8 @@ function objectCopy(authInfo, request, sourceBucket,
                     if (options.extraMD) {
                         Object.assign(storeMetadataParams, options.extraMD);
                     }
-                    const dataToDelete = options.dataToDelete;
+                    const dataToDelete = _orphanedDataLocations(
+                        options.dataToDelete, destDataGetInfoArr);
                     return next(null, storeMetadataParams, destDataGetInfoArr,
                         destObjMD, serverSideEncryption, destBucketMD,
                         dataToDelete);
@@ -589,11 +615,7 @@ function objectCopy(authInfo, request, sourceBucket,
         function deleteExistingData(dataToDelete, storingNewMdResult,
             destBucketMD, storeMetadataParams, serverSideEncryption,
             sourceObjSize, destObjPrevSize, next) {
-            // Clean up any potential orphans in data if object
-            // put is an overwrite of already existing
-            // object with same name, so long as the source is not
-            // the same as the destination
-            if (!sourceIsDestination && dataToDelete) {
+            if (dataToDelete) {
                 const newDataStoreName = storeMetadataParams.dataStoreName;
                 return data.batchDelete(dataToDelete, request.method,
                     newDataStoreName, log, err => {
@@ -688,3 +710,5 @@ function objectCopy(authInfo, request, sourceBucket,
 }
 
 module.exports = objectCopy;
+// Exposed for unit testing only.
+module.exports._orphanedDataLocations = _orphanedDataLocations;
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -32,6 +32,7 @@ const { initializeInternalLogRequestQueue, queueInternalLogRequest } = require('
 const { algorithms, arsenalErrorFromChecksumError, getCopyObjectChecksumAlgorithm } =
     require('./apiUtils/integrity/validateChecksums');
 const ChecksumTransform = require('../auth/streamingV4/ChecksumTransform');
+const ChecksumWritable = require('../auth/streamingV4/ChecksumWritable');
 const kms = require('../kms/wrapper');
 
 const versionIdUtils = versioning.VersionID;
@@ -146,11 +147,12 @@ function _getCopyObjectChecksumAlgorithm(requestedAlgo, sourceObjMD) {
 }
 
 /**
- * Recompute the destination's checksum by streaming the source bytes
- * through a ChecksumTransform. When the bytes don't have to move
- * (copy-to-self, same location, no SSE, not versioned) the source is GET'ed
- * solely to compute the new digest; the bytes are then discarded and the
+ * Recompute the destination's checksum by streaming the source bytes through
+ * a checksum stream. When the bytes don't have to move (copy-to-self, same
+ * location, no SSE, not versioned) the source is GET'ed solely to compute the
+ * new digest via a ChecksumWritable sink; the bytes are then discarded and the
  * existing data location is reused — only cloudserver's metadata is updated.
+ * Otherwise the source is piped through a ChecksumTransform into a single put.
  *
  * @param {object} log - request logger
  * @param {object} request - HTTP request
@@ -177,19 +179,19 @@ function _recomputeChecksumAndStore(log, request, requestedAlgo, sourceObjMD, da
     const wrapChecksumErr = err => Object.assign(err, { checksumStream: { algorithm: algoName } });
 
     // Copy-to-self where the bytes don't have to move (same location, no SSE
-    // change, not versioned): GET the source through a ChecksumTransform to
-    // compute the new digest, discard the bytes, and reuse the existing data
-    // locations. No PUT, no DELETE — only the metadata gets updated.
+    // change, not versioned): GET the source into a ChecksumWritable to compute
+    // the new digest, discard the bytes, and reuse the existing data locations.
+    // No PUT, no DELETE — only the metadata gets updated.
     if (sourceIsDestination && storeMetadataParams.locationMatch
         && !needsEncryption && !isVersionedObj) {
         log.debug('computing checksum on copy-to-self without rewriting data',
             { algorithm: algoName, size: storeMetadataParams.size });
         const sourceStream = _pipeSourcePartsThrough(dataLocator, log);
-        const checksumStream = new ChecksumTransform(algoName, undefined, false, log);
+        const checksumSink = new ChecksumWritable(algoName, log);
         const finish = jsutil.once(err => {
             if (err) {
                 sourceStream.destroy(err);
-                checksumStream.destroy(err);
+                checksumSink.destroy(err);
                 if (request.sourceServerAccessLog) {
                     // eslint-disable-next-line no-param-reassign
                     request.sourceServerAccessLog.error = err;
@@ -199,17 +201,16 @@ function _recomputeChecksumAndStore(log, request, requestedAlgo, sourceObjMD, da
             // eslint-disable-next-line no-param-reassign
             storeMetadataParams.checksum = {
                 algorithm: algoName,
-                value: checksumStream.digest,
+                value: checksumSink.digest,
                 type: 'FULL_OBJECT',
             };
             return next(null, storeMetadataParams, dataLocator, destObjMD,
                 serverSideEncryption, destBucketMD);
         });
         sourceStream.once('error', finish);
-        checksumStream.once('error', err => finish(wrapChecksumErr(err)));
-        checksumStream.on('data', () => {});
-        checksumStream.once('end', () => finish());
-        sourceStream.pipe(checksumStream);
+        checksumSink.once('error', err => finish(wrapChecksumErr(err)));
+        checksumSink.once('finish', () => finish());
+        sourceStream.pipe(checksumSink);
         return undefined;
     }
 
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -8,18 +8,15 @@ const validateHeaders = s3middleware.validateConditionalHeaders;
 
 const constants = require('../../constants');
 const collectCorsHeaders = require('../utilities/collectCorsHeaders');
-const locationConstraintCheck
-    = require('./apiUtils/object/locationConstraintCheck');
-const { checkQueryVersionId, versioningPreprocessing, decodeVID }
-    = require('./apiUtils/object/versioning');
+const locationConstraintCheck = require('./apiUtils/object/locationConstraintCheck');
+const { checkQueryVersionId, versioningPreprocessing, decodeVID } = require('./apiUtils/object/versioning');
 const getReplicationInfo = require('./apiUtils/object/getReplicationInfo');
 const { data } = require('../data/wrapper');
 const services = require('../services');
 const { pushMetric } = require('../utapi/utilities');
 const removeAWSChunked = require('./apiUtils/object/removeAWSChunked');
 const { standardMetadataValidateBucketAndObj } = require('../metadata/metadataUtils');
-const validateWebsiteHeader = require('./apiUtils/object/websiteServing')
-    .validateWebsiteHeader;
+const validateWebsiteHeader = require('./apiUtils/object/websiteServing').validateWebsiteHeader;
 const { config } = require('../Config');
 const monitoring = require('../utilities/monitoringHandler');
 const applyZenkoUserMD = require('./apiUtils/object/applyZenkoUserMD');
@@ -29,8 +26,11 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
 const { setSSEHeaders } = require('./apiUtils/object/sseHeaders');
 const { updateEncryption } = require('./apiUtils/bucket/updateEncryption');
 const { initializeInternalLogRequestQueue, queueInternalLogRequest } = require('../utilities/serverAccessLogger');
-const { algorithms, arsenalErrorFromChecksumError, getCopyObjectChecksumAlgorithm } =
-    require('./apiUtils/integrity/validateChecksums');
+const {
+    algorithms,
+    arsenalErrorFromChecksumError,
+    getCopyObjectChecksumAlgorithm,
+} = require('./apiUtils/integrity/validateChecksums');
 const ChecksumTransform = require('../auth/streamingV4/ChecksumTransform');
 const ChecksumWritable = require('../auth/streamingV4/ChecksumWritable');
 const kms = require('../kms/wrapper');
@@ -38,8 +38,8 @@ const kms = require('../kms/wrapper');
 const versionIdUtils = versioning.VersionID;
 const locationHeader = constants.objectLocationConstraintHeader;
 const versioningNotImplBackends = constants.versioningNotImplBackends;
-const externalVersioningErrorMessage = 'We do not currently support putting ' +
-'a versioned object to a location-constraint of type AWS or Azure or GCP.';
+const externalVersioningErrorMessage =
+    'We do not currently support putting ' + 'a versioned object to a location-constraint of type AWS or Azure or GCP.';
 
 /**
  * Compute the prior data locations that are orphaned.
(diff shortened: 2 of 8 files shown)


tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -1535,3 +1535,166 @@ describe('Object Copy with object lock enabled on both destination ' +
...
+
+        afterEach(async () => {
+            await bucketUtil.empty(sourceBucketName, true);
+            await bucketUtil.empty(destBucketName, true);
+        });
+
+        after(async () => bucketUtil.deleteMany([sourceBucketName, destBucketName]));
+
+        const { algorithms } = require(
+            '../../../../../lib/api/apiUtils/integrity/validateChecksums');
```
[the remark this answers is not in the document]
leif-scality: fixed

lib/api/objectCopy.js
```diff
@@ -468,30 +566,109 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                            log.error('failed to compute empty checksum digest',
+                                { algorithm: algoName, error: err });
+                            return next(errors.InternalError, destBucketMD);
                         });
                 }
-                return next(null, storeMetadataParams, dataLocator, destObjMD,
-                    serverSideEncryption, destBucketMD);
+                return finishZeroByte();
+            }
+            if (shouldRecomputeChecksum) {
```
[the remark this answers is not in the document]
leif-scality: fixed

lib/api/objectCopy.js
```diff
@@ -468,30 +566,109 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                            // eslint-disable-next-line no-param-reassign
+                            request.sourceServerAccessLog.error = err;
+                        }
+                        return next(err, destBucketMD);
+                    }
+                    return next(null, storeMetadataParams, results, destObjMD, serverSideEncryption, destBucketMD);
+                });
+                const sourceStream = _pipeSourcePartsThrough(dataLocator, log);
+                const checksumStream = new ChecksumTransform(algoName, undefined, false, log);
+                sourceStream.once('error', done);
```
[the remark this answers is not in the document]
leif-scality: We don't do that kind of error defensive error handling

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -39,6 +39,12 @@ const errMPUTypeWithoutAlgo = errorInstances.InvalidRequest.customizeDescription
     'The x-amz-checksum-type header can only be used ' + 'with the x-amz-checksum-algorithm header.',
 );
 
+// TODO: Update with 'MD5', 'SHA512', 'XXHASH128', 'XXHASH3', 'XXHASH64' when they are introduced.
```
BourgoisMickael: you can add the ticket number in the comment if you have a ticket for this TODO

tcarmet (approved):

lib/api/objectCopy.js
```diff
@@ -28,13 +29,93 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+/**
+ * Decide whether the destination's checksum needs to be recomputed by
+ * streaming the source bytes through a ChecksumTransform.
+ *
+ * @param {object} headers - request headers
+ * @param {object} sourceObjMD - source object metadata
+ * @returns {boolean}
+ */
+function _shouldRecomputeChecksum(headers, sourceObjMD) {
+    const requestedAlgo = headers['x-amz-checksum-algorithm'] && headers['x-amz-checksum-algorithm'].toLowerCase();
```
BourgoisMickael: I think we can use optional chaining operator in cloudserver. 
```suggestion
    const requestedAlgo = headers['x-amz-checksum-algorithm']?.toLowerCase();
```

Maybe this can be used in other places as well to simplify

lib/api/objectCopy.js
```diff
@@ -28,13 +29,93 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+            // instead of returning a Readable. Pipe a per-part PassThrough
+            // into the master passthrough and use its 'end' as the completion
+            // signal — same pattern arsenal's data.copyObject uses.
+            const perPart = new PassThrough();
+            perPart.once('error', done);
+            perPart.once('end', () => done());
+            perPart.pipe(passthrough, { end: false });
+            return data.get(part, perPart, log, err => {
+                if (err) {
+                    done(err);
```
BourgoisMickael: When this error happens, it could be any error and the `perPart` is still piped.

So it could theoretically receive data from azure and continue to stream during a short time window until the final callback calls `passthrough.destroy` and unpipes the source `perPart`.

Maybe the `perPart` should be destroyed here to stop immediately the streaming

lib/api/objectCopy.js
```diff
@@ -28,13 +29,93 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+    const passthrough = new PassThrough();
+    async.eachSeries(dataLocator, (part, cb) => {
+        const done = jsutil.once(cb);
+        if (part.dataStoreType === 'azure') {
+            // Azure's data.get writes part bytes into the provided writable
+            // instead of returning a Readable. Pipe a per-part PassThrough
+            // into the master passthrough and use its 'end' as the completion
+            // signal — same pattern arsenal's data.copyObject uses.
+            const perPart = new PassThrough();
+            perPart.once('error', done);
```
BourgoisMickael: Should we consider encapsulating the error to include some part details in the error for better troubleshooting if the error is ever logged somewhere else in the streaming path.

lib/api/objectCopy.js
```diff
@@ -417,37 +513,39 @@ function objectCopy(authInfo, request, sourceBucket,
...
+            const shouldRecomputeChecksum = _shouldRecomputeChecksum(request.headers, sourceObjMD);
             // skip if source and dest and location constraint the same and
-            // versioning is not enabled
+            // versioning is not enabled, unless we need to recompute a
+            // checksum (which requires streaming the bytes through us)
             // still send along serverSideEncryption info so algo
             // and masterKeyId stored properly in metadata
             if (sourceIsDestination && storeMetadataParams.locationMatch
-                && !isVersionedObj && !needsEncryption) {
+                && !isVersionedObj && !needsEncryption && !shouldRecomputeChecksum) {
```
BourgoisMickael: If you don't skip, you'll trigger a data GET + PUT + DELETE (for previous location).

This only for checksum recomputation. Should you rather define another path where if it's only recompute checksum, you allow only the GET to stream into the ChecksumTransform stream and then you discard the end of the data stream, and avoid having to do a data PUT + DELETE

lib/api/objectCopy.js
```diff
@@ -583,17 +843,21 @@ function objectCopy(authInfo, request, sourceBucket,
...
+            sourceObjSize, destObjPrevSize, sourceObjMD, next) {
             // Clean up any potential orphans in data if object
             // put is an overwrite of already existing
             // object with same name, so long as the source is not
-            // the same as the destination
-            if (!sourceIsDestination && dataToDelete) {
+            // the same as the destination — but a copy-to-self that
+            // recomputes the checksum writes new data locations, so the
+            // previous data must still be reclaimed in that case.
+            if ((!sourceIsDestination
```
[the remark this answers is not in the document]
leif-scality: fixed

tests/unit/api/objectCopy.js
```diff
@@ -640,3 +644,635 @@ describe('objectCopy with objectKeyByteLimit', () => {
...
+                'response XML should not carry a checksum');
+            metadata.getObjectMD(destBucketName, objectKey, {}, log, (err, md) => {
+                assert.ifError(err);
+                assert(!md.checksum, 'destination metadata should not carry a checksum');
+                done();
+            });
+        });
+    });
+
+    it('should early-return on copy-to-self and not call data.copyObject or data.put', done => {
```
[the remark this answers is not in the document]
leif-scality: fixed

lib/api/objectCopy.js
```diff
@@ -28,13 +29,230 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+        checksumStream.once('error', finish);
+        checksumStream.on('data', () => {});
+        checksumStream.once('end', () => finish());
+        sourceStream.pipe(checksumStream);
+        return undefined;
+    }
+
+    // Stream source bytes through a ChecksumTransform and write them out as a single put.
+    log.debug('recomputing checksum on CopyObject', { algorithm: algoName, size: storeMetadataParams.size });
+    const done = jsutil.once((err, results) => {
```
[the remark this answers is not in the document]
leif-scality: fixed

lib/api/objectCopy.js
```diff
@@ -28,13 +29,221 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+    // change, not versioned): GET the source through a ChecksumTransform to
+    // compute the new digest, discard the bytes, and reuse the existing data
+    // locations. No PUT, no DELETE — only the metadata gets updated.
+    if (sourceIsDestination && storeMetadataParams.locationMatch
+        && !needsEncryption && !isVersionedObj) {
+        log.debug('computing checksum on copy-to-self without rewriting data',
+            { algorithm: algoName, size: storeMetadataParams.size });
+        const sourceStream = _pipeSourcePartsThrough(dataLocator, log);
+        const checksumStream = new ChecksumTransform(algoName, undefined, false, log);
+        const finish = jsutil.once(err => {
```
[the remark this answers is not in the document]
leif-scality: fixed

lib/api/objectCopy.js
leif-scality: Fixed I now Get and recompute but dont PUT and DELETE. I also removed recompute for external backend cases, so if the data does not go through cloudserver we dont GET, and the dest object receives no checksum if a recompute was required

lib/api/objectCopy.js
leif-scality: added a commit improving theerrors

lib/api/objectCopy.js
leif-scality: done

lib/api/apiUtils/integrity/validateChecksums.js

lib/api/objectCopy.js
leif-scality: This would make every thing more complex for little gain, lets just let error propagate naturally

lib/api/objectCopy.js
```diff
@@ -28,13 +30,246 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+    storeMetadataParams, destObjMD, destBucketMD, serverSideEncryption,
+    dataStoreContext, backendInfoDest, next) {
+    const algoName = _getCopyObjectChecksumAlgorithm(requestedAlgo, sourceObjMD);
+
+    // Copy-to-self where the bytes don't have to move (same location, no SSE
+    // change, not versioned): GET the source through a ChecksumTransform to
+    // compute the new digest, discard the bytes, and reuse the existing data
+    // locations. No PUT, no DELETE — only the metadata gets updated.
+    if (sourceIsDestination && storeMetadataParams.locationMatch
+        && !needsEncryption && !isVersionedObj) {
```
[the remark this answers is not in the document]
leif-scality: known issue will be fixed in another PR

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -930,3 +931,89 @@ describe('getChecksumDataFromMPUHeaders', () => {
...
+        assert.strictEqual(result.algorithm, 'sha256');
+    });
+
+    it('should accept algorithms already in lowercase', () => {
+        const result = getCopyObjectChecksumAlgorithm({
+            'x-amz-checksum-algorithm': 'crc32',
+        });
+        assert.strictEqual(result.error, null);
+        assert.strictEqual(result.algorithm, 'crc32');
+    });
```
BourgoisMickael: you could include these 2 inside the `validAlgorithms` array above to run in the loop, the test body is the same

tests/unit/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -930,3 +931,89 @@ describe('getChecksumDataFromMPUHeaders', () => {
...
+
+    it('should produce an arsenal InvalidRequest error via the standard mapping', () => {
+        const result = getCopyObjectChecksumAlgorithm({
+            'x-amz-checksum-algorithm': 'GARBAGE',
+        });
+        const err = arsenalErrorFromChecksumError(result.error);
+        assert.strictEqual(err.is.InvalidRequest, true);
+        assert(err.description.includes('Checksum algorithm provided is unsupported'),
+            'expected AWS-shaped error description');
+    });
```
BourgoisMickael: This doesn't seem to be related to `getCopyObjectChecksumAlgorithm` but rather `arsenalErrorFromChecksumError` maybe you should have another describe separate for that function

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: fixed

tests/unit/api/apiUtils/integrity/validateChecksums.js
leif-scality: fixed

lib/api/objectCopy.js
```diff
@@ -28,13 +29,253 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+                    perPart.destroy(err);
+                    done(wrapErr(err));
+                }
+            });
+        }
+        return data.get(part, null, log, (err, partStream) => {
+            if (err) {
+                return done(err);
+            }
+            partStream.once('error', done);
```
BourgoisMickael: the error is not wrapped here

lib/api/objectCopy.js
```diff
@@ -28,13 +29,253 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+    const passthrough = new PassThrough();
+    async.eachSeries(dataLocator, (part, cb) => {
+        const done = jsutil.once(cb);
+        if (part.dataStoreType === 'azure') {
+            // Azure's data.get writes part bytes into the provided writable
+            // instead of returning a Readable. Pipe a per-part PassThrough
+            // into the master passthrough and use its 'end' as the completion
+            // signal — same pattern arsenal's data.copyObject uses.
+            const perPart = new PassThrough();
+            const wrapErr = err => Object.assign(err, {
```
BourgoisMickael: You could pass the `part` as param to that function and define it once outside the `async.eachSeries`

lib/api/objectCopy.js
```diff
@@ -28,13 +29,253 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+                algorithm: algoName,
+                value: checksumStream.digest,
+                type: 'FULL_OBJECT',
+            };
+            return next(null, storeMetadataParams, dataLocator, destObjMD,
+                serverSideEncryption, destBucketMD);
+        });
+        sourceStream.once('error', finish);
+        checksumStream.once('error', finish);
+        checksumStream.on('data', () => {});
```
BourgoisMickael: It's not optimal to use a Transform only for the side effect of calculating checksum while discarding the output.
The Transform still works on pushing bytes into it's buffer.
Maybe we could rework the Checksum by implementing a new ChecksumWritable stream that does the same checksum computation but discards immediately the bytes without pushing.

lib/api/objectCopy.js
```diff
@@ -28,13 +29,253 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+            storeMetadataParams.checksum = {
+                algorithm: algoName,
+                value: checksumStream.digest,
+                type: 'FULL_OBJECT',
+            };
+            return next(null, storeMetadataParams, dataLocator, destObjMD,
+                serverSideEncryption, destBucketMD);
+        });
+        sourceStream.once('error', finish);
+        checksumStream.once('error', finish);
```
BourgoisMickael: Maybe this error should be wrapped as well to easily troubleshoot if an error is coming from the source or checksum stream

tests/unit/api/objectCopy.js
```diff
@@ -640,3 +645,961 @@ describe('objectCopy with objectKeyByteLimit', () => {
...
+        });
+    });
+
+    it('should reject an unknown x-amz-checksum-algorithm value with InvalidRequest', done => {
+        const req = _createObjectCopyRequest(destBucketName, {
+            'x-amz-checksum-algorithm': 'GARBAGE',
+        });
+        objectCopy(authInfo, req, sourceBucketName, objectKey, undefined, log, err => {
+            assert(err);
+            assert.strictEqual(err.is.InvalidRequest, true);
```
BourgoisMickael: not ideal to use boolean comparison with `.is`
```suggestion
            assert.strictEqual(err.message, 'InvalidRequest');
```

tests/unit/api/objectCopy.js
```diff
@@ -640,3 +645,961 @@ describe('objectCopy with objectKeyByteLimit', () => {
...
+        const dataPutSpy = sinon.spy(data, 'put');
+        Promise.resolve(algorithms.sha256.digest(objData[0])).then(expectedDigest => {
+            const req = _createObjectCopyRequest(destBucketName, {
+                'x-amz-checksum-algorithm': 'SHA256',
+                'x-amz-server-side-encryption': 'AES256',
+            });
+            objectCopy(authInfo, req, sourceBucketName, objectKey, undefined, log,
+                assertRecomputed('sha256', 'ChecksumSHA256', expectedDigest, err => {
+                    cipherBundleSpy.restore();
+                    dataPutSpy.restore();
```
BourgoisMickael: Ideally those restore should be in a hook `beforeEach` or `afterEach` to cover the case where an error is thrown before without executing this callback, that would leave the spy in place

tests/unit/api/objectCopy.js
```diff
@@ -640,3 +645,961 @@ describe('objectCopy with objectKeyByteLimit', () => {
...
+                    const req = _createObjectCopyRequest(destBucketName, {
+                        'x-amz-checksum-algorithm': 'SHA256',
+                    });
+                    objectCopy(authInfo, req, sourceBucketName, objectKey, undefined, log,
+                        (err, xml) => {
+                            dataGetStub.restore();
+                            assertRecomputed('sha256', 'ChecksumSHA256', expectedDigest, done)(err, xml);
+                        });
+                }, err => {
+                    dataGetStub.restore();
```
BourgoisMickael: Same here, the stub should be tracked outside this test to be restored in a hook to avoid missing it if any code path throws without reaching this callbacks

tests/unit/api/objectCopy.js
```diff
@@ -640,3 +645,961 @@ describe('objectCopy with objectKeyByteLimit', () => {
...
+                        headers: {
+                            'x-amz-checksum-algorithm': header,
+                            'x-amz-metadata-directive': 'REPLACE',
+                        },
+                        url: `/${sourceBucketName}/${objectKey}`,
+                        socket: {},
+                    });
+                    objectCopy(authInfo, req, sourceBucketName, objectKey, undefined, log, (err, xml) => {
+                        dataPutSpy.restore();
+                        batchDeleteSpy.restore();
```
BourgoisMickael: same here, the first `assert.ifError` from `metadata.getObjectMD` would not restore the spys

tests/unit/api/objectCopy.js
```diff
@@ -640,3 +645,961 @@ describe('objectCopy with objectKeyByteLimit', () => {
...
+            checksumAlgorithm: 'crc32',
+            checksumValue: 'AAAAAA==',
+            checksumType: 'FULL_OBJECT',
+        }, err => {
+            assert.ifError(err);
+            const req = _createObjectCopyRequest(destBucketName);
+            objectCopy(authInfo, req, sourceBucketName, objectKey, undefined, log, (err, xml) => {
+                assert.ifError(err);
+                assert(xml.includes('<ChecksumCRC32>AAAAAA==</ChecksumCRC32>'));
+                assert(xml.includes('<ChecksumType>FULL_OBJECT</ChecksumType>'));
```
BourgoisMickael: Prefer `assert.match` with a regex to see the full detail of the xml + expected pattern in error message, or include a message to the assert to provide more details about actual vs expected value

tests/unit/api/objectCopy.js
```diff
@@ -640,3 +645,961 @@ describe('objectCopy with objectKeyByteLimit', () => {
...
+                const req = new DummyRequest({
+                    bucketName: sourceBucketName,
+                    namespace,
+                    objectKey,
+                    headers: { 'x-amz-metadata-directive': 'REPLACE' },
+                    url: `/${sourceBucketName}/${objectKey}`,
+                    socket: {},
+                });
+                objectCopy(authInfo, req, sourceBucketName, objectKey, undefined, log, err => {
+                    batchDeleteSpy.restore();
```
BourgoisMickael: Overall many sinon spy/stub in many tests seems they could skip restore if some error branches is reached (here the `ifError` of `getObjectMD` and `putObjectMD`).

Maybe we should include in the claude review skill that in test files any mock/spy/stub should be tracked and cleaned in an `afterEach` hook to make sure every branch is covered. Otherwise one test failure might impact another test.

lib/api/objectCopy.js
leif-scality: I don't think it is worth it to create a ChecksumWritable just to avoid a NOP. The checksum transform already handles everything and it is already tested

lib/api/objectCopy.js
leif-scality: done

lib/api/objectCopy.js
leif-scality: done

lib/api/objectCopy.js
leif-scality: done

tests/unit/api/objectCopy.js
leif-scality: fixed, and added a rule to the review skill to detect this issue

tests/unit/api/objectCopy.js
leif-scality: fixed

tests/unit/api/objectCopy.js
leif-scality: fixed

tests/unit/api/objectCopy.js
leif-scality: fixed

tests/unit/api/objectCopy.js
leif-scality: fixed, and added a rule to the review skill to detect this issue

tests/unit/api/objectCopy.js
leif-scality: fixed

lib/api/objectCopy.js
BourgoisMickael: This is not NOP, the `_transform` when passing the input to the callback (https://github.com/scality/cloudserver/blob/2679e31cd7e43790277153ae6e7c64072df80318/lib/auth/streamingV4/ChecksumTransform.js#L90) triggers a `push` of that data into the readable side of the stream.

You then have CPU and memory impact and perf impact of running the readable side of the transform, only to discard it at the end:
- send the data into the readable stream internal buffer
- internal accounting of buffer size
- schedule events in the event loop (for data listener and other stream event)
- invoke the empty callback of data listener
-  now data is discarded, and GC of the data is delayed until now

---


Ideally the input should be discarded asap to avoid unnecessary processing.

A simpler solution to avoid reworking the stream and avoid implementing another ChecksumWritable stream could be to pass a `discard` boolean flag to the ChecksumTransform and conditionally pass the input to the `_transform` callback

```js
callback(null, this.discard ? undefined : input);
```

leif-scality: added the writable

BourgoisMickael (approved):

lib/api/objectCopy.js
```diff
@@ -28,12 +26,280 @@ const { verifyColdObjectAvailable } = require('./apiUtils/object/coldStorage');
...
+const ChecksumWritable = require('../auth/streamingV4/ChecksumWritable');
+const kms = require('../kms/wrapper');
 
 const versionIdUtils = versioning.VersionID;
 const locationHeader = constants.objectLocationConstraintHeader;
 const versioningNotImplBackends = constants.versioningNotImplBackends;
-const externalVersioningErrorMessage = 'We do not currently support putting ' +
-'a versioned object to a location-constraint of type AWS or Azure or GCP.';
+const externalVersioningErrorMessage =
+    'We do not currently support putting ' + 'a versioned object to a location-constraint of type AWS or Azure or GCP.';
```
BourgoisMickael:
```suggestion
    'We do not currently support putting a versioned object to a location-constraint of type AWS or Azure or GCP.';
```

lib/api/objectCopy.js
```diff
@@ -510,23 +534,21 @@ function objectCopy(authInfo, request, sourceBucket,
...
+        monitoring.promMetrics('PUT', destBucketName, errorInstances.InvalidStorageClass.code, 'copyObject');
         return callback(errors.InvalidStorageClass);
     }
     if (!validateWebsiteHeader(websiteRedirectHeader)) {
         const err = errors.InvalidRedirectLocation;
-        log.debug('invalid x-amz-website-redirect-location' +
-            `value ${websiteRedirectHeader}`, { error: err });
-        monitoring.promMetrics(
-            'PUT', destBucketName, err.code, 'copyObject');
+        log.debug('invalid x-amz-website-redirect-location' + `value ${websiteRedirectHeader}`, { error: err });
```
BourgoisMickael: string concatenation here

lib/api/objectCopy.js
```diff
@@ -540,440 +562,609 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                            log,
+                            { skipObject: true },
+                            (err, destBucketMD, destObjMD) => {
+                                if (err) {
+                                    log.debug('error validating put part of request', { error: err });
+                                    return next(err, destBucketMD);
+                                }
+                                const flag = destBucketMD.hasDeletedFlag() || destBucketMD.hasTransientFlag();
+                                if (flag) {
+                                    log.trace('deleted flag or transient flag ' + 'on destination bucket', { flag });
```
BourgoisMickael: string concat here

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -156,221 +171,326 @@ describe('Object Copy', () => {
...
+                    });
+            },
+        );
 
-        it('should return 400 InvalidArgument if invalid tagging ' +
-        'directive', done => {
-            s3.send(new CopyObjectCommand({ Bucket: destBucketName, Key: destObjName,
-                CopySource: `${sourceBucketName}/${sourceObjName}`,
-                TaggingDirective: 'COCO' })).then(() => {
+        it('should return 400 InvalidArgument if invalid tagging ' + 'directive', done => {
```
BourgoisMickael: string concat

lib/api/objectCopy.js
leif-scality: fixed

lib/api/objectCopy.js
leif-scality: fixed

lib/api/objectCopy.js
leif-scality: fixed

leif-scality:
ping

PR #6177, opened 2026-05-28
CLDSRV-909: Reject CopyObject when source exceeds 5 GiB

tcarmet:
Align cloudserver's CopyObject with the AWS S3 contract by rejecting requests whose source object exceeds 5 GiB. AWS S3 returns HTTP 400 `InvalidRequest` ("The specified copy source is larger than the maximum allowable size for a copy source: 5368709120") and expects clients to use the multipart copy flow instead; cloudserver currently accepts these requests, leaving a gap with the published contract.

Targeting `development/9.4` because changing a previously-accepted response is a behavior break for clients that relied on cloudserver's permissive behavior, happy to retarget `development/9.3` if we prefer wider rollout.

Last commit contains a prettier fmt command and is kept separate from the actual code change for review. We should not focus on it as most of it is existing change.

### DIFF
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -641,6 +641,17 @@ function objectCopy(authInfo, request, sourceBucket, sourceObject, sourceVersion
                             request.sourceServerAccessLog && (request.sourceServerAccessLog.error = err);
                             return next(err, destBucketMD);
                         }
+                        const sourceSize = parseInt(sourceObjMD['content-length'], 10);
+                        if (sourceSize > constants.maximumAllowedUploadSize) {
+                            log.debug('copy source object too large', { sourceSize });
+                            const err = errorInstances.InvalidRequest.customizeDescription(
+                                'The specified copy source is larger than the maximum ' +
+                                    `allowable size for a copy source: ${constants.maximumAllowedUploadSize}`,
+                            );
+                            // eslint-disable-next-line no-param-reassign
+                            request.sourceServerAccessLog && (request.sourceServerAccessLog.error = err);
+                            return next(err, destBucketMD);
+                        }
                         const headerValResult = validateHeaders(
                             request.headers,
                             sourceObjMD['last-modified'],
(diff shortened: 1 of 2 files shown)


lib/api/objectCopy.js
```diff
@@ -351,6 +351,16 @@ function objectCopy(authInfo, request, sourceBucket,
                         request.sourceServerAccessLog && (request.sourceServerAccessLog.error = err);
                         return next(err, destBucketMD);
                     }
+                    const sourceSize = parseInt(sourceObjMD['content-length'], 10);
+                    if (sourceSize > constants.maximumAllowedUploadSize) {
```
BourgoisMickael:
```suggestion
                    if (sourceSize > constants.maximumAllowedUploadSize && !config.bypassMaxPutObjectSize) {
```

Same as https://github.com/scality/cloudserver/pull/5935 use a flag to let customer disable this limitation

tcarmet: nice, thank you for catching that.

lib/api/objectCopy.js
```diff
@@ -351,6 +351,16 @@ function objectCopy(authInfo, request, sourceBucket,
...
+                        log.debug('copy source object too large', { sourceSize });
+                        const err = errorInstances.InvalidRequest.customizeDescription(
+                            'The specified copy source is larger than the maximum ' +
+                            `allowable size for a copy source: ${constants.maximumAllowedUploadSize}`);
+                        // eslint-disable-next-line no-param-reassign
+                        request.sourceServerAccessLog && (request.sourceServerAccessLog.error = err);
```
jonathan-gramain: This is a strange construct, I would rather do a plain `if` statement

jonathan-gramain (approved):

lib/api/objectCopy.js
tcarmet: it is, but what I'm doing here is following the same pattern found across the file/project. I could go for a if and leave a bit of inconsistency or change it as whole. What I went with now is just consistency.

jonathan-gramain: ok, generally speaking I tend to prefer having an inconsistent but incrementally correct style than a consistently subpar style, but that's my take, do as you think is best!

tcarmet: I'm fine with that, I think I needed an extra opinion to make a call so thanks, will have a look into it.

dvasilas (approved):

PR #6178, opened 2026-05-29
Bump microVersionId and add isReplica handling

maeldonn:
Cloudserver-side prep for cascaded CRR ([design](https://github.com/scality/citadel/blob/development/1.0/docs/design/replication/crr-cascaded.md)): bumps arsenal to the `improvement/ARSN-578/micro-version-id` branch, bumps `microVersionId` on every user-facing object metadata write through a single `bumpMicroVersionId` helper (gated on `replicationInfo` presence to keep storage tight, with a `force` flag preserving the existing overwriting-version path), sources `x-amz-replication-status` from the new `ReplicationInfo.isReplica` field with a legacy `status === 'REPLICA'` fallback, and clears `isReplica` on direct user writes (tagging / ACL / retention / legal hold) so a modified replica stops looking like one. The backbeat-route side of the design (setting `isReplica = true` on incoming replicas, cascade comparison, MultiBackend guard) lands in a follow-up ticket — until then the new fields are read but never set, so observable behavior is unchanged.

Issue: CLDSRV-906

### DIFF
--- a/lib/api/apiUtils/object/bumpMicroVersionId.js
+++ b/lib/api/apiUtils/object/bumpMicroVersionId.js
@@ -1,0 +1,27 @@
+const { versioning } = require('arsenal');
+const { config } = require('../../../Config');
+
+/**
+ * Bump objectMD.microVersionId. microVersionId is a generic
+ * metadata-revision marker, not a CRR-specific field, but cascaded CRR
+ * is its only consumer today - so we gate on replicationInfo to avoid
+ * inflating storage on objects that wouldn't use it. The gate can be
+ * widened later if another consumer needs it on every object.
+ * Pass `force = true` to bump unconditionally.
+ *
+ * @param {object} objectMD - object MD POJO or `md.getValue()`
+ * @param {boolean} [force] - bump even without replicationInfo
+ * @return {undefined}
+ */
+function bumpMicroVersionId(objectMD, force) {
+    if (!force && !objectMD?.replicationInfo) {
+        return;
+    }
+
+    const { instanceId, replicationGroupId } = config;
+
+    // eslint-disable-next-line no-param-reassign
+    objectMD.microVersionId = versioning.VersionID.generateVersionId(instanceId, replicationGroupId);
+}
+
+module.exports = bumpMicroVersionId;
--- a/lib/api/apiUtils/object/getReplicationInfo.js
+++ b/lib/api/apiUtils/object/getReplicationInfo.js
@@ -100,6 +100,8 @@ function getReplicationInfo(s3config, objKey, bucketMD, isMD, objSize, operation
         content,
         role: ReplicationConfiguration.resolveSourceRole(config.role),
         isNFS: bucketMD.isNFS(),
+        // `undefined` so JSON/BSON drop the field on persist (saves bytes).
+        isReplica: undefined,
     };
 }
 
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -10,8 +10,7 @@ const { scaledMsPerDay } = config.getTimeOptions();
 const versionIdUtils = versioning.VersionID;
 // Use Arsenal function to generate a version ID used internally by metadata
 // for null versions that are created before bucket versioning is configured
-const nonVersionedObjId =
-    versionIdUtils.getInfVid(config.replicationGroupId);
+const nonVersionedObjId = versionIdUtils.getInfVid(config.replicationGroupId);
 
 /** decodeVID - decode the version id
  * @param {string} versionId - version ID
@@ -101,25 +100,24 @@ function _storeNullVersionMD(bucketName, objKey, nullVersionId, objMD, log, cb) 
     nullVersionMD.originOp = 's3:StoreNullVersion';
     metadata.putObjectMD(bucketName, objKey, nullVersionMD, { versionId }, log, err => {
         if (err) {
-            log.debug('error from metadata storing null version as new version',
-            { error: err });
+            log.debug('error from metadata storing null version as new version', { error: err });
         }
-        
+
         cb(err);
     });
 }
 
 /** check existence and get location of null version data for deletion
-* @param {string} bucketName - name of bucket
-* @param {string} objKey - name of object key
-* @param {object} options - metadata options for getting object MD
-* @param {string} options.versionId - version to get from metadata
-* @param {object} mst - info about the master version
-* @param {string} mst.versionId - the master version's version id
-* @param {RequestLogger} log - logger instanceof
-* @param {function} cb - callback
-* @return {undefined} - and call callback with (err, dataToDelete)
-*/
+ * @param {string} bucketName - name of bucket
+ * @param {string} objKey - name of object key
+ * @param {object} options - metadata options for getting object MD
+ * @param {string} options.versionId - version to get from metadata
+ * @param {object} mst - info about the master version
+ * @param {string} mst.versionId - the master version's version id
+ * @param {RequestLogger} log - logger instanceof
+ * @param {function} cb - callback
+ * @return {undefined} - and call callback with (err, dataToDelete)
+ */
 function _prepareNullVersionDeletion(bucketName, objKey, options, mst, log, cb) {
     const nullOptions = {};
     if (!options.deleteData) {
@@ -135,38 +133,40 @@ function _prepareNullVersionDeletion(bucketName, objKey, options, mst, log, cb) 
         // PUT via this option
         nullOptions.deleteNullKey = true;
     }
-    return metadata.getObjectMD(bucketName, objKey, options, log,
-        (err, versionMD) => {
-            if (err) {
-                // the null key may not exist, hence it's a normal
-                // situation to have a NoSuchKey error, in which case
-                // there is nothing to delete
-                if (err.is.NoSuchKey) {
-                    log.debug('null version does not exist', {
-                        method: '_prepareNullVersionDeletion',
-                    });
-                } else {
-                    log.warn('could not get null version metadata', {
-                        error: err,
-                        method: '_prepareNullVersionDeletion',
-                    });
-                }
-                return cb(err);
+    return metadata.getObjectMD(bucketName, objKey, options, log, (err, versionMD) => {
+        if (err) {
+            // the null key may not exist, hence it's a normal
+            // situation to have a NoSuchKey error, in which case
+            // there is nothing to delete
+            if (err.is.NoSuchKey) {
+                log.debug('null version does not exist', {
+                    method: '_prepareNullVersionDeletion',
+                });
+            } else {
+                log.warn('could not get null version metadata', {
+                    error: err,
+                    method: '_prepareNullVersionDeletion',
+                });
             }
-            if (versionMD.location) {
-                const dataToDelete = Array.isArray(versionMD.location) ?
-                      versionMD.location : [versionMD.location];
-                nullOptions.dataToDelete = dataToDelete;
-            }
-            return cb(null, nullOptions);
-        });
+            return cb(err);
+        }
+        if (versionMD.location) {
+            const dataToDelete = Array.isArray(versionMD.location) ? versionMD.location : [versionMD.location];
+            nullOptions.dataToDelete = dataToDelete;
+        }
+        return cb(null, nullOptions);
+    });
 }
 
 function _deleteNullVersionMD(bucketName, objKey, options, log, cb) {
     return metadata.deleteObjectMD(bucketName, objKey, options, log, err => {
         if (err) {
-            log.warn('metadata error deleting null versioned key',
-                     { bucketName, objKey, error: err, method: '_deleteNullVersionMD' });
+            log.warn('metadata error deleting null versioned key', {
+                bucketName,
+                objKey,
+                error: err,
+                method: '_deleteNullVersionMD',
+            });
         }
         return cb(err);
     });
@@ -193,7 +193,7 @@ function _deleteNullVersionMD(bucketName, objKey, options, log, cb) {
      version key, if needed
  */
 function processVersioningState(mst, vstat, nullVersionCompatMode) {
-    const versioningSuspended = (vstat === 'Suspended');
+    const versioningSuspended = vstat === 'Suspended';
     const masterIsNull = mst.exists && (mst.isNull || !mst.versionId);
 
     if (versioningSuspended) {
@@ -244,7 +244,7 @@ function processVersioningState(mst, vstat, nullVersionCompatMode) {
     if (masterIsNull) {
         // if master is a null version or a non-versioned key,
         // copy it to a new null key
-        const nullVersionId = (mst.isNull && mst.versionId) ? mst.versionId : nonVersionedObjId;
+        const nullVersionId = mst.isNull && mst.versionId ? mst.versionId : nonVersionedObjId;
         if (nullVersionCompatMode) {
             options.extraMD = {
                 nullVersionId,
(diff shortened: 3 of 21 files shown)


lib/api/objectDeleteTagging.js
```diff
@@ -76,13 +77,18 @@ function objectDeleteTagging(authInfo, request, log, callback) {
...
+                objectMD.replicationInfo.isReplica = false;
+            }
             const replicationInfo = getReplicationInfo(config,
                 objectKey, bucket, true, 0, REPLICATION_ACTION, objectMD);
             if (replicationInfo) {
                 // eslint-disable-next-line no-param-reassign
                 objectMD.replicationInfo = Object.assign({},
                     objectMD.replicationInfo, replicationInfo);
             }
+            bumpMicroVersionId(objectMD);
```
francoisferrand:
```suggestion
		const replicationInfo = getReplicationInfo(config, objectKey, bucket, true, 0, REPLICATION_ACTION, objectMD);
		if (replicationInfo) {
				// eslint-disable-next-line no-param-reassign
				objectMD.replicationInfo = Object.assign({},
				objectMD.replicationInfo, replicationInfo);

				if (objectMD.replicationInfo?.isReplica) {
	                // eslint-disable-next-line no-param-reassign
    	            objectMD.replicationInfo.isReplica = false;
    	         }

    	         bumpMicroVersionId(objectMD);
		}
```

francoisferrand: (same in all files...)

lib/api/objectPutLegalHold.js
```diff
@@ -87,13 +88,18 @@ function objectPutLegalHold(authInfo, request, log, callback) {
             // eslint-disable-next-line no-param-reassign
             objectMD.legalHold = legalHold;
             const params = getVersionSpecificMetadataOptions(objectMD, config.nullVersionCompatMode);
+            if (objectMD.replicationInfo?.isReplica) {
+                // eslint-disable-next-line no-param-reassign
+                objectMD.replicationInfo.isReplica = false;
+            }
```
francoisferrand: should this not be handled in `getReplicationInfo` ? (I think this function is not really a getter, but really "builds" the new replication info...)

lib/api/apiUtils/object/bumpMicroVersionId.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+function bumpMicroVersionId(objectMD, force) {
+    if (!force && !objectMD?.replicationInfo) {
+        return;
+    }
+
+    const { instanceId, replicationGroupId } = config;
+
+    // eslint-disable-next-line no-param-reassign
+    objectMD.microVersionId = versioning.VersionID.generateVersionId(instanceId, replicationGroupId);
```
francoisferrand, line 24: isn't there a method `ObjectMD.updateMicroVersionId()` in arsenal ? should it not be used instead?

(if we can't because of POJO, then I wonder if that function was really useful....)

lib/services.js
```diff
@@ -190,9 +191,7 @@ const services = {
             options.replayId = uploadId;
         }
         // update microVersionId when overwriting metadata.
-        if (updateMicroVersionId) {
-            md.updateMicroVersionId();
-        }
+        bumpMicroVersionId(md.getValue(), updateMicroVersionId);
```
francoisferrand: - is this not redundant with the other calls in APIs ?
- calling it here we ALWAYS generate a new microVersionId when _creating_ a new object: which is not needed since it comes with a new versionId... (except for non-versioned case, but no replication there: so the function does nothing)

package.json
```diff
@@ -33,7 +33,7 @@
         "@azure/storage-blob": "^12.28.0",
         "@hapi/joi": "^17.1.1",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/Arsenal#8.4.2",
+        "arsenal": "git+https://github.com/scality/Arsenal#improvement/ARSN-578/micro-version-id",
```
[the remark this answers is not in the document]
maeldonn: Waiting for arsenal release. Will be updated before merge.

tests/unit/api/apiUtils/object/bumpMicroVersionId.js
```diff
@@ -0,0 +1,31 @@
+const assert = require('assert');
+
+const bumpMicroVersionId = require('../../../../../lib/api/apiUtils/object/bumpMicroVersionId');
+
+describe('bumpMicroVersionId', () => {
+    it('sets a fresh microVersionId when replicationInfo is present', () => {
```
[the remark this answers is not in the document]
maeldonn: Updated.

tests/unit/api/objectReplicationMD.js
```diff
@@ -718,3 +719,55 @@ describe('Replication object MD without bucket replication config', () => {
...
+    const objectPutTaggingAsync = promisify(objectPutTagging);
+
+    beforeEach(() => {
+        cleanup();
+        createBucket();
+    });
+
+    afterEach(() => cleanup());
+
+    it('sets microVersionId on objectPut', async () => {
```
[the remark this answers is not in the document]
maeldonn: Updated.

package.json
```diff
@@ -33,7 +33,7 @@
         "@azure/storage-blob": "^12.28.0",
         "@hapi/joi": "^17.1.1",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/Arsenal#8.4.4",
+        "arsenal": "git+https://github.com/scality/Arsenal#2c429ab35a5ac82c3dafa5a0296a49a23a9c8a4a",
```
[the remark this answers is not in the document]
maeldonn: Yes , arsenal is not yet released.

package.json
```diff
@@ -33,7 +33,7 @@
         "@azure/storage-blob": "^12.28.0",
         "@hapi/joi": "^17.1.1",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/Arsenal#8.4.4",
+        "arsenal": "git+https://github.com/scality/Arsenal#improvement/ARSN-578/micro-version-id",
```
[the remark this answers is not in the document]
maeldonn: Yes, arsenal is not yet released.

lib/api/objectDeleteTagging.js
maeldonn: Updated.

lib/api/objectPutLegalHold.js
maeldonn: Yes, very clean.

lib/api/apiUtils/object/bumpMicroVersionId.js
maeldonn: `ObjectMD.updateMicroVersionId()` is used by s3utils that holds `ObjectMD` instances. Cloudserver's call sites here all hold POJOs from `metadata.getObjectMD`, so we use the lower-level `versioning.VersionID.generateVersionId` (which is exactly what `updateMicroVersionId` calls internally). Both arsenal entry points are valid for their respective consumer shapes.

package.json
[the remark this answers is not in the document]
maeldonn: arsenal is not yet released.

lib/api/apiUtils/object/versioning.js
```diff
@@ -523,7 +528,10 @@ function restoreMetadata(objMD, metadataStoreParams) {
  *  version id of the null version
  */
 function overwritingVersioning(objMD, metadataStoreParams) {
-    metadataStoreParams.updateMicroVersionId = true;
+    metadataStoreParams.microVersionId = versionIdUtils.generateVersionId(
```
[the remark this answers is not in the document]
maeldonn: Fixed.

delthas (approved):

lib/utilities/collectResponseHeaders.js
```diff
@@ -89,8 +89,11 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg, returnTagC
     if (objectMD.legalHold !== undefined) {
         responseMetaHeaders['x-amz-object-lock-legal-hold'] = objectMD.legalHold ? 'ON' : 'OFF';
     }
-    if (objectMD.replicationInfo && objectMD.replicationInfo.status) {
-        responseMetaHeaders['x-amz-replication-status'] = objectMD.replicationInfo.status;
+    if (objectMD.replicationInfo) {
+        const { isReplica, status } = objectMD.replicationInfo;
+        if (isReplica || status) {
```
francoisferrand, line 94: do we need that condition?
- It was not present before...
- Should we not always have a `x-amz-replication-status` if `objectMD.replicationInfo` is set?

package.json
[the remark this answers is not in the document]
maeldonn: Arsenal is not released.

lib/utilities/collectResponseHeaders.js
maeldonn: The guard existed before as `if (objectMD.replicationInfo && objectMD.replicationInfo.status)`. I just split it to destructure both fields. Keeping it avoids emitting `x-amz-replication-status: undefined` for legacy or partially-cleared metadata where `status` is missing and `isReplica` is false.

lib/api/apiUtils/object/bumpMicroVersionId.js
SylvainSenechal, line 1: nit : I think putting it in versioning.js instead of creating a new file could have been enough 🤔

lib/api/apiUtils/object/getReplicationInfo.js
```diff
@@ -100,6 +100,7 @@ function getReplicationInfo(s3config, objKey, bucketMD, isMD, objSize, operation
         content,
         role: ReplicationConfiguration.resolveSourceRole(config.role),
         isNFS: bucketMD.isNFS(),
+        isReplica: false,
```
SylvainSenechal: mixed feeling on this : I say let's keep it, it makes sense as we return status pending, but also weird because we have objectMd available in the function, so we should return objectMd.isReplica 🤔

lib/utilities/collectResponseHeaders.js
SylvainSenechal: ok for me

lib/api/objectDeleteTagging.js
```diff
@@ -93,6 +94,7 @@ function objectDeleteTagging(authInfo, request, log, callback) {
                 if (replicationInfo) {
                     // eslint-disable-next-line no-param-reassign
                     objectMD.replicationInfo = Object.assign({}, objectMD.replicationInfo, replicationInfo);
+                    bumpMicroVersionId(objectMD);
```
SylvainSenechal, line 97: On my implementation, i was also resetting isReplica to false during these metadata updates, shouldn't we do it ?

lib/services.js
```diff
@@ -220,9 +220,8 @@ const services = {
             md.setUploadId(uploadId);
             options.replayId = uploadId;
         }
-        // update microVersionId when overwriting metadata.
         if (updateMicroVersionId) {
-            md.updateMicroVersionId();
+            md.updateMicroVersionId(config.instanceId, config.replicationGroupId);
```
SylvainSenechal, line 224: On my version, i removed the if statement, and the params update
microVersionId.

I found this param was only set in overwritingVersioning() at the beginning : 
which I also removed.

I did it because I think microVersionId should always be updated when metadataStoreObject is called what do you think 🤔

SylvainSenechal: not fully sure though, because maybe we wanna avoid useless microVersionId storage in mondoDb, and also bump microVersionId on more specific path ?

lib/api/apiUtils/object/bumpMicroVersionId.js
maeldonn: I'd rather keep it as its own file: `versioning.js` is already 585 lines and handles `versionId` semantics, whereas `microVersionId` is a separate revision marker (used by cascaded CRR loop detection). Six call sites import it, so a focused helper file felt clearer than growing `versioning.js`. Happy to move it if you feel strongly though.

lib/api/apiUtils/object/getReplicationInfo.js
maeldonn: Intentional, same as hardcoding `status: 'PENDING'`: this builds a fresh outgoing `replicationInfo` for a user write, which by definition isn't a replica. The hardcoded `false` is precisely what flips the previous `isReplica: true` to `false` when merged onto `objectMD.replicationInfo` at the call sites. Reading `objectMD.isReplica` would defeat that.

lib/api/objectDeleteTagging.js
maeldonn: Already handled: `getReplicationInfo` returns `isReplica: false` and the `Object.assign` merge applies it.

lib/services.js
maeldonn: `metadataStoreObject` is called by multiple operations (PUT, COPY, MPU complete, etc.) where we don't want to bump `microVersionId`. Keeping the existing gated behavior.

For now we want to avoid unnecessary metadata writes since `microVersionId` is only consumed by CRR. That doesn't mean it's coupled to CRR conceptually, just that there's no other consumer today, so bumping it unconditionally would be wasted writes.

lib/api/apiUtils/object/getReplicationInfo.js
SylvainSenechal: true, other fields are also hardcoded

SylvainSenechal (approved):

lib/api/apiUtils/object/getReplicationInfo.js
francoisferrand: nit: since we need to keep "backwards" compatibility (i.e. support the missing `isReplica` field), can we / should we skip this field when set to `false`, to spare a few bytes in mongo?

lib/utilities/collectResponseHeaders.js
```diff
@@ -89,8 +89,11 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg, returnTagC
...
+            responseMetaHeaders['x-amz-replication-status'] = isReplica ? 'REPLICA' : status;
```
francoisferrand, line 95: it's been discussed in the design, but not sure about this now: when doing cascaded replication, we may want to know that we are actually still replicating (to the cascaded destination), and even possibly that replication is complete/... (i.e. "report" replica only when we detected the loop / have not further replication pending)

maybe we should also provide a (non-standard) way to _display_ the status of each destination and indicate if we are a replica or not.

→ let's no discuss this further here, but create a followup ticket for this so we confirm the solution (and adjust if needed) before shipping the feature

francoisferrand (approved):

package.json
```diff
@@ -37,7 +37,7 @@
         "@opentelemetry/instrumentation-ioredis": "~0.64.0",
         "@opentelemetry/instrumentation-mongodb": "~0.69.0",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/Arsenal#8.4.7",
+        "arsenal": "git+https://github.com/scality/Arsenal#f6b6e2a37675e76186326e22c17b7a4cef77f5e0",
```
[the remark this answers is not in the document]
maeldonn: Not yet released.

lib/utilities/collectResponseHeaders.js
maeldonn: Create a follow-up ticket: https://scality.atlassian.net/browse/CLDSRV-931.

package.json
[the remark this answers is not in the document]
maeldonn: not yet released.

lib/api/apiUtils/object/getReplicationInfo.js
maeldonn: Done. `getReplicationInfo` now returns `isReplica: undefined` instead of `false`, so JSON/BSON drop the field on persist (file, bucketclient, in-memory backends via `JSON.stringify`; mongo via `ignoreUndefined: true` on the full-subdocument `$set` write). `Object.assign` still copies the undefined value at call sites, which clears any prior `isReplica: true` in memory before persist.

PR #6179, opened 2026-05-29
Add CRR Cascaded capabilities

SylvainSenechal:
ISSUE : CLDSRV-897

Crr cascaded design : https://github.com/scality/citadel/pull/349

Related PRs :
Arsenal : https://github.com/scality/Arsenal/pull/2628
CloudserverClient : https://github.com/scality/cloudserverclient/pull/24
Backbeat : https://github.com/scality/backbeat/pull/2747
S3utils : https://github.com/scality/s3utils/pull/395

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -1,4 +1,6 @@
-const { constants: { HTTP_STATUS_CONFLICT } } = require('http2');
+const {
+    constants: { HTTP_STATUS_CONFLICT },
+} = require('http2');
 const url = require('url');
 const async = require('async');
 const httpProxy = require('http-proxy');
@@ -143,7 +145,9 @@ function _respondWithHeaderCrrConflict(response, log, callback, code, message, m
         response,
         { code, message },
         { 'x-scal-micro-version-id': mvId ? encode(mvId) : '' },
-        log, callback, HTTP_STATUS_CONFLICT,
+        log,
+        callback,
+        HTTP_STATUS_CONFLICT,
     );
 }
 
@@ -435,16 +439,16 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
 
     const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
     if (incomingVersionIdEncoded !== undefined) {
-        const incomingVersionIdDecoded = incomingVersionIdEncoded !== 'null'
-            ? decode(incomingVersionIdEncoded)
-            : 'null';
+        const incomingVersionIdDecoded =
+            incomingVersionIdEncoded !== 'null' ? decode(incomingVersionIdEncoded) : 'null';
         if (incomingVersionIdDecoded instanceof Error) {
             log.error('crr putData: failed to decode x-scal-version-id header', {
                 method: 'putData',
                 error: incomingVersionIdDecoded.message,
             });
-            return callback(errorInstances.BadRequest.customizeDescription(
-                'bad request: invalid x-scal-version-id header'));
+            return callback(
+                errorInstances.BadRequest.customizeDescription('bad request: invalid x-scal-version-id header'),
+            );
         }
         if (objMd && objMd.versionId === incomingVersionIdDecoded) {
             // Data already at destination for this version; return 409 with the existing
@@ -457,7 +461,9 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
             });
             request.resume();
             return _respondWithHeaderCrrConflict(
-                response, log, callback,
+                response,
+                log,
+                callback,
                 VersionIdCollisionException.name,
                 'version id already at destination',
                 objMd.microVersionId,
@@ -608,13 +614,15 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 method: 'putMetadata',
                 error: incomingMicroVersionId.message,
             });
-            return callback(errorInstances.BadRequest.customizeDescription(
-                'bad request: invalid x-scal-micro-version-id header'));
+            return callback(
+                errorInstances.BadRequest.customizeDescription('bad request: invalid x-scal-micro-version-id header'),
+            );
         }
 
         const isIncomingOlderThanCurrent = (incoming, current) => {
-            if (current === null) { // nothing to be stale against
-                return false; 
+            if (current === null) {
+                // nothing to be stale against
+                return false;
             }
             // null incoming : source has no cascade revision => oldest possible state
             // larger string = older revision (reverse-chronological order)
@@ -643,7 +651,9 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             });
             request.resume();
             return _respondWithHeaderCrrConflict(
-                response, log, callback,
+                response,
+                log,
+                callback,
                 conflictErr.err,
                 conflictErr.message,
                 conflictErr.mvId,
@@ -669,8 +679,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         }
 
         if (incomingMicroVersionId !== null && incomingMicroVersionId !== (omVal.microVersionId ?? null)) {
-            return callback(errors.BadRequest.customizeDescription(
-                'bad request: x-scal-micro-version-id header does not match body microVersionId'));
+            return callback(
+                errors.BadRequest.customizeDescription(
+                    'bad request: x-scal-micro-version-id header does not match body microVersionId',
+                ),
+            );
         }
 
         const { headers } = request;
@@ -850,28 +863,41 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             const objSize = omVal['content-length'] || 0;
 
             const nextReplInfo = getReplicationInfo(
-                config, objectKey, bucketInfo, isMDOnly, objSize,
-                null, null, null, constants.crrCascadeBlockedLocationTypes);
+                config,
+                objectKey,
+                bucketInfo,
+                isMDOnly,
+                objSize,
+                null,
+                null,
+                null,
+                constants.crrCascadeBlockedLocationTypes,
+            );
 
             const hasNextHop = nextReplInfo && nextReplInfo.backends.length > 0;
 
             // Replicating further requires x-scal-micro-version-id for loop detection.
             if (hasNextHop && !hasMicroVersionId) {
-                return callback(errors.InternalError.customizeDescription(
-                    'x-scal-micro-version-id is required when replication would trigger a next hop'));
+                return callback(
+                    errors.InternalError.customizeDescription(
+                        'x-scal-micro-version-id is required when replication would trigger a next hop',
+                    ),
+                );
             }
 
-            omVal.replicationInfo = hasNextHop ? nextReplInfo : {
-                status: '',
-                backends: [],
-                content: [],
-                destination: undefined,
-                storageClass: undefined,
-                role: undefined,
-                storageType: undefined,
-                dataStoreVersionId: undefined,
-                isNFS: undefined,
-            };
+            omVal.replicationInfo = hasNextHop
+                ? nextReplInfo
+                : {
+                      status: '',
+                      backends: [],
+                      content: [],
+                      destination: undefined,
+                      storageClass: undefined,
+                      role: undefined,
+                      storageType: undefined,
+                      dataStoreVersionId: undefined,
+                      isNFS: undefined,
+                  };
             omVal.replicationInfo.isReplica = true;
         }
 
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -1,3 +1,4 @@
+const { constants: { HTTP_STATUS_CONFLICT } } = require('http2');
 const url = require('url');
 const async = require('async');
 const httpProxy = require('http-proxy');
@@ -7,7 +8,13 @@ const joi = require('@hapi/joi');
 const backbeatProxy = httpProxy.createProxyServer({
     ignorePath: true,
 });
-const { auth, errors, errorInstances, s3middleware, s3routes, models, storage } = require('arsenal');
+const { auth, errors, errorInstances, s3middleware, s3routes, models, storage, versioning } = require('arsenal');
+const { decode, encode } = versioning.VersionID;
+const {
+    VersionIdCollisionException,
+    StaleMicroVersionIdException,
+    MicroVersionIdAlreadyStoredException,
+} = require('@scality/cloudserverclient');
 
 const { responseJSONBody } = s3routes.routesUtils;
 const { getSubPartIds } = s3middleware.azureHelper.mpuUtils;
@@ -21,6 +28,7 @@ const locationStorageCheck = require('../api/apiUtils/object/locationStorageChec
 const { dataStore } = require('../api/apiUtils/object/storeObject');
 const prepareRequestContexts = require('../api/apiUtils/authorization/prepareRequestContexts');
 const { decodeVersionId } = require('../api/apiUtils/object/versioning');
+const getReplicationInfo = require('../api/apiUtils/object/getReplicationInfo');
 const locationKeysHaveChanged = require('../api/apiUtils/object/locationKeysHaveChanged');
 const { standardMetadataValidateBucketAndObj, metadataGetObject } = require('../metadata/metadataUtils');
 const { config } = require('../Config');
@@ -32,6 +40,7 @@ const {
 } = require('../api/apiUtils/integrity/validateChecksums');
 const { BackendInfo } = models;
 const { pushReplicationMetric } = require('./utilities/pushReplicationMetric');
+const writeContinue = require('../utilities/writeContinue');
 const kms = require('../kms/wrapper');
 const { listLifecycleCurrents } = require('../api/backbeat/listLifecycleCurrents');
 const { listLifecycleNonCurrents } = require('../api/backbeat/listLifecycleNonCurrents');
@@ -93,7 +102,7 @@ function _isObjectRequest(req) {
     return ['data', 'metadata', 'multiplebackenddata', 'multiplebackendmetadata'].includes(req.resourceType);
 }
 
-function _respondWithHeaders(response, payload, extraHeaders, log, callback) {
+function _respondWithHeaders(response, payload, extraHeaders, log, callback, statusCode = 200) {
     let body = '';
     if (typeof payload === 'string') {
         body = payload;
@@ -115,10 +124,10 @@ function _respondWithHeaders(response, payload, extraHeaders, log, callback) {
         // eslint-disable-next-line no-param-reassign
         response.serverAccessLog.endTurnAroundTime = process.hrtime.bigint();
     }
-    response.writeHead(200, httpHeaders);
+    response.writeHead(statusCode, httpHeaders);
     response.end(body, 'utf8', () => {
         log.end().info('responded with payload', {
-            httpCode: 200,
+            httpCode: statusCode,
             contentLength: Buffer.byteLength(body),
         });
         callback();
@@ -127,6 +136,15 @@ function _respondWithHeaders(response, payload, extraHeaders, log, callback) {
 
 function _respond(response, payload, log, callback) {
     _respondWithHeaders(response, payload, {}, log, callback);
+}
+
+function _respondWithHeaderCrrConflict(response, log, callback, code, message, mvId) {
+    return _respondWithHeaders(
+        response,
+        { code, message },
+        { 'x-scal-micro-version-id': mvId ? encode(mvId) : '' },
+        log, callback, HTTP_STATUS_CONFLICT,
+    );
 }
 
 function _getRequestPayload(req, cb) {
@@ -414,6 +432,40 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         log.error(errMessage);
         return callback(errorInstances.BadRequest.customizeDescription(errMessage));
     }
+
+    const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
+    if (incomingVersionIdEncoded !== undefined) {
+        const incomingVersionIdDecoded = incomingVersionIdEncoded !== 'null'
+            ? decode(incomingVersionIdEncoded)
+            : 'null';
+        if (incomingVersionIdDecoded instanceof Error) {
+            log.error('crr putData: failed to decode x-scal-version-id header', {
+                method: 'putData',
+                error: incomingVersionIdDecoded.message,
+            });
+            return callback(errorInstances.BadRequest.customizeDescription(
+                'bad request: invalid x-scal-version-id header'));
+        }
+        if (objMd && objMd.versionId === incomingVersionIdDecoded) {
+            // Data already at destination for this version; return 409 with the existing
+            // microVersionId so backbeat can decide if putMetadata is still needed.
+            log.debug('crr putData: version already at destination', {
+                method: 'putData',
+                bucketName: request.bucketName,
+                objectKey: request.objectKey,
+                hasMicroVersionId: !!objMd.microVersionId,
+            });
+            request.resume();
+            return _respondWithHeaderCrrConflict(
+                response, log, callback,
+                VersionIdCollisionException.name,
+                'version id already at destination',
+                objMd.microVersionId,
+            );
+        }
+    }
+
+    writeContinue(request, response);
     const context = {
         bucketName: request.bucketName,
         owner: canonicalID,
@@ -539,6 +591,66 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
 }
 
 function putMetadata(request, response, bucketInfo, objMd, log, callback) {
+    const { bucketName, objectKey } = request;
+
+    const encodedMicroVersionId = request.headers['x-scal-micro-version-id'];
+    // When the header is present this is a conditional write: accept only if
+    // the incoming microVersionId is newer than the one already stored.
+    // '' is a valid value meaning the source has no microVersionId (null revision).
+    // Note: '' is falsy so the presence check must be !== undefined, not truthy.
+    const hasMicroVersionId = encodedMicroVersionId !== undefined;
+    let incomingMicroVersionId = null;
+    if (hasMicroVersionId && objMd) {
+        // '' means source has no microVersionId, treated as older revision
+        incomingMicroVersionId = encodedMicroVersionId === '' ? null : decode(encodedMicroVersionId);
+        if (incomingMicroVersionId instanceof Error) {
+            log.error('putMetadata: failed to decode x-scal-micro-version-id header', {
+                method: 'putMetadata',
+                error: incomingMicroVersionId.message,
+            });
+            return callback(errorInstances.BadRequest.customizeDescription(
+                'bad request: invalid x-scal-micro-version-id header'));
+        }
+
+        const isIncomingOlderThanCurrent = (incoming, current) => {
+            if (current === null) { // nothing to be stale against
+                return false; 
+            }
+            // null incoming : source has no cascade revision => oldest possible state
+            // larger string = older revision (reverse-chronological order)
+            return incoming === null || incoming > current;
+        };
+
+        const objectMicroVersionId = objMd.microVersionId || null;
+        let conflictErr = null;
+        if (incomingMicroVersionId === objectMicroVersionId) {
+            conflictErr = {
+                err: MicroVersionIdAlreadyStoredException.name,
+                message: 'incoming microVersionId already at destination',
+            };
+        } else if (isIncomingOlderThanCurrent(incomingMicroVersionId, objectMicroVersionId)) {
+            conflictErr = {
+                err: StaleMicroVersionIdException.name,
+                message: 'incoming revision is older than destination',
+                mvId: objMd?.microVersionId,
+            };
+        }
+        if (conflictErr) {
+            log.debug(`putMetadata: ${conflictErr.message}`, {
+                method: 'putMetadata',
+                bucketName,
+                objectKey,
+            });
+            request.resume();
+            return _respondWithHeaderCrrConflict(
+                response, log, callback,
+                conflictErr.err,
+                conflictErr.message,
+                conflictErr.mvId,
+            );
+        }
+    }
+
     return _getRequestPayload(request, (err, payload) => {
         if (err) {
             return callback(err);
@@ -552,7 +664,16 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             return callback(errors.MalformedPOSTRequest);
         }
 
-        const { headers, bucketName, objectKey } = request;
+        if (omVal.replicationInfo?.status === 'REPLICA') {
+            omVal.replicationInfo.isReplica = true;
+        }
+
+        if (incomingMicroVersionId !== null && incomingMicroVersionId !== (omVal.microVersionId ?? null)) {
+            return callback(errors.BadRequest.customizeDescription(
+                'bad request: x-scal-micro-version-id header does not match body microVersionId'));
+        }
+
+        const { headers } = request;
 
         // Destination-side delete-marker replication.
         // We need the REPLICA status to distinguish from
@@ -560,7 +681,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         if (
             omVal.isDeleteMarker &&
             omVal.replicationInfo &&
-            omVal.replicationInfo.status === 'REPLICA' &&
+            omVal.replicationInfo.isReplica &&
             request.serverAccessLog
         ) {
             // eslint-disable-next-line no-param-reassign
@@ -576,7 +697,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // The REPLICA status excludes source-side replication-status updates.
         if (
             omVal.replicationInfo &&
-            omVal.replicationInfo.status === 'REPLICA' &&
+            omVal.replicationInfo.isReplica &&
             (omVal.originOp === 's3:ObjectTagging:Put' || omVal.originOp === 's3:ObjectTagging:Delete') &&
             request.serverAccessLog
         ) {
@@ -593,7 +714,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // The REPLICA status excludes source-side replication-status updates.
         if (
             omVal.replicationInfo &&
-            omVal.replicationInfo.status === 'REPLICA' &&
+            omVal.replicationInfo.isReplica &&
             omVal.originOp === 's3:ObjectAcl:Put' &&
             request.serverAccessLog
         ) {
@@ -672,9 +793,8 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // then we want to create a version for the replica object even though
         // none was provided in the object metadata value.
         if (omVal.replicationInfo.isNFS) {
-            const { isReplica } = omVal.replicationInfo;
-            versioning = isReplica;
-            omVal.replicationInfo.isNFS = !isReplica;
+            versioning = omVal.replicationInfo.isReplica;
+            omVal.replicationInfo.isNFS = !omVal.replicationInfo.isReplica;
         }
 
         const options = {
@@ -722,6 +842,37 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // If the new null keys logic (S3C-7352) is not supported (compatibility mode), 'isNull' remains undefined.
         if (!nullVersionCompatMode) {
             options.isNull = isNull;
+        }
+
+        const isReplicationWrite = !!headers['x-scal-replication-content'];
+        if (isReplicationWrite) {
+            const isMDOnly = headers['x-scal-replication-content'] === 'METADATA';
+            const objSize = omVal['content-length'] || 0;
+
+            const nextReplInfo = getReplicationInfo(
+                config, objectKey, bucketInfo, isMDOnly, objSize,
+                null, null, null, constants.crrCascadeBlockedLocationTypes);
+
+            const hasNextHop = nextReplInfo && nextReplInfo.backends.length > 0;
+
+            // Replicating further requires x-scal-micro-version-id for loop detection.
+            if (hasNextHop && !hasMicroVersionId) {
+                return callback(errors.InternalError.customizeDescription(
+                    'x-scal-micro-version-id is required when replication would trigger a next hop'));
+            }
+
+            omVal.replicationInfo = hasNextHop ? nextReplInfo : {
+                status: '',
+                backends: [],
+                content: [],
+                destination: undefined,
+                storageClass: undefined,
+                role: undefined,
+                storageType: undefined,
+                dataStoreVersionId: undefined,
+                isNFS: undefined,
+            };
+            omVal.replicationInfo.isReplica = true;
         }
 
         return async.series(
--- a/package.json
+++ b/package.json
@@ -64,11 +64,11 @@
         "vaultclient": "scality/vaultclient#8.5.3",
         "werelogs": "scality/werelogs#semver:^8.2.4",
         "ws": "^8.18.0",
+        "@scality/cloudserverclient": "1.0.9",
         "xml2js": "^0.6.2"
     },
     "devDependencies": {
         "@eslint/compat": "^1.2.2",
-        "@scality/cloudserverclient": "1.0.7",
         "@scality/eslint-config-scality": "scality/Guidelines#8.3.1",
         "eslint": "^9.14.0",
         "eslint-plugin-import": "^2.31.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenko/cloudserver",
-    "version": "9.4.0-preview.5",
+    "version": "9.4.0-preview.6",
     "description": "Zenko CloudServer, an open-source Node.js implementation of a server handling the Amazon S3 protocol",
     "main": "index.js",
     "engines": {
--- a/tests/functional/backbeat/putMetadata.js
+++ b/tests/functional/backbeat/putMetadata.js
@@ -42,38 +42,43 @@ function makeMicroVersionId() {
 }
 
 function buildMetadataBody(overrides) {
-    const obj = Object.assign({
-        'content-length': Buffer.byteLength(OBJECT_BODY),
-        'content-type': 'text/plain',
-        'last-modified': new Date().toISOString(),
-        'x-amz-version-id': 'null',
-        'owner-id': CANONICAL_ID,
-        'owner-display-name': 'test',
-        'content-md5': OBJECT_MD5_HEX,
-        replicationInfo: {
-            status: 'REPLICA',
-            isReplica: true,
-            backends: [],
-            content: [],
-            destination: '',
-            storageClass: '',
-            role: '',
-            storageType: '',
-            dataStoreVersionId: '',
+    const obj = Object.assign(
+        {
+            'content-length': Buffer.byteLength(OBJECT_BODY),
+            'content-type': 'text/plain',
+            'last-modified': new Date().toISOString(),
+            'x-amz-version-id': 'null',
+            'owner-id': CANONICAL_ID,
+            'owner-display-name': 'test',
+            'content-md5': OBJECT_MD5_HEX,
+            replicationInfo: {
+                status: 'REPLICA',
+                isReplica: true,
+                backends: [],
+                content: [],
+                destination: '',
+                storageClass: '',
+                role: '',
+                storageType: '',
+                dataStoreVersionId: '',
+            },
         },
-    }, overrides || {});
+        overrides || {},
+    );
     return Buffer.from(JSON.stringify(obj));
 }
 
 // putMetadata without x-scal-replication-content — tests pure conditional update semantics
 async function putMetadata(key, mvId) {
     const bodyOverrides = mvId ? { microVersionId: mvId.raw } : {};
-    return backbeatClient.send(new PutMetadataCommand({
-        Bucket: TEST_BUCKET,
-        Key: key,
-        MicroVersionId: mvId ? mvId.encoded : undefined,
-        Body: buildMetadataBody(bodyOverrides),
-    }));
+    return backbeatClient.send(
+        new PutMetadataCommand({
+            Bucket: TEST_BUCKET,
+            Key: key,
+            MicroVersionId: mvId ? mvId.encoded : undefined,
+            Body: buildMetadataBody(bodyOverrides),
+        }),
+    );
 }
 
 before(async () => {
@@ -89,37 +94,46 @@ before(async () => {
     });
 
     await s3.send(new CreateBucketCommand({ Bucket: TEST_BUCKET }));
-    await s3.send(new PutBucketVersioningCommand({
-        Bucket: TEST_BUCKET,
-        VersioningConfiguration: { Status: 'Enabled' },
-    }));
+    await s3.send(
+        new PutBucketVersioningCommand({
+            Bucket: TEST_BUCKET,
+            VersioningConfiguration: { Status: 'Enabled' },
+        }),
+    );
 
     await s3.send(new CreateBucketCommand({ Bucket: DEST_BUCKET }));
-    await s3.send(new PutBucketVersioningCommand({
-        Bucket: DEST_BUCKET,
-        VersioningConfiguration: { Status: 'Enabled' },
-    }));
+    await s3.send(
+        new PutBucketVersioningCommand({
+            Bucket: DEST_BUCKET,
+            VersioningConfiguration: { Status: 'Enabled' },
+        }),
+    );
 
     await s3.send(new CreateBucketCommand({ Bucket: TEST_BUCKET_CRR }));
-    await s3.send(new PutBucketVersioningCommand({
-        Bucket: TEST_BUCKET_CRR,
-        VersioningConfiguration: { Status: 'Enabled' },
-    }));
-    await s3.send(new PutBucketReplicationCommand({
-        Bucket: TEST_BUCKET_CRR,
-        ReplicationConfiguration: {
-            Role: 'arn:aws:iam::account-id:role/src-resource,' +
-                'arn:aws:iam::account-id:role/dest-resource',
-            Rules: [{
-                Status: 'Enabled',
-                Prefix: '',
-                Destination: {
-                    Bucket: `arn:aws:s3:::${DEST_BUCKET}`,
-                    StorageClass: 'zenko',
-                },
-            }],
-        },
-    }));
+    await s3.send(
+        new PutBucketVersioningCommand({
+            Bucket: TEST_BUCKET_CRR,
+            VersioningConfiguration: { Status: 'Enabled' },
+        }),
+    );
+    await s3.send(
+        new PutBucketReplicationCommand({
+            Bucket: TEST_BUCKET_CRR,
+            ReplicationConfiguration: {
+                Role: 'arn:aws:iam::account-id:role/src-resource,' + 'arn:aws:iam::account-id:role/dest-resource',
+                Rules: [
+                    {
+                        Status: 'Enabled',
+                        Prefix: '',
+                        Destination: {
+                            Bucket: `arn:aws:s3:::${DEST_BUCKET}`,
+                            StorageClass: 'zenko',
+                        },
+                    },
+                ],
+            },
+        }),
+    );
 });
 
 // These tests send no x-scal-replication-content header — microVersionId is used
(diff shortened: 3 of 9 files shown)


lib/api/apiUtils/object/versioning.js
```diff
@@ -566,6 +565,24 @@ function overwritingVersioning(objMD, metadataStoreParams) {
...
 }
 
+function generateMicroVersionId() {
+    return versionIdUtils.generateVersionId(config.instanceId, config.replicationGroupId);
+}
+
+/**
+ * @param {Object} objectMD - plain object metadata (not an ObjectMD instance)
+*/
+function prepareMetadataForCascadedCrr(objectMD) {
```
SylvainSenechal: Could be in arsenal 🤔

lib/routes/routeBackbeat.js
```diff
@@ -430,6 +433,30 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
+            return _respondWithHeaders(response, [],
+                { 'x-scal-cascade-loop-detected': 'true' }, log, callback);
+        }
+        if (event === 'stale') {
+            log.debug('crr cascade putData: stale event, rejecting', {
+                method: 'putData',
+                bucketName: request.bucketName,
+                objectKey: request.objectKey,
+            });
+            return callback(errors.OperationAborted);
```
SylvainSenechal: many 409 available in arsenal. we can pick this one, maybe some other ones, or create our own

lib/routes/routeBackbeat.js
```diff
@@ -669,7 +721,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
         // then we want to create a version for the replica object even though
         // none was provided in the object metadata value.
         if (omVal.replicationInfo.isNFS) {
-            const isReplica = omVal.replicationInfo.status === 'REPLICA';
+            // Accept both the legacy status='REPLICA' and the new isReplica flag
+            // so that cascade hops (which arrive with status='PENDING') are also
+            // recognised as replicas.
+            const isReplica = omVal.replicationInfo.isReplica === true
```
SylvainSenechal: Should probably use the new helper function from arsenal : getReplicationIsReplica ? 

edit : cannot, working on json, not object

lib/routes/routeBackbeat.js
```diff
@@ -721,6 +777,53 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+                nextReplInfo.backends = nextReplInfo.backends.filter(b => {
+                    const loc = config.locationConstraints[b.site];
+                    return !loc || !BLOCKED_LOCATION_TYPES.includes(loc.type);
+                });
+            }
+
+            if (nextReplInfo && nextReplInfo.backends.length > 0) {
+                omVal.replicationInfo = nextReplInfo;
+            } else {
+                omVal.replicationInfo = {
```
SylvainSenechal: maybe this should be an arsenal function as its updating metadata

edit: no, cannot as we are working with json

lib/routes/routeBackbeat.js
SylvainSenechal: Yeah i think we need to use the helper function from Arsenal to check if an object is a replica, using both odl and new field for compatibility. 
A few places in the code where we are still hard checking ''REPLICA' that are gonna break 🤔

tests/functional/backbeat/putMetadata.js
SylvainSenechal, line 1: all these worked locally, need to wait for the bumps now

package.json
```diff
@@ -33,7 +33,7 @@
         "@azure/storage-blob": "^12.28.0",
         "@hapi/joi": "^17.1.1",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/Arsenal#8.4.2",
+        "arsenal": "file:../Arsenal",
```
SylvainSenechal: TODO 🧐

lib/api/apiUtils/object/versioning.js
SylvainSenechal: no i think its fine here (talking to myself)

tests/functional/backbeat/putMetadata.js
```diff
@@ -0,0 +1,268 @@
...
+                Status: 'Enabled',
+                Prefix: '',
+                Destination: {
+                    Bucket: `arn:aws:s3:::${DEST_BUCKET}`,
+                    StorageClass: 'zenko',
+                },
+            }],
+        },
+    }));
+});
```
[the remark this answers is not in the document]
SylvainSenechal: bucket/keys names are randomized
I intentionnally didnt add any after to keep the test and file lighter.

Do we want to add after cleanup ?

scality-cloudserverclient-v1.0.9.tgz
maeldonn, line 1: Remove this file

Dockerfile
```diff
@@ -25,7 +25,7 @@ ENV PYTHON=python3
 RUN npm install -g \
     node-gyp \
     typescript@4.9.5
-COPY package.json yarn.lock /usr/src/app/
+COPY package.json yarn.lock scality-cloudserverclient-v1.0.9.tgz /usr/src/app/
```
maeldonn: Don't forget to revert, I guess it's for testing

lib/routes/routeBackbeat.js
```diff
@@ -541,6 +564,28 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+                { 'x-scal-micro-version-id-exists': 'true' }, log, callback);
+        }
+        if (event === 'stale') {
+            log.debug('crr cascade putMetadata: stale event, rejecting', {
+                method: 'putMetadata', bucketName, objectKey,
+            });
+            return callback(errors.OperationAborted);
+        }
+    }
+
```
maeldonn, line 663: You need to use the helper `writeContinue()` to allow the client to streamData.

maeldonn (changes requested):
I think we can remove the changes from this PR that are already covered in the other one. Once that's done, we can rebase this branch onto the other PR (since Prettier is already included there).

lib/routes/routeBackbeat.js
SylvainSenechal: ok let me see I thought this was somewhat automatic

scality-cloudserverclient-v1.0.9.tgz
SylvainSenechal: yeah for testing i shipped cloudserverclient with it

lib/routes/routeBackbeat.js
```diff
@@ -541,6 +566,31 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+            log.debug('crr cascade putMetadata: stale event, rejecting', {
+                method: 'putMetadata', bucketName, objectKey,
+            });
+            return _respondWithHeaders(response,
+                { code: StaleMicroVersionIdException.name, message: 'incoming revision is older than destination' },
+                {}, log, callback, 409);
+        }
+    }
+
+    writeContinue(request, response);
```
maeldonn:
```suggestion
```

lib/routes/routeBackbeat.js
```diff
@@ -430,6 +434,27 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
+        log.debug('crr cascade putData: version already at destination', {
+            method: 'putData',
+            bucketName: request.bucketName,
+            objectKey: request.objectKey,
+            hasMicroVersionId: !!objMd.microVersionId,
+        });
+        return _respondWithHeaders(response,
+            { code: VersionIdCollisionException.name, message: 'version id already at destination' },
+            { 'x-scal-micro-version-id': objMd.microVersionId ? encode(objMd.microVersionId) : '' },
+            log, callback, 409);
```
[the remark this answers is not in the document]
SylvainSenechal: this will be defensive but yeah calling request.resume()

scripts/crr-cascade-test.ts
```diff
@@ -0,0 +1,281 @@
...
+// ─── location names as configured in the Artesca UI ──────────────────────────
+// https://s3.workloadplane.scality.local
+// 7MB2SGXF3LMZMCFZOD22 / zDKH+EKDlOKG6Ivmdewg9veQJ6FvqUzd+j5yKbT=
+const LOCATION_B = 'cascade-b';
+// 2H6QL1SOWZBF0KGEBS1B / sODDxuUoeW+R3w4pKoZ5CyZGFGXhVv/cBfL98jUD
+const LOCATION_C = 'cascade-c';
+// ─────────────────────────────────────────────────────────────────────────────
+
+const requestHandler = new NodeHttpHandler({
+    httpsAgent: new https.Agent({ rejectUnauthorized: false }),
```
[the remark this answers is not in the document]
SylvainSenechal: file wrongly commited

scripts/crr-cascade-test.ts
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+        if (args.request?.body && typeof args.request.body !== 'string') {
+            // Force body to string to avoid chunked encoding
+            const chunks: Buffer[] = [];
+            for await (const chunk of args.request.body) chunks.push(chunk);
+            args.request.body = Buffer.concat(chunks).toString();
+        }
+        return next(args);
+    },
```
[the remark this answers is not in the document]
SylvainSenechal: file wrongly commited

lib/routes/routeBackbeat.js
```diff
@@ -414,6 +422,32 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
+            hasMicroVersionId: !!objMd.microVersionId,
+        });
+        request.resume();
+        return _respondWithHeaders(
+            response,
+            { code: VersionIdCollisionException.name, message: 'version id already at destination' },
+            { 'x-scal-micro-version-id': objMd.microVersionId ? encode(objMd.microVersionId) : '' },
+            log,
+            callback,
+            409,
```
maeldonn: Use node http constants everywhere instead of hardcoding

lib/routes/routeBackbeat.js
```diff
@@ -724,6 +801,48 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+        // Always mark isReplica=true.
+        if (incomingMicroVersionId) {
+            const isMDOnly = headers['x-scal-replication-content'] === 'METADATA';
+            const objSize = omVal['content-length'] || 0;
+
+            // These S3-compatible Scality locations are excluded
+            // as cascade targets because they use the MultiBackend S3 path which
+            // bypasses the putData/putMetadata routes, so loop detection cannot fire
+            // on those destinations.
+            const BLOCKED_LOCATION_TYPES = ['location-scality-ring-s3-v1', 'location-scality-artesca-s3-v1'];
```
maeldonn: Could it be tested on a real lab ?

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -157,6 +157,7 @@ function createAndStoreObject(bucketName, bucketMD, objectKey, objMD, authInfo,
         replicationInfo: getReplicationInfo(config,
             objectKey, bucketMD, false, size, null, null, authInfo),
         overheadField,
+        updateMicroVersionId: true,
```
maeldonn: Do we really need this ?

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +573,48 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+            request.resume();
+            return _respondWithHeaders(
+                response,
+                { code: MicroVersionIdAlreadyStoredException.name,
+                    message: 'incoming microVersionId already at destination' },
+                {},
+                log,
+                callback,
+                409,
+            );
```
maeldonn: could you remove duplication to simplify here ?

lib/routes/routeBackbeat.js
```diff
@@ -414,6 +422,32 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         log.error(errMessage);
         return callback(errorInstances.BadRequest.customizeDescription(errMessage));
     }
+
+    const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
+    const decoded = incomingVersionIdEncoded ? decode(incomingVersionIdEncoded) : null;
+    const incomingVersionIdDecoded = decoded instanceof Error ? null : decoded;
+    if (incomingVersionIdDecoded && objMd && objMd.versionId === incomingVersionIdDecoded) {
```
francoisferrand: > `objMd.versionId === incomingVersionIdDecoded` 

- Can this condition actually fail?
- Is it really part of the condition?

AFAIU, `versionId` is used (or should be?) to find the existing objMD:
- so if not found, we would get objMD=null ;
- and if found, it should have the versionID...
- Unless there is a some logic here which gets the matching version if it exist, else the latest one : in that case objMd.versionId !== incomingVersionIdDecoded means we actually did not an object already

→ should clarify how we get there, and nice if we can drop this condition
(remember that with CRR with cannot work in non-versioned buckets -so no "overwrites"-, and that versions may be replicated out of order)

lib/api/apiUtils/object/createAndStoreObject.js
SylvainSenechal: yes, actually this is important, I added this after making my tests in codespaces : 
Without this :
You put an object : object isn't assigned an micro version Id
You setup the whole replication, backbeat picks up that source object, it doesn't have any microversion id so cloudserver just doesn't try to process it as a potential cascade replication

lib/routes/routeBackbeat.js
SylvainSenechal: Ok I did something, you take a look but honestly it's really not that great 🤔

lib/routes/routeBackbeat.js
SylvainSenechal: should be honestly, i think i need to double check this together with the design and the code

maeldonn (approved):
LGTM. Please bump package.json and create a new release. Could you just confirm that the location guard is working ?

lib/api/apiUtils/object/createAndStoreObject.js
francoisferrand: - should "new" objects not keep having an empty microVersionId (to avoid wasting metadata space)?
and just consider "no micro version id" is the oldest micro versionId (the only thing it cannot compare to is another no micro version id - but that would not happen on new object, each get their own versionId)
- in case of un-versioned buckets (i.e. only master objects), we may need to update the micro version id (because we have no version id) ; however there is no replication in that case, so no problem not to do it

lib/routes/routeBackbeat.js
```diff
@@ -414,6 +432,30 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         log.error(errMessage);
         return callback(errorInstances.BadRequest.customizeDescription(errMessage));
     }
+
+    const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
+    const decoded = incomingVersionIdEncoded ? decode(incomingVersionIdEncoded) : null;
+    const incomingVersionIdDecoded = decoded instanceof Error ? null : decoded;
```
francoisferrand: should we not actually _fail_ if the version id cannot be decoded? (this would be a bad request)

lib/routes/routeBackbeat.js
```diff
@@ -414,6 +432,30 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         log.error(errMessage);
         return callback(errorInstances.BadRequest.customizeDescription(errMessage));
     }
+
+    const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
+    const decoded = incomingVersionIdEncoded ? decode(incomingVersionIdEncoded) : null;
```
francoisferrand: nit: should we not use an empty string instead of null, so it can be compared more easily?
```suggestion
    const decoded = incomingVersionIdEncoded ? decode(incomingVersionIdEncoded) : '';
```

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +581,42 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
 }
 
 function putMetadata(request, response, bucketInfo, objMd, log, callback) {
+    const { bucketName, objectKey } = request;
+
+    const encodedMicroVersionId = request.headers['x-scal-micro-version-id'];
```
francoisferrand, line 602: we probably need to differentiate between
- the absence of microVersionId (ie. x-scal-micro-version-id is not present in the request) → no extra check
- microVersionId is empty (something like `x-scal-micro-version-id: ""`, when referring to the first metadata revision when object was created) → need the extra check, and check if the object already has newer micro version id (i.e. either it does not have one → "equal", or it does → "newer")

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +581,42 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+    const decoded = encodedMicroVersionId ? decode(encodedMicroVersionId) : null;
+    const incomingMicroVersionId = decoded instanceof Error ? null : decoded;
```
francoisferrand: if micro version id header is set, it MUST either be empty or decode cleanly; otherwise request is invalid

lib/routes/routeBackbeat.js
francoisferrand: the code is completely duplicated, the only difference is really name of exception and "text" of message.

as discussed in cloudserverclient, do we really need cloudserver to make that distinction? just returning the same error (micro version id conflict) AND the micro version id of the object stored allows the caller to make the distinction, if they need - without duplication.

francoisferrand: ahhh, I missed the extra type MicroVersionIdAlreadyStoredException, c.f. https://github.com/scality/cloudserverclient/pull/24/changes#r3496549728

we had the discussion and ended adding microVersionId in StaleMicroVersionIdException, but it is useless if we generate 2 kinds of errors...
→ IMHO we should drop `MicroVersionIdAlreadyStoredException`, which allows deduplicating here and simplifying backbeat?

---
otherwise we shoud still dedup, something like:

```suggestion
if (cmp !=== Ordering.NEWER) {
   request.resume();
            return _respondWithHeaderCrrConflict(
                response, log, callback,
                cmp === Ordering.EQUAL ? MicroVersionIdAlreadyStoredException.name : StaleMicroVersionIdException.name,
                'incoming microVersionId already at destination',
            );
}
```

lib/routes/routeBackbeat.js
francoisferrand: the function for updating replicationInfo is really `getReplicationInfo()`.

but in all other places (c.f. createAndStoreObject, completeMultiPart...) with do a much simpler `replicationInfo = getReplicationInfo()` : shouldn't we do the same? Why do we need to do it differently here?

lib/routes/routeBackbeat.js
```diff
@@ -724,6 +803,48 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+            // on those destinations.
+            const BLOCKED_LOCATION_TYPES = ['location-scality-ring-s3-v1', 'location-scality-artesca-s3-v1'];
+
+            const nextReplInfo = getReplicationInfo(config, objectKey, bucketInfo, isMDOnly, objSize, null, null, null);
+
+            if (nextReplInfo) {
+                nextReplInfo.backends = nextReplInfo.backends.filter(b => {
+                    const loc = config.locationConstraints[b.site];
+                    return !loc || !BLOCKED_LOCATION_TYPES.includes(loc.type);
+                });
```
francoisferrand: this logic should be in `getReplicationInfo`, this is the function responsible to build the list of backend/destinations. Doing it there will also avoid create a backend entry only to remove it.

lib/routes/routeBackbeat.js
```diff
@@ -724,6 +803,48 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+        // next hop. If not, clear the source-side replicationInfo fields
+        // Always mark isReplica=true.
+        if (incomingMicroVersionId) {
+            const isMDOnly = headers['x-scal-replication-content'] === 'METADATA';
+            const objSize = omVal['content-length'] || 0;
+
+            // These S3-compatible Scality locations are excluded
+            // as cascade targets because they use the MultiBackend S3 path which
+            // bypasses the putData/putMetadata routes, so loop detection cannot fire
+            // on those destinations.
```
francoisferrand: is this part of the design? I may be wrong, but I thought we wanted (needed maybe, to be check with product team) to trigger multi-backend replication after CRR ?
→ please double check

lib/routes/routeBackbeat.js
SylvainSenechal: rewrote this part a bit, but this is all just refactoring

lib/routes/routeBackbeat.js
SylvainSenechal: yes actually i changed the single if into 2 ifs :
If the versionId is provided, but we can't decode it, it means something is wrong and we shouldn't continue anything

lib/routes/routeBackbeat.js
SylvainSenechal: I rewrote some part of the if statement into 2 stages, but your comment still stands.

I am not sure i understand it though, I think you are saying that if we were able to get objMd, then necessarily objMd will be equal to incomingVersionIdDecoded ? 

- objMd is obtained through a middleware, from the provided bucket + the key, and it's the latest available version of that bucket
- the x-scal-micro-version-id is anything people using the client decide to put inside. I don't see a situation where we would skip checking objMd.version === incomingVersionId

lib/api/apiUtils/object/createAndStoreObject.js
maeldonn: - Similar thread here: https://github.com/scality/citadel/pull/349#discussion_r3137680994
- ~~Should be done in objectPut instead: [`7a63acf` (#6178)](https://github.com/scality/cloudserver/pull/6178/changes/7a63acf389198acd80ca4d39bd7ce6de5c30c688#diff-a0fc5123110176e6643145a82c4d9858b9114a68e0f32a4a75dcd4f30621b2bbR97)~~

SylvainSenechal: We discussed this a bit with Maël. 
I think we could try to not do it all the time, and do it only for putObject (but not even sure).
But we were wondering what happens if we don't write microVersionId on putObject : how do we compare two objects that don't have any micro version ID, it may be possible on the first replication to consider no micro version id => source is newer, but then we must write the microVersionID anyways to block loop and make later write work.
Also in the design we said microVersionId would necessarily be used for this feature only but more generally to be able to track and compare update time, I'm very worried we're gonna have issues later if we start writing microVersionID only on certain conditiion. Actually the code in arsenal, backbeat and cloudserver is already filled with quite a lot of defensive and smelly if (microVersionID) because we never bothered adding this from the start

lib/routes/routeBackbeat.js
SylvainSenechal: I change the code a bit to do what you mention in the first point, and also return an error when the header is present but fail to be decoded like in putData.

For the other point,
The current implementation is : new source will always have a microVersionId, it's talked about in another comment at the top of this review, but I don't see how we can properly deal with empty microversion ids on new objects written after this feature is added, without having issues later. It means we would have situation like 

A->B->C
I put on A, A has no microVersionID, but replicates to B and C which now have a microVersionID.
But A still doesn't have a microVersionID. What happens if there is a loop configuration and C goes back to A
What happens on the second write on A when a first write has already replicated on B and C ? B and C will have a microVersionID but A won't ?

It seems to me that deciding to not assign a microVersionID on all new putObject writes is gonna get real tricky in the long run

lib/routes/routeBackbeat.js
SylvainSenechal: Sending back the microVersionID is indeed not super useful, I added it because it was suggested it could become useful or nice to have the info for the caller. I'm not really gonna focus on this to much, i can remove it or keep it, imo I think it's fine that an api returns it.

But for the double error, well tbh i don't really mind either as I just wanna put an end to this feature, but for me it's better to have backbeat check the errors like this : 

Image

Than having it just receive a microVersion id, then does another loop if microversion == null, if microversion < > xxx, then we don't do exactly the same thing (cbOnce(cascadeLoopDetected) vs cbOnce(cascadeDataStale)).

In the screenshot, we directly have errors that we can nicely catch and that's pretty much it.

I just see calling the client like calling any functions, and if the function I call return an error, I want the error to be as detailed as possible and not redo the checks that was already done, because anyways cloudserver has to do those comparisons here.

SylvainSenechal: I will recheck the suggested simpification after we settle the discussion, not fully sure it should be done though because compare can return 4 values
older
younger
equal
not comparable

so doing this cmp !=== Ordering.NEWER
still leaves room for NOT_COMPARABLE, I think it's safer to assert on what we want that exclude only value

lib/routes/routeBackbeat.js
SylvainSenechal: it's here https://github.com/scality/citadel/pull/349/changes#diff-f4fbff9b43d54385a8778f028f6950744393bb5235c5860afa7eec1b89072f83R266

but yeah i need to triple check it because yes i don't really like this part with these hardcoded locations

lib/routes/routeBackbeat.js
SylvainSenechal: Are you sure you wanna do this ? It means I have to add some kind of flag to getReplicationInfo to do the current logic in getReplicationInfo 🤔

SylvainSenechal: Ok I see, maybe it can be done cleanly if, in getReplicationInfo.js I add a helper function, leave the existing function almost as it is and just call that extra helper at the end of getReplicationInfo if the flag is on ?

lib/api/apiUtils/object/createAndStoreObject.js
francoisferrand: > we were wondering what happens if we don't write microVersionId on putObject : how do we compare two objects that don't have any micro version ID

no issue there: 
- there are 2 "independent" putObject on 2 different sites : they will have different versionId, so no need to compare microVersionId
- "within" a specific version (versionID), the empty/missing versionID is inequivoquality the oldest revision of metadata (the one from the original putObject). Any followup updates to that versionID need to bump the micro version id

This case may be a "lack" of microVersionID in the metadata (field not set, for metadata size efficiency), but it is really just a sentinel value for the oldest microVersionID; and it compares exactly the same:
- "empty" versionID on both side → already up to date (equality)
- replicate "empty" versoinID while the target already have a versionID → stale versionID (older)
- replicate versionID while the target has an empty versionID → proceed with replication (newer)

> we start writing microVersionID only on certain conditiion.

this case will be present anyway: all objects before the upgrade will not have versionID, so we must be able to manage this state: we cannot assume it will be present.
(and crrExistingObject may add microVersionID, but I'd rather not be protected by this alone: with DR/... there may be other cases coming, best be prepared)

lib/routes/routeBackbeat.js
francoisferrand: > Sending back the microVersionID is indeed not super useful, I added it because it was suggested it could become useful or nice to have the info for the caller

not sure where it was lost in the discussion, but the suggestion was really to add microVersionID in the result so we could have a single error : i.e. reduce the logic to minimal in cloudserver, and keep the most in backbeat.

(btw, the code you mention also has similar comment about duplication: my _feeling_ -to be confirmed- is that this duplication at every layer is due at least in part to the qualification happening too early, and would be avoided by propagating a single error [wiht microVersionId] accross all layers, and differentiating the cases only at the end of the chain)

lib/routes/routeBackbeat.js
francoisferrand: > objMd is obtained through a middleware, from the provided bucket + the key, and it's the latest available version of that bucket

- if by middleware you mean the "router" in the same routeBackbeat file, indeed; but still this is something under our control and closely related to this code
- I thought this would return the specific `versionId` associated with the result ; if it returns the  _latest_, then the issue is much larger : we need to try and check if we have this specific `versionId`. We may be replicating the "previous" version, but the logic must apply in that case as well...

> the x-scal-micro-version-id is anything people using the client decide to put inside. I don't see a situation where we would skip checking objMd.version === incomingVersionId

- this API cannot be used by "people", it is used only by scality software
- if the `x-scal-version-id` is set to tell which version we are interested in (and should be retrieved). If there is no such revision, there is no conflict (and responsibility/bug of the caller if they did not provide the right versionId) ; if this revision is present, this is a conflict. We don't care if there is another revision of the object, we should only attempt to retrieve the ObjMD of the revision we are interested in.

lib/routes/routeBackbeat.js
francoisferrand: > new source will always have a microVersionId

I really don't understand why we need this.

This was discussed in the design, and is a critical point to mitigate the impact on MongoDB sizing.

- It really just mean "no micro versionID" is also a valid micro version id, and older than any other valid microVersionID value
- There is no additional risk due to concurrent putObject : as replication is possible only in versioned buckets, so object will be versioned and not have the same VersionID → so no conflict is possible (conflict is only ever checks on objects with THE SAME versionId)

> I put on A, A has no microVersionID, but replicates to B and C which now have a microVersionID.

when replicating, we do putMetadata : so B and C should have the same "empty" microVersionID.
the replication itself MUST NOT change the microVersionID (or metadata, for that matter): it just performs an (almost) verbatim copy of the metadata.

francoisferrand (commented):
@SylvainSenechal seems there are quite the a few open points, which don't look like they will converge quickly through review comments.

Can you prepare a small doc to list each of these topics (both on this PR and the backbeat one, likely the same but best be sure), with the different approaches and arguments in favor of each?
→ then we can go through them in a short meeting (30min should be enough?), settle the question and move forward

(note: some of these questions may be refined/addressed incrementally in followup PRs, so we could merge this PR already; and for others it may make sense to make the change first. But either way we need to settle the points ASAP -and implement the followup, if any- before we are ready to ship)

lib/routes/routeBackbeat.js
SylvainSenechal: We rediscussed with Maël, I think this is ok to keep, we don't want cascade replication to continue on these locations

lib/routes/routeBackbeat.js
SylvainSenechal: This was discussed live and it's fixed.
Making the difference between no header at all, empty value '', and real micro version id now

lib/routes/routeBackbeat.js
SylvainSenechal: up, just wanna double check whats the potential issue here

lib/api/apiUtils/object/createAndStoreObject.js
SylvainSenechal: discussed live and code is updated :
microVersionId is only used for metadata diff checks now

header microVersionId absent => no crr cascade
header microVersionId present but empty string '' => replicate without asking question
header microVersionId present with a value => do comparison

lib/routes/routeBackbeat.js
SylvainSenechal: I reviewed everything, but I don't really see any code duplication on putMetadata, backbeat does a putMetada, receive the error which directly goes to "handleReplicationOutcome", without needing to do any comparison again. Looks good to me, the functions called sends an error that's as precise as possible

lib/routes/routeBackbeat.js
```diff
@@ -414,6 +432,38 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         log.error(errMessage);
         return callback(errorInstances.BadRequest.customizeDescription(errMessage));
     }
+
+    const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
+    if (incomingVersionIdEncoded) {
```
francoisferrand: header may be present but empty, for `null versions`

```suggestion
    if (incomingVersionIdEncoded != undefined) {
```

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
 
 function putMetadata(request, response, bucketInfo, objMd, log, callback) {
+    const { bucketName, objectKey } = request;
+
+    const encodedMicroVersionId = request.headers['x-scal-micro-version-id'];
+    const isCascadeWrite = encodedMicroVersionId !== undefined;
+    let incomingMicroVersionId = null;
+    if (isCascadeWrite) {
+        // '' means source has no microVersionId, treated as older revision
+        incomingMicroVersionId = encodedMicroVersionId === '' ? null : decode(encodedMicroVersionId);
```
francoisferrand: cloudserver **DOES NOT** know this is replication: this route is `putMetadata`, not `write replicated metadata`, and is used everywhere in backbeat (lifecycle, gc...)

The semantics here is not that of replication, it is a generic conditional-write : "_when `microVersionId` header is set AND the object/version already exist, update metadata only if microVersoinId is newer than  the one already stored_"

```suggestion
if (encodedMicroVersionId !== undefined) {
    const microVersionId = encodedMicroVersionId === '' ? null : decode(encodedMicroVersionId);

File: lib/routes/routeBackbeat.js [hunk 26]
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+        if (incomingMicroVersionId instanceof Error) {
+            log.error('crr putMetadata: failed to decode x-scal-micro-version-id header', {
+                method: 'putMetadata',
+                error: incomingMicroVersionId.message,
+            });
+            return callback(errorInstances.BadRequest.customizeDescription(
+                'bad request: invalid x-scal-micro-version-id header'));
+        }
+        // Loop/stale detection only applies when the object already exists at destination
```
francoisferrand: same: this is not loop/stale detection, but really `microVersionId` conditional update/conflict detection.

loop/stale is the feature of _replication_ we build on top - in backbeat only, where the replication lives.

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+        if (objMd) {
+            // null = oldest revision (original putObject, before any metadata update).
+            // VersionID strings are in reverse chronological order: a lexicographically
+            // larger string is an older revision.
+            const isSourceOlderThanDestination = (sourceMicroVersionId, destinationMicroVersionId) =>
+                destinationMicroVersionId !== null &&
+                (sourceMicroVersionId === null || sourceMicroVersionId > destinationMicroVersionId);
```
francoisferrand: is this not handled already in arsenal's `compareMicroVersionId` function ?

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+
+            // Treat '' and undefined as null (no microVersionId set).
+            const sourceMicroVersionId = incomingMicroVersionId || null;
```
francoisferrand: * `incomingMicroVersionId` is already `null` if false, c.f. definition above: `incomingMicroVersionId = encodedMicroVersionId === '' ? null : decode(encodedMicroVersionId);` ?
* do we really need a second variable?

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+            const destinationMicroVersionId = objMd.microVersionId || null;
```
francoisferrand: source and destination are a bit replication-specific : just `microVersionId` and `objMicroVersionId` variables should be enough?

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+        if (incomingMicroVersionId instanceof Error) {
+            log.error('crr putMetadata: failed to decode x-scal-micro-version-id header', {
+                method: 'putMetadata',
+                error: incomingMicroVersionId.message,
+            });
+            return callback(errorInstances.BadRequest.customizeDescription(
+                'bad request: invalid x-scal-micro-version-id header'));
+        }
+        // Loop/stale detection only applies when the object already exists at destination
+        if (objMd) {
```
francoisferrand: do we need to validate microVersionId if `obj` (i.e. the specified object/version document) was not found?

We could keep things simpler with `if (encodedMicroVersionId !== undefined && obj) {` at the beginning, what do you think?

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+                    objectKey,
+                });
+                request.resume();
+                return _respondWithHeaderCrrConflict(
+                    response, log, callback,
+                    MicroVersionIdAlreadyStoredException.name,
+                    'incoming microVersionId already at destination',
+                );
+            }
+            if (isSourceOlderThanDestination(sourceMicroVersionId, destinationMicroVersionId)) {
```
francoisferrand: I am afraid this may not be enough : we are basically
1. reading metadata from mongo (obj)
2. comparing the microVersionId → abort if too old
3. writing metadata

This is not atomic, so 2 such calls could race each-other, both pass the check 2 and write the metadata... To make it safe, we need to make a conditional write in mongodb somehow: i.e. let mongo do the `microVersionId` comparison (or implement a generic "racing writing" detection logic, probably out of scope here...)

lib/routes/routeBackbeat.js
```diff
@@ -724,6 +840,48 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             options.isNull = isNull;
         }
 
+        // Cascade triggering
+        // If the bucket receiving this replica has its own CRR rules, set
+        // status to PENDING so the queue populator here picks it up for the
+        // next hop. If not, clear the source-side replicationInfo fields
+        // Always mark isReplica=true.
+        if (isCascadeWrite) {
```
francoisferrand: `microVersionId` header semantics is not "try to replicate" but really "do not write older metadata":
- if another feature (lifecycle) uses putMetadata sets this header → may unexpectedly trigger replication ;
- conversely, another feature may use putMetadata to make a change which requires replication

so this is not the right condition: microVersionId is about making the call conditioned, while this block  is really about "begin called as part of replication"...

→ For the first point, checking `headers['x-scal-replication-content']` is the correct & precise way to identify if this put metadata is part of replication. If it is, then we must (now) try to cascade.
→ Second point is not addressed, and not needed at the moment AFAIK
→ As a _safety check_, we may here _require_ that microVersionId condition is specified when making a replication's putMetadata : and reject the call otherwise (internal error) instead of just (silently) skipping cascaded replication...

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
 }
 
 function putMetadata(request, response, bucketInfo, objMd, log, callback) {
+    const { bucketName, objectKey } = request;
+
+    const encodedMicroVersionId = request.headers['x-scal-micro-version-id'];
```
francoisferrand, line 602: when the header is set, maybe we should have a sanity check (later) to verify the metadata contains the same `microVersionId` as the header

lib/routes/routeBackbeat.js
```diff
@@ -607,17 +716,23 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
 
         if (headers['x-scal-replication-content'] === 'METADATA') {
             if (!objMd) {
-                return callback(errors.ObjNotFound);
+                // For metadata-only writes, the destination's data location must be
+                // preserved from the existing object. If there is no existing object
+                // and the source has data, there is nothing to copy the location from.
+                // Zero-byte objects have no data location, so this is safe to skip.
```
francoisferrand: it is safe to skip, but also safe to keep as well : and the previous code was much simpler. If there is no need change it, best not to change it.

Especially semantically, `headers['x-scal-replication-content'] === 'METADATA'` means we expect to _update_ the metadata of an object: so that object MUST exist. Removing the check could mask bugs (e.g. backbeat set METADATA instead of DATA+METADATA) or lead to re-creating the object/version (empty) after it was deleted.

→ keep the previous chunk, this should not be changed

lib/routes/routeBackbeat.js
```diff
@@ -724,6 +840,48 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+        // Always mark isReplica=true.
+        if (isCascadeWrite) {
+            const isMDOnly = headers['x-scal-replication-content'] === 'METADATA';
+            const objSize = omVal['content-length'] || 0;
+
+            // These S3-compatible Scality locations are excluded
+            // as cascade targets because they use the MultiBackend S3 path which
+            // bypasses the putData/putMetadata routes, so loop detection cannot fire
+            // on those destinations.
+            const BLOCKED_LOCATION_TYPES = ['location-scality-ring-s3-v1', 'location-scality-artesca-s3-v1'];
```
francoisferrand: can we know the "source" (sender) of the request here, and map it to a location?
It would not solve the loop problem, but passing that information would allow skipping extra API call(s) in case of bidirectional CRR, and keep same "load" as the existing code.

→ please create followup, could be a nice optimization?

(kind of flimsy, but we could extract the "replication group" from versionId ; not sure how to map to locations though. Or checking if source IP is in the location's bootstrap list. Other maybe other ways...)

lib/routes/routeBackbeat.js
francoisferrand: I think it would be better to move it out of here indeed : easier to test (isolated), easier to find (logic in a single place), and makes the current function smaller...

Not sure if we could do with the existing params, extend them, or if we need a new one ; but (esp if we add one) we should also take the chance to refactor the API of `getReplicationInfo` to have a take a structure instead of a long series of `null`...

(may be in followup PR though, but should be done either way I think)

tests/functional/backbeat/putMetadata.js
francoisferrand, line 1: `microVersionID` may be specified as well when NOT doing replication, i.e. when `headers['x-scal-replication-content']` is not set.
→ need to test both microVersionID semantics (i.e. conditional updates) AND the interaction with replication (when replicating, should cascade).
→ maybe more readable to split into 2 files `putMetadata.js` and `putData.js` tests matching the actual APIs, and cover both aspects (and possibly combinations)

lib/routes/routeBackbeat.js
francoisferrand: if `headers['x-scal-replication-content']` is set, but `microVersionID` is not:
- we must still mark the object as `isReplica`. This must be done as long as this is a replication request
- what we may skip or reject is _only the cascade trigger_, i.e. updating replication info to PENDING. Not sure what we should put instead, if we skip

This could be caused by 2 things:
- bug in code → should reject/fail with an error
- or may be the case of "older artesca" replicating to "newer artesca" → in that case, skipping cascade may be better indeed to avoid breaking customer workflow, but need to double-think if this could break further... At the very least we would need a 
- maybe a compromise would be to fail only if microVersionId is not set AND we would trigger a cascade (could be risky? or maybe not, since at least _we_ would set it...) and ignore otherwise...

```suggestion
        if (headers['x-scal-replication-content']) {
        
            const replInfo = getReplicationInfo(config, objectKey, bucketInfo, isMDOnly, objSize, null, null, null);
            replInfo.isReplica = true
        
            if (!headers['x-scal-micro-version-id'] && replInfo.status == PENDING) {
                return respondWithError(...)
            }
            
            obj.relicationInfo = replInfo;
```

→ in any case `headers['x-scal-replication-content']` must be part of the condition, this is the real "first level" condition for cascade
→ not sure how best to handle the missing `microVersionId` in that case: e.g. if ignoring is safe (but still we must set "isReplica"), if it should fail (strictly safe, but may break in upgrade scenarios), if it should replicate anyway (trust the user to deploy cascaded/... only when all systems are ugpraded) or if we should fail only when actually triggering replication... what do you think?

francoisferrand (changes requested):
the condition microVersionId update is not atomic → must somehow involve the mongo/metadata layer, i.e. perform an actual conditional metadata update on `microVersionId`

delthas:
_Removing myself as reviewer since alrdy covered by 2 people; and quite complex. Re-add me if needed._

lib/routes/routeBackbeat.js
SylvainSenechal: Mhh, looking at the way this is sent from backbeat, we are very unlikely to send '' unless we hardcode '' in the client.
This is because in backbeat, we are getting the versionID from an Arsenal helper that will returns null when there is no version ID

Also, if we let '' go though this if, it will fail right after on the decode function as its an invalid version id. 

We should still make the api here a bit more self documenting, how about this : 


"We expect either no version ID, or a real version id that isnt an empty string" :
if (incomingVersionIdEncoded != null && incomingVersionIdEncoded !== '') ?

lib/routes/routeBackbeat.js
SylvainSenechal: A check that the microVersionId in the header is the same as the microVersionId in the metadata body ? Yes why not, I added it as soon as the request payload is decoded

lib/routes/routeBackbeat.js
SylvainSenechal: right, i simplified this as its nested in an if that already does some of these checks

lib/routes/routeBackbeat.js
SylvainSenechal: I did this, I think it's nice to do to make the code more readable

One thing to note : in the event that the api receive a microVersionId but objMd is null (so on first replication), the microVerisonId wouldnt be validated by decode(), which we may want to do. 
I think it's very corner case and unlikely though

lib/routes/routeBackbeat.js
SylvainSenechal: So, I ended up not using the Arsenal utility function. If found it was getting over complicated to use, as I tried to make some kind of generic Arsenal comparison function that deals with null values, and multiple cases, but here we already have more context and focus only on doing comparison that we care about. 

Basically, I think if we wanted to use Arsenal utility, we would have to introduce Crr cascade business into arsenal.
And if we want to use the generic Arsenal function here, we will need to add logic on top of it to fit it for Crr Cascade

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,65 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+                    response, log, callback,
+                    StaleMicroVersionIdException.name,
+                    'incoming revision is older than destination',
+                    objMd?.microVersionId,
+                );
+            }
+        }
+    }
+
+
```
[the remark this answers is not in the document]
SylvainSenechal: Mhhh, no ? 

We introduce 100-continue for putData as it's data heavy
We don't particularly care about it for putMetada 🤔 Or else we will start using it everywhere

lib/routes/routeBackbeat.js
SylvainSenechal: Ok I went back on this :
This modification should indeed not be there, and instead it requires some small changes on backbeat to deal with metadata only writes on zero sized object, but yeah there is no need to make that modification here

lib/routes/routeBackbeat.js
SylvainSenechal: Applied + created low priority follow up https://scality.atlassian.net/browse/CLDSRV-950

lib/routes/routeBackbeat.js
SylvainSenechal: Ok I see, the change is mostly about semantics which make sense, I will make the change

For dealing with the absence of microVersionID, I would return an error, and I think this is fine as it will not just return an error when microVersionID is missing, it will return an error when 
- microVersionID is missing
- We are in a replication context
- The destination bucket also has a replication configuration which is basically a cascade scenario

lib/routes/routeBackbeat.js
SylvainSenechal: Ah you want to have something like 
a<->b, and when b receives the requests, it identifies the next replication is A, which is also where the current request came from, so it stops

I think from just 2 minutes of thinking we would need to pass a new header, that contains the location name. And here in backbeat compare the 2 locations names. 

Anyways creating a follow-up ticket

https://scality.atlassian.net/browse/CLDSRV-951

lib/routes/routeBackbeat.js
SylvainSenechal: This is outdated (was discussed in meeting and addressed with code change)

lib/routes/routeBackbeat.js
SylvainSenechal: I'm creating a ticket for this : https://scality.atlassian.net/browse/CLDSRV-952

tests/functional/backbeat/putMetadata.js
SylvainSenechal: Yeah the file is big enough that a split makes sense, also some tests can be added

SylvainSenechal:
Btw I will deal with prettier at the end

lib/routes/routeBackbeat.js
```diff
@@ -414,6 +432,38 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
+        const incomingVersionIdDecoded = decode(incomingVersionIdEncoded);
+        if (incomingVersionIdDecoded instanceof Error) {
+            log.error('crr putData: failed to decode x-scal-version-id header', {
+                method: 'putData',
+                error: incomingVersionIdDecoded.message,
+            });
+            return callback(errorInstances.BadRequest.customizeDescription(
+                'bad request: invalid x-scal-version-id header'));
+        }
+        if (objMd && objMd.versionId === incomingVersionIdDecoded) {
```
francoisferrand, line 453: need to go in this check for `incomingVersionIdEncoded === ''` as well (i.e. null version object in a versioned bucket)

```suggestion
const encodedVersionId = request.headers['x-scal-version-id'];
if (encodedVersionId !== undefined) {
   const versionId = encodedVersionId ? decode(encodedVersionId) : '';
   if (error) {
       ... 
   }
   if (objMD && objMd.versionId === versionId) {
       ...
   }
```

francoisferrand, line 453: we need to confirm how `objMD` is "retrieved", i.e. that it will always return the _version_ we require : for this specific check, we don't care that there is any other version or a maser, but only that this specific version is available...

→ If done, it would allow dropping the parsing of versionID here and comparison with the object. We would only need to check the condition "versionID specified && object found" : `if (request.headers['x-scal-version-id']!==undefined && objMD) { //conflict !!!`
→ If not done, the conflict detection may not work in some case, for exemple if replication a previous (non-current) version...

lib/routes/routeBackbeat.js
```diff
@@ -414,6 +432,38 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         log.error(errMessage);
         return callback(errorInstances.BadRequest.customizeDescription(errMessage));
     }
+
+    const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
+    if (incomingVersionIdEncoded != null && incomingVersionIdEncoded !== '') {
```
francoisferrand: should we not check `incomingVersionIdEncoded` against `undefined` to detect the absence/presence of field ?

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +589,64 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+            return callback(errorInstances.BadRequest.customizeDescription(
+                'bad request: invalid x-scal-micro-version-id header'));
+        }
+
+        // null = oldest revision (original putObject, before any metadata update).
+        // VersionID strings are in reverse chronological order: a lexicographically
+        // larger string is an older revision.
+        const isIncomingOlderThanCurrent = (incoming, current) =>
+            current !== null &&
+            (incoming === null || incoming > current);
```
francoisferrand: kind of hard to read (and validate that it works), would be best not to see this: i.e. use existing compareVersionId...

lib/routes/routeBackbeat.js
```diff
@@ -724,6 +838,44 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+            const objSize = omVal['content-length'] || 0;
+
+            // These S3-compatible Scality locations are excluded as cascade
+            // targets because they use the MultiBackend S3 path which bypasses
+            // the putData/putMetadata routes, so loop detection cannot fire
+            // on those destinations.
+            const cascadeBlockedLocationTypes = [
+                'location-scality-ring-s3-v1',
+                'location-scality-artesca-s3-v1',
+            ];
```
francoisferrand: should either be in the Config object (if we may need to change) or in constants.js

lib/routes/routeBackbeat.js
```diff
@@ -552,15 +660,20 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+        const { headers } = request;
 
         // Destination-side delete-marker replication.
         // We need the REPLICA status to distinguish from
         // source-side replication status updates that also carry isDeleteMarker=true.
         if (
             omVal.isDeleteMarker &&
             omVal.replicationInfo &&
-            omVal.replicationInfo.status === 'REPLICA' &&
+            (omVal.replicationInfo.isReplica === true || omVal.replicationInfo.status === 'REPLICA') &&
```
francoisferrand: nit: should we introduce a helper function? or maybe "standardize" `omVal.replicationInfo.isReplica` after the JSON.parse() call on line 658 ?

```js
        try {
            omVal = JSON.parse(payload);
			omVal.replicationInfo.isReplica ||= omVal.replicationInfo.status === 'REPLICA';
        } catch {
        }
```

lib/routes/routeBackbeat.js
francoisferrand: looking at the code of compare versionId in arsenal : it seems that function does not support comparing "null" versionID → should be reworked to support that case

Will need to standardize on what "null" is: `null`, `''` or `falsish`... Falsish is more flexible (does not require null coalescence after loading document...) but means there may not be an actual equality when the function return 0 (null != undefined)

lib/routes/routeBackbeat.js
```diff
@@ -724,6 +838,44 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
...
+
+            omVal.replicationInfo = hasNextHop ? nextReplInfo : {
+                status: '',
+                backends: [],
+                content: [],
+                destination: '',
+                storageClass: '',
+                role: '',
+                storageType: '',
+                dataStoreVersionId: '',
```
francoisferrand: nit: do we need to put an empty replicationInfo? can't we keep the field undefined?

lib/routes/routeBackbeat.js
SylvainSenechal: I don't have a big opinion on this. 
VersionID was already always passed from backbeat before, what's sure is right here for the collision check we need it

But I don't really know whether this function should work without a versionID provided. 
What I know is In cloudserverClient, version id header field is not mandatory, and in backbeat, the api is called in only one place with the versionId (and I doubt it's used anywhere else)

lib/routes/routeBackbeat.js
SylvainSenechal: Yes, and actually it was more tricky because I found the null versionID is stored as string "null"

lib/routes/routeBackbeat.js
SylvainSenechal: Prefer to keep, anyways we will always have at least
omVal.replicationInfo.isReplica = true;

lib/routes/routeBackbeat.js
SylvainSenechal: Correct, objMd is always the master, not queried with the specific versionID.

Follow-up : https://scality.atlassian.net/browse/CLDSRV-953

lib/routes/routeBackbeat.js
SylvainSenechal: We have this doc, although its not referenced directly in the code 🤔 https://scality.atlassian.net/wiki/spaces/OS/pages/edit-v2/4141449224

lib/routes/routeBackbeat.js
```diff
@@ -539,6 +591,66 @@ function getCanonicalIdsByAccountId(accountId, log, cb) {
...
+        }
+
+        const isIncomingOlderThanCurrent = (incoming, current) => {
+            if (current === null) { // nothing to be stale against
+                return false; 
+            }
+            // null incoming : source has no cascade revision => oldest possible state
+            // larger string = older revision (reverse-chronological order)
+            return incoming === null || incoming > current;
+        };
```
francoisferrand, line 630: is this not compareMicroVersionId ? Or is that "refactor" moved to later commit?

(still quite hard to read, even with the comments... but somehow can't see the original comment about this...)

lib/routes/routeBackbeat.js
francoisferrand: there is a bit of duplication everywhere:
- 2x 10 lines here (as initially pointed by Mael)
- matching 2 different errors in backbeat
- doing 2 separate comparison (== then strict older)

code here may be dedup by using compareVersionId, something like:
```
const diff = compareVersionId(current, incoming):
if (diff <= 0) {
    const err := diff === 0 ? MicroVersionIdAlreadyStoredException.name : StaleMicroVersionIdException.name;
   log.debug(....)
   request.resume();
   return _respondWithHeaderCrrConflict( ..., err, ... )
}
```

francoisferrand (approved):

PR #6182, opened 2026-06-02
Bump arsenal to 8.4.4

maeldonn:
Issue: CLDSRV-913


SylvainSenechal (approved):
we could run in zenko ci before merging if we want to be safer

maeldonn:
> we could run in zenko ci before merging if we want to be safer

Great idea @SylvainSenechal !

DarkIsDude (approved):

PR #6185, opened 2026-06-03
CLDSRV-915: Fix service user provisioning broken by AWS SDK v3 migration

tcarmet:
The AWS SDK v3 migration changed how IAM "entity not found" errors surface, and the service user provisioning script relied on the old shape. Since then a fresh environment can never create its service user: the lookup error is treated as fatal instead of triggering creation, and installs that provision one fail.

The script had no test coverage, which is how this slipped through. This PR also adds a functional test suite that exercises it end to end against a real IAM backend in CI, to catch mistakes in the script sooner.

### DIFF
--- a/bin/ensureServiceUser
+++ b/bin/ensureServiceUser
@@ -16,6 +16,7 @@ const {
     PutUserPolicyCommand,
     ListAccessKeysCommand,
     CreateAccessKeyCommand,
+    NoSuchEntityException,
 } = require('@aws-sdk/client-iam');
 const { version } = require('../package.json');
 
@@ -181,7 +182,7 @@ function collectResource(v, done) {
     v.collect()
         .then(res => done(null, res))
         .catch((err) => {
-            if (err.name === 'NoSuchEntity') {
+            if (err instanceof NoSuchEntityException) {
                 return done();
             }
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenko/cloudserver",
-    "version": "9.3.8",
+    "version": "9.3.9",
     "description": "Zenko CloudServer, an open-source Node.js implementation of a server handling the Amazon S3 protocol",
     "main": "index.js",
     "engines": {
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
         "ft_healthchecks": "cd tests/functional/healthchecks && yarn test",
         "ft_s3cmd": "cd tests/functional/s3cmd && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
         "ft_s3curl": "cd tests/functional/s3curl && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
+        "ft_scripts": "cd tests/functional/scripts && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
         "ft_util": "cd tests/functional/utilities && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
         "ft_test": "npm-run-all -s ft_awssdk ft_s3cmd ft_s3curl ft_node ft_healthchecks ft_management ft_util ft_backbeat",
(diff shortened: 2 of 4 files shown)


bin/ensureServiceUser
```diff
@@ -181,7 +181,9 @@ function collectResource(v, done) {
     v.collect()
         .then(res => done(null, res))
         .catch((err) => {
-            if (err.name === 'NoSuchEntity') {
```
SylvainSenechal, line 184: this is kinda weird as I spent quite a lot of time on sdkv2/v3 difference, and i'm pretty sure we need to look at err.name for aws sdk v3 🤔

SylvainSenechal: Could it be Vault not being fully compliant with s3 standard when dealing with errors ?

SylvainSenechal: btw many cases https://github.com/search?q=org%3Ascality%20err.name%20%3D%3D%3D%20%27NoSuchEntity%27&type=code

SylvainSenechal: yeah looking at other codebases and what we did for the migration, I think the proper way is 

err instanceof NoSuchEntityException
with NoSuchEntityException imported from @aws-sdk/client-iam
 
https://github.com/search?q=org%3Ascality+instanceof+NoSuchEntityException&type=code

tcarmet: alright yeah using `err instanceof NoSuchEntityException` looks good and better, thanks.

But yeah if I was to use `err.name` I would need to do `err.name === 'NoSuchEntityException'` for the code to work properly (you can play around with the new testsuite if you'd like). So I am a bit concerned by all the matches, if they are from the sdkv2 or arsenal that's fine, they should work, otherwise it's worth a overall look on the usage.

tests/functional/scripts/ensureServiceUser.js
```diff
@@ -0,0 +1,178 @@
...
+            AWS_SECRET_ACCESS_KEY: secretAccessKey,
+        },
+    });
+}
+
+async function ignoreNoSuchEntity(promise) {
+    try {
+        return await promise;
+    } catch (err) {
+        if (err.Error && err.Error.Code === 'NoSuchEntity') {
```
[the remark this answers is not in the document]
tcarmet: fair enough

tmacro (approved):

nicolas2bert (approved):

tests/functional/scripts/ensureServiceUser.js
```diff
@@ -0,0 +1,179 @@
...
+
+    it('should fail when the user exists outside the scality-internal path', async () => {
+        await iamClient.send(new CreateUserCommand({ UserName: userName }));
+
+        await assert.rejects(runScript(userName), err => {
+            assert.strictEqual(err.code, 1);
+            assert.ok(
+                err.stdout.includes('EntityAlreadyExists'),
+                `expected EntityAlreadyExists in output: ${err.stdout}`,
+            );
```
BourgoisMickael: Can be simplfied into one line
```suggestion
            assert.match(err.stdout, /EntityAlreadyExists/);
```

BourgoisMickael (approved):

package.json
```diff
@@ -126,6 +126,7 @@
         "ft_healthchecks": "cd tests/functional/healthchecks && yarn test",
         "ft_s3cmd": "cd tests/functional/s3cmd && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
         "ft_s3curl": "cd tests/functional/s3curl && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
+        "ft_scripts": "cd tests/functional/scripts && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
         "ft_util": "cd tests/functional/utilities && mocha --reporter mocha-multi-reporters --reporter-options configFile=$INIT_CWD/tests/reporter-config.json -t 40000 *.js --exit",
         "ft_test": "npm-run-all -s ft_awssdk ft_s3cmd ft_s3curl ft_node ft_healthchecks ft_management ft_util ft_backbeat",
```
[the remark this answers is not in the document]
tcarmet: on purpose, `ft_scripts` requires vault and and the only scenario where it is up is on the s3c tests. (today at least, we could change that but that's a different topic). 

Adding it here for now we should be fine.

PR #6187, opened 2026-06-10
CLDSRV-918: raise SDK request timeout in functional test client

tcarmet:
The AWS SDK v3 migration introduced a 5-second request timeout in the functional test client where SDK v2 previously ran with its 120-second default. Test operations that get no response bytes until the server finishes working - a 50MB server-side part copy, 1000-key batch deletes - regularly exceed 5 seconds of socket silence under CI load, producing `Connection timed out after 5000 ms` failures on more than a dozen unrelated branches over the past 60 days, including one merge-queue hit.

Raising the timeout to 30 seconds restores the pre-migration behavior within mocha's 40-second global cap: a genuine hang still fails the test, surfacing as the SDK's informative `TimeoutError` instead of a generic mocha timeout.

**Why this also bumps arsenal (ARSN-594, scality/Arsenal#2640):**

Raising the timeout uncovered a latent server-crash bug. In `@smithy/node-http-handler`, a `requestTimeout` below 6s is registered through `request.setTimeout()`, which Node cleans up when the request completes. At 6s and above, registration is deferred by 3s; for any request lasting longer than 3s the timer then attaches directly to the socket (`socket.setTimeout()`) and is never removed when the request finishes. With a keep-alive agent, the pooled socket keeps that stale timer and its handler later destroys the connection out from under whichever request is using it - so the 30s config turns every >3s test request into a delayed, random mid-stream disconnect.

A client vanishing mid-stream then hit a latent bug in arsenal's `retrieveData`: its `once()` guard around the iteration callback was created per data-layer callback invocation rather than per iteration, so a double callback during the teardown race ran the iteration callback twice, raising `Callback was already called` as an uncaughtException and killing the server. This crashed the `file-ft-tests` suites deterministically (3 out of 3 runs, 200+ cascading ECONNREFUSED failures each). The arsenal fix makes the guard iteration-scoped; a client disconnect can no longer take the server down.

**Reference for the previous 120s behavior:**
- The SDK v2 default: `aws-sdk@2.1692.0` ships `httpOptions: { timeout: 120000 }` ([`lib/config.js#L626-L628`](https://github.com/aws/aws-sdk-js/blob/v2.1692.0/lib/config.js#L626-L628)), documented as "Sets the socket to timeout after timeout milliseconds of inactivity on the socket. Defaults to two minutes (120000)" ([AWS.Config docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Config.html#httpOptions-property)).
- The pre-migration test config never overrode it: before the SDK v3 migration, [`tests/functional/aws-node-sdk/test/support/config.js` at `ea4659029`](https://github.com/scality/cloudserver/blob/ea46590291db2d2cd4af42a61c52aa68f32a25ad/tests/functional/aws-node-sdk/test/support/config.js) configured no timeout at all, so every functional test ran with the 120s default.

### DIFF
--- a/tests/functional/aws-node-sdk/test/support/config.js
+++ b/tests/functional/aws-node-sdk/test/support/config.js
@@ -22,15 +22,23 @@ const DEFAULT_GLOBAL_OPTIONS = {
     httpOptions,
 };
 
+// Heavy server-side operations (multi-megabyte UploadPartCopy, 1000-key
+// batch deletes) send no response bytes until they complete, so they can
+// exceed a short request timeout under CI load. 30s gives ample headroom
+// while staying under mocha's 40s global timeout, so a genuine hang still
+// surfaces as the SDK's informative TimeoutError rather than a bare mocha
+// timeout.
+const REQUEST_TIMEOUT = 30000;
+
 const DEFAULT_MEM_OPTIONS = {
     endpoint: `${transport}://127.0.0.1:8000`,
     port: 8000,
     forcePathStyle: true,
-    region: 'us-east-1', 
+    region: 'us-east-1',
     maxAttempts: 3,
     requestHandler: new NodeHttpHandler({
         connectionTimeout: 5000,
-        requestTimeout: 5000,
+        requestTimeout: REQUEST_TIMEOUT,
         httpAgent: new (ssl ? https : http).Agent({
             maxSockets: 200,
             keepAlive: true,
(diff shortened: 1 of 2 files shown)


jonathan-gramain (approved):

tests/functional/aws-node-sdk/test/support/config.js
```diff
@@ -22,15 +22,23 @@ const DEFAULT_GLOBAL_OPTIONS = {
     httpOptions,
 };
 
+// The SDK v2 config set no timeout, so tests ran with the v2 default of
+// 120s; the v3 migration introduced a 5s timeout that heavy server-side
+// operations (multi-megabyte UploadPartCopy, 1000-key batch deletes) exceed
+// under CI load, as the server sends no response bytes until they complete.
+// 30s stays below mocha's 40s global timeout so a genuine hang still
+// surfaces as the SDK TimeoutError rather than a generic mocha timeout.
```
dvasilas: I would argue that when we read this comment in the future, we will not care about the history (migration changed it to 5s), but only about why we chose 30s as the value.

dvasilas (approved):

PR #6188, opened 2026-06-12
CLDSRV-925: UploadPartCopy handle checksums

leif-scality:
- Handle checksums in UploadPartCopy. Don't compute and store a checksum for parts in external backends (same as PutPart).
- fix: CompleteMPU don't require per-part checksums for external backends. UploadPart and UploadPartCopy don't store a checksum for external backend parts so we cannot require them. The final object will not have a checksum.
- fix: UploadPart reject Parts missing a COMPOSITE checksum when required
- fix: UploadPart returning the default checksum when it should not
- fix: CopyObject allow in-place checksum change on self-copy

# UploadPartCopy — when the checksum is recomputed vs reused vs skipped

| # | MPU algorithm | Dest backend | `copy-source-range` present | Source's stored checksum | Checksum action | Returned in CopyPartResult? |
|---|---|---|---|---|---|---|
| 1 | none (legacy MPU, pre-feature) | any | any | any | **None** — nothing stored/returned | – |
| 2 | client-provided (explicit) or default (implicit crc64nvme) | **external** (aws_s3 / azure / gcp / …) | any | any | **None** — skipped, matches `UploadPart` | no |
| 3 | client-provided (explicit) or default (implicit crc64nvme) | local (sproxyd / file / mem) | **yes (ranged)** | any | **Recompute** — over the copied range | explicit: yes · default: no (stored only) |
| 4 | client-provided (explicit) or default (implicit crc64nvme) | local (sproxyd / file / mem) | no | **FULL_OBJECT, same algorithm** | **Reuse** — copy source's stored value | explicit: yes · default: no (stored only) |
| 5 | client-provided (explicit) or default (implicit crc64nvme) | local (sproxyd / file / mem) | no | COMPOSITE, different algorithm, or none | **Recompute** — over the whole part | explicit: yes · default: no (stored only) |

### DIFF
--- a/lib/api/objectCopy.js
+++ b/lib/api/objectCopy.js
@@ -1,5 +1,4 @@
 const async = require('async');
-const { PassThrough } = require('stream');
 
 const { errors, errorInstances, jsutil, versioning, s3middleware, s3routes } = require('arsenal');
 const { validateObjectKeyLength } = s3routes.routesUtils;
@@ -33,6 +32,7 @@ const {
 } = require('./apiUtils/integrity/validateChecksums');
 const ChecksumTransform = require('../auth/streamingV4/ChecksumTransform');
 const ChecksumWritable = require('../auth/streamingV4/ChecksumWritable');
+const { buildSourcePartsStream } = require('./apiUtils/object/sourceChecksum');
 const kms = require('../kms/wrapper');
 
 const versionIdUtils = versioning.VersionID;
@@ -64,63 +64,6 @@ function _orphanedDataLocations(dataToDelete, newDataGetInfo) {
     const newIds = new Set((newDataGetInfo || []).map(locId));
     const orphans = (dataToDelete || []).filter(loc => !newIds.has(locId(loc)));
     return orphans.length > 0 ? orphans : null;
-}
-
-/**
- * Concatenate the source object's parts into a single Readable stream by
- * reading them sequentially through `data.get`. Returns the PassThrough
- * immediately; consumers should pipe it to the next stage and observe
- * `error` on this stream for any read failure along the way.
- *
- * @param {Array} dataLocator - ordered source parts
- * @param {object} log - request logger
- * @return {PassThrough}
- */
-function _pipeSourcePartsThrough(dataLocator, log) {
-    const passthrough = new PassThrough();
-    const wrapErr = (err, part) =>
-        Object.assign(err, {
-            copyPart: { key: part.key, dataStoreName: part.dataStoreName, dataStoreType: part.dataStoreType },
-        });
-    async.eachSeries(
-        dataLocator,
-        (part, cb) => {
-            const done = jsutil.once(cb);
-            if (part.dataStoreType === 'azure') {
-                // Azure's data.get writes part bytes into the provided writable
-                // instead of returning a Readable. Pipe a per-part PassThrough
-                // into the master passthrough and use its 'end' as the completion
-                // signal — same pattern arsenal's data.copyObject uses.
-                const perPart = new PassThrough();
-                perPart.once('error', err => done(wrapErr(err, part)));
-                perPart.once('end', () => done());
-                perPart.pipe(passthrough, { end: false });
-                return data.get(part, perPart, log, err => {
-                    if (err) {
-                        perPart.destroy(err);
-                        done(wrapErr(err, part));
-                    }
-                });
-            }
-            return data.get(part, null, log, (err, partStream) => {
-                if (err) {
-                    return done(wrapErr(err, part));
-                }
-                partStream.once('error', err => done(wrapErr(err, part)));
-                partStream.once('end', () => done());
-                partStream.pipe(passthrough, { end: false });
-                return undefined;
-            });
-        },
-        err => {
-            if (err) {
-                passthrough.destroy(err);
-            } else {
-                passthrough.end();
-            }
-        },
-    );
-    return passthrough;
 }
 
 /**
@@ -207,7 +150,7 @@ function _recomputeChecksumAndStore(
             algorithm: algoName,
             size: storeMetadataParams.size,
         });
-        const sourceStream = _pipeSourcePartsThrough(dataLocator, log);
+        const sourceStream = buildSourcePartsStream(dataLocator, log);
         const checksumSink = new ChecksumWritable(algoName, log);
         const finish = jsutil.once(err => {
             if (err) {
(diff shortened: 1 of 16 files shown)


tests/unit/api/objectCopyPart.js
```diff
@@ -155,3 +162,238 @@ describe('objectCopyPart', () => {
         });
     });
 });
+
+describe('objectPutCopyPart._shouldRecomputeChecksum', () => {
+    const { _shouldRecomputeChecksum } = objectPutCopyPart;
+    const noRange = { headers: {} };
+    const withRange = { headers: { 'x-amz-copy-source-range': 'bytes=0-1' } };
+
+    it('returns true when a copy-source-range is requested', () => {
```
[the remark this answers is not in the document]
leif-scality: fixed

tests/functional/aws-node-sdk/test/object/mpuUploadPartChecksum.js
```diff
@@ -129,7 +129,16 @@ describe('UploadPart checksum validation', () =>
                                 PartNumber: 3 + idx, Body: partBody,
                                 [checksumField[otherAlgo]]: correctDigest[otherAlgo],
                             })),
-                            { name: 'InvalidRequest' },
+                            err => {
+                                assert.strictEqual(err.name, 'InvalidRequest',
+                                    `expected InvalidRequest, got ${err.name}: ${err.message}`);
+                                // AWS names the expected (MPU) and actual (sent) algorithms.
+                                assert(err.message.includes(
```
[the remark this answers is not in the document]
leif-scality: fiedx

lib/api/objectPutCopyPart.js
```diff
@@ -297,12 +380,35 @@ function objectPutCopyPart(authInfo, request, sourceBucket,
...
+                        // eslint-disable-next-line no-param-reassign
+                        request.actionImplicitDenies = originalIdentityAuthzResults;
+                        if (err) {
+                            // eslint-disable-next-line no-param-reassign
+                            request.sourceServerAccessLog && (request.sourceServerAccessLog.error = err);
+                            return next(err, destBucketMD);
+                        }
+                        return next(null, destBucketMD, result.locations, result.totalHash, copyObjectSize,
+                            sourceVerId, sse, new Date().toJSON(), splitter, mpuOverviewMD, result.checksum);
+                    });
```
anurag4DSB: Do we know how much time do we spend here? How many milliseconds do we add?

anurag4DSB (approved):

lib/api/objectPutCopyPart.js
leif-scality: No. There is a task open to measure the cost of checksums, it is going to be done next

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -2135,5 +2135,73 @@ describe('Object Copy checksum behavior', () => {
...
+                        Key: sourceObjName,
+                        Body: selfCopyBody,
+                        ChecksumAlgorithm: sourceHeader,
+                    }),
+                );
+                const expectedDigest = await Promise.resolve(algorithms[algo].digest(Buffer.from(selfCopyBody)));
+
+                // Copy the object onto itself, changing only the checksum
+                // algorithm. MetadataDirective is omitted, so it defaults to
+                // COPY — the case the guard wrongly rejects.
```
dvasilas:
```suggestion
                // COPY
```

tests/functional/aws-node-sdk/test/object/objectCopy.js
```diff
@@ -2135,5 +2135,73 @@ describe('Object Copy checksum behavior', () => {
...
+                );
+                assert.strictEqual(headRes[key], expectedDigest);
+                assert.strictEqual(headRes.ChecksumType, 'FULL_OBJECT');
+            });
+        });
+
+        // The checksum carve-out must not open the guard for a genuine no-op:
+        // a self-copy with no checksum algorithm and no other change is still
+        // rejected.
+        it('should still reject a no-change self-copy (COPY directive)', async () => {
```
dvasilas:
```suggestion
        it('should reject a no-change self-copy (COPY directive)', async () => {
```

lib/api/objectCopy.js
```diff
@@ -362,11 +362,14 @@ function _prepMetadata(
...
     // Cannot copy from same source and destination if no MD
-    // changed and no source version id
+    // changed and no source version id. A requested checksum algorithm counts
+    // as a change (it (re)computes the object's checksum in place), matching
+    // AWS, so it is excluded from this guard.
     if (
         sourceIsDestination &&
         whichMetadata === 'COPY' &&
         Object.keys(overrideMetadata).length === 0 &&
+        !headers['x-amz-checksum-algorithm'] &&
```
dvasilas: This means we will do the copy even if the source object already has the requested checksum (and we requested the same algorithm, so no change in the checksum), right?

Is this what AWS does?

Should we instead use ` _shouldRecomputeChecksum`?

leif-scality: The original guard change was wrong. I checked AWS on a bucket with Bucket Keys turned on, where every self-copy looks like an encryption change so they all succeed; on a normal bucket AWS rejects a COPY self-copy that only sets a checksum and needs MetadataDirective=REPLACE instead.

The commit was changed and I only added a unit test to verify we follow the AWS behaviour

tests/functional/aws-node-sdk/test/object/objectCopy.js
leif-scality: done

tests/functional/aws-node-sdk/test/object/objectCopy.js
leif-scality: done

dvasilas (approved):

lib/api/objectPutPart.js
```diff
@@ -330,16 +339,21 @@ function objectPutPart(authInfo, request, streamingV4Params, log,
...
-                    `Checksum algorithm '${headerChecksum.algorithm}' is not the same ` +
-                    `as the checksum algorithm '${mpuChecksumAlgo}' specified during ` +
-                    'CreateMultipartUpload.'
-                ), destinationBucket);
+                return next(checksumTypeMismatchErr(mpuChecksumAlgo, headerChecksum.algorithm), destinationBucket);
+            }
+
+            // A COMPOSITE MPU's final checksum is composed from the per-part
+            // checksums, so every part must carry one.
+            if (!headerChecksum && mpuChecksumType === 'COMPOSITE') {
```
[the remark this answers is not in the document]
leif-scality: Not an issue — the UploadPart COMPOSITE check (objectPutPart.js:355) sits after the external early-returns (aws_s3 :272, azure/gcp :320), so it only applies to local backends, matching CompleteMPU's exemption. _initiateExternalMpu uploads onto an aws_s3 MPU, so its checksum-less part returns at :272 before the check — which is why the external-backends suite passes (locally and in the CI unit-tests job).

leif-scality:
ping

PR #6189, opened 2026-06-15
CLDSRV-926: skip async/await diff check when PR has skip-async-migration label

DarkIsDude:
## Summary

- Adds `pull-requests: read` permission to the `tests.yaml` workflow
- The "Check async/await compliance in diff" step now checks whether the current branch's open PR carries the `skip-async-migration` label before running `check-diff-async`
- If the label is present the step exits cleanly; otherwise the check runs as before

## Behaviour

| Scenario | Result |
|---|---|
| PR has `skip-async-migration` label | Step skips, exits 0 |
| PR does not have the label | `check-diff-async` runs normally |
| No open PR (direct push / `workflow_dispatch`) | `check-diff-async` runs normally |

## Pre-requisite

The `skip-async-migration` label must be created in the repository settings before use.

Closes CLDSRV-926

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -109,11 +109,10 @@ jobs:
         run: yamllint -c yamllint.yml $(git ls-files "*.yml")
       - name: Check async/await compliance in diff
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ github.token }}
           BRANCH_NAME: ${{ github.ref_name }}
         run: |
-          if gh pr list --head "$BRANCH_NAME" --state open --json labels \
-               --jq '.[].labels[].name' 2>/dev/null | grep -qx 'skip-async-migration'; then
+          if gh pr list --head "$BRANCH_NAME" --state open --label "skip-async-migration" 2>/dev/null | grep -q .; then
             echo "Skipping async/await check: 'skip-async-migration' label found"
           else
             yarn run check-diff-async
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -73,6 +73,7 @@ env:
 permissions:
   contents: read
   packages: write
+  pull-requests: read
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -107,7 +108,16 @@ jobs:
       - name: Lint Yaml
         run: yamllint -c yamllint.yml $(git ls-files "*.yml")
       - name: Check async/await compliance in diff
-        run: yarn run check-diff-async
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: ${{ github.ref_name }}
```
francoisferrand, line 113: do we need to list PRs from bash script?
can't we use an `if` condition, something like:

```suggestion
        `if: ${{ contains(github.event.pull_request.labels.*.name, 'skip-async-migration' }}`
```

francoisferrand: esp. check would fail if PR is not opened...

francoisferrand: thinking on it: maybe both these refactoring checks should ONLY be performed on PR event (not on push), so they check only the code we modify and don't force people to add reformatting/... in waterflow branches?

DarkIsDude: Was my first guess to use an if but as you said, to have something working we must trigger only PR events.

We talk about not doing that on waterflow but we decided to don't do it because we can mess the convergence so the answer should be no.

I didn't find a better pattern to have something simpler.

.github/workflows/tests.yaml
```diff
@@ -107,7 +108,16 @@ jobs:
       - name: Lint Yaml
         run: yamllint -c yamllint.yml $(git ls-files "*.yml")
       - name: Check async/await compliance in diff
-        run: yarn run check-diff-async
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
francoisferrand: nit: `github.token` can be used instead, and better IMHO (can be used in default parameters or actions, not a "fake" secret created temporarily...)

.github/workflows/tests.yaml
```diff
@@ -107,7 +108,16 @@ jobs:
...
+          BRANCH_NAME: ${{ github.ref_name }}
+        run: |
+          if gh pr list --head "$BRANCH_NAME" --state open --json labels \
+               --jq '.[].labels[].name' 2>/dev/null | grep -qx 'skip-async-migration'; then
```
francoisferrand: nit: can't we use the `--label` parameter to get only skip-async-migration PRs?

```
if gh pr list --head "$BRANCH_NAME" --state open --label skip-async-migration ; then
```

francoisferrand (approved):

maeldonn (approved):

delthas (approved):

PR #6190, opened 2026-06-15
CLDSRV-922: do not reference null version from master in null-key mode

tcarmet:
When replicating a version on top of a pre-existing null version, the destination master could end up referencing that null version in a way the storage layout no longer supports, which intermittently left a bucket non-empty after all its objects were deleted. This change makes replication follow the same null-version contract as the regular write path so the leftover key can no longer occur.

Added a deterministic test that asserts the master references the null version only in compatibility mode; it fails on the old behavior and passes on the new. The full backbeat-route functional suite was run against a bucketd-backed stack in both compatibility modes with no regressions.

### DIFF
--- a/tests/multipleBackend/routes/routeBackbeatForReplication.js
+++ b/tests/multipleBackend/routes/routeBackbeatForReplication.js
@@ -59,6 +59,8 @@ const dst = {
 
 const testData = 'testkey data';
 
+const isNullVersionCompatMode = process.env.ENABLE_NULL_VERSION_COMPAT_MODE === 'true';
+
 // Generate unique bucket name function
 function generateUniqueBucketName(prefix) {
     return `${prefix}-${uuidv4().substring(0, 8)}`;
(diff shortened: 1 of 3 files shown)


anurag4DSB (approved):

dvasilas (approved):

tests/multipleBackend/routes/routeBackbeatForReplication.js
```diff
@@ -1179,6 +1181,102 @@ describe(`backbeat routes for replication (${name})`, () => {
...
+                    return next(err);
+                }
+                return next(null, data);
+            }),
+        }, (err, results) => {
+            if (err) {
+                return done(err);
+            }
+
+            const masterMD = objectMDFromRequestBody(results.getMasterMetadata);
```
nicolas2bert: nice to have: check that we can empty the bucket?

nicolas2bert (approved):

PR #6192, opened 2026-06-18
Support Days=0 lifecycle expiration: bump arsenal and add functional tests

delthas:
## What

CloudServer side of `Days = 0` lifecycle support — bumps arsenal to the released **8.5.6** and wires it up:

1. **Apply prettier** to the touched `putBucketLifecycle` handler + test.
2. **Pass the werelogs logger** to `LifecycleConfiguration` so arsenal can emit the 0-day audit log.
3. **Functional tests** (`tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js`):
   - `Expiration` `Days: 0`, `NoncurrentVersionExpiration` `NoncurrentDays: 0`, and `AbortIncompleteMultipartUpload` `DaysAfterInitiation: 0` accepted and round-trip via GetBucketLifecycle
   - negative day values rejected with the exact AWS wording (`'<field>' for <action> action must be a positive integer`), `> MAX_DAYS` rejected (`MalformedXML`)
4. **Bump arsenal to `8.5.6`** — the released tag carrying the Days=0 validation, the AWS-matching error wording, and the multi-backend sproxyd `head()` crash fix (ARSN-607).
5. **Migrate `bucketPutLifecycle` to async/await** — clears the CodeQL "callback-style function" alert, using the dual callback+async trampoline (per the async-migration TAD).

## Why

`Days = 0` is an explicit *"empty this bucket"* lifecycle signal. It was rejected everywhere until now, so it's unambiguous (unlike `Days = 1`, used by legitimate short-retention buckets) and lets backbeat (BB-779) skip building costly lifecycle indexes for buckets being drained. CloudServer delegates lifecycle validation to arsenal, so the substantive change is the dependency bump; the rest wires the logger, adds coverage, and migrates the handler.

## Related

- ARSN-597 (arsenal) — Days=0 validation + AWS error wording; merged, released in **8.5.6** (scality/Arsenal#2648)
- ARSN-607 (arsenal) — multi-backend sproxyd `head()` crash fix, also in 8.5.6
- BB-779 / backbeat#2748 — skip-index heuristic that consumes Days=0

Issue: CLDSRV-928

### DIFF
--- a/lib/api/bucketPutLifecycle.js
+++ b/lib/api/bucketPutLifecycle.js
@@ -1,4 +1,4 @@
-const { waterfall } = require('async');
+const { promisify } = require('util');
 const uuid = require('uuid').v4;
 const LifecycleConfiguration = require('arsenal').models.LifecycleConfiguration;
 
@@ -15,11 +15,16 @@ const monitoring = require('../utilities/monitoringHandler');
  * @param {AuthInfo} authInfo - Instance of AuthInfo class with requester's info
  * @param {object} request - http request object
  * @param {object} log - Werelogs logger
- * @param {function} callback - callback to server
- * @return {undefined}
+ * @param {function} [callback] - callback to server
+ * @return {Promise} - resolves with the CORS response headers
  */
+async function bucketPutLifecycle(authInfo, request, log, callback) {
+    if (callback) {
+        return bucketPutLifecycle(authInfo, request, log)
+            .then(corsHeaders => callback(null, corsHeaders))
+            .catch(err => callback(err, err.additionalResHeaders));
+    }
 
-function bucketPutLifecycle(authInfo, request, log, callback) {
     log.debug('processing request', { method: 'bucketPutLifecycle' });
 
     const { bucketName } = request;
@@ -29,51 +34,33 @@ function bucketPutLifecycle(authInfo, request, log, callback) {
         requestType: request.apiMethods || 'bucketPutLifecycle',
         request,
     };
-    return waterfall(
-        [
-            next => parseXML(request.post, log, next),
-            (parsedXml, next) => {
-                const lcConfigClass = new LifecycleConfiguration(parsedXml, log, config);
-                // if there was an error getting lifecycle configuration,
-                // returned configObj will contain 'error' key
-                process.nextTick(() => {
-                    const configObj = lcConfigClass.getLifecycleConfiguration();
-                    if (configObj.error) {
-                        return next(configObj.error);
-                    }
-                    return next(null, configObj);
-                });
-            },
-            (lcConfig, next) =>
-                standardMetadataValidateBucket(metadataValParams, request.actionImplicitDenies, log, (err, bucket) => {
-                    if (err) {
-                        return next(err, bucket);
-                    }
-                    return next(null, bucket, lcConfig);
-                }),
-            (bucket, lcConfig, next) => {
-                if (!bucket.getUid()) {
-                    bucket.setUid(uuid());
-                }
-                bucket.setLifecycleConfiguration(lcConfig);
-                metadata.updateBucket(bucket.getName(), bucket, log, err => next(err, bucket));
-            },
-        ],
-        (err, bucket) => {
-            const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
-            if (err) {
-                log.trace('error processing request', { error: err, method: 'bucketPutLifecycle' });
-                monitoring.promMetrics('PUT', bucketName, err.code, 'putBucketLifecycle');
-                return callback(err, corsHeaders);
-            }
-            pushMetric('putBucketLifecycle', log, {
-                authInfo,
-                bucket: bucketName,
-            });
-            monitoring.promMetrics('PUT', bucketName, '200', 'putBucketLifecycle');
-            return callback(null, corsHeaders);
-        },
-    );
+
+    let bucket;
+    try {
+        const parsedXml = await promisify(parseXML)(request.post, log);
+        const lcConfig = new LifecycleConfiguration(parsedXml, log, config).getLifecycleConfiguration();
+        if (lcConfig.error) {
+            throw lcConfig.error;
+        }
+        bucket = await promisify(standardMetadataValidateBucket)(metadataValParams, request.actionImplicitDenies, log);
+        if (!bucket.getUid()) {
+            bucket.setUid(uuid());
+        }
+        bucket.setLifecycleConfiguration(lcConfig);
+        await promisify(metadata.updateBucket).call(metadata, bucket.getName(), bucket, log);
+    } catch (err) {
+        log.trace('error processing request', { error: err, method: 'bucketPutLifecycle' });
+        monitoring.promMetrics('PUT', bucketName, err.code, 'putBucketLifecycle');
+        err.additionalResHeaders ||= collectCorsHeaders(request.headers.origin, request.method, bucket);
+        throw err;
+    }
+
+    pushMetric('putBucketLifecycle', log, {
+        authInfo,
+        bucket: bucketName,
+    });
+    monitoring.promMetrics('PUT', bucketName, '200', 'putBucketLifecycle');
+    return collectCorsHeaders(request.headers.origin, request.method, bucket);
 }
 
 module.exports = bucketPutLifecycle;
--- a/lib/api/bucketPutLifecycle.js
+++ b/lib/api/bucketPutLifecycle.js
@@ -1,7 +1,6 @@
 const { waterfall } = require('async');
 const uuid = require('uuid').v4;
-const LifecycleConfiguration =
-    require('arsenal').models.LifecycleConfiguration;
+const LifecycleConfiguration = require('arsenal').models.LifecycleConfiguration;
 
 const config = require('../Config').config;
 const parseXML = require('../utilities/parseXML');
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
         "@azure/storage-blob": "^12.28.0",
         "@hapi/joi": "^17.1.1",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/Arsenal#8.4.4",
+        "arsenal": "git+https://github.com/scality/Arsenal#f5d43abeb8139ec8835ef0354888ff13ddb14d8b",
```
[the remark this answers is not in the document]
delthas: Thanks Sherlock

tests/functional/aws-node-sdk/test/bucket/putBucketLifecycle.js
```diff
@@ -89,7 +90,69 @@ describe('aws-sdk test put bucket lifecycle', () => {
...
+                LifecycleConfiguration: {
+                    Rules: [{
+                        ID: 'test-id',
+                        Status: 'Enabled',
+                        Prefix: '',
+                        NoncurrentVersionExpiration: { NoncurrentDays: 0 },
+                    }],
+                },
+            };
+            await s3.send(new PutBucketLifecycleConfigurationCommand(params));
```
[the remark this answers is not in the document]
delthas: Addressed in 59fb8a7b — the `NoncurrentDays=0` and `DaysAfterInitiation=0` tests now call `GetBucketLifecycleConfiguration` and assert the value round-trips as `0`, matching the `Expiration Days=0` test.

DarkIsDude (changes requested):
@delthas CI is broken

delthas:
> @delthas CI is broken

Yes thats due to https://github.com/scality/Arsenal/pull/2651 https://github.com/scality/cloudserver/pull/6198 

Reviews welcome :D

lib/api/bucketPutLifecycle.js
```diff
@@ -16,11 +15,16 @@
...
  * @param {object} log - Werelogs logger
- * @param {function} callback - callback to server
- * @return {undefined}
+ * @param {function} [callback] - callback to server
+ * @return {Promise} - resolves with the CORS response headers
  */
+async function bucketPutLifecycle(authInfo, request, log, callback) {
+    if (callback) {
+        return bucketPutLifecycle(authInfo, request, log)
+            .then(corsHeaders => callback(null, corsHeaders))
```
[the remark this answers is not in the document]
delthas: This `.then().catch()` is the intended async-migration **trampoline**, kept deliberately — not an oversight.

It's the reviewed pattern established for the async/await migration effort:

- Introduced in #6045 (commit `da1caeb7d`, *"use trampoline function to keep backward compatibility"*) and used identically in `bucketGet` and `objectGetLegalHold`.
- The dual callback+async interface enables *progressive* migration: callers still passing a callback behave exactly as before; callers migrated later can `await` the promise (omit the callback).
- It is intentionally `.then().catch()` rather than `await` + `try/catch` with the callback inside the `try`. In #6045's review, that shape was explicitly rejected as a pattern that *"must not be done … risk to invoke the callback multiple times"* — if the continuation callback throws, the `catch` re-invokes it.
- Documented in the migration TAD: https://scality.atlassian.net/wiki/spaces/OS/pages/3523346468/

So this is a known false-positive against the migration idiom (the same reason the matching handlers carry the eslint `prefer-await-to-then` warning without suppression). Leaving the trampoline as-is.

SylvainSenechal (approved):

lib/api/bucketPutLifecycle.js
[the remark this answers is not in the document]
francoisferrand: you can use callbackify to avoid the warning?
(this is a message from CodeQL, not Claude: so you cannot "convince" him you are right...)

maeldonn (approved):

lib/api/bucketPutLifecycle.js
```diff
@@ -16,11 +15,16 @@ const monitoring = require('../utilities/monitoringHandler');
...
+            .catch(err => callback(err, err.additionalResHeaders));
```
francoisferrand, line 25: shouldn't we use `callbackify` to streamline code and avoid mistake?

lib/api/bucketPutLifecycle.js
```diff
@@ -30,53 +34,34 @@ function bucketPutLifecycle(authInfo, request, log, callback) {
...
         request,
     };
-    return waterfall([
-        next => parseXML(request.post, log, next),
-        (parsedXml, next) => {
-            const lcConfigClass =
-                new LifecycleConfiguration(parsedXml, config);
-            // if there was an error getting lifecycle configuration,
-            // returned configObj will contain 'error' key
-            process.nextTick(() => {
```
francoisferrand, line 40: do you know why `nextTick` here? we are now removing it, would it create some issue?

lib/api/bucketPutLifecycle.js
```diff
@@ -30,53 +34,34 @@ function bucketPutLifecycle(authInfo, request, log, callback) {
...
-            return callback(err, corsHeaders);
+
+    let bucket;
+    try {
+        const parsedXml = await promisify(parseXML)(request.post, log);
+        const lcConfig = new LifecycleConfiguration(parsedXml, log, config).getLifecycleConfiguration();
+        if (lcConfig.error) {
+            throw lcConfig.error;
+        }
+        bucket = await promisify(standardMetadataValidateBucket)(metadataValParams, request.actionImplicitDenies, log);
```
francoisferrand, line 45: this is not the same as before, bucket would be lost on error: `standardMetadataValidateBucket` may return both an error AND the bucket -typically on validation error- which is needed for `collectCorsHeaders()` in the error path

package.json
```diff
@@ -37,7 +37,7 @@
         "@opentelemetry/instrumentation-ioredis": "~0.64.0",
         "@opentelemetry/instrumentation-mongodb": "~0.69.0",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/arsenal#8.5.2",
+        "arsenal": "git+https://github.com/scality/arsenal#58effc7702e14110cfb1564c6fe09242b57ed804",
```
DarkIsDude: Warning 🙏

DarkIsDude (approved):

francoisferrand (changes requested):

lib/api/bucketPutLifecycle.js
DarkIsDude: Also good point but this function are already promisied here :

```
standardMetadataValidateBucketAndObj[promisify.custom] = (params, ...rest) =>
    new Promise((resolve, reject) => {
        standardMetadataValidateBucketAndObj(params, ...rest, (err, bucket, objectMD) => {
            if (err) {
                return reject(attachCorsHeadersToError(err, bucket, params));
            }
            return resolve({ bucket, objectMD });
        });
    });
```

lib/api/bucketPutLifecycle.js
delthas: The error path passes the CORS headers as the callback's 2nd argument — the router does `(err, corsHeaders)` (see `lib/api/api.js`: *"Callback arity varies per route ((err, corsHeaders), ...)"*). `util.callbackify` only calls `callback(err)` on rejection, so it would drop `corsHeaders` on the error path.

The `.then().catch()` trampoline is the established async-migration pattern in cloudserver for exactly this reason (introduced with #6045 / the async-migration TAD), so I kept it for consistency with the other migrated handlers.

lib/api/bucketPutLifecycle.js
delthas: The `process.nextTick` in the original wrapped the *synchronous* `getLifecycleConfiguration()` + `next()` inside the `async.waterfall` step, deferring the synchronous `next()` off the current tick. In async/await there is no `next`, and the preceding `await promisify(parseXML)(...)` already yields to the event loop — so running `getLifecycleConfiguration()` synchronously and `throw`ing on error is equivalent. No behavior change from removing it.

lib/api/bucketPutLifecycle.js
delthas: Confirmed — as @DarkIsDude noted, `standardMetadataValidateBucket[promisify.custom]` rejects via `attachCorsHeadersToError(err, bucket, params)`, which sets `err.additionalResHeaders = collectCorsHeaders(origin, method, bucket)` from the bucket. So the bucket is not lost for CORS on validation errors — the headers ride on the rejected error, and the `catch` keeps them (`err.additionalResHeaders || collectCorsHeaders(...)`). The `||` fallback only applies to errors with no bucket (parseXML / config / updateBucket), where `collectCorsHeaders(undefined)` yields the same default the original produced.

lib/api/bucketPutLifecycle.js
```diff
@@ -30,53 +34,34 @@ function bucketPutLifecycle(authInfo, request, log, callback) {
...
-        });
-        monitoring.promMetrics('PUT', bucketName, '200', 'putBucketLifecycle');
-        return callback(null, corsHeaders);
+        bucket.setLifecycleConfiguration(lcConfig);
+        await promisify(metadata.updateBucket).call(metadata, bucket.getName(), bucket, log);
+    } catch (err) {
+        log.trace('error processing request', { error: err, method: 'bucketPutLifecycle' });
+        monitoring.promMetrics('PUT', bucketName, err.code, 'putBucketLifecycle');
+        err.additionalResHeaders =
+            err.additionalResHeaders || collectCorsHeaders(request.headers.origin, request.method, bucket);
```
francoisferrand:
```suggestion
        err.additionalResHeaders ||= collectCorsHeaders(request.headers.origin, request.method, bucket);
```

francoisferrand (approved):

package.json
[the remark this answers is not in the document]
delthas: Now pinned to the released `8.5.6` tag (arsenal 8.5.6) instead of a commit hash — resolving.

package.json
delthas: Done — now pinned to the released `8.5.6` tag.

lib/api/bucketPutLifecycle.js
delthas: Applied in 77689ebb3 (`||=`) — thanks.

PR #6125, opened 2026-03-31
CLDSRV-880 Fix formatting bugs in Claude review-pr skill

nicolas2bert:
## Summary

- Suggestion blocks now use heredoc (`-F body=@-` with `<<'COMMENT_BODY'`) instead of `$'...'` quoting, which broke on single quotes
- Summary comments now ban markdown headings (rendered as giant bold text in GitHub)
- Simplified formatting section (~43 lines removed, ~36 added)

PR #6194, opened 2026-06-18
Adopt the shared Claude review skill and modernize the review workflow

delthas:
## Summary

Two related changes bringing CloudServer's Claude review CI up to the shared standard (vault2 / Zenko / pensieve-api / zenko-operator).

### 1. Adopt the shared review skill

- Move CloudServer's repo-specific review criteria from the vendored `.claude/skills/review-pr/SKILL.md` into `.github/copilot-instructions.md` (read by the marketplace skill).
- Delete the vendored skill — the reusable `claude-code-review.yml@v2` workflow already loads `scality-skills@scality-agent-hub`, and the vendored copy otherwise shadows it at runtime.
- `CLAUDE.md` stays as repo context.

This supersedes the vendored-skill formatting fix in CLDSRV-880 / #6125 — once the skill is deleted, it lives only in agent-hub.

### 2. Modernize the review workflow

- Forward secrets via `secrets: inherit` (instead of enumerating each one in both jobs).
- Add a dynamic `run-name` and a `workflow_dispatch` trigger (with `pr_number` input) for manual re-runs, plus the `allowed-tools` passthrough.
- The `review` job now also runs on `workflow_dispatch` (guarded off only for `pull_request_target` + dependabot). The existing `review-dependency-bump` job is unchanged.

Note: prettier re-indented `review.yml` to the repo's 4-space YAML style (it had never been prettier-checked before, since prettier only runs on changed files), which inflates the diff — the semantic changes are the secrets/run-name/dispatch ones above.

Issue: CLDSRV-930


francoisferrand (approved):
don't we need to upgrade the GHA workflow file to latest standards as well?

DarkIsDude (approved):

delthas:
> don't we need to upgrade the GHA workflow file to latest standards as well?

Done

PR #6196, opened 2026-06-19
Fix empty error object logged for AWS location healthcheck failures

maeldonn:
AWS SDK throws S3ServiceException, whose inherited Error fields (message, name, stack) are non-enumerable and dropped by JSON.stringify, so failures surfaced as "error":{} in startup logs. Normalize Error instances in per-location healthcheck results into plain objects so message, name, and AWS SDK $metadata survive serialization.

Issue: [CLDSRV-933](https://scality.atlassian.net/browse/CLDSRV-933)

[CLDSRV-933]: https://scality.atlassian.net/browse/CLDSRV-933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/utilities/healthcheckHandler.js
+++ b/lib/utilities/healthcheckHandler.js
@@ -51,24 +51,6 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
         }
     });
     async.parallel(clientTasks, (err, results) => {
-        // Error's message/name/stack are non-enumerable, so JSON.stringify
-        // would drop them and log "error":{}. Copy them onto a plain object.
-        results.forEach(clientResult => {
-            Object.keys(clientResult).forEach(k => {
-                const e = clientResult[k] && clientResult[k].error;
-                if (e instanceof Error) {
-                    // eslint-disable-next-line no-param-reassign
-                    clientResult[k].error = {
-                        ...e,
-                        message: e.message,
-                        name: e.name,
-                        ...(e.code !== undefined && { code: e.code }),
-                        ...(e.statusCode !== undefined && { statusCode: e.statusCode }),
-                        ...(e.$metadata !== undefined && { $metadata: e.$metadata }),
-                    };
-                }
-            });
-        });
         // obj will be an object of the healthcheck results of
         // every backends. No errors were returned directly to
         // async.parallel in order to complete the check, so a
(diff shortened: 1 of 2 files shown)


lib/utilities/healthcheckHandler.js
```diff
@@ -56,6 +51,24 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
         }
     });
     async.parallel(clientTasks, (err, results) => {
+        // Error's message/name/stack are non-enumerable, so JSON.stringify
+        // would drop them and log "error":{}. Copy them onto a plain object.
+        results.forEach(clientResult => {
```
DarkIsDude: We can't do that in the following reduce ?

DarkIsDude (approved):

delthas (approved):

lib/utilities/healthcheckHandler.js
maeldonn: Good catch

maeldonn:
@scality/raving-robots Which version do you think I should target ?

lib/utilities/healthcheckHandler.js
```diff
@@ -65,21 +60,44 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
...
-        const { obj, fail } = results.reduce((acc, clientResult) => {
-            Object.assign(acc.obj, clientResult);
+        const { obj, fail } = results.reduce(
+            (acc, clientResult) => {
+                const keys = Object.keys(clientResult);
+                keys.forEach(k => {
+                    const e = clientResult[k] && clientResult[k].error;
+                    if (e instanceof Error) {
+                        // Error's message/name/stack are non-enumerable, so
+                        // JSON.stringify would drop them and log "error":{}.
```
francoisferrand, line 70: instead of fixing it manually (and just once) here, should we instead use a `replacer` in the call to `JSON.stringify`?

something like:
```js
const json = JSON.stringify(msg, (key, value) => {
	if (value instanceof Error) {
       return JSON.stringify(value, Object.getOwnPropertyNames(value));
    }
	return value;
});
```

francoisferrand: and/or update `ArsenalError` type to have a toJSON method?

lib/utilities/healthcheckHandler.js
```diff
@@ -62,10 +62,27 @@ function clientCheck(flightCheckOnStartUp, log, cb) {
...
+                        // Copy them onto a plain object.
+                        // eslint-disable-next-line no-param-reassign
+                        clientResult[k].error = {
+                            ...e,
+                            message: e.message,
+                            name: e.name,
+                            ...(e.code !== undefined && { code: e.code }),
+                            ...(e.statusCode !== undefined && { statusCode: e.statusCode }),
+                            ...(e.$metadata !== undefined && { $metadata: e.$metadata }),
```
francoisferrand, line 79: what kind of error do we actually get here?
If I remember correctly, we should always (or almost) receive ArsenalError's : mongo/aws/... errors should be wrapped already when we reach here, and thus the shape of errors should be standardized by now...

@benzekrimaha may remember more precisely about this?

francoisferrand (approved):

maeldonn:
Moved the fix to werelogs. I will open another PR to bump werelogs.

PR #6207, opened 2026-06-25
Bump werelogs to 8.2.4

maeldonn:
Bump werelogs to 8.2.4 to make error logging work everywhere in cloudserver.

Issue: CLDSRV-939

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
         "@azure/storage-blob": "^12.28.0",
         "@hapi/joi": "^17.1.1",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/Arsenal#8.4.11",
+        "arsenal": "git+https://github.com/scality/Arsenal#8.4.13",
         "async": "2.6.4",
         "bucketclient": "scality/bucketclient#8.2.7",
         "bufferutil": "^4.0.8",
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
         "utf8": "^3.0.0",
         "uuid": "^11.0.3",
         "vaultclient": "scality/vaultclient#8.5.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
         "tv4": "^1.3.0"
     },
     "resolutions": {
+        "werelogs": "scality/werelogs#8.2.4",
```
DarkIsDude: Why ?

DarkIsDude (approved):

package.json
maeldonn: Some of our dependencies also use werelogs. This resolution forces every one of them onto 8.2.4 so they all pick up the fix.

francoisferrand: - best to fix dependencies IMHO
- may use `^` (here and/or in deps) to ensure same version can be used where compatible
- yarn knows how to have different copies of dependency : not the role of the "top level" package to override resolution/dependency inside each package it depends on

maeldonn: Removed from resolutions and updated the only dependency with aws-sdk-js-v3

package.json
```diff
@@ -56,7 +56,7 @@
         "utf8": "^3.0.0",
         "uuid": "^11.0.3",
         "vaultclient": "scality/vaultclient#8.5.3",
-        "werelogs": "scality/werelogs#8.2.2",
+        "werelogs": "scality/werelogs#semver:^8.2.4",
```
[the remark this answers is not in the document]
maeldonn: Intentional, as @francoisferrand suggested: the `^` lets werelogs dedupe across our git deps without a `resolutions` override. Reproducibility is fine since we install with `--frozen-lockfile` and the commit is pinned in `yarn.lock`.

francoisferrand (approved):

delthas (approved):

PR #6211, opened 2026-06-29
Fix flaky server access log test by matching entries instead of exact count

delthas:
## Symptom

`testServerAccessLogFile.js` flakes with `AssertionError: Expected N log entries, got N+1` in `file-ft-tests` / `file-ft-tests-null-compat`, on a different operation each run.

## Root cause (captured in CI)

The tests asserted an exact access-log entry count per operation and isolated tests by truncating the **shared** access-log file. Instrumenting the assertion to dump the collected entries caught the extra one red-handed, in two independent failures:

```
objectDeleteTagging: expected 4, got 5
  [0] CreateBucket        test-server-access-log-bucket
  [1] PutObject           test-server-access-log-bucket/test-object-key
  [2] GetObject           buckettestgetobject/someObject   <-- FOREIGN (unique req_id, http 200)
  [3] PutObjectTagging    test-server-access-log-bucket/test-object-key
  [4] DeleteObjectTagging test-server-access-log-bucket/test-object-key
```

The extra entry is a `GetObject` on `buckettestgetobject` — a bucket owned by **`test/object/get.js`**, a different suite. It has a unique `req_id` (so it is **not** an SDK-retry duplicate, contrary to the ticket's original hypothesis) and a foreign bucket/key.

Access-log lines are written on the server's `res.on('close')`, which under HTTP keep-alive can fire long after the response (on socket teardown). A request from another suite can therefore have its line written *during* a serverAccessLogs test, landing inside the truncate window. This is outside the test's control, so no count-based barrier in the test can fix it.

## Fix

Stop asserting an exact count. Match each expected entry against the collected entries by its fields (the same field semantics as `validateLogEntry`) and ignore anything else. Unordered groups are flattened and matched independently, so interleaved foreign entries are tolerated regardless of when they arrive.

The first commit is an isolated prettier pass on the file (it predated prettier adoption); the second is the functional change.

Issue: CLDSRV-923

### DIFF
--- a/tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
+++ b/tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
@@ -23,35 +23,76 @@ function truncateLogFileIfExists(filePath) {
     }
 }
 
-async function waitForLogs(filePath, expectedLines, maxRetries, delayMs) {
-    for (let attempt = 0; attempt < maxRetries; attempt++) {
-        const logEntries = fs.readFileSync(filePath, 'utf8');
-        const lines = logEntries
-            .trim()
-            .split('\n')
-            .filter(line => line.length > 0);
-        if (lines.length >= expectedLines) {
-            try {
-                return lines.map(line => JSON.parse(line));
-            } catch (err) {
-                // FIXME(CLDSRV-800): readFileSync may read partial lines making JSON.parse fail, so we need to retry.
-                if (attempt == maxRetries) {
-                    throw new Error(`Failed to read log entries from ${filePath} after ${maxRetries} attempts: ${err}`);
-                }
+function readLogLines(filePath) {
+    return fs
+        .readFileSync(filePath, 'utf8')
+        .trim()
+        .split('\n')
+        .filter(line => line.length > 0);
+}
+
+// Whether a collected log entry satisfies an expected entry's constrained
+// fields (same semantics as validateLogEntry). Lets us ignore unrelated lines
+// from other suites sharing the server's access log file (CLDSRV-923).
+function entryMatchesExpected(entry, properties) {
+    for (const [key, val] of Object.entries(properties)) {
+        if (key === 'unordered') {
+            continue;
+        }
+        if (val === null) {
+            if (key in entry) {
+                return false;
             }
+        } else if (entry[key] !== val) {
+            return false;
+        }
+    }
+    return true;
+}
+
+async function waitForExpectedLogs(filePath, expectedEntries, maxRetries, delayMs) {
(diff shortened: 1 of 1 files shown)


tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -23,29 +23,61 @@ function truncateLogFileIfExists(filePath) {
...
-                return lines.map(line => JSON.parse(line));
-            } catch (err) {
-                // FIXME(CLDSRV-800): readFileSync may read partial lines making JSON.parse fail, so we need to retry.
-                if (attempt == maxRetries) {
-                    throw new Error(`Failed to read log entries from ${filePath} after ${maxRetries} attempts: ${err}`);
+        const content = fs.readFileSync(filePath, 'utf8');
+        const lines = content
+            .trim()
+            .split('\n')
+            .filter(line => line.length > 0);
```
tcarmet: I'll leave up to you to review if you think a little helper for those line would be useful since I see the same pattern below.

tcarmet (approved):
Thanks for having a look at it, much appreciated.

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
delthas: Good call — extracted a `readLogLines` helper (read + trim + split + filter) now used by both `waitForExpectedLogs` and `waitForAction`. Thanks!

maeldonn (approved):
Please remove `Claude-Session: xxx` from the commit description...

delthas:
> Please remove `Claude-Session: xxx` from the commit description...

Done, thanks!

tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
```diff
@@ -23,29 +23,61 @@ function truncateLogFileIfExists(filePath) {
...
+            if (allFound) {
+                return entries;
             }
+        } catch {
+            // FIXME(CLDSRV-800): readFileSync may read partial lines making JSON.parse fail, so we need to retry.
         }
         await sleep(delayMs);
     }
-    throw new Error(`Failed to read log entries from ${filePath} after ${maxRetries} attempts`);
+    throw new Error(`Did not find all expected log entries in ${filePath} after ${maxRetries} attempts`);
```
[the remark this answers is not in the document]
delthas: Good idea — on timeout, `waitForExpectedLogs` now recomputes the unmatched expected entries and reports them in the error (`Missing: <actions>`). Thanks!

benzekrimaha (approved):

PR #6218, opened 2026-07-06
Send X-Scal-Request-Uids (not x-scal-request-id) on the management config poll

delthas:
The management config-overlay poll (`lib/management/poll.js`,
`loadRemoteOverlay`) sent the werelogs `req_id` chain under the wrong
header name `x-scal-request-id`. Rename the outgoing header to
`x-scal-request-uids`, which matches cloudserver's own reader in
`lib/server.js`. The value (`log.getSerializedUids()`) is unchanged.

Pure rename: no dual-send. Also exports `loadRemoteOverlay` and adds a
focused unit test asserting the outgoing request carries
`x-scal-request-uids` and not `x-scal-request-id`.

Part of the OS-1089 reqUids-propagation audit. Note: the receiving
`config/overlay` endpoint on pensieve-api doesn't declare this header in
its swagger params, but pensieve reads it via a global middleware
(`WithRequestId`), so it applies to every endpoint regardless. Related
pensieve-side reqUids work: PSVAPI-131.

Issue: CLDSRV-948

### DIFF
--- a/lib/management/poll.js
+++ b/lib/management/poll.js
@@ -16,7 +16,7 @@ function loadRemoteOverlay(managementEndpoint, instanceId, remoteToken, cachedOv
     const opts = {
         headers: {
             'x-instance-authentication-token': remoteToken,
-            'x-scal-request-id': log.getSerializedUids(),
+            'x-scal-request-uids': log.getSerializedUids(),
         },
         json: true,
     };
@@ -143,4 +143,5 @@ function startPollingManagementClient(managementEndpoint, instanceId, remoteToke
 
 module.exports = {
     startPollingManagementClient,
+    loadRemoteOverlay,
 };
--- a/lib/management/poll.js
+++ b/lib/management/poll.js
@@ -5,18 +5,13 @@ const request = require('../utilities/request');
 const _config = require('../Config').config;
 const logger = require('../utilities/logger');
 const metadata = require('../metadata/wrapper');
-const {
-    loadCachedOverlay,
-    patchConfiguration,
-    saveConfigurationVersion,
-} = require('./configuration');
+const { loadCachedOverlay, patchConfiguration, saveConfigurationVersion } = require('./configuration');
 const { reshapeExceptionError } = arsenal.errorUtils;
 
 const pushReportDelay = 30000;
 const pullConfigurationOverlayDelay = 60000;
 
-function loadRemoteOverlay(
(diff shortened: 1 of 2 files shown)


SylvainSenechal (approved):

lib/management/poll.js
```diff
@@ -5,67 +5,65 @@ const request = require('../utilities/request');
...
-            log, wcb),
-        (cachedOverlay, wcb) =>
-            loadRemoteOverlay(managementEndpoint, instanceId, remoteToken,
-                cachedOverlay, log, wcb),
-        (cachedOverlay, remoteOverlay, wcb) =>
-            saveConfigurationVersion(cachedOverlay, remoteOverlay, log, wcb),
-        (remoteOverlay, wcb) => patchConfiguration(remoteOverlay,
-            log, wcb),
-    ], error => {
+    request.get(`${managementEndpoint}/${instanceId}/config/overlay`, opts, (error, response, body) => {
```
benzekrimaha, line 23: the poll hits `GET /instance/{uuid}/config/overlay` on pensieve-api, and AFAIK that specific endpoint does not declare X-Scal-Request-Uids in its swagger parameters while most other pensieve endpoints do.  upon checking I found https://scality.atlassian.net/browse/PSVAPI-131 might be useful to put it in the summary to avoid confusion

benzekrimaha (approved):
LGTM.
side note and not blocking : 
blobserver's `lib/management/poll.js` has the same x-scal-request-id header , we might need a follow up ticket

lib/management/poll.js
delthas: Confirmed it works regardless of the swagger params: pensieve reads `X-Scal-Request-Uids` in the global `WithRequestId` middleware (`gen/restapi/configure_pensieve.go:202`, wired through `setupGlobalMiddleware`), which wraps every endpoint — so `config/overlay` picks it up even without declaring the param. Before this rename cloudserver sent `x-scal-request-id`, so the middleware found nothing and minted a fresh id (no correlation); with the rename it continues cloudserver's chain. Added a note + PSVAPI-131 to the description.

Re blobserver: it looks EOL (last real commit Feb 2021), so I don't think a follow-up is worth it — shout if you disagree and I'll file one.

delthas:
Re blobserver: it looks EOL (last real commit Feb 2021), so I don't think a follow-up is worth it. Let me know and I can create one as needed.

PR #6219, opened 2026-07-06
Forward X-Scal-Request-Uids on Backbeat report/CRR-metrics requests

delthas:
Forward the werelogs request-uid chain (`log.getSerializedUids()`) on cloudserver's Backbeat report / CRR-metrics HTTP calls, via the `x-scal-request-uids` header.

Previously `traceparent` flowed on these calls but the reqUids chain did not. The header set here uses the same lowercase name cloudserver already reads on incoming admin requests (`lib/server.js`) and that arsenal's `RESTServer` reads, so req_id continuity is preserved across the Backbeat metric and state/schedule calls.

Scope: only the Backbeat metric/state calls that go through `reportHandler`'s `_makeRequest` (CRR/ingestion metrics, replication/ingestion states and schedules). The vault/metadata/data calls in reportHandler already forward reqUids through their arsenal clients and are unchanged.

Issue: CLDSRV-947

### DIFF
--- a/lib/utilities/reportHandler.js
+++ b/lib/utilities/reportHandler.js
@@ -151,28 +151,34 @@ function getSystemStats() {
     };
 }
 
-const _makeRequest = (endpoint, path, log, cb) => {
+async function _makeRequest(endpoint, path, log, cb) {
+    if (typeof cb === 'function') {
+        return _makeRequest(endpoint, path, log).then(
+            res => cb(null, res),
+            err => cb(err),
+        );
+    }
     const url = `${endpoint}${path}`;
-    request.get(
-        url,
-        {
-            json: true,
-            headers: { 'x-scal-request-uids': log.getSerializedUids() },
-        },
-        (error, response, body) => {
-            if (error) {
-                return cb(error);
-            }
-            if (response.statusCode >= 400) {
-                return cb('responseError', body);
-            }
-            if (body) {
-                return cb(null, body);
-            }
-            return cb(null, {});
-        },
-    );
-};
+    const body = await new Promise((resolve, reject) => {
+        request.get(
+            url,
+            {
+                json: true,
+                headers: { 'x-scal-request-uids': log.getSerializedUids() },
+            },
+            (error, _response, body) => {
+                if (error) {
+                    return reject(error);
+                }
+                return resolve(body);
+            },
+        );
+    });
+    if (body) {
+        return body;
+    }
+    return {};
+}
 
 function _crrMetricRequest(endpoint, site, log, cb) {
     const path = `${REQ_PATHS.crrMetricPrefix}/${site}`;
--- a/lib/utilities/reportHandler.js
+++ b/lib/utilities/reportHandler.js
@@ -151,25 +151,32 @@ function getSystemStats() {
     };
 }
 
-const _makeRequest = (endpoint, path, cb) => {
+const _makeRequest = (endpoint, path, log, cb) => {
     const url = `${endpoint}${path}`;
-    request.get(url, { json: true }, (error, response, body) => {
-        if (error) {
-            return cb(error);
-        }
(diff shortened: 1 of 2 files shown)


lib/utilities/reportHandler.js
```diff
@@ -143,25 +151,43 @@ function getSystemStats() {
     };
 }
 
-const _makeRequest = (endpoint, path, cb) => {
+async function _makeRequest(endpoint, path, log, cb) {
+    if (typeof cb === 'function') {
+        return _makeRequest(endpoint, path, log)
+            .then(res => cb(null, res))
```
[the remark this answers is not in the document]
delthas: Nack, documented way to migrate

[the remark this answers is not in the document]
DarkIsDude: @delthas maybe Claude is smarter than us. I found his suggestion pertinent. Can you check if we should or not consider it and update the migration doc (or follow-up) or agent-hub and confluence ?

lib/utilities/reportHandler.js
```diff
@@ -143,25 +151,43 @@ function getSystemStats() {
...
+            url,
+            {
+                json: true,
+                headers: { 'x-scal-request-uids': log.getSerializedUids() },
+            },
+            (error, response, body) => {
+                if (error) {
+                    return reject(error);
+                }
+                return resolve({ response, body });
```
DarkIsDude: Why do we need response ?

DarkIsDude: I mean with > 400, we'll have an error ?

SylvainSenechal: I legit can't wrap my head around this function lmao
sight

lib/utilities/reportHandler.js
```diff
@@ -151,25 +151,32 @@ function getSystemStats() {
     };
 }
 
-const _makeRequest = (endpoint, path, cb) => {
+const _makeRequest = (endpoint, path, log, cb) => {
```
SylvainSenechal: nit: maybe makeRequest should take reqUuid directly 🤔

lib/utilities/reportHandler.js
SylvainSenechal: I think its ok, looks like response is mostly because it was there before but yeah technically not used, probably ok to keep it, maybe _response ?

SylvainSenechal (approved):

lib/utilities/reportHandler.js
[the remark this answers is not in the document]
delthas: Good call — you were right to push on it. I've opened scality/agent-hub#62 to fix the prescribed trampoline: the two-arg `.then(onFulfilled, onRejected)` form (which is exactly what `util.callbackify` does internally) instead of `.then().catch()`, plus a skill version bump. The Confluence TAD is the remaining follow-up. Applied the two-arg form here too.

lib/utilities/reportHandler.js
delthas: Keeping `log` — threading `log` through call chains is the pervasive convention across cloudserver/arsenal, so `_makeRequest(…, log, cb)` stays consistent with the rest of the codebase.

lib/utilities/reportHandler.js
delthas: Good catch — it read as unused because the `statusCode >= 400` branch it fed is actually unreachable: `request.js` already converts 4xx/5xx into an `Error`, so the resolved path only ever sees <400. Removed that dead branch — the inner Promise resolves `body` directly, `response` is now genuinely unused (`_response`), and the error handling simplifies to `err => cb(err)`. Behavior is unchanged (a ≥400 still arrives as an `Error` → `cb(err)` → callers map it to `(null, {})`).

DarkIsDude (approved):
LGTM

PR #6220, opened 2026-07-06
Forward X-Scal-Request-Uids on Scuba quota requests

delthas:
Forward the werelogs req_id chain (`log.getSerializedUids()`) as the `X-Scal-Request-Uids` header on cloudserver's Scuba utilization requests, so Scuba can correlate quota lookups back to the originating cloudserver request. `traceparent` already flows; the req_id chain did not.

`ScubaClientImpl.getUtilizationMetrics` now takes a `log` parameter and, when the logger exposes `getSerializedUids()`, merges the header into the per-call options (arsenal `RESTClient` convention). A missing/mock logger never sends an undefined header. The quota (`_evaluateQuotas`) and Veeam capacity-metrics call sites pass `log`; the background health-check path is intentionally left untouched.

Issue: CLDSRV-946

### DIFF
--- a/lib/routes/veeam/utils.js
+++ b/lib/routes/veeam/utils.js
@@ -10,7 +10,6 @@ const UtilizationService = require('../../utilization/instance');
 const metadata = require('../../metadata/wrapper');
 
 const pipeline = promisify(streamPipeline);
-const getUtilizationMetrics = promisify((...args) => UtilizationService.getUtilizationMetrics(...args));
 const getBucket = promisify((...args) => metadata.getBucket(...args));
 
 /**
@@ -234,7 +233,7 @@ function getFileToBuild(request, data, inlineLastModified = false) {
 async function fetchCapacityMetrics(bucketMd, request, log) {
     const bucketKey = `${bucketMd._name}_${new Date(bucketMd._creationDate).getTime()}`;
     try {
-        return await getUtilizationMetrics('bucket', bucketKey, null, {}, log);
+        return await UtilizationService.getUtilizationMetrics('bucket', bucketKey, null, {}, log);
     } catch (err) {
         const statusCode = err.response?.status || err.statusCode || err.code;
         if (statusCode === 404) {
--- a/lib/routes/veeam/utils.js
+++ b/lib/routes/veeam/utils.js
@@ -37,12 +37,15 @@ async function receiveData(request, log) {
     // Prevent memory overloads by limiting the size of the
     // received data.
     if (parsedContentLength > ContentLengthThreshold) {
-        throw errorInstances.InvalidInput
-            .customizeDescription(`maximum allowed content-length is ${ContentLengthThreshold} bytes`);
+        throw errorInstances.InvalidInput.customizeDescription(
+            `maximum allowed content-length is ${ContentLengthThreshold} bytes`,
+        );
     }
     return await new Promise((resolve, reject) => {
         const settle = jsutil.once((err, result) => {
-            if (err) { return reject(err); }
+            if (err) {
+                return reject(err);
+            }
             return resolve(result);
         });
         let totalLength = 0;
@@ -51,8 +54,7 @@ async function receiveData(request, log) {
             write(chunk, _enc, cb) {
                 totalLength += chunk.length;
                 if (totalLength > parsedContentLength) {
-                    log.error('data stream exceed announced size',
-                        { parsedContentLength, overflow: totalLength });
+                    log.error('data stream exceed announced size', { parsedContentLength, overflow: totalLength });
                     return cb(errors.InternalError);
                 }
                 chunks.push(chunk);
@@ -88,17 +90,18 @@ function buildHeadXML(xmlContent) {
  * @returns {object} - response headers
  */
 function getResponseHeader(request, bucket, dataBuffer, lastModified, log) {
-    const corsHeaders = collectCorsHeaders(request.headers.origin,
-        request.method, bucket);
-    const responseMetaHeaders = collectResponseHeaders({
-        'last-modified': lastModified || new Date().toISOString(),
-        'content-md5': crypto
-            .createHash('md5')
-            .update(dataBuffer)
-            .digest('hex'),
-        'content-length': dataBuffer.byteLength,
-        'content-type': 'text/xml',
-    }, corsHeaders, null, false);
+    const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
+    const responseMetaHeaders = collectResponseHeaders(
+        {
+            'last-modified': lastModified || new Date().toISOString(),
+            'content-md5': crypto.createHash('md5').update(dataBuffer).digest('hex'),
+            'content-length': dataBuffer.byteLength,
+            'content-type': 'text/xml',
+        },
+        corsHeaders,
+        null,
+        false,
+    );
     responseMetaHeaders.versionId = 'null';
     responseMetaHeaders['x-amz-id-2'] = log.getSerializedUids();
     responseMetaHeaders['x-amz-request-id'] = log.getSerializedUids();
@@ -136,10 +139,10 @@ async function respondWithData(request, response, log, bucket, data, lastModifie
                 try {
                     response.setHeader(key, responseMetaHeaders[key]);
                 } catch (e) {
-                    log.debug('header can not be added ' +
-                        'to the response', {
+                    log.debug('header can not be added ' + 'to the response', {
                         header: responseMetaHeaders[key],
-                        error: e.stack, method: 'routeVeeam/respondWithData'
+                        error: e.stack,
+                        method: 'routeVeeam/respondWithData',
                     });
                 }
             }
@@ -199,7 +202,7 @@ function getFileToBuild(request, data, inlineLastModified = false) {
         return { error: errors.NoSuchKey };
     }
 
-    const modified = fileToBuild.LastModified || (new Date()).toISOString();
+    const modified = fileToBuild.LastModified || new Date().toISOString();
     const fieldName = _isSystemXML ? 'SystemInfo' : 'CapacityInfo';
 
     if (inlineLastModified) {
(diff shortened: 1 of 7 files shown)


lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +27,73 @@ class ScubaClientImpl extends ScubaClient {
...
-    getUtilizationMetrics(metricsClass, resourceName, options, body, callback) {
+    async getUtilizationMetrics(metricsClass, resourceName, options, body, log, callback) {
+        // Dual callback+async trampoline: legacy callers pass a callback, new
+        // callers await the returned promise. The callback is invoked from the
+        // promise chain (never inside try/catch) so it is not re-invoked if it
+        // throws.
+        if (typeof callback === 'function') {
+            return this.getUtilizationMetrics(metricsClass, resourceName, options, body, log)
+                .then(data => callback(null, data))
+                .catch(callback);
```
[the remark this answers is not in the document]
delthas: Intentional documented migration

lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +27,69 @@ class ScubaClientImpl extends ScubaClient {
...
+        }
         const requestStartTime = process.hrtime.bigint();
-        return this._getLatestMetricsCallback(metricsClass, resourceName, options, body, (err, data) => {
+        const reqUids = log && typeof log.getSerializedUids === 'function' ? log.getSerializedUids() : undefined;
+        const mergedOptions = reqUids
+            ? { ...options, headers: { ...(options && options.headers), 'X-Scal-Request-Uids': reqUids } }
+            : options;
+        const getLatestMetrics = util.promisify((...args) => this._getLatestMetricsCallback(...args));
```
benzekrimaha: `_getLatestMetricsCallback` is util.callbackify(this.getLatestMetrics) (constructor), and this line wraps it back with util.promisify , so this is promisify(callbackify(fn)), i.e. fn. The SDK method is already async; now that the wrapper is async too, call it directly:

```suggestion
        return await super.getLatestMetrics(metricsClass, resourceName, mergedOptions, body);
```

lib/routes/veeam/utils.js
benzekrimaha:
```suggestion
return await UtilizationService.getUtilizationMetrics('bucket', bucketKey, null, {}, log);
```

lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +27,69 @@ class ScubaClientImpl extends ScubaClient {
...
+                this._healthCheck();
+            },
+            Number(process.env.SCUBA_HEALTHCHECK_FREQUENCY) || externalBackendHealthCheckInterval,
+        );
     }
 
-    getUtilizationMetrics(metricsClass, resourceName, options, body, callback) {
+    async getUtilizationMetrics(metricsClass, resourceName, options, body, log, callback) {
+        if (typeof callback === 'function') {
+            return this.getUtilizationMetrics(metricsClass, resourceName, options, body, log)
```
benzekrimaha: if callback(null, data) throws, the .catch(callback) re-invokes the callback with the thrown error, the classic double-callback. We should use the two-argument `.then` so the fulfilled-handler's throw is not swallowed into a second call
```suggestion
           return this.getUtilizationMetrics(metricsClass, resourceName, options, body, log)
    .then(data => callback(null, data), err => callback(err));
```

lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +27,69 @@ class ScubaClientImpl extends ScubaClient {
...
+                .then(data => callback(null, data))
+                .catch(callback);
+        }
         const requestStartTime = process.hrtime.bigint();
-        return this._getLatestMetricsCallback(metricsClass, resourceName, options, body, (err, data) => {
+        const reqUids = log && typeof log.getSerializedUids === 'function' ? log.getSerializedUids() : undefined;
```
benzekrimaha:
```suggestion
        const reqUids = log?.getSerializedUids?.() ? log.getSerializedUids() : undefined;
```

lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +27,69 @@ class ScubaClientImpl extends ScubaClient {
...
+        const mergedOptions = reqUids
+            ? { ...options, headers: { ...(options && options.headers), 'X-Scal-Request-Uids': reqUids } }
```
benzekrimaha: nit:
```suggestion
            ? { ...options, headers: { ...options?.headers, 'X-Scal-Request-Uids': reqUids } }
```

lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +27,69 @@ class ScubaClientImpl extends ScubaClient {
...
         }
         this._healthCheck();
-        this._healthCheckTimer = setInterval(async () => {
-            this._healthCheck();
-        }, Number(process.env.SCUBA_HEALTHCHECK_FREQUENCY)
-            || externalBackendHealthCheckInterval);
+        this._healthCheckTimer = setInterval(
+            async () => {
+                this._healthCheck();
+            },
```
benzekrimaha: the async adds nothing and the promise isn't awaited here
```suggestion
            () => this._healthCheck()
```

lib/utilization/scuba/wrapper.js
maeldonn:
```suggestion
const reqUids = log?.getSerializedUids?.();
```

tests/unit/quotas/scuba/wrapper.js
maeldonn, line 1: Could you use async/await instead of callbacks ?

lib/utilization/scuba/wrapper.js
delthas: Applied the two-arg `.then(data => callback(null, data), err => callback(err))` form. Also opened scality/agent-hub#62 to fix the prescribed trampoline in the migration guide/skill so future migrations use it.

lib/utilization/scuba/wrapper.js
delthas: Done — `return await super.getLatestMetrics(metricsClass, resourceName, mergedOptions, body)` directly. Removed the now-dead `_getLatestMetricsCallback` (`util.callbackify`) and the unused `util` import; this also cleared the DEP0174 warning.

lib/utilization/scuba/wrapper.js
delthas: Done — `const reqUids = log?.getSerializedUids?.();` (thanks @maeldonn).

lib/utilization/scuba/wrapper.js
delthas: Done — `{ ...options, headers: { ...options?.headers, 'X-Scal-Request-Uids': reqUids } }`.

lib/utilization/scuba/wrapper.js
delthas: Done — dropped the no-op `async`, now `() => this._healthCheck()`.

lib/routes/veeam/utils.js
delthas: Done — removed the redundant `promisify` wrapper; `fetchCapacityMetrics` now awaits the async `getUtilizationMetrics` directly. This also cleared the DEP0174 warning.

tests/unit/quotas/scuba/wrapper.js
delthas: Done — rewrote the new scuba wrapper tests to async/await (stubbing `ScubaClient.prototype.getLatestMetrics` with `.resolves`). The `quotaUtils` and veeam test stubs were migrated from callbacks to promises accordingly.

maeldonn: Should promisify `getUtilizationMetrics`

maeldonn (approved):

tests/unit/quotas/scuba/wrapper.js
delthas: Already async/awaited, nothing to do(?). Closing.

lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +25,67 @@ class ScubaClientImpl extends ScubaClient {
...
-        }, Number(process.env.SCUBA_HEALTHCHECK_FREQUENCY)
-            || externalBackendHealthCheckInterval);
+        this._healthCheckTimer = setInterval(
+            () => this._healthCheck(),
+            Number(process.env.SCUBA_HEALTHCHECK_FREQUENCY) || externalBackendHealthCheckInterval,
+        );
     }
 
-    getUtilizationMetrics(metricsClass, resourceName, options, body, callback) {
+    async getUtilizationMetrics(metricsClass, resourceName, options, body, log, callback) {
```
benzekrimaha: why keep both callback and promise interfaces here? This leaves a .then() inside an async method, CodeQL still flags it, and the positional API changes from callback as the last argument to log instead. can we make getUtilizationMetrics promise-only and migrate/adapt the remaining callback callers at their boundary?

delthas: Done in d083ab7 — `getUtilizationMetrics` is now promise-only: dropped the callback branch and the trailing `callback` param, and migrated the only remaining callback caller (the `bucketQuota`/`accountQuota` tasks in `_evaluateQuotas`' `async.parallel`) to `await` it directly. `async.parallel` accepts async-function tasks, so there's no `.then()` glue in the caller either. veeam already awaited it.

That removes the `.then()` inside the async method and the callback-as-last-arg signature you flagged. Behaviour is unchanged — the quota tests stub `getLatestMetrics` directly, so they pass as-is (75 green).

lib/utilization/scuba/wrapper.js
```diff
@@ -27,51 +25,61 @@ class ScubaClientImpl extends ScubaClient {
     }
 
     _healthCheck() {
-        return this.healthCheck().then(data => {
-            if (data?.date) {
-                const date = new Date(data.date);
-                if (Date.now() - date.getTime() > this.maxStaleness) {
-                    throw new Error('Data is stale, disabling quotas');
+        return this.healthCheck()
```
benzekrimaha: you migrated `getUtilizationMetrics` to async/await for CodeQL, but left .then()/.catch() here in the same file (CodeQL still open). no point stopping halfway IMO

benzekrimaha (approved):
LGTM , but I'd strongly advice converting `_healthCheck` to async/await too (CodeQL `.then()` still open)

lib/utilization/scuba/wrapper.js
delthas: Fair — done in 488ba0d. Migrated `_healthCheck` to async/await (try/catch), so the Scuba wrapper is now fully `.then`-free (clears the CodeQL `.then()` flag at wrapper.js:41 and the two eslint `prefer-await-to-then` warnings). Behaviour unchanged; the `_healthCheck` tests already await it (75 green).

Left the `quotaUtils` finalizer `async.waterfall` (the remaining CodeQL cb/done) alone for now — clearing those means migrating `_evaluateQuotas` (callback→promise) and its callers, which is a broader async migration than this reqUids PR should carry. Happy to do it in a dedicated follow-up if you'd prefer.

PR #6221, opened 2026-07-06
Seed zero-value quota metric on bucket create / quota enable

delthas:
## Context

Bucket/account quota enforcement reads utilization metrics from Scuba, which serves the `__infostore` collection. Those documents are written today only by the s3utils `count-items` cron (every 6–24h). A freshly created bucket — or a bucket that has a quota enabled while empty — has no metric document until the next cron run, so `validateQuotas` → `_evaluateQuotas` gets `NoSuchEntity` from Scuba and **fails open**: the quota is not enforced for up to 24h (ARTESCA-17063).

## Change

Seed a zero-value bucket metric document in `__infostore` at the moments CloudServer knows the bucket is empty, so the quota is enforceable immediately:

1. **On `createBucket`** — the bucket is empty by definition, so seed unconditionally.
2. **On `bucketUpdateQuota`** — seed only when the bucket is **verifiably empty** (a `maxKeys:1` `DelimiterVersions` listing). Never default to zero for a bucket that may hold uncounted data (that would silently under-enforce).

Both paths:
- are gated on `config.isQuotaEnabled()`;
- are **best-effort / non-fatal** — a seeding failure is logged and never fails the request (quota simply stays disabled until count-items, i.e. today's behavior);
- use the new arsenal `MetadataWrapper.initializeBucketCapacity` (idempotent `$setOnInsert`, so an existing count-items document is never overwritten), keyed by the bucket's **metastore** `creationDate` — exactly what `_evaluateQuotas` queries with, so the create-window lookup is an exact match.

Account-level metrics are intentionally out of scope: a brand-new bucket's account may hold data in other, not-yet-counted buckets, so seeding `account_<id>=0` would under-report. Empty-account coverage stays with count-items (S3UTILS-224).

## Dependency

Requires **ARSN-610** ([arsenal#2666](https://github.com/scality/arsenal/pull/2666)). The first commit here temporarily pins arsenal to that branch commit so the new method resolves; it should be replaced with a released arsenal 8.5 tag once ARSN-610 merges.

## Tests

New unit tests: `bucketUpdateQuota` seeds on an empty bucket, does **not** seed on a non-empty bucket, still succeeds when seeding errors, and does not seed when quotas are disabled; `bucketPut` seeds on creation, stays non-fatal on seed error, and does not seed when quotas are disabled.

Issue: CLDSRV-949

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const async = require('async');
+const { promisify } = require('util');
 const { errors } = require('arsenal');
 
 const acl = require('../../../metadata/acl');
@@ -85,38 +86,36 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
     return metadata.updateBucket(bucketName, bucket, log, callback);
 }
 
-function seedBucketQuotaCapacity(bucket, log, done) {
+const createBucketMD = promisify((...args) => metadata.createBucket(...args));
+const initializeBucketCapacity = promisify((...args) => metadata.initializeBucketCapacity(...args));
+const addToUsersBucketAsync = promisify(addToUsersBucket);
+const removeTransientOrDeletedLabelAsync = promisify(removeTransientOrDeletedLabel);
+
+async function seedBucketQuotaCapacity(bucket, log) {
     if (!config.isQuotaEnabled()) {
-        return done();
+        return;
     }
-    return metadata.initializeBucketCapacity(bucket.getName(), bucket.getCreationDate(), log, err => {
-        if (err) {
-            log.error('error seeding bucket quota capacity metric', { error: err });
-        }
-        return done();
-    });
+    try {
+        await initializeBucketCapacity(bucket.getName(), bucket.getCreationDate(), log);
+    } catch (err) {
+        // Best-effort: a missing metric self-heals at the next count-items run,
+        // so a seeding failure must never fail the bucket operation.
+        log.warn('error seeding bucket quota capacity metric', { error: err });
+    }
 }
 
-function freshStartCreateBucket(bucket, canonicalID, log, callback) {
+async function freshStartCreateBucket(bucket, canonicalID, log, callback) {
+    if (callback) {
+        return freshStartCreateBucket(bucket, canonicalID, log).then(() => callback(null), callback);
+    }
     const bucketName = bucket.getName();
-    metadata.createBucket(bucketName, bucket, log, err => {
-        if (err) {
-            log.debug('error from metadata', { error: err });
-            return callback(err);
-        }
-        log.trace('created bucket in metadata');
-        return addToUsersBucket(canonicalID, bucketName, bucket, log, err => {
-            if (err) {
-                return callback(err);
-            }
-            return removeTransientOrDeletedLabel(bucket, log, labelErr => {
-                if (labelErr) {
-                    return callback(labelErr);
-                }
-                return seedBucketQuotaCapacity(bucket, log, () => callback(null));
-            });
-        });
-    });
+    await createBucketMD(bucketName, bucket, log);
+    log.trace('created bucket in metadata');
+    await addToUsersBucketAsync(canonicalID, bucketName, bucket, log);
+    // Seed before clearing the transient/deleted flag so the bucket is never usable without its metric doc.
+    await seedBucketQuotaCapacity(bucket, log);
+    await removeTransientOrDeletedLabelAsync(bucket, log);
+    return undefined;
 }
 
 /**
@@ -131,19 +130,16 @@ function freshStartCreateBucket(bucket, canonicalID, log, callback) {
  * @param {function} callback - callback with error or null as arguments
  * @return {undefined}
  */
-function cleanUpBucket(bucketMD, canonicalID, log, callback) {
+async function cleanUpBucket(bucketMD, canonicalID, log, callback) {
+    if (callback) {
+        return cleanUpBucket(bucketMD, canonicalID, log).then(() => callback(null), callback);
+    }
     const bucketName = bucketMD.getName();
-    return addToUsersBucket(canonicalID, bucketName, bucketMD, log, err => {
-        if (err) {
-            return callback(err);
-        }
-        return removeTransientOrDeletedLabel(bucketMD, log, labelErr => {
-            if (labelErr) {
-                return callback(labelErr);
-            }
-            return seedBucketQuotaCapacity(bucketMD, log, () => callback(null));
-        });
-    });
+    await addToUsersBucketAsync(canonicalID, bucketName, bucketMD, log);
+    // Seed before clearing the transient/deleted flag so the bucket is never usable without its metric doc.
+    await seedBucketQuotaCapacity(bucketMD, log);
+    await removeTransientOrDeletedLabelAsync(bucketMD, log);
+    return undefined;
 }
 
 /**
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -85,6 +85,18 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
     return metadata.updateBucket(bucketName, bucket, log, callback);
 }
 
+function seedBucketQuotaCapacity(bucket, log, done) {
+    if (!config.isQuotaEnabled()) {
+        return done();
+    }
+    return metadata.initializeBucketCapacity(bucket.getName(), bucket.getCreationDate(), log, err => {
+        if (err) {
+            log.error('error seeding bucket quota capacity metric', { error: err });
+        }
+        return done();
+    });
+}
+
 function freshStartCreateBucket(bucket, canonicalID, log, callback) {
     const bucketName = bucket.getName();
     metadata.createBucket(bucketName, bucket, log, err => {
@@ -97,7 +109,12 @@ function freshStartCreateBucket(bucket, canonicalID, log, callback) {
             if (err) {
                 return callback(err);
             }
-            return removeTransientOrDeletedLabel(bucket, log, callback);
+            return removeTransientOrDeletedLabel(bucket, log, labelErr => {
+                if (labelErr) {
+                    return callback(labelErr);
+                }
+                return seedBucketQuotaCapacity(bucket, log, () => callback(null));
+            });
         });
     });
 }
@@ -120,7 +137,12 @@ function cleanUpBucket(bucketMD, canonicalID, log, callback) {
         if (err) {
             return callback(err);
         }
-        return removeTransientOrDeletedLabel(bucketMD, log, callback);
+        return removeTransientOrDeletedLabel(bucketMD, log, labelErr => {
+            if (labelErr) {
+                return callback(labelErr);
+            }
+            return seedBucketQuotaCapacity(bucketMD, log, () => callback(null));
+        });
     });
 }
 
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -18,7 +18,6 @@ const oldUsersBucket = constants.oldUsersBucket;
 const zenkoSeparator = constants.zenkoSeparator;
 const userBucketOwner = 'admin';
 
-
 function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
     // BACKWARD: Simplify once do not have to deal with old
     // usersbucket name and old splitter
@@ -28,8 +27,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
         if (err && !err.is.NoSuchBucket && !err.is.BucketAlreadyExists) {
             return cb(err);
         }
-        const splitter = usersBucketAttrs ?
-            constants.splitter : constants.oldSplitter;
+        const splitter = usersBucketAttrs ? constants.splitter : constants.oldSplitter;
         let key = createKeyForUserBucket(canonicalID, splitter, bucketName);
         const omVal = {
             creationDate: new Date().toJSON(),
@@ -38,46 +36,44 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
         // If the new format usersbucket does not exist, try to put the
         // key in the old usersBucket using the old splitter.
         // Otherwise put the key in the new format usersBucket
-        const usersBucketBeingCalled = usersBucketAttrs ?
-            usersBucket : oldUsersBucket;
-        return metadata.putObjectMD(usersBucketBeingCalled, key,
-            omVal, {}, log, err => {
-                if (err?.is?.NoSuchBucket) {
-                    // There must be no usersBucket so createBucket
-                    // one using the new format
-                    log.trace('users bucket does not exist, ' +
-                        'creating users bucket');
-                    key = `${canonicalID}${constants.splitter}` +
-                        `${bucketName}`;
-                    const creationDate = new Date().toJSON();
-                    const freshBucket = new BucketInfo(usersBucket,
-                        userBucketOwner, userBucketOwner, creationDate,
-                        BucketInfo.currentModelVersion());
-                    return metadata.createBucket(usersBucket,
-                        freshBucket, log, err => {
-                            // Note: In the event that two
-                            // users' requests try to create the
-                            // usersBucket at the same time,
-                            // this will prevent one of the users
-                            // from getting a BucketAlreadyExists
-                            // error with respect
-                            // to the usersBucket.
-                            // TODO: move to `.is` once BKTCLT-9 is done and bumped in Cloudserver
-                            if (err && !err.BucketAlreadyExists) {
-                                log.error('error from metadata', {
-                                    error: err,
-                                });
-                                return cb(err);
-                            }
-                            log.trace('Users bucket created');
-                            // Finally put the key in the new format
-                            // usersBucket
-                            return metadata.putObjectMD(usersBucket,
-                                key, omVal, {}, log, cb);
+        const usersBucketBeingCalled = usersBucketAttrs ? usersBucket : oldUsersBucket;
+        return metadata.putObjectMD(usersBucketBeingCalled, key, omVal, {}, log, err => {
+            if (err?.is?.NoSuchBucket) {
+                // There must be no usersBucket so createBucket
+                // one using the new format
+                log.trace('users bucket does not exist, ' + 'creating users bucket');
+                key = `${canonicalID}${constants.splitter}` + `${bucketName}`;
+                const creationDate = new Date().toJSON();
+                const freshBucket = new BucketInfo(
+                    usersBucket,
+                    userBucketOwner,
+                    userBucketOwner,
+                    creationDate,
+                    BucketInfo.currentModelVersion(),
+                );
+                return metadata.createBucket(usersBucket, freshBucket, log, err => {
+                    // Note: In the event that two
+                    // users' requests try to create the
+                    // usersBucket at the same time,
+                    // this will prevent one of the users
+                    // from getting a BucketAlreadyExists
+                    // error with respect
+                    // to the usersBucket.
+                    // TODO: move to `.is` once BKTCLT-9 is done and bumped in Cloudserver
+                    if (err && !err.BucketAlreadyExists) {
+                        log.error('error from metadata', {
+                            error: err,
                         });
-                }
-                return cb(err);
-            });
+                        return cb(err);
+                    }
+                    log.trace('Users bucket created');
+                    // Finally put the key in the new format
+                    // usersBucket
+                    return metadata.putObjectMD(usersBucket, key, omVal, {}, log, cb);
+                });
+            }
+            return cb(err);
+        });
     });
 }
 
@@ -139,16 +135,15 @@ function cleanUpBucket(bucketMD, canonicalID, log, callback) {
  * @callback called with (err, sseInfo: object)
  */
 function bucketLevelServerSideEncryption(bucket, headers, log, cb) {
-    kms.bucketLevelEncryption(
-        bucket, headers, log, (err, sseInfo) => {
-            if (err) {
-                log.debug('error getting bucket encryption info', {
-                    error: err,
-                });
-                return cb(err);
-            }
-            return cb(null, sseInfo);
-        });
+    kms.bucketLevelEncryption(bucket, headers, log, (err, sseInfo) => {
+        if (err) {
+            log.debug('error getting bucket encryption info', {
+                error: err,
+            });
+            return cb(err);
+        }
+        return cb(null, sseInfo);
+    });
 }
 
 /**
@@ -163,28 +158,44 @@ function bucketLevelServerSideEncryption(bucket, headers, log, cb) {
  * @param {function} cb - callback to bucketPut
  * @return {undefined}
  */
-function createBucket(authInfo, bucketName, headers,
-    locationConstraint, log, cb) {
+function createBucket(authInfo, bucketName, headers, locationConstraint, log, cb) {
     // Prefer using undefined instead of null for unused properties so they are not present in metadata payload
     log.trace('Creating bucket');
     assert.strictEqual(typeof bucketName, 'string');
     const canonicalID = authInfo.getCanonicalID();
-    const ownerDisplayName =
-        authInfo.getAccountDisplayName();
+    const ownerDisplayName = authInfo.getAccountDisplayName();
     const creationDate = new Date().toJSON();
     const isNFSEnabled = headers['x-scal-nfs-enabled'] === 'true' || undefined;
     const headerObjectLock = headers['x-amz-bucket-object-lock-enabled'];
-    const objectLockEnabled
-        = headerObjectLock && headerObjectLock.toLowerCase() === 'true';
-    const bucket = new BucketInfo(bucketName, canonicalID, ownerDisplayName,
-        creationDate, BucketInfo.currentModelVersion(), undefined, undefined, undefined, undefined,
-        undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, isNFSEnabled,
-        undefined, undefined, objectLockEnabled);
+    const objectLockEnabled = headerObjectLock && headerObjectLock.toLowerCase() === 'true';
+    const bucket = new BucketInfo(
+        bucketName,
+        canonicalID,
+        ownerDisplayName,
+        creationDate,
+        BucketInfo.currentModelVersion(),
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        isNFSEnabled,
+        undefined,
+        undefined,
+        objectLockEnabled,
+    );
     let locationConstraintVal = null;
 
     if (locationConstraint) {
-        const [locationConstraintStr, ingestion] =
-            locationConstraint.split(zenkoSeparator);
+        const [locationConstraintStr, ingestion] = locationConstraint.split(zenkoSeparator);
         if (locationConstraintStr) {
             locationConstraintVal = locationConstraintStr;
             bucket.setLocationConstraint(locationConstraintStr);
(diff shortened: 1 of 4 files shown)


lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,121 @@
...
+    } catch (err) {
+        log.error('error seeding bucket quota capacity metric', { error: err });
+    }
+}
 
+async function bucketUpdateQuota(authInfo, request, log, callback) {
+    if (callback) {
+        return bucketUpdateQuota(authInfo, request, log)
+            .then(corsHeaders => callback(null, corsHeaders))
+            .catch(err => callback(err, err.code, err.additionalResHeaders));
```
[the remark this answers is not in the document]
delthas: No, this trampoline is the documented async/await migration

package.json
```diff
@@ -37,7 +37,7 @@
         "@opentelemetry/instrumentation-ioredis": "~0.64.0",
         "@opentelemetry/instrumentation-mongodb": "~0.69.0",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/arsenal#8.5.4",
+        "arsenal": "git+https://github.com/scality/arsenal#05db0fa2268958fac066204ad2a48d2e083311dd",
```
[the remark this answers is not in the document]
delthas: I know

package.json
```diff
@@ -37,7 +37,7 @@
         "@opentelemetry/instrumentation-ioredis": "~0.64.0",
         "@opentelemetry/instrumentation-mongodb": "~0.69.0",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/arsenal#8.5.6",
+        "arsenal": "git+https://github.com/scality/arsenal#43a1d0577dffd6e77a712ab813abaff57ace43ae",
```
DarkIsDude: Be careful 🙏

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -89,6 +85,18 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
     return metadata.updateBucket(bucketName, bucket, log, callback);
 }
 
+function seedBucketQuotaCapacity(bucket, log, done) {
```
DarkIsDude: This function is a new one and must use async/await. We agreed to don't do it for prettier one, but this one is a new one 🙏

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -101,7 +109,12 @@ function freshStartCreateBucket(bucket, canonicalID, log, callback) {
             if (err) {
                 return callback(err);
             }
-            return removeTransientOrDeletedLabel(bucket, log, callback);
+            return removeTransientOrDeletedLabel(bucket, log, labelErr => {
```
DarkIsDude: Same here, you touch the function, it's not only a formatting 🙏

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -89,6 +85,18 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
     return metadata.updateBucket(bucketName, bucket, log, callback);
 }
 
+function seedBucketQuotaCapacity(bucket, log, done) {
+    if (!config.isQuotaEnabled()) {
+        return done();
+    }
+    return metadata.initializeBucketCapacity(bucket.getName(), bucket.getCreationDate(), log, err => {
+        if (err) {
+            log.error('error seeding bucket quota capacity metric', { error: err });
```
DarkIsDude: Why we just log and don't propagate it ?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -101,7 +109,12 @@ function freshStartCreateBucket(bucket, canonicalID, log, callback) {
             if (err) {
                 return callback(err);
             }
-            return removeTransientOrDeletedLabel(bucket, log, callback);
+            return removeTransientOrDeletedLabel(bucket, log, labelErr => {
+                if (labelErr) {
+                    return callback(labelErr);
+                }
+                return seedBucketQuotaCapacity(bucket, log, () => callback(null));
```
DarkIsDude: Why `callback(null)` ? Why not `callback()` ?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -163,28 +180,44 @@ function bucketLevelServerSideEncryption(bucket, headers, log, cb) {
...
     const isNFSEnabled = headers['x-scal-nfs-enabled'] === 'true' || undefined;
     const headerObjectLock = headers['x-amz-bucket-object-lock-enabled'];
-    const objectLockEnabled
-        = headerObjectLock && headerObjectLock.toLowerCase() === 'true';
-    const bucket = new BucketInfo(bucketName, canonicalID, ownerDisplayName,
-        creationDate, BucketInfo.currentModelVersion(), undefined, undefined, undefined, undefined,
-        undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, isNFSEnabled,
-        undefined, undefined, objectLockEnabled);
+    const objectLockEnabled = headerObjectLock && headerObjectLock.toLowerCase() === 'true';
+    const bucket = new BucketInfo(
```
DarkIsDude, line 189: 😂 this is where you know that a class is too large 😂

DarkIsDude: (nothing to do)

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,114 @@
...
 
-function bucketUpdateQuota(authInfo, request, log, callback) {
-    log.debug('processing request', { method: 'bucketUpdateQuota' });
+async function seedEmptyBucketCapacity(bucket, log) {
+    if (!config.isQuotaEnabled()) {
+        return;
+    }
+    const bucketName = bucket.getName();
+    try {
+        const list = await promisify(metadata.listObject.bind(metadata))(
```
DarkIsDude: To avoid promisify always, you can do `promise.custom`. See the confluence page.  does it make sense here ?

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,114 @@
...
+            log,
+        );
+        const isEmpty =
+            (list.Versions ? list.Versions.length : 0) + (list.DeleteMarkers ? list.DeleteMarkers.length : 0) === 0;
+        if (!isEmpty) {
+            return;
+        }
+        await promisify(metadata.initializeBucketCapacity.bind(metadata))(bucketName, bucket.getCreationDate(), log);
+    } catch (err) {
+        log.error('error seeding bucket quota capacity metric', { error: err });
```
DarkIsDude: Why just a log and not a propagate ?

DarkIsDude (changes requested):

package.json
delthas: Yes, known and flagged by Claude already, keeping open until the Arsenal part is merged

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,114 @@
...
-            bucket: bucketName,
+
+    let bucket;
+    try {
+        bucket = await promisify(standardMetadataValidateBucket)(metadataValParams, request.actionImplicitDenies, log);
+        const requestBody = await parseRequestBody(request.post, request.headers['content-type']);
+        const quotaValue = validateBucketQuotaProperty(requestBody);
+        bucket.setQuota(quotaValue);
+        await promisify(metadata.updateBucket.bind(metadata))(bucket.getName(), bucket, log);
+        await seedEmptyBucketCapacity(bucket, log);
```
benzekrimaha, line 116: why seed after updateBucket persists the quota? Done in this order there's still a window where the quota is set but no metric exists (fail-open, the exact bug we're fixing here). Listing + seeding first, then enabling the quota, closes the window entirely IMO and the emptiness check can't be invalidated by the quota becoming visible in between

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,114 @@
...
+    try {
+        const list = await promisify(metadata.listObject.bind(metadata))(
+            bucketName,
+            {
+                maxKeys: 1,
+                listingType: 'DelimiterVersions',
+            },
+            log,
+        );
+        const isEmpty =
```
benzekrimaha: a `DelimiterVersions` listing won't see in-progress MPU parts (shadow bucket), but those hold real storage. If count-items includes MPU parts in the bucket metric, seeding 0 here under-enforces for a bucket that only has uncommitted uploads , the exact case the description says must never happen ("never default to zero for a bucket that may hold uncounted data"). Does count-items count MPU parts? and if yes, shouldn't the emptiness check also probe the MPU shadow bucket?
```suggestion
        const isEmpty = (list.Versions?.length ?? 0) + (list.DeleteMarkers?.length ?? 0) === 0;
```

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -124,7 +137,12 @@ function cleanUpBucket(bucketMD, canonicalID, log, callback) {
         if (err) {
             return callback(err);
         }
-        return removeTransientOrDeletedLabel(bucketMD, log, callback);
+        return removeTransientOrDeletedLabel(bucketMD, log, labelErr => {
+            if (labelErr) {
+                return callback(labelErr);
+            }
+            return seedBucketQuotaCapacity(bucketMD, log, () => callback(null));
```
benzekrimaha: seeding also fires on the transient/deleted-flag cleanup path (re-creating a bucket left in transient state), which the PR description doesn't mention and no test covers, codecov flags 5 missed lines in this file. Is the creationDate used here guaranteed to be the new incarnation's metastore date that `_evaluateQuotas` will query? can we add a test for this path (or explain why it can't be hit with quotas enabled).

tests/unit/api/bucketPut.js
```diff
@@ -934,3 +904,50 @@ describe('bucketPut API with SSE Configurations', () => {
...
+
+    it('should seed a zero-value quota metric on bucket creation when quotas are enabled', done => {
+        sinon.stub(config, 'isQuotaEnabled').returns(true);
+        const initStub = sinon
+            .stub(metadata, 'initializeBucketCapacity')
+            .callsFake((name, creationDate, l, cb) => cb(null));
+        bucketPut(authInfo, testRequest, log, err => {
+            assert.ifError(err);
+            assert(initStub.calledOnce, 'expected initializeBucketCapacity to be called once');
+            assert.strictEqual(initStub.firstCall.args[0], bucketName);
```
benzekrimaha, line 922: the exact-match creationDate keying is the load-bearing part of the design; the test should also pin it:
```suggestion
            assert.strictEqual(initStub.firstCall.args[0], bucketName);
            assert.strictEqual(initStub.firstCall.args[1], md.getCreationDate());
```

tests/unit/api/bucketUpdateQuota.js
```diff
@@ -0,0 +1,122 @@
...
+    });
+
+    it('should still update the quota if seeding the metric fails', done => {
+        bucketPut(authInfo, bucketPutRequest, log, err => {
+            assert.ifError(err);
+            sinon
+                .stub(metadata, 'initializeBucketCapacity')
+                .callsFake((name, creationDate, l, cb) => process.nextTick(() => cb(errors.InternalError)));
+            return bucketUpdateQuota(authInfo, updateQuotaRequest, log, err => {
+                assert.ifError(err);
```
benzekrimaha, line 108: the test only asserts no error we should `assert md.getQuota() === 1000` after the call, otherwise "still update the quota" isn't actually verified.

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,114 @@
...
+            (list.Versions ? list.Versions.length : 0) + (list.DeleteMarkers ? list.DeleteMarkers.length : 0) === 0;
+        if (!isEmpty) {
+            return;
+        }
+        await promisify(metadata.initializeBucketCapacity.bind(metadata))(bucketName, bucket.getCreationDate(), log);
+    } catch (err) {
```
benzekrimaha, line 85: can we add the test case where `listObject` errors (seed skipped, request still succeeds).

lib/api/bucketUpdateQuota.js
benzekrimaha: warn, not error: IMO the failure is best-effort and self-heals at the next count-items run; nothing for a human to do. Also worth a one-line comment stating why it's non-fatal

lib/api/apiUtils/bucket/bucketCreation.js
delthas: Done — `seedBucketQuotaCapacity` is now async/await (along with `freshStartCreateBucket` and `cleanUpBucket`).

lib/api/apiUtils/bucket/bucketCreation.js
delthas: Migrated — `freshStartCreateBucket` (and `cleanUpBucket`) are now async/await, with a callback trampoline so `createBucket` and the other callback-style callers keep working unchanged.

lib/api/apiUtils/bucket/bucketCreation.js
delthas: By design it's best-effort: if the seed fails, the metric self-heals on the next count-items run and there's nothing for an operator to act on, so we must not fail bucket creation over it. Switched it to `log.warn` with a comment stating why it's non-fatal.

lib/api/apiUtils/bucket/bucketCreation.js
delthas: `callback(null)` explicitly signals "no error" (the first callback arg is the error). It's also deliberate here: a stray second argument would shift the args of `async.waterfall` callers — that actually caused a bug earlier in `objectPut` — so exactly one arg is intended. With the migration this now lives in the trampoline (`.then(() => callback(null))`).

lib/api/bucketUpdateQuota.js
delthas: `promisify.custom` is really for callbacks whose *shape* isn't the standard `(err, result)` (multiple results / a custom resolve object) — these metadata methods are standard-shaped, so it wouldn't add anything, and attaching it would mean mutating arsenal's shared `MetadataWrapper` methods. What the comment is really after — not calling `promisify()` on every invocation — I did by declaring the promisified wrappers once at module scope (arrow form so `this` is preserved and sinon stubs still resolve), matching the existing convention in `lib/routes/veeam/utils.js` (`const getBucket = promisify((...args) => metadata.getBucket(...args))`). Happy to switch to `promisify.custom` on the wrapper if you'd still prefer it.

lib/api/bucketUpdateQuota.js
delthas: Switched to `log.warn` and added a one-line comment: it's best-effort and self-heals at the next count-items run, so a seeding failure never fails the quota update.

lib/api/bucketUpdateQuota.js
delthas: Reordered — the seed now runs before `setQuota`/`updateBucket`, so the metric exists before the quota becomes enforceable and the emptiness check can't be invalidated in between.

lib/api/bucketUpdateQuota.js
delthas: Good catch — count-items does count MPU parts, so `seedEmptyBucketCapacity` now probes the MPU shadow bucket after the DelimiterVersions check and skips seeding if any in-progress upload exists, so a bucket with only uncommitted uploads is never seeded to zero. I mirrored `bucketDelete`'s own emptiness check — a single bounded `metadata.listObject(mpuShadowBucket, { prefix: 'overview', maxKeys: 1 })` (one op, scans only overview keys, tolerates `NoSuchBucket` = none). Also took the `?.length ?? 0` simplification.

lib/api/apiUtils/bucket/bucketCreation.js
delthas: Added a test for the transient/deleted cleanup path (bucketPut re-creating a flagged bucket). And yes — the creationDate is the re-created bucket's metastore date: `cleanUpBucket` seeds with `bucketMD.getCreationDate()` then persists that same `bucketMD` via `removeTransientOrDeletedLabel`→`updateBucket`, so it matches what `_evaluateQuotas` queries. The test asserts `initStub.firstCall.args[1] === md.getCreationDate()`.

tests/unit/api/bucketPut.js
delthas: Done — the test now asserts `initStub.firstCall.args[1] === md.getCreationDate()` (fetched back from metadata), pinning the exact-match keying.

tests/unit/api/bucketUpdateQuota.js
delthas: Done — the test now fetches the bucket and asserts `getQuota() === 1000n` (it's a BigInt), so the quota update is actually verified.

lib/api/bucketUpdateQuota.js
delthas: Added — a test stubs `listObject` to error and asserts no seed happens while the quota is still updated.

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -85,21 +86,37 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
...
+        // Best-effort: a missing metric self-heals at the next count-items run,
+        // so a seeding failure must never fail the bucket operation.
+        log.warn('error seeding bucket quota capacity metric', { error: err });
+    }
+}
+
+async function freshStartCreateBucket(bucket, canonicalID, log, callback) {
+    if (callback) {
+        return freshStartCreateBucket(bucket, canonicalID, log)
+            .then(() => callback(null))
```
DarkIsDude: Should we use the new then(success, error) pattern ?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -114,14 +131,17 @@ function freshStartCreateBucket(bucket, canonicalID, log, callback) {
  * @param {function} callback - callback with error or null as arguments
  * @return {undefined}
  */
-function cleanUpBucket(bucketMD, canonicalID, log, callback) {
+async function cleanUpBucket(bucketMD, canonicalID, log, callback) {
+    if (callback) {
+        return cleanUpBucket(bucketMD, canonicalID, log)
+            .then(() => callback(null))
```
DarkIsDude: Should we use the new pattern then(success, error) ?

lib/api/bucketUpdateQuota.js
```diff
@@ -1,97 +1,138 @@
...
-    );
+
+    let bucket;
+    try {
+        bucket = await validateBucket(metadataValParams, request.actionImplicitDenies, log);
+        const requestBody = await parseRequestBody(request.post, request.headers['content-type']);
+        const quotaValue = validateBucketQuotaProperty(requestBody);
+        // Seed the zero metric before persisting the quota: otherwise there is a
+        // window where the quota is enforceable but no metric exists, which
+        // disables quotas cluster-wide (the exact bug this fixes).
```
DarkIsDude: We should add least drop `(the exact bug this fixes)`.  And maybe try to simplify 🙏

DarkIsDude (approved):
LGTM

lib/api/apiUtils/bucket/bucketCreation.js
delthas: Done — switched to `.then(() => callback(null), callback)`. Applied the same two-arg form to `cleanUpBucket` and `bucketUpdateQuota` for consistency, and updated the CLAUDE.local.md trampoline convention to match — it avoids re-invoking the callback if the success continuation itself throws.

lib/api/apiUtils/bucket/bucketCreation.js
delthas: Done here too — `.then(() => callback(null), callback)`, same two-arg form.

lib/api/bucketUpdateQuota.js
delthas: Simplified to a single line and dropped the parenthetical: `// Seed before enabling the quota, so the metric exists before quotas can be enforced.`

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,137 @@
...
+    }
+}
+
+async function seedEmptyBucketCapacity(bucket, log) {
+    if (!config.isQuotaEnabled()) {
+        return;
+    }
+    const bucketName = bucket.getName();
+    try {
+        const list = await listObjects(bucketName, { maxKeys: 1, listingType: 'DelimiterVersions' }, log);
```
benzekrimaha, line 73: this is check-then-act: an object can be written after this empty listing and before the quota is persisted. Since quota tracking is still disabled, that object is not reflected in Scuba, while `$setOnInsert` leaves the zero metric in place. This silently under-enforces the newly enabled quota until count-items runs — the same problem this PR is meant to fix. How do we prevent this race?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -89,21 +86,35 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
...
-            if (err) {
-                return callback(err);
-            }
-            return removeTransientOrDeletedLabel(bucket, log, callback);
-        });
-    });
+    await createBucketMD(bucketName, bucket, log);
+    log.trace('created bucket in metadata');
+    await addToUsersBucketAsync(canonicalID, bucketName, bucket, log);
+    await removeTransientOrDeletedLabelAsync(bucket, log);
```
benzekrimaha: the bucket becomes visible before its metric is initialized, so a concurrent PUT can land between these two calls and then be hidden by the zero seed. I think we should initialize the metric before removing the transient or deleted flag.

benzekrimaha (changes requested):
I think that the current ordering still allows usage to be silently under-reported => 
- the quota-enable path has a check-then-act race between the empty listing, metric initialization, and quota persistence;
- bucket creation clears the transient/deleted flag before initializing the metric, exposing a similar race;
The callback trampolines also leave the new CodeQL `.then()` findings open; async helpers should be promise-only, with callback adaptation at their callers IMO

lib/api/apiUtils/bucket/bucketCreation.js
delthas: Good catch — reordered so `seedBucketQuotaCapacity` runs before `removeTransientOrDeletedLabel` in both `freshStartCreateBucket` and `cleanUpBucket`. While the transient/deleted flag is still set the bucket isn't finalized/usable, so nothing can be written before the metric exists; the creationDate is unchanged so the key still matches. Added a test asserting the seed is called before the flag is cleared (`updateBucket`).

lib/api/bucketUpdateQuota.js
delthas: You're right that it's check-then-act, and there's no atomic fix here without a bucket lock, since the bucket is already live (unlike the creation path, which has the transient-flag gate we can seed behind).

We're accepting this deliberately: the seed is a best-effort floor, not an authoritative count — count-items is the source of truth and reconciles at its next run. The failure mode is one bucket briefly under-enforced until then, which is far narrower than the bug this PR fixes (Scuba's deep-health failing → quotas disabled cluster-wide, indefinitely). I've documented it inline at the seed call.

If you'd want a stronger guarantee it'd need a lock; otherwise I think accept-and-self-heal is the right tradeoff. Happy to discuss — leaving this open for your call.

package.json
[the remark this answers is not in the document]
delthas: Done — now pinned to the released `8.5.11` tag (`git+https://github.com/scality/arsenal#8.5.11`), replacing the temporary commit-hash pin. ARSN-610 merged and was tagged 8.5.11.

package.json
delthas: ARSN-610 is merged and released as arsenal `8.5.11`; the pin now points at that tag instead of the branch commit. This also fixes the `yarn install --frozen-lockfile` failure (the old commit ref had been orphaned by force-pushes). Thanks for the nudge.

delthas:
The MR is ready for me, if all OK feel free to /approve / merge without me.

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,138 @@
...
+        await updateBucketMD(bucket.getName(), bucket, log);
+    } catch (err) {
+        const corsHeaders = collectCorsHeaders(request.headers.origin, request.method, bucket);
+        log.debug('error processing request', {
+            error: err,
+            method: 'bucketUpdateQuota',
         });
-        return callback(null, corsHeaders);
+        monitoring.promMetrics('PUT', bucketName, err.code, 'updateBucketQuota');
+        err.additionalResHeaders = err.additionalResHeaders || corsHeaders;
```
benzekrimaha, line 126: arsenal errors coming from metadata are shared singletons (only `customizeDescription()` clones), so this pins the first request's CORS headers on the instance and replays them on every later failure with the same error, for any bucket/origin. Can we carry the headers out without mutating `err` (e.g. a small result carrier shared with the callback shim)?

benzekrimaha (changes requested):
Thanks, all my previous comments are addressed, and the documented accept-and-self-heal trade-off works for me.
One last issue in the error path (inline).

lib/api/bucketUpdateQuota.js
```diff
@@ -1,95 +1,138 @@
...
-                    throw new Error('Invalid JSON');
-                }
-                return next(null, jsonData);
-            } catch {
-                return next(errorInstances.InvalidArgument.customizeDescription('Request body must be a JSON object'));
-            }
+async function parseRequestBody(requestBody, contentType) {
+    if (contentType === 'application/xml') {
+        try {
+            return await promisify(parseString)(requestBody, { explicitArray: false });
```
benzekrimaha: nit: `promisify(parseString)` is rebuilt on every XML request , we can hoist it to module scope next to the other promisified helpers.

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -89,21 +86,36 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
...
-        }
-        log.trace('created bucket in metadata');
-        return addToUsersBucket(canonicalID, bucketName, bucket, log, err => {
-            if (err) {
-                return callback(err);
-            }
-            return removeTransientOrDeletedLabel(bucket, log, callback);
-        });
-    });
+    await createBucketMD(bucketName, bucket, log);
```
benzekrimaha, line 112: the old `log.debug('error from metadata', ...)` on createBucket failure is gone is that intentional? the caller still gets the error, just checking we don't lose the trace.

benzekrimaha (commented):
Two last minor points, non-blocking.

lib/api/bucketUpdateQuota.js
delthas: I looked into this and I don't think it leaks here. arsenal's `errors.*` are fresh per-access getters (`ArsenalError.errors()` does `new ArsenalError(...)` on every access) — only `errorInstances.*` are shared singletons. Every error reaching this catch is fresh or a clone:
- `updateBucket` failures come from arsenal's `MetadataWrapper`, which throws `errors.InternalError` etc. → a new instance per call;
- `validateBucket` errors get their CORS headers attached per-request by `metadataUtils.attachCorsHeadersToError`;
- the quota/parse validation errors are `errorInstances.X.customizeDescription(...)`, and `customizeDescription` returns a clone.

So `err.additionalResHeaders = err.additionalResHeaders || …` mutates a per-request object, not a shared one — no cross-request replay. It's also the documented trampoline convention and the same mutation `attachCorsHeadersToError` already does across the codebase, so I'd keep it consistent here. If you're seeing a specific path that reaches this with a shared `errorInstances` error, point me at it and I'll revisit.

lib/api/bucketUpdateQuota.js
delthas: Done — hoisted `const parseStringAsync = promisify(parseString)` to module scope alongside the other promisified helpers.

lib/api/apiUtils/bucket/bucketCreation.js
delthas: Intentional, and consistent with the convention: the recently-migrated `bucketPutLifecycle` doesn't log per-step metadata errors either — it uses a single catch and lets the error propagate. The old `log.debug('error from metadata')` here was only on the `createBucket` branch (not `addToUsersBucket` / `removeTransientOrDeletedLabel`), so it wasn't a uniform trace. The error still propagates through the trampoline to `createBucket` → `bucketPut`, which hands it to the API layer where it's logged. So nothing is lost.

benzekrimaha (approved):

PR #6222, opened 2026-07-08
Handle BucketAlreadyExists races and bump arsenal for MPU afterEach fix

benzekrimaha:
## Context

ARSN-613 changes MongoDB `createBucket()` to return
`BucketAlreadyExists` when another request wins a concurrent metastore upsert.
Cloudserver must handle this explicit result instead of surfacing an
`InternalError`.

This fixes an intermittent MPU test-cleanup failure where concurrent
`ListMultipartUploads` calls race while creating the hidden MPU shadow bucket.

Paired Arsenal change:
[Arsenal #2672](https://github.com/scality/Arsenal/pull/2672).

## Arsenal dependency

Cloudserver is temporarily pinned to Arsenal commit:

`506014399741d3fd011c25288fc38c985337babe`

Before release, this SHA will be replaced with the Arsenal release tag containing
ARSN-613.

Issue: CLDSRV-919

### DIFF
--- a/lib/api/apiUtils/bucket/bucketCreation.js
+++ b/lib/api/apiUtils/bucket/bucketCreation.js
@@ -18,7 +18,6 @@ const oldUsersBucket = constants.oldUsersBucket;
 const zenkoSeparator = constants.zenkoSeparator;
 const userBucketOwner = 'admin';
 
-
 function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
     // BACKWARD: Simplify once do not have to deal with old
     // usersbucket name and old splitter
@@ -28,8 +27,7 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
         if (err && !err.is.NoSuchBucket && !err.is.BucketAlreadyExists) {
             return cb(err);
         }
-        const splitter = usersBucketAttrs ?
-            constants.splitter : constants.oldSplitter;
+        const splitter = usersBucketAttrs ? constants.splitter : constants.oldSplitter;
         let key = createKeyForUserBucket(canonicalID, splitter, bucketName);
         const omVal = {
             creationDate: new Date().toJSON(),
@@ -38,46 +36,44 @@ function addToUsersBucket(canonicalID, bucketName, bucketMD, log, cb) {
         // If the new format usersbucket does not exist, try to put the
         // key in the old usersBucket using the old splitter.
         // Otherwise put the key in the new format usersBucket
-        const usersBucketBeingCalled = usersBucketAttrs ?
-            usersBucket : oldUsersBucket;
-        return metadata.putObjectMD(usersBucketBeingCalled, key,
-            omVal, {}, log, err => {
-                if (err?.is?.NoSuchBucket) {
-                    // There must be no usersBucket so createBucket
-                    // one using the new format
-                    log.trace('users bucket does not exist, ' +
-                        'creating users bucket');
-                    key = `${canonicalID}${constants.splitter}` +
-                        `${bucketName}`;
-                    const creationDate = new Date().toJSON();
-                    const freshBucket = new BucketInfo(usersBucket,
-                        userBucketOwner, userBucketOwner, creationDate,
-                        BucketInfo.currentModelVersion());
-                    return metadata.createBucket(usersBucket,
-                        freshBucket, log, err => {
-                            // Note: In the event that two
-                            // users' requests try to create the
-                            // usersBucket at the same time,
-                            // this will prevent one of the users
-                            // from getting a BucketAlreadyExists
-                            // error with respect
-                            // to the usersBucket.
-                            // TODO: move to `.is` once BKTCLT-9 is done and bumped in Cloudserver
-                            if (err && !err.BucketAlreadyExists) {
-                                log.error('error from metadata', {
-                                    error: err,
-                                });
-                                return cb(err);
-                            }
-                            log.trace('Users bucket created');
-                            // Finally put the key in the new format
-                            // usersBucket
-                            return metadata.putObjectMD(usersBucket,
-                                key, omVal, {}, log, cb);
+        const usersBucketBeingCalled = usersBucketAttrs ? usersBucket : oldUsersBucket;
+        return metadata.putObjectMD(usersBucketBeingCalled, key, omVal, {}, log, err => {
+            if (err?.is?.NoSuchBucket) {
+                // There must be no usersBucket so createBucket
+                // one using the new format
+                log.trace('users bucket does not exist, ' + 'creating users bucket');
+                key = `${canonicalID}${constants.splitter}` + `${bucketName}`;
+                const creationDate = new Date().toJSON();
+                const freshBucket = new BucketInfo(
+                    usersBucket,
+                    userBucketOwner,
+                    userBucketOwner,
+                    creationDate,
+                    BucketInfo.currentModelVersion(),
+                );
+                return metadata.createBucket(usersBucket, freshBucket, log, err => {
+                    // Note: In the event that two
+                    // users' requests try to create the
+                    // usersBucket at the same time,
+                    // this will prevent one of the users
+                    // from getting a BucketAlreadyExists
+                    // error with respect
+                    // to the usersBucket.
+                    // TODO: move to `.is` once BKTCLT-9 is done and bumped in Cloudserver
+                    if (err && !err.BucketAlreadyExists) {
+                        log.error('error from metadata', {
+                            error: err,
                         });
-                }
-                return cb(err);
-            });
+                        return cb(err);
+                    }
+                    log.trace('Users bucket created');
+                    // Finally put the key in the new format
+                    // usersBucket
+                    return metadata.putObjectMD(usersBucket, key, omVal, {}, log, cb);
+                });
+            }
+            return cb(err);
+        });
     });
 }
 
@@ -145,16 +141,15 @@ function cleanUpBucket(bucketMD, canonicalID, log, callback) {
  * @callback called with (err, sseInfo: object)
  */
 function bucketLevelServerSideEncryption(bucket, headers, log, cb) {
-    kms.bucketLevelEncryption(
-        bucket, headers, log, (err, sseInfo) => {
-            if (err) {
-                log.debug('error getting bucket encryption info', {
-                    error: err,
-                });
-                return cb(err);
-            }
-            return cb(null, sseInfo);
-        });
+    kms.bucketLevelEncryption(bucket, headers, log, (err, sseInfo) => {
+        if (err) {
+            log.debug('error getting bucket encryption info', {
+                error: err,
+            });
+            return cb(err);
+        }
+        return cb(null, sseInfo);
+    });
 }
 
 /**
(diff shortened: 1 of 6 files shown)


SylvainSenechal (approved):

tcarmet (approved):

delthas (approved):

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -92,11 +88,15 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
...
             log.debug('error from metadata', { error: err });
             return callback(err);
         }
-        log.trace('created bucket in metadata');
+        if (err?.is?.BucketAlreadyExists) {
+            log.trace('bucket already exists in metadata');
+        } else {
+            log.trace('created bucket in metadata');
+        }
         return addToUsersBucket(canonicalID, bucketName, bucket, log, err => {
```
francoisferrand, line 102: should not be be done if bucket already exists, to avoid race and/or incorrect content?

lib/services.js
```diff
@@ -935,6 +935,19 @@ const services = {
                 // a hidden bucket.
                 return metadata.createBucket(MPUBucketName, mpuBucket, log,
                     err => {
+                        if (err?.is?.BucketAlreadyExists) {
+                            return metadata.getBucket(MPUBucketName, log,
```
francoisferrand: is this function `getMPUBucket()` called a lot (it is not an s3 api, not sure if it is called everywhere in mpu code, or just in a few places) : if that is the case, maybe we should plan another improvement to let createBucket return the bucket data as well it case it already exists.

...or maybe we could just return the (existing) `mpuBucket` variable? for actual buckets, there may be more in the BucketInfo -tags, lifecycle rules...-, but likely not the case for mpu's BucketInfo  ?

francoisferrand (commented):
nit: this is exactly the kind of PR where prettier is really adding noise... 5-10 lines of "real" change, but prettier fully rewrote 2 large files...

lib/api/apiUtils/bucket/bucketCreation.js
SylvainSenechal: I also don't understand why bucket already exists error does not cause the function to return early with an error ? 

More generally i don't understand why this wasn't catched before in the err ? 

I would've expect current behavior to be : 

CreateBucket(bucketName) => return err because collision in the db on bucket name, then return callback(err)

I think I'm missing some context information about : what trigger this bug precisely, cuz the pr talks about MPU but here we are fixing a function called freshStartCreateBucket, is this not called all the time to create any bucket ?

lib/services.js
SylvainSenechal, line 921: Mhh. I find this very weird :

It's called
GetMpuBucket
It **currently** does :

GetBucket, if not found, create it. 
So ok this is like a "get or create if not found".

But then with the new code changes, this function becomes 

GetBucket, if not found, create it. When creating it, if it already exists, getBucket 
But if the bucket was not found in the first place, why would we get an "alreadyExists" error when creating it 🤔  ? Race condition ? But then, this is infinite, because in the last getBucket you can also get a notExist error 🤔

lib/api/apiUtils/bucket/bucketCreation.js
benzekrimaha: On `BucketAlreadyExists` we now only run `addToUsersBucket` and skip `removeTransientOrDeletedLabel` / `updateBucket` with our local `BucketInfo`, so we don't overwrite metadata the concurrent winner already wrote. The winner path still runs full `cleanUpBucket` when we actually created the metastore entry. No extra metastore read.

lib/services.js
benzekrimaha: `getMPUBucket` is only called from `initiateMultipartUpload` and `listMultipartUploads`. On `BucketAlreadyExists` we now return the local `mpuBucket` , MPU shadow `BucketInfo` is minimal (no tags/lifecycle), so that should be enough and avoids an extra `getBucket`. Longer term, Arsenal could return existing bucket data on duplicate so all callers benefit.

lib/api/apiUtils/bucket/bucketCreation.js
benzekrimaha: ARSN-613 changes `metadata.createBucket()` to return `BucketAlreadyExists` on a metastore duplicate instead of silent success. Without handling it we surface `InternalError`. Previously cloudserver always continued because there was no `err` to catch.

The **MPU flake** is in `getMPUBucket` (`InitiateMultipartUpload` / `ListMultipartUploads`). `freshStartCreateBucket` is the normal `CreateBucket` path when two requests both see `NoBucketYet` and race on `metadata.createBucket` — same contract change.

We don't return `BucketAlreadyExists` to the S3 client here: the parallel check already decided this is our bucket being created. On duplicate we only ensure the users-bucket listing; we skip `updateBucket` with our local object.

lib/services.js
benzekrimaha: Two concurrent MPU calls can both get `NoSuchBucket`, one wins `createBucket`, the other gets `BucketAlreadyExists`. We return the local `mpuBucket` we built for the create attempt — same owner/displayName as the winner's shadow bucket, no second read. Not a loop: `BucketAlreadyExists` means the metastore entry exists; we don't refetch.

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -163,135 +162,154 @@
...
+                    if (err) {
+                        log.debug('error parsing acl from headers', {
+                            error: err,
+                        });
+                        return callback(err);
+                    }
+                    bucket.setFullAcl(parsedACL);
+                    return callback(null, bucket);
+                });
+            },
```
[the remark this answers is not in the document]
benzekrimaha: Out of scope for CLDSRV-919 — pre-existing callback style in this file. Can dismiss.

delthas:
Removing self as there are already 3 other reviewers, more involved in the issue. Feel free to add me again if needed.

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -92,17 +88,18 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
 function freshStartCreateBucket(bucket, canonicalID, log, callback) {
     const bucketName = bucket.getName();
     metadata.createBucket(bucketName, bucket, log, err => {
-        if (err) {
+        if (err && !err.is.BucketAlreadyExists) {
```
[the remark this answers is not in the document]
SylvainSenechal: possibly relevant

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -88,21 +88,18 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
...
             log.debug('error from metadata', { error: err });
             return callback(err);
         }
         if (err?.is?.BucketAlreadyExists) {
             log.trace('bucket already exists in metadata');
-        } else {
-            log.trace('created bucket in metadata');
+            // Ensure users-bucket listing only; do not updateBucket with our
+            // local BucketInfo when a concurrent request won createBucket.
+            return addToUsersBucket(canonicalID, bucketName, bucket, log, callback);
```
SylvainSenechal: I don't have much knowledge of this, but isn't it problematic to run addToUsersBucket when the bucket already exists ? or is it idempotent or something ? But even then, isn't addToUsersBucket already called when the bucket was created ?

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -88,21 +88,18 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
...
         }
-        return addToUsersBucket(canonicalID, bucketName, bucket, log, err => {
-            if (err) {
-                return callback(err);
-            }
-            return removeTransientOrDeletedLabel(bucket, log, callback);
-        });
+        log.trace('created bucket in metadata');
+        return cleanUpBucket(bucket, canonicalID, log, callback);
```
SylvainSenechal: I don't really know this functions but yeah it's hard to understand what's happening here, why are we dropping removeTransientOrDeletedLabel and using cleanUpBucket ?

SylvainSenechal (changes requested):
I think the pr is fine but some changes are complicated to understand (https://github.com/scality/cloudserver/pull/6222#discussion_r3589284259), also the linter with some commits not squashed yet is very hard to read

lib/api/apiUtils/bucket/bucketCreation.js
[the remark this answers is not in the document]
SylvainSenechal: ah i think its an older commit 🤔

lib/api/apiUtils/bucket/bucketCreation.js
francoisferrand: > On `BucketAlreadyExists` we now only run `addToUsersBucket`

why do we still need to run `addToUsersBucket` ? Should be done by the "successful" bucket creation?

lib/services.js
francoisferrand: > Arsenal could return existing bucket data on duplicate

that would not be `createBucket` anymore (and the semantics of "error, it already exists but here is the data" is not very nice to map) ; and not just an arsenal / mongoclientinterface  : it would change the whole "metadata" interface, and require consistent changes in MetadataWrapper and all implementation (in-mem, mongo, metadata).

since "BucketInfo is minimal [...] should be enough", no reason to go there I think.

(another change could be the opposite actually: let `getMPUBucket` only retrieve what is there, use it for read -e.g. in listMultipartUploads, handling the error appropriately- and call `createBucket` if/when needed -e.g. in initiateMultipartUpload, handling the already created error. But likely some impact still, and certainly out of scope of this ticket)

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -92,17 +88,18 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
 function freshStartCreateBucket(bucket, canonicalID, log, callback) {
     const bucketName = bucket.getName();
     metadata.createBucket(bucketName, bucket, log, err => {
-        if (err) {
+        if (err && !err?.is?.BucketAlreadyExists) {
             log.debug('error from metadata', { error: err });
             return callback(err);
         }
+        if (err?.is?.BucketAlreadyExists) {
```
francoisferrand, line 91: seems weird to do this ONLY when BucketAlreadyExists...
- this was always done previously, especially on success
- in case of error, some other process created the bucket, and should handle addToUsersBucket ?

```suggestion
        if (!err?.is?.BucketAlreadyExists) {
```

lib/api/apiUtils/bucket/bucketCreation.js
```diff
@@ -92,6 +88,12 @@ function removeTransientOrDeletedLabel(bucket, log, callback) {
 function freshStartCreateBucket(bucket, canonicalID, log, callback) {
     const bucketName = bucket.getName();
     metadata.createBucket(bucketName, bucket, log, err => {
+        if (err?.is?.BucketAlreadyExists) {
```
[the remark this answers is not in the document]
benzekrimaha: With the current ordering, `BucketAlreadyExists` returns success without modifying the winner’s metadata, ordinary errors propagate, and successful creation performs registration and cleanup. Inverting the condition would break both the success and race paths.

lib/api/apiUtils/bucket/bucketCreation.js
benzekrimaha: `addToUsersBucket` and transient cleanup run only after a successful `createBucket`; `BucketAlreadyExists` now returns success without side effects, leaving cleanup to the winning request. I did not apply the literal suggestion because inverting this condition would skip the successful cleanup path and let the race error fall through.

lib/api/apiUtils/bucket/bucketCreation.js
benzekrimaha: The request that successfully creates the bucket performs `addToUsersBucket` and transient cleanup. The concurrent loser now returns success without modifying the winner’s metadata.

lib/api/apiUtils/bucket/bucketCreation.js
benzekrimaha: That was an intermediate refactoring to reuse the existing cleanup sequence. The final version no longer does this: after successful creation it explicitly runs `addToUsersBucket` followed by `removeTransientOrDeletedLabel`; `BucketAlreadyExists` performs no cleanup.

lib/api/apiUtils/bucket/bucketCreation.js
benzekrimaha: the request that successfully creates the bucket performs `addToUsersBucket` and transient cleanup. The concurrent loser now returns success without modifying the winner’s metadata.

lib/services.js
benzekrimaha: Changing the metadata interface is unnecessary here. The MPU bucket is minimal, so the race path returns the locally constructed `mpuBucket`. The alternative split between read-only lookup and creation is broader and remains out of scope for this PR.

SylvainSenechal (approved):

delthas:
@maeldonn , please review in my place as I'm going on PTO! Thanks

maeldonn (approved):
LGTM

tests/unit/bucket/bucketCreation.js
```diff
@@ -52,32 +50,69 @@ describe('bucket creation', () => {
...
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    it('should complete creation when metadata returns BucketAlreadyExists', done => {
+        sandbox.stub(metadata, 'getBucket').callsFake((name, log, cb) => {
+            if (name === raceBucketName) {
+                return cb(errors.NoSuchBucket);
+            }
+            return cb(errors.NoSuchBucket);
```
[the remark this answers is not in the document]
benzekrimaha: Addressed in bc2264c6e (folded into the race-fix commit).

PR #6224, opened 2026-07-08
Show onward replication status on replicas

maeldonn:
Report the onward status (PENDING/PROCESSING/FAILED) instead of always REPLICA, and add x-amz-meta-replica to keep the replica flag visible.

Issue: CLDSRV-931

### DIFF
--- a/lib/utilities/collectResponseHeaders.js
+++ b/lib/utilities/collectResponseHeaders.js
@@ -91,8 +91,12 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg, returnTagC
     }
     if (objectMD.replicationInfo) {
         const { isReplica, status } = objectMD.replicationInfo;
-        if (isReplica || status) {
-            responseMetaHeaders['x-amz-replication-status'] = isReplica ? 'REPLICA' : status;
+        if (isReplica) {
+            const hasReplicated = !status || status === 'COMPLETED';
+            responseMetaHeaders['x-amz-replication-status'] = hasReplicated ? 'REPLICA' : status;
+            responseMetaHeaders['x-amz-meta-scal-replica'] = 'true';
+        } else if (status) {
+            responseMetaHeaders['x-amz-replication-status'] = status;
         }
     }
     if (Array.isArray(objectMD?.replicationInfo?.backends)) {
--- a/tests/unit/utils/collectResponseHeaders.js
+++ b/tests/unit/utils/collectResponseHeaders.js
@@ -8,12 +8,46 @@ describe('Middleware: Collect Response Headers', () => {
         assert.deepStrictEqual(headers['x-amz-replication-status'], 'REPLICA');
     });
 
-    it('should set REPLICA header from isReplica even when status is PENDING', () => {
+    it('should mark a replica with x-amz-meta-scal-replica', () => {
         const objectMD = {
-            replicationInfo: { status: 'PENDING', isReplica: true },
+            replicationInfo: { status: 'REPLICA', isReplica: true },
         };
         const headers = collectResponseHeaders(objectMD);
         assert.deepStrictEqual(headers['x-amz-replication-status'], 'REPLICA');
+        assert.deepStrictEqual(headers['x-amz-meta-scal-replica'], 'true');
+    });
+
+    it('should default to REPLICA when isReplica is true and status is absent', () => {
+        const objectMD = { replicationInfo: { isReplica: true } };
+        const headers = collectResponseHeaders(objectMD);
+        assert.deepStrictEqual(headers['x-amz-replication-status'], 'REPLICA');
+        assert.deepStrictEqual(headers['x-amz-meta-scal-replica'], 'true');
+    });
+
+    ['PENDING', 'PROCESSING', 'FAILED'].forEach(status => {
+        it(`should expose onward status ${status} on a cascaded replica`, () => {
+            const objectMD = { replicationInfo: { status, isReplica: true } };
+            const headers = collectResponseHeaders(objectMD);
+            assert.deepStrictEqual(headers['x-amz-replication-status'], status);
+            assert.deepStrictEqual(headers['x-amz-meta-scal-replica'], 'true');
+        });
+    });
+
+    it('should report REPLICA once onward replication is COMPLETED', () => {
+        const objectMD = {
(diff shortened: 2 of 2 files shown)


lib/utilities/collectResponseHeaders.js
```diff
@@ -91,8 +91,12 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg, returnTagC
     }
     if (objectMD.replicationInfo) {
         const { isReplica, status } = objectMD.replicationInfo;
-        if (isReplica || status) {
-            responseMetaHeaders['x-amz-replication-status'] = isReplica ? 'REPLICA' : status;
+        if (isReplica) {
+            const replicated = !status || ['COMPLETED', 'REPLICA'].includes(status);
+            responseMetaHeaders['x-amz-replication-status'] = replicated ? 'REPLICA' : status;
```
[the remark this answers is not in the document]
maeldonn: The commit is missing a file...

lib/utilities/collectResponseHeaders.js
```diff
@@ -91,8 +91,12 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg, returnTagC
...
+            responseMetaHeaders['x-amz-meta-replica'] = 'true';
```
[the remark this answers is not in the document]
maeldonn: Good catch on the collision. Keeping it in the `x-amz-meta-*` namespace though (stock AWS CLI/SDKs only surface `x-amz-meta-*` under `Metadata`; `x-amz-scal-*` is dropped), but adding the `scal` prefix to avoid clobbering a user key, matching the existing `x-amz-meta-scal-s3-transition-in-progress` pattern. Renamed to `x-amz-meta-scal-replica`.

tests/unit/utils/collectResponseHeaders.js
```diff
@@ -8,12 +8,39 @@ describe('Middleware: Collect Response Headers', () => {
...
+    });
+
+    it('should report REPLICA once onward replication is COMPLETED', () => {
+        const objectMD = {
+            replicationInfo: { status: 'COMPLETED', isReplica: true },
+        };
+        const headers = collectResponseHeaders(objectMD);
+        assert.deepStrictEqual(headers['x-amz-replication-status'], 'REPLICA');
+        assert.deepStrictEqual(headers['x-amz-meta-replica'], 'true');
+    });
```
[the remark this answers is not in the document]
maeldonn: Added, covering the `!status` branch.

lib/utilities/collectResponseHeaders.js
```diff
@@ -91,8 +91,12 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg, returnTagC
...
     if (objectMD.replicationInfo) {
         const { isReplica, status } = objectMD.replicationInfo;
-        if (isReplica || status) {
-            responseMetaHeaders['x-amz-replication-status'] = isReplica ? 'REPLICA' : status;
+        if (isReplica) {
+            const replicated = !status || ['COMPLETED', 'REPLICA'].includes(status);
+            responseMetaHeaders['x-amz-replication-status'] = replicated ? 'REPLICA' : status;
+            responseMetaHeaders['x-amz-meta-scal-replica'] = 'true';
+        } else if (status) {
+            responseMetaHeaders['x-amz-replication-status'] = status;
```
SylvainSenechal, line 99: shouldn't we have             
responseMetaHeaders['x-amz-meta-scal-replica'] = 'false';

Maybe not always in the else, but in some situation

delthas (approved):
LGTM code-wise.

lib/utilities/collectResponseHeaders.js
```diff
@@ -91,8 +91,12 @@ function collectResponseHeaders(objectMD, corsHeaders, versioningCfg, returnTagC
     }
     if (objectMD.replicationInfo) {
         const { isReplica, status } = objectMD.replicationInfo;
-        if (isReplica || status) {
-            responseMetaHeaders['x-amz-replication-status'] = isReplica ? 'REPLICA' : status;
+        if (isReplica) {
+            const replicated = !status || ['COMPLETED', 'REPLICA'].includes(status);
```
SylvainSenechal: It's a bit nit, but I would use temporary variables so that we understand why replicated is equal to this, it's hard to understand why replicated is true when we have no status for example.

Or maybe even if else blocks that treat different each situation we could face one by one instead of the ternary

For example : 

```suggestion
if (isReplica) {
  const isReplicaWithoutStatus = !status
  if (isReplicaWithoutStatus) {
    responseMetaHeaders['x-amz-replication-status'] = 'REPLICA'
  } else {
  ..
  }
} else {}
```

Or no top level if (isReplica)


```suggestion
const isReplicaWithoutStatus = isReplica & !status
if (isReplicaWithoutStatus) {
  responseMetaHeaders['x-amz-replication-status'] = 'REPLICA'
} else {
..
}
```

SylvainSenechal (changes requested):

lib/utilities/collectResponseHeaders.js
SylvainSenechal: Even defining a bool 'replicated' inside an if (isReplica) is weird 🤔

lib/utilities/collectResponseHeaders.js
maeldonn: We don't need the header if it's not a replica, as it might confuse users. Do you think we should keep it?

lib/utilities/collectResponseHeaders.js
maeldonn: I think that the real problem is the variable name ?

maeldonn: Simplified the condition and renamed the variable.

lib/utilities/collectResponseHeaders.js
SylvainSenechal: Mhh, it's personal opinion but I think it's always nicer to be specific than to have optional fields.

But this looks like a product question

SylvainSenechal (approved):
Maybe can ask Florent if he wants 
responseMetaHeaders['x-amz-meta-scal-replica'] = 'false';

When the object is not a replica, or if current solution (returns nothing is enough)

maeldonn:
@SylvainSenechal We omit this header to maintain consistency with the other headers.

PR #6225, opened 2026-07-09
CLDSRV-945: Support cross account rate limiting

tmacro:
- Cloudserver now passes the bucket owner canonicalId to Vault using `targetAccount` if it has been previously cached.
  - (Vault's side) If `targetAccount` differs from the requesting account targetAccount's rate limit config is retrieved and returned in response


tcarmet (approved):

anurag4DSB (approved):
Thanks for making prettier a separate commit :)

PR #6228, opened 2026-07-09
CLDSRV-955: Parallelize docker builds into a matrix

tcarmet:
Speeds up CI by building the container images in parallel instead of one after another, so the test jobs that depend on them start sooner. The release-only images that depend on the freshly built server image and are not needed by any test are moved off the critical path to run alongside the tests. Image names and tags are unchanged. This is the second of two changes tracked under CLDSRV-955 to reduce the `tests` workflow wall time; the first caches the installed dependency tree.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -192,35 +192,33 @@ jobs:
       matrix:
         include:
           # Image url and tags are preserved exactly via matrix.name/matrix.tags.
+          # cloudserver builds production then testcoverage (target2) in the same
+          # entry: testcoverage is FROM production, so building it right after -
+          # sharing the cloudserver cache scope - lets it reuse the production
+          # layers (a few seconds) instead of rebuilding them. Sequential within
+          # one job, so there is no cache-to race on scope=cloudserver.
           - name: cloudserver
             context: .
             target: production
             tags: ghcr.io/${{ github.repository }}:${{ github.sha }}
             cache-from: type=gha,scope=cloudserver
             cache-to: type=gha,mode=max,scope=cloudserver
-          # testcoverage is FROM production; it writes to its own cache scope to
-          # avoid racing the production build's cache-to on scope=cloudserver,
-          # while still reading the production scope for layer reuse.
-          - name: cloudserver-testcoverage
-            context: .
-            target: testcoverage
-            tags: ghcr.io/${{ github.repository }}:${{ github.sha }}-testcoverage
-            cache-from: |
-              type=gha,scope=cloudserver-testcoverage
-              type=gha,scope=cloudserver
-            cache-to: type=gha,mode=max,scope=cloudserver-testcoverage
+            target2: testcoverage
+            tags2: ghcr.io/${{ github.repository }}:${{ github.sha }}-testcoverage
           - name: pykmip
             context: .github/pykmip
             target: ''
             tags: ghcr.io/${{ github.repository }}/pykmip:${{ github.sha }}
             cache-from: type=gha,scope=pykmip
             cache-to: type=gha,mode=max,scope=pykmip
+            target2: ''
           - name: ci-mongodb
             context: .github/docker/mongodb
             target: ''
             tags: ghcr.io/${{ github.repository }}/ci-mongodb:${{ github.sha }}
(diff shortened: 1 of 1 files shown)


tcarmet:
## Docker build matrix - CI timing impact

Measured on a **warm** run of this branch (run `29056334730`, warm attempt) vs a recent successful **development/9.3** run (`28062248343`).

### Build phase
| | development/9.3 | this PR | delta |
|---|---|---|---|
| Build phase wall time | **623s** (one sequential `build` job) | **462s** (matrix, max of 4 parallel entries) | **−161s (~26%)** |
| Time from run start → first test job starts | 630s | 470s | **−160s** |

Matrix entry breakdown (warm):
| entry | duration |
|---|---|
| `cloudserver` (production) | 208s |
| `cloudserver-testcoverage` | **462s** ← matrix bottleneck |
| `pykmip` | 14s |
| `ci-mongodb` | 13s |
| `build-federation` (separate job) | 113s - now runs **in parallel with the tests**, no longer blocking them |

### Honest caveats
- The win is real (~160s off the pre-test critical path) but **smaller than it could be**. `cloudserver-testcoverage` is `FROM production`; splitting it into its own parallel entry with a separate cache scope means it re-builds the production layers instead of reusing them (in the old sequential job it was only ~14s, right after production). A follow-up that couples production+testcoverage in a single matrix entry would drop the matrix wall to roughly `max(~220s, pykmip, mongodb)` ≈ **~220s**, i.e. a ~400s build-phase reduction rather than ~160s.
- The `development/9.3` GHA docker layer cache warmth varies (those runs are ~weekly and the cache evicts after 7 days), so the 623s baseline is not guaranteed fully-warm; treat these as directional.

_Note: the functional suite is independently flaky on the shared MPU abort-race teardown (CLDSRV-938), unrelated to this change; this run needed reruns of a couple of single flaky tests to go green. All build/matrix jobs passed on every attempt._

tcarmet:
## Docker build - updated timing (warm, after restructure)

Structure is now a dedicated **`build-cloudserver`** job (builds production then testcoverage, sharing the `cloudserver` cache scope) + a **`build-images`** matrix (`pykmip`, `ci-mongodb`), with **`build-federation`** running off the critical path. Measured on a **warm** run (`29100971369`, attempt 2) vs `development/9.3` (`28062248343`, one sequential `build` job = 623s).

| job | duration (warm) | detail |
|---|---|---|
| `build-cloudserver` | **510s** | production **459s** + testcoverage **18s** |
| `build-images` (pykmip) | 16s | parallel |
| `build-images` (ci-mongodb) | 20s | parallel |
| `build-federation` | ~113s | now parallel with the tests, off the pre-test path |
| **build-phase wall (gates tests)** | **~510s** | max(build-cloudserver, build-images) |

**vs `development/9.3`:** 623s (sequential) → ~510s to first test job (**~113s / ~18% faster**). The gain comes almost entirely from moving `build-federation` (~113s) and the small pykmip/ci-mongodb builds off the pre-test critical path.

### Honest findings
- **The production image build dominates (~460s) and does not cache well across runs** - it was 208s on one earlier run but 459–509s on every other, warm or cold. The yarn-install + TypeScript-compile builder stage is the cost; its GHA docker-layer cache reuse is inconsistent. This is the real floor on the build phase and is unchanged by this PR.
- **Coupling testcoverage into `build-cloudserver` is a compute win, not a wall-time win.** testcoverage now reuses the production layers (18s) instead of rebuilding them (~460s as a standalone parallel entry) - so we no longer pay for a *duplicate* production build. In wall-clock it adds only ~18s after production (vs overlapping in the old parallel design), which is why the wall is ~510s rather than ~462s; the trade is ~50s of wall time for roughly one full production build's worth of saved CI compute.
- Follow-up worth considering (separate ticket): make the cloudserver **production** image build cache reliably (e.g. dedicated buildx cache for the yarn/tsc builder layer) - that's where the remaining ~250s would come from.

_Flaky note: the functional suite still trips the known MPU abort-race teardown (CLDSRV-938), unrelated to this change._

tcarmet:
Pausing the docker build parallelization work for now; focusing CLDSRV-955 on the node_modules caching change (#6227). Analysis and comparison numbers are captured in the PR comments for future reference.

PR #6229, opened 2026-07-10
CLDSRV-955: Cache cloudserver builder yarn/tsc layer with a BuildKit mount

tcarmet:
Speeds up the cloudserver image build by making the expensive dependency-install work reusable across CI runs. Most of that image's build time is the builder stage installing the git dependencies and compiling their TypeScript; today the image layer cache is all-or-nothing and is thrown away whenever the lockfile changes. A BuildKit cache mount keeps the downloaded and already-compiled packages so a build reuses them even on a layer-cache miss, and the mount is persisted between runs. This is a follow-up experiment under CLDSRV-955 aimed at the build-time floor that the job-parallelization work (#6228) can't reach.

## How it works
- Dockerfile: the builder's `yarn install` gets a `--mount=type=cache` on yarn's cache folder (which also holds the compiled git deps), so a warm cache skips the re-clone and re-`tsc`.
- Workflow: the mount lives in the BuildKit builder, not the image layers, so it is saved/restored across runs via `buildkit-cache-dance` backed by `actions/cache` (keyed on the lockfile hash).

## Validation
Draft until CI shows the effect: compare the cloudserver image build time on a warm run against a `development/9.3` build. Keeping it draft so we can confirm the mount is actually restored (look for the yarn install reusing cached git deps) before calling it a win.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -202,6 +202,24 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ github.token }}
 
+      # Persist the Dockerfile builder's yarn cache mount across runs. The mount
+      # (holding downloaded + compiled git deps) is not part of the image layer
+      # cache, so buildkit-cache-dance saves/restores it via actions/cache.
+      - name: Restore Docker yarn cache
+        id: docker-yarn-cache
+        uses: actions/cache@v4
+        with:
+          path: docker-yarn-cache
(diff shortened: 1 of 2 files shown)


tcarmet:
Pausing this BuildKit cache-mount experiment; not investing further effort for now. (Also blocked at startup by the repo Actions allow-list, which would need buildkit-cache-dance added.)

PR #6230, opened 2026-07-20
fix putData : detect collision on specific provided version id

SylvainSenechal:
Issue: CLDSRV-953

The putData collision detection previously fetched the master object and compared its versionId against the incoming x-scal-version-id header. This caused issues when replicating a non-current version : the master's versionId wouldn't match, so no collision was detected and data was re-uploaded unnecessarily

Fix : decode the x-scal-version-id header in the router and use it as the metadata lookup key, so objMd is always the specific version being replicated

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -432,38 +432,26 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
         return callback(errorInstances.BadRequest.customizeDescription(errMessage));
     }
 
-    const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
-    if (incomingVersionIdEncoded !== undefined) {
-        const incomingVersionIdDecoded =
-            incomingVersionIdEncoded !== 'null' ? decode(incomingVersionIdEncoded) : 'null';
-        if (incomingVersionIdDecoded instanceof Error) {
-            log.error('crr putData: failed to decode x-scal-version-id header', {
-                method: 'putData',
-                error: incomingVersionIdDecoded.message,
-            });
-            return callback(
-                errorInstances.BadRequest.customizeDescription('bad request: invalid x-scal-version-id header'),
-            );
-        }
-        if (objMd && objMd.versionId === incomingVersionIdDecoded) {
-            // Data already at destination for this version; return 409 with the existing
-            // microVersionId so backbeat can decide if putMetadata is still needed.
-            log.debug('crr putData: version already at destination', {
-                method: 'putData',
-                bucketName: request.bucketName,
-                objectKey: request.objectKey,
-                hasMicroVersionId: !!objMd.microVersionId,
-            });
-            request.resume();
-            return _respondWithHeaderCrrConflict(
-                response,
-                log,
-                callback,
-                VersionIdCollisionException.name,
-                'version id already at destination',
-                objMd.microVersionId,
-            );
-        }
+    const incomingVersionIdEncoded = request.query?.versionId;
+    if (incomingVersionIdEncoded !== undefined && objMd) {
+        // objMd is the specific version requested via the versionId query param.
+        // Its existence means the data is already at the destination. Return 409 with the
+        // existing microVersionId so backbeat can decide if putMetadata is still needed.
+        log.debug('crr putData: version already at destination', {
+            method: 'putData',
+            bucketName: request.bucketName,
+            objectKey: request.objectKey,
+            hasMicroVersionId: !!objMd.microVersionId,
+        });
+        request.resume();
+        return _respondWithHeaderCrrConflict(
+            response,
+            log,
+            callback,
+            VersionIdCollisionException.name,
+            'version id already at destination',
+            objMd.microVersionId,
+        );
     }
 
     writeContinue(request, response);
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
         "vaultclient": "scality/vaultclient#8.5.8",
         "werelogs": "scality/werelogs#semver:^8.2.4",
         "ws": "^8.18.0",
-        "@scality/cloudserverclient": "1.0.9",
+        "@scality/cloudserverclient": "1.0.12",
         "xml2js": "^0.6.2"
     },
     "devDependencies": {
--- a/lib/api/apiUtils/object/versioning.js
+++ b/lib/api/apiUtils/object/versioning.js
@@ -1,4 +1,5 @@
 const { errorInstances, versioning } = require('arsenal');
+const { ExternalNullVersionId } = versioning.VersioningConstants;
 const async = require('async');
 
 const metadata = require('../../../metadata/wrapper');
@@ -18,7 +19,7 @@ const nonVersionedObjId = versionIdUtils.getInfVid(config.replicationGroupId);
  * fails due to improper format, otherwise undefined or the decoded version id
  */
 function decodeVID(versionId) {
-    if (versionId === 'null') {
+    if (versionId === ExternalNullVersionId) {
         return versionId;
     }
 
--- a/tests/functional/backbeat/putData.js
+++ b/tests/functional/backbeat/putData.js
@@ -14,6 +14,7 @@ const { BackbeatRoutesClient, PutDataCommand, VersionIdCollisionException } = re
 const { generateVersionId, encode: encodeVersionId } = versioning.VersionID;
 
 const TEST_BUCKET = `bucket-putdata-${uuidv4().split('-')[0]}`;
+const TEST_BUCKET_UNVERSIONED = `bucket-putdata-unver-${uuidv4().split('-')[0]}`;
 const OBJECT_BODY = 'imAboutToBeCascadedWitNoParachuteInMyBack';
 const OBJECT_MD5_HEX = createHash('md5').update(OBJECT_BODY).digest('hex');
 const CANONICAL_ID = '79a59df900b949e55d96a1e698fbacedfd6e09d98eacf8f8d5218e7cd47ef2be';
@@ -55,6 +56,8 @@ before(async () => {
             VersioningConfiguration: { Status: 'Enabled' },
         }),
     );
+
+    await s3.send(new CreateBucketCommand({ Bucket: TEST_BUCKET_UNVERSIONED }));
 });
 
 describe('putData : VersionId collision detection', () => {
@@ -106,17 +109,63 @@ describe('putData : VersionId collision detection', () => {
         const output = await putData(key, { versionId: differentVersionId });
         assert.ok(output.Location, 'should return a Location when data is written normally');
     });
+
+    it('should throw VersionIdCollisionException when a non-current version already exists', async () => {
+        const key = 'putdata-non-current-collision';
+
+        const v1Result = await s3.send(
+            new PutObjectCommand({
+                Bucket: TEST_BUCKET,
+                Key: key,
+                Body: Buffer.from(OBJECT_BODY),
+                ContentType: 'text/plain',
+            }),
+        );
+        const v1VersionId = v1Result.VersionId;
+        assert.ok(v1VersionId, 'first PutObject should return a VersionId');
+
+        const v2Result = await s3.send(
+            new PutObjectCommand({
+                Bucket: TEST_BUCKET,
+                Key: key,
+                Body: Buffer.from(OBJECT_BODY),
+                ContentType: 'text/plain',
+            }),
+        );
+        assert.ok(v2Result.VersionId, 'second PutObject should return a VersionId');
+
+        // putData on v1 (non-current) must detect the collision, not just compare against master
+        try {
+            await putData(key, { versionId: v1VersionId });
+            assert.fail('expected VersionIdCollisionException');
+        } catch (err) {
+            assert.ok(
+                err instanceof VersionIdCollisionException,
+                `expected VersionIdCollisionException, got ${err.constructor.name}`,
+            );
+            assert.strictEqual(err.microVersionId, '', 'microVersionId should be empty for original write state');
+        }
+    });
 });
 
 describe('putData : null-version objects (ExternalNullVersionId)', () => {
-    // Null-version objects created before versioning was enabled use Arsenal constant ExternalNullVersionId = 'null'
-    // getEncodedVersionId() returns 'null' as-is (no base62 encoding), and objMd.versionId is
-    // undefined in metadata : collision detection is not possible, so putData must write normally.
-    it('should write normally when VersionId is "null" (ExternalNullVersionId)', async () => {
+    it('should write normally when VersionId is "null" and destination has no null version', async () => {
+        // Versioned bucket: existing object has a real versionId, not a null version.
+        // Fetching with ExternalNullVersionId returns no objMd => no collision => write normally.
+        const key = 'putdata-null-version-no-collision';
+        await s3.send(
+            new PutObjectCommand({
+                Bucket: TEST_BUCKET,
+                Key: key,
+                Body: Buffer.from(OBJECT_BODY),
+                ContentType: 'text/plain',
+            }),
+        );
+
         const output = await backbeatClient.send(
             new PutDataCommand({
                 Bucket: TEST_BUCKET,
-                Key: 'putdata-null-version',
+                Key: key,
                 ContentMD5: OBJECT_MD5_HEX,
                 CanonicalID: CANONICAL_ID,
                 VersioningRequired: true,
@@ -126,10 +175,44 @@ describe('putData : null-version objects (ExternalNullVersionId)', () => {
         );
         assert.ok(output.Location, 'putData with null-version versionId should write normally');
     });
+
+    it('should throw VersionIdCollisionException when a null version already exists at destination', async () => {
+        // Unversioned bucket: objects have no versionId in metadata (they are null versions).
+        // putData with ExternalNullVersionId must detect the collision.
+        const key = 'putdata-null-version-collision';
+        await s3.send(
+            new PutObjectCommand({
+                Bucket: TEST_BUCKET_UNVERSIONED,
+                Key: key,
+                Body: Buffer.from(OBJECT_BODY),
+                ContentType: 'text/plain',
+            }),
+        );
+
+        try {
+            await backbeatClient.send(
+                new PutDataCommand({
+                    Bucket: TEST_BUCKET_UNVERSIONED,
+                    Key: key,
+                    ContentMD5: OBJECT_MD5_HEX,
+                    CanonicalID: CANONICAL_ID,
+                    VersionId: ExternalNullVersionId,
+                    Body: Buffer.from(OBJECT_BODY),
+                }),
+            );
+            assert.fail('expected VersionIdCollisionException');
+        } catch (err) {
+            assert.ok(
+                err instanceof VersionIdCollisionException,
+                `expected VersionIdCollisionException, got ${err.constructor.name}`,
+            );
+            assert.strictEqual(err.microVersionId, '', 'microVersionId should be empty for null-version collision');
+        }
+    });
 });
 
-describe('putData : baseline (no cascade headers)', () => {
-    it('should succeed normally when putData has no VersionId header', async () => {
+describe('putData : baseline', () => {
+    it('should succeed normally when putData has no VersionId query param', async () => {
         const key = 'putdata-baseline-no-version-id';
         const output = await putData(key);
         assert.ok(output.Location, 'putData without VersionId should return a Location');
--- a/tests/unit/routes/routeBackbeat.js
+++ b/tests/unit/routes/routeBackbeat.js
@@ -229,10 +229,9 @@ describe('routeBackbeat', () => {
             'content-md5': '1234',
             'content-length': '0',
             'x-scal-versioning-required': 'true',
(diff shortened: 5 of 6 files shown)


lib/routes/routeBackbeat.js
```diff
@@ -450,8 +450,9 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
                 errorInstances.BadRequest.customizeDescription('bad request: invalid x-scal-version-id header'),
             );
         }
-        if (objMd && objMd.versionId === incomingVersionIdDecoded) {
```
SylvainSenechal, line 448: No need for objMd.versionId === incomingVersionIdDecoded anymore as the middleware is already fetching objMd for the specific versionID provided in the header

delthas (approved):

lib/routes/routeBackbeat.js
```diff
@@ -438,20 +439,21 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
-                error: incomingVersionIdDecoded.message,
+                error: decoded.message,
             });
             return callback(
                 errorInstances.BadRequest.customizeDescription('bad request: invalid x-scal-version-id header'),
             );
         }
-        if (objMd && objMd.versionId === incomingVersionIdDecoded) {
-            // Data already at destination for this version; return 409 with the existing
+        if (objMd) {
```
maeldonn: `if (objMd)` only means "this version exists" because the router's `isPutDataApi` block fetched the header version instead of the master; that coupling is easy to break. Clearer to drop the router block and fetch the version here with `metadataGetObject` (returns `undefined` when missing):

```js
const incomingVersionId = decode(incomingVersionIdEncoded);
// ... BadRequest on decode error ...
return metadataGetObject(request.bucketName, request.objectKey, incomingVersionId, null, log, (err, versionMd) => {
    if (err) {
        return callback(err);
    }
    if (versionMd) {
        // existing 409 conflict path, using versionMd.microVersionId
    }
    return writeData();
});
```

lib/routes/routeBackbeat.js
```diff
@@ -2069,7 +2071,20 @@ function routeBackbeat(clientIP, request, response, log) {
                     });
                     return next(errors.InvalidArgument);
                 }
-                const versionId = decodedVidResult;
+                let versionId = decodedVidResult;
+                // For putData api: the version to check is passed
+                // via x-scal-version-id header, not the URL query. Fetch that specific
+                // version so objMd matches the replicated version, not always the master.
+                const isPutDataApi = request.method === 'PUT' && request.resourceType === 'data';
```
maeldonn: The router shouldn't special-case one handler, and these guards duplicate putData's. Move the version lookup into putData itself (see handler comment) so this block goes away.

lib/routes/routeBackbeat.js
```diff
@@ -2069,7 +2071,20 @@ function routeBackbeat(clientIP, request, response, log) {
...
+                if (isPutDataApi) {
+                    const versionIdHeader = request.headers['x-scal-version-id'];
+                    if (versionIdHeader !== undefined && versionIdHeader !== ExternalNullVersionId) {
+                        const decoded = decode(versionIdHeader);
+                        if (!(decoded instanceof Error)) {
```
maeldonn: Decode failure is ignore, should it be handled ?

maeldonn (commented):
Logic is fine, need a little refactor

lib/routes/routeBackbeat.js
SylvainSenechal: Mhh I'll let other reviewers comment on that but I don't fully agree : 

Middleware/router is usually a good place to pre-extract generic info, here, the router already get the object after validating query params (bucket/object names), and it does it for all backbeat routes.
If we re add a metadataGetObject in this specific api, we are gonna call it twice because it's already called in the router.

I think it's an ok design to have the router handle the get object, the only thing a bit dirty is indeed that new logic where the router fetches an object specific version id only for putData

lib/routes/routeBackbeat.js
SylvainSenechal: This is debatable if we consider that the router acts as a middleware that fetches objMd for all apis, in that case, we wouldn't want each apis functions to have their own extra ways of dealing with fetching again objMd.

Let's discuss with the first comment

lib/routes/routeBackbeat.js
SylvainSenechal: it's handled in putData, and I prefer to leave it in putData.

Here, this handler only decide "try to fetch specific version id if provided and valid, else, fallback to versionId provided in the query"

lib/routes/routeBackbeat.js
SylvainSenechal, line 1: General comment regarding Maël 3 feedbacks : I mostly agree with him that there is weirdness in the code changes, and that the current solution doesn't feel fully satisfying, but I don't think there is any perfect way to do it, else we would need a bigger refactoring

lib/routes/routeBackbeat.js
```diff
@@ -433,20 +434,21 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
     }
 
     const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
-    if (incomingVersionIdEncoded !== undefined) {
-        const incomingVersionIdDecoded =
-            incomingVersionIdEncoded !== 'null' ? decode(incomingVersionIdEncoded) : 'null';
-        if (incomingVersionIdDecoded instanceof Error) {
+    // ExternalNullVersionId means a pre-versioning null object: collision detection is not applicable.
+    if (incomingVersionIdEncoded !== undefined && incomingVersionIdEncoded !== ExternalNullVersionId) {
```
SylvainSenechal: abit hard to read but the diff is equivalent, in both case, the logic is : 
"do not try to decode a 'null' versionId", as this would throw an error, and we dont wanna throw on 'null' versionId, they are just special case.

Prefer this new code, to avoid weird ternary and start using ExternalNullVersionId arsenal constant which will remind developers that 'null' versionId is a thing 👀

SylvainSenechal:
@francoisferrand added as Maël not here

lib/routes/routeBackbeat.js
francoisferrand: I think this is not a theoretical problem, we need to be practical:
- the router makes one metadata call already; we must not make a second metadata to avoid modifying the router on principle. Metadata calls are costly at scale, we should not make a useless one.
- in addition, the router does auth check on the object it requests. So pointing at the wrong object could lead to incorrect auth (esp. in cross-account case)
- modifying the router must be done carefully though, to ensure we don't break any case: neither `putData`, nor other cases
- I think there is a case in putMetadata where we also make the call without versionId, and eventually need to make a second metadata call to get actual version. Bonus point if we can fix it at the same time (but anyway should look at it, as it could steer the change in another direction)

lib/routes/routeBackbeat.js
francoisferrand: the router already has many cases, which are kind of hard-coded.

Maybe we can do this based on the presence of the header (if versionIdHeader ...)? I.e. set the `query.versionId` field from this header if it was not present already, before calling `decodeVersionId` ?

either way, even if we keep the "is put api" guard, no point doing both `decodeVersionId` and handling if this header:

```suggestion
const versionId = ...... ? // whatever condition: header present, put api, ...
    decodeVersionId(request.query) : decodeScalVersionId(header);
```
(or move it in a separate function, out of here)

lib/routes/routeBackbeat.js
```diff
@@ -433,20 +434,21 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
+                error: decoded.message,
             });
             return callback(
                 errorInstances.BadRequest.customizeDescription('bad request: invalid x-scal-version-id header'),
             );
         }
-        if (objMd && objMd.versionId === incomingVersionIdDecoded) {
-            // Data already at destination for this version; return 409 with the existing
+        if (objMd) {
+            // objMd is the specific version (fetched by versionId from x-scal-version-id header)
```
francoisferrand: not true : in case of decode error, you fallback to the "old" path.... so you may still receive an objMD from the master, even though a version id was provided.

i.e. same as https://github.com/scality/cloudserver/pull/6230/changes#r3637279080 : on header decode error the API should fail. It should not fallback to requesting master if a versionId is provided.

lib/routes/routeBackbeat.js
```diff
@@ -433,20 +434,21 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
     }
 
     const incomingVersionIdEncoded = request.headers['x-scal-version-id'];
-    if (incomingVersionIdEncoded !== undefined) {
-        const incomingVersionIdDecoded =
-            incomingVersionIdEncoded !== 'null' ? decode(incomingVersionIdEncoded) : 'null';
-        if (incomingVersionIdDecoded instanceof Error) {
+    // ExternalNullVersionId means a pre-versioning null object: collision detection is not applicable.
```
francoisferrand: > `pre-versioning null object`

that is called a "null version"

> collision detection is not applicable

why? and especially how is it relevant to this block: here the condition just skip the _decode()_ call, there is nothing about collision detection here.

```suggestion
    // ExternalNullVersionId means a null version, which does not need decoding
```

--- 
I don't really mind removing the ternary, but it conveying the indent better: "if a versionId is set, decode it" (null version does not have a different meaning/intent, it is really an implementation detail of the decoding).
However I wonder if it really needs to be "decoded" as `"null"` : don't we use an empty string as the decoded null-version marker? (not an issue with this PR, but I wonder if there wasn't a bug here...)

lib/routes/routeBackbeat.js
```diff
@@ -2047,7 +2049,20 @@ function routeBackbeat(clientIP, request, response, log) {
                     });
                     return next(errors.InvalidArgument);
                 }
-                const versionId = decodedVidResult;
+                let versionId = decodedVidResult;
+                // For putData api: the version to check is passed
+                // via x-scal-version-id header, not the URL query. Fetch that specific
+                // version so objMd matches the replicated version, not always the master.
+                const isPutDataApi = request.method === 'PUT' && request.resourceType === 'data';
+                if (isPutDataApi) {
```
francoisferrand: don't we have the same issue in `putMetadata` ?
e.g. if the router is not modified, that may also receive the "master" instead of the specific version, and thus not behave as expected...

lib/routes/routeBackbeat.js
SylvainSenechal: Ok so lets keet it here, will recheck putmetadata

lib/routes/routeBackbeat.js
SylvainSenechal: nvm, considering in this branch we have 
isPutData true
versionIdHeader provided

you're right
it means it was specifically requested to query a certain versionId => error if it fails

lib/routes/routeBackbeat.js
SylvainSenechal: Yeah so ideally, decode should handle null version without throwing error, and be able to differentiate for us nullVersion/validDecoded/ErrorDecoded but I don't feel like changing decode, sounds like it could break a bunch of other stuff

lib/routes/routeBackbeat.js
SylvainSenechal: you're right, considering that versionIdHeader is provided, it means the client specifically asks for a version ID, so now the middleware will return an error when failing to decode

lib/routes/routeBackbeat.js
SylvainSenechal: I checked cloudserverClient, we only have 3 other apis (MPUxx) using the versionId in header, and these request are treated differently. We can remove the "if isPutData" and do something simpler

lib/routes/routeBackbeat.js
SylvainSenechal: And you're right about the intent, its better to have 

if versionId is defined
if versionId is a null version: continue
decode version id

than 

if versionId is defined and versionId is not a null version
decode version id

result is the same but intent feels different

lib/routes/routeBackbeat.js
SylvainSenechal: putMetadata is safe, versionId is provided by backbeat through query parameters, and objMd is queried based on that specific versionid.

Although, the current design is weird, we have a backbeat api that accepts versionIDs sometimes through headers (putdata), sometimes through query params (putMetadata)... 

The header was actually added for crr cascade, I think the reason we used header was thats the encoded characters are better handled than in query params (because versoin id can contains whitespace), and also maybe something about header being directly readable without having to decode the whole body but Im not even sure this is true. Still, we end up with something thats not super clean, and had we use query params instead of header, wouldnt even need this pr

lib/routes/routeBackbeat.js
```diff
@@ -434,35 +435,28 @@ function putData(request, response, bucketInfo, objMd, log, callback) {
...
-                response,
-                log,
-                callback,
-                VersionIdCollisionException.name,
-                'version id already at destination',
-                objMd.microVersionId,
-            );
+        // ExternalNullVersionId means a null version, which does not need decoding
+        if (incomingVersionIdEncoded !== ExternalNullVersionId) {
+            if (objMd) {
```
SylvainSenechal: Actually removed the whole decode here : It's already done in the middleware, we could almost do the whole remaining logic in the middleware but its probably not the right place

lib/routes/routeBackbeat.js
francoisferrand: > else we would need a bigger refactoring

generally : if the proper way requires a refactoring, it should be done. The risk otherwise is to end up with an accumulation of code which is just slightly off - but taken together ends up being a huge pile of mess...

so one must dig into this "not fully satisfying feeling" : is it because we are introducing debt? is that debt something that may come back at us, or something easily manageable? what refactoring would be required? would this refactoring really help the problem _here_, or just cleanup the rest (but the change would still be as complex)? How long would this refactor take, is it worth it? is it worth it to solve the specific problem at hand? Is it because the problem should be solved more simply? in that case, does the complexity come from the problem it self (i.e. not so easy because of some corner case/...), from some implementation choice, ...? are there some conditions (context) which could allow for a better solution? ...

lib/routes/routeBackbeat.js
francoisferrand: should have a conflict as well in case of an existing object with `null` versionId!

```suggestion
        if (incomingVersionIdEncoded !== undefined && objMd) {
```

lib/routes/routeBackbeat.js
```diff
@@ -2039,15 +2033,31 @@ function routeBackbeat(clientIP, request, response, log) {
...
                     return routeNonObjectRequest(request, response, userInfo, log, next);
                 }
-                const decodedVidResult = decodeVersionId(request.query);
-                if (decodedVidResult instanceof Error) {
-                    log.trace('invalid versionId query', {
-                        versionId: request.query.versionId,
-                        error: decodedVidResult,
-                    });
-                    return next(errors.InvalidArgument);
+                const versionIdHeader = request.headers['x-scal-version-id'];
```
francoisferrand: this makes the `x-scal-version-id` _overrides_ the (possibly missing) `query.versionId` field.

- Can you please double check which _routes_ are affected / in which case (i.e. which route use x-scal-versionId), and especially confirm in each case that it should indeed look up this object instead of the query.versionId / master one?
- Should we abort/fail -or at least log a warning- if the request has both x-scal-version-id and query.versionId ? If fear this code could allow abused to silently override the query.versionId field, and possibly break stuff...
- Do we have some validation about _allowed_/_required_ fields, eg. putData has an `x-scal-version-id`, deleteObjectFromLifecycle has a required `query.versionId`, etc... It does not seem to be handled here, is it handled in each handler? (maybe not a new issue, but we should consider if we need to add something for putData/putMetadata, and possibly a followup to cover the other/existing API...)

lib/routes/routeBackbeat.js
```diff
@@ -2039,15 +2033,31 @@ function routeBackbeat(clientIP, request, response, log) {
...
+                let versionId;
+                if (versionIdHeader !== undefined) {
+                    if (versionIdHeader !== ExternalNullVersionId) {
+                        const decoded = decode(versionIdHeader);
+                        if (decoded instanceof Error) {
+                            return next(
+                                errorInstances.BadRequest.customizeDescription(
+                                    'bad request: invalid x-scal-version-id header',
```
francoisferrand: should we not use the same error for both x-scal-version-id and request.query.versionid decoding?

lib/routes/routeBackbeat.js
```diff
@@ -2039,15 +2033,31 @@ function routeBackbeat(clientIP, request, response, log) {
...
-                        versionId: request.query.versionId,
-                        error: decodedVidResult,
-                    });
-                    return next(errors.InvalidArgument);
+                const versionIdHeader = request.headers['x-scal-version-id'];
+                let versionId;
+                if (versionIdHeader !== undefined) {
+                    if (versionIdHeader !== ExternalNullVersionId) {
+                        const decoded = decode(versionIdHeader);
+                        if (decoded instanceof Error) {
```
francoisferrand: is this not the `decodeVID()` function ?

lib/routes/routeBackbeat.js
```diff
@@ -2039,15 +2033,31 @@ function routeBackbeat(clientIP, request, response, log) {
...
+                } else {
+                    const decodedVidResult = decodeVersionId(request.query);
+                    if (decodedVidResult instanceof Error) {
+                        log.trace('invalid versionId query', {
+                            versionId: request.query.versionId,
+                            error: decodedVidResult,
+                        });
+                        return next(errors.InvalidArgument);
+                    }
+                    versionId = decodedVidResult;
```
francoisferrand: nit: code is kind of duplicated, and kind of not the right level for this function

could we change "remap" the x-scal-version-id for some routes maybe?
e.g. 
```js
if (request.headers['x-scal-version-id'] !== undefined) {
   request.query = request.headers['x-scal-version-id'];
}
const decodedVidResult = decodeVersionId(request.query);
[...]
```

or just move this logic in a dedicated function?
```js
func decodeRequestVersionId(request) {
   const versionIdHeader = request.headers['x-scal-version-id']
   if (versionIdHeader !== undefined) {
      return versionIdHeader;
   }
   return decodeVersionId(request.query);
}
[...]
const decodedVidResult = decodeRequestVersionId(request);
if (decodedVidResult instanceof Error) {
    return next(errors.InvalidArgument.customizeDescription(decodedVidResult.error));
}
const versionId = decodedVidResult;
```

lib/routes/routeBackbeat.js
francoisferrand: the conflict logic does not belong in middleware - it is specific to some routes only. It's not about _code_ (we could do anything), but about abstractions and layers: and that logic certainly does not belong in the middleware, while decoding the versionId does.

lib/routes/routeBackbeat.js
SylvainSenechal: no decodeVersionId is for request.query (helper that extract the version id from the query before decoding)

and for the header we use directly decode from arsenal

edit: ok there is a third one decodeVID... 

I'm gonna check together with your other comments, I think this will be clearer if extracted into a function

lib/routes/routeBackbeat.js
SylvainSenechal: I did some refactoring, its nice the decodeVID function handle the 'null' case. 
I think it's simple enough that we don't need a helper function and can inline it.

i can't help but notice we could refactor it further... between arsenal and cloudserver we have 3 decodes functions....
 
decodeVersionId
decodeVID
decode

SylvainSenechal:
@francoisferrand I think the decode version id is clearer now although as I said in one of my response comment we could probably go even further in the refactoring of this function.

What I wanna address though is this comment, not sure you saw it : https://github.com/scality/cloudserver/pull/6230#discussion_r3808033867

I realize the version id passed in the header was added for CRR Cascade, and it doesn't seem like we had a strong argument to do it vs just passing it in query params like the other.

If we changed cloudserver client to pass it in the query param, it would be less troublesome and cleaner, actually this whole pr wouldn't have existed because we would've fetched the right version from the query directly..

lib/routes/routeBackbeat.js
SylvainSenechal: Code has changed, comment is automatically solved by new code

lib/routes/routeBackbeat.js
SylvainSenechal: This is true, I just verified and added a test that when provided version id is 'null', the backend does fetch the null version instead of master which i wasnt sure about

lib/routes/routeBackbeat.js
SylvainSenechal: Good remarks, and I talked about it in these 2 comments : 
https://github.com/scality/cloudserver/pull/6230#discussion_r3808033867
https://github.com/scality/cloudserver/pull/6230#issuecomment-5344471196

The bottom line is, I feel we messed up something in the design of CRR Cascade by adding version id in the header instead of query params, and well we still have time to change it so maybe we should 🤔

francoisferrand (changes requested):
to be changed with https://github.com/scality/cloudserverclient/pull/28

SylvainSenechal:
Okay hopefully this time we're done.
Still a few changes and new tests

francoisferrand (approved):

package.json
francoisferrand, line 1: should release cloudserver as well?

PR #6231, opened 2026-07-21
Refactor getReplicationInfo with object parameters

SylvainSenechal:
Issue: CLDSRV-950

Refactor getReplicationInfo with object parameters
Not super useful but we said we would do it during the crr cascade so here we are


SylvainSenechal:
messed up pr because i used base 9.3 first

PR #6235, opened 2026-07-23
Enforce microVersionId conflicts at the metadata write level

maeldonn:
The JS pre-checks in putMetadata are not atomic: two concurrent CRR writes can
both pass them and let a stale revision overwrite a newer one. This attaches a
conditional put ($gt on microVersionId) so the write itself rejects stale
revisions with PreconditionFailed, mapped to the same conflict response as the
pre-checks. Enabled only on backends supporting conditional puts (mongodb).

Issue: CLDSRV-952

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -587,6 +587,9 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
     // Note: '' is falsy so the presence check must be !== undefined, not truthy.
     const hasMicroVersionId = encodedMicroVersionId !== undefined;
     let incomingMicroVersionId = null;
+    const options = {
+        overheadField: constants.overheadField,
+    };
     if (hasMicroVersionId && objMd) {
         // '' means source has no microVersionId, treated as older revision
         incomingMicroVersionId = encodedMicroVersionId === '' ? null : decode(encodedMicroVersionId);
@@ -639,6 +642,13 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 conflictErr.message,
                 conflictErr.mvId,
             );
+        }
+
+        // Atomic counterpart of the JS pre-checks above, performed at database level.
+        if (incomingMicroVersionId !== null) {
+            options.conditions = {
+                $or: [{ microVersionId: { $exists: false } }, { microVersionId: { $gt: incomingMicroVersionId } }],
+            };
         }
     }
 
@@ -790,10 +800,6 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             versioning = omVal.replicationInfo.isReplica;
             omVal.replicationInfo.isNFS = !omVal.replicationInfo.isReplica;
         }
-
-        const options = {
-            overheadField: constants.overheadField,
-        };
 
         // NOTE: When 'versioning' is set to true and no 'versionId' is specified,
         // it results in the creation of a "new" version, which also updates the master.
@@ -1045,7 +1051,23 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     });
                 },
             ],
-            callback,
+            (err, results) => {
+                if (err?.is?.PreconditionFailed && mongoOptions.conditions) {
+                    log.debug('putMetadata: write rejected, incoming microVersionId is not newer than stored', {
+                        method: 'putMetadata',
+                        bucketName,
+                        objectKey,
+                    });
+                    request.resume();
+                    return _respondWithHeaderCrrConflict(
+                        response, log, callback,
+                        StaleMicroVersionIdException.name,
+                        'incoming revision is not newer than stored',
+                        objMd?.microVersionId,
+                    );
+                }
+                return callback(err, results);
+            },
         );
     });
 }
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -587,7 +587,7 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
     // Note: '' is falsy so the presence check must be !== undefined, not truthy.
     const hasMicroVersionId = encodedMicroVersionId !== undefined;
     let incomingMicroVersionId = null;
-    const options = {
+    const metadataOptions = {
         overheadField: constants.overheadField,
     };
     if (hasMicroVersionId && objMd) {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
         "@opentelemetry/instrumentation-ioredis": "~0.64.0",
         "@opentelemetry/instrumentation-mongodb": "~0.69.0",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/arsenal#8.5.6",
+        "arsenal": "git+https://github.com/scality/arsenal#fa9abec316dc0a48b9511eec5b41f7b9a626d5f4",
```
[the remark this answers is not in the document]
maeldonn: Not yet released, will be updated before merge.

lib/routes/routeBackbeat.js
```diff
@@ -1067,7 +1072,23 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     });
                 },
             ],
-            callback,
+            (err, results) => {
+                if (err?.is?.PreconditionFailed) {
+                    log.debug('putMetadata: write rejected, stored microVersionId is not newer than incoming', {
```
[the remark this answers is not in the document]
maeldonn: Updated.

package.json
benzekrimaha: reminder to point to the released arsenal tag containing ARSN-619 before merging (`/after_pull_request` on scality/Arsenal#2678 if needed).

lib/routes/routeBackbeat.js
```diff
@@ -814,6 +814,11 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             overheadField: constants.overheadField,
         };
 
+        // atomic counterpart of the pre-checks above ($gt cannot match a missing microVersionId)
+        if (incomingMicroVersionId !== null && objMd?.microVersionId && metadata.supportsConditionalPutObjectMD?.()) {
+            options.conditions = { microVersionId: { $gt: incomingMicroVersionId } };
```
benzekrimaha: could the condition also cover the missing-field case, e.g. `$or: [{ microVersionId: { $exists: false } }, { microVersionId: { $gt: … } }]`? otherwise this window stays unprotected , if `translateConditions` cannot express it, worth mentionning in the comment.

benzekrimaha (changes requested):
Logic LGTM.

benzekrimaha (approved):

lib/routes/routeBackbeat.js
SylvainSenechal: Yeah, known limitation, I will rediscuss with Maël next week, but yeah we would need to update arsenal with some more logic to deal with $exist, $or which may be more ocmplex

package.json
```diff
@@ -35,7 +35,7 @@
         "@opentelemetry/instrumentation-ioredis": "~0.64.0",
         "@opentelemetry/instrumentation-mongodb": "~0.69.0",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/arsenal#8.5.12",
+        "arsenal": "git+https://github.com/scality/arsenal#6675ce3895cb779b0d07756a373762c82ff5e335",
```
SylvainSenechal: todo

lib/routes/routeBackbeat.js
```diff
@@ -1064,7 +1070,25 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                     });
                 },
             ],
-            callback,
+            (err, results) => {
+                if (err?.is?.PreconditionFailed) {
```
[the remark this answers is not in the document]
SylvainSenechal: fair enough 🧐

delthas (approved):
nit: Can you squash your 2 last commits into the main Mael commit? So that we get proper atomic commits in history.

lib/routes/routeBackbeat.js
```diff
@@ -606,6 +606,9 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
     // Note: '' is falsy so the presence check must be !== undefined, not truthy.
     const hasMicroVersionId = encodedMicroVersionId !== undefined;
     let incomingMicroVersionId = null;
+    const mongoOptions = {
```
francoisferrand: this cannot be **mongo**Options : as the backend can be mongo or metadata (or memory) → `mdOptions` or `metadataOptions` maybe ?

lib/routes/routeBackbeat.js
```diff
@@ -659,6 +662,13 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
                 conflictErr.mvId,
             );
         }
+
+        // Atomic counterpart of the JS pre-checks above, performed at database level.
+        // Only effective when the stored revision already carries a
+        // microVersionId, need $exists and $or implementation on Arsenal for the rest
```
francoisferrand: should be implemented first, so we fix once in cloudserver?

SylvainSenechal: Yes, added in arsenal

SylvainSenechal:
will rerun prettier later

SylvainSenechal:
@francoisferrand its approved but you can retake a look if you want, most of the logic is happening in arsenal anyways

lib/routes/routeBackbeat.js
```diff
@@ -791,25 +801,21 @@ function putMetadata(request, response, bucketInfo, objMd, log, callback) {
             omVal.replicationInfo.isNFS = !omVal.replicationInfo.isReplica;
         }
```
francoisferrand, line 803: nit: could we not have kept `options` _**here**_?

less changes, more locality... and either way we need to "share" a variable from the main function to the continuation callback, so moving does not really help...

```
		const options = {
            overheadField: constants.overheadField,
        };

		// Atomic counterpart of the JS pre-checks above, performed at database level.
        if (incomingMicroVersionId !== null) {
            options.conditions = {
                $or: [{ microVersionId: { $exists: false } }, { microVersionId: { $gt: incomingMicroVersionId } }],
            };
        }
```

francoisferrand (approved):

lib/routes/routeBackbeat.js
SylvainSenechal: its a bit weird to have re add an isolated 
if (incomingMicroVersionId !== null) 
later in the code when we already have it at the top of the function 🤔

SylvainSenechal (approved):

lib/routes/routeBackbeat.js
francoisferrand: to me feels to me much weirder (and harder to read) to spread the construction of the options...
`incomingMicroVersionId` is kind of like a parameter, so no problem using it throughout the function

PR #6243, opened 2026-07-31
CLDSRV-943: remove unused integrityChecks config option

leif-scality:
`integrityChecks` allowed us to disable integrity checks for an specific endpoint, this is no longer something we want, nobody uses the flag, and it is also broken today. For all this reasons we are removing it.


fredmnl (approved):

dvasilas (approved):

leif-scality:
ping

PR #6247, opened 2026-08-04
CLDSRV-968: Expire idle keep-alive sockets in the functional test client

tcarmet:
The functional test client keeps pooled keep-alive connections indefinitely, so after an idle gap it can send a request over a connection the server has already closed and fail with a spurious connection reset. Because the client also runs with retries disabled, one reset is enough to fail a hook and take a whole group of tests down with it, which surfaces as intermittent CI red on slower full-deployment environments where such idle gaps actually occur.

This makes the client expire its own idle connections before the server does, so the reuse race cannot arise.

### DIFF
--- a/tests/functional/aws-node-sdk/test/support/config.js
+++ b/tests/functional/aws-node-sdk/test/support/config.js
@@ -30,6 +30,14 @@ const DEFAULT_GLOBAL_OPTIONS = {
 // timeout.
 const REQUEST_TIMEOUT = 30000;
 
+// Idle timeout for pooled keep-alive sockets, kept below the server's 5s
+// keep-alive so the client retires a socket before the server closes it.
+// A value is required: Node clamps this against the server's advertised
+// `Keep-Alive` hint, but only ever downwards from a non-zero timeout
+// (`agentTimeout = this.options.timeout || 0`), so leaving it unset means
+// pooled sockets never expire. See CLDSRV-968.
+const AGENT_IDLE_TIMEOUT = 4000;
+
 const DEFAULT_MEM_OPTIONS = {
     endpoint: `${transport}://127.0.0.1:8000`,
     port: 8000,
@@ -43,6 +51,7 @@ const DEFAULT_MEM_OPTIONS = {
             maxSockets: 200,
             keepAlive: true,
             keepAliveMsecs: 1000,
+            timeout: AGENT_IDLE_TIMEOUT,
         }),
     }),
 };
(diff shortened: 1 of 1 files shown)


leif-scality (approved):

anurag4DSB (approved):

tests/functional/aws-node-sdk/test/support/config.js
```diff
@@ -43,6 +55,7 @@ const DEFAULT_MEM_OPTIONS = {
             maxSockets: 200,
             keepAlive: true,
             keepAliveMsecs: 1000,
+            timeout: AGENT_IDLE_TIMEOUT,
```
jonathan-gramain, line 54: FWIW, nodejs 22.20 has a new option `agentKeepAliveTimeoutBuffer` of 1 second by default, which would naturally have the same effect here but in a more perennial way (https://nodejs.org/docs/latest/api/http.html#new-agentoptions). Currently the node version is pinned to 22.14 so it's not present, I think the proposed fix is good in the meantime but we could consider reverting it in the future (maybe worth a TODO)

tcarmet: Oh nice! Definitely worth a todo yeah. I'll create a ticket for nodejs upgrade add a todo in the comments to remove this config when tackling it.

tests/functional/aws-node-sdk/test/support/config.js
```diff
@@ -30,6 +30,18 @@ const DEFAULT_GLOBAL_OPTIONS = {
 // timeout.
 const REQUEST_TIMEOUT = 30000;
 
+// Idle timeout for pooled keep-alive sockets, which must stay below the
```
BourgoisMickael: This comment could be more succinct

BourgoisMickael (approved):

tests/functional/aws-node-sdk/test/support/config.js
tcarmet: I did some tests to review this and since this runs on the client side, we're already running on `nodejs > 22.20`, we have `22.23.1` here on cloudserver CI and integration have `22.22.0` where we faced the issue. I also did a test run with a quick scenario to reproduce a similar situation with node and confirm that it only clamps when a timeout is set, with no timeout it stays at 0.

Upgrading node won't make any change so I won't add any mention.

tests/functional/aws-node-sdk/test/support/config.js
tcarmet: done

tests/functional/aws-node-sdk/test/support/config.js
jonathan-gramain: OK, I assume this is because the server doesn't advertise its timeout value. Could be a nice addition in a future improvement maybe (no need to do anything now).

PR #6248, opened 2026-08-06
CLDSRV-965: checksums feature flag

leif-scality:
- Add checksum toggle
- Enabled by default
- When set to false, checksums are no longer handled (no header is verified, no checksum is computed, no checksum is compared, no checksum is stored)
- `Content-MD5` et `x-amz-checksum-sha256` not impacted by the flag


dvasilas (approved):
As discussed, there are cases when `completeMultipartUpload` could permanently fail because parts have been enabled with checksums disabled (during upgrades, if checksums are disabled and then re-enabled).
Let's create a documentation ticket to document these cases and how they should be resolved.

leif-scality:
> As discussed, there are cases when `completeMultipartUpload` could permanently fail because parts have been enabled with checksums disabled (during upgrades, if checksums are disabled and then re-enabled). Let's create a documentation ticket to document these cases and how they should be resolved.

I think I will add an error log `"unable to calculate final object checksum, part X has no stored checksum. Object will be created with no checksum"` and accept the request. This case could also occur during the upgrades so I think the accept plus log is a better solution

dvasilas:
> > As discussed, there are cases when `completeMultipartUpload` could permanently fail because parts have been enabled with checksums disabled (during upgrades, if checksums are disabled and then re-enabled). Let's create a documentation ticket to document these cases and how they should be resolved.
> 
> I think I will add an error log `"unable to calculate final object checksum, part X has no stored checksum. Object will be created with no checksum"` and accept the request. This case could also occur during the upgrades so I think the accept plus log is a better solution

Makes sense :+1: . That's a better choice than requiring a manual intervention for each problematic MPU.
We should maybe still document the checksum coverage gap in those cases.

.github/workflows/tests.yaml
```diff
@@ -558,6 +558,55 @@ jobs:
...
+      S3BACKEND: mem
+      S3VAULT: mem
+      CLOUDSERVER_IMAGE: ghcr.io/${{ github.repository }}:${{ github.sha }}-testcoverage
+      MONGODB_IMAGE: ghcr.io/${{ github.repository }}/ci-mongodb:${{ github.sha }}
+      MPU_TESTING: 'yes'
+      S3_INTEGRITY_CHECKS_ENABLED: 'false'
+      JOB_NAME: ${{ github.job }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
SylvainSenechal, line 577: nit: checkout v4 is getting old 🧐 v7 available

lib/api/apiUtils/integrity/validateChecksums.js
```diff
@@ -558,6 +577,13 @@ function arsenalErrorFromChecksumError(err) {
             return errMPUTypeInvalid;
         case ChecksumError.MPUTypeWithoutAlgo:
             return errMPUTypeWithoutAlgo;
+        case ChecksumError.MPUTypeNotConfigured:
+            return errMPUTypeNotConfigured;
+        case ChecksumError.MPUTypeModeMismatch:
+            return errorInstances.InvalidRequest.customizeDescription(
```
SylvainSenechal, line 583: nit: a bit weird that you bothered defining errMPUTypeNotConfigured at the top of the file, but not this error 🤔 If it's because you need the details.type, you can still define it at the top with a function that take a param and return the error

SylvainSenechal: nvm its already whats done for other custom error 🤷

leif-scality: All errors have a cloudserver definition for unit tests, in this case `ChecksumError.MPUTypeNotConfigured` only maps to `errMPUTypeNotConfigured`, but sometimes `ChecksumError` share the same arsenal error.

The goal is to separate domain errors from XML arsenal / AWS S3 error messages

lib/Config.js
```diff
@@ -584,6 +584,58 @@ function parseServerAccessLogs(config) {
...
+ */
+function parseIntegrityChecks(config) {
+    const res = { enabled: true };
+
+    if (config && config.integrityChecks) {
+        assert(
+            typeof config.integrityChecks === 'object' && !Array.isArray(config.integrityChecks),
+            'bad config: integrityChecks must be an object',
+        );
+
```
SylvainSenechal, line 619: nit: isn't this kind of stuff preferably handled by the Joi library directly 🤔

SylvainSenechal: I think its doable with joi, with like joi.default({enabled: true}) or something

But honestly whatever, half the file is joi and half the file is manual check

SylvainSenechal (approved):

lib/Config.js
leif-scality: Yea, I followed the existing pattern

lib/Config.js
```diff
@@ -584,6 +584,58 @@ function parseServerAccessLogs(config) {
     return res;
 }
 
+/**
```
BourgoisMickael, line 587: This comment looks too verbose, it concise

BourgoisMickael (approved):

lib/Config.js
leif-scality: It is verbose but it documents what the feature flag does and doesn't do. There is no useless info in it

leif-scality:
ping

PR #6251, opened 2026-08-12
Read the quota staleness window from the correct config key

delthas:
`lib/Config.js` read the quota staleness window from `quota.maxStatenessMS`
instead of `quota.maxStalenessMS`, so a value set in `config.json` was
silently ignored and the hardcoded 24h default was used instead. Only the
`QUOTA_MAX_STALENESS_MS` environment variable could change the window.

The setting is now validated like the options around it: apply the default,
then assert that an explicitly provided value is a positive integer rather
than silently falling back on a malformed one. The default moves to
`constants.js` as `defaultQuotaMaxStalenessMS`.

The existing test could not catch this. The fixture set the correctly
spelled key to `86400000`, which equals the default, so the assertion passed
through the fallback branch either way. It now uses a distinct value and
asserts that value differs from the default.

Issue: CLDSRV-970

### DIFF
--- a/tests/unit/Config.js
+++ b/tests/unit/Config.js
@@ -13,6 +13,7 @@ const {
 
 const { LOCATION_NAME_DMF } = require('../constants');
 const constants = require('../../constants');
+const allOptsQuota = require('./testConfigs/allOptsConfig/config.json').quota;
 
 const { ValidLifecycleRules: supportedLifecycleRules } = require('arsenal').models;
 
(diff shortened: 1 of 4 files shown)


tests/unit/Config.js
```diff
@@ -545,16 +546,31 @@ describe('Config', () => {
             process.env.S3_CONFIG_FILE = oldConfig;
         });
 
-        it('should set up quota', () => {
+        it('should set up quota from the config file', () => {
+            // The fixture value must differ from the default, otherwise this test would
```
SylvainSenechal: seems like comment from claude thinking about ongoing issue to fix, maybe not useful to keep

SylvainSenechal (approved):

tests/unit/Config.js
```diff
@@ -535,6 +535,7 @@ describe('Config', () => {
 
     describe('quota option setup', () => {
         let oldConfig;
+        const allOptsQuota = require('./testConfigs/allOptsConfig/config.json').quota;
```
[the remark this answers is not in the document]
delthas: Moved to the top of the file, alongside the other requires (after `const constants = require('../../constants')`). The relative path is unchanged since `require` resolves relative to the file, not the block.

Amended into the existing commit and force-pushed, so the branch is now at `564393413`.

tests/unit/Config.js
delthas: Dropped

delthas:
Requested @DarkIsDude in place of Maha Benzekri, who is currently on PTO.

DarkIsDude (approved):

delthas:
!done 2h

delthas:
[`CLDSRV-970`](https://scality.atlassian.net/browse/CLDSRV-970): logged `2h`; transitioned to Done (`Fixed`)

PR #6252, opened 2026-08-12
Label quota metrics retrieval with the real status code

delthas:
`lib/utilization/scuba/wrapper.js` derived the Prometheus `code` label from
`err.statusCode`. scubaclient does not wrap errors, so axios failures arrive
with the status on `err.response.status` and nothing on `err.statusCode`, and
the label was always `500`. A scuba 404 was indistinguishable from a genuine
5xx or from a connection failure.

The label is now derived the way `lib/routes/veeam/utils.js` already does it:

    code = err.response?.status || err.statusCode || err.code || 500;

Transport failures surface their errno (`ECONNREFUSED`, `ETIMEDOUT`) rather
than being reported as 500, so the label is no longer strictly numeric.

Issue: CLDSRV-971

### DIFF
--- a/lib/utilization/scuba/wrapper.js
+++ b/lib/utilization/scuba/wrapper.js
@@ -71,7 +71,7 @@ class ScubaClientImpl extends ScubaClient {
         try {
             return await super.getLatestMetrics(metricsClass, resourceName, mergedOptions, body);
         } catch (err) {
-            code = err.statusCode || 500;
+            code = err.response?.status || err.code || 500;
             throw err;
         } finally {
             const responseTimeInNs = Number(process.hrtime.bigint() - requestStartTime);
--- a/tests/unit/quotas/scuba/wrapper.js
+++ b/tests/unit/quotas/scuba/wrapper.js
@@ -1,6 +1,7 @@
 const assert = require('assert');
 const sinon = require('sinon');
 const { ScubaClientImpl } = require('../../../../lib/utilization/scuba/wrapper');
+const monitoring = require('../../../../lib/utilities/monitoringHandler');
 const { default: ScubaClient } = require('scubaclient');
 
 describe('ScubaClientImpl', () => {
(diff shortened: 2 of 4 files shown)


lib/utilization/scuba/wrapper.js
```diff
@@ -71,7 +71,9 @@ class ScubaClientImpl extends ScubaClient {
         try {
             return await super.getLatestMetrics(metricsClass, resourceName, mergedOptions, body);
         } catch (err) {
-            code = err.statusCode || 500;
+            // scubaclient does not wrap errors, so axios failures carry the status
+            // on err.response.status, and transport failures only have err.code.
+            code = err.response?.status || err.statusCode || err.code || 500;
```
DarkIsDude: That's a pretty standard axios code, IMO the comment is not useful

DarkIsDude (approved):

lib/utilization/scuba/wrapper.js
delthas: Dropped

lib/utilization/scuba/wrapper.js
```diff
@@ -71,7 +71,7 @@ class ScubaClientImpl extends ScubaClient {
         try {
             return await super.getLatestMetrics(metricsClass, resourceName, mergedOptions, body);
         } catch (err) {
-            code = err.statusCode || 500;
+            code = err.response?.status || err.statusCode || err.code || 500;
```
francoisferrand: if `err.response?.status` is axios error code, do we really need the fallback to `err.statusCode || err.code` ?

tests/unit/quotas/scuba/wrapper.js
```diff
@@ -139,5 +140,63 @@ describe('ScubaClientImpl', () => {
             assert.strictEqual(forwardedOptions.headers['X-Existing'], 'v');
             assert.strictEqual(forwardedOptions.headers['X-Scal-Request-Uids'], 'req1:req2');
         });
+
+        describe('retrieval duration metric', () => {
```
francoisferrand, line 144: these tests are Lind of artificial : they are purely unit test, but do not help to be sure the code works...
e.g. if actually the error were wrapped, if another lib (not axios was used), ... (and they validate all kinds of error, even if some were actually useless...)

Would be best to handle this in functional tests I guess...

francoisferrand (approved):

lib/utilization/scuba/wrapper.js
delthas: Good catch on `err.statusCode` — dropped it. Nothing in this path sets it, which is exactly why the label was always 500: the old `err.statusCode || 500` never once took the first branch.

I checked against the axios that actually produces these errors. Note scubaclient bundles its **own** axios (1.13.4) rather than using cloudserver's root 0.18.1:

| case | `err.response.status` | `err.statusCode` | `err.code` |
|---|---|---|---|
| 404 | `404` | `undefined` | `ERR_BAD_REQUEST` |
| connection refused | – | `undefined` | `ECONNREFUSED` |

So it is now:

```js
code = err.response?.status || err.code || 500;
```

`err.code` still has to stay: it is the only signal when there is no response, and telling a connection failure apart from a scuba 5xx is the point of the change. Worth noting the ordering is load-bearing rather than cosmetic — axios sets `code: 'ERR_BAD_REQUEST'` on a 404, so putting `err.code` first would label every 4xx `ERR_BAD_REQUEST`. There is now a unit test pinning that precedence.

Note this makes the line diverge from `lib/routes/veeam/utils.js:272`, which still carries the same dead `err.statusCode` fallback. Happy to clean that up too, here or separately, if you'd prefer them consistent.

tests/unit/quotas/scuba/wrapper.js
delthas: Moved to a functional test.

<details>
Fair — they asserted my assumption about the error shape instead of verifying it. Added functional coverage in `tests/sur/quota.js`: the scuba mock can now fail metrics with a given status, and the test asserts the histogram records a `code="404"` observation while the `code="500"` count stays put (the actual regression), through real HTTP / scubaclient / axios — so wrapping would fail it.

Kept the transport case as a unit test on purpose: stopping the mock lets the health check disable retrieval within ~100ms (`SCUBA_HEALTHCHECK_FREQUENCY: 100`), so it would be racy. The Veeam path is the deterministic spot if you want that covered too.

Also dropped the dead `statusCode` case, and the 404 case now sets `ERR_BAD_REQUEST` so it proves precedence rather than assuming it.
</details>

delthas:
!done 3h

delthas:
[`CLDSRV-971`](https://scality.atlassian.net/browse/CLDSRV-971): logged `3h`; transitioned to Done (`Fixed`)

PR #6253, opened 2026-08-12
Register quota metrics unconditionally

delthas:
Stacked on #6252 — review that one first. This PR's diff is only the last two
commits; retarget to `development/9.4` once #6252 merges.

Both quota gates in `lib/utilities/monitoringHandler.js` tested
`config.isQuotaEnabled` — the function reference rather than a call — so both
were always true, and all six quota metrics have always been registered.

Rather than calling the function, the conditionals are removed. Always
registering is what has actually happened since these checks were written,
nothing depends on the metrics disappearing when quota is off, and gating
registration would mean guarding every metric touch point against a metric that
was never created: `crrCacheToProm` reached `bucketsWithQuota` unguarded, and
the scuba wrapper observes its histogram in a `finally` block, on a path the
Veeam capacity route reaches with no quota gating at all.

Behaviour is unchanged — the six `s3_cloudserver_quota_*` series are exposed
exactly as before — so this is a dead code removal rather than a metric surface
change.

Issue: CLDSRV-972

### DIFF
--- a/lib/utilities/monitoringHandler.js
+++ b/lib/utilities/monitoringHandler.js
@@ -65,48 +65,39 @@ const httpResponseSizeBytes = new client.Summary({
     help: 'Cloudserver HTTP response sizes in bytes',
 });
 
-let quotaEvaluationDuration;
-let utilizationMetricsRetrievalDuration;
-let utilizationServiceAvailable;
-let bucketsWithQuota;
-let accountsWithQuota;
-let requestWithQuotaMetricsUnavailable;
+const quotaEvaluationDuration = new client.Histogram({
+    name: 's3_cloudserver_quota_evaluation_duration_seconds',
+    help: 'Duration of the quota evaluation operation',
+    labelNames: ['action', 'code', 'type'],
+    buckets: [0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1],
+});
 
-if (config.isQuotaEnabled) {
-    quotaEvaluationDuration = new client.Histogram({
-        name: 's3_cloudserver_quota_evaluation_duration_seconds',
-        help: 'Duration of the quota evaluation operation',
-        labelNames: ['action', 'code', 'type'],
-        buckets: [0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1],
-    });
+const utilizationMetricsRetrievalDuration = new client.Histogram({
+    name: 's3_cloudserver_quota_metrics_retrieval_duration_seconds',
+    help: 'Duration of the utilization metrics retrieval operation',
+    labelNames: ['code', 'class'],
+    buckets: [0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5],
+});
 
-    utilizationMetricsRetrievalDuration = new client.Histogram({
-        name: 's3_cloudserver_quota_metrics_retrieval_duration_seconds',
-        help: 'Duration of the utilization metrics retrieval operation',
-        labelNames: ['code', 'class'],
-        buckets: [0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5],
-    });
+const utilizationServiceAvailable = new client.Gauge({
+    name: 's3_cloudserver_quota_utilization_service_available',
+    help: 'Availability of the utilization service',
+});
 
-    utilizationServiceAvailable = new client.Gauge({
-        name: 's3_cloudserver_quota_utilization_service_available',
-        help: 'Availability of the utilization service',
-    });
+const bucketsWithQuota = new client.Gauge({
+    name: 's3_cloudserver_quota_buckets_count',
+    help: 'Total number of buckets quota',
+});
 
-    bucketsWithQuota = new client.Gauge({
-        name: 's3_cloudserver_quota_buckets_count',
-        help: 'Total number of buckets quota',
-    });
+const accountsWithQuota = new client.Gauge({
+    name: 's3_cloudserver_quota_accounts_count',
+    help: 'Total number of account quota',
+});
 
-    accountsWithQuota = new client.Gauge({
-        name: 's3_cloudserver_quota_accounts_count',
-        help: 'Total number of account quota',
-    });
-
-    requestWithQuotaMetricsUnavailable = new client.Counter({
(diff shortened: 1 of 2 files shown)


lib/utilities/monitoringHandler.js
```diff
@@ -72,7 +72,7 @@ let bucketsWithQuota;
 let accountsWithQuota;
 let requestWithQuotaMetricsUnavailable;
 
-if (config.isQuotaEnabled) {
+if (config.isQuotaEnabled()) {
```
SylvainSenechal: lmao dealing with this in 2026 is madness

SylvainSenechal (approved):

lib/utilities/monitoringHandler.js
delthas: Haha yeah, glad we have LLM reviews for this now

delthas:
Requested @francoisferrand in place of Maha Benzekri, who is currently on PTO.

lib/utilities/monitoringHandler.js
```diff
@@ -187,7 +190,7 @@ function crrCacheToProm(crrResults) {
             numberOfBuckets.set(crrResults.getObjectCount.buckets || 0);
             numberOfObjects.set(crrResults.getObjectCount.objects || 0);
         }
-        if (config.isQuotaEnabled) {
+        if (config.isQuotaEnabled()) {
```
francoisferrand: - if we go the alternative route, should we always set the metrics (as was done before), or fix the condition (as you did) ? This not only creates the metric (which has no impact), but also (and more significantly) sets the value so the metric is scrapped...
- if we keep the current code, the condition would be better made on `bucketsWithQuota` and `accountsWithQuota` I think

```suggestion
        if (bucketsWithQuota) {
            bucketsWithQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
        }
        if (accountsWithQuota) {
            accountsWithQuota.set(crrResults?.getVaultReport?.accountWithQuotaCount || 0);
        }
```

lib/utilities/monitoringHandler.js
francoisferrand: do we actually need to make this conditional?

I see this was mentioned as an alternative in the ticket:

> Drop the dead conditional entirely and register the six metrics unconditionally. That preserves today's behaviour byte-for-byte, removes the misleading check, and carries no crash risk or metric-surface change — at the cost of exposing quota metrics when quota is off, which is already what happens today.

so I wonder why it was not rejected?
- it keeps the current behavior
- nobody complained about this behavior;
- it simplifies the code (no conditional, no ?. on member access)

francoisferrand (commented):
- I think _creating_ the metrics is harmless (it was already the case); and it avoids adding all the ?. → [just create them always](https://github.com/scality/cloudserver/pull/6253/changes#r3782840588)
- less sure how to fix [crrCacheToPom](https://github.com/scality/cloudserver/pull/6253/changes#r3782824783)

lib/utilities/monitoringHandler.js
delthas: Moved to always creating the metrics.

lib/utilities/monitoringHandler.js
delthas: Moved to always creating the metrics.

tests/unit/utils/monitoring.js
```diff
@@ -112,10 +135,40 @@ describe('Monitoring: endpoint', () => {
...
+        's3_cloudserver_quota_unavailable_count',
+    ];
+
+    it('should register quota metrics even when quota is disabled', async () => {
+        assert.strictEqual(config.isQuotaEnabled(), false);
+
+        await fetchMetrics({ method: 'GET', url: '/metrics' }, res);
+        const metrics = res.end.args[0][0];
+
+        quotaMetricNames.forEach(name => assert(metrics.includes(name), `${name} is not registered`));
```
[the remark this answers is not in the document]
delthas: Keeping `includes` here. I measured the failure output against a scrape-sized input (~6KB), and `assert.match` doesn't produce a diff — it dumps the whole input:

| form | failure output |
|---|---|
| `assert(includes)` | 113 chars (prints the expression) |
| `assert(includes, msg)` | 50 chars |
| `assert.match` | **7971 chars** (entire scrape) |
| `assert.match(…, msg)` | 50 chars |

A real `/metrics` body is bigger still. And since the suggested line 159 passes a message, the message replaces the diff, making it output-identical to what's already there; the snippet for 171–172 passes no message, so that one would be the 8KB dump.

`includes` is also this file's existing idiom (7+ instances, e.g. the `# TYPE` assertions above).

lib/utilities/monitoringHandler.js
```diff
@@ -187,10 +181,8 @@ function crrCacheToProm(crrResults) {
             numberOfBuckets.set(crrResults.getObjectCount.buckets || 0);
             numberOfObjects.set(crrResults.getObjectCount.objects || 0);
         }
-        if (config.isQuotaEnabled) {
-            bucketsWithQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
-            accountsWithQuota.set(crrResults?.getVaultReport?.accountWithQuotaCount || 0);
-        }
+        bucketsWithQuota.set(crrResults?.getObjectCount?.bucketWithQuotaCount || 0);
```
francoisferrand: nit:

the first ? is useless - we are in an `if (crrResults)` block
```suggestion
        bucketsWithQuota.set(crrResults.getObjectCount?.bucketWithQuotaCount || 0);
```

also this line maybe moved int he previous block, which already asserts crrResults.getObjectCount

francoisferrand (approved):
LGTM, but I think this warrants another set of eyes from product, to confirm we want/accept to have the metrics always (not just the presence of the metrics, but especially publishing actual _data points_ for the bucketsWithQuota/accountsWithQuota metrics)

lib/utilities/monitoringHandler.js
delthas: Done, and applied the same treatment to the accounts line so both are gated on their own source — no `?.` left in the block.

delthas:
Asking product.

delthas:
Product said both options are OK, let's do simplest. So, merging as is.

delthas:
!done 3h

delthas:
[`CLDSRV-972`](https://scality.atlassian.net/browse/CLDSRV-972): logged `3h`; transitioned to Done (`Fixed`)

PR #6258, opened 2026-08-19
CLDSRV-978: Stop installing tcpdump in the kmip cluster job

tcarmet:
The KMIP cluster functional test job was hanging for hours on installing tcpdump, because of an unreliable Azure-hosted Ubuntu mirror. tcpdump already ships with the runner image, so the install was never needed - dropping it removes our exposure to that mirror.

The step is commented out rather than deleted, so it can be restored quickly if a future runner or image change drops the package.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -877,8 +877,11 @@ jobs:
           python-version: 3.9
       - name: Setup CI environment
         uses: ./.github/actions/setup-ci
-      - name: Install tcpdump to analyze traffic on kmip cluster interfaces
-        run: sudo apt-get update && sudo apt-get install -y tcpdump
+      # tcpdump is already present on the runner image, and the apt mirrors are
+      # unreliable enough to hang a job for hours, so we don't install it.
+      # Uncomment if a runner or image change ever drops the package.
+      # - name: Install tcpdump to analyze traffic on kmip cluster interfaces
+      #   run: sudo apt-get install -y tcpdump
       - name: Copy KMIP certs
         run: cp -r ./certs /tmp/ssl-kmip
         working-directory: .github/pykmip


jonathan-gramain (approved):

.github/workflows/tests.yaml
```diff
@@ -878,7 +878,7 @@ jobs:
       - name: Setup CI environment
         uses: ./.github/actions/setup-ci
       - name: Install tcpdump to analyze traffic on kmip cluster interfaces
-        run: sudo apt-get update && sudo apt-get install -y tcpdump
+        run: sudo apt-get install -y tcpdump
```
BourgoisMickael: This might not fix the issue with `apt`.  It looks like the azure mirror used for `ubuntu-latest` is flaky and the issue can arise for any request (`update` or `install` of a package). (https://github.com/Frazzled-Productions/poke-memory/issues/644).

Because the connection to the server stays open, apt doesn't timeout and hangs for hours.
Maybe we should add a timeout on those steps:
```yaml
timeout-minutes: 5  # Forces GitHub to kill the step if it hangs
```

Or use an action to timeout + retry: https://github.com/nick-fields/retry.
Or use linux `timeout` with a custom retry loop in shell

tcarmet: I mean, it did fix the issue, but that's because tcpdump is already installed on those runners. So I was considering removing this package install.

But yes the issue remains cause those repos are down, but less we depend on it less are the chances something impact us.

I kept it here cause wanted to be safe in case a runner OS upgrade would remove tcpdump but what do you think? Should we remove it or do we tweak the step with a timeout and all?

BourgoisMickael: Well if it's already present you can remove by commenting the line with a description saying it's already present. 

So if we change the runner or the image loses it, we can bring it back quickly

tcarmet: sounds good, done

BourgoisMickael (approved):

PR #6265, opened 2026-08-31
Allow enabling versioning on buckets with a GCP location constraint

maeldonn:
### Not handled yet

| Gap | Impact | What it would take |
|---|---|---|
| S3 has a suspended versioning state, with `null` version ids. GCS versioning is only on or off | A GCP location cannot represent a suspended bucket | Decide whether to reject the configuration or emulate the state |
| Object versioning on the GCS bucket is a hard requirement: without it, the delete issued behind a delete marker destroys the data instead of archiving it | Silent data loss | Block writes on a non-versioned bucket, instead of only reporting it in the healthcheck |
| GCS lifecycle rules on noncurrent versions (`isLive: false`, `daysSinceNoncurrentTime`, `numNewerVersions`) can reap generations that S3 still serves as current | Deferred and silent data loss, with no error at write time | Document the precondition, and consider detecting such rules at configuration time |
| GCS never promotes a noncurrent generation, so the backend state diverges from the S3 view after an undelete or a version delete | Confusing during support, and the root cause of the lifecycle exposure above | Optional: promote by copy, accepting the new generation, the non-atomicity and the storage class reset |

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -153,7 +153,7 @@ const constants = {
     // for external backends, don't call unless at least 1 minute
     // (60,000 milliseconds) since last call
     externalBackendHealthCheckInterval: 60000,
-    versioningNotImplBackends: { azure: true, gcp: true },
+    versioningNotImplBackends: { azure: true },
     mpuMDStoredExternallyBackend: { aws_s3: true, gcp: true },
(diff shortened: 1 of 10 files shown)


maeldonn:
Replaced with #6266

PR #6266, opened 2026-08-31
Allow enabling versioning on buckets with a GCP location constraint

maeldonn:
### Not handled yet

| Gap | Impact | What it would take |
|---|---|---|
| Object versioning on the GCS bucket is a hard requirement: without it, the delete issued behind a delete marker destroys the data instead of archiving it | Silent data loss | Block writes on a non-versioned bucket, instead of only reporting it in the healthcheck |
| GCS lifecycle rules on noncurrent versions (`isLive: false`, `daysSinceNoncurrentTime`, `numNewerVersions`) can reap generations that S3 still serves as current | Deferred and silent data loss, with no error at write time | Document the precondition, and consider detecting such rules at configuration time |
| GCS never promotes a noncurrent generation, so the backend state diverges from the S3 view after an undelete or a version delete | Confusing during support, and the root cause of the lifecycle exposure above | Optional: promote by copy, accepting the new generation, the non-atomicity and the storage class reset |

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -22,10 +22,7 @@ const {
     areChecksumsEnabled,
 } = require('../integrity/validateChecksums');
 
-const { externalBackends, versioningNotImplBackends } = constants;
-
-const externalVersioningErrorMessage =
-    'We do not currently support putting a versioned object to a location-constraint of type Azure or GCP.';
+const { externalBackends, versioningNotImplBackends, externalVersioningErrorMessage } = constants;
 
 /**
  * Validate and compute the checksum for a zero-size object body.
--- a/lib/api/bucketPutVersioning.js
+++ b/lib/api/bucketPutVersioning.js
@@ -6,23 +6,23 @@ const collectCorsHeaders = require('../utilities/collectCorsHeaders');
 const metadata = require('../metadata/wrapper');
 const { standardMetadataValidateBucket } = require('../metadata/metadataUtils');
 const { pushMetric } = require('../utapi/utilities');
-const versioningNotImplBackends =
-    require('../../constants').versioningNotImplBackends;
+const versioningNotImplBackends = require('../../constants').versioningNotImplBackends;
 const { config } = require('../Config');
 const monitoring = require('../utilities/monitoringHandler');
 
-const externalVersioningErrorMessage = 'We do not currently support putting ' +
-'a versioned object to a location-constraint of type Azure or GCP.';
+const externalVersioningErrorMessage =
+    'We do not currently support putting ' + 'a versioned object to a location-constraint of type Azure or GCP.';
 
-const replicationVersioningErrorMessage = 'A replication configuration is ' +
-'present on this bucket, so you cannot change the versioning state. To ' +
-'change the versioning state, first delete the replication configuration.';
+const replicationVersioningErrorMessage =
+    'A replication configuration is ' +
+    'present on this bucket, so you cannot change the versioning state. To ' +
+    'change the versioning state, first delete the replication configuration.';
 
-const ingestionVersioningErrorMessage = 'Versioning cannot be suspended for '
-+ 'buckets setup with Out of Band updates from a location';
+const ingestionVersioningErrorMessage =
+    'Versioning cannot be suspended for ' + 'buckets setup with Out of Band updates from a location';
 
-const objectLockErrorMessage = 'An Object Lock configuration is present on ' +
-  'this bucket, so the versioning state cannot be changed.';
+const objectLockErrorMessage =
+    'An Object Lock configuration is present on ' + 'this bucket, so the versioning state cannot be changed.';
 
 /**
  * Format of xml request:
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
         "@opentelemetry/instrumentation-ioredis": "~0.64.0",
         "@opentelemetry/instrumentation-mongodb": "~0.69.0",
         "@smithy/node-http-handler": "^3.0.0",
-        "arsenal": "git+https://github.com/scality/arsenal#8.5.15",
+        "arsenal": "git+https://github.com/scality/arsenal#2136c02ff1d3dfd43a5ff7ca59f955d105946dba",
```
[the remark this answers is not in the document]
maeldonn: Arsenal is not yet merged

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -322,7 +322,10 @@ function createAndStoreObject(
                     }
                 }
 
-                const headerChecksum = areChecksumsEnabled() ? getChecksumDataFromHeaders(request.headers) : null;
+                // A delete marker has no body: an x-amz-checksum-* header
+                // describes the request payload (e.g. a multi-object delete XML).
+                const headerChecksum =
+                    isDeleteMarker || !areChecksumsEnabled() ? null : getChecksumDataFromHeaders(request.headers);
```
francoisferrand: Is this not an issue for other data locations as well? Ie a separate bugfix for integrity check feature?

tests/unit/api/bucketPutVersioning.js
francoisferrand, line 1: no change here AFAICS, only removal of " or GCP" in the `externalVersioningErrorMessage` const?

tests/unit/multipleBackend/VersioningBackendClient.js
```diff
@@ -7,8 +7,7 @@ const DummyService = require('../DummyService');
 const { DummyRequestLogger } = require('../helpers');
 
 const missingVerIdInternalError = errorInstances.InternalError.customizeDescription(
-    'Invalid state. Please ensure versioning is enabled ' +
-    'in AWS for the location constraint and try again.'
+    'Invalid state. Please ensure versioning is enabled ' + 'in AWS for the location constraint and try again.',
```
francoisferrand:
```suggestion
    'Invalid state. Please ensure versioning is enabled in AWS for the location constraint and try again.',
```

francoisferrand: same for a few below

lib/api/bucketPutVersioning.js
francoisferrand, line 1: - the GPC removal should not go in 9.4, but 9.5
- I wonder if we should build the strings, instead of having the source of truth in 2 constants (one in arsenal + one in cloudserver) AND having manually duplicated that info in multiple error strings...

francoisferrand (commented):
> S3 has a suspended versioning state, with nullversion ids. GCS versioning is only on or off

Suspended versioning is really just a name : in our implementation it differs (so in particular performance may not be the same), but functionally suspended is about the same as off?
i.e. is suspended not just a different name for no versioning (no versioning → versioning ←→ suspended versioning), so we could map both no & suspended versioning on GCP's disabled versioning?

> Object versioning on the GCS bucket is a hard requirement: without it, the delete issued behind a delete marker destroys the data instead of archiving it

- not sure what you mean, there is no "archiving" in S3...
- if using GCP versioning is not practical, another option would be to "flatten" the versions: i.e. store each version side-by-side, as `<objKey>.<scalVersionId>`

> Block writes on a non-versioned bucket, instead of only reporting it in the healthcheck

If there a GCP "extension" to make the putObject/deleteObject conditional on versioning? Otherwise this may be very costly....

> GCS lifecycle rules on noncurrent versions (isLive: false, daysSinceNoncurrentTime, numNewerVersions) can reap generations that S3 still serves as current

- Should be part of the documentation, and may be verified when creating location and/or enabling versioning (though none of these are foolproof, and we still risk a data loss)
- Or a reason to switch to "emulating" versions by flattening them....

> GCS never promotes a noncurrent generation, so the backend state diverges from the S3 view after an undelete or a version delete

That is not really a problem _for us_, right: The issue is more about what the user would see in GCP's bucket?

All in all, I don't think we should merge this -and esp. in 9.4- but take the time to make a (short) design in Citadel : so we can align on the requirements and limits of the solution (i.e. do we want to keep the same "view" on GCP, or is this really just data storage, accessed only through the product ? what to do with GCP-side lifecycle rules ? ...)

tests/unit/multipleBackend/VersioningBackendClient.js
maeldonn: Linked to prettier... Fixed.

lib/api/apiUtils/object/createAndStoreObject.js
maeldonn: Not only for GCP, moved to a separate PR.

lib/api/bucketPutVersioning.js
maeldonn: Moved to 9.5.
Tried to deduplicate and keep it simple.

maeldonn:
> but functionally suspended is about the same as off?

Checked and you're right, nothing to do. I removed the line from the table.

> there is no "archiving" in S3

GCS calls noncurrent versions "archived" (no delete markers, no promotion back to live, and lifecycle rules can target them via `isLive: false` / `daysSinceNoncurrentTime` / `numNewerVersions`)

> another option would be to "flatten" the versions

Possible, and it doesn't even need the S3 version id: the backend key isn't recomputed on read, put returns it and it's stored per-version in location[0].key, so any unique suffix minted at put time works. Objects already written keep resolving at their plain key, so no migration either.

The cost is elsewhere: every path that recomputes the key via _createAwsKey instead of reading it from MD (whole MPU lifecycle, copyObject/uploadPartCopy, and toObjectGetInfo which the backbeat routes use with only objectKey + bucket), plus _shouldSkipDelete, which skips the old-version delete assuming overwrites are in-place. Real work, but a design-doc call rather than a blocker.

> If there a GCP "extension" to make the putObject/deleteObject conditional on versioning?

No, GCS preconditions cover generation/metageneration only, nothing on versioning state. We already fail after the fact (no x-goog-generation -> missingVerIdInternalError), but too late, and delete markers are exempted.

> Should be part of the documentation, and may be verified when creating location and/or enabling versioning

Agreed on documenting it. Verifying is doable but not free. We talk to GCS through the S3 interop API, so the AWS SDK is what sends our requests. It can call `?lifecycle`, but GCS answers with its own XML format, which the SDK doesn't know how to read, so we'd have to parse that response ourselves. It's the same kind of adaptation code we already have in `GcpApis/*` for the other places where GCS and S3 differ. And checking only at location creation wouldn't help much, since rules can be added later and we'd never notice. Better to read it in the healthcheck we already run every 60s, together with the versioning state.

> The issue is more about what the user would see in GCP's bucket?

Right, not a problem while the bucket is only used as a Scality backend: we always read a specific generation, never the live one. It only bites if the bucket is used without us in front (direct reads, OOB, and GCS lifecycle above all). So we need to pin down the supported usage and document it. Same drift on the AWS backend, so it's the mirrored-bucket model rather than GCS-specific.

PR #6267, opened 2026-09-01
Ignore the request checksum when creating a delete marker 

maeldonn:
A delete marker has no body, but createAndStoreObject took its checksum
from the request headers, which on a multi-object delete cover the request
XML. On an external location the mismatch failed the put, and the rollback
error then hit an unguarded result.versionId in multiObjectDelete, killing
the worker with a 502. Skip the header checksum for delete markers, and
guard the result so any later error is returned as a proper S3 error.

Issue: [CLDSRV-984](https://scality.atlassian.net/browse/CLDSRV-984)

[CLDSRV-984]: https://scality.atlassian.net/browse/CLDSRV-984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/api/apiUtils/object/createAndStoreObject.js
+++ b/lib/api/apiUtils/object/createAndStoreObject.js
@@ -322,14 +322,17 @@ function createAndStoreObject(
                     }
                 }
 
-                const headerChecksum = areChecksumsEnabled() ? getChecksumDataFromHeaders(request.headers) : null;
-                if (headerChecksum && headerChecksum.error) {
-                    return next(arsenalErrorFromChecksumError(headerChecksum));
+                // A delete marker has no body and no checksum: an
+                // x-amz-checksum-* header describes the request payload
+                // (e.g. a multi-object delete XML).
+                let checksums = { primary: null, secondary: null };
+                if (!isDeleteMarker) {
+                    const headerChecksum = areChecksumsEnabled() ? getChecksumDataFromHeaders(request.headers) : null;
+                    if (headerChecksum && headerChecksum.error) {
+                        return next(arsenalErrorFromChecksumError(headerChecksum));
+                    }
+                    checksums = { primary: headerChecksum || defaultChecksumData, secondary: null };
                 }
-                const checksums = {
-                    primary: headerChecksum || defaultChecksumData,
-                    secondary: null,
-                };
                 return dataStore(
                     objectKeyContext,
                     cipherBundle,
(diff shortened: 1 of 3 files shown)


lib/api/apiUtils/object/createAndStoreObject.js
leif-scality: If think it is better to do, so we dont calculate the defaultChecksum and may get weird results
```js
let checksums = {
                    primary: null,
                    secondary: null,
};
if (!isDeleteMarker) {
                const headerChecksum = areChecksumsEnabled() ? getChecksumDataFromHeaders(request.headers) : null;
                if (headerChecksum && headerChecksum.error) {
                    return next(arsenalErrorFromChecksumError(headerChecksum));
                }
                checksums = {
                    primary: headerChecksum || defaultChecksumData,
                    secondary: null,
                };
}
```

leif-scality, line 42: Does this need to be skipped also?

maeldonn: Already skipped by an [early return](https://github.com/scality/cloudserver/blob/0d7205f8acc6feea8b75811ef7b29890c105c435/lib/api/apiUtils/object/createAndStoreObject.js#L290-L292).

maeldonn: Updated.

leif-scality (approved):

lib/api/apiUtils/object/createAndStoreObject.js
```diff
@@ -322,14 +322,19 @@ function createAndStoreObject(
                     }
                 }
 
-                const headerChecksum = areChecksumsEnabled() ? getChecksumDataFromHeaders(request.headers) : null;
-                if (headerChecksum && headerChecksum.error) {
-                    return next(arsenalErrorFromChecksumError(headerChecksum));
+                // A delete marker has no body and no checksum: an
```
DarkIsDude, line 325: This function is very long, I would try to extract logic to a new function. It's nits but this function are a nightmare to debug and follow.

DarkIsDude (approved):
A "nits" for you

lib/api/apiUtils/object/createAndStoreObject.js
maeldonn: Extracting this logic into a new function might not be the right approach. This function needs a larger refactor that is out of scope for this PR.

PR #6268, opened 2026-09-02
CLDSRV-986: Pass overhead fields when abortMPU cleans up after completeMPU

dvasilas:
When `AbortMultipartUpload` finds an object created by an MPU whose upload ID is still addressable, it cleans up the created object.

It issues the `deleteObjectMD` request to delete the object without passing the `overheadField` parameter, which instructs Metadata to copy the overhead fields into the oplog entry.
So the resulting oplog entry has an `overhead` that includes only the `originOp` field (does not include the `owner id` field required by SUR).

This state can arise from `CompleteMultipartUpload`'s ordering.
The shadow bucket holds one overview key per MPU. That key is what makes an upload ID addressable (`metadataValidateMultipart` fetches it and returns `NoSuchUpload` when it is missing).
`CompleteMultipartUpload` writes the resulting object first, and then deletes the MPU part keys and the overview key. 

There are two cases in which an `AbortMultipartUpload` finds both the overview key in the shadow bucket and the object created by the MPU, and cleans up the object:
- The shadow bucket cleanup fails, and the overview key stays indefinitely.
- An `AbortMultipartUpload` arrives in the window where the resulting object has been created but the MPU state has not yet been cleaned up from the shadow bucket.

SUR cannot count an entry with no owner ID. It asserts on the existence of the owner ID field in `overhead`.
An oplog entry without the owner ID field completely blocks the SUR ingestor for the corresponding raft session.

### Testing

A test that reproduces this scenario already exists: `'should delete the correct orphaned object version'` in `tests/unit/api/apiUtils/object/abortMultipartUpload.spec.js`

### DIFF
--- a/lib/api/apiUtils/object/abortMultipartUpload.js
+++ b/lib/api/apiUtils/object/abortMultipartUpload.js
@@ -10,8 +10,7 @@ const metadata = require('../../../metadata/wrapper');
 const { preprocessingVersioningDelete } = require('./versioning');
 const { config } = require('../../../Config');
 
-function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log,
-    callback, request) {
+function abortMultipartUpload(authInfo, bucketName, objectKey, uploadId, log, callback, request) {
     const metadataValMPUparams = {
         authInfo,
         bucketName,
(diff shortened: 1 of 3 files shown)


fredmnl (approved):

anurag4DSB (approved):

dvasilas:
:robot:

jonathan-gramain (approved):

PR #6274, opened 2026-09-07
Allow partialFilterExpression in backbeat index add payload

maeldonn:
Since BB-773 (backbeat >= 9.4.1) the lifecycle conductor sends its v2 index
specs with a `partialFilterExpression`, rejected as an unknown field by the
Joi schema of `POST /_/backbeat/index?operation=add`. The v2 lifecycle
indexes were never created and buckets fell back to v1 listings (BB-866).

Declare the field as an optional object (the rest of the chain already
forwarded it to MongoDB) and add route-level tests.

Issue: CLDSRV-991

### DIFF
--- a/lib/routes/routeBackbeat.js
+++ b/lib/routes/routeBackbeat.js
@@ -1786,6 +1786,7 @@ const indexEntrySchema = joi.object({
             }),
         )
         .required(),
+    partialFilterExpression: joi.object(),
 });
 
 const indexingSchema = joi.array().items(indexEntrySchema).min(1);
--- a/tests/unit/routes/routeBackbeat.js
+++ b/tests/unit/routes/routeBackbeat.js
@@ -1540,3 +1540,117 @@ describe('routeBackbeat authorization', () => {
         });
     });
 });
+
+describe('routeBackbeat index add payload validation', () => {
+    const bucketName = 'bucketname';
+    let endPromise;
+    let resolveEnd;
+    let response;
+
+    function makeIndexRequest(payload) {
+        const body = JSON.stringify(payload);
+        return new DummyRequest(
+            {
+                method: 'POST',
+                headers: { 'content-length': body.length },
+                url: `/_/backbeat/index/${bucketName}?operation=add`,
+            },
+            body,
+        );
+    }
+
+    beforeEach(() => {
+        endPromise = new Promise(resolve => {
+            resolveEnd = resolve;
+        });
+        response = {
+            setHeader: sinon.stub(),
+            writeHead: sinon.stub(),
+            end: sinon.stub().callsFake((body, encoding, callback) => {
+                resolveEnd();
+                if (callback) {
+                    callback();
+                }
+            }),
+        };
+        sinon.stub(auth.server, 'doAuth').yields(
+            null,
+            new AuthInfo({
+                canonicalID: 'abcdef/lifecycle',
+                accountDisplayName: 'Lifecycle Service Account',
+            }),
+            undefined,
+            undefined,
+            {},
+        );
+        sinon.stub(metadata, 'putBucketIndexes').yields(null);
+    });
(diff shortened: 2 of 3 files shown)


delthas (approved):
Oh wow, good catch 😅  Thanks!

lib/routes/routeBackbeat.js
```diff
@@ -1786,6 +1786,7 @@ const indexEntrySchema = joi.object({
             }),
         )
         .required(),
+    partialFilterExpression: joi.object(),
```
DarkIsDude, line 1789: should we validate it ?

DarkIsDude (approved):

lib/routes/routeBackbeat.js
DarkIsDude: I mean declare the schema ? Can be hard as it's mongo but pretty sure we can find some joi lib to validate mongo ?

tests/unit/routes/routeBackbeat.js
```diff
@@ -1540,3 +1540,101 @@ describe('routeBackbeat authorization', () => {
         });
     });
 });
+
+describe('routeBackbeat index add payload validation', () => {
```
DarkIsDude, line 1544: Missing a test : should reject if the field is not well formatted ?

lib/routes/routeBackbeat.js
francoisferrand: it is validated by mongo during the call, so is it worth duplicating the validation?
- we don't actually _own_ the format, we just pass it through: so not sure we add value.
- it's fine not to own it, since this is an internal API only.

(the alternative would be to define our own "index definition format", and map it to mongo format in cloudserver : in that case we should indeed do validation)

tests/unit/routes/routeBackbeat.js
```diff
@@ -1540,3 +1540,101 @@ describe('routeBackbeat authorization', () => {
...
+            },
+        ];
+        routeBackbeat('127.0.0.1', makeIndexRequest(payload), response, log);
+        void (await endPromise);
+
+        const err = JSON.parse(response.end.getCall(0).args[0]);
+        assert.strictEqual(err.code, 'BadRequest');
+        assert.strictEqual(metadata.putBucketIndexes.called, false);
+    });
+});
```
francoisferrand, line 1656: do we have "integration" test, where we actually call mongo?
would be nice to have these, and thus test that we reject the call if either `keys` or `partialFilterExpression` is not a valid mongo format and that we pass the fields appropriately

tests/unit/routes/routeBackbeat.js
```diff
@@ -1540,3 +1540,101 @@ describe('routeBackbeat authorization', () => {
...
+            undefined,
+            {},
+        );
+        sinon.stub(metadata, 'putBucketIndexes').yields(null);
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
```
francoisferrand, line 1592: missing test where the request fails if backend is not mongo? or is it in some integration tests?

francoisferrand (approved):

tests/unit/routes/routeBackbeat.js
maeldonn: Added

lib/routes/routeBackbeat.js
maeldonn: We only check it's an object, mongo validates the content on the call

tests/unit/routes/routeBackbeat.js
maeldonn: added

tests/unit/routes/routeBackbeat.js
maeldonn: Added

PR #6275, opened 2026-09-08
CLDSRV-992: Match server access log entries instead of asserting an exact count

anurag4DSB:
Backport of CLDSRV-923, from #6211 (`6aa67d9b6`, plus its prettier prerequisite `06a0bfe8e`). The fix landed on development/9.4 and 9.5 only and was never forward-merged down, so the flake is still live on the 9.3 line. hotfix/9.3.13 needs the same pick and has its own PR, #6276, since nothing forward-ports between hotfix lines.

The access-log file is shared by the whole server process and a line is written on the response `close` event, which under keep-alive can fire much later, during another suite's truncate window. A foreign GetObject on `buckettestgetobject` from `test/object/get.js` therefore inflates the count, which is why the victim varied between runs: bucketPut in one, objectDelete in the next. This reproduced independently while working on CLDSRV-992: the same job on #6277, which lacks this fix, failed 2 of 10 stress runs on exactly this assertion, and turned up a third victim (objectGetAttributes) beyond the two named in the ticket.

**Review commit 1 only.** Commit 2 is a pure prettier pass, cherry-picked with its original author, and accounts for most of the line count; the functional change is 83 lines and the resulting file is byte-identical to development/9.4's, green there since July. The reorder relative to upstream is deliberate, so the fix reviews on its own rather than sitting on top of a reformat.

### DIFF
--- a/tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
+++ b/tests/functional/aws-node-sdk/test/serverAccessLogs/testServerAccessLogFile.js
@@ -23,29 +23,76 @@ function truncateLogFileIfExists(filePath) {
     }
 }
 
-async function waitForLogs(filePath, expectedLines, maxRetries, delayMs) {
-    for (let attempt = 0; attempt < maxRetries; attempt++) {
-        const logEntries = fs.readFileSync(filePath, 'utf8');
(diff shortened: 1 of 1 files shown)


tcarmet (approved):

jonathan-gramain (approved):

anurag4DSB:
ping

PR #6280, opened 2026-09-08
CLDSRV-992: Build the uploadPartCopy fixtures once per suite, not per test

anurag4DSB:

NEW FIX. Both `uploadPartCopy` describes in the PartNumber tests rebuild a ~200 MB MPU fixture in a `beforeEach`, though all four of their tests only issue a GetObject. The redundant second build is what exceeds the 40 s mocha timeout. Clears the timeout half of CLDSRV-992 row F8 (run 33647777908).

One functional test file, setup and teardown only. No product code, no assertions touched. development/9.3 needs the same change and has #6281.

The fixture is identical and both tests read exactly what they read before. No timeout raised, no retry added, nothing skipped. That row's separate ServiceUnavailable on "part 8 when ordered MPU" is a different transient and is not addressed here.

Setup moves to `before` and cleanup to `after`, so each describe builds its fixture once. The enclosing describe already owns the bucket in its own `before`/`after`, so the fixture safely outlives the tests.

The cost is arithmetic from the fixture itself: a serial 10x5 MB completeMPU, a 50 MB UploadPartCopy, a 50 MB UploadPart, then a further 5 MB part and second 50 MB copy. Both tests are pure GetObject calls, so the second build did no work they depended on. s3c-ft-tests-v0 stressed 10x on this branch.

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/get.js
+++ b/tests/functional/aws-node-sdk/test/object/get.js
@@ -923,7 +923,7 @@ describe('GET object', () => {
                 const partOneBody = Buffer.concat(bufs, partOneSize);
                 const partTwoBody = Buffer.alloc(partSize, 4);
 
-                beforeEach(done => async.waterfall([
+                before(done => async.waterfall([
                     next => completeMPU(orderedPartNumbers, next),
                     next => createMPUAndPutTwoParts(partTwoBody, next),
                     (uploadId, ETags, next) =>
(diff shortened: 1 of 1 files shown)


anurag4DSB:
Moving back to draft along with #6281 — same reasoning. The change was justified as fixture cost, but the CI logs show the fixture is not slow (the same build succeeds three or four times per run then stalls once), so this halves exposure to an unexplained stall while giving up per-test independence.

These s3c jobs capture no cloudserver log, which is why the stall cannot be diagnosed today. #6283 adds that capture first.

anurag4DSB:
closing this cuz we actually want our tests to be independent, so will come back to this

PR #6281, opened 2026-09-08
CLDSRV-992: Build the uploadPartCopy fixtures once per suite, not per test

anurag4DSB:

NEW FIX. Both `uploadPartCopy` describes in the PartNumber tests rebuild a ~200 MB MPU fixture in a `beforeEach`, though all four of their tests only issue a GetObject. The redundant second build is what exceeds the 40 s mocha timeout. Clears CLDSRV-992 row F4.

One functional test file, setup and teardown only. No product code, no assertions touched. hotfix/9.2.36 needs the same change and has #6280.

The fixture is identical and both tests read exactly what they read before. No timeout raised, no retry added, nothing skipped.

Setup moves to `before` and cleanup to `after`, so each describe builds its fixture once. The enclosing describe already owns the bucket in its own `before`/`after`, so the fixture safely outlives the tests.

The cost is arithmetic from the fixture itself: a serial 10x5 MB completeMPU, a 50 MB UploadPartCopy, a 50 MB UploadPart, then a further 5 MB part and second 50 MB copy. Both tests are pure GetObject calls, so the second build did no work they depended on. s3c-ft-tests-v0-null-compat stressed 10x on this branch.

### DIFF
--- a/tests/functional/aws-node-sdk/test/object/get.js
+++ b/tests/functional/aws-node-sdk/test/object/get.js
@@ -908,7 +908,7 @@ describe('GET object', () => {
                 const partOneBody = Buffer.concat(bufs, partOneSize);
                 const partTwoBody = Buffer.alloc(partSize, 4);
 
-                beforeEach(done => async.waterfall([
+                before(done => async.waterfall([
                     next => completeMPU(orderedPartNumbers, next),
                     next => createMPUAndPutTwoParts(partTwoBody, next),
                     (uploadId, ETags, next) =>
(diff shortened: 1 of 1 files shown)


anurag4DSB:
Moving back to draft — the rationale for this change does not hold up.

I justified it as "the fixture is too expensive for the 40 s budget", but measuring the CI logs shows the fixture is not slow: the tests around it run in 38–657 ms, and in the failing runs the same fixture builds successfully three or four times and then stalls once. That is an intermittent stall, not slow work, so building it once per suite only halves the exposure without explaining the cause — and it gives up per-test independence to do so, which is a bad trade.

Independence here is also load-bearing rather than incidental: sibling tests in the same describe do plain `PutObject` to `objectName` (get.js:813, :831, :851, :880), clobbering the MPU, which is exactly why `completeMPU` is re-run per test.

These jobs are also the ones that capture no cloudserver log, so the stall is currently undiagnosable. #6283 adds that capture; I will come back to this row once a recurrence can name its own cause.

anurag4DSB:
closing this cuz we actually want our tests to be independent, so will come back to this

PR #6282, opened 2026-09-08
CLDSRV-992: Wait for a mongo primary before waiting for the S3 API

anurag4DSB:

NEW FIX. The mongodb-backed jobs wait up to 40 s for port 8000 but nothing waits for MongoDB, and cloudserver's start-up blocks on its metadata backend, so a mongo that is up-but-not-yet-elected appears as the S3 API never binding 8000. The job then dies on "Server did not start in less than 40 seconds" with no clue why. Clears CLDSRV-992 row F5.

CI only, no product or test code. One new helper at the repo root beside `wait_for_local_port.bash`, and four jobs gain the gate: multiple-backend, mongo-v0-ft-tests, mongo-v1-ft-tests, sur-tests.

A healthy run pays nothing; the wait returns as soon as a primary is reported. No test, assertion or timeout is changed, and the existing 8000 wait keeps its 40 s.

Each job waits for an elected primary, `rs.status().myState == 1`, before waiting for 8000. A plain `nc -z 27018` would not do: `mongo-run.sh` starts mongod immediately but backgrounds `rs.initiate()` behind a `sleep 5`, so the port answers at least five seconds before the replica set exists. On timeout the helper dumps `rs.status()`.

**Probable cause, not proven, and I would rather say so.** The failing attempt's server log is gone: artifacts bundles are per run, so the successful re-run overwrote it. What the surviving log establishes is that multiple-backend bound 8000 at t+19.4 s on a *passing* attempt, against a 40 s budget that must also cover image pull, container create and replica-set init; a runner half as fast fails, matching a job that dies at the port wait and passes on re-run. The eval was checked against a real container, which reported `myState: 1`. The change also has diagnostic value independent of whether it is the cause.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -336,6 +336,8 @@ jobs:
       - name: Run multiple backend test
         run: |-
           set -o pipefail;
+          # start-up blocks on mongo, so a slow mongo looks like 8000 never binding
+          bash wait_for_mongo_primary.bash 120
           bash wait_for_local_port.bash 8000 40
           bash wait_for_local_port.bash 81 40
           yarn run multiple_backend_test | tee /tmp/artifacts/${{ github.job }}/tests.log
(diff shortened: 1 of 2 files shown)


anurag4DSB:
ideally cloudserver should not be healthy if mongo id not ready so 8000 will not be available so this logic doesn't make sense to me, I will come back to this.

PR #6283, opened 2026-09-08
CLDSRV-992: Capture cloudserver's log on the s3c functional jobs

anurag4DSB:
The s3c-ft-tests jobs upload sproxyd, bucketd, repd and vault logs but never cloudserver's own, because its output goes to the container's stdout and is discarded. Any failure originating in cloudserver on these jobs is therefore undiagnosable and becomes hypothetical.

Chasing a fresh occurrence of F8's ServiceUnavailable through the artifacts that do exist: nginx logged zero 5xx, sproxyd zero errors, and bucketd zero 503s with its only connection errors at a single startup instant. So cloudserver produced that 503 itself and nothing recorded why. S

The job now dumps cloudserver's container log to its artifacts directory before teardown, while the containers still exist. The step is `if: always()` and ends in `|| true`, so it runs on failure, which is when it matters, and can never itself fail a job; a passing run pays one `docker compose logs` call. If it were wrong we would see an empty or missing s3.log in the very next run's artifacts.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -666,6 +666,15 @@ jobs:
         run: |-
           set -o pipefail;
           yarn run ft_scripts | tee /tmp/artifacts/${{ matrix.job-name }}/ft_scripts.log
+      # cloudserver runs as a container on these jobs and its log is never
+      # captured, so a server-side stall or 503 here cannot be diagnosed after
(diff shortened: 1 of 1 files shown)


anurag4DSB:
claude code hallucination, I checked manually we do collect the logs.

PR #6292, opened 2026-09-14
Server side encryption migrations should not be reported as an object…

SylvainSenechal:
Issue: CLDSRV-990

Taking a look at the ticket :
We have this updateObjectEncryption function setting 
objMD.originOp = 's3:ObjectCreated:Copy';
all the time which is indeed problematic as it sends bucket notifications (although this is debatable, sending a notif is not exactly a P0..), and also can trigger transition (which is more problematic and the reason why this ticket was created https://github.com/scality/backbeat/pull/2830#discussion_r3947338244)

But, upon looking, the function is called from 5 paths :
- objectPut
- objectCopy
- initMpu

- objectGet
- objectHead

The first 3 are writes, which calls updateEncryption with a flag, so that the originOp rewrite path is not even exercised at all (because these 3 writes will have their own logic to set origin op)

Only the 2 reads get/head can make a call that would wrongly set originOp to copy. But even then, this only happens in the even of a sse migration.

So I'm making a change on the originOp to use something that will not trigger bucket notif/transition

2 remarks :
- Could consider sending this pr on a lower cloudserver version
- updateEncryption is not tested at all, neither for this particular fix, or for anytging else at all, I could maybe add a test, didnt do it its too easy to create 500 lines of ai test crap


lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -77,18 +77,17 @@ function updateObjectEncryption(bucket, objMD, objectKey, log, keyArnPrefix, opt
             dataLocator.masterKeyId = `${keyArnPrefix}${dataLocator.masterKeyId}`;
         }
     }
+    // Reformating the key arn is not an operation that should trigger any
+    // oplog driven logic (CRR, bucket notifications).
     // eslint-disable-next-line no-param-reassign
-    objMD.originOp = 's3:ObjectCreated:Copy';
-    // Copy should be tested for 9.5 in INTGR-1038
```
SylvainSenechal, line 82: @BourgoisMickael On this old ticket : https://scality.atlassian.net/browse/INTGR-1038

you wrote a comment "It will trigger new bucket notification, but it will not trigger a replication for CRR"

With this change, it will not trigger bucket notif anymore. But I think you have integration tests that may expect it, so I'm letting you know you may have to update your tests

SylvainSenechal:
Adding Maha as you kinda created the ticket from the discussion on the bb pr, and Mickael which may be affected on the Ring tests

benzekrimaha (approved):

lib/api/apiUtils/bucket/updateEncryption.js
BourgoisMickael: I did that to follow the AWS way of updating the encryption: you do a copy of the object on itself while changing the encryption in metadata.

This is an internal operation but follows the way a user would do.

I can't find that action`UpdateObjectEncryption` in the bucket notification documentation.

SylvainSenechal: Well that's the point 🤔 actionUpdateObjectEncryption is not in the list of notifications, so it wont trigger a notif unlike the existing ObjectCreated

BourgoisMickael: I though we wanted to actually receive a notification (even if it's internal change) as the metadata changes.

Is it how it's done in AWS as well ?

Otherwise I'm fine with it. Should this fix ship in development/9.4 ?

BourgoisMickael (approved):

lib/api/apiUtils/bucket/updateEncryption.js
SylvainSenechal: Im gonna put it on 9.4

I don't know about AWS, but I guess there is no concept of "sse migration" in aws

lib/api/apiUtils/bucket/updateEncryption.js
```diff
@@ -77,18 +77,17 @@ function updateObjectEncryption(bucket, objMD, objectKey, log, keyArnPrefix, opt
             dataLocator.masterKeyId = `${keyArnPrefix}${dataLocator.masterKeyId}`;
         }
     }
+    // Reformating the key arn is not an operation that should trigger any
```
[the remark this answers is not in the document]
SylvainSenechal: Spot it faster next time its at least the third time that I push and I just approved and created the waterfall branch -_-

